PR #8, opened 2018-06-26
force set gpg language to en

realityone:
GnuPG takes the `LANG` environment variable as output language.

If we using the English to expect the output, I prefer set the `LANG` explicitly.


thedataking:
Thanks @realityone, we've made the change on an internal branch and updated the public mirror. Closing. Let us know if you encounter additional issues.

PR #36, opened 2019-01-24
Update the Travis script to run on Linux and OS X

miguelsaldivar:
In this commit, the `.travis.yml` (Travis build script) has been
updated to build and test on both Linux and OS X. To achieve this
the `script` portion of the script, checks for the os type and runs
the correct provisioning script accordingly. Another change is, `matrix`
was added, and this sets the correct environment variables depending on
the os.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ git:
 
 env:
   global:
-    - CXX=clang++
-    - CC=clang
     - USE_CCACHE=1
     - CCACHE_COMPRESS=1
     - CCACHE_CPP2=1
@@ -64,17 +62,17 @@ install:
 script:
   - |
     case "$CACHE_NAME" in
-      LINUX_JOB|OSX_JOB)
+      LINUX_DEV_BUILD|OSX_DEV_BUILD)
         ccache -s
         python3 ./scripts/build_translator.py --debug || travis_terminate 1
         ccache -s
         python3 ./scripts/test_translator.py --debug ./tests || travis_terminate 1
         ;;
-      LINUX_JOB2)
+      LINUX_FAST_BUILD)
         cargo build
         python3 ./scripts/test_translator.py --debug ./tests || travis_terminate 1
         ;;
-      OSX_JOB2)
+      OSX_FAST_BUILD)
         cargo build
         python3 ./scripts/test_translator.py --debug ./tests || travis_terminate 1
         ;;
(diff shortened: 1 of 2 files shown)


miguelsaldivar:
@thedataking that's what I seen from various other sources (a lot of the `rust-lang` projects) mix the two, so I followed suit. Although if that's not okay, the `matrix` style will be the one to use.

.travis.yml
```diff
@@ -1,48 +1,78 @@
-os: osx
+language: generic
 # note: xcode9.4/macOS 10.13 is the default for macOS
```
thedataking: This comment should go above line 21 now that you are using the build matrix feature.

.travis.yml
```diff
@@ -1,48 +1,78 @@
...
     - $TRAVIS_BUILD_DIR/ast-importer/target.travis
     - $TRAVIS_BUILD_DIR/cross-checks/rust-checks/config/target
     - $TRAVIS_BUILD_DIR/cross-checks/rust-checks/rustc-plugin/target
     - $TRAVIS_BUILD_DIR/cross-checks/rust-checks/runtime/target
     - $TRAVIS_BUILD_DIR/cross-checks/rust-checks/derive-macros/target
     - $TRAVIS_BUILD_DIR/rust-refactor/target
 
+
 install:
 #   - HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade python
```
thedataking: This comment should go in the "osx" arm of the conditional below.

thedataking:
@miguelsaldivar I had two minor comments, other than that, this looks ready to merge!

.travis.yml
miguelsaldivar: Fixed :+1:

.travis.yml
```diff
@@ -1,48 +1,78 @@
...
     - $TRAVIS_BUILD_DIR/rust-refactor/target
 
+
 install:
-#   - HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade python
   - /usr/bin/env python3 --version
-  - ./scripts/provision_mac.sh
+  - |
+    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      # - HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade python
```
thedataking: Should be `# HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade python`. Otherwise, uncommenting the line would not work.

miguelsaldivar: Yup, that makes sense. :+1:

PR #37, opened 2019-01-31
[WIP]Add `ccache` to the path, for the Linux dev build

miguelsaldivar:


thedataking:
@miguelsaldivar not sure why you kept `dist: xenial` on the top level since that obviously only makes sense for Linux hosts. Is there a way to move it into the matrix?

PR #42, opened 2019-03-07
Add pkg-config to the list of dependencies

Kixunil:
`pkg-config` is required for building openssl, so it was added to the list of dependencies.

I'm unsure if modifying Arch Linux or brew dependencies is needed and if yes, then how the package is called there.

Also, what's the purpose of OpenSSL in a tool for translating C to Rust? Such dependency seems rather weird to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ C2Rust requires LLVM 6 or 7 and its corresponding libraries and clang compiler. 
 
 - **Ubuntu 16.04, 18.04 & 18.10:**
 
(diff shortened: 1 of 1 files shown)


rinon:
Good catch on `pkg-config`, thanks. It's part of `base-devel` on Arch, so we already have it covered there.

OpenSSL is a dependency of `cargo` via the `crates-io` crate. Our refactoring tool depends on the `cargo` crate to figure out how to compile a crate (and thus parse it for our refactoring purposes).

PR #48, opened 2019-04-08
LLVM 8 Support

rinon:
LLVM 8 is out and a few APIs have changed.


rinon:
Need to make sure we have a buildbot that builds with LLVM 8.

Issue #63, opened 2019-04-14
Allow whitelisting of flexible array members


thedataking:
We have two options here since we always need the programmer to make a determination:
1. update the sources to use C99-style flexible array members,
2. mark/whitelist non-C99 flexible array members in a configuration file.

The second option seems most appropriate in two cases:
- when tracking upstream C sources,
- when the non-C99 flexible array members are in system headers.

thedataking:
We also have the option to assume that one-sized arrays at the end of structures are flex array members without breaking the translation. After some discussion, this was the approach we decided to take – plus adding a strict-flex-array-member mode that errors on one-sized potential.

rinon:
I don't see a good reason to have a strict flex array member mode that errors on 1 sized arrays in the final slot. Why don't we just conservatively handle them as potentially flexible and emit slightly less readable code for that case?

rinon:
Resolved by 1abf09f99fad528e78f53f10807c4c6411524642

PR #68, opened 2019-04-17
Conservatively handle potential flexible array members

rinon:
We don't need to break on flexible array member fields if we handle these fields with `.offset()` rather than array indexing.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -1720,10 +1720,12 @@ class TranslateASTVisitor final
 
         if (auto CA = dyn_cast_or_null<ConstantArrayType>(
                 D->getType().getTypePtr())) {
-            // If the array has a size of 1, and struct field count is
-            // greater than 1, and if the (struct field count - 1) is equal to
-            // the index, it is most likely a flexible array.
-            return (CA->getSize() == 1 && FieldCount > 1 &&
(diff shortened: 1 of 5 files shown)


rinon:
Resolves #63

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -1709,24 +1713,20 @@ class TranslateASTVisitor final
...
-            // the index, it is most likely a flexible array.
-            if (CA->getSize() == 1 && FieldCount > 1 &&
-                FieldCount - 1 == D->getFieldIndex() &&
-                !warnedFlexibleArrayDecls.count(D)) {
-                // Insert the Decl into the set, if it has not been warned about
-                // yet.
-                warnedFlexibleArrayDecls.insert(D);
-                return true;
-            }
+            // If the array has a size of 1, struct field count is greater
```
TheDan64: Can't the array size be 0 as well?

rinon: Yes, that is a GNU extension. We don't need to warn about that because people don't need to fix their code, though. We handle it at https://github.com/immunant/c2rust/pull/68/files#diff-4c1124a9b33d10318140e1ae99bd3189R138

I actually want to get rid of the warning in AstExporter entirely, but am waiting on a response from @perl on the bug before I do.

PR #73, opened 2019-04-18
Fix papercuts preventing translation of curl

thedataking:


thedataking:
Blocked on #85.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1010,6 +1010,35 @@ impl<'c> Translation<'c> {
         }
     }
 
+    /// Determines whether we need to use the `const_transmute` feature.
+    /// Modeled after `static_initializer_is_unsafe`.
+    fn static_initializer_needs_const_transmute(&self, expr_id: Option<CExprId>) -> bool {
```
rinon: This isn't necessary after some changes made in #87, so I would like to merge that in first.

PR #77, opened 2019-04-20
Make build.rs template less Linux specific

wezm:
There are many platforms that are unix-like, or UNIX derived that aren't Linux, such as the [BSDs](https://runbsd.info/) and [Illumos](https://www.illumos.org/). Frequently when encountering crate build issues on these platforms it's because of overly specific conditional compilation. In the interests of avoiding this in generated crates I propose this change.


rinon:
Good point, thanks for the PR.

PR #92, opened 2019-05-06
Update support for C variadics

thedataking:
Reflect anticipated changes in Rust's implementation of C-compatible variadics.


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -130,6 +130,25 @@ impl TypedAstContext {
         }
     }
 
+    /// Predicate for (pointers to) structs that are binary compatible with C's `va_list`.
```
ahomescu: Does this return true for both `struct __va_list_tag` and `struct __va_list_tag*`? If so, I'm not sure that's correct.

ahomescu:
@thedataking  Should there be another commit that converts `va_copy` calls to `clone`?

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -256,6 +256,33 @@ impl TypedAstContext {
...
+                } else {
+                    false
+                }
+            },
+            _ => false
+        }
+    }
+
+    /// Predicate for pointers to structs that are binary compatible with C's `va_list`.
+    pub fn is_pointer_to_va_list(&self, typ: CTypeId) -> bool {
```
ahomescu: It's a bit more complex than this, depending on the architecture:

- On struct-architectures, e.g., x86_64, there's a structure called `struct __va_list_tag` and a `typedef struct __va_list_tag __builtin_va_list[1];` definition which decays to `struct __va_list_tag*` when passed to callees.

- On pointer-architectures like x86 and ARM, `__va_list_tag` doesn't exist, and the other one is just `typedef char* __builtin_va_list;` (or `void*` in some cases). I'm not sure if the type you'll see in the transpiler is `char*` or `__builtin_va_list`.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3411,7 +3394,35 @@ impl<'c> Translation<'c> {
...
+                    // when calling a function expecting to receive an object that is binary
+                    // compatible with C's `va_list`.
+                    if let Some(CTypeKind::Function(_, params, _, _, _)) = fn_ty {
+                        // look for parameters of type `va_list`
+                        let positions: Vec<usize> = params
+                            .iter()
+                            .enumerate()
+                            .filter_map(|(pos, param)| {
+                                if self.ast_context.is_pointer_to_va_list(param.ctype)
+                                { Some(pos) } else { None }
```
ahomescu: I'm not sure this is how `rust-fmt` would indent it. You could try running it on this file, but that's probably going to produce a ton of other changes.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -270,33 +297,33 @@ impl TypedAstContext {
...
-            // Allow a decayed reference to the array to count as a va_list
-            CTypeKind::Pointer(pointee_qty) => {
-                return self.is_va_list(pointee_qty.ctype);
-            }
-
-            _ => {}
-        }
-        false
-    }
+//    pub fn is_va_list(&self, typ: CTypeId) -> bool {
```
ahomescu: Do you still need this code? If not, just delete it.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -210,72 +177,16 @@ impl<'c> Translation<'c> {
                         VaPart::Start(va_id) | VaPart::End(va_id) => va_id,
                         VaPart::Copy(dst_va_id, _src_va_id) => dst_va_id,
```
ahomescu: Could just use `_` instead of `_src_va_id`.

PR #93, opened 2019-05-07
Lua scripting ast visitor for refactoring tool

TheDan64:
Adds an initial lua ast visitor pattern. Also adds a lua script to clean up function params and locals when they are erroneously marked as mutable and not prefixed with an underscore when unused.

I'm currently working on updating that script the use the more generic visitor pattern than what's currently in `cleanup_param_locals.lua`.

These changes try to start deprecating the interner in favor of just using lua tables/objects.


c2rust-refactor/src/scripting.rs
```diff
@@ -135,16 +151,255 @@ impl<'lua> IntoLuaAst<'lua> for P<ast::Expr> {
...
+                    ast.set("path", path.into_lua_ast(ctx, lua_ctx)?)?;
+                    ast.set("args", iter_to_lua_array(args?.into_iter(), lua_ctx)?)?;
+                },
+                ExprKind::MethodCall(_, _) => {
+                    ast.set("kind", "MethodCall")?;
+
+                },
+                ExprKind::Tup(_) => {
+                    ast.set("kind", "Tup")?;
+
```
ahomescu: I think at some point in the future we'll want to enable access to the tuple fields. Maybe add a `FIXME` or `TODO` comment here for now (unless you're still implementing this)?

c2rust-refactor/src/scripting.rs
```diff
@@ -135,16 +151,255 @@ impl<'lua> IntoLuaAst<'lua> for P<ast::Expr> {
...
+                    ast.set("rhs", rhs.into_lua_ast(ctx, lua_ctx)?)?;
+                },
+                ExprKind::Unary(op, expr) => {
+                    ast.set("kind", "Unary")?;
+                    ast.set("op", syntax::ast::UnOp::to_string(op))?;
+                    ast.set("expr", expr.into_lua_ast(ctx, lua_ctx)?)?;
+                },
+                ExprKind::Cast(_, _) => {
+                    ast.set("kind", "Cast")?;
+
```
ahomescu: See above: add `TODO` comment.

c2rust-refactor/src/scripting.rs
```diff
@@ -495,17 +750,13 @@ impl<'a, 'tcx> TransformCtxt<'a, 'tcx> {
...
         match self.clone_ast(node) {
-            RustAstNode::Stmt(s) => s.into_lua_ast(self, &mut ast)?,
-            RustAstNode::Expr(e) => e.into_lua_ast(self, &mut ast)?,
-            _ => {
-                return Err(LuaError::external(
-                    "get_ast not implemented for this type of RustAstNode",
-                ));
-            }
+            RustAstNode::Stmt(s) => s.into_lua_ast(self, lua_ctx),
+            RustAstNode::Expr(e) => e.into_lua_ast(self, lua_ctx),
```
ahomescu: You could match both at the same time using `RustAstNode::Stmt(e) | RustAstNode::Expr(e)`.

c2rust-refactor/src/scripting.rs
```diff
@@ -587,5 +838,64 @@ impl<'a, 'tcx> UserData for TransformCtxt<'a, 'tcx> {
...
+
+        /// Visits all function like items
+        // @function visit_fn_like
+        // @tparam LuaAstNode krate Crate in its entirety
+        // @tparam function() callback Function called for each function like item.
+        methods.add_method_mut("visit_fn_like", |lua_ctx, this, (krate, callback): (LuaAstNode, LuaFunction)| {
+            let mut krate = ast::Crate::try_from(this.remove_ast(krate)).expect("Did not find crate input");
+            let mut found_err = Ok(());
+
+            mut_visit_fns(&mut krate, |fn_like| {
```
ahomescu: I don't like that the names are different: `visit_fn_like` in Lua and `visit_fns`/`mut_visit_fns` in Rust. Ideally, the names and perhaps signatures would be identical between the two languages, if the Lua version is just a thin wrapper over the Rust one.

c2rust-refactor/src/scripting/ast_visitor.rs
```diff
@@ -0,0 +1,493 @@
...
+                    ast.set("decl", decl.into_lua_ast(ctx, lua_ctx)?)?;
+                    ast.set("block", block.into_lua_ast(ctx, lua_ctx)?)?;
+                },
+                ItemKind::Struct(_variant_data, _generics) => {
+                    ast.set("kind", "Struct")?;
+                },
+                ItemKind::Impl(..) => {
+                    ast.set("kind", "Impl")?;
+                }
+                ref e => warn!("Item IntoLuaAst kind: {:?}", e),
```
ahomescu: Maybe use `unimplemented!` here so we get an error and we know why.

c2rust-refactor/src/scripting.rs
rinon: I think this is getting moved to the `Visitor` interface, right @TheDan64? If so, it will be a method that isn't part of the libsyntax `Visitor` trait, since the concept of `fn_like`s is something we have for the refactorer. I think the name is fine if it is moved.

TheDan64: That's the general idea, yes. But I don't think I'll be able to use the existing visitor trait as I need to walk the Lua table AST rather than the rust AST

c2rust-refactor/src/scripting/ast_visitor.rs
TheDan64: I did this originally but it became annoying because it would often panic on unimplemented nodes there weren't even relevant to what I was scripting

c2rust-refactor/src/scripting/utils.rs
```diff
@@ -0,0 +1,16 @@
+use rlua::prelude::{LuaContext, LuaResult, LuaTable, ToLua};
+
+pub(crate) fn iter_to_lua_array<'lua, I, T: 'lua>(iter: I, lua_ctx: LuaContext<'lua>) -> LuaResult<LuaTable<'lua>>
```
ahomescu: I wonder if it would be more ergonomic to make this into a mini-trait with a method `into_lua_array` implemented for any `Iterator`.

TheDan64: It's an interesting idea, but with a trait you still have to import it and call it which is pretty similar to this function, no? I guess it just doesn't seem any more or less ergonomic?

c2rust-refactor/src/scripting.rs
TheDan64: Are you thinking of `Semi`? `Stmt` won't work with `Expr`

ahomescu: Oh right, I was thinking of `Semi`.

c2rust-refactor/src/scripting/utils.rs
ahomescu: Yeah a trait would require an additional import.

ahomescu:
The PR generally looks good to me.

TheDan64:
@rinon Any final thoughts on these changes? I'm pretty happy with it at this point.

c2rust-refactor/scripts/reconstruct_for_range.lua
```diff
@@ -37,7 +37,7 @@ $'label:?Ident: while $cond:Expr {
                   print(incr)
                   if (incr.kind == "Semi" or
                       incr.kind == "Expr") then
-                     incr = incr.expr
+                     incr = incr.expr_old
```
rinon: Can we get rid of `expr_old` and use the new style here?

c2rust-refactor/src/scripting.rs
```diff
@@ -495,17 +755,13 @@ impl<'a, 'tcx> TransformCtxt<'a, 'tcx> {
...
-                    "get_ast not implemented for this type of RustAstNode",
-                ));
-            }
+            RustAstNode::Stmt(s) => s.into_lua_ast(self, lua_ctx),
+            RustAstNode::Expr(e) => e.into_lua_ast(self, lua_ctx),
+            _ => Err(LuaError::external(
+                "get_ast not implemented for this type of RustAstNode",
+            )),
         }
-        Ok(ast)
```
rinon: Good catch on removing this :+1:

c2rust-refactor/src/scripting.rs
```diff
@@ -587,5 +843,18 @@ impl<'a, 'tcx> UserData for TransformCtxt<'a, 'tcx> {
         methods.add_method("get_ast", |lua_ctx, this, node: LuaAstNode| {
             this.get_lua_ast(lua_ctx, node)
         });
+
+        /// Visits an entire crate via a lua object's methods
+        // @function run_visitor
+        // @tparam function() callback Function called for each function like item.
+        methods.add_method_mut("run_visitor", |lua_ctx, this, (visitor_obj, krate): (LuaTable, LuaAstNode)| {
```
rinon: I would expect this to be called `visit_crate` as it is analogous to `LuaAstVisitor::visit_crate`.

c2rust-refactor/src/scripting/utils.rs
rinon: Isn't this basically https://docs.rs/rlua/0.16.3/rlua/struct.Context.html#method.create_sequence_from?

c2rust-refactor/src/scripting/ast_visitor.rs
```diff
@@ -0,0 +1,655 @@
...
+                return Ok(());
+            }
+        }
+    };
+}
+
+pub(crate) struct LuaAstVisitor<'a, 'lua, 'tctx: 'a> {
+    ctx: &'a TransformCtxt<'a, 'tctx>,
+    lua_ctx: LuaContext<'lua>,
+    visitor_obj: LuaTable<'lua>
```
rinon:
```suggestion
    visitor: LuaTable<'lua>
```

c2rust-refactor/src/scripting/ast_visitor.rs
```diff
@@ -0,0 +1,655 @@
...
+
+        self.visit_module(lua_krate.get("module")?)?;
+        self.finish()?;
+
+        krate.merge_lua_ast(lua_krate)?;
+
+        Ok(())
+    }
+
+    pub fn visit_module(&self, module: LuaTable<'lua>) -> LuaResult<()> {
```
rinon:
```suggestion
    pub fn visit_mod(&self, module: LuaTable<'lua>) -> LuaResult<()> {
```
I'd like to match the API in the libsyntax `Visitor` when it makes sense.

c2rust-refactor/src/scripting/ast_visitor.rs
```diff
@@ -0,0 +1,655 @@
...
+    }
+
+    pub fn finish(&self) -> LuaResult<()> {
+        call_lua_visitor_method!(self.visitor_obj,finish());
+
+        Ok(())
+    }
+}
+
+impl<'lua> IntoLuaAst<'lua> for FnLike {
```
rinon: I think we should move all the `IntoLuaAst` impls into their own module so they're not spread between scripting.rs and ast_visitor.rs

c2rust-refactor/src/scripting/ast_visitor.rs
```diff
@@ -0,0 +1,655 @@
...
+
+    pub fn visit_crate(&self, krate: &mut Crate) -> LuaResult<()> {
+        let lua_krate = krate.clone().into_lua_ast(self.ctx, self.lua_ctx)?;
+
+        call_lua_visitor_method!(self.visitor_obj,visit_crate(lua_krate));
+
+        self.visit_module(lua_krate.get("module")?)?;
+        self.finish()?;
+
+        krate.merge_lua_ast(lua_krate)?;
```
rinon: Is it correct that `merge_lua_ast` is only called when visiting the full crate, and not on any child nodes? That seems odd to me. What if I only visit a single function? I think the recursion should be removed from `merge_lua_ast` and put into the (Mut)Visitor functions.

c2rust-refactor/src/scripting/ast_visitor.rs
```diff
@@ -0,0 +1,655 @@
...
+            },
+            ref e => unimplemented!("IntoLuaAst for {:?}", e),
+        }
+
+        Ok(ast)
+    }
+}
+
+pub(crate) trait MergeLuaAst {
+    fn merge_lua_ast<'lua>(self, table: LuaTable<'lua>) -> LuaResult<()>;
```
rinon: Can we use `&mut self` here instead? The function shouldn't discard `self` (which is what this signature implies).

c2rust-refactor/scripts/reconstruct_for_range.lua
TheDan64: Yes - I suppose I didn't want to touch this at the time since I didn't know how the script works and since it expects an interned node, I wasn't sure how to fix it. But I can take the time to look at it I suppose

c2rust-refactor/src/scripting/utils.rs
TheDan64: Nice find! Didn't see this.

TheDan64: I wonder why my version is much shorter and straightforward than their impl though. huh.

rinon: Their's is probably a lot faster.

TheDan64: Goood point

c2rust-refactor/src/scripting/ast_visitor.rs
TheDan64: Good idea - do you think we should just have one `traits.rs` with scripting traits and impls? Or split them up into `traits/into_lua_ast.rs` and friends?

rinon: Just call it `scripting/into_lua_ast.rs`

c2rust-refactor/src/scripting/ast_visitor.rs
TheDan64: > Is it correct that merge_lua_ast is only called when visiting the full crate, and not on any child nodes? That seems odd to me. What if I only visit a single function?

Interesting point, I suppose I set it up such that `visit_crate` is the main entry point. Do you think it makes sense to leave merging to the caller of `visit_crate` (and friends) then?

> I think the recursion should be removed from merge_lua_ast and put into the (Mut)Visitor functions.

One of the big problems I found was that the `Mut`/`Visitor` traits don't return results, so our error handling gets completely borked. The lua code returns a ton of results and so need to be unwrapped or saved to self which seemed a really unwieldy.

c2rust-refactor/src/scripting/ast_visitor.rs
TheDan64: I thought there was a reason I did it this way and impl'd for `&mut T`, but I can't remember why. I'll try and switch it back and see if I run into that issue again

c2rust-refactor/src/scripting/ast_visitor.rs
rinon: It would be nice to allow visiting sub-crate AST nodes and their children.

AFAICT you're not actually using the MutVisitor trait? I meant your `LuaAstVisitor` methods.

PR #102, opened 2019-06-04
Moved bitfields derive code into separate crate for re-export

TheDan64:
Now the `BoolOrInt` trait is only generated once, not number of fields times. And similarly I moved `generate_calc_total_bit_size_fn` which generates code to do a sizeof, to the trait rather than being codegened. This should hopefully give `quote` a lot less work to do.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -203,9 +203,16 @@ dependencies = [
 
 [[package]]
 name = "c2rust-bitfields"
+version = "0.1.1"
+dependencies = [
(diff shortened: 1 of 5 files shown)


rinon:
@TheDan64 as far as I'm concerned this is ready to merge if you're happy with it. We should probably release new crates for bitfields and bitfields-derive.

TheDan64:
Sounds good, I can take care of the releases

PR #115, opened 2019-06-20
Add new LuaAstNode and refactor Lua interface to use it

rinon:


c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -0,0 +1,362 @@
...
+        methods.add_method("get_node", |lua_ctx, this, ()| {
+            match this.0.borrow().node.clone() {
+                ItemKind::Use(e) => Ok(e.to_lua(lua_ctx)),
+                node => Err(Error::external(format!("Item node {:?} not implemented yet", node))),
+            }
+        });
+    }
+}
+
+unsafe impl Send for LuaAstNode<P<ForeignItem>> {}
```
TheDan64: Is this a good idea? I take it `ForeignItem` isn't send because of a span?

c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -0,0 +1,362 @@
...
+        methods.add_method("get_kind", |_lua_ctx, this, ()| {
+            Ok(this.0.borrow().node.ast_name())
+        });
+
+        methods.add_method("get_id", |lua_ctx, this, ()| {
+            Ok(this.0.borrow().id.to_lua(lua_ctx))
+        });
+
+        methods.add_method("get_ident", |_lua_ctx, this, ()| {
+            Ok(this.0.borrow().ident.to_string())
```
TheDan64: This will reallocate both in rust and then in lua, I think we can avoid it on the rust side by doing:
```suggestion
            Ok(this.0.borrow().ident.name.as_str().get())
```

c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -0,0 +1,362 @@
...
+        });
+
+        methods.add_method("set_ident", |_lua_ctx, this, ident: String| {
```
TheDan64: Another possible optimization: use `ident: LuaString` and then `Ident::from_str(&ident.to_str()?);` to reference the lua string directly

c2rust-refactor/src/scripting/to_lua_ast_node.rs
rinon: Lots of things aren't `Send` because of `Span`. I gave up on thread-safety because the API is too unwieldily otherwise. `LuaAstNode` is not thread-safe and we cannot spawn new threads in Lua. I need to document this with a big warning in the Lua API docs.

c2rust-refactor/src/scripting/to_lua_ast_node.rs
rinon: Good point, fixed now.

PR #116, opened 2019-06-20
Add linkage attribute to (non-gnu89) inline functions

thedataking:
This change was prompted by a linker error in bzip2. Using the [linkage](https://github.com/rust-lang/rust/issues/29603) feature and adding the `#[linkage = "external"]` attribute to inlined Rust functions (translated from non-gnu89 inlined extern functions) avoids the linker issue. This due to the way `rustc` handles functions with explicit linkage (see [mono.rs](https://github.com/rust-lang/rust/blob/9f22ddf6bb7001b884006b9528f8fc8c5b94ae8f/src/librustc/mir/mono.rs#L90)).

After additional testing, it does not seem necessary to add a `#[linkage = "internal"]` attribute on static inlined functions.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2040,10 +2040,9 @@ impl<'c> Translation<'c> {
                     if is_global && is_extern && !attrs.contains(&c_ast::Attribute::GnuInline) {
                         // ensures that public inlined rust function can be used in other modules
                         mk_ = mk_.single_attr("linkage = \"external\"");
-                    } else {
-                        // ensures that non-public inlined rust functions do not clash btw. modules
-                        mk_ = mk_.single_attr("linkage = \"internal\"");
                     }
+                    // NOTE: it does not seem necessary to have an else branch here that
+                    // specifies internal linkage in all other cases due to name mangling by rustc.
                 }
 
                 Ok(ConvertedDecl::Item(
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -2022,6 +2022,27 @@ impl<'c> Translation<'c> {
...
+                    //   not externally visible and does not prevent other translation units from
+                    //   defining the same function. This makes the inline keyword an alternative to
+                    //   static for defining functions inside header files, which may be included in
+                    //   multiple translation units of the same program.
+                    // * always_inline implies inline -
+                    //   https://gcc.gnu.org/ml/gcc-help/2007-01/msg00051.html
+                    //   even if the `inline` keyword isn't present
+                    // * gnu_inline instead applies gnu89 rules. extern inline will not emit an
+                    //   externally visible function.
+                    self.use_feature("linkage");
```
TheDan64: Shouldn't this feature be contained in the following if stmt?

scripts/test_translator.py
```diff
@@ -343,7 +343,15 @@ def run(self) -> List[TestOutcome]:
...
-        rust_file_builder.add_features(["libc", "extern_types", "simd_ffi", "stdsimd", "const_transmute", "nll", "custom_attribute"])
+        rust_file_builder.add_features([
+            "libc",
+            "extern_types",
+            "simd_ffi",
+            "stdsimd",
+            "const_transmute",
+            "nll",
+            "custom_attribute",
+            "linkage"])
```
TheDan64: Nit:
```suggestion
            "linkage",
        ])
```

PR #117, opened 2019-06-24
Add shell.nix to pull in environment

boj:
Allows a `nix` user to pull together an ad-hoc environment to build everything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,10 @@ C2Rust requires LLVM 6, 7, or 8 with its corresponding clang compiler and librar
 - **Arch Linux:**
 
         pacman -S base-devel llvm clang cmake openssl
+        
+- **NixOS / nix:**
 
+        nix-shell
 
 - **OS X:** XCode command-line tools and recent LLVM (we recommend the Homebrew version) are required.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -33,6 +33,12 @@ C2Rust requires LLVM 6, 7, or 8 with its corresponding clang compiler and librar
 - **Arch Linux:**
 
         pacman -S base-devel llvm clang cmake openssl
+        
+- **NixOS / nix:**
+
+        nix-shell
+        rustup toolchain install nightly
```
rinon: We depend on a specific nightly (currently nightly-2019-04-12), which rustup should automatically install if you cargo build in the project directory because of the `rust-toolchain` file. You shouldn't need these rustup comands?

rinon:
Thanks for the nix support, looks great. Re: the documentation `rustup` commands, could you test out whether the build works without those?

README.md
boj: Ok that makes sense. I am a novice when it comes to rust.

boj:
@rinon It works fine without. I have amended the commit.

rinon:
Thanks, @boj!

leo60228:
Shouldn't `$LIBCLANG_PATH` be `libclang.so`, not the directory that contains it?

PR #118, opened 2019-06-24
Just ask rustc where its libraries are

jameysharp:
Instead of trying to duplicate complicated and non-portable path-construction logic, let the `rustc` which rustup picks for us just tell us what we wanted to know.

Assuming this simplification works for everyone, then we can delete a bunch of other code from `scripts/common.py`, and remove the `distro` dependency. For ease of review, though, this is just the minimal change, sans all those deletions.

### DIFF
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -206,19 +206,19 @@ def get_rust_toolchain_binpath() -> str:
 
 def _get_rust_toolchain_path(dirtype: str) -> str:
     """
-    Ask rustc for the correct path to its {lib,bin} directory. 
+    Ask rustc for the correct path to its {lib,bin} directory.
     """
 
-    if 'RUSTUP_HOME' in pb.local.env:
-        home = pb.local.env['RUSTUP_HOME']
-    else:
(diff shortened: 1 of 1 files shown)


rinon:
Thanks for the cleanup, good point. Looks like the script isn't running in the source tree, so there's no `rust-toolchain` to specify the right nightly version. We should specify the rustc version explicitly.

jameysharp:
Huh, I don't understand all the "error while loading shared library" messages from the CI builds. Is that what you mean about not picking up the right nightly Rust build?

I actually wrote this patch because I'm not using rustup at all; instead I use a combination of https://direnv.net and Nix to get the right versions of all of a project's dependencies when I `cd` into that project. So `common.py` couldn't figure out what distro I'm using, and then once I fixed that, it still couldn't find the right path. So I'm not set up to test locally in an environment that's much like what anyone else is using; I was really hoping this patch would completely paper over that difference.

jameysharp:
The patch which @thedataking merged did fix this issue for me, but I've forced-pushed a new patch to this PR that builds on his but is simpler and still seems to work. (No results from Azure at the moment, but at least the Travis builds passed.)

PR #120, opened 2019-06-26
Don't rely on distro detection in common.on_linux

jameysharp:
Instead, just ask the platform module if this is Linux, regardless of
which distribution of Linux it might be.


thedataking:
Much appreciated!

PR #130, opened 2019-07-17
Lua script for raw ptr -> ref/slice ref promotion

TheDan64:


c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -295,12 +297,87 @@ impl UserData for LuaAstNode<P<Expr>> {
 // thread that did not acquire it.
 // @type TyAstNode
 unsafe impl Send for LuaAstNode<P<Ty>> {}
-impl UserData for LuaAstNode<P<Ty>> {}
+impl UserData for LuaAstNode<P<Ty>> {
+    fn add_methods<'lua, M: UserDataMethods<'lua, Self>>(methods: &mut M) {
+        methods.add_method("get_kind", |lua_ctx, this, ()| {
+            match &this.0.borrow().node {
```
rinon: This should use the `AstName` trait instead of a match here if `TyKind` implements it (see `ast_manip/ast_names.rs`).

c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -462,3 +539,83 @@ impl ToLuaExt for AstNode {
...
+/// FnLike AST node handle
+//
+// This object is NOT thread-safe. Do not use an object of this class from a
+// thread that did not acquire it.
+// @type FnLikeAstNode
+unsafe impl Send for LuaAstNode<P<FnLike>> {}
+impl UserData for LuaAstNode<FnLike> {
+    fn add_methods<'lua, M: UserDataMethods<'lua, Self>>(methods: &mut M) {
+        methods.add_method("get_kind", |lua_ctx, this, ()| {
+            Ok(this.0.borrow().kind.to_lua(lua_ctx))
```
rinon: Just return the name with `AstName`, as noted above.

TheDan64: I don't think this is supported for `FnLike` since that's a custom struct of ours. Should I give it a manual impl?

rinon: Hm, good point. Yeah, a manual impl is more generic, and can just delegate to `AstName` for the contained type.

c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -286,15 +375,211 @@ unsafe impl Send for LuaAstNode<P<Expr>> {}
...
+
+        methods.add_method("get_lit", |_lua_ctx, this, ()| {
+            if let ExprKind::Lit(lit) = &this.borrow().node {
+                return Ok(Some(LuaAstNode::new(lit.clone())))
+            }
+
+            Ok(None)
+        });
+
+        methods.add_method("to_lit", |_lua_ctx, this, lit: LuaAstNode<Lit>| {
```
rinon: I’m not sure about all these `to_*` methods. Why not use a single `set` method that just replaces the node?

TheDan64: How would the set method know what what kind of node you're creating?

IE for lit you need a literal, for binary you need an op and two exprs. A generic set method would, I think, only work if it takes an expr and replaces self with it? But then you'd still need to create that expr somehow?

c2rust-refactor/src/scripting.rs
```diff
@@ -478,6 +489,31 @@ impl<'a, 'tcx> UserData for TransformCtxt<'a, 'tcx> {
...
+            let def_id = match this.cx.try_resolve_ty(&ty.borrow()) {
+                Some(id) => id,
+                None => return Ok(None),
+            };
+
+           Ok(this.cx.hir_map().as_local_hir_id(def_id).map(|id| LuaHirId(id)))
+        });
+
+        methods.add_method(
+            "get_nodeid_hirid",
```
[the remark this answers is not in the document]
TheDan64: What about `nodeid_to_hirid`? It doesn't actually take a whole node but an id, so the compiler fn is kinda misnamed

c2rust-refactor/src/scripting.rs
```diff
@@ -601,5 +637,151 @@ impl<'a, 'tcx> UserData for TransformCtxt<'a, 'tcx> {
...
+                id: DUMMY_NODE_ID,
+                node: ExprKind::Lit(lit),
+                span: DUMMY_SP,
+                attrs: ThinVec::new(),
+            });
+
+            Ok(LuaAstNode::new(expr))
+        });
+
+        methods.add_method("int_lit_expr", |_lua_ctx, _this, (int, _suffix): (i64, Option<LuaString>)| {
```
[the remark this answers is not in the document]
TheDan64: Good catch, not sure how that happened.

TheDan64:
> This is looking good so far. A bit of a higher level comment: I want to eventually auto generate most of the LuaAstNode API from the AST types themselves, so we don’t have to keep implementing all that by hand. That’s got to be a pain for you right now. How is working in Lua for this going? Do you think it’s worthwhile?

I think it'll be worthwhile once there's a full API. I remember at one point I (somehow) managed to write nothing but lua for three hours straight and it felt pretty great and productive. But most of the time I'm have to go back and forth between rust and lua.

I do think we'll be able to autogenerate much of it. But there will definitely be functions we can't - ie fns for perf optimizations, fns that need to interact with the refactoring ctxt, and convenience functions (they're very much needed when working with, ie, generics which are very cumbersome)

And I do think we should integrate penlight and/or other libs at some point. They haven't been entirely required so far so I haven't done it yet. But for example there was one point yesterday where I wanted a set of unique values, but needed a stable iteration order. Lua doesn't have that, so I left it as an array and iterated to check for duplicates (luckily, I don't think that particular list will ever be large, but still).

c2rust-refactor/src/scripting/to_lua_ast_node.rs
rinon: Ok, I get what you're saying. Basically these are just `ExprKind` constructors. I guess it's fine for them to be on `Expr` then.

c2rust-refactor/tests/upgrade_ptr_to_ref/new.rs
```diff
@@ -0,0 +1,206 @@
...
+
+unsafe fn destroy_buf(sd: &mut SizedData) {
+    if (sd).buf.is_none() {
+        return;
+    }
+    (sd).buf.take();
+    (sd).buf = None;
+}
+
+unsafe fn explicit_lifetimes<'a>(_ptrs: &'a mut Ptrs<'r, 's>) {}
```
rinon: Doesn't this require adding `'r` and `'s` to the function signature generics?

c2rust-refactor/scripts/upgrade_ptr_to_ref.lua
```diff
@@ -0,0 +1,708 @@
...
+        end
+    end
+
+    return self.node_id_cfgs[node_id]
+end
+
+-- HrIds may be reused in different functions, so we should clear them out
+-- so we don't accidentally access old info
+function Visitor:visit_fn_decl(fn_decl)
+    self.vars = {}
```
rinon: Technically functions can be nested in rust, so this would clear in the middle of processing a function if so.

c2rust-refactor/scripts/upgrade_ptr_to_ref.lua
```diff
@@ -0,0 +1,708 @@
...
+
+        local pat_hirid = self.tctx:nodeid_to_hirid(locl:get_pat_id())
+
+        self:add_var(pat_hirid, Variable.new(local_id, true))
+    end
+end
+
+refactor:transform(
+    function(transform_ctx)
+        local node_id_cfgs = {
```
rinon: Let's at least put a TODO here that these ids should be passed to this transform, and they are specific to the test case.

c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -286,15 +450,258 @@ unsafe impl Send for LuaAstNode<P<Expr>> {}
...
+                e => unimplemented!("LuaAstNode<P<Expr>>:set_exprs() for {}", e.ast_name()),
+            }
+
+            Ok(())
+        });
+
+        methods.add_method("get_op", |_lua_ctx, this, ()| {
+            match &this.borrow().node {
+                ExprKind::Unary(op, _) => Ok(Some(op.ast_name())),
+                // ExprKind::Binary(op, ..) |
```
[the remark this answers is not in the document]
TheDan64: I think I didn't need them implemented, but acknowledged that they probably belong there too. I think `BinOp` didn't have `AstName` implemented for some reason, so I didn't go through the trouble of manually adding it since I didn't need it. (But weirdly `UnOp` had `AstName` impl'd already)

c2rust-refactor/tests/upgrade_ptr_to_ref/new.rs
TheDan64: Yes, indeed. But it seemed like we didn't want to venture too far into explicit lifetimes so I didn't get that far. (It would have been the next step)

c2rust-refactor/scripts/upgrade_ptr_to_ref.lua
TheDan64: That's a good point. I don't think we'll see much of that in translated code. But certainly should be considered.

Wouldn't those be two separate visits? IE visit outer then visit inner? Or would the inner item get visited mid outer visit?

c2rust-refactor/src/scripting/to_lua_ast_node.rs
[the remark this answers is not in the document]
rinon: Sounds good, just leave a TODO

c2rust-refactor/tests/upgrade_ptr_to_ref/new.rs
rinon: Ok. I'm inclined to never emit code that doesn't compile, so we either need a big warning here or we need to disable the feature in this case.

c2rust-refactor/scripts/upgrade_ptr_to_ref.lua
rinon: Inner would get visited inside the outer.

c2rust-refactor/src/scripting/to_lua_ast_node.rs
```diff
@@ -305,12 +712,232 @@ impl UserData for LuaAstNode<P<Expr>> {
...
+                TyKind::Rptr(_, mut_ty) => {
+                    *mut_ty = mt.borrow().clone();
+
+                    Ok(())
+                },
+                _ => Ok(()),
+            }
+        });
+
+        methods.add_method("get_path", |_lua_ctx, this, ()| {
```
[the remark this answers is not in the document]
TheDan64: Even for enum variants which have multiple fields?

c2rust-refactor/scripts/upgrade_ptr_to_ref.lua
TheDan64: Interesting. I wonder if I'll need like a stack of variables to push and pop or something

rinon: That's what I've done for nested modules. It would be nice to assert or something on this case, but I don't think it's critical. Needs a least a fat warning comment.

c2rust-refactor/src/scripting.rs
```diff
@@ -601,5 +641,144 @@ impl<'a, 'tcx> UserData for TransformCtxt<'a, 'tcx> {
...
+        });
+
+        methods.add_method("get_expr_ty", |_lua_ctx, this, expr: LuaAstNode<P<Expr>>| {
+            let rty = this.cx.node_type(expr.borrow().id);
+            let ty = reflect_tcx_ty(this.cx.ty_ctxt(), rty);
+
+            Ok(LuaAstNode::new(ty))
+        });
+
+        methods.add_method("get_field_expr_hirid", |_lua_ctx, this, expr: LuaAstNode<P<Expr>>| {
```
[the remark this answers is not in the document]
TheDan64: Not sure - the field id, in this case, is a `DefId`, not a `NodeId`. I've managed to avoid exposing `DefId`s to Lua until now. Do you think we should expose it?

PR #131, opened 2019-07-24
Split structures into inner and outer definitions to support packed-aligned structures

ahomescu:
Implement support for packed-aligned structures by splitting them into an outer-aligned newtype containing an inner-packed structure with the actual fields.


ahomescu:
I now realize this isn't ready yet, it doesn't support packed structures that contain aligned structures (Rust doesn't allow that intentionally), and I have some ideas on workarounds.

ahomescu:
I wound up doing the following things:
 
1. If a structure is both packed and aligned, then I split it into two parts: the aligned outer structure and the packed inner structure.

2. If a packed structure directly contains an aligned structure (as a field), I replace that with the unaligned version, followed by a padding pseudo-field that covers the size difference, like this:
```Rust
#[allow(dead_code , non_upper_case_globals)]
const Bar_PADDING: usize =
    ::std::mem::size_of::<Bar>() - ::std::mem::size_of::<Bar_Inner>();

pub struct Foo {
   pub bar: Bar_Inner,
   _pad: [u8; Bar_PADDING],
}
```

3. If a packed structure indirectly contains an aligned structure, we don't support this case and let rustc complain. I have some ideas on how to handle this case, but they're really complex and the kernel headers don't need this anyway.

rinon:
This did pass CI, was just a intermittent failure. Is there anything blocking a merge?

ahomescu:
> Is there anything blocking a merge?

No, I've just been waiting for everyone's comments.

TheDan64:
I didn't really have anything to add - the struct changes looked fine to me

PR #135, opened 2019-07-31
Track comments by source code span

rinon:
We want to track comments that should not be attached to the beginning of items or statements, e.g., comments at the end of a block. In order to do so, we need to keep track of all C AST node spans and map comments from their place inside those spans to analogous positions with the Rust AST spans. All comment spans are assigned before translation so that we can properly move non-definition function decl comments to their definition before either is traversed during translation.


rinon:
Comments in the middle or on the end of lines are not placed correctly, and comments on statements that don't get translated at all are missing (e.g. `(void)foo`).

PR #160, opened 2019-09-11
Static analysis for automated rewriting

TheDan64:


c2rust-refactor/src/analysis/ownership/mod.rs
```diff
@@ -196,6 +197,46 @@ fn analyze_inter<'lty, 'tcx>(cx: &mut Ctxt<'lty, 'tcx>) {
...
+        if func_summ.sig.inputs.len() == 2 {
+            let param0_is_mut_t = is_mut_t(func_summ.sig.inputs[0].ty);
+            let param1_is_isize = if let TyKind::Int(int_ty) = func_summ.sig.inputs[1].ty.sty {
+                int_ty == IntTy::Isize
+            } else {
+                false
+            };
+            let ret_is_mut_t = is_mut_t(func_summ.sig.output.ty);
+            if param0_is_mut_t && param1_is_isize && ret_is_mut_t {
+                if tctxt.def_path(*def_id).to_string_no_crate() == "::ptr[0]::{{impl}}[1]::offset[0]" {
```
rinon: Why don't you check this condition early? It's the most restrictive of the nested conditions.

c2rust-refactor/src/analysis/ownership/mod.rs
```diff
@@ -196,6 +197,46 @@ fn analyze_inter<'lty, 'tcx>(cx: &mut Ctxt<'lty, 'tcx>) {
...
+            if let TyKind::Param(param_ty) = mut_ty.ty.sty {
+                return param_ty.name.as_str().get() == "T";
+            }
+        }
+    }
+
+    false
+}
+
+fn register_std_constraints<'a, 'tcx, 'lty>(
```
rinon: Eventually we may want a more generic way to add constraints to lib functions, but I think we shouldn't add that now. A TODO comment mentioning that we should make it more generic if more functions are added would be nice though.

c2rust-refactor/src/scripting.rs
```diff
@@ -511,6 +540,33 @@ impl<'a, 'tcx> TransformCtxt<'a, 'tcx> {
...
+        );
+
+        methods.add_method_mut(
+            "clear_marks",
+            |_lua_ctx, this, ()| Ok(this.st.marks_mut().clear()),
+        );
+
+        methods.add_method(
+            "get_marks",
+            |lua_ctx, this, ()| {
```
rinon: Is there any way to not duplicate this and related functionality in `RefactorState` and `TransformCtxt`?

c2rust-refactor/scripts/upgrade_ptr_to_ref.lua
```diff
@@ -264,8 +306,6 @@ end
 function Visitor:visit_expr(expr)
```
rinon: `visit_expr` has gotten pretty long. I'd recommend splitting it up into separate methods for each type of rewrite.

c2rust-refactor/src/analysis/ownership/mod.rs
TheDan64: Gooood point. I think it was the last check I added, so checking early slipped my mind

c2rust-refactor/src/scripting.rs
TheDan64: Not sure I follow - what functionality is this duplicating? All this function does is serialize the marks to lua

rinon: Just above that there's a `get_marks` method on `RefactorState` that is identical.

TheDan64: Oh lol. Not sure how that happened. I definitely think we don't need both

c2rust-refactor/src/scripting.rs
```diff
@@ -97,6 +104,24 @@ pub fn run_lua_file(
     Ok(())
 }
 
+/// Takes a set of marks and turns it into a lua table
+fn lua_serialize_marks<'lua>(
```
ahomescu: Should this be called `get_marks`? It would be nice to have a 1:1 mapping from Lua functions to their Rust implementations.

c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -144,7 +144,15 @@ impl PrintParse for Block {
 
 impl PrintParse for Arg {
     fn to_string(&self) -> String {
-        pprust::arg_to_string(self)
```
ahomescu: Is this something that will be added to nightly soon?

TheDan64: CC @rinon - Think you added this?

c2rust-refactor/src/scripting.rs
TheDan64: Think so? It seemed like serializing was more literally what it's doing from rust's perspective, while "getting" them from Lua's

c2rust-refactor/src/rewrite/strategy/print.rs
rinon: It's already in nightly, but after the change that made the pretty printer impossible for us to use.

c2rust-refactor/src/scripting.rs
rinon: I think `serialize_marks` is fine here, since it's not getting the marks out of the command state but taking them as input.

PR #165, opened 2019-09-26
Roll to nightly 2019-09-...

rinon:
We need proper support for parameter attributes, which requires a newer nightly.

I split out the rustc/libsyntax pretty-printer into our own crate so we can once again forge a list of comments to be emitted in the pretty-printed output. It seemed to invasive to request this change upstream, and the pretty-printer should be easy to maintain out-of-tree.


rinon:
Blocked until rust-lang/rust#64793 lands upstream.

rinon:
@TheDan64 The upgrade_ptr_to_ref test is failing, probably because the node ids are all wrong.

rinon:
Make that the `ownership_upgrade_ptrs` test.

rinon:
Would like to push this to `nightly-2019-10-03` to pick up https://github.com/rust-lang/rust/pull/64991. Will test tomorrow.

XVilka:
Please, if possible, to check if https://github.com/rust-lang/rust-clippy/pull/4590 is included. Since running clippy is a common task after the conversion is done, it's important to not let it fail.

XVilka:
Last buildable nightly toolchain with that ICE fix landed is `c27f7568b 2019-10-13`. What about using it instead?

rinon:
> Last buildable nightly toolchain with that ICE fix landed is `c27f7568b 2019-10-13`. What about using it instead?

Something changed in the compiler between 2019-10-09 and 2019-10-13 that broke one of our tests. I don't have time to fix it right now, so I'm going to merge this. We can revisit rolling to something newer if/when it is needed.

Issue #187, opened 2019-10-29
Build fails


samuela:
I'm installing c2rust following exactly the instructions provided in the readme, but it fails:

```
root@e581c80974df:/src# RUST_BACKTRACE=1 cargo +nightly-2019-10-04 install c2rust
    Updating crates.io index
  Installing c2rust v0.13.0
   Compiling libc v0.2.65
...
   Compiling itertools v0.8.1
error: failed to compile `c2rust v0.13.0`, intermediate artifacts can be found at `/tmp/cargo-installdotAug`

Caused by:
  failed to run custom build command for `c2rust-refactor v0.13.0`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installdotAug/release/build/c2rust-refactor-dead6bcacbd887b2/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'failed to run process_ast.py: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:1165:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:76
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:60
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:64
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:196
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:210
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:473
  11: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:380
  12: rust_begin_unwind
[...]

PR #188, opened 2019-10-29
Improve python3 dependency awareness

samuela:
Fixes https://github.com/immunant/c2rust/issues/187. Adds a bit more detail to the panic message and adds python3 to the dependencies listed in the README.

### DIFF
--- a/c2rust-refactor/build.rs
+++ b/c2rust-refactor/build.rs
@@ -11,7 +11,7 @@ fn process_ast(mode: &str, dest: &Path) {
         .arg(mode)
         .arg(dest)
         .spawn()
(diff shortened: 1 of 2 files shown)


c2rust-refactor/build.rs
```diff
@@ -11,7 +11,7 @@ fn process_ast(mode: &str, dest: &Path) {
         .arg(mode)
         .arg(dest)
         .spawn()
-        .expect("failed to run process_ast.py");
+        .expect("failed to run process_ast.py. Make sure python3 is in your path.");
```
samuela: maybe `PATH` instead of `path`?

PR #236, opened 2020-02-16
Update instruction to add `rustc-dev` component

JohnTitor:
Now we don't contain `rustc-dev` by default so it's worth to add it explicitly.


rinon:
LGTM. I'm surprised you ran into an issue, because I didn't think `rustc-dev` was removed from the default set of components until https://github.com/rust-lang/rust/issues/67425.

@thedataking ?

thedataking:
@JohnTitor thanks a bunch for making us aware that this will be an issue for us when we roll to a newer nightly version. Much appreciation!

PR #237, opened 2020-02-17
Format code using 'cargo fmt'

Atul9:


rinon:
Formatting the project is something I've been wanting to do for awhile. However, I'd prefer to do it a single module at a time so we can manually review the changes and make sure that the formatting still makes sense. There are some blocks that I know we will want to whitelist from formatting as the rustfmt rules aren't universally applicable.

That said, this caused tests to fail, which definitely needs to be investigated and fixed.

PR #255, opened 2020-05-07
add %f support to the 'convert_format_args' command

marcograss:
I was refactoring some translated code and I noticed there is no support for %f format string specifiers, so I added it.

I looked up online and seems in recent C standards (but maybe not all of them?) the %f format string specifier will promote the float to double.

so given "double" is 8 byte in size, I chose to use f64 as a rust data type.

Let me know if you see any problem in this, I'm still new to c2rust,

thanks

marco

### DIFF
--- a/c2rust-refactor/src/transform/format.rs
+++ b/c2rust-refactor/src/transform/format.rs
@@ -278,6 +278,7 @@ enum CastType {
     Usize,
     Char,
     Str,
+    Float,
 }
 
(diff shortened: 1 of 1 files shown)


XVilka:
@thedataking @ahomescu is there anything preventing merging this? Are there any problems with such conversion?

ahomescu:
Sorry, I think this slipped through the cracks.

PR #273, opened 2020-07-02
Keep top-level inline function decls. Closes #272

### LINKED ISSUE #272: Translation of Zebra misses inlined functions
Issue surfaced by @panstromek while translating https://github.com/panstromek/zebra-1/tree/minz

The file `counter.c` and others contain top-level function declarations marked inline. These are not emitted as Rust because `prune_unused_decls` only marks global non-inlined functions as roots. See https://github.com/immunant/c2rust/blob/9f9d86ec86daa2341fd81645d5a661175df62046/c2rust-transpile/src/c_ast/mod.rs#L539

thedataking:


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -536,7 +536,6 @@ impl TypedAstContext {
                 CDeclKind::Function {
                     body: Some(_),
                     is_global: true,
-                    is_inline: false,
```
ahomescu, line 539: I think the check should become something like `!is_inline || (is_inline && is_extern)`.

ahomescu: Although I'm not 100% sure even that is correct.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -536,7 +536,6 @@ impl TypedAstContext {
...
                     ..
                 } => {
```
ahomescu, line 540: If you replace `is_inline: false,` with `is_inline,` above and add `is_extern,`, then you can do:
```suggestion
                } if !is_inline || is_extern => {
```

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -536,11 +536,14 @@ impl TypedAstContext {
...
                     body: Some(_),
                     is_global: true,
-                    is_inline: false,
+                    is_inline: is_inl,
+                    is_extern: is_ext,
                     ..
                 } => {
-                    to_walk.push(decl_id);
-                    used.insert(decl_id);
+                    if !is_inl || (is_inl && is_ext) {
```
ahomescu, line 543: Didn't my suggestion work? You shouldn't need a separate `if`, it could just be a guard in the match arm.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -536,11 +536,14 @@ impl TypedAstContext {
                 CDeclKind::Function {
                     body: Some(_),
                     is_global: true,
-                    is_inline: false,
+                    is_inline: is_inl,
```
ahomescu, line 539: Doesn't it work without the rename to `is_inl`?

PR #277, opened 2020-07-08
Handle _Static_assert gracefully. Closes #276

### LINKED ISSUE #276: Warn that `_Static_assert` is unsupported.
The transpiler currently chokes on `_Static_assert`s rather than gracefully emitting a warning that this is an unsupported C2X feature. This problem was surfaced when transpiling `postgres`. 

Rust has a [`static_assertions`](https://crates.io/crates/static_assertions) crate that could possibly be used to support translations of C static asserts.

thedataking:
Without this PR, the c2rust transpiler barfs on translation units that use this feature. With it, we emit a warning instead.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -857,6 +857,13 @@ class TranslateASTVisitor final
         abort();
     }
 
+    bool VisitStaticAssertDecl(StaticAssertDecl *SAD) {
+        std::vector<void *> childIds = {SAD->getAssertExpr(), SAD->getMessage()};
+        auto T = QualType(); // don't care about the type
+        encode_entry(SAD, TagStaticAssertDecl, childIds, T);
+        return true;
+    }
+
     bool VisitLabelStmt(LabelStmt *LS) {
 
         std::vector<void *> childIds = {LS->getSubStmt()};
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -858,7 +858,10 @@ class TranslateASTVisitor final
     }
 
     bool VisitStaticAssertDecl(StaticAssertDecl *SAD) {
-        std::vector<void *> childIds = {SAD->getAssertExpr(), SAD->getMessage()};
+        std::vector<void *> childIds = {SAD->getAssertExpr()};
+        auto msg = SAD->getMessage();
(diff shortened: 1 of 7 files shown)


c2rust-transpile/src/c_ast/iterators.rs
```diff
@@ -174,6 +174,13 @@ fn immediate_decl_children(kind: &CDeclKind) -> Vec<SomeId> {
         Field { typ, .. } => intos![typ.ctype],
         MacroObject { .. } | MacroFunction { .. } => vec![],
         NonCanonicalDecl { canonical_decl } => intos![canonical_decl],
+        StaticAssert {assert_expr, message} => {
```
ahomescu, line 177:
```suggestion
        StaticAssert { assert_expr, message } => {
```

c2rust-transpile/src/c_ast/iterators.rs
```diff
@@ -174,6 +174,13 @@ fn immediate_decl_children(kind: &CDeclKind) -> Vec<SomeId> {
...
+            if message.is_some() {
```
ahomescu, line 178:
```suggestion
            if let Some(message) = message {
```

c2rust-transpile/src/c_ast/iterators.rs
```diff
@@ -174,6 +174,13 @@ fn immediate_decl_children(kind: &CDeclKind) -> Vec<SomeId> {
...
+                intos![assert_expr, message.unwrap()]
```
ahomescu, line 179:
```suggestion
                intos![assert_expr, message]
```

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -857,6 +857,13 @@ class TranslateASTVisitor final
         abort();
     }
 
+    bool VisitStaticAssertDecl(StaticAssertDecl *SAD) {
+        std::vector<void *> childIds = {SAD->getAssertExpr(), SAD->getMessage()};
+        auto T = QualType(); // don't care about the type
```
ahomescu, line 862: Does this need to be a separate variable?

thedataking: no, but if I just pass `QualType()` as the 4th argument how do I communicate that we don't care about this parameter?

ahomescu: Move the comment to that line?

PR #290, opened 2020-09-03
Map one and zero to true and false in bool expressions

thedataking:


fw-immunant:
I rebased this. It seems like a fairly straightforward bit of low-hanging fruit; we now transpile `_Bool cond = 1;` as `let mut cond: bool = true;` (as opposed to `let mut cond: bool = 1 as std::ffi::c_int != 0;`).

thedataking:
@kkysen can this land now or is it obsoleted by your work?

PR #295, opened 2020-09-14
Use raw identifier for reserved keywords

### LINKED ISSUE #279: Translate Rust keywords into raw identifiers
We can be better at picking variable names when the C code uses an identifier that shadows a Rust keyword. The example buffer library uses `self` as an identifier which is named `self_0` in the output. We should emit `r#self` instead.

davidMcneil:
Resolves #279

Unfortunately, we cannot use a raw identifier for all keywords see the comment [here](https://github.com/immunant/c2rust/blob/99e56b7e659d43f7b0a1dcae2c3d7144aa5a20e3/c2rust-transpile/src/convert_type.rs#L135). 

Signed-off-by: David McNeil <mcneil.david2@gmail.com>

### DIFF
--- a/c2rust-bitfields-derive/src/lib.rs
+++ b/c2rust-bitfields-derive/src/lib.rs
@@ -193,7 +193,10 @@ fn bitfield_struct_impl(struct_item: ItemStruct) -> Result<TokenStream, Error> {
     let field_types_setter_arg = &field_types;
     let method_names: Vec<_> = bitfields
         .iter()
-        .map(|field| Ident::new(&field.name, Span::call_site().into()))
+        .map(|field| {
+            // We cannot use `Ident::new` here because it does not support raw identifiers
+            quote::format_ident!("{}", field.name, span = Span::call_site().into())
+        })
         .collect();
     let field_names: Vec<_> = bitfields.iter().map(|field| &field.field_name).collect();
     let field_names_setters = &field_names;
@@ -202,7 +205,9 @@ fn bitfield_struct_impl(struct_item: ItemStruct) -> Result<TokenStream, Error> {
         .iter()
         .map(|field_ident| {
             let span = Span::call_site().into();
-            let setter_name = &format!("set_{}", field_ident);
+            let raw_ident = &field_ident.to_string();
+            let ident = strip_raw_ident(&raw_ident);
+            let setter_name = &format!("set_{}", ident);
 
             Ident::new(setter_name, span)
         })
@@ -262,4 +267,14 @@ fn bitfield_struct_impl(struct_item: ItemStruct) -> Result<TokenStream, Error> {
     };
 
     Ok(q.into())
+}
+
+fn strip_raw_ident(ident: &str) -> &str {
+    const RAW_IDENT_PREFIX: &str = "r#";
+
+    if ident.starts_with(RAW_IDENT_PREFIX) {
+        &ident[RAW_IDENT_PREFIX.len()..]
+    } else {
+        ident
+    }
 }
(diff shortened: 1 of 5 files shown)


c2rust-transpile/src/renamer.rs
```diff
@@ -176,6 +178,14 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
...
         self.pick_name(&format!("fresh{}", fresh))
     }
+
+    fn raw_identifier_if_reserved_name(basename: &str) -> Option<String> {
+        if RESERVED_NAMES.contains(&basename) && !RESERVED_PATH_SEGMENT_NAMES.contains(&basename) {
+            Some(format!("r#{}", basename))
+        } else {
+            None
+        }
+    }
```
davidMcneil, line 188: It is a bummer that `RESERVED_NAMES` and `RESERVED_PATH_SEGMENT_NAMES` leak into `Renamer`. I could make the `Renamer` take a closure allowing the detection of reserved names to be configurable.

c2rust-bitfields-derive/src/lib.rs
```diff
@@ -263,3 +268,13 @@ fn bitfield_struct_impl(struct_item: ItemStruct) -> Result<TokenStream, Error> {
...
+
+fn strip_raw_ident(ident: &str) -> &str {
+    const RAW_IDENT_PREFIX: &str = "r#";
+
+    if ident.starts_with(RAW_IDENT_PREFIX) {
+        &ident[RAW_IDENT_PREFIX.len()..]
+    } else {
+        ident
+    }
+}
```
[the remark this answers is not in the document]
ahomescu: It's a small function, probably not worth de-duplicating.

c2rust-transpile/src/renamer.rs
ahomescu: Renamer already takes a reference to `RESERVED_NAMES` as the argument to its `new` constructor, so you could take that reference and stick it inside `Renamer` but I'm not sure that's worth it.

davidMcneil: Agreed, that was my initial approach (ie treat `reserved_names` as names that should be converted to a raw identifier). However, when I had to add `RESERVED_PATH_SEGMENT_NAMES` there did not seem to be value in keeping the reserved list internally.

thedataking:
@davidMcneil I've fixed the problem that caused Azure pipelines to fail. Can you rebase your PR on top of master?

davidMcneil:
@thedataking I rebased, but it is still failing. It looks like there is a problem reading this [token](https://github.com/immunant/c2rust/blob/ca54f19f1dd930eb22a7055ae402b4b46fdfbafe/.github/workflows/internal-testsuite.yml#L32)?

thedataking:
Correct, David that the github actions fails because of the secret token. I turned off actions on forked PRs for that reason so we can ignore that failure. If Azure Pipelines testing succeeds and Andrei is happy, we can can merge this PR.

ahomescu (approved):

XVilka:
Should be this PR rebased and updated? @davidMcneil

c2rust-bitfields-derive/src/lib.rs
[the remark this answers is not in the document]
kkysen: We can use [`str::strip_prefix`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix) now.

```rust
ident.strip_prefix("r#").unwrap_or(ident)
```

cydparser:
I've rebased this PR on master and fixed a new bug: https://github.com/cydparser/c2rust/compare/aca484279...c509a9f52

@davidMcneil, if you are too busy to finish this PR, let me know and I'll open a new one.

Rua:
@cydparser Are you willing to continue this one? It looks pretty useful!

cydparser:
@Rua, I am, but I was hoping to receive consent from the original author.

Rua:
@cydparser I don't think they will be coming back. It's been 6 years...

thedataking:
@cydparser the PR is approved and carries meta-data identifying the original author so I also do not see a problem with you landing this without waiting any longer for @davidMcneil to respond. Looks like they've had 3mo to do so already.

Issue #294, opened 2020-09-11
Consider emitting the c-types provided by std, instead of depending on `libc`


Luro02:
The standard library provides [`std::os::raw`](https://doc.rust-lang.org/std/os/raw/index.html), which defines `c_*`-types that can be used for ffi. Those types are identical to the types provided by `libc`.

It would be nice if c2rust would emit the types provided by the standard library, instead of depending on an external crate. (it is annoying to replace all occurences of `libc::*` with the std equivalent).

ahomescu:
The main thing to consider is that the `std::os::raw` types are only available with `libstd` but not in `no_std` mode, while the `libc` crate supports `no_std`.

> It would be nice if c2rust would emit the types provided by the standard library, instead of depending on an external crate.

We could add an option `--use-std-types` to the transpiler, but I think we still need to support both sets of types. Alternatively, we could switch to `std::os::raw` types as the default and add an option for the libc types.

> (it is annoying to replace all occurences of libc::* with the std equivalent).

The c2rust-refactor tool should be able to automate that, e.g., using `rewrite_ty`.

Luro02:
Relevant issue https://github.com/rust-lang/rust/issues/36193

ahomescu:
@Luro02 I implemented this and created a pull request, feel free to use the `std_raw_types` branch while it gets reviewed.

Luro02:
Very cool, thank you 🤗

bungcip:
i think we can close this issue because https://github.com/immunant/c2rust/pull/1381 solves the problem

PR #297, opened 2020-09-15
Std raw types

ahomescu:
Based on https://github.com/immunant/c2rust/issues/294, switches to `std::os::raw` types by default and adds optional `--use-libc-types` transpiler argument.

### DIFF
--- a/c2rust-transpile/src/convert_type.rs
+++ b/c2rust-transpile/src/convert_type.rs
@@ -20,6 +20,7 @@ pub struct TypeConverter {
     fields: HashMap<CDeclId, Renamer<FieldKey>>,
     suffix_names: HashMap<(CDeclId, &'static str), String>,
     features: HashSet<&'static str>,
+    use_libc_types: bool,
     emit_no_std: bool,
 }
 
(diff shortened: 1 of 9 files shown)


chrysn:
Bindgen has an option `--ctypes-prefix "some_crate::libc"` that allows the user to set the path, which is convenient in embedded environments. It defaults to `::std::os::raw`. I think the option would make sense for C2Rust as well.

kkysen:
We should update this to use the `core::ffi::c_*` types now that they're in `core`, so we don't have to worry about not being able to use the `std::os::raw` types in `no_std` crates.  They're still unstable, but they look like they're going to be stabilized shortly: https://github.com/rust-lang/rust/issues/94501.  They're still on nightly, but we depend on many nightly features already.  I'd think the `core` types should be preferable to the `libc` ones, since you could be translating C code that doesn't use `libc` at all.  And once the `core::ffi::c_*` types are stabilized, we can also use the `std::ffi::c_*` re-exports if we ever want to generate `std` instead of `core`.

chrysn:
Much of the generated code is already well usable on stable Rust; even if it's stabilized soon it'd be months until it's on stable.

I thus still advocate a `--ctype-prefix` option -- when used with nightly, `core::ffi` and `std::ffi` (any good reason to use that, now that core doesn't even need to be an extern crate any more?) can be good defaults.

kkysen:
Yeah, I agree.  A `--ctypes-prefix` option is much more flexible, and in the future once things are stabilized, we can default to `core::ffi`.

kkysen:
`core::ffi::c_*` was just stabilized in 1.64: https://github.com/rust-lang/rust/issues/94501#issuecomment-1186301451.

Issue #293, opened 2020-09-03
stdatomic.h raises missing types for AstNode { ... __ATOMIC_SEQ_CST }


chrysn:
Whenever c2rust encounters stdatomic.h in my setup, it fails irrecoverably with messages like this:

```
  warning: Missing type 94827692380208 for node: AstNode { tag: TagTypedefDecl, children: [], loc: SrcSpan { fileid: 81, begin_line: 97, begin_column: 1, end_line: 97, end_column: 37 }, type_id: Some(94827692380208), rvalue: LValue, macro_expansions: [], macro_expansion_text: Some("__ATOMIC_SEQ_CST"), extras: [Text("atomic_uint_least32_t"), Bool(false)] }
  Exported Clang AST was invalid. Check warnings above for unimplemented features.
  --> /usr/lib/llvm-10/lib/clang/10.0.1/include/stdatomic.h:97:1
   [-Wclang-ast]
  thread 'main' panicked at 'Could not find CTypeId(4932) in TypedAstContext', c2rust-transpile/src/c_ast/mod.rs:816:21
```

(The block from `warning:` to `[-Wclang-ast]` is repeated dozens of times; all earlier warnings are unrelated "macro redefined" ones).

As support for LLVM 10.0.1 is rather recent, it might be related to that. I'm using Rust nightly-2020-08-29, and LLVM 10.0.1 as in the above error messages..

<details><summary>Full compile_commands.json in case it's relevant</summary>
```
[
  {
    "arguments": [
      "any-cc",
      "-DTHREAD_STACKSIZE_MAIN=(THREAD_STACKSIZE_DEFAULT + 5000)",
      "-DDEVELHELP",
      "-Wall",
      "-Wextra",
      "-pedantic",
      "-g3",
      "-std=gnu11",
      "-m32",
      "-fstack-protector-all",
      "-ffunction-sections",
      "-fdata-sections",
      "-DDEBUG_ASSERT_VERBOSE",
      "-DRIOT_APPLICATION=shell_threads",
      "-DBOARD_NATIVE=native",
      "-DRIOT_BOARD=BOARD_NATIVE",
      "-DCPU_NATIVE=native",
      "-DRIOT_CPU=CPU_NATIVE",
      "-DMCU_NATIVE=native",
      "-DRIOT_MCU=MCU_NATIVE",
      "-fno-common",
      "-Wall",
      "-Wextra",
      "-Wmissing-include-dirs",
      "-fdiagnostics-color",
[...]

PR #291, opened 2020-09-03
Add --preserve-unused-functions

chrysn:
to keep otherwise pruned static / inline functions.

Closes: https://github.com/immunant/c2rust/issues/287

---

This is an initial simple approach to #287. It looks like it's generating valid code, but I'm too early in my whole evaluation of c2rust to say whether what comes out of this will work, so please take it with due care. (Also, tests don't pass, but neither do they pass on my machine on current master, so...)

I'm playing fast and lose with API stability here, as I don't have the overview to tell whether prune_unused_decls is part of any practically public API. If it is, I can replace the added parameter with a second function `prune_unused_decls_but_not_functions` and rename the current implementation to a private impl one.

PR #302, opened 2020-09-18
Check for C11 atomics in and around VisitQualType

chrysn:
This does not help towards the support of C11 atomics, but ensures that
the existing error messages about the trouble with them get shown:

* In VisitQualType at latest, where the source location information is
  already gone so it's not that helpful (but at least it should catch a
  wide range of sources where those types could come from)

* In VisitTypedefNameDecl, where the `typedef _Atomic(...) atomic_...;`
  declarations of clang's stdatomic.h run through. It's but one of many
  possible callers, but it's the one where one more check can make the
  difference between a usable and a cryptic error for many cases.

  (And there's no point in cluttering everything with checks as we're soon gonna have atomic support anyway, aren't we? ;-) )

Closes: https://github.com/immunant/c2rust/issues/293

---

Testing procedure: Run the minimal example of the closed issue. Instead of the long errors, there's now a concise:

```
In file included from test.c:1:
/usr/lib/llvm-10/bin/../lib/clang/10.0.1/include/stdatomic.h:76:37: error: c2rust: C11 Atomics are not supported. Aborting.
typedef _Atomic(_Bool)              atomic_bool;
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

If it weren't for the front-end check in VisitTypedefNameDecl (and removing that was how I tested the back-end check), the error would be instead:

```
error: c2rust: C11 Atomics are not supported. No precise location information available. Aborting.
```


chrysn:
As with #291, builds pass locally:

```shell
$ ./scripts/test_translator.py ./tests
[...]
Test summary:
  unexpected failures: 0
  unexpected successes: 0
  expected failures: 3
  successes: 108
```

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -157,6 +157,24 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
           astEncoder(ast) {}
 
     void VisitQualType(const QualType &QT) {
+        if(isa<AtomicType>(QT)) {
```
thedataking, line 160: nit: looks like this could be `QT.isAtomicType()`.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2104,6 +2122,17 @@ class TranslateASTVisitor final
                          cbor_encode_boolean(array, D->isImplicit());
                      });
 
+        if(isa<AtomicType>(typeForDecl)) {
```
thedataking, line 2125: see previous comment.

thedataking (approved):

thedataking:
@chrysn thanks for this PR. I fixed the CI so if you push the minor changes I recommend, we should be able to get this tested and merged in.

c2rust-ast-exporter/src/AstExporter.cpp
chrysn: Nope -- gives `error: ‘const class clang::QualType’ has no member named ‘isAtomicType’`. (When applied only here, and also for other locations which was what I first tried as TBH the `isa<AtomicType>` is something I plainly copy-pasted from other locations in this C++ file).

chrysn:
Force-pushed with new time stamp to trigger CI rebuild.

chrysn:
Hm, stuck in CI again. Should I try to get it unstuck where it is, or just rebase onto master and see how things go there?

fw-immunant:
If you have time to poke at this again, it would be nice to rebase it onto master and make sure everything is green. I've seen our CI being randomly flaky before.

chrysn:
Rebased, and checked that the copy-pasted idioms are still around the same way at their source location.

After I've discovered that I've always been running with outdated builds (`cargo run -- transpile ...` isn't enough, it needs to be `cargo build && cargo run` because c2rust is accessing c2rust-transpile through target).

The original testing procedure is still accurate. If you want to trigger the "No precise location information" case, try C like

```c
void x(_Atomic(char) *c) { }
```

I'm sure there are places in the code where this could be caught with location information, but I'd say that constructs like the above are rare anyway, and that it's good to have a fallback in case we missed any other location.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -189,6 +189,24 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
           astEncoder(ast) {}
 
     void VisitQualType(const QualType &QT) {
+        if(isa<AtomicType>(QT)) {
```
fw-immunant, line 192: This check needs to be moved inside the `if (!QT.isNull()) {` conditional below; otherwise we get an internal Clang abort doing `isa<AtomicType>(QT)` when QT has a null inner pointer.

fw-immunant (changes requested):
Commented inline about issue causing test failures.

kkysen:
By the way, the issue with `cargo run -- transpile` not rebuilding `c2rust-transpile` and having to do `cargo build && cargo run` is really annoying and not easy to figure out what's going wrong.  Let me see if I can fix that.  I had no idea that was happening either.

kkysen:
* Superseded by #814.

Issue #305, opened 2020-09-19
Platform generic transpilation


chrysn:
The FAQ are explicit about not supporting cross-transpilation, but I'd like to use that and work on fixing it (primarily to get https://github.com/rust-lang/rust-bindgen/issues/1344 working); I'd appreciate a bit of high-level review before I get head-on into it.

According to the FAQ entry, the preprocessor is run first, and that's what's keeping transpilation from working. With the right flags supplied to the preprocessor, that should not be too much trouble. (After all, flags are already supplied for platform dependent stuff, like any defines the program may use to determine endianness and such).

Looks to me that the more difficult step will be getting the types right, ie. putting i32 or i64 where C has its `int` or `long`. The cty crate may be of some help there.

chrysn:
My initial experimentation on this indicates that working around the host preprocessor expansion might even be sufficient to transpile on generic platforms. Since I've started using a C compiler with the right target platform in `-E` mode to expand any defines and includes, all the non-matching sizes disappeared. That preprocessing goes down even into stdint.h, so C2Rust only sees native C types like "int" or "long", and defers to `::libc` (or `::cty` as I replace it, manually although I could merge in #297 and do it there).

The downside of this approach so far is that I'm using the actual platform C compiler and not clang with suitable flags -- which means that under some circumstances, GCC idiom winds up there and I get errors like "Type conversion not implemented for TagTypeUnknown expecting 3" (At least I think that's what's happening there).

Eventually, I think that if all arguments are passed around right (and I'm honestly not quite sure whether they aren't already), clang could work from the data already in compile_commands.json and produce output suitable for any platform, provided it's specified the right way in the compile commands.

chrysn:
Further experimentation has shown that expanding preprocessor is only really necessary when using a C compiler different from clang -- that then has its own path discovery, and eventually finds the platform-specific definitions for variously sized integers that clang does not find. Where clang is used, I see no real difference in the results between using preprocessed and original data. (Function sequence differs, and preprocessing loses the data C2Rust would need to `--translate-const-macros`).

Would it be correct to change the FAQ entry to

> > I translated code on platform X but it didn't work correctly on platform Y
> 
> We run the C preprocessor before translation to Rust. This specializes the code to the platform given in the compile_commands, defaulting to the host platform. For this reason, in order to cross-tanspile, the same platform must be selected during transpilation and compilation of the Rust code.

?

[...]

PR #307, opened 2020-09-20
asm-casts: Run on no_std

chrysn:
There's nothing in the asm-casts crate that necessitates std (and the std::marker::PhantomData was changed to its alias core::marker::PhantomData). This PR just sets `#![no_std]` and switches over the PhantomData import.

This allows code produced with asm casts to be used in no_std environments. (Admittedly, that's a bit far-fetched as there's no official support for cross-platform transpilation yet, see https://github.com/immunant/c2rust/issues/305 -- but it's another stepping stone towards https://github.com/rust-lang/rust-bindgen/issues/1344).

I've tested this in a practical application (it builds and runs -- but I can't tell yet whether it actually executes any place with those casts), and run the local `cargo test` which runs the crate-internal coverage combo successfully.

### DIFF
--- a/c2rust-asm-casts/src/lib.rs
+++ b/c2rust-asm-casts/src/lib.rs
@@ -1,4 +1,5 @@
-use std::marker::PhantomData;
+#![no_std]
+use core::marker::PhantomData;
 
 /// Pseudo-structure that provides the inner type definition
 /// and cast functions for every pair of types used


ahomescu (approved):

ahomescu:
I approved this, but we need to see what's up with that failing test (might not be caused by your change).

chrysn:
It's the `##[error]Input required and not supplied: token` error I've seen in other failed build runs as well. (The later `##[error]Cannot read property 'replace' of null` might be a follow-up error from the token not being set, and thus possibly `null`).

Looks to me like https://github.com/immunant/c2rust/blob/ca54f19f1dd930eb22a7055ae402b4b46fdfbafe/.github/workflows/internal-testsuite.yml#L32 is missing its input; can't do much about that from the outside as that probably needs configuring somewhere in the CI setup.

thedataking:
The failing test is because the forked version does not have access to the repo secret needed to pass the GitHub Actions tests. I tried disabling GitHub Actions on third-party PRs but apparently I wasn't successful. Merging.

chrysn:
Thanks for merging.

Would it be possible to upload this to crates.io? I'm approaching the release of the new c2rust based version of riot-sys, and cargo insists all dependencies are uploaded.

(The c2rust which users need to have installed depends on, basically, all the other PRs I currently have open, but that's not so bad because that dependency is outside cargo anyway as c2rust can't be a build dependency AFAIU, but the c2rust-asm-casts are a regular dependency).

ahomescu:
Pinging @rinon for crates.io releases.

chrysn:
@rinon, is there anything left to do / where I contribute to getting this to crates.io? That dependency is currently the largest stopper to a new round of RIOT sys-/wrapper crates being published.

rinon:
@chrysn new version is published, sorry it took so long.

chrysn:
I still see https://crates.io/crates/c2rust-asm-casts as being the 0.1.1 of last 2019-10-30, might something have gone wrong during the upload?

rinon:
I was dumb and missed this crate in the publish. Should be good now @chrysn

PR #313, opened 2020-11-11
llvm 11 fixes

kaspar030:
My Arch Linux bumped me to clang/llvm 11 today, which broke c2rust compilation.

This PR contains the fixes I needed to make it compile:

1. ASTContext::getRawCommentList() was removed. It was a getter for ASTContext::Comments, which is a public field. So there's a commit that directly accesses the field.

2. SmallString::str() changed semantics slightly, causing it to not map to std::string. A double `.str().str()` goes through StringRef to std::string. LLVM 11 offers a direct `std::string(X)` operator, but to keep this compatible, I kept the double call.

I did not test this on clang/llvm < 11...

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2562,7 +2562,7 @@ static std::vector<const char *> augment_argv(int argc, const char *argv[]) {
     SmallString<128> P("-extra-arg=-resource-dir=" CLANG_BIN_PATH);
     llvm::sys::path::append(P, "..", Twine("lib") + CLANG_LIBDIR_SUFFIX,
                             "clang", CLANG_VERSION_STRING);
-    std::string resource_dir = P.str();
+    std::string resource_dir = P.str().str();
     char *resource_dir_cstr = new char[resource_dir.length() + 1];
     strncpy(resource_dir_cstr, resource_dir.c_str(), resource_dir.length() + 1);
 
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2463,7 +2463,7 @@ class TranslateConsumer : public clang::ASTConsumer {
             }
 #else  // CLANG_VERSION_MAJOR >= 10
             const FileID file = sourceMgr.getMainFileID();
-            auto comments = Context.getRawCommentList().getCommentsInFile(file);
+            auto comments = Context.Comments.getCommentsInFile(file);
             if (comments != nullptr) {
                 cbor_encoder_create_array(&outer, &array, comments->size());
                 for (auto comment : *comments) {


chrysn:
I've merged this into my integration branch (for-riot, which contains all my PRs), and it's still working fine with -10.

rinon:
LGTM but we should test this on older LLVM versions.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2463,7 +2463,7 @@ class TranslateConsumer : public clang::ASTConsumer {
             }
 #else  // CLANG_VERSION_MAJOR >= 10
             const FileID file = sourceMgr.getMainFileID();
-            auto comments = Context.getRawCommentList().getCommentsInFile(file);
+            auto comments = Context.Comments.getCommentsInFile(file);
```
ahomescu, line 2466: If this is new to clang 11, it should be wrapped in a version check (see `CLANG_VERSION_MAJOR` above).

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2562,7 +2562,7 @@ static std::vector<const char *> augment_argv(int argc, const char *argv[]) {
     SmallString<128> P("-extra-arg=-resource-dir=" CLANG_BIN_PATH);
     llvm::sys::path::append(P, "..", Twine("lib") + CLANG_LIBDIR_SUFFIX,
                             "clang", CLANG_VERSION_STRING);
-    std::string resource_dir = P.str();
+    std::string resource_dir = P.str().str();
```
ahomescu, line 2565: Same as above: wrap in a check on `CLANG_VERSION_MAJOR`.

c2rust-ast-exporter/src/AstExporter.cpp
kaspar030: New in clang 11 is the removal of the getter. The field is public since at least clang 6. Does that warrant an `#ifdef`?

c2rust-ast-exporter/src/AstExporter.cpp
kaspar030: I compiled this in a ubuntu bionic container with clang-6 and libclang-6:

```
root@71976200ca48:/src/c2rust# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.5 LTS"
root@71976200ca48:/src/c2rust# clang --version
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
root@71976200ca48:/src/c2rust# cargo --version
cargo 1.41.0-nightly (626f0f40e 2019-12-03)
root@71976200ca48:/src/c2rust# rustc --version
rustc 1.41.0-nightly (6d77e45f0 2019-12-04)
root@71976200ca48:/src/c2rust# git describe  
0.1.0-3140-g8e7131e6
```

The test suite passed:
```
root@71976200ca48:/src/c2rust# python3 ./scripts/test_translator.py --debug ./tests
[...]
Test summary:                                                            
  unexpected failures: 0                                                 
  unexpected successes: 0                                                
  expected failures: 3                                                   
  successes: 108
```

I don't think an extra ifdef is necessary, but it is of course your call!

kaspar030:
> I don't think an extra ifdef is necessary, but it is of course your call!

Please advise. No ifdefs are necessary down to at least clang 6. I can add them if you'd prefer that.

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: > Does that warrant an #ifdef?

No, that's fine, as long as the new code works on older versions.

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: > I don't think an extra ifdef is necessary, but it is of course your call!

It's fine, no ifdef needed.

ahomescu (approved):

Issue #338, opened 2021-06-25
Some integer constants fail to convert properly


mjgarton:
Using current master ( ed1af8e4f829689a8cbeedf6ea5b33cb199045e5 ) the following code

```
#include <stdio.h>

int main()
{
	int x = 1;
	switch (x) {
	case 123:
		break;
	}
}
```
converts to
```
#![allow(dead_code, mutable_transmutes, non_camel_case_types, non_snake_case,
         non_upper_case_globals, unused_assignments, unused_mut)]
#![register_tool(c2rust)]
#![feature(main, register_tool)]
use ::rust::*;
unsafe fn main_0() -> libc::c_int {
    let mut x: libc::c_int = 1 as libc::c_int;
    match x { 0 | _ => { } }
    return 0;
}
#[main]
pub fn main() { unsafe { ::std::process::exit(main_0() as i32) } }
```

Note that the `match` case has `0` when it should have `123`

This is using llvm & clang 12.0.0

PR #339, opened 2021-06-25
Fix evaluateConstantInt

mjgarton:
Ensure we set the value when there is one, instead of when there isn't.

This fixes https://github.com/immunant/c2rust/issues/338 for me, although I am not certain it is fully correct.


rinon:
Thanks for the fix! I made a slight modification, can you try this out for your case?

mjgarton:
> Thanks for the fix! I made a slight modification, can you try this out for your case?

Yes, this works for me @rinon .  Thanks.

PR #343, opened 2021-07-30
CMake correctness; streamline `nix-shell` environment

bb010g:
These allow `nix-shell --pure --show-trace --keep-going --run 'cargo build --release'` from a clean checkout to succeed.

## Don't assume LLVM & Clang share the same lib dir
Also, don't assume `llvm-config --libdir` shares the same `lib/` root as `llvm-config --cmakedir`.

## streamline `nix-shell` environment
Rustup is no longer required when using `nix-shell`. Instead, the required Rust nightly is already available.

### DIFF
--- a/c2rust-ast-exporter/build.rs
+++ b/c2rust-ast-exporter/build.rs
@@ -118,6 +118,10 @@ fn generate_bindings() -> Result<(), &'static str> {
 fn build_native(llvm_info: &LLVMInfo) {
     // Find where the (already built) LLVM lib dir is
     let llvm_lib_dir = &llvm_info.lib_dir;
+    // Find where the (already built) LLVM cmake module dir is
+    let llvm_cmake_dir = &llvm_info.cmake_dir;
+    // Find where the Clang cmake module dir is
+    let clang_cmake_dir = &llvm_info.clang_cmake_dir;
 
     match env::var("C2RUST_AST_EXPORTER_LIB_DIR") {
         Ok(libdir) => {
@@ -127,8 +131,8 @@ fn build_native(llvm_info: &LLVMInfo) {
             // Build libclangAstExporter.a with cmake
             let dst = Config::new("src")
                 // Where to find LLVM/Clang CMake files
-                .define("LLVM_DIR", &format!("{}/cmake/llvm", llvm_lib_dir))
-                .define("Clang_DIR", &format!("{}/cmake/clang", llvm_lib_dir))
+                .define("LLVM_DIR", llvm_cmake_dir)
+                .define("Clang_DIR", clang_cmake_dir)
                 // What to build
                 .build_target("clangAstExporter")
                 .build();
@@ -204,6 +208,10 @@ fn build_native(llvm_info: &LLVMInfo) {
 struct LLVMInfo {
     /// LLVM lib dir containing libclang* and libLLVM* libraries
     pub lib_dir: String,
+    /// LLVM cmake dir containing cmake modules
+    pub cmake_dir: String,
+    /// Clang cmake dir containing cmake modules
+    pub clang_cmake_dir: String,
 
     /// List of libs we need to link against
     pub libs: Vec<String>,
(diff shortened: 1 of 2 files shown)


c2rust-ast-exporter/build.rs
```diff
@@ -273,23 +281,62 @@ impl LLVMInfo {
 
         let llvm_config = find_llvm_config();
         let llvm_config_missing = "
+        Couldn't find `llvm-config`. Make sure `llvm-config` is on $PATH then
+        re-build.
+        ";
```
fw-immunant, line 286: As far as I can tell, with these changes this string becomes dead. But I'm not opposed to keeping it around as a reminder to add better error handling for this situation.

c2rust-ast-exporter/build.rs
```diff
@@ -273,23 +281,62 @@ impl LLVMInfo {
...
+        let llvm_config_libdir_missing = "
         Couldn't find LLVM lib dir. Try setting the `LLVM_LIB_DIR` environment
         variable or make sure `llvm-config` is on $PATH then re-build. For example:
 
           $ export LLVM_LIB_DIR=/usr/local/opt/llvm/lib
         ";
+        let llvm_config_cmakedir_missing = "
```
fw-immunant, line 293: Arguably these should be named `llvm_libdir_missing`, and `llvm_cmakedir_missing`.

fw-immunant (approved):
This looks good overall and improves error reporting, which is always a plus. I'm not a nix expert but the nix changes look plausible.

kkysen:
@bb010g @fw-immunant, can this be merged now?  @fw-immunant approved it already and I just fixed the merge conflicts.

@thedataking, the `c2rust-testsuite` CI is still failing, though, even though [`c2rust-testsuite`](https://github.com/immunant/c2rust-testsuite) is now public.  I think that's because the `.github/workflows/internal-testsuite.yml` still uses a secret token, which it seems is what the CI is failing on.  Then there's also the slack fail in the same CI workflow.

kkysen:
CI is now failing after I re-triggered it with the merge:

[CI failure on Ubuntu 18](https://dev.azure.com/immunant/c2rust/_build/results?buildId=1806&view=logs&j=ae6377c6-aed8-50bb-2c16-24013b3441a6&t=46edcbf0-4390-5c30-2ae1-53f44109c6d5&l=382)

```shell
          Couldn't find Clang cmake dir. Try setting the `CLANG_CMAKE_DIR` environment
          variable or make sure `llvm-config` is on $PATH then re-build. For example:

            $ export CLANG_CMAKE_DIR=/usr/local/opt/llvm/lib/cmake/clang
          : Os { code: 2, kind: NotFound, message: "No such file or directory" }', c2rust-ast-exporter/build.rs:343:22
```

This might be flaky CI again since it only happened on Ubuntu 18.

kkysen:
Ubuntu 18 CI failed again, so it's probably not flaky.  I'm not sure why it's only happening in Ubuntu 18, and not even in Ubuntu 20.  The build scripts for the docker containers for those should be exactly the same.

kkysen:
https://github.com/immunant/c2rust/pull/343/commits/c54f7cab7521ac75f13235cf99eb829e936f7cfd, removing the `token` field, did fix the [`immunant/c2rust-testsuite`](https://github.com/immunant/c2rust-testsuite) checkout issue in GitHub Actions.

kkysen:
It's failing in Ubuntu 18 because Ubuntu 18 uses llvm 6, which doesn't have a `$(llvm-config --cmake-dir)/../clang` directory, and in fact, llvm doesn't have this directory until llvm 9.  We'll have to find a fix for that.  @bb010g, do llvm and clang need separate directories before llvm 9?  The changes were in https://github.com/immunant/c2rust/pull/343/commits/289d44a4b998da053e583cddf2b934f2bca622fe.  Should we have a fallback to the old behavior?

bb010g:
@kkysen A fallback sounds reasonable. I don't think I tested on older LLVMs.

Issue #346, opened 2021-08-27
packed attribute not transferred on unions


chrysn:
Unions with `__attribute__((packed))` are translated by C2Rust without the `#[repr(C, packed)]` which bindgen produces; comparing memory layouts with C struct shows that it should be packed.

PR (of "hey I think I know which lines to copy around" quality) coming up.

PR #311, opened 2020-09-22
Disarm C's doc comments even inside lines

chrysn:
This is a straightforward fix for #310, accompanied by an extension to the test cases to cover the new behavior. The new test failed with the old code, and now passes.

Issue #309, opened 2020-09-22
ASM wrappers produce code that violates borrow rules


chrysn:
With the following ARM assembly function in `test.c` (found in the ARM CMSIS standard library as used in RIOT-OS; disclaimer: I have no clue what this actually does, it looks like an inline wrapper around some CPU instruction), the resulting Rust code is not buildable:

```c
#include <stdint.h>

uint64_t __SMLALDX (uint32_t op1, uint32_t op2, uint64_t acc)
{
  union llreg_u{
    uint32_t w32[2];
    uint64_t w64;
  } llr;
  llr.w64 = acc;

  asm volatile ("smlaldx %0, %1, %2, %3" : "=r" (llr.w32[1]), "=r" (llr.w32[0]): "r" (op1), "r" (op2) , "0" (llr.w32[1]), "1" (llr.w32[0]) );

  return llr.w64;
}
```

After manually fixing #306 by s/asm/llvm_asm/ on c2rust-lib.rs and test.rs, the error can be shown with:

```shell
$ c2rust transpile compile_commands.json --emit-build-file
Transpiling test.c
$ cargo +nightly build
   Compiling c2rust_out v0.0.0 (/tmp/asm)
warning: unused `#[macro_use]` import
  --> c2rust-lib.rs:13:1
   |
13 | #[macro_use]
   | ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

error[E0499]: cannot borrow `llr.w32[_]` as mutable more than once at a time
  --> test.rs:19:18
   |
17 |     let fresh0 = &mut llr.w32[1 as libc::c_int as usize];
   |                  --------------------------------------- first mutable borrow occurs here
18 |     let fresh1;
19 |     let fresh2 = &mut llr.w32[0 as libc::c_int as usize];
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ second mutable borrow occurs here
...
24 |          "r" (op2), "0" (c2rust_asm_casts::AsmCast::cast_in(fresh0, fresh4)),
   |                                                             ------ first borrow later used here

error[E0503]: cannot use `llr.w32[_]` because it was mutably borrowed
  --> test.rs:21:18
   |
17 |     let fresh0 = &mut llr.w32[1 as libc::c_int as usize];
   |                  --------------------------------------- borrow of `llr.w32[_]` occurs here
...
[...]

Issue #136, opened 2019-08-05
Packed structures indirectly containing aligned structures aren't supported


ahomescu:
Commit https://github.com/immunant/c2rust/commit/49d2b577a39d7d13f93468347cd3f235e0644c77 adds support for packed structures that are also aligned, and also packed structures that directly contain aligned structure fields. However, packed structures that indirectly contain aligned structures aren't supported by the current implementation, e.g.
```C
struct A {
  int x;
  struct B b;
} __attribute__((packed));

struct B {
  int x;
  struct C c;
}

struct C {
  int x;
} __attribute__((aligned(16)));
```

I have some ideas on how to support this, but the implementation would be somewhat complex.
Basically, we need to produce `X_Inner` and `X_PADDING` for all types that are directly or indirectly included in a packed structure, and manually lay out the fields of the `X_Inner` structure to perfectly match what C does.
This is tricky because, in addition to the post-field padding that the current implementation adds, we'd need to insert pre-field padding inside non-packed structures, e.g., between the `B::x` and `B::c` fields above.
We cannot rely on rustc inserting that padding by itself, since we're not using the `#[repr(align(N))]` attribute for alignment.

We need to control the exact layout of every structure ourselves using padding fields.
To start, we will distinguish between manually-aligned structures, i.e., structures whose layouts are directly or indirectly controlled by an `aligned(N)` attribute, and rustc-aligned ones.
I propose the following algorithm to emit a manually-aligned structure in Rust:
 1. Find the last manually-aligned field in the structure
 2. Split the structure, e.g., `Foo`, into the following components:
    a. The sub-structure containing all fields before the manually-aligned one (called `Foo_Pre` below)
    b. The pre-field padding
    c. The field itself (of type `T`)
    d. The post-field padding
    e. The substructure with all the following fields, if any (all rustc-managed), e.g.
```Rust
struct Foo {
  pre: Foo_Pre,
  pre_padding: [u8; field_PREPADDING],
  field: T,
  post_padding: [u8; field_POSTPADDING],
  post: Foo_Post,
}
```
 3. Compute the alignment `T_ALIGNMENT` of `T` as the maximum between the manually-specified alignment and `align_of::<T>()` from Rust.
 4. Compute `field_PREPADDING` as `align_to(size_of::<Pre>(), T_ALIGNMENT)` after computing `T_ALIGNMENT`.
 5. Further split the first 3 fields into a separate substructure, in order to compute the post-padding size:
```Rust
[...]

PR #347, opened 2021-08-27
Transfer the packed attribute on unions to repr(packed)

chrysn:
Closes: https://github.com/immunant/c2rust/issues/346

---

This should be a straightforward solution to #346, and works for me on RIOT (which also means I can't easily test it on the main branch, as I still need my for-riot vendor branch that has #311 and the branch mentioned in #309 merged).

There are no tests yet; is

```
$ cargo +nightly-20109-12-05 build --locked --release
$  python3 scripts/test_translator.py ./tests
```

the right procedure?

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -1946,6 +1946,8 @@ impl ConversionContext {
                     let name = expect_opt_str(&node.extras[0]).unwrap().map(str::to_string);
                     let has_def = from_value(node.extras[1].clone())
                         .expect("Expected has_def flag on struct");
+                    let attrs = from_value::<Vec<Value>>(node.extras[2].clone())
+                        .expect("Expected attribute array on record");
                     let fields: Option<Vec<CDeclId>> = if has_def {
                         Some(
                             node.children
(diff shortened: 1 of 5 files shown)


thedataking:
@chrysn the Darwin failures were due to timeouts as far as I can tell. I've opened a PR to get CI working properly again.

chrysn:
Is there anything that should be addressed about this PR? Does it help if I rebase it onto current master?

thedataking:
LTGM but I'm no expert on packed unions so I'd like a second pair of eyes before we merge. And yes, rebasing onto current master would be nice.

chrysn:
As for the correctness of the built unions, the way I'm using them
frequently transmutes them back and forth between parts generated by
bindgen and those generated by C2Rust. As part of these transmutes, I'm
checking the size behind the pointer -- and they match ever since I'm on
this branch.

Rebased onto master and pushed for consumption by the CI; I haven't run
my full use case (which does these tests as part of everyday operation)
on the rebased branch yet.

chrysn:
I've given the rebased packed-unions a test run. Most tests also merged in the minimal-309 branch referenced in https://github.com/immunant/c2rust/issues/309#issuecomment-696671353, as that's necessary for other parts of the code. As expected, that branch has passed the inline casting tests (asserting that bindgen's view of the C structs is as large as C2Rust's), which current master plus minimal-309 does not).

fw-immunant (approved):
This LGTM and does what it says on the tin. I think we should merge.

fw-immunant:
I think it's worth noting separately, however, that we still have some holes in our support of the `packed` attribute--it isn't respected yet for struct or union fields, and somewhat relatedly we still have #136 outstanding. In `docs/known-limitations.md` we mention missing GNU packed struct support--this should probably be clarified to describe exactly which meanings/contexts of `__attribute__((packed))` we support.

PR #354, opened 2021-12-03
c2rust-ast-builder: add dependency cargo

milahu:
fix #332 error: ```can't find crate for `rustc` ```

PR #355, opened 2021-12-03
ci: test only on arch linux

milahu:
reduce the test madness

* makes no sense to test c2rust on 6 "different" linux distros
* darwin loves to hang and block the linux tests &rarr; bye bye darwin

todo ...

* produce debug builds, not release builds. debug builds are much faster
* avoid the "compile a million dependencies" madness
  * https://bheisler.github.io/post/efficient-use-of-travis-ci-cache-for-rust/
  * https://blog.travis-ci.com/2021-09-03-rust
* avoid the "lets run everything in docker" madness


thedataking:
It is correct that the Darwin build sometimes fail due to networking issues. At the same time, it is easy to write code that works on Linux but fails on Darwin and our CI setup can catch that. For example, PR #354 seems to hit a compiler error on that platform:

```
/Users/runner/work/1/s/c2rust-ast-exporter/src/AstExporter.cpp:589:68: error: no member named 'isRValue' in 'clang::Expr'
        encode_entry_raw(ast, tag, ast->getSourceRange(), ty, ast->isRValue(), isVaList,
                                                              ~~~  ^
/Users/runner/work/1/s/c2rust-ast-exporter/src/AstExporter.cpp:2642:25: error: calling a protected constructor of class 'clang::tooling::CommonOptionsParser'
    CommonOptionsParser OptionsParser(argc_, argv_.data(), MyToolCategory);
                        ^
/usr/local/Cellar/llvm/13.0.0_1/include/clang/Tooling/CommonOptionsParser.h:76:3: note: declared protected here
  CommonOptionsParser(
  ^
2 errors generated.
```

milahu:
> PR #354 seems to hit a compiler error on that platform
>
> `error: no member named 'isRValue' in 'clang::Expr'`

sounds more like a version mismatch = api change

```sh
# c2rust/scripts/provision_mac.sh
brew install llvm@11

# c2rust/scripts/provision_deb.sh
apt-get install -qq clang-7 libclang-7-dev 
```

on my machine, the build fails just before that step, so cannot debug that right now
(build fails, cos nix does offline install, so cannot download tinycbor)

fw-immunant:
The mentioned Darwin failure here does look to be a LLVM version issue--for some reason brew installs both `llvm@11` and `llvm` (which is 13.0.0), and Clang 13 has removed `clang::Expr::isRValue`.

thedataking:
> The mentioned Darwin failure here does look to be a LLVM version issue--for some reason brew [installs](https://dev.azure.com/immunant/c2rust/_build/results?buildId=1480&view=logs&jobId=24cae736-ab81-5834-0476-aadc3600cb65&j=24cae736-ab81-5834-0476-aadc3600cb65&t=48582f96-fe01-5767-45d0-0612462fe3a5) both `llvm@11` and `llvm` (which is 13.0.0), and Clang 13 has removed `clang::Expr::isRValue`.

Please see `feature/ci-dev` for fixes to those issues. At one point in time, I had to pin LLVM to `llvm@11` but that is no longer necessary. The reason I didn't merge my changes is that I'm seeing unit tests fail.

thedataking:
Thanks for the contribution, there are some great ideas in there such as using caching to avoid compiling all the dependencies all the time (which should be tracked as a separate issue). However, we won't be reducing the supported/tested platforms down to just Arch Linux.

PR #368, opened 2022-02-07
Remove cross checking functionality

thedataking:


ahomescu (approved):
Do we have a tag for the commit before this?

PR #375, opened 2022-03-22
Limited support for va_list in structs

thedataking:
Trying to transpile apache httpd 2.4.52 uncovered a pattern in `util_script.c` that the transpiler cannot handle. This patch (partially) addresses this limitation by supporting va_lists in structs. Fully supporting the pattern found in apache httpd will likely require substantial engineering effort.

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -553,6 +553,12 @@ impl Builder {
         Builder { span: span, ..self }
     }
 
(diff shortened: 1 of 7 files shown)


fw-immunant (approved):
Looking this over, it seems broadly sane. I agree the OOB way the `'a` lifetime shows up in two places is less than ideal, but it would definitely take more design to do something cleaner and more general here.

This is passing the added test, and I have no particular issues with it. Should we merge or are more eyes needed?

PR #377, opened 2022-04-13
Fix mdbook manual after deprecating c2rust-refactor

rinon:
Now that we no longer build c2rust-refactor, we need to fix the manual to not depend on it to build. Also removes outdated references to cross-checks and fixes build issues with the new pulldown-cmark-to-cmark crate (by removing the utility that depended on it, since we no longer need it).


manual/SUMMARY.md
```diff
@@ -23,7 +25,7 @@
     - [genann](examples/genann/README.md)
     - [lil](examples/lil/README.md)
     - [xzoom](examples/xzoom/README.md)
-    - [robotfindskitten](examples/robotfindskitten/README.md)
+    <!-- - [robotfindskitten](examples/robotfindskitten/README.md) -->
```
fw-immunant, line 28: I noticed this wasn't tested by `scripts/test_examples.py` in the c2rust repo. What's the status of this example?

fw-immunant (approved):
LGTM, left one question but not blocking merge.

manual/SUMMARY.md
rinon: The example was primarily for testing the refactoring engine, which is now deprecated.

PR #378, opened 2022-04-15
c2rust-transpile/c_ast/conversion: workaround for unknown node

orangecms:
Not exactly sure what would need to be done here. Feedback would help a lot! :-)

`c2rust transpile compile_commands.json` yielded:

```
thread 'main' panicked at 'Type conversion not implemented for TagTypeUnknown expecting 3', c2rust-transpile/src/c_ast/conversion.rs:795:22
```

Sorry for the big change; it looks like the code was unformatted, and my setup runs `rustfmt` automatically.


thedataking:
Did you isolate the code construct that causes the problem? Suggestions on how to proceed:

1. Provide a code snippet that triggers the issue in the PR description (or at least info on how to reproduce)
2. If you want to reformat the code, that should be a separate, non-functional commit.

fw-immunant:
It's more likely that what you want to do here (instead of adding a case for `TypeTag::TagTypeUnknown` to the match on `ty_node.tag` in `ConversionContext::visit_node`) is add a case to `VisitBuiltinType` in `c2rust-ast-exporter/src/AstExporter.cpp`. It may help to print out some debugging info in the `default` case there first, to find out what the `kind` of the unhandled builtin type you're encountering is. You may have to consult clang's [`BuiltinTypes.def`](https://github.com/llvm/llvm-project/blob/main/clang/include/clang/AST/BuiltinTypes.def) (or, less likely, one of the other `.def` files mentioned [here](https://clang.llvm.org/doxygen/Type_8h_source.html#l02502)) to decode the integral kind value to something human-readable.

fw-immunant:
Closing as the root cause of this bug may have been resolved by now and if not, when the issue recurs it'll come with more debugging information to distinguish it from issues we've already solved. Please file an issue if you still encounter issues with `master`.

PR #379, opened 2022-04-20
ast-exporter: fix: No such file llvm-config

milahu:
use `or_else` instead of `or`

`or` evaluates all branches
`or_else` stops after first success

in my case, this branch fails

```rs
                // Relative to LLVM_LIB_DIR
                env::var("LLVM_LIB_DIR").ok().map(|d| {
                    String::from(
                        Path::new(&d)
                            .join("../bin/llvm-config")
                            .canonicalize()
                            .unwrap()
                            .to_string_lossy(),
                    )
                })
```

so `cargo build` throws

```
error: failed to run custom build command for `c2rust-ast-exporter v0.15.0 (/tmp/c2rust/c2rust-ast-exporter)`

Caused by:
  process didn't exit successfully: `/tmp/c2rust/target/debug/build/c2rust-ast-exporter-15a5bcc950a5a7b0/build-script-build` (exit status: 101)

  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', c2rust-ast-exporter/build.rs:253:30
```

### DIFF
--- a/c2rust-ast-exporter/build.rs
+++ b/c2rust-ast-exporter/build.rs
@@ -214,9 +214,10 @@ impl LLVMInfo {
         fn find_llvm_config() -> Option<String> {
             // Explicitly provided path in LLVM_CONFIG_PATH
             env::var("LLVM_CONFIG_PATH")
-                .ok()
+            .ok()
+            .or_else(|| {
                 // Relative to LLVM_LIB_DIR
-                .or(env::var("LLVM_LIB_DIR").ok().map(|d| {
+                env::var("LLVM_LIB_DIR").ok().map(|d| {
                     String::from(
                         Path::new(&d)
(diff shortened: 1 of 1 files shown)


milahu:
> possible if you have an unusual LLVM installation layout.

yes, i forgot to mention, this is on nixos, where llvm/lib and llvm/bin are in different prefixes

```
/nix/store/sfydnaab0wxn2qm3pkaab5x1fcagzxpf-llvm-11.1.0-lib/lib/libLLVM-11.so
/nix/store/w8p79yfq5059141frg9p4z17nr6s999w-llvm-11.1.0-dev/bin/llvm-config
```

thanks!

PR #383, opened 2022-04-25
Target-dependent translation of inline asm

fw-immunant:
This would benefit greatly from feedback on how it's detecting target architecture, which is currently pretty hacky and completely undocumented, and will also need some consideration on how we can test target-architecture-specific functionality in CI (I want to make sure installing toolchains for multiple architectures doesn't slow down CI as a whole too much, and we need to have a way to validate translation of programs for non-host architectures, perhaps without running them).


c2rust-transpile/src/lib.rs
```diff
@@ -218,9 +219,32 @@ fn get_module_name(
     file.to_str().map(String::from)
 }
 
+fn target_arch_from_args(clang_args: &[& str]) -> Option<String> {
```
ahomescu, line 222: Clang knows the target for the original source file, so couldn't we export that from `AstExporter` instead?

fw-immunant: Good call, done.

fw-immunant:
I've fixed a couple more bugs and added tests which will also test other architectures (just arm for now) as much as their runtime environment permits.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -101,7 +131,24 @@ fn parse_constraints(mut constraints: &str) ->
...
+            if !(is_explicit_reg || is_tied) {
+                // Attempt to parse machine-specific constraints
+                if let Some((machine_constraints, is_mem)) =
+                translate_machine_constraint(&*constraints, arch) {
+                    constraints = machine_constraints.into();
+                    mem_only = is_mem;
+                } else {
+                    warn!("Did not recognize inline asm constraint: {}\n\
+                    It is likely that this will cause compilation errors or \
+                    incorrect semantics in the translated program; please
```
rinon, line 147: Do you need the escapes? If so it's missing on this line.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -262,13 +391,15 @@ fn asm_is_att_syntax(asm: &str) -> bool {
     } else if asm.contains("word ptr") {
         false
     } else {
-        asm.contains('$')
+        // Guess based on sigils used in AT&T assembly. This would be more
+        // robust if it stripped comments first.
+        asm.contains('$') || asm.contains('%') || asm.contains('(')
```
rinon, line 396: I'm a bit concerned that '(' is too broad and will end up with false positives.

rinon (approved):
LGTM, but maybe Andrei should take another look because I haven't really looked at inline asm in a while.

c2rust-transpile/src/translator/assembly.rs
fw-immunant: The escape prevents the leading whitespace on the next line from being part of the string, so it is needed. Will fix.

c2rust-transpile/src/translator/assembly.rs
fw-immunant: '(' is intended to pick out `disp(base,idx,scale)` address calculations in AT&T syntax. Is there anywhere it can appear in Intel syntax? The only places I can think of are comments and macros. It's not hard to remove comments, but parsing macros seems like it's more trouble than we should bother with.

fw-immunant: We now remove comments before searching for these sigils.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -186,7 +314,8 @@ fn reg_is_reserved(constraint: &str) -> Option<(&str, &str)> {
 ///
 /// Modifies operands and returns a pair of prefix and suffix strings that
 /// should be appended to the assembly template.
-fn rewrite_reserved_reg_operands(att_syntax: bool, operands: &mut Vec<BidirAsmOperand>) -> (String, String) {
+fn rewrite_reserved_reg_operands(att_syntax: bool, arch: Arch,
```
ahomescu, line 317: Run rustfmt? This doesn't look like how it would break this line.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -239,6 +368,35 @@ fn rewrite_reserved_reg_operands(att_syntax: bool, operands: &mut Vec<BidirAsmOp
     (prolog, epilog)
 }
 
+/// Removes comments from an x86 assembly template.
+fn remove_comments(mut asm: &str) -> String {
```
ahomescu, line 372: Could we keep the comments and prepend them to the Rust assembly statement?

tests/asm.arm/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
```
ahomescu, line 1: Could you also add an arm64 test? I think people might care about that more than 32-bit.

c2rust-transpile/src/translator/assembly.rs
fw-immunant: The comment-stripped asm is only used for the "AT&T or Intel" heuristic. I can add a comment to clarify that.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -117,27 +164,102 @@ fn parse_constraints(mut constraints: &str) ->
     Ok((mode, mem_only, constraints))
 }
 
+fn is_regname_or_int(parsed_constraint: &str) -> bool {
+    parsed_constraint.contains('"') ||
+        parsed_constraint.starts_with(|c: char| c.is_ascii_digit())
+}
+
+/// Translate an architecture-specific assembly constraint from llvm/gcc
```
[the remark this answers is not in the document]
fw-immunant: According to https://llvm.org/docs/LangRef.html#constraint-codes, constraint codes generally match between LLVM and GCC, but [GCC's semantics are considered canonical](https://llvm.org/docs/LangRef.html#supported-constraint-code-list).

The assembly template we get has been processed by Clang and c2rust-ast-exporter, and has register references prefixed with dollar signs, so I believe it's LLVM's format. But I don't know where to find docs on that format itself, so I've mostly developed the parsing here by trial-and-error.

tests/asm.arm/Cargo.toml
fw-immunant: Will do.

c2rust-transpile/src/translator/assembly.rs
fw-immunant: Done.

tests/asm.arm/Cargo.toml
fw-immunant: Done.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -291,6 +498,12 @@ fn rewrite_asm<F: Fn(&str) -> bool>(asm: &str, is_mem_only: F) -> String {
             continue;
         }
 
+        // Note empty chunks
+        if chunk == "" {
```
ahomescu, line 502: Minor nit: `chunk.is_empty()` could have worked too.

ahomescu (approved):

PR #385, opened 2022-04-27
Export attributes on non-canonical RecordDecls

### LINKED ISSUE #384: Translation drops the packed attribute if attached to non-canonical decls in the C input
Issue discovered by transpiling `lighttpd-1.4.64` but it likely applies to all applications that call the `epoll_*` functions.

```C
int epoll_ctl(int epfd, int op, int fd, struct epoll_event *event);
int epoll_wait(int epfd, struct epoll_event *events, int maxevents, int timeout);
```

These functions take a pointer to a `struct epoll_event` which is defined in `sys/epoll.h`. glibc provides this definition for i386 and x86_64 architectures

```C
/* defined in  arch-specific header */
#define __EPOLL_PACKED __attribute__ ((__packed__))
/* defined in sys/epoll.h */
struct epoll_event
{
  uint32_t events;	/* Epoll events */
  epoll_data_t data;	/* User data variable */
} __EPOLL_PACKED;
```

The `libc` Rust crate defines `epoll_event` like so

```Rust
#[cfg_attr(
        any(
            all(
                target_arch = "x86",
                not(target_env = "musl"),
                not(target_os = "android")),
            target_arch = "x86_64"),
        repr(packed))]
    pub struct epoll_event {
        pub events: u32,
        pub u64: u64,
    }
```

Transpiling `lighttpd`, we get the following:

```Rust
#[derive(Copy, Clone)]
#[repr(C)]
pub struct epoll_event {
    pub events: uint32_t,
    pub data: epoll_data_t,
}
```

Which is missing the `#[repr(packed)]`. This causes a pointer to get truncated which ultimately causes a segfault. We have two options to avoid this issue:

1. Improve detection of when structures need to be `packed`.
2. Detect when we are using types from the standard library and use those instead of emitting our own.

thedataking:
This fixes an issue where the translated Rust structures miss the
lighttpd and (presumably) other programs that use the epoll APIs.

Closes #384

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -1912,9 +1912,21 @@ class TranslateASTVisitor final
      */
     bool VisitRecordDecl(RecordDecl *D) {
         if (!D->isCanonicalDecl()) {
-            // Emit non-canonical decl so we have a placeholder to attach comments to
+            // Emit non-canonical decl so we have a placeholder to attach comments to.
+            // Attributes may also be attached to the non-canonical declaration so
+            // we emit them too.
             std::vector<void *> childIds = {D->getCanonicalDecl()};
-            encode_entry(D, TagNonCanonicalDecl, D->getLocation(), childIds, QualType());
+            encode_entry(D, TagNonCanonicalDecl, D->getLocation(), childIds, QualType(),
+                [D](CborEncoder *local) {
+                // 1. Attributes stored as an array of attribute names
+                CborEncoder attrs;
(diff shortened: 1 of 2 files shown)


fw-immunant (approved):
This LGTM, but I had to go digging to see that the for-loop that just scans for "packed" is what we do in several places already and probably the right thing to do here. If we factored that loop out of the 3 places we have it, it would clarify that we're intentionally only looking at the "packed" attribute because the other possible attributes that structs have are passed through their own CBOR encoding.

Relatedly, I guess we don't support aligned unions?

thedataking:
Excellent points. I've extracted the duplicated packed attribute code into its own function and extended the patch to support Unions too. Side note: it seems that enums can have attributes too; not sure we handle those at all.

PR #386, opened 2022-04-28
Enable compiling c2rust on stable

fw-immunant:
I was reviewing our usage of nightly features and it turns out that the transpiler doesn't have a meaningful need for any nightly features. This PR compiles with rustc 1.58 stable.

Our tests don't pass with a stable compiler because the generated code contains `#![feature]`, but I think it's worthwhile to enable transpiling with stable as it's pretty easy to do so.


rinon (approved):
LGTM :tada:

PR #391, opened 2022-05-05
Unfork `prettyplease`

fw-immunant:
We vendored and forked `prettyplease` because it has an overly minimal API that did not expose some prettyprinting entrypoints we wanted, and it needed a patch to be able to emit named C-variadics. We can work around the former and the latter was fixed in version 0.1.9, so we can now unfork/unvendor this dependency. I think the somewhat ugly glue code for pretty-printing exprs/statements/patterns/paths is worth it in order to be able to reference `prettyplease` via crates.io, as the glue is self-contained and relatively unlikely to break. That said, before merging, we should add tests for said glue code to verify that it strips the right context.


fw-immunant:
Tests added.

rinon (approved):
Well, that's a _creative_ way to do this. Works for me!

Issue #390, opened 2022-05-03
Renumber operand references in assembly template strings in the presence of tied operands


fw-immunant:
When we generate translate inline asm to `asm!`, we merge tied inputs and outputs into single inout arguments. We use this merging when we rewrite the argument list, but we do not currently renumber by-index operand references to account for the changes to the argument list.

For example, this test case currently fails if uncommented, as the asm template string ends up referencing operand 1 while we tie the two operands together resulting in only operand 0 existing in the translated `asm!` expression:

https://github.com/immunant/c2rust/blob/65288c66edecd927397b32eb17ebf27edf2b7585/tests/asm.aarch64/src/asm.c#L8-L13

We'll want to fix this in [`rewrite_asm`](https://github.com/immunant/c2rust/blob/65288c66edecd927397b32eb17ebf27edf2b7585/c2rust-transpile/src/translator/assembly.rs#L478), which currently does not have access to the tied operands that we determine elsewhere.

PR #392, opened 2022-05-06
map input operand index to tied output index in asm string

aneksteind:
addresses #390

### DIFF
--- a/c2rust-transpile/src/translator/assembly.rs
+++ b/c2rust-transpile/src/translator/assembly.rs
@@ -472,10 +472,50 @@ fn asm_is_att_syntax(asm: &str) -> bool {
     }
 }
 
+/// If applicable, maps the index of an input operand to the
+/// output operand it is tied to.
+///
+/// Utilizes the fact that GNU inline assembly specifies all
+/// operands in a particular order of [output1, ... outputN]
+/// followed by [input1, ..., inputN] where the indices for all
+/// input operands are indexed relative to the size of the
+/// output operand sequence.
+fn map_input_op_idx(
+    idx: usize,
+    num_output_operands: usize,
+    tied_operands: &HashMap<(usize, bool), usize>,
+) -> usize {
+    if let Some(adj_idx) = idx.checked_sub(num_output_operands) {
+        // will only be Some(idx) if it was an input operand
+        match tied_operands.get(&(adj_idx, false)) {
+            Some(&out_idx) => {
+                return out_idx;
+            }
+            None => {
+                // get number of tied inputs before this index
+                // TODO: can calculate this once before the call, not once for each input
+                let num_tied_before = tied_operands
+                    .keys()
+                    .filter(|&&(iidx, is_out)| !is_out && iidx < adj_idx)
+                    .count();
+                // shift the original index by the number of tied input operands prior to the current one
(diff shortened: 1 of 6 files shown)


c2rust-transpile/src/translator/assembly.rs
```diff
@@ -472,10 +472,55 @@ fn asm_is_att_syntax(asm: &str) -> bool {
...
 /// by translating its references to operands (of the form $0 or $x0) to {0} or
 /// {0:y} (and wrapping mem-only references in square brackets).
-fn rewrite_asm<F: Fn(&str) -> bool>(asm: &str, is_mem_only: F, arch: Arch) -> String {
+fn rewrite_asm<F: Fn(&str) -> bool>(
+    asm: &str,
+    num_output_operands: usize,
+    tied_operands: &HashMap<(usize, bool), usize>,
+    is_mem_only: F,
+    arch: Arch,
+) -> String {
```
fw-immunant, line 523: Given that `num_output_operands` and `tied_operands` are only used to remap given operand reference indices, I think it would be cleaner to accept an `Fn` closure that remaps operand indices here.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -472,10 +472,55 @@ fn asm_is_att_syntax(asm: &str) -> bool {
...
+/// Utilizes the fact that GNU inline assembly specifies all
+/// operands in a particular order of [output1, ... outputN]
+/// followed by [input1, ..., inputN] where the indices for all
+/// input operands are indexed relative to the size of the
+/// output operand sequence.
+fn remapped_idx(
+    idx_str: &str,
+    num_output_operands: usize,
+    tied_operands: &HashMap<(usize, bool), usize>,
+) -> String {
```
fw-immunant, line 487: I think this would compose better if it took `idx: usize` (rather than `&str`) and returned a `usize`, as this shrinks the domain and pushes parse errors back into the parser in `rewrite_asm` (which might want to return a `Result`). We really don't want malformed asm to cause the whole transpiler to fail if we can avoid it.

fw-immunant (changes requested):
A couple comments inline. Also please add a test case to `asm.x86_64` that exercises tied operands so we have CI coverage all the way through transpilation, compilation, and run.

aneksteind:
> A couple comments inline. Also please add a test case to `asm.x86_64` that exercises tied operands so we have CI coverage all the way through transpilation, compilation, and run.

The trouble with x86_64 tests is that they get run in the pipeline and the Rust output in `assembly.rs` is incorrect, so many non-trivial examples don't compile. I've added a trivial example for now and will look into either fixing it or contriving a more complicated example that compiles.

fw-immunant (approved):
The trivial example is a bit more complex than it has to be--the following seems more minimal (no `dummy` or second `add`):
```
int mul2(int var64)
{
    int out;
    asm("add %0, %1\n\t"
        : "=r"(out)
        : "0"(var64));
    return out;
}
```

Otherwise, looks good. Add the more minimal example above to the tests and then we can merge.

aneksteind:
I added your simpler example in addition to the existing example, because the latter exercises the mapping of an untied operand as well.

PR #396, opened 2022-05-10
Patch tinycbor with upstream fix for build

rinon:
This is a follow-up from #389 because I can't push to the branch that PR was created from. Adds the upstream build system fix as a patch to our build of tinycbor, which should allow both recursive make and cargo invocation of the build.


chrysn:
This resolves #389 in my tests, thank you.

I don't quite understand why the ubuntu test failed, but I had to do a `cargo clean` when applying this, so maybe that can be resolved by removing the caches.

PR #404, opened 2022-05-13
asm: Ignore the `i` constraint

chrysn:
RISC-V code that contains things like

```c
    __asm__ volatile (
        "csrrc %[dest], mstatus, %[mask]"
        :[dest]    "=r" (state)
        :[mask]    "i" (MSTATUS_MIE)
        : "memory"
        );
```

currently doesn't recognize the `i` constraint, and turns the code into

```rust
    asm!(
        "csrrc {0}, mstatus, {1}", lateout(reg) state, inlateout(i) MSTATUS_MIE => _,
        options(preserves_flags)
    );
```

which fails to compile with errors like

```
error: invalid register class `i`: unknown register class
    --> .../out/riot_c2rust_replaced.rs:1576:56
     |
1576 |         "csrrc {0}, mstatus, {1}", lateout(reg) state, inlateout(i) MSTATUS_MIE => _,
     |       
```

The attached patch ignores the `i` constraint -- [according to GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Simple-Constraints.html#Simple-Constraints) it means "An immediate integer operand (one with constant value) is allowed". The Rust assembly not only has no constraint for that, it appears to allow that by default -- after applying that patch, it generated code like `csrrci	s3,mstatus,8` which is the immediate version of `csrrc`.

### DIFF
--- a/c2rust-transpile/src/translator/assembly.rs
+++ b/c2rust-transpile/src/translator/assembly.rs
@@ -142,6 +142,11 @@ fn parse_constraints(
         "r" => {
             constraints = "reg".into();
         }
+        "i" => {
+            // Rust inline assembly has no constraint for that, but uses the argument as an
+            // immediate value anyway
+            constraints = "reg".into();
+        }
         _ => {
             let is_explicit_reg = constraints.starts_with('"');
             let is_tied = !constraints.contains(|c: char| !c.is_ascii_digit());
@@ -271,9 +276,6 @@ fn translate_machine_constraint(constraint: &str, arch: Arch) -> Option<(&str, b
         },
         Arch::Riscv => match constraint {
             "f" => "freg",
-            /* Rust inline assembly has no constraint for that, but uses the argument as an
-            * immediate value anyway */
-            "i" => "reg",
             _ => return None,
         },
     };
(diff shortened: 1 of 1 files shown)


chrysn:
I'm not fully sure how to test this (or whether it's small / riscv asm support is experimental enough to not need an explicit test). I started creating one off tests/asm.aarch64 into tests/asm.riscv, but there's no libtest.a for that platform, and either way I wouldn't want to *run* that code (it's interrupt-disabling, that's only supported in a bare metal context anyway, which is also what the riscv32imac-unknown-none-elf tuple indicates), but only test if the produced assemblies are alike (or even just test if it does build).

fw-immunant (changes requested):
I think this is a simple enough change that validating against docs rather than having an explicit test is probably fine. However, because this is a simple constraint rather than a machine-specific constraint, it should be handled in [this match](https://github.com/immunant/c2rust/blob/4f9a30cace56a4d287dad3b3d86f636729207359/c2rust-transpile/src/translator/assembly.rs#L136-L167) in `parse_constraints` rather than `translate_machine_constraint`.

chrysn:
Thanks, updated. (I'm running it through another round of testing here, but I don't expect that anything comes up; if so, I'll have word on it before CI is done rerunning this here.)

fw-immunant (approved):

chrysn:
Thanks for reviewing and merging.

Would you consider publishing a 0.16.1 that includes this? That'd make installing a version that works for RIOT users a lot easier.

fw-immunant:
I'll bring it up in an upcoming internal discussion of c2rust development priorities.

Issue #397, opened 2022-05-10
Assembly translation doesn't account for defaults correctly


aneksteind:
GNU inline assembly default dialect is [ATT](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#x86-Options), while Rust's is [Intel](https://blog.rust-lang.org/inside-rust/2020/06/08/new-inline-asm.html). `asm_is_att_syntax` aims to default to the ATT dialect for the assembly string but the following last resort in checking ATT syntax will cause a simple assembly instruction, with no additional helpful context, to default to Intel syntax:

https://github.com/immunant/c2rust/blob/6b05591a149bb2460f706c3c2810bfd188a88767/c2rust-transpile/src/translator/assembly.rs#L471

An example would be:
```c
int six(void) {
    int out = 0;
    int zero = 0;
    int six = 6;

    asm("add %1, %0\n\t"
        : "=r"(out)
        : "r"(six), "0"(zero));

    return out;
}
```

One alternative approach would be to default to `true` but use commonly-used intel-syntax sigils such as `;` for comments, known registers specified without `%`, `[` for indirect, etc. to rule out ATT syntax

aneksteind:
The issue is that the following last resort in checking att syntax will cause a simple assembly instruction, with no additional helpful context, to default to intel syntax:

https://github.com/immunant/c2rust/blob/6b05591a149bb2460f706c3c2810bfd188a88767/c2rust-transpile/src/translator/assembly.rs#L471

An example would be:
```c
int six(void) {
    int out = 0;
    int zero = 0;
    int six = 6;

    asm("add %1, %0\n\t"
        : "=r"(out)
        : "r"(six), "0"(zero));

    return out;
}
```

PR #405, opened 2022-05-13
default to AT&T dialect, add hint for Intel

aneksteind:
addresses #397


fw-immunant (approved):
LGTM. At some point it might be worth trying to check for `-masm=<intel|att>` in the command line for the current file to determine the default behavior, but this is more correct in the absence of any flags.

aneksteind:
> LGTM. At some point it might be worth trying to check for -masm=<intel|att> in the command line for the current file to determine the default behavior, but this is more correct in the absence of any flags.

Agreed, I'm looking into something along those lines

Issue #400, opened 2022-05-11
inline assembly translator doesn't check dialect against host clang version


aneksteind:
Unlike GCC, inline assembly with Intel syntax is only supported in [clang versions >= 14](https://reviews.llvm.org/D113707). There should be a check that at least warns the user that their inline assembly may not be interpreted correctly if their clang version is less than 14.

PR #406, opened 2022-05-13
warn if clang version is less than 14 when translating inline asm with Intel dialect

aneksteind:
addresses #400


aneksteind:
closing because this is most relevant for the way in which C code in compiled and is consequential for any output assembly from clangs integrated assembler (the user could have used a different assembler)

PR #408, opened 2022-05-15
Fix warnings

thedataking:
Builds should complete cleanly so it is easier to spot new warnings and potential issues.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -18,10 +18,9 @@ use syn::{BinOp, UnOp}; // To override c_ast::{BinOp,UnOp} from glob import
 use crate::rust_ast::comment_store::CommentStore;
 use crate::rust_ast::item_store::ItemStore;
 use crate::rust_ast::set_span::SetSpan;
-use crate::rust_ast::traverse::Traversal;
 use crate::rust_ast::{pos_to_span, SpanExt, DUMMY_SP};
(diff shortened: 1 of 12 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -4502,10 +4501,7 @@ impl<'c> Translation<'c> {
             CExprKind::DeclRef(_, decl_id, _) if variants.contains(&decl_id) => {
                 return val.map(|x| match **unparen(&x) {
                     Expr::Cast(ExprCast { ref expr, .. }) => expr.clone(),
-                    _ => panic!(format!(
-                        "DeclRef {:?} of enum {:?} is not cast",
-                        expr, enum_decl
-                    )),
+                    _ => panic!("DeclRef {:?} of enum {:?} is not cast",expr, enum_decl),
```
fw-immunant, line 4504: rustfmt would add a space here.

c2rust/build.rs
```diff
@@ -2,9 +2,6 @@ use std::env;
 use std::path::PathBuf;
 use std::process::Command;
 
-use chrono::{Duration, NaiveDate};
-use rustc_version::Channel;
-
```
fw-immunant, line 7: We should also remove the corresponding dependencies from `c2rust/Cargo.toml`.

fw-immunant (approved):
Mostly LGTM, one missing space (we should put rustfmt in a git hook) and one place where we should remove now-unused deps from `Cargo.toml`.

PR #413, opened 2022-05-21
fix incorrect generation of `'static` lifetime name

afetisov:
The syn based parser would crash with the apostrophe prefixed lifetime name


fw-immunant (approved):
Strange Darwin failure in CI, but I don't think it could possibly be code-related; this looks like a good fix. Thank you!

PR #415, opened 2022-05-21
Keep original label names from C code

afetisov:
I have added an `Option<Rc<str>>` field to the `cfg::Label::FromC`. This allows to preserve the names of blocks which are provided in the C source, potentially increasing the readability of transpiled code and making it easier to match original and transpiled code blocks. The label names are used as control flow variable values when `--ddebug-labels` is enabled. Unfortunately, Relooper really likes to dump those names in the `_` branch of matches, bringing the automatic label names to the front. Still, there are some situations where the old labels can be traced in the transpiled code. This will also significantly increase readability if Relooper is changed to use labeled blocks & breaks instead of a match on variable: the C label names can be directly used as block label names.

I had to make Label non-Copy. This is not a painful transition, and arguably there was no reason to make it Copy in the first place, since it was likely to carry some non-Copy data.

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -3,6 +3,7 @@ use c2rust_ast_exporter::clang_ast::*;
 use failure::err_msg;
 use serde_bytes::ByteBuf;
 use std::collections::HashMap;
+use std::rc::Rc;
 use std::vec::Vec;
 
 use super::Located;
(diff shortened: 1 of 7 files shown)


fw-immunant (approved):
This looks good overall. My first thought was that it would be nice to keep `Label: Copy`, but that would mean accessing `label_names` in `Label::pretty_print`, and most callers of the latter don't have access to the `TypedAstContext`. So this approach seems sane and the implementation looks reasonable. The use of `Rc` here does seem worthwhile given how many `.clone()`s we end up introducing.

Most of the panics introduced seem to be only in places where we're relying on internal consistency of `label_names` with the set of AST locations where we expect names, so they shouldn't hurt overally reliability. The one exception is:

```rust
                    let label_name = from_value::<Rc<str>>(node.extras[0].clone())
                        .expect("unnamed label in C source code");
```

where we're relying on the serialized C AST having the structure we expect. This is the same pattern as all the adjacent code, so I'm satisfied with tests passing to validate it.

As far as improving output of translation goes, there are a few improvements in the output for our test cases, e.g. `gotos/irreducible.c` where this change now brings over the `l1` label. :+1:

PR #420, opened 2022-05-27
Updated c2rust crates to edition 2021.

kkysen:
There are quite minimal changes needed to move everything to edition 2021, and we're already pinned to a nightly that is after edition 2021 came out, so I don't see a reason not to update everything, and it's a pretty simple change.


fw-immunant (approved):
LGTM. I think you're right that with our version requirements this isn't a compat hazard.

PR #421, opened 2022-05-27
`c2rust_transpile::transpile` can only be called once; this is starting to fix it.

kkysen:
`c2rust_transpile::transpile` is a public library function, so it can be called by anyone, but calling it more than once panics.  This is because it initializes a global, init-once logger.

This adds a dependency on `log_reroute` so that the new logger can be swapped in every time.  Alternatively, the logging logic could be moved out of the library and into only the binary, but that seems like more work than this simple `log_reroute` change.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2712,7 +2712,6 @@ ExportResult *make_export_result(const Outputs &outputs) {
 // Extract clang AST for the source file specified in the argument vector.
 // Note: The arguments should only reference one source file at a time.
 Outputs process(int argc, const char *argv[], int *result) {
-    static uint64_t source_path_count = 0;
     auto argv_ = augment_argv(argc, argv);
     int argc_ = argv_.size() - 1; // ignore the extra nullptr
 
@@ -2729,9 +2728,14 @@ Outputs process(int argc, const char *argv[], int *result) {
 #endif
 
     // the logic below assumes we're only translating one source file
-    assert(OptionsParser.getSourcePathList().size() - 1 ==
-               source_path_count++ &&
-           "Expected exactly one source path");
+    static size_t source_path_count = 0;
+    source_path_count++;
+    const size_t num_sources = OptionsParser.getSourcePathList().size();
+    assert(
+        (num_sources == 1 // newer clang versions
+        || num_sources == source_path_count // older clang versions
+        ) && "Expected exactly one source path"
+    );
 
     // CommonOptionsParser is stateful so the vector returned by
     // getSourcePathList() includes paths from past invocations.
--- a/c2rust-transpile/Cargo.toml
+++ b/c2rust-transpile/Cargo.toml
@@ -26,6 +26,7 @@ indexmap = { version = "1.0.1", features = ["serde-1"] }
 itertools = "0.10"
 libc = "0.2"
 log = "0.4"
+log-reroute = "0.1"
 pathdiff = "0.2"
 proc-macro2 = "1.0"
 regex = "1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c5d78ce20460b82d3fa150275ed9d55e21064fc7951177baacf86a145c4a4b1f"
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -221,6 +227,7 @@ dependencies = [
  "itertools",
  "libc",
  "log",
+ "log-reroute",
  "pathdiff",
  "proc-macro2",
  "regex",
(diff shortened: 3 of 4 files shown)


kkysen:
I just also added a fix in `c2rust-ast-exporter` where only one source path is being evaluated at a time.
Previously, it assumed sources would accumulate in `OptionsParser.getSourcePathList()`
and not be cleared out between successive calls, so it maintained a static counter to check against.
But in clang 13 or 14, this changed, and now `OptionsParser.getSourcePathList().size() == 1`,
so the assert would fail whenever `process` was called multiple times,
or when `c2rust_transpile::transpile` was called multiple times.

This changes the code to handle both cases.
That is, either `OptionsParser.getSourcePathList().size() == 1`,
or `OptionsParser.getSourcePathList().size() == source_path_count`.

I tested this on clang/llvm versions 9 through 14.  Is that as far back as we need to support?

c2rust-transpile/Cargo.toml
```diff
@@ -26,6 +26,7 @@ indexmap = { version = "1.0.1", features = ["serde-1"] }
 itertools = "0.10"
 libc = "0.2"
 log = "0.4"
+log-reroute = "0.1"
```
aneksteind, line 29: any reason not to have this at the latest version 0.1.8?

aneksteind (approved):
lgtm after the assertion is gated by clang version, although I'm not sure about minimum supported clang version. we test llvm version 10 at least

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2729,9 +2728,14 @@ Outputs process(int argc, const char *argv[], int *result) {
...
 
     // the logic below assumes we're only translating one source file
-    assert(OptionsParser.getSourcePathList().size() - 1 ==
-               source_path_count++ &&
-           "Expected exactly one source path");
+    static size_t source_path_count = 0;
+    source_path_count++;
+    const size_t num_sources = OptionsParser.getSourcePathList().size();
+    assert(
+        (num_sources == 1 // newer clang versions
```
aneksteind, line 2735: could we gate this by clang major version similar to a few lines above?

c2rust-transpile/Cargo.toml
kkysen: I meant to not specify the patch version so it could upgrade without manually bumping it, like the `itertools`, `libc`, and `log` versions above.  Or should I specify the specific the version here?  It's 0.1.8 in `Cargo.lock`.

c2rust-ast-exporter/src/AstExporter.cpp
kkysen: I was thinking about this, but I thought it would be better to have the check based more on a feature test, in case anything else ever changes.  Doing it by clang version seemed more brittle to me, though I think it's an easy enough change.  I'd have to check if it was changed in clang 13 or 14, though.

c2rust-transpile/Cargo.toml
aneksteind: no i think thats a good reason to keep it as-is

c2rust-ast-exporter/src/AstExporter.cpp
aneksteind: if this fix is specifically a result of differences between clang versions, i think it'd make sense but i'm not sure i understand your point about a feature test. could you explain what you mean?

kkysen: By feature test, I meant something like a JS feature test testing for a feature you want to use/polyfill, rather than checking the browser version.  That is, it's usually better to test for the feature you actually care about (in this case if `OptionsParser.getSourcePathList()` accumulates all the paths across multiple calls) instead of testing indirectly through a version check that may not always be tied to the exact feature.  It could, for example, change in a different way or back in a later clang version.

So we check that either the paths accumulate by the `num_sources == source_path_count` check, and if that's not the case, we can check `num_sources == 1` instead.  That way we don't need to depend on clang version checks where we don't need to, which aren't inherently tied to what we really want to check.  Does that make more sense?

aneksteind: Yes that makes sense to me. Is it possible for this assertion to evaluate to true for either set of clang versions where it otherwise wouldn't with one of the two clauses compiled out? I.e. is the feature still meant to work when the clang version is new but only the second clause is true, or if it's old and only the first is true? Or does it not make a difference?

kkysen: Sorry, I'm not exactly sure what you're saying here.  The assertion logic should work regardless of clang version.  A valid (non-null) source path is always chosen and it's always the last one that is chosen.

aneksteind: It's alright. It doesn't seem like the scenario I'm thinking about can happen anyway.

aneksteind (approved):

fw-immunant (approved):
LGTM. I think it's kind of awful that we have a `fn init(...) -> ()` at all (*crying in RAII*), much less one that has hard-coded stringification of a foreign datatype and an ANSI format template string in it... but this is definitely an improvement.

kkysen:
Stephen had said we could also move all the logging setup logic (which is very global and un-RAII) to the binary crate, instead of the library one.  But I think this is the simplest fix for now.

PR #422, opened 2022-05-27
Got `cargo test` to run again.

kkysen:
`c2rust-bitfields/tests` had been blocking `cargo test` from working at all.  It seems like it should be in `tests` instead (though I'm not sure), since it needs to have `c2rust transpile` run on it first before it can link successfully.   So I just moved it to `c2rust-bitfields/c2rust-tests` so that `cargo test` doesn't try to compile it and fail.

With that unblocking `cargo test`, I was able to run `cargo test` again and fix all the errors, mostly in doctests, since it hadn't been tested in a while probably.

There are still some warnings from `bindgen` generated files (for nullptr derefs) in `c2rust-ast-exporter`, but those are due to `bindgen` and should be fixed in the next `bindgen` release.

Then I added a test than simply runs `scripts/test_translator.py tests`.  I'm not sure if we want this to keep being completely separate or if it's okay to just call it so it runs as part of `cargo test`.  Also, should any of the other `scripts/test_*.py` test scripts be run, too?


fw-immunant (changes requested):
Looks good overall, but I'm not sure about running `test_translator.py` from `cargo test` on the transpiler--that'll involve nested `cargo` invocations, which is a bit scary even if it happens to work here. I would say to split that out into a separate PR where folks can bikeshed it; keep the other changes here and add `cargo test` to the CI workflow so that this fix stays fixed and our CI coverage improves.

kkysen:
That makes sense.  Should I just add an `#[ignore]` to it, or revert that commit?

fw-immunant:
Just revert for now, thanks.

fw-immunant (approved):

Issue #410, opened 2022-05-17
Need support for tests that target ARM


aneksteind:
Tests such as `tests/asm.arm` and `tests/asm.aarch64` are unable to execute in the test pipeline because their inline assembly is meant to target ARM.

kkysen:
Can we run these tests using qemu?  And then cross-compile maybe using `zig cc` (through something like [`cargo zigbuild`](https://crates.io/crates/cargo-zigbuild))?

aneksteind:
> Can we run these tests using qemu?

Yes, we discussed this a while ago; the work just needs to be done. One alternative is to add an ARM image/container to the pipeline and test just the relevant tests, which would at least remove the need to cross compile. I don't have a preference which one to choose, probably whichever is easiest.

kkysen:
If we can easily test locally, too, that seems preferable.

PR #429, opened 2022-06-01
Add Support for Testing Non-Native Targets

kkysen:
Fixes https://github.com/immunant/c2rust/issues/410.

Added support to `scripts/test_translator.py` for cross-compiling using `cargo zigbuild` and `zig cc` and for emulation using `qemu`.

More specifically, `cargo zigbuild` is used in place of `cargo` for building and running the test crates.  It uses `zig cc` as the linker, which has seamless cross-compilation that just works.

`qemu` is used to emulate other targets and architectures.  These are small test programs, so this should be fine, and doing it this way allows all tests to be run locally, which is very convenient.

I wasn't sure how dependencies should be managed, but for this, I just had them installed in `test_translator.py` itself, assuming it's running Ubuntu Linux.  I assume there's a better way to do this that's more cross-platform, but I wasn't sure how to do that yet, so I just did it inline first to make things quick.  We can fix that next.

Specifically for dependencies,
* `cargo-zigbuild` is installed using `cargo install`
* `zig cc` is installed through `pip install ziglang` (added it to `requirements.txt`)
* `rustc` targets are installed through `rustup target add`
* `qemu` is installed through `apt` (`qemu-user`, `qemu-user-static`)
* `qemu` dynamic libraries are installed through `apt` (`gcc-${triple}`)
* `update-binfmts` (for enabling `qemu` `binfmt_misc` support) is installed through `apt` (`binfmt-support`)
* `qemu` `binfmt_misc` entries are added through `update-binfmts --enable`

Note, however, that the `asm.arm` and `asm.aarch64` tests, the ones I was trying to test, don't appear to work out of the box.  I'm not sure if the errors in them are due to something wrong in the cross compilation and emulation, are they are simply errors in the tests themselves since I assume they weren't being tested as regularly.  Specifically, `asm.arm` fails at runtime on an assertion, and `asm.aarch64` fails to compile, saying an asm argument is unused.

### DIFF
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -15,6 +15,8 @@ from typing import List, Callable
 
 import plumbum as pb
 
+from plumbum.machines import LocalCommand as Command
+
 
 class Colors:
     # Terminal escape codes
@@ -316,9 +318,6 @@ def _invoke(console_output, cmd, *arguments):
         msg = "cmd exited with code {}: {}".format(pee.retcode, cmd[arguments])
         logging.critical(pee.stderr)
         die(msg, pee.retcode)
-
-
-Command = pb.machines.LocalCommand
 
 
 def get_cmd_or_die(cmd: str) -> Command:
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -330,6 +330,17 @@ def get_cmd_or_die(cmd: str) -> Command:
         die("{} not in path".format(cmd), errno.ENOENT)
 
 
+def get_cmd_or_install(cmd: str, install_cmd: Command) -> Command:
+    """
+    lookup named command or install it first
+    """
+    try:
+        return pb.local[cmd]
+    except pb.CommandNotFound:
+        install_cmd()
+        return pb.local[cmd]
+
+
 def ensure_dir(path):
     if not os.path.exists(path):
         logging.debug("creating dir %s", path)
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -9,3 +9,4 @@ pygments
 typing;python_version<"3.5"
 scan-build
 pyyaml
+ziglang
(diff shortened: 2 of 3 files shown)


kkysen:
I moved the installation stuff to `scripts/provision_cross_tests.sh`, which I only run in `scripts/provision_deb.sh` now.  If installs  all the same stuff as before, but since it's not installed at use-site, it installs necessary things for all targets found in `target-tuple` files.

`scripts/test_translator.py` doesn't install anything anymore.  If `cargo-zigbuild` is available, it uses that (which uses `zig cc`), otherwise just normal `cargo`.  And it goes back to the logic of only running tests for targets if those (`rustup`) targets are already installed.

We can keep discussing whether to use `zig cc` or not, but I wanted to add these changes first, which are simpler, and see how it works in CI.

kkysen:
Removed any dependencies on `zig cc` and `cargo-zigbuild` now.  It now just uses the `gcc-${c_target}` `apt` packages for cross-compiling.

It does this by generating a `.cargo/config.toml` for every cross-compiling test.  This sets the correct linker, rustflags, and runner (`qemu-${arch}`, which means we don't need to mess with `binfmt_misc` either).  It also sets the default target, so we don't need to always pass `--target ${target}` everywhere.

`provision_cross_tests.sh` is also now split into just additions to `provision_deb.sh` and `provision_rust.sh`.

Also, the arm tests that are failing are temporarily fixed/skipped until we figure out how to actually fix them so that we don't cause CI failures.  Speaking of CI, these changes need to also be pushed to [`feature/ci-dev`](/immunant/c2rust/tree/feature/ci-dev) and have `./scripts/docker_build.sh push-all` be run.

kkysen:
https://github.com/immunant/c2rust/issues/433 is now fixed by https://github.com/immunant/c2rust/pull/441, so that's down to 1/2 issues blocking the `tests/asm.aarch64` test from working.

PR #435, opened 2022-06-07
Added a `.gitignore` under `tests/` to ignore files generated by `test_translator.py`

kkysen:
Pretty simple.  This just adds `tests/.gitignore` to ignore all the files generated by `test_translator.py` with `--keep all`.


fw-immunant (changes requested):
This is good to have in general, but the `build.rs` files in each test subdir are tracked in git, so we shouldn't gitignore those.

kkysen:
Ah, I missed that about `build.rs`, I'll change that to `!build.rs` instead.

fw-immunant (approved):

PR #437, opened 2022-06-08
Buffer read of `compile_commands.json`

fw-immunant:
Not a performance issue in any situation I've seen, but makes `strace` output cleaner which improves debugging.

### DIFF
--- a/c2rust-transpile/src/compile_cmds/mod.rs
+++ b/c2rust-transpile/src/compile_cmds/mod.rs
@@ -171,7 +171,7 @@ pub fn get_compile_commands(
     compile_commands: &Path,
     filter: &Option<Regex>,
 ) -> Result<Vec<LinkCmd>, Error> {
-    let f = File::open(compile_commands)?; // open read-only
+    let f = std::io::BufReader::new(File::open(compile_commands)?); // open read-only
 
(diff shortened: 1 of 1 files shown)


kkysen:
Can we just do a `fs::read` and read it in one `read`?  I imagine `compile_commands.json` won't ever be too big to do that.

fw-immunant:
> Can we just do a `fs::read` and read it in one `read`? I imagine `compile_commands.json` won't ever be too big to do that.

I've seen hundreds-of-MB `compile_commands.json` files in C/C++ projects before, and this fix isn't any more complex than `fs::read` would be, so I think this is preferable (if just slightly).

kkysen:
I don't really think it matters which we use here, but 100s of MBs wouldn't be too big, and `fs::read` and `serde_json::from_slice` is usually about 10x faster than `BufReader` and `serde_json::from_reader`.

fw-immunant:
> I don't really think it matters which we use here, but 100s of MBs wouldn't be too big, and `fs::read` and `serde_json::from_slice` is usually about 10x faster than `BufReader` and `serde_json::from_reader`.

I see, I wasn't aware of [serde-rs/json#160](https://github.com/serde-rs/json/issues/160). Feel free to issue a follow-up PR.

Issue #433, opened 2022-06-07
`c2rust transpile` always crashes on `aarch64-unknown-linux-gnu` using clang 10


kkysen:
When linked to clang 10 (and 7, 8, and 9, all the version supported on Ubuntu 20), as in the Docker image and CI, `c2rust transpile` always crashes on `aarch64-unknown-linux-gnu`.  This includes crashing on a completely empty C file and a minimal `compile_commands.json`, as well as any other C files I test.  It does not happen in clang >10, but it does in clang 10, 9, 8, and 7, and I'm not sure about <7.

For example, using this compilation database at `tests/empty.aarch64/compile_commands.json` (you need to adjust the `directory` since it must be an absolute path):
```json
[
  {
    "arguments": [
      "cc",
      "-c",
      "-target",
      "aarch64-linux-gnu",
      "empty.c"
    ],
    "directory": "/home/kkysen/work/rust/c2rust/tests/empty.aarch64",
    "file": "empty.c"
  }
]
```
and running:
```shell
> touch tests/empty.aarch64/empty.c
> LLVM_CONFIG_PATH=llvm-config-10 RUST_BACKTRACE=1 cargo run transpile tests/empty.aarch64/compile_commands.json
    Finished release [optimized] target(s) in 0.17s
     Running `target/release/c2rust transpile tests/empty.aarch64/compile_commands.json`
warning: c2rust: Encountered unsupported BuiltinType kind 48 for type __SVInt8_t
warning: c2rust: Encountered unsupported BuiltinType kind 49 for type __SVInt16_t
warning: c2rust: Encountered unsupported BuiltinType kind 50 for type __SVInt32_t
warning: c2rust: Encountered unsupported BuiltinType kind 51 for type __SVInt64_t
warning: c2rust: Encountered unsupported BuiltinType kind 52 for type __SVUint8_t
warning: c2rust: Encountered unsupported BuiltinType kind 53 for type __SVUint16_t
warning: c2rust: Encountered unsupported BuiltinType kind 54 for type __SVUint32_t
warning: c2rust: Encountered unsupported BuiltinType kind 55 for type __SVUint64_t
warning: c2rust: Encountered unsupported BuiltinType kind 56 for type __SVFloat16_t
warning: c2rust: Encountered unsupported BuiltinType kind 57 for type __SVFloat32_t
warning: c2rust: Encountered unsupported BuiltinType kind 58 for type __SVFloat64_t
[...]

PR #441, opened 2022-06-10
Handle arm64 Scalable Vector Extension (SVE) types on clang 10

kkysen:
Fixes https://github.com/immunant/c2rust/issues/433.

This correctly handles arm64 Scalable Vector Extension (SVE) types on clang 10.  SVE types were added in clang 10, but the API we had used to handle them was only added in clang 11, so they were being skipped in clang 10 and causing errors when transpiling any C file on `aarch64-unknown-linux-gnu` due to the SVE types not being recognized.  In clang <10, SVE types are not yet present, so this is not an issue there.  Now the `aarch64-unknown-linux-gnu` should work in CI, which uses clang 10.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -31,9 +31,6 @@
 #include "clang/Basic/LangStandard.h"
 #endif // CLANG_VERSION_MAJOR < 10
 #include "clang/Tooling/Tooling.h"
-#if CLANG_VERSION_MAJOR < 11
-// TODO(kkysen) CodeGenTypes
-#endif // CLANG_VERSION_MAJOR < 11
 
 #include "AstExporter.hpp"
 #include "ExportResult.hpp"
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -322,7 +322,6 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
     }
 
     void VisitBuiltinType(const BuiltinType *T) {
-        TypeTag tag;
         auto kind = T->getKind();
 
 #if CLANG_VERSION_MAJOR >= 10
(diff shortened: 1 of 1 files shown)


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -319,73 +319,97 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
...
             auto Info = Context->getBuiltinVectorTypeInfo(T);
-            auto t = Info.ElementType;
-            auto qt = encodeQualType(t);
+            auto ElemType = Info.ElementType;
 
-            #if CLANG_VERSION_MAJOR >= 12
-            auto ElemCount = Info.EC.getKnownMinValue();
-            #else
+#if CLANG_VERSION_MAJOR >= 12
+            auto ElemCount = Info.EC.getKnownMinValue() * Info.NumVectors;
```
ahomescu, line 333: Was the missing `NumVectors` a bug in the old code?

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -319,73 +319,97 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
...
-        case BuiltinType::UChar:      tag = TagUChar;       break;
-        case BuiltinType::Char_U:     tag = TagChar;        break;
-        case BuiltinType::Char_S:     tag = TagChar;        break;
-        case BuiltinType::Void:       tag = TagVoid;        break;
-        case BuiltinType::Bool:       tag = TagBool;        break;
-        case BuiltinType::WChar_S:    tag = TagSWChar;      break;
-        case BuiltinType::WChar_U:    tag = TagUWChar;      break;
-        }
-        // clang-format on
+        const TypeTag tag = [&] {
```
ahomescu, line 375: Is there anything changed semantically here? Looks more like a style change, these should generally be separate patches but I'll allow this one.

ahomescu (approved):

c2rust-ast-exporter/src/AstExporter.cpp
fw-immunant: No--before, we multipled by `Info.NumVectors` inside the call to `cbor_encode_uint`; now since we need to calculate sizes separately for each Clang version range, `ElemCount` itself accounts for `NumVectors` so we can consistently capture the same set of variables in the closure.

c2rust-ast-exporter/src/AstExporter.cpp
fw-immunant: I don't think dbe85a1f36dc4f1145e2fd345fd39a34375b88c4 changes any semantics, no.

kkysen: No, that part was just a pure refactor to make the code a bit more concise.

PR #446, opened 2022-06-15
Remove unused files and directories

thedataking:
I'm not aware of any uses of the vagrant environment any longer.
We can also get rid of .travis.yml since we're no longer using
Travis CI for testing.


kkysen:
We should also remove mentions of vagrant from the other files, right?
* `docs/README-developers.md`
* `manual/SUMMARY.md`
* `azure-pipelines.yml`.

kkysen:
"three" is changed to "two" now.

fw-immunant (approved):

Issue #449, opened 2022-06-15
`c2rust transpile` crashes on `long double` => `f128`


kkysen:
When running the `tests/longdouble` test as below, `c2rust transpile` crashes:
```shell
> ./scripts/test_translator.py tests/ --only-directories longdouble --test-longdoubles
longdouble:
 [ FAILED ] translate long_double.c
thread 'main' panicked at '"1.3" is not a valid Ident', /home/kkysen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.38/src/fallback.rs:701:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/library/core/src/panicking.rs:143:14
   2: proc_macro2::fallback::validate_ident
             at /home/kkysen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.38/src/fallback.rs:701:9
   3: proc_macro2::fallback::Ident::_new
             at /home/kkysen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.38/src/fallback.rs:642:9
   4: proc_macro2::fallback::Ident::new
             at /home/kkysen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.38/src/fallback.rs:652:9
   5: proc_macro2::imp::Ident::new
             at /home/kkysen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.38/src/wrapper.rs:691:50
   6: proc_macro2::Ident::new
             at /home/kkysen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.38/src/lib.rs:952:21
   7: <alloc::string::String as c2rust_ast_builder::builder::Make<proc_macro2::Ident>>::make
             at ./c2rust-ast-builder/src/builder.rs:180:9
   8: <I as c2rust_ast_builder::builder::Make<syn::path::PathSegment>>::make
             at ./c2rust-ast-builder/src/builder.rs:312:20
   9: <alloc::vec::Vec<S> as c2rust_ast_builder::builder::Make<syn::path::Path>>::make
[...]

PR #450, opened 2022-06-15
Fixed the `c2rust transpile` invalid idents panic on transpiling `long double`

kkysen:
Fixes https://github.com/immunant/c2rust/issues/449.

Changed `ident_expr()`s that were float literals to use `float_unsuffixed_lit()`, which fixes the `c2rust transpile` crash on invalid idents when transpiling `long double` to `f128`.  Now `tests/longdouble` works, so we should enable it by default and remove `--test-longdoubles`.  I think it makes more sense to put that in a separate PR, however.

The old `ident_expr` was a pre-`syn` way of inserting literals into the output.  That doesn't work with `syn` now, so we need to create them properly as literals.


fw-immunant (approved):
LGTM. These were missed in the original syn port. I did a grep for other situations with this issue and didn't find any.

PR #452, opened 2022-06-15
Disable developer build test steps on Azure

thedataking:
Azure builds are too slow. As an interim solution before we move to Github Actions, we can speed them up by disabling the developer builds and only test against the host version of clang/LLVM.


kkysen:
I'll merge this once I see how it does in CI (there's a long backlog right now, which this should hopefully prevent in the future).

kkysen:
Much faster now, about 12-15 minutes on average instead of 1+ hours.

PR #455, opened 2022-06-16
Fixed bare URL warnings in `cargo doc`

kkysen:
When running `cargo doc` and `cargo doc --document-private-items`, there were warnings about bare URLs in doc comments, which don't turn into hyperlinks in the generated documentation.  This fixes all of those by either making them named `[name](link)` links or surrounding them with `<>`.  Now `cargo doc` runs without warnings except for a clash between the `c2rust-transpile` library crate in the `c2rust-transpile` package and the `c2rust-transpile` binary crate in the `c2rust` package, which is harder to fix easily/might not be worth it.

### DIFF
--- a/c2rust-transpile/src/translator/assembly.rs
+++ b/c2rust-transpile/src/translator/assembly.rs
@@ -200,9 +200,9 @@ fn is_regname_or_int(parsed_constraint: &str) -> bool {
 /// Translate an architecture-specific assembly constraint from llvm/gcc
 /// to those accepted by the Rust asm! macro. "Simple" (arch-independent)
 /// constraints are handled in `parse_constraints`, not here.
-/// See https://gcc.gnu.org/onlinedocs/gcc/Machine-Constraints.html,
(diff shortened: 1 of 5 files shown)


c2rust-transpile/src/translator/assembly.rs
```diff
@@ -444,7 +444,7 @@ fn remove_comments(mut asm: &str) -> String {
     without_comments
 }
 
-/// Detect whether an x86[_64] gcc inline asm string uses Intel or AT&T syntax.
+/// Detect whether an x86(_64) gcc inline asm string uses Intel or AT&T syntax.
```
fw-immunant, line 447: Kind of sad that we have to change this as `[]` was intended to indicate an optional string a la EBNF. Oh well.

fw-immunant (approved):

c2rust-transpile/src/translator/assembly.rs
kkysen: `()` is more often used to indicate optionality in text, though, so I don't think it's too bad.

PR #459, opened 2022-06-17
Refactoring, cleanup, and generic testing script

kkysen:
I added a script, `./scripts/pdg.sh <test crate root> <args...>`, that instruments and runs a test crate, such as `analysis/test` or `lighttpd`.  This is a more generic version of @aneksteind's saved command.
* It only re-builds/instruments if the `c2rust-instrument` binary is newer than `metadata.bc`.
* It saves the `c2rust instrument` output to `instrument.{out,err}.log` and the `pdg` output to `pdg.log`.
* It supresses the many warnings from the test crate with `RUSTFLAGS="-Awarnings"`.
* It uses `./scripts/get-get-binary-names-from-cargo-metadata.mjs` to parse `cargo metadata` to know what default binary targets to run (because `cargo run` doesn't work with the instrumented binary).

Then I cleaned up a bunch of things:
* Removed unused imports.
* `_`-prefixed unused variables, functions, types, etc.
* Ran `rustfmt`.
Now everything builds without warnings.

And then I refactored some things in `c2rust_pdg`:
* Switched to using the RAII `c2rust_analysis_rt::Runtime` instead of `initialize` and `finalize` calls.
* Added `anyhow` for error-handling instead of `.unwrap()`ing everywhere.
* Made `EVENT_TRACE_FILE_PATH` a normal variable in `fn main` instead of a `lazy_static!`, since it's only used once.
* Gave some variables more expressive names.
* Used new inline formatting strings (e.x. `"{x:?}"`).
* Skipped some `String` allocs that were unneeded.
* Switched an `Into` `impl` to a `From` one.
* Fixed doc comments on `newtype_index!` types.
* `#[derive(Default)]`ed some structs that had similar `new` methods.
* Refactored `read_event_log` to use `iter::from_fn`, which is way cleaner.
* Added `use EventKind::*` wildcard imports right before large `match`es to reduce boilerplate.
* Added a `trait EventKindExt` `impl`ed for `EventKind` for all the functions that took an `&EventKind` that were method-like.
* Refactored `has_parent` out of `parent` (previously `get_parent_object`), since it only returned `None` or `Some(obj)`.
* Replaced uses of `usize` that were really `Pointer` (`c2rust_analysis_rt::Pointer`) with `Pointer` for clarity.
* Removed some unnecessary `.clone()`s (as in they were no-ops).
* Switched `&Vec<_>` args to `&[_]`.

### DIFF
--- a/analysis/runtime/src/mir_loc.rs
+++ b/analysis/runtime/src/mir_loc.rs
@@ -40,7 +40,7 @@ pub enum MirProjection {
     Deref,
     Field(usize),
     Index(usize),
-    Unsupported
+    Unsupported,
 }
 
 #[derive(Eq, PartialEq, Hash, Clone, Serialize, Deserialize)]
@@ -66,8 +66,15 @@ pub struct DefPathHash(pub u64, pub u64);
 
 impl fmt::Debug for DefPathHash {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let entry = "unknown".to_string();
-        write!(f, "{}", MIR_LOCS.functions.get(self).unwrap_or(&entry))
+        write!(
+            f,
+            "{}",
+            MIR_LOCS
+                .functions
+                .get(self)
+                .map(|s| s.as_str())
+                .unwrap_or("unknown")
+        )
     }
 }
 
@@ -77,9 +84,9 @@ impl From<(u64, u64)> for DefPathHash {
     }
 }
 
-impl Into<(u64, u64)> for DefPathHash {
-    fn into(self) -> (u64, u64) {
-        (self.0, self.1)
+impl From<DefPathHash> for (u64, u64) {
+    fn from(other: DefPathHash) -> Self {
+        (other.0, other.1)
     }
 }
 
--- a/pdg/src/builder.rs
+++ b/pdg/src/builder.rs
@@ -1,135 +1,148 @@
 use crate::graph::{Func, Graph, GraphId, Graphs, Node, NodeId, NodeKind};
 use bincode;
-use c2rust_analysis_rt::events::{Event, EventKind};
+use c2rust_analysis_rt::events::{Event, EventKind, Pointer};
 use c2rust_analysis_rt::mir_loc::{EventMetadata, Metadata, TransferKind};
 use c2rust_analysis_rt::{mir_loc, MirLoc};
 use log;
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_hir::def_id::DefPathHash;
-use rustc_middle::mir::Local;
 use std::collections::HashMap;
-use std::fs::File;
-use std::io::BufReader;
+use std::fs::{self, File};
+use std::io::{self, BufReader};
+use std::iter;
+use std::path::Path;
 
-pub fn read_event_log(path: String) -> Vec<Event> {
-    let file = File::open(path).unwrap();
-    let mut events = vec![];
+pub fn read_event_log(path: &Path) -> io::Result<Vec<Event>> {
+    let file = File::open(path)?;
     let mut reader = BufReader::new(file);
-    loop {
-        match bincode::deserialize_from(&mut reader) {
-            Ok(e) => events.push(e),
-            _ => break,
+    let events = iter::from_fn(|| bincode::deserialize_from(&mut reader).ok()).collect::<Vec<_>>();
+    Ok(events)
+}
+
+pub fn _read_metadata(path: &Path) -> anyhow::Result<Metadata> {
+    let bytes = fs::read(path)?;
+    let metadata = bincode::deserialize(&bytes)?;
+    Ok(metadata)
+}
+
+pub trait EventKindExt {
+    fn ptr(&self, metadata: &EventMetadata) -> Option<Pointer>;
+    fn has_parent(&self) -> bool;
+    fn parent(&self, obj: (GraphId, NodeId)) -> Option<(GraphId, NodeId)>;
+    fn to_node_kind(&self) -> Option<NodeKind>;
+}
+
+impl EventKindExt for EventKind {
+    /// return the ptr of interest for a particular event
+    fn ptr(&self, _metadata: &EventMetadata) -> Option<Pointer> {
+        use EventKind::*;
+        Some(match *self {
+            CopyPtr(lhs) => lhs,
+            Field(ptr, ..) => ptr,
+            Free { ptr } => ptr,
+            Ret(ptr) => ptr,
+            LoadAddr(ptr) => ptr,
+            StoreAddr(ptr) => ptr,
+            LoadValue(ptr) => ptr,
+            StoreValue(ptr) => ptr,
+            CopyRef => return None, // FIXME
+            ToInt(ptr) => ptr,
+            Realloc { old_ptr, .. } => old_ptr,
+            FromInt(lhs) => lhs,
+            Alloc { ptr, .. } => ptr,
+            AddrOfLocal(lhs, _) => lhs,
+            Offset(ptr, _, _) => ptr,
+            Done => return None,
+        })
+    }
+
+    fn has_parent(&self) -> bool {
+        use EventKind::*;
+        match self {
+            Realloc { new_ptr: _, .. } => false,
+            Alloc { ptr: _, .. } => false,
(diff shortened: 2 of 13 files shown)


analysis/runtime/src/mir_loc.rs
```diff
@@ -77,9 +84,9 @@ impl From<(u64, u64)> for DefPathHash {
     }
 }
 
-impl Into<(u64, u64)> for DefPathHash {
-    fn into(self) -> (u64, u64) {
-        (self.0, self.1)
+impl From<DefPathHash> for (u64, u64) {
```
aneksteind, line 87: why is this preferred?

pdg/src/builder.rs
```diff
@@ -1,136 +1,148 @@
...
+    let file = File::open(path)?;
     let mut reader = BufReader::new(file);
-    loop {
-        match bincode::deserialize_from(&mut reader) {
-            Ok(e) => events.push(e),
-            _ => break,
-        }
-    }
-    events
+    let events = iter::from_fn(|| bincode::deserialize_from(&mut reader).ok()).collect::<Vec<_>>();
```
aneksteind, line 18: nice, wasn't aware of `from_fn`

pdg/src/builder.rs
```diff
@@ -1,136 +1,148 @@
...
-        EventKind::StoreValue(..) => NodeKind::StoreValue,
-        EventKind::AddrOfLocal(_, l) => NodeKind::AddrOfLocal(Local::from(l)),
-        EventKind::ToInt(_) => NodeKind::PtrToInt,
-        EventKind::FromInt(_) => NodeKind::IntToPtr,
-        EventKind::Ret(_) => return None,
-        EventKind::Offset(_, offset, _) => NodeKind::Offset(offset),
-        EventKind::Done => return None,
-    })
+    fn parent(&self, obj: (GraphId, NodeId)) -> Option<(GraphId, NodeId)> {
+        if self.has_parent() {
```
aneksteind, line 71: I think the logic here may be reversed. Why not have this defined the other way where `has_parent` is an alias for `parent(..).is_some()`?

aneksteind: And by reversed I mean we should be returning `Some(obj)` if it has a parent

pdg/src/builder.rs
kkysen: Yeah, it's nice.  There are so many cool iterator functions and impls.

analysis/runtime/src/mir_loc.rs
kkysen: `From` is generally preferred over `Into` unless you can't due to orphan rules, because `From` implies `Into` due to the blanket `impl`.

See [the docs for `Into`](https://doc.rust-lang.org/std/convert/trait.Into.html):

One should avoid implementing [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) and implement [From](https://doc.rust-lang.org/std/convert/trait.From.html) instead. Implementing [From](https://doc.rust-lang.org/std/convert/trait.From.html) automatically provides one with an implementation of [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) thanks to the blanket implementation in the standard library.

Prefer using [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) over [From](https://doc.rust-lang.org/std/convert/trait.From.html) when specifying trait bounds on a generic function to ensure that types that only implement [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) can be used as well.

pdg/src/builder.rs
kkysen: Yeah, that is backwards, oops.

kkysen: I think it makes sense to have `parent` call `has_parent` because then `has_parent` doesn't need to take the `obj` argument.  If we had `has_parent` call `self.parent(obj).is_some()`, then we'd need that `obj`.

kkysen:
Actually I'm getting an ICE somewhere; let me fix that first.

analysis/runtime/src/mir_loc.rs
aneksteind: got it, thanks for linking all of that!

pdg/src/builder.rs
aneksteind: Granted, I'm not sure I see the appeal of breaking them up this way though?

pdg/src/builder.rs
```diff
@@ -192,15 +208,19 @@ pub fn add_node(
...
         function: Func(dest_fn),
-        block: basic_block_idx.clone().into(),
-        index: statement_idx.clone().into(),
+        block: basic_block_idx.into(),
+        index: statement_idx.into(),
         kind: node_kind,
-        source: source.and_then(|p| get_parent_object(&event.kind, p)).map(|n| n.1),
+        source: source
+            .and_then(|p| event.kind.parent(p))
+            .map(|(_, nid)| nid),
```
aneksteind, line 216: just FYI this is the place to tweak to write a test against, to ensure that all Nodes have source GraphId's (here ignored) that point to the graph whose node list they are a part of.

kkysen:
Actually, the ICE was from something I did later, not in the commits here.

aneksteind:
If the formatting and semantics of the output are maintained, and since we don't have unit tests yet, could you compare the output of this branch on an existing event and metadata log to a previous branch? Maybe in another PR we can make some of these into proper snapshot tests with https://github.com/mitsuhiko/insta

kkysen:
For now, is testing on `analysis/test` enough, or should I check a larger `event.bc` and `metadata.bc`?

[`insta`](https://github.com/mitsuhiko/insta) looks really cool.  I think I saw it once before.  We should definitely try that out.

aneksteind:
I think the event and metadata file I provided on slack would be best to test against because it provides edge cases that `analysis/test` doesn't fully cover yet.

pdg/src/builder.rs
kkysen: I fixed the reversal now.

kkysen: I wasn't sure the appeal of having the `obj` as an argument, but I left that function there since there might've been other reasons for it.  But we could drop that and just do `.has_parent().then(|| obj)`.  To me, `has_parent` seems like it has the core logic.  `obj` doesn't really need to be in it at all.

kkysen:
Ok, I'll test on those `lighttpd` `event.bc` and `metadata.bc` files.

pdg/src/builder.rs
aneksteind: Yeah, I agree with you on the `obj` piece. Okay, let's have it this way, and we can find a better way to handle `obj` later on.

kkysen:
I'm getting different pdg outputs, but I think that's because `HashMap`s are non-deterministic.  `rustc` had this problem, too.

aneksteind:
Are you referring to the order out the output objects?

kkysen:
Yeah, their order is totally different.

kkysen:
Let me see if I can just sort them.

aneksteind:
We can also use indexmap or a hasher that doesn't use a random seed

kkysen:
Either sorted or insertion order would be better for human reading, right?

kkysen:
Also, can you rebase `master` into `feat.pdg` so we can cut down the CI times?

aneksteind:
Now that I think about it, I'm not sure why the order should be different because the output just iterates over the IndexVec portion of the Graphs struct, not the hashmap. Wouldn't the output be different from run to run on the same branch if it were a HashMap issue? If you're comparing debug output of the objects themselves, just comparing the output of the program (for objects alone) should be enough.

aneksteind:
There was confusion on my part regarding output: the `latest_assignment` is getting output as well and differs between runs. If the portion of the output corresponding to just the graph objects matches, then that's enough for now to show that there have been no regressions:

```rust
for (g, graph) in pdg.graphs.iter().enumerate() {
    println!("-- Object {:?} ---", g);
    for (n, node) in graph.nodes.iter().enumerate() {
        println!("{:?}:{:?}", n, node)
    }
    println!()
}
```

pdg/src/builder.rs
kkysen: Got it.  I'll come back to check this once the tests are in place.

aneksteind (approved):

PR #460, opened 2022-06-21
Fix Clippy lints

afetisov:
This PR fixes about 50 different Clippy lints. The diff is huge, but I have tried to organize all changes into separate commits, with each commit fixing a single lint. Most of the fixes are essentially applying the automatic suggestions from Clippy, but some are a bit more complicated.

I have also disabled the following lints in the transpiler:

    clippy::borrowed_box,
    clippy::duplicate_underscore_argument,
    clippy::enum_variant_names,
    clippy::large_enum_variant,
    clippy::match_ref_pats,
    clippy::new_without_default,
    clippy::single_match,
    clippy::too_many_arguments,
    clippy::try_err,
    clippy::type_complexity,
    clippy::useless_format,

In my opinion they are unreasonably noisy (like `match_ref_pats`), too hard to fix (`type_complexity` and `too_many_arguments`) or just don't fit in the current code style and would require unreasonably far-reaching changes (`borrowed_box`, which is triggered by boxed syntax nodes, which are used everywhere). 

Some of those lints should probably be fixed and re-enabled in the future.

Still, about 30 Clippy warnings are left unfixed. I have deemed them too difficult to fix in the current state, but we can't just disable them since they warn about things which are real issues. This includes stuff like `unsafe` safety docs, which don't fit into a PR with mostly automatic fixes, identical blocks in `if` expressions, which may signify error and require careful consideration, or `try_err` lint, which asks to change `Err(E)?` into `return Err(E.into())`. I don't want to spend time on that in the current PR since I plan to significantly rework that part anyway (e.g. restructure that code, or move to `anyhow::bail!` error returns).

### DIFF
--- a/analysis/runtime/src/mir_loc.rs
+++ b/analysis/runtime/src/mir_loc.rs
@@ -22,7 +22,7 @@ lazy_static! {
             .clone()
             .expect("MIR_LOC_FILE_PATH not initialized by the instrumented code");
         let file = File::open(&path)
-            .unwrap_or_else(|_| panic!("Could not open span file: {:?}", path.to_string_lossy()));
+            .unwrap_or_else(|_| panic!("Could not open span file: {:?}", path.display()));
         bincode::deserialize_from(file).expect("Error deserializing span file")
     };
 }
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -242,8 +242,8 @@ fn display_loc(ctx: &AstContext, loc: &Option<SrcSpan>) -> Option<DisplaySrcSpan
 
 fn has_packed_attribute(attrs: Vec<Value>) -> bool {
     attrs.iter().any(|attr| {
-        let attr_name: String =
-            from_value(attr.clone()).expect("Record attributes should be strings");
+        let attr_name =
+            from_value::<String>(attr.clone()).expect("Record attributes should be strings");
         attr_name == "packed"
     })
 }
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -385,22 +385,22 @@ impl ConversionContext {
     }
 
     /// Add a `CType`node into the `TypedAstContext`
-    fn add_type(&mut self, id: ImporterId, typ: CType) -> () {
+    fn add_type(&mut self, id: ImporterId, typ: CType) {
         self.typed_context.c_types.insert(CTypeId(id), typ);
     }
 
     /// Add a `CStmt` node into the `TypedAstContext`
-    fn add_stmt(&mut self, id: ImporterId, stmt: CStmt) -> () {
+    fn add_stmt(&mut self, id: ImporterId, stmt: CStmt) {
         self.typed_context.c_stmts.insert(CStmtId(id), stmt);
     }
 
     /// Add a `CExpr` node into the `TypedAstContext`
-    fn add_expr(&mut self, id: ImporterId, expr: CExpr) -> () {
+    fn add_expr(&mut self, id: ImporterId, expr: CExpr) {
         self.typed_context.c_exprs.insert(CExprId(id), expr);
     }
 
     /// Add a `CDecl` node into the `TypedAstContext`
-    fn add_decl(&mut self, id: ImporterId, decl: CDecl) -> () {
+    fn add_decl(&mut self, id: ImporterId, decl: CDecl) {
         self.typed_context.c_decls.insert(CDeclId(id), decl);
     }
 
@@ -413,7 +413,7 @@ impl ConversionContext {
         new_id: ImporterId,
         node: &AstNode,
         expr: CExprKind,
-    ) -> () {
+    ) {
         if expected_ty & node_types::STMT != 0 {
             // This is going to be an extra node not present in the Clang AST
             let new_expr_id = self.id_mapper.fresh_id();
@@ -437,7 +437,7 @@ impl ConversionContext {
     /// into the `ConversionContext` on creation.
     ///
     /// This populates the `typed_context` of the `ConversionContext` it is called on.
-    fn convert(&mut self, untyped_context: &AstContext) -> () {
+    fn convert(&mut self, untyped_context: &AstContext) {
         for raw_comment in &untyped_context.comments {
             let comment = Located {
                 loc: Some(raw_comment.loc.into()),
@@ -497,7 +497,7 @@ impl ConversionContext {
         node_id: ClangId,      // Clang ID of node to visit
         new_id: ImporterId,    // New ID of node to visit
         expected_ty: NodeType, // Expected type of node to visit
-    ) -> () {
+    ) {
         use self::node_types::*;
 
         if expected_ty & TYPE != 0 {
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -245,17 +245,11 @@ fn display_loc(ctx: &AstContext, loc: &Option<SrcSpan>) -> Option<DisplaySrcSpan
 }
 
 fn has_packed_attribute(attrs: Vec<Value>) -> bool {
-    for attr in attrs {
-        match from_value::<String>(attr.clone())
-            .expect("Record attributes should be strings")
-            .as_str()
-        {
-            "packed" => return true,
-            _ => {}
-        }
-    }
-
-    return false;
+    attrs.iter().any(|attr| {
+        let attr_name: String =
+            from_value(attr.clone()).expect("Record attributes should be strings");
+        attr_name == "packed"
+    })
 }
 
 impl ConversionContext {
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -57,7 +57,7 @@ type ImporterId = u64;
 /// We need to re-ID nodes since the mapping from Clang's AST to ours is not one-to-one. Sometimes
 /// we need to add nodes (such as 'Semi' nodes to make the lifting of expressions into statements
 /// explicit), sometimes we need to collapse (such as inlining 'FieldDecl' into the 'StructDecl').
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct IdMapper {
     new_id_source: ImporterId,
     old_to_new: HashMap<ClangId, ImporterId>,
@@ -65,12 +65,8 @@ pub struct IdMapper {
 }
 
 impl IdMapper {
-    pub fn new() -> IdMapper {
-        IdMapper {
-            new_id_source: 0,
-            old_to_new: HashMap::new(),
-            new_to_old: HashMap::new(),
-        }
+    pub fn new() -> Self {
+        Self::default()
     }
 
     /// Create a fresh NEW_ID not corresponding to a CLANG_ID
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -241,11 +241,12 @@ fn display_loc(ctx: &AstContext, loc: &Option<SrcSpan>) -> Option<DisplaySrcSpan
 }
 
 fn has_packed_attribute(attrs: Vec<Value>) -> bool {
-    attrs.iter().any(|attr| {
-        let attr_name =
-            from_value::<String>(attr.clone()).expect("Record attributes should be strings");
-        attr_name == "packed"
-    })
+    attrs
+        .into_iter()
+        .map(|attr| {
+            from_value::<String>(attr).expect("Record attributes should be strings")
+        })
+        .any(|attr_name| attr_name == "packed")
 }
 
 impl ConversionContext {
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -243,9 +243,7 @@ fn display_loc(ctx: &AstContext, loc: &Option<SrcSpan>) -> Option<DisplaySrcSpan
 fn has_packed_attribute(attrs: Vec<Value>) -> bool {
     attrs
         .into_iter()
-        .map(|attr| {
-            from_value::<String>(attr).expect("Record attributes should be strings")
-        })
+        .map(|attr| from_value::<String>(attr).expect("Record attributes should be strings"))
         .any(|attr_name| attr_name == "packed")
 }
 
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -988,24 +988,19 @@ pub enum CDeclKind {
 
 impl CDeclKind {
     pub fn get_name(&self) -> Option<&String> {
-        match self {
-            &CDeclKind::Function { name: ref i, .. } => Some(i),
-            &CDeclKind::Variable { ident: ref i, .. } => Some(i),
-            &CDeclKind::Typedef { name: ref i, .. } => Some(i),
-            &CDeclKind::EnumConstant { name: ref i, .. } => Some(i),
-            &CDeclKind::Enum {
-                name: Some(ref i), ..
-            } => Some(i),
-            &CDeclKind::Struct {
-                name: Some(ref i), ..
-            } => Some(i),
-            &CDeclKind::Union {
-                name: Some(ref i), ..
-            } => Some(i),
-            &CDeclKind::Field { name: ref i, .. } => Some(i),
-            &CDeclKind::MacroObject { ref name, .. } => Some(name),
-            _ => None,
-        }
+        use CDeclKind::*;
+        Some(match self {
+            Function { name: i, .. } => i,
+            Variable { ident: i, .. } => i,
+            Typedef { name: i, .. } => i,
+            EnumConstant { name: i, .. } => i,
+            Enum { name: Some(i), .. } => i,
+            Struct { name: Some(i), .. } => i,
+            Union { name: Some(i), .. } => i,
+            Field { name: i, .. } => i,
+            MacroObject { name, .. } => name,
+            _ => return None,
+        })
     }
 }
 
@@ -1243,12 +1238,57 @@ pub enum UnOp {
     Coawait,       // [C++ Coroutines] co_await x
 }
 
+impl UnOp {
+    pub fn as_str(&self) -> &'static str {
+        use UnOp::*;
+        match self {
+            AddressOf => "&",
+            Deref => "*",
+            Plus => "+",
+            PreIncrement => "++",
+            PostIncrement => "++",
+            Negate => "-",
+            PreDecrement => "--",
+            PostDecrement => "--",
+            Complement => "~",
+            Not => "!",
+            Real => "__real",
+            Imag => "__imag",
+            Extension => "__extension__",
+            Coawait => "co_await",
+        }
+    }
+}
+
+impl Display for UnOp {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
 /// Represents a unary type operator in C
 #[derive(Debug, Clone, Copy)]
 pub enum UnTypeOp {
     SizeOf,
     AlignOf,
     PreferredAlignOf,
+}
+
+impl UnTypeOp {
+    pub fn as_str(&self) -> &'static str {
+        use UnTypeOp::*;
+        match self {
+            SizeOf => "sizeof",
+            AlignOf => "alignof",
+            PreferredAlignOf => "__alignof",
+        }
+    }
+}
+
+impl Display for UnTypeOp {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
 }
 
 impl UnOp {
@@ -1293,6 +1333,52 @@ pub enum BinOp {
 
     Assign, // =
     Comma,  // ,
+}
+
+impl BinOp {
+    pub fn as_str(&self) -> &'static str {
+        use BinOp::*;
+        match self {
+            Multiply => "*",
+            Divide => "/",
+            Modulus => "%",
+            Add => "+",
+            Subtract => "-",
+            ShiftLeft => "<<",
+            ShiftRight => ">>",
+            Less => "<",
+            Greater => ">",
+            LessEqual => "<=",
+            GreaterEqual => ">=",
+            EqualEqual => "==",
+            NotEqual => "!=",
+            BitAnd => "&",
+            BitXor => "^",
+            BitOr => "|",
+            And => "&&",
+            Or => "||",
+
+            AssignAdd => "+=",
+            AssignSubtract => "-=",
+            AssignMultiply => "*=",
+            AssignDivide => "/=",
+            AssignModulus => "%=",
+            AssignBitXor => "^=",
+            AssignShiftLeft => "<<=",
+            AssignShiftRight => ">>=",
+            AssignBitOr => "|=",
+            AssignBitAnd => "&=",
+
+            Assign => "=",
+            Comma => ", ",
+        }
+    }
+}
+
+impl Display for BinOp {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
 }
 
 impl BinOp {
@@ -1589,6 +1675,41 @@ pub enum CTypeKind {
 
     Half,
     BFloat16,
+}
+
+impl CTypeKind {
+    pub fn as_str(&self) -> &'static str {
+        use CTypeKind::*;
+        match self {
+            Void => "void",
+            Bool => "_Bool",
+            Char => "char",
+            SChar => "signed char",
+            Short => "signed short",
+            Int => "int",
+            Long => "long",
+            LongLong => "long long",
+            UChar => "unsigned char",
+            UShort => "unsigned short",
+            UInt => "unsigned int",
+            ULong => "unsigned long",
+            ULongLong => "unsigned long long",
+            Float => "float",
+            Double => "double",
+            LongDouble => "long double",
+            Int128 => "__int128",
+            UInt128 => "unsigned __int128",
+            Half => "half",
+            BFloat16 => "bfloat16",
+            _ => unimplemented!("Printer::print_type({:?})", self),
+        }
+    }
+}
+
+impl Display for CTypeKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
 }
 
 #[derive(Copy, Clone, Debug)]
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -598,23 +598,20 @@ impl TypedAstContext {
                 use SomeId::*;
                 match some_id {
                     Type(type_id) => {
-                        match self.c_types[&type_id].kind {
+                        if let CTypeKind::Elaborated(decl_type_id) = self.c_types[&type_id].kind {
                             // This is a reference to a previously declared type.  If we look
                             // through it we should(?) get something that looks like a declaration,
                             // which we can mark as wanted.
-                            CTypeKind::Elaborated(decl_type_id) => {
-                                let decl_id = self.c_types[&decl_type_id]
-                                    .kind
-                                    .as_decl_or_typedef()
-                                    .expect("target of CTypeKind::Elaborated isn't a decl?");
-                                if wanted.insert(decl_id) {
-                                    to_walk.push(decl_id);
-                                }
+                            let decl_id = self.c_types[&decl_type_id]
+                                .kind
+                                .as_decl_or_typedef()
+                                .expect("target of CTypeKind::Elaborated isn't a decl?");
+                            if wanted.insert(decl_id) {
+                                to_walk.push(decl_id);
                             }
-
+                        } else {
                             // For everything else (including `Struct` etc.), DFNodes will walk the
                             // corresponding declaration.
-                            _ => {}
                         }
                     }
 
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -335,21 +335,18 @@ impl TypedAstContext {
         let resolved_ctype = self.resolve_type(typ);
         match resolved_ctype.kind {
             CTypeKind::Struct(record_id) => {
-                let r#struct = &self[record_id];
                 if let CDeclKind::Struct {
                     name: Some(ref nam),
                     ..
-                } = r#struct.kind
+                } = &self[record_id].kind
                 {
-                    return nam == "__va_list_tag" || nam == "__va_list";
+                    nam == "__va_list_tag" || nam == "__va_list"
                 } else {
                     false
                 }
             }
             // va_list is a 1 element array; return true iff element type is struct __va_list_tag
-            CTypeKind::ConstantArray(typ, 1) => {
-                return self.is_va_list(typ);
-            }
+            CTypeKind::ConstantArray(typ, 1) => self.is_va_list(typ),
             _ => false,
         }
     }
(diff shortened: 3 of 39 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -152,9 +152,8 @@ impl TypedAstContext {
         }
 
         let mut include_map = vec![];
-        for fileid in 0..files.len() {
+        for (fileid, mut cur) in files.iter().enumerate() {
```
fw-immunant, line 155: It's somewhat wild to me to bind an iteration variable mutably, but sure.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -376,17 +376,13 @@ fn vec_expr(val: Box<Expr>, count: Box<Expr>) -> Box<Expr> {
...
-            }
-        }
+    match stmts.as_slice() {
+        [Stmt::Expr(Expr::Block(ExprBlock {
+            block: b,
+            label: None,
+            ..
+        }))] => return Box::new(b.clone()),
+        _ => {}
     }
```
fw-immunant, line 386: This seems better written as an `if let` with a slice pattern, no? I would expect the `single_match` lint to complain about this.

fw-immunant (approved):
These all look good, with a couple tiny nits mentioned inline. Thanks for making this easy to review.

c2rust-transpile/src/c_ast/mod.rs
fw-immunant: It's just like any other `mut` variable, but unlike the vast majority of iteration variables, it gets rebound in the body. The contrast is with a C-style `for` loop where altering the iteration variable in the body would change the iteration structure. It's not wrong, but I was surprised to see this idiom.

c2rust-transpile/src/translator/mod.rs
afetisov: It does, that's why it's silenced 😕 But imho using if-let with huge patterns is very clunky. It also somewhat obscures the scrutinee, and causes an inversion of flow (pattern before the match value) which I find confusing for big patterns.

However, I agree that it's more of a my personal preference in this specific case than some overarching policy. I have changed some matches into if-lets with big patterns in this PR. I think this specific form looks better, but if you would insist on consistency I can change it into an if-let.

I'm not a fan of huge patterns, but I don't have a better solution.

kkysen:
Hi @afetisov, thanks for all the help!  Can I push a few changes to your fork if I figure out how to do fix something better, like that `match` statement?

afetisov:
@kkysen Sure.

c2rust-transpile/src/cfg/structures.rs
```diff
@@ -640,26 +630,24 @@ impl StructureState {
...
+                                }),
+                                _token,
+                            )] => {
+                                let e = mk().while_expr(
+                                    not(cond),
+                                    mk().span(body_span)
+                                        .block(body.iter().skip(1).cloned().collect()),
+                                    lbl.map(|l| l.pretty_print()),
+                                );
+                                return (vec![mk().span(span).expr_stmt(e)], ast.span);
```
kkysen, line 648: `clippy` complains about the single-branch `match` here, right?  If the warning is being suppressed, move the `#[allow(...)]` here.

c2rust-transpile/src/translator/structs.rs
```diff
@@ -46,19 +46,17 @@ fn contains_block(expr_kind: &Expr) -> bool {
...
-        Expr::Unary(ExprUnary { expr, .. }) => contains_block(&expr),
-        Expr::MethodCall(ExprMethodCall { args, .. }) => {
-            args.iter().map(|e| contains_block(&e)).any(|b| b)
-        }
-        Expr::Cast(ExprCast { expr, .. }) => contains_block(&expr),
+        Expr::Unary(ExprUnary { expr, .. }) => contains_block(expr),
+        Expr::MethodCall(ExprMethodCall { args, .. }) => args.iter().any(contains_block),
+        Expr::Cast(ExprCast { expr, .. }) => contains_block(expr),
         _ => false,
     }
```
kkysen, line 61: Can the `contains_block` be refactored out of the `match` by returning an `Iterator` and then calling `.iter().any(contains_block)` on it outside of the `match`?

c2rust-transpile/src/translator/structs.rs
```diff
@@ -125,8 +123,7 @@ impl<'a> Translation<'a> {
                             mk().angle_bracketed_args(vec![mk().lifetime("a")]),
                         ),
                     ];
-                    let ty = mk().path_ty(mk().abs_path(path));
-                    ty
+                    mk().path_ty(mk().abs_path(path))
```
kkysen, line 126: `clippy` warns on this, right?  The `ty` variable name also serves as documentation.

c2rust-transpile/src/translator/mod.rs
kkysen: Wait, that's not equivalent.  I'll figure it out.

c2rust-transpile/src/lib.rs
```diff
@@ -1,3 +1,17 @@
...
+    clippy::enum_variant_names,
+    clippy::large_enum_variant,
+    clippy::match_ref_pats,
+    clippy::new_without_default,
+    clippy::single_match,
+    clippy::too_many_arguments,
+    clippy::try_err,
+    clippy::type_complexity,
+    clippy::useless_format
+)]
```
kkysen, line 13: Same as the other blanket `#![allow(...)]`: either `#[allow(...)]` them on a specific item with a comment explaining why, or leave them as warnings to be fixed by less-trivial changes.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -488,7 +484,7 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&path
                 .parent()
                 .unwrap()
                 .iter()
-                .map(|os| PathBuf::from(os))
+                .map(PathBuf::from)
```
kkysen, line 487: An `|s| s.into()` would work, too.  Generally `.into()` is preferred over `::from()`, but I'm not sure when you can avoid the closure.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1052,7 +1047,7 @@ fn arrange_header(t: &Translation, is_binary: bool) -> (Vec<syn::Attribute>, Vec
     } else {
         let pragmas = t.get_pragmas();
         for (key, mut values) in pragmas {
-            values.sort();
+            values.sort_unstable();
```
kkysen, line 1050: How do we know an unstable sort is okay?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4058,16 +4042,17 @@ impl<'c> Translation<'c> {
...
-                }) = *expr
-                {
-                    if blbl.ident == mk().label(lbl.pretty_print()).name.ident {
-                        return Some(ret_val.clone());
-                    }
+                }),
+                _,
+            ) = stmt
+            {
+                if blbl.ident == mk().label(lbl.pretty_print()).name.ident {
```
kkysen, line 4054: `blbl` is pretty cryptic.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1637,63 +1621,52 @@ pub enum Attribute {
...
-        match *self {
-            CTypeKind::Float => true,
-            CTypeKind::Double => true,
-            CTypeKind::LongDouble => true,
-            _ => false,
-        }
+        matches!(
+            self,
+            CTypeKind::Float | CTypeKind::Double | CTypeKind::LongDouble
+        )
```
kkysen, line 1669: `use CTypeKind::*`

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1637,63 +1621,52 @@ pub enum Attribute {
...
+        matches!(
+            self,
+            CTypeKind::Char // true on the platforms we handle
+                | CTypeKind::SChar
+                | CTypeKind::Int
+                | CTypeKind::Short
+                | CTypeKind::Long
+                | CTypeKind::LongLong
+                | CTypeKind::Int128
+        )
```
kkysen, line 1662: `use CTypeKind::*`

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1637,63 +1621,52 @@ pub enum Attribute {
...
+        matches!(
+            self,
+            CTypeKind::Bool
+                | CTypeKind::UChar
+                | CTypeKind::UInt
+                | CTypeKind::UShort
+                | CTypeKind::ULong
+                | CTypeKind::ULongLong
+                | CTypeKind::UInt128
+        )
```
kkysen, line 1649: `use CTypeKind::*`

c2rust-transpile/src/translator/named_references.rs
```diff
@@ -56,16 +56,16 @@ impl<'c> Translation<'c> {
...
-                    | Expr::Index(..) => true,
-                    _ => false,
-                }
+                        | Expr::Unary(ExprUnary {
+                            op: syn::UnOp::Deref(_),
+                            ..
+                        })
+                        | Expr::Field(..)
+                        | Expr::Index(..)
+                )
```
kkysen, line 68: `use Expr::*`

c2rust-transpile/src/translator/named_references.rs
```diff
@@ -56,16 +56,16 @@ impl<'c> Translation<'c> {
         reference.and_then(|reference| {
             /// Check if something is a valid Rust lvalue. Inspired by `librustc::ty::expr_is_lval`.
             fn is_lvalue(e: &Box<Expr>) -> bool {
```
kkysen, line 58: Can just take an `&Expr` I think.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -894,29 +894,27 @@ impl<'c> Translation<'c> {
...
                     } else {
-                        let k = if up {
-                            BinOp::Add(Default::default())
-                        } else {
-                            BinOp::Sub(Default::default())
-                        };
-                        mk().binary_expr(k, read.clone(), one)
-                    }
+                        BinOp::Sub(Default::default())
+                    };
```
kkysen, line 916: Could be
```rust
let k = (if up { BinOp::Add } else { BinOp::Sub })(Default::default());
```

c2rust-transpile/src/diagnostics.rs
```diff
@@ -42,7 +42,7 @@ pub fn init(mut enabled_warnings: HashSet<Diagnostic>, log_level: log::LevelFilt
                 Level::Trace => "trace",
             };
             let target = record.target();
-            let warn_flag = if let Ok(_) = Diagnostic::from_str(target) {
+            let warn_flag = if Diagnostic::from_str(target).is_ok() {
                 format!(" [-W{}]", target)
             } else {
                 String::new()
```
kkysen, line 48: Could be
```rust
let warn_flag = Diagnostic::from_str(target).map(|_| format!(" [-W{}]", target)).unwrap_or_default();
```

c2rust-transpile/src/cfg/inc_cleanup.rs
```diff
@@ -136,32 +130,28 @@ impl IncCleanup {
...
+                block: blk,
+                label: None,
+                ..
+            }) if blk.stmts.is_empty() => {
+                return Stmt::Expr(Expr::If(ExprIf {
+                    cond: cond.clone(),
+                    then_branch: body.clone(),
+                    else_branch: None,
+                    attrs: vec![],
+                    if_token: Default::default(),
```
kkysen, line 151: Maybe just do `Default::default()` for all of these fields to be consistent.

c2rust-transpile/src/cfg/inc_cleanup.rs
```diff
@@ -136,32 +130,28 @@ impl IncCleanup {
...
+    if let Stmt::Expr(Expr::If(ExprIf {
+        cond,
+        then_branch: body,
+        else_branch: Some((_token, block)),
+        ..
+    })) = &stmt
+    {
+        match &**block {
+            Expr::Block(ExprBlock {
+                block: blk,
```
kkysen, line 142: No need to rename `block`.

c2rust-transpile/src/cfg/inc_cleanup.rs
```diff
@@ -136,32 +130,28 @@ impl IncCleanup {
...
-                            cond: cond.clone(),
-                            then_branch: body.clone(),
-                            else_branch: None,
-                        }));
-                    }
-                }
+    if let Stmt::Expr(Expr::If(ExprIf {
+        cond,
+        then_branch: body,
+        else_branch: Some((_token, block)),
```
kkysen, line 136: Keep `body` as `then_branch` and `block` as `else_branch`.

kkysen (changes requested):
I made a bunch of small suggestions for individual changes, some of which I went ahead and fixed, as well as some other `clippy` warnings.  In general, I think we should avoid any blanket `#![allow(...)]`s.  If we need them, we should apply them to the smallest scope possible with a comment explaining why.  And if it's something we can fix later, just leave the warning as is.

analysis/runtime/src/mir_loc.rs
```diff
@@ -21,8 +21,8 @@ lazy_static! {
             .expect("MIR_LOC_FILE_PATH was locked")
             .clone()
             .expect("MIR_LOC_FILE_PATH not initialized by the instrumented code");
-        let file =
-            File::open(&path).expect(&format!("Could not open span file: {:?}", path.to_str()));
+        let file = File::open(&path)
+            .unwrap_or_else(|_| panic!("Could not open span file: {:?}", path.to_string_lossy()));
```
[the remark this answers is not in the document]
kkysen: Fixed in https://github.com/immunant/c2rust/pull/460/commits/4acae2bf5de31e71afa300e2913630a51fea98ab.

c2rust-transpile/src/cfg/inc_cleanup.rs
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/4b4257d8b69048d9012e44db027d3104ab2ca2c3.

c2rust-transpile/src/cfg/inc_cleanup.rs
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/4b4257d8b69048d9012e44db027d3104ab2ca2c3.

c2rust-transpile/src/translator/operators.rs
kkysen: Actually, `BinOp::Add` and `BinOp::Sub` have different types since `Token![+]` and `Token![-]` are different types, so this can't be done.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1318,20 +1302,20 @@ impl BinOp {
...
+                | BinOp::AssignSubtract
+                | BinOp::AssignMultiply
+                | BinOp::AssignDivide
+                | BinOp::AssignModulus
+                | BinOp::AssignBitXor
+                | BinOp::AssignShiftLeft
+                | BinOp::AssignShiftRight
+                | BinOp::AssignBitOr
+                | BinOp::AssignBitAnd
+                | BinOp::Assign
```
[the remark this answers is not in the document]
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/4d37cfe651bd8688575b64916d712e0801fdbf5b.

c2rust-transpile/src/c_ast/mod.rs
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/01f375a473594b3d3efcd3db852d075c44b3d686.

c2rust-transpile/src/c_ast/mod.rs
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/01f375a473594b3d3efcd3db852d075c44b3d686.

c2rust-transpile/src/diagnostics.rs
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/3ac46569f1955c3e06b6c770ebc1c04568dafffa.

c2rust-transpile/src/translator/named_references.rs
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/efa9f2a74169cc649e33097d9624d28604523934.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -686,26 +674,26 @@ impl TypedAstContext {
...
+                is_packed: true,
+                ..
+            } | CDeclKind::Struct {
                 max_field_alignment: Some(_),
                 ..
-            } => true,
-            _ => false,
-        }
+            }
+        )
```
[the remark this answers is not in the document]
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/99be3a65c87c774c8d27872a9ed295f5605909d9.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -397,13 +389,12 @@ impl TypedAstContext {
...
+            field_ty.kind,
             CTypeKind::IncompleteArray(_)
-            | CTypeKind::ConstantArray(_, 0)
-            | CTypeKind::ConstantArray(_, 1) => true,
-
-            _ => false,
-        }
+                | CTypeKind::ConstantArray(_, 0)
+                | CTypeKind::ConstantArray(_, 1)
+        )
```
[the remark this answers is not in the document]
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/99be3a65c87c774c8d27872a9ed295f5605909d9.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -291,15 +290,15 @@ impl TypedAstContext {
...
-            },
+            Some(decl_id) => matches!(
+                self[decl_id].kind,
+                CDeclKind::Struct { fields: None, .. }
+                    | CDeclKind::Union { fields: None, .. }
+                    | CDeclKind::Enum {
+                        integral_type: None,
+                        ..
+                    }
+            ),
```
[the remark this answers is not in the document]
kkysen: Fixed by https://github.com/immunant/c2rust/pull/460/commits/99be3a65c87c774c8d27872a9ed295f5605909d9.

c2rust-transpile/src/translator/mod.rs
afetisov: I don't think such significant restructuring of a function fits in a PR with mostly automatically applicable changes. If you want to change it, it would be better to make it a separate PR.

There are, in fact, a lot of such functions which could benefit from some cleanup.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -240,22 +240,16 @@ pub struct ConversionContext {
...
-            _ => {}
-        }
-    }
-
-    return false;
+    attrs.iter().any(|attr| {
+        let attr_name: String =
+            from_value(attr.clone()).expect("Record attributes should be strings");
+        attr_name == "packed"
+    })
```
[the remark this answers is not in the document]
afetisov: Maybe? I don't know. I have focused on fixing the lints, performing equivalent code transformations, and this change requires thinking long and hard about how all code fits together. Since it was good enough before the change, it should be good enough after the change as well.

Overall, the codebase is too panic-happy, and would benefit from a cleanup of error handling, but that's a separate issue.

c2rust-transpile/src/c_ast/mod.rs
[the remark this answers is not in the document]
afetisov: Personally I prefer `matches!` for code which can be expressed using it. One benefit is it is more concise. Even if it is formatted like in this PR, there are still less tokens obscuring the expression. The bigger benefit is that is a significantly weaker construct that a match: it always returns a `bool`, always does just a simple match and cannot include blocks as branches. I prefer to apply strength reduction when it doesn't compromise other code qualities.

c2rust-transpile/src/cfg/inc_cleanup.rs
```diff
@@ -56,14 +56,8 @@ impl IncCleanup {
...
-                            match *case.body {
-                                Expr::Block(ExprBlock {
-                                    block: ref mut blk, ..
-                                }) => {
-                                    removed_tail_expr =
-                                        removed_tail_expr || self.remove_tail_expr(&mut blk.stmts)
-                                }
-                                _ => (),
+                            if let Expr::Block(ExprBlock { block, .. }) = case.body.as_mut() {
+                                removed_tail_expr |= self.remove_tail_expr(&mut block.stmts);
```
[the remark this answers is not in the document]
afetisov: Indeed. I think this part will be more readable as an explicit `if`-statement.

c2rust-transpile/src/translator/mod.rs
afetisov: I prefer using `From` where it doesn't result in convoluted code. It makes the type clear, and it is easy to use as a free function. But eventually it depends on the context.

c2rust-transpile/src/translator/mod.rs
afetisov: Clippy told me ) More seriously, the difference between stable and unstable sorting is that unstable sort doesn't need to preserve the order of equal elements. But for primitives, including the `&str` type used here, the elements compare equal under `PartialEq::eq` only if they are identical. For this reason there is literally no difference between stable and unstable sort, but unstable sort is faster.

The [lint description](https://rust-lang.github.io/rust-clippy/master/#stable_sort_primitive) says that sometimes stable sort of primitives is faster than unstable sort, but in the absence of benchmark data I would always go with the lint.

c2rust-transpile/src/diagnostics.rs
afetisov: I will disagree here. Using the `Result` combinators in a context where you don't actually use any of the payloads and just treat the `Result` as a boolean is confusing. The first version makes it clear that the contents of the parse are irrelevant, and isn't significantly worse looking.

c2rust-transpile/src/c_ast/conversion.rs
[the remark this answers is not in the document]
fw-immunant: Panicking if this is not a `String` is a consistency check between the CBOR generator and consumer, and should stay in.

c2rust-transpile/src/translator/mod.rs
kkysen: Hmm, I did clean up a few other functions already.  You're probably right that they should go in separate PRs, but that also makes it more work and takes longer to separate everything out.

For this one and any similar ones, how about we revert them to whatever is fully `clippy`-approved and then we can refactor them later in a separate PR to make them cleaner.

c2rust-transpile/src/c_ast/conversion.rs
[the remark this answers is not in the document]
kkysen: Ok, we'll leave it.  It would make sense to fix it in a separate PR if we did want to change it, but I wanted to mark it here before it gets lost.

c2rust-transpile/src/c_ast/mod.rs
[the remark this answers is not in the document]
kkysen: I generally don't like `matches!` for large patterns because it puts a pattern into non-first parameter position, which looks very un-idiomatic compared to other Rust code.

Still, I think it's a good idea to use `matches!` in places like these because of the strength reduction.  But if the code could be refactored to make it simpler elsewhere as well and in doing so eliminate a large `matches!`, I tend to prefer that.

c2rust-transpile/src/cfg/inc_cleanup.rs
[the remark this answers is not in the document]
kkysen: I don't think this

```rust
removed_tail_expr = removed_tail_expr || self.remove_tail_expr(&mut blk.stmts)
```

is less readable than the equivalent `if`:

```rust
if !removed_tail_expr {
    removed_tail_expr  = self.remove_tail_expr(&mut blk.stmts)
} 
```

c2rust-transpile/src/translator/mod.rs
kkysen: I think it's fine here.

c2rust-transpile/src/translator/named_references.rs
kkysen: I think we can save this for a more-involved `clippy::borrowed_box` fix that affects a lot of functions at once.

kkysen:
I think once we remove all the blanket `#![allow(...)]`s and a few other things, we can merge this.

For the few other things:
* [ ] Change the single-arm `match`es to `matches!` if `clippy` says so.
* [ ] Decide on `sort` vs. `sort_unstable`.
* [ ] Undo the `|=` change to use `_ = _ || ...`.
* [ ] Use `from_value::<String>(...)` instead of `let _: String = from_value(...)` for consistency.

afetisov:
Ok, I'll implement the minor fixes when I have the time. But I _really_ don't want to remove the blanket suppressions. I have explained my reasoning above, and I strongly believe that it would make the code harder rather than easier to maintain. Specific lints which will soon be fixed may be dropped, of course.

Specifically, `duplicate_underscore_argument`, `new_without_default`, `large_enum_variant`, `match_ref_pats`, `single_match` and `try_err` are more of a case of me being lazy. We can fix them easily and we probably should, I'm just afraid of the churn.

`too_many_arguments`, `type_complexity` and `useless_format` are just Clippy being annoying. It's hard to do anything about them in a way which doesn't just obscure the issue (like introducing ad hoc structs). `useless_format` is particularly annoying, `format!("")` instead of `String::new()` isn't in any way worse than `vec![]` instead of `Vec::new()`. That lint should really be `pedantic` rather than `complexity`.

kkysen:
I really disagree about the `clippy` warnings.  They've been added for a reason, and I think they're all legitimate issues.  Moreover, I don't think we should ever globally disable them.  That prevents new code from seeing any of these warnings, which is bad.  If we really do want to apply them, they should be applied individually to each instance that causes the warning.

Specifically,
* [`clippy::borrowed_box`](https://rust-lang.github.io/rust-clippy/master/#borrowed_box)
  This is similar to the `&Vec<T>` to `&[T]` changes, and we should fix these as well to avoid extra indirections and possibly some unnecessary boxes.  We don't necessarily need to fix them in this PR, as the fix touches a lot of functions, but we shouldn't just suppress this warning.

* [`clippy::enum_variant_names`](https://rust-lang.github.io/rust-clippy/master/#enum_variant_names)
  I don't see why this isn't a good suggestion.  We're repeating the `enum` name.  And anyways, this only happens to a single `enum`, so if we really want to keep the variant names as is, why wouldn't we just apply a local `#[allow(clippy::enum_variant_names)]` directly to that type?

* [`clippy::large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/#large_enum_variant)
  This is a great warning, one that would be quite hard to catch just by looking at the code.  We should inspect each `enum` that does this, and if it's passed around by value often, maybe we should `Box` the large variants to keep the `enum` small.  Or if it's not worth it, then we should signal that we determined that ourselves by applying a local `#[allow(clippy::large_enum_variant)]` that `enum` directly with a comment explaining our reasoning.

* [`clippy::match_ref_pats`](https://rust-lang.github.io/rust-clippy/master/#match_ref_pats)
  I don't see why we wouldn't want to apply these suggestions, probably in this PR as they seem simple and automatic.  Adding a `*` dereference to the `match` expression instead of adding `&` to all `match` arm patterns seems very much like the right decision.

* [`clippy::single_match`](https://rust-lang.github.io/rust-clippy/master/#single_match)
  We should either fix these warnings with the suggested `if let` or leave them as is and fix them by refactoring to avoid the single-arm `match`.

* [`clippy::try_err`](https://rust-lang.github.io/rust-clippy/master/#try_err)
  This one is quite annoying I admit.  `?` is a lot cleaner here, though we really could use a `yeet` actually.  But using `Err(...)?` instead of `return Err(....into())` (and I'm not sure if we always need the `.into()`) unconditionally returns, and this can help in certain cases, so I think this is still preferable.  Furthermore, as it's a `clippy` lint, a lot of other Rust code will be following this as well, and it'd be better to stick to the most idiomatic Rust.

* [`clippy::type_complexity`](https://rust-lang.github.io/rust-clippy/master/#type_complexity)
  Some of these types are indeed overly complex and it would greatly aid readability if they were refactored into some type aliases, which give names to smaller parts of the type.  I've already done this for a few of the `clippy::type_complexity` warnings.

* [`clippy::duplicate_underscore_argument`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  By the way, this `#![allow(clippy::duplicate_underscore_argument)]` wasn't being put in the right place, so the warning was still popping up.  It's a legitimate warning, and I'm not sure why there is both a `ptr: usize` and a `_ptr: ()` argument.  We should figure this out, not suppress the warning.

* [`clippy::new_without_default`](https://rust-lang.github.io/rust-clippy/master/#new_without_default)
  For many of these `struct`s that have a `Self::new()` method that is identical to what `#[derive(Default)]` would generate, we should instead just use a `#[derive(Default)]` and have `Self::new()` call `Self::default()`.  It's a lot simpler and less error-prone.

* [`clippy::too_many_arguments`](https://rust-lang.github.io/rust-clippy/master/#too_many_arguments)
  There are some functions with way too many arguments.  14 arguments seems a bit overboard.  These should be refactored into an arg `struct`, which also lets you have named arguments in a way.  A small fix like this might not be a priority to fix, but that doesn't mean the warning isn't valid.

* [`clippy::useless_format`](https://rust-lang.github.io/rust-clippy/master/#useless_format)
  This is pretty easy to fix and I just pushed a fix already.  For the `format!("")` case, that can just be turned into `"".into()` or `String::new()`.  I don't see why that's such a big problem.  For cases like `format!("{}", s)`, that can just be turned into `s.to_string()`, though I understand why you might not want to do that in a large `match` where all the other arms use `format!`.  For that case (there was 1), I just refactored the code a bit, made it simpler anyways, and got rid of the useless format.

c2rust-transpile/src/translator/mod.rs
kkysen: I think we'll just leave it as `.sort_unstable()`.  It's unlikely to become a problem I think.

kkysen:
I've fixed most of the remaining `clippy` warnings.  Here are the remaining ones and the plan to fix them:

* [ ] [`clippy::duplicate_underscore_argument`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  I'll ask about the rationale for this because it's about naming.

* [ ] [`clippy::large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  These (a few) have been marked with comments explaining their size.  We'll leave these as warnings for now and perhaps fix them in a separate PR.

* [ ] [`clippy::type_complexity`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  There are still 2 remaining `clippy::type_complexity` warnings, which are not trivial to fix with `type` alias refactorings (not clear how to name things).  These might get simplified with future allocation and indirection fixes (see below), so I'll wait on that.  For now they remain warnings.

* [ ] [`clippy::borrowed_box`](https://rust-lang.github.io/rust-clippy/master/#borrowed_box)
* [ ] [`clippy::vec_box`](https://rust-lang.github.io/rust-clippy/master/#vec_box)
   These are unnecessary allocations and indirections lints.  They are likely to cause a lot of code churn having to fix levels of indirection everywhere, so we should fix this in a separate PR.

* [ ] [`clippy::if_same_then_else`](https://rust-lang.github.io/rust-clippy/master/#if_same_then_else)
   For 1case of this, I'm not sure why this is the same in both branches.  Should ask about before fixing, so it remains a warning for now.
   For the other 2 cases of this, I have a fix, but I want to put it in a separate PR for more review and to make sure it keeps the same semantics.

* [ ] [`clippy::same_item_push`](https://rust-lang.github.io/rust-clippy/master/#same_item_push)
  I have a fix, but I want to put it in a separate PR for more review and to make sure it keeps the same semantics.

* [ ] [`clippy::needless_bool`](https://rust-lang.github.io/rust-clippy/master/#needless_bool)
  I have a fix, but I want to put it in a separate PR for more review and to make sure it keeps the same semantics.

kkysen:
Hi @afetisov, do you have any remaining concerns?  I outlined what the plan for the rest of the warnings is going forward.  For all of the `#![allow(...)]`s, I either fixed them now, or left them for future PRs, as I explained in the above comment.  Otherwise, I'm ready to merge now.  Finally!  That was a lot.  Thanks for the help!

afetisov:
Honestly, I have no idea. This PR is 111 commits now, it's unreviewable. So... just merge it and get over with. If there are any issues they may be fixed later.

PR #461, opened 2022-06-22
Use `syn::parse_quote!` to reimplement the generation of `fn main()`

afetisov:
This PR reimplements `c2rust_transpile::translator::main_function` using `syn::parse_quote!` macro instead of explicitly constructed AST via c2rust-ast-builder. The PR consists of two commits. The first one purely swaps usages of `mk()` for `syn::parse_quote!`. The second restructures `convert_main` into a more idiomatic form, leveraging slice patterns and avoiding panicking item accesses.

This PR is intended foremost as a test drive of the `syn::parse_quote!` in the context of the transpiler. It is meant to showcase on a relatively simple and compact test case the benefits of that approach, and to gain experience with porting code to `parse_quote!`.

In my view, the benefits of `parse_quote!`-based solution are obvious. The code is much more compact, it doesn't require any redundant boxing of AST nodes, the API is familiar to any Rust programmer (it's just the usual Rust code), and the output of code generation is immediately visible. There is also some mild help available from the IDEs with regard to interpolated code (mostly related to syntax highlighting and brace matching).

The only downside that I have encountered is that the code is no longer typechecked at compile time: while the builder explicitly constructs the well-typed AST, turning any syntax errors into compile errors, the `parse_quote!` invocation is parsed only at runtime. This means that if the code within the macro is not well-formed, it will be uncovered only at runtime, and the errors may be confusing (the exact position of the error is not provided by the macro). Unfortunately, `parse_quote!` relies heavily on Rust's trait system, and const traits are still very unstable, so this will not improve in the foreseeable future.

On the other hand, the possibility of purely syntactic errors is quite low, since the code snippets within the macro are usually very short and natural. There was only one case which stumped me: the return type of function signature. I have tried to write `fn foo() -> #ret`, while the correct syntax is `fn foo() #ret`, with `ret` an `Option<ReturnType>`. In retrospect, it should have been obvious since otherwise `->` would be dangling, but it's lack in the correct syntax is still mildly disturbing for me.

In all other cases during my experiments (which included not just `convert_main`), I have not encountered any confusing behaviour.

Another potential issue is that `parse_quote!` is less amenable to incremental construction of AST than the builder. This may be side-stepped with intermediate well-typed parses, or with `quote!` macro (but it increases the likelihood of runtime errors). However, I have generally experienced that `parse_quote!` makes AST generation simple enough that some minor code duplication is a nice alternative to incremental builders.

If this PR is acceptable, I would like to gradually switch most code to the `syn::parse_quote!` API. I plan to do it in several PRs, since such sweeping changes are likely to cause merge conflicts with branches. Also, it's hard to test such sweeping changes, and I'm not sure that I would like to migrate _all_ usages of c2rust-ast-builder (some cases look simpler with the builder API, but they are rare).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -677,19 +677,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d32e9bd16cc02eae7db7ef620b392808b89f6a5e16bb3497d159c6b92a0f4f86"
 dependencies = [
  "backtrace",
- "failure_derive",
-]
-
-[[package]]
-name = "failure_derive"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aa4da3c766cd7a0db8242e326e9e4e081edd567072893ed320008189715366a4"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
- "synstructure",
 ]
 
 [[package]]
@@ -1676,18 +1663,6 @@ dependencies = [
  "proc-macro2",
  "quote",
  "unicode-ident",
-]
-
-[[package]]
-name = "synstructure"
-version = "0.12.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f36bdaa60a83aca3921b5259d5400cbf5e90fc51931376a9bd4a0eb79aa7210f"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
- "unicode-xid",
 ]
 
 [[package]]
--- a/c2rust-transpile/Cargo.toml
+++ b/c2rust-transpile/Cargo.toml
@@ -20,7 +20,7 @@ c2rust-bitfields = { version = "0.3.0", path = "../c2rust-bitfields" }
 clap = {version = "2.34", features = ["yaml"]}
 colored = "2.0"
 dtoa = "1.0"
-failure = "0.1.5"
+failure = { version = "0.1.5", default-features = false, features = ["std"] }
 fern = { version = "0.6", features = ["colored"] }
 handlebars = "4.2"
 indexmap = { version = "1.0.1", features = ["serde-1"] }
--- a/c2rust-transpile/src/translator/main_function.rs
+++ b/c2rust-transpile/src/translator/main_function.rs
@@ -32,194 +32,69 @@ impl<'c> Translation<'c> {
                 .get(&main_id)
                 .expect("Could not find main function in renamer");
 
-            let decl = mk().fn_decl("main", vec![], None, ReturnType::Default);
-
-            let main_fn = mk().path_expr(vec![main_fn_name]);
-
-            let exit_fn = mk().abs_path_expr(vec!["std", "process", "exit"]);
-            let args_fn = mk().abs_path_expr(vec!["std", "env", "args"]);
-            let vars_fn = mk().abs_path_expr(vec!["std", "env", "vars"]);
-
-            let no_args: Vec<Box<Expr>> = vec![];
+            let main_fn = syn::Ident::new(&main_fn_name, Span::dummy());
 
             let mut stmts: Vec<Stmt> = vec![];
             let mut main_args: Vec<Box<Expr>> = vec![];
+
+            let get_arg_ty = |idx: CParamId, name: &str| match self.ast_context[idx].kind {
+                CDeclKind::Variable { ref typ, .. } => self.convert_type(typ.ctype),
+                _ => Err(TranslationError::generic(
+                    format!("Cannot find type of '{}' argument in main function", name),
+                )),
+            };
 
             let n = parameters.len();
 
             if n >= 2 {
                 // `argv` and `argc`
 
-                stmts.push(mk().local_stmt(Box::new(mk().local(
-                    mk().mutbl().ident_pat("args"),
-                    Some(mk().path_ty(vec![mk().path_segment_with_args(
-                        "Vec",
-                        mk().angle_bracketed_args(vec![
-                            mk().mutbl().ptr_ty(mk().path_ty(vec!["libc", "c_char"])),
-                        ]),
-                    )])),
-                    Some(
-                        mk().call_expr(
-                            mk().path_expr(vec!["Vec", "new"]),
-                            vec![] as Vec<Box<Expr>>,
-                        ),
-                    ),
-                ))));
-                stmts.push(mk().semi_stmt(mk().for_expr(
-                    mk().ident_pat("arg"),
-                    mk().call_expr(args_fn, vec![] as Vec<Box<Expr>>),
-                    mk().block(vec![mk().semi_stmt(mk().method_call_expr(
-                        mk().path_expr(vec!["args"]),
-                        "push",
-                        vec![mk().method_call_expr(
-                            mk().method_call_expr(
-                                mk().call_expr(
-                                    mk().abs_path_expr(vec!["std", "ffi", "CString", "new"]),
-                                    vec![mk().path_expr(vec!["arg"])],
-                                ),
-                                "expect",
-                                vec![mk().lit_expr("Failed to convert argument into CString.")],
-                            ),
-                            "into_raw",
-                            vec![] as Vec<Box<Expr>>,
-                        )],
-                    ))]),
-                    None as Option<Ident>,
-                )));
-                stmts.push(mk().semi_stmt(mk().method_call_expr(
-                    mk().path_expr(vec!["args"]),
-                    "push",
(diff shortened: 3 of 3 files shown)


kkysen:
I haven't worked much on this code—I think @fw-immunant would have more detailed feedback—but in general, I really like the `syn::parse_quote!` approach.  The loss of type checking isn't the greatest, but it is tremendously more readable and writable, which I think is more important here.  Thank you!

kkysen:
For documentation, and for making things more readable and writable, I think it would be useful to do something similar to what `syn::parse_quote!` is doing: allowing you to easily see what the output of the code will look like.  That is, we should include a lot more example outputs as type-checked doctests that are as inline as possible (to the function boundary where doctests can go).  We'd have to keep it up-to-date with the builder code, but it'd definitely help a lot for both readability and for knowing how to modify and extend things.

Furthermore, I think it would be beneficial to get some of the benefits of the `syn::parse_quote!` approach in terms of readability, but with increased type-checking, more than the status quo, by having it type-checked as first-class code within the transpiler.  For example, instead of using things like

```rust
mk().abs_path_expr(vec!["std", "process", "exit"])
```

we should use

```rust
use std::process;

exit.type_name_of_val()
```

with the following blanket `impl`:

```rust
pub trait TypeName {
    fn type_name() -> &'static str {
        std::any::type_name::<Self>()
    }
    
    fn type_name_of_val(&self) -> &'static str {
        Self::type_name()
    }
}

impl<T> TypeName for T {}
```

This has a bunch of benefits:
* The type is fully type-checked, which it currently isn't.
* We use the canonical, fully-qualified path that works in all contexts.  For example, the above `null_mut::<T>.type_name_of_val()` expands to `core::ptr::null_mut<T>`, using `core` instead of `std`.
* It works well with generics, too, at least concrete generic instantiations.
* The type/function in our code can be imported and referred to with just its local name, while we still get the fully-qualified path.
* We use the type/function itself, which means Intellisense works on it.

However, it doesn't work with:
* macros
* generic, non-instantiated types/functions
* aliases and re-exports

Furthermore, for more involved code we generate, I think it would be better to move that to a simple library crate.  Then the code can be fully type-checked there and we can just invoke small wrappers in the generated code, and with the type name trick above, we ensure all the paths match up at compile time.  This would also make the generated code easier to read, less redundant, and more amenable to static analysis, because we preserve more information and static analysis can know about our library helper functions.

Maybe I should move this to another issue.

afetisov:
@fw-immunant I must object to your assessment.

> Runtime performance (parsing tokens at runtime)

Maybe, but most likely there will be no difference, maybe even the `parse_quote!` code will be more performant. Parsing tokens is a relatively trivial thing to do, that's not a bottleneck in any compiler. On the other hand, `parse_quote!` will result in much fewer memory allocations. Take a look at the builder API, it passes boxed nodes everywhere, and the arguments commonly take vectors. `parse_quote!`, on the other hand, can directly output unboxed nodes, with an absolutely minimum number of allocations required for the AST. It also doesn't use any intermediate allocations, since the nodes are serialized into a single preallocated vector of tokens, and then directly deserialized.

Take a look at the implementations of `syn::parse::Parse` for AST nodes. They are almost literally the same as the explicit construction of nodes that is in `c2rust-ast-builder`, all the recursive `parse` calls will be monomorphized and inlined. Basically the only difference is that it checks for error on every token. You will also see very, very few allocations in that code. For comparison, around 35 of the methods in `builder::Builder` return `Box<Expr>`, and around 15 return `Box<Item>`. Those hundreds of small one-off unshared allocations will absolutely _trump_ any possible overhead of token parsing.

If you don't believe me, I can make a benchmark to compare the performance of two approaches. I would be _very_ surprised if `Builder` turned out more performant on non-trivial code snippets.

Actually I don't understand why `Builder` uses _so many boxes_. I just assume there was some good reason which I don't see. But if it's just an artifact of the MVP, removing all of those allocations will likely cause just as much churn as moving to a syn-based API. These assumptions are used everywhere, with explicit boxes and vectors created, exposed in functions signatures, written in explicit types and used in pattern matching, which requires extra dereferences of the scrutinee.

> Compile-time performance (more proc macros)

Ah, here is the great part! _The `syn` and `quote` crates  don't contain any proc macros!_ (well, unless you enable the `proc-macro` feature, which is used when _writing_ proc-macros and not required to parse tokenstreams at runtime). If you haven't looked at the implementations of `parse_quote!` and `quote!`, I encourage you to do so. I find them a real marvel of engineering. They implement the quote-unquote functionality purely via surprisingly simple declarative macros! The heavyweight magic is offloaded to the trait system.

Now, I don't know how performant declarative macros actually are on the code we require, but at least it doesn't require compiling the proc-macro machinery. 

> Compile-time checking for AST well-formedness

The AST is well-formed in either case. Parsing cannot produce invalid AST. The only difference is that constructing a node via `Builder` is always infallible, while constructing it via `parse_quote!` can panic at runtime if the quote was ill-formed. This is an issue, but a minor one. It is as simple as having tests which execute every AST interpolation (you don't even need any assertions on that quote, if it parses without error than it's okay). If you don't have tests for some code generation, type checking won't help you in ensuring its correctness (you can't check the semantics of transpiled code at compile time).

Now, it is _theoretically_ possible that an incorrectly written quote will splice a node in a way which will result in well-formed but wrong AST. But:

1. It is _extremely_ unlikely to arise in practice, if you're splicing parsed AST nodes and not arbitrary quotes. I have never encountered it, and it is hard to even construct a counterexample where an invalid quote would accidentally parse for some specific value of splices types, items and expressions.
2. This error is also possible with a `Builder` API, since it boils down to a misuse of strongly typed API. The exact errors will be different, but nothing stops you from using the `Builder` to attach attributes to integer literals, visibility to trait methods, or binding by reference instead of by value. The verbosity and indirection of `Builder` make those errors somewhat more likely, if anything.
3. It's _really_ hard to mess up something like `let x: usize = ::std::mem::size_of::<#T>();`, while messing up its construction via the builder is quite trivial. Here is a comparison:

```rust
syn::parse_quote! { let x: usize = ::std::mem::size_of::<#T>(); }

mk().local_stmt(Box::new(
    mk().local(
        mk().ident_pat("x"),
        Some(mk().ident_ty("usize")),
        Some(mk().call_expr(
            mk().abs_path_expr(
                vec![
                    mk().path_segment("std"),
                    mk().path_segment("mem"),
                    mk().path_segment_with_args(
                        "size_of",
                        mk().angle_bracketed_args(vec![ty]),
                    ),
                ]
            ),
            vec![] as Vec<Box<Expr>>,
        )),
    )
))
```

Dedicated methods which encapsulate the construction of common AST nodes, like `compute_size_of_ty` for `size_of` above, certainly help, but far from everything can be packaged in a pretty method. You can see how much ad-hoc AST generation code is in the `main_function.rs`, and there are many similar complex generators of statements and expression in the codebase. I don't think you can dismiss is just as "Code is more approachable", this code is much more maintainable.

> Churn--this will change a bunch of code, which is especially concerning because it simultaneously removes compile-time checks from it

Well, I can't object to the churn part, but I believe it is well justified, just like fixing warnings and implementing proper error handling is justified, even if it's a lot of churn. But the "removes compile-time checks from it" part I disagree with, and I have argued above why it is not the case. The only compile-check that a `Builder` guarantees is that the AST node has the required type - but this is also true of `parse_quote!`, as long as you test once that it actually _does_ parse. It's not like using an untyped language where testing requires testing literally all inputs and you still don't get any hard guarantees.

Messing up not the syntactic structure but the actual _semantics_ of transpiled code is equally easy with both APIs. Nothing prevents you from using a wrong binding or an incorrect method.

> We still need to manually construct and splice code as values fairly often (e.g., any place the .span() method of the current builder API is used)

That's not really a cost, it applies equally to both approaches. On the flip side, `parse_quote!` allows to construct more complex snippets of code directly, without complex construction code and intermediate splicing. Again, compare the implementations of `compute_size_of_ty` with those approaches:

```rust
fn compute_size_of_ty(&self, ty: Box<Type>) -> Result<WithStmts<Box<Expr>>, TranslationError> {
    let name = "size_of";
    let params = mk().angle_bracketed_args(vec![ty]);
    let path = vec![
        mk().path_segment("core"),
        mk().path_segment("mem"),
        mk().path_segment_with_args(name, params),
    ];
    let call = mk().call_expr(mk().abs_path_expr(path), vec![] as Vec<Box<Expr>>);

    Ok(WithStmts::new_val(call))
}

fn compute_size_of_ty(&self, ty: Box<Type>) -> Result<WithStmts<Box<Expr>>, TranslationError> {
    Ok(WithStmts::new_val(syn::parse_quote!{ ::core::mem::size_of::<#ty>() }))
}
```

The second one is so simple, it could be easily inlined into the quote at the use site!

PR #464, opened 2022-06-27
Disable building `json2cbor.c` in `tinycbor`, and which doesn't build on macOS/BSDs when `cjson` is installed

kkysen:
Fixes https://github.com/immunant/c2rust/issues/353.

This unconditionally disables building `tools/json2cbor/` in [`tinycbor`](https://github.com/intel/tinycbor), which isn't needed for our uses of `tinycbor` and is an optional dependency.  Normally it is autodetected when [`cjson`](https://github.com/DaveGamble/cJSON) is found, and it still is, but now we append to the generated `.config` file to explicitly disable it.

On macOS/BSD, `tools/json2cbor/json2cbor.c` doesn't build because it defines `_POSIX_C_SOURCE`, which in BSD libc, prevents `asprintf` from being declared, which is used by `tools/json2cbor/json2cbor.c`.  This can be fixed by removing the `_POSIX_C_SOURCE` define, but that'd have to be fixed in a patch or upstream in https://github.com/intel/tinycbor, which it probably should be, but the fix would take longer.

However, either way, we don't need `json2cbor`, so we can just always disable it, and it's better to be explicit about our dependencies than unknowingly having an optional dependency on `cjson` that affects the build.

### DIFF
--- a/c2rust-ast-exporter/src/CMakeLists.txt
+++ b/c2rust-ast-exporter/src/CMakeLists.txt
@@ -19,7 +19,7 @@ ExternalProject_Add(tinycbor_build
             # patch from upstream:
             # https://github.com/intel/tinycbor/commit/6176e0a28d7c5ef3a5e9cbd02521999c412de72c
             PATCH_COMMAND patch --forward -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/tinycbor_fix_build.patch || true
-            CONFIGURE_COMMAND ""
+            CONFIGURE_COMMAND make .config && cat ${CMAKE_CURRENT_SOURCE_DIR}/tinycbor.config >> .config
             BUILD_COMMAND make --quiet prefix=<INSTALL_DIR> CFLAGS=-fPIC
             INSTALL_COMMAND make --quiet prefix=<INSTALL_DIR> install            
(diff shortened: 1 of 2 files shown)


kkysen:
@gcxfd, can you see if this fixes your `tinycbor` issue when compiling from `cargo install c2rust`?

kkysen:
It's fixed upstream, too, now: https://github.com/intel/tinycbor/pull/221.

So we don't need this fix anymore (if we upgrade our tinycbor version), but I think it's still worth it to have, because otherwise we're pulling in an unneeded dependency, but only sometimes, and there could be some other future error there.

kkysen:
https://github.com/intel/tinycbor/pull/221 has now been tested on macOS and it works, so we just need to test this one now (with [`cjson`](https://github.com/DaveGamble/cJSON) installed globally) if anyone with a mac can:

```sh
cargo install c2rust --git https://github.com/immunant/c2rust --branch kkysen/tinycbor-disable-cjson
```

Once that's tested, we should be able to merge this I think.

boyland-pf:
This also compiles on mac.

fw-immunant (approved):
Thanks for tracking down the issue here. Fix looks good.

PR #465, opened 2022-06-27
Switched `tinycbor` to use `GIT_REPOSITORY` and `GIT_TAG` instead of `URL` and `URL_HASH` in `ExternalProject_Add`

kkysen:
Switched `tinycbor` to use `GIT_REPOSITORY` and `GIT_TAG` instead of `URL` and `URL_HASH` in `ExternalProject_Add`.

`tinycbor` is already downloaded from GitHub, so we can just use the built-in git support.  This makes upgrading `tinycbor` versions easier.  Also, we can use either git tags/branches or commit hashes.  I used a commit hash here so we keep the integrity checks (which are SHA-1 now vs. MD5), but we could use git tags for more flexibility (easy to upgrade and they could patch/update the tag).

I added this because I was thinking about upgrading the `tinycbor` version, as we might want to upstream a fix for https://github.com/immunant/c2rust/issues/353, where the `_POSIX_C_SOURCE` define causes `asprintf` to not be declared in BSD libc.  I found a workaround for that by appending to the generating `tinycbor/.config` to disable `cjson` and `json2cbor` support since we don't need it anyways, but I'm not sure if that's the best long-term approach.

### DIFF
--- a/c2rust-ast-exporter/src/CMakeLists.txt
+++ b/c2rust-ast-exporter/src/CMakeLists.txt
@@ -4,17 +4,16 @@ project(ASTExporter)
 #################################################
 # TinyCBOR                                      #
 #################################################
-set(TINYCBOR_URL "https://github.com/intel/tinycbor/archive/v0.5.3.tar.gz"
-    CACHE STRING "tinycbor download URL")
-set(TINYCBOR_MD5 "2cd3af70d8749a7ddd5a8d04d09ea8f6" CACHE STRING "tinycbor archive md5 sum")
+# v0.5.3 tag, but using the commit hash instead for integrity checks (worth it?)
+set(TINYCBOR_TAG "755f9ef932f9830a63a712fd2ac971d838b131f1" CACHE STRING "tinycbor git tag/branch/commit hash")
 set(TINYCBOR_PREFIX "${CMAKE_BINARY_DIR}/tinycbor" CACHE STRING "tinycbor install prefix")
 
 include(ExternalProject)
 ExternalProject_Add(tinycbor_build
             PREFIX ${TINYCBOR_PREFIX}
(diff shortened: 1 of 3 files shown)


kkysen:
Failed on Fedora because `git` isn't installed.  I can fix the docker scripts to install `git`.  It's also quite useful for using the docker container as I discovered last time debugging Fedora issues.

kkysen:
Failed on Darwin because `rust-src` and `rustc-dev` components aren't installed for the right rust version.  I'm not sure why this happened now, but our current setup is brittle to changes.  We should specify the needed components in a `rust-toolchain.toml` file so they are always downloaded on `cargo build` if needed.

kkysen:
See, this time CI worked on Darwin even though I changed nothing related to it (only a Fedora CI fix that's not even used yet).

kkysen:
Should be all ready to go now if Darwin is working again.  Just needs a docker push from @thedataking for Fedora to pass CI.

thedataking:
It seems this change will require everyone who wants to build c2rust to have git installed. Most folks probably have git installed but on the other hand, I don't think we should take on more dependencies than strictly necessary. I think we should hold off on this.

kkysen:
I agree.  We likely want to just vendor `tinycbor` as a git submodule anyways, so in that case, this wouldn't matter.

I would like to add `git` to the Docker containers, though, as that's generally useful, and only Fedora doesn't have `git`.

thedataking:
> I would like to add git to the Docker containers, though, as that's generally useful, and only Fedora doesn't have git.

If you have a specific reason to add `git` to the containers, I'm fine with that. If you don't need it yet, let's wait until you do.

kkysen:
I don't strictly need it yet, but I need to add it every time I have to debug Fedora, and it's weird that all the other containers have it but not Fedora.

PR #466, opened 2022-06-28
Clarify names in dynamic instrumentation code

fw-immunant:
We had a lot of very similar names in the dynamic instrumentation code, and this set of renamings disambiguates so it's easier to follow what's going on.

The former `FunctionInstrumenter` (now `CollectFunctionInstrumentationPoints`) did not actually instrument functions, it just collected a set of points to instrument.

Similarly, `add_instrumentation` (now `add_instrumentation_point`) does not add the instrumentation itself, but adds an instrumentation point to the set.

`instrument` (now `instrument_body`) is clearer if its name includes the scope.

Finally, `do_instrumentation` (now `apply_instrumentation`) is clearer if it actually says what it does to the instrumentation--it applies it to a function by rewriting the function body.

A bonus tweak is that this moves the definition of `main_did` to where it's used.

### DIFF
--- a/dynamic_instrumentation/src/instrument_memory.rs
+++ b/dynamic_instrumentation/src/instrument_memory.rs
@@ -50,7 +50,7 @@ impl InstrumentMemoryOps {
             tcx.item_name(body_did).to_string(),
         );
         debug!("Body before instrumentation: {:#?}", body);
-        instrument(self, tcx, body, body_did);
+        instrument_body(self, tcx, body, body_did);
         debug!("Body after instrumentation: {:#?}", body);
     }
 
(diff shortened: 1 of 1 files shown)


dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -664,29 +664,31 @@ fn instrument<'tcx>(
...
         state,
-        &instrumenter.into_instrumentation_points(),
+        &collect_points.into_instrumentation_points(),
         tcx,
         body,
         body_def_hash,
     );
 
+    // Apply `main`-specific instrumentation if this fn is main
+    let main_did = tcx.entry_fn(()).map(|(def_id, _)| def_id);
```
kkysen, line 685: This is a bit separate from the changes here, but `did` as an abbreviation for `def_id` is pretty bad.  `did` is another word, and `def_id` is very long, and both `def`, `id`, and `def_id` would make more sense than `did` to me.

fw-immunant: Agreed, but `did` for `DefId` is a naming convention already present in MIR.

kkysen:
I think all the renames make sense, so it should be good to merge once it passes CI.

dynamic_instrumentation/src/instrument_memory.rs
kkysen: That's unfortunate, but it makes sense if they already do it.

kkysen (approved):

PR #469, opened 2022-06-29
Added pdg tests, the query, graph deduplication, and an improved textual representation

kkysen:
Added:
* pdg tests that check if:
    * there are no duplicate graphs
    * all graphs' first/head node has no source
* the pdg nodes that need write permissions query:
    * implemented with @aneksteind from his pseudocode
* graph deduplication:
    * ended up being very simple to do at the end with an `Itertools::unique` call
* an improved textual representation that's much more concise and readable
* a bunch of `clippy` fixes, mostly autofixes

### DIFF
--- a/pdg/src/graph.rs
+++ b/pdg/src/graph.rs
@@ -210,9 +210,15 @@ impl Display for Node {
     }
 }
 
+impl Display for GraphId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "graph {}", self.as_usize())
+    }
+}
+
 impl Display for Graph {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        writeln!(f, "graph {{")?;
+        writeln!(f, "{{")?;
         for (node_id, node) in self.nodes.iter_enumerated() {
             writeln!(f, "\t{node_id}: {node},")?;
         }
@@ -223,11 +229,11 @@ impl Display for Graph {
 
 impl Display for Graphs {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        for (i, graph) in self.graphs.iter().enumerate() {
-            if i != 0 {
+        for (graph_id, graph) in self.graphs.iter_enumerated() {
+            if graph_id.as_usize() != 0 {
                 write!(f, "\n\n")?;
             }
-            write!(f, "{graph}")?;
+            write!(f, "{graph_id} {graph}")?;
         }
         Ok(())
     }
--- a/pdg/src/graph.rs
+++ b/pdg/src/graph.rs
@@ -47,8 +47,14 @@ pub struct Func(pub DefPathHash);
 
 impl Debug for Func {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        let x = c2rust_analysis_rt::DefPathHash::from(self.0 .0.as_value());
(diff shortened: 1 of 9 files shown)


pdg/src/graph.rs
```diff
@@ -71,7 +81,7 @@ pub struct Node {
     /// operation.  As in `rustc_middle::mir::Location`, an index less than the number of
     /// statements in the block refers to that statement, and an index equal to the number of
     /// statements refers to the terminator.
-    pub index: usize,
+    pub statement_index: usize,
```
aneksteind, line 84: we can change to `statement_idx` so that elsewhere we don't need to do:
```
statement_index: statement_idx
```

kkysen: Sure, I'll change that.

pdg/src/query.rs
```diff
@@ -0,0 +1,58 @@
+use linked_hash_set::LinkedHashSet;
+
+use crate::graph::{Graph, NodeId, NodeKind};
+
+impl Graph {
+    /// it would be nice to test properties like "an object is created in foo,
```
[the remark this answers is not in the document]
aneksteind: A paraphrase of the purpose and behavior is what I would put instead

pdg/src/util.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    pub fn len(&self) -> usize {
+        self.duplicates.len()
+    }
+}
+
+impl<T> Duplicates<T> {
+    const DEFAULT_UP_TO_N: usize = 5;
+
+    pub fn fmt_up_to_n(
```
aneksteind, line 65: are there existing libraries that give you something close to this amount of control? seems like a lot of custom code

aneksteind (approved):
lgtm other than the other comments I posted. if addressed to some extent you find reasonable, this is good to merge

pdg/src/util.rs
kkysen: I don't think so.  I customized the error message quite a bit.  Maybe there's some existing crate but I'm not quite sure how to search for something so specific.  `Itertools` is doing the heavy lifting of the actual grouping/deduplication though.

I probably wouldn't have spent this much time on the deduplication test seeing how simple the fix is now, but I hadn't realized it was that simple yet.  Still, I already wrote the test so might as well keep it, and it might help later if we need to switch to on-the-fly deduplication for memory reasons.

pdg/src/graph.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/469/commits/95d35db08266a55eaf806cf3adec9ae2844fe0eb.

pdg/src/query.rs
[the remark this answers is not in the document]
kkysen: Fixed in https://github.com/immunant/c2rust/pull/469/commits/bf47a0b6a46f7af9ac2bc6110a97fb2ed017db30.  Does it look good?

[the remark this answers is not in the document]
aneksteind: Made one small suggestion

aneksteind:
@kkysen it's definitely an improvement, I'm satisfied with it

pdg/src/query.rs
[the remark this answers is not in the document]
kkysen: Fixed in https://github.com/immunant/c2rust/pull/469/commits/146f299a034889ec89d9c94eb5fc043a794f576e.

Issue #447, opened 2022-06-15
Qualified paths like `libc::c_int` should be globally qualified like `::libc::c_int` to avoid name clashes


kkysen:
Right now, most paths that `c2rust transpile` generates are fully-qualified, like `libc::c_int`.  This can be ambiguous, however, when there is another item named `libc`, which would shadow the `libc` crate.  To fix this, we should make the qualified paths global, i.e., prefix them with a `::` like `::libc::c_int`, which causes the lookup to be only within crates, not within all in-scope items, which is what we want.

This can cause practical issues, like in the `tests/longdouble` test.  When I added a `typedef long double f128` to the C code, then the generated Rust code included a `pub type f128 = f128::f128`.  That is, the `f128` type alias clashed with the `f128` crate.  If `pub type f128 = ::f128::f128` was generated, it would work.

PR #475, opened 2022-06-30
Refactor repeated `std_or_core` logic

kkysen:
Refactored out `std_or_core = if self.tcfg.emit_no_std { "core" } else { "std" }` into `Translation::std_or_core(&self)` and `fn std_or_core(emit_no_std: bool)`.  This was repeated in lots of places all over the transpiler.  There are lots of similar de-duplication refactors we could do, but this one was simple and I did while working on another PR first.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4608,13 +4608,7 @@ impl<'c> Translation<'c> {
 
         if self.ast_context.is_va_list(resolved_ty_id) {
             // generate MaybeUninit::uninit().assume_init()
-            let name = "uninit";
-            let path = vec![
-                mk().path_segment(self.std_or_core()),
-                mk().path_segment("mem"),
-                mk().path_segment("MaybeUninit"),
-                mk().path_segment(name),
-            ];
+            let path = vec![self.std_or_core(), "mem", "MaybeUninit", "uninit"];
             let call = mk().call_expr(mk().abs_path_expr(path), vec![] as Vec<Box<Expr>>);
             let call = mk().method_call_expr(call, "assume_init", vec![] as Vec<Box<Expr>>);
             return Ok(WithStmts::new_val(call));
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -281,7 +281,11 @@ pub struct Translation<'c> {
 }
 
 pub fn std_or_core(emit_no_std: bool) -> &'static str {
-    if emit_no_std { "core" } else { "std" }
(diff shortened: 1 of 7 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -4593,10 +4598,9 @@ impl<'c> Translation<'c> {
...
         if self.ast_context.is_va_list(resolved_ty_id) {
             // generate MaybeUninit::uninit().assume_init()
-            let std_or_core = if self.tcfg.emit_no_std { "core" } else { "std" };
             let name = "uninit";
             let path = vec![
-                mk().path_segment(std_or_core),
+                mk().path_segment(self.std_or_core()),
                 mk().path_segment("mem"),
                 mk().path_segment("MaybeUninit"),
                 mk().path_segment(name),
```
fw-immunant, line 4606: I think this can just be a `Vec<&str>`, actually.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3177,11 +3183,10 @@ impl<'c> Translation<'c> {
...
     fn compute_size_of_ty(&self, ty: Box<Type>) -> TranslationResult<WithStmts<Box<Expr>>> {
-        let std_or_core = if self.tcfg.emit_no_std { "core" } else { "std" };
         let name = "size_of";
         let params = mk().angle_bracketed_args(vec![ty]);
         let path = vec![
-            mk().path_segment(std_or_core),
+            mk().path_segment(self.std_or_core()),
             mk().path_segment("mem"),
             mk().path_segment_with_args(name, params),
         ];
```
fw-immunant, line 3192: This can also be `Vec<&str>` (without `mk().path_segment` calls).

fw-immunant (approved):
Looks good in general; I mentioned inline a couple additional cleanups that could be made.

We could also make shorthand for `mk().abs_path(vec![self.std_or_core(), ...]);`, because that's how `std_or_core` is generally used--as a way to build an absolute item path.

c2rust-transpile/src/translator/mod.rs
kkysen: It can't because `mk().path_segment_with_args(name, params)` is `PathSegment`.

kkysen:
Yeah, we can definitely make a lot more shorthands, but there's a ton of duplicated code so I'll save that for later.  I think I'll fix that when I standardize the paths (make them all absolute (https://github.com/immunant/c2rust/issues/447) and always use `std` or `core` right).  Speaking of that, why don't we just use `core` everywhere?  A later refactor by the user of replacing `core::` with `std::` is very easy; just search and replace.

PR #476, opened 2022-06-30
Fixed remaining `clippy::borrowed_box` warnings

kkysen:
Some of these were fixed by turning `&Box<Expr>`s into `&Expr`s and then re-adding explicit `Box::new`s where needed.  And some others were fixed by turning `&Box<Expr>` into `Box<Expr>` when all uses were `.clone()`ed anyways and call sites were supplied by `&arg` reference and were never otherwise moved out of.


fw-immunant (approved):
LGTM.

Issue #470, opened 2022-06-29
docs.rs/c2rust has been broken since `0.10.0`


kkysen:
https://docs.rs/c2rust has been broken since `0.10.0`.

`cargo doc` works locally, but on docs.rs, it [fails](https://docs.rs/crate/c2rust-transpile/0.16.0/builds/557447):

```shell
[INFO] [stderr]   running: "cmake" "--build" "." "--target" "clangAstExporter" "--config" "Debug" "--"
[INFO] [stderr]   [  8%] Performing download step (download, verify and extract) for 'tinycbor_build'
[INFO] [stderr]   -- verifying file...
[INFO] [stderr]          file='/opt/rustwide/target/x86_64-unknown-linux-gnu/debug/build/c2rust-ast-exporter-1eea402c4d720905/out/build/tinycbor/src/v0.5.3.tar.gz'
[INFO] [stderr]   -- MD5 hash of
[INFO] [stderr]       /opt/rustwide/target/x86_64-unknown-linux-gnu/debug/build/c2rust-ast-exporter-1eea402c4d720905/out/build/tinycbor/src/v0.5.3.tar.gz
[INFO] [stderr]     does not match expected value
[INFO] [stderr]       expected: '2cd3af70d8749a7ddd5a8d04d09ea8f6'
[INFO] [stderr]         actual: 'd41d8cd98f00b204e9800998ecf8427e'
[INFO] [stderr]   -- File already exists but hash mismatch. Removing...
[INFO] [stderr]   -- Downloading...
[INFO] [stderr]      dst='/opt/rustwide/target/x86_64-unknown-linux-gnu/debug/build/c2rust-ast-exporter-1eea402c4d720905/out/build/tinycbor/src/v0.5.3.tar.gz'
[INFO] [stderr]      timeout='none'
[INFO] [stderr]   -- Using src='https://github.com/intel/tinycbor/archive/v0.5.3.tar.gz'
[INFO] [stderr]   -- Retrying...
[INFO] [stderr]   -- Using src='https://github.com/intel/tinycbor/archive/v0.5.3.tar.gz'
[INFO] [stderr]   -- Retry after 5 seconds (attempt #2) ...
[INFO] [stderr]   -- Using src='https://github.com/intel/tinycbor/archive/v0.5.3.tar.gz'
[INFO] [stderr]   -- Retry after 5 seconds (attempt #3) ...
[...]

Issue #471, opened 2022-06-29
Make `tinycbor` dependency build offline


kkysen:
Building `tinycbor` by downloading it in `cmake` causes many issues:
* https://github.com/immunant/c2rust/issues/470
* https://github.com/immunant/c2rust/issues/356
* https://github.com/immunant/c2rust/pull/440

We should figure out how to build it offline or switch to a solution that can be.

kkysen:
We just use these functions:

```
cbor_encode_boolean
cbor_encode_byte_string
cbor_encode_double
cbor_encode_int
cbor_encode_null
cbor_encoder_close_container
cbor_encoder_create_array
cbor_encoder_get_buffer_size
cbor_encoder_get_extra_bytes_needed
cbor_encoder_init
cbor_encode_string
cbor_encode_string_array
cbor_encode_text_string
cbor_encode_text_stringz
cbor_encode_uint
```

and [`minicbor::encode::Encode`](https://docs.rs/minicbor/latest/minicbor/encode/struct.Encoder.html) has all the equivalent methods.

kkysen:
Making `tinycbor` a git submodule might work (https://github.com/immunant/c2rust/issues/470#issuecomment-1172257062).  But I don't think creating C bindings for `minicbor`, specifically the type-specific ones in [`minicbor::encode::Encode<Vec<u8>>`](https://docs.rs/minicbor/latest/minicbor/encode/struct.Encoder.html) to match exactly the ones we use listed above, would be that hard.  This comes with a few extra benefits as well:

* We don't have an assertion if the CBOR-encoded C AST is > 64 MBs because we use a growable `Vec<u8>` as the output, not a `(uint8_t *buffer, size_t size)`:
https://github.com/immunant/c2rust/blob/5d4e7a637196727eb6cf118628792d8b726dd556/c2rust-ast-exporter/src/AstExporter.cpp#L2604-L2615

* We could also use `minicbor` for deserializing from the Rust side instead of `serde_cbor`.  `serde_cbor` is no longer maintained and suggests using either `minicbor` or `ciborium`.  Of the two, `minicbor` had a lot more downloads (and is `no_std`-compatible, too).  If we switch to using `minicbor` on the Rust side, then we use the same library for serializing and deserializing, which has less of a chance for any bugs.

[...]

PR #482, opened 2022-07-01
Skip downloading `tinycbor` in docs.rs

kkysen:
Should fix https://github.com/immunant/c2rust/issues/470.

Skip doing `cmake` build (which includes downloading `tinycbor`) in docs.rs () because it restricts network access so otherwise our docs.rs build will fail.

According to https://docs.rs/about/builds#detecting-docsrs, we can check `env::var("DOCS_RS").is_ok()`.

It tested this with `DOCS_RS=1 cargo doc`, which builds fine.  `DOCS_RS=1 cargo build` fails with linker errors, but I *don't think* docs.rs does that, hopefully.

This is the simplest fix to get docs.rs working again for `c2rust`, but we should still f ix https://github.com/immunant/c2rust/issues/471 later.

### DIFF
--- a/c2rust-ast-exporter/build.rs
+++ b/c2rust-ast-exporter/build.rs
@@ -111,6 +111,11 @@ fn generate_bindings() -> Result<(), &'static str> {
 ///   - if `opt-level=0`                              then `CMAKE_BUILD_TYPE=Debug`
 ///   - if `opt-level={1,2,3}` and not `debug=false`, then `CMAKE_BUILD_TYPE=RelWithDebInfo`
 fn build_native(llvm_info: &LLVMInfo) {
+    if env::var("DOCS_RS").is_ok() {
+        // Don't build `cmake` things because it downloads `tinycbor`,
+        // and docs.rs has no network access
+        return;
+    }
     // Find where the (already built) LLVM lib dir is
     let llvm_lib_dir = &llvm_info.lib_dir;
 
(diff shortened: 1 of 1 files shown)


c2rust-ast-exporter/build.rs
```diff
@@ -119,6 +119,12 @@ fn build_native(llvm_info: &LLVMInfo) {
             println!("cargo:rustc-link-search=native={}", libdir);
         }
         _ => {
+            if env::var("DOCS_RS").is_ok() {
```
rinon, line 122: Could we just skip `build_native()` entirely if this env var is set? That would be simpler.

kkysen: Yeah, does pretty much the same thing.  Should it go at the top of `build_native` (I'd put it here) or around the call to `build_native`?

rinon: I'd prefer outside of build_native, to indicate that we should not even try to build in this docs mode.

kkysen:
Also, when do we plan on publishing a new release so we can test/use this for real?

c2rust-ast-exporter/build.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/482/commits/50ab8099512ce1c93bca801e053d2839e2d86860.

rinon:
> Also, when do we plan on publishing a new release so we can test/use this for real?

Not sure, but this seems like a good reason to do that, so let's plan for it. Maybe we can do that next week.

rinon (approved):

Issue #477, opened 2022-06-30
Catch more compile-time errors in CI


kkysen:
Right now in CI, we just test

```sh
cargo build
./scripts/test_translator.py tests/
```

We should test more:

```sh
cargo fmt --check
RUSTFLAGS='-D warnings' cargo clippy --tests --all --all-features # includes cargo build
RUSTDOCFLAGS='-D warnings' cargo doc --all-features --document-private-items
cargo test --all-features
./scripts/test_translator.py tests/
# immunant/c2rust-testsuite CI
```

Furthermore, we should run the external tests, `./scripts/test_translator.py tests/` and [`immunant/c2rust-testsuite`](https://github.com/immunant/c2rust-testsuite) CI inside of `cargo test`.  This not only simplifies testing workflows, but also hooks into external testing, such as crater runs.

I think we should definitely add these at least by the time we move CI to GitHub Actions instead of Azure Pipelines, but preferably earlier, as `cargo clippy` is almost passing (https://github.com/immunant/c2rust/issues/474), and once it is, I'd like to keep it that way.

thedataking:
> Furthermore, we should run the external tests, ./scripts/test_translator.py tests/ and [immunant/c2rust-testsuite](https://github.com/immunant/c2rust-testsuite) CI inside of cargo test.

I think `cargo test` should not run `immunant/c2rust-testsuite` since the latter has complex dependencies. Let's keep it separate.

kkysen:
Okay, I haven't really looked into the details of what `immunant/c2rust-testsuite` does, so that might make sense.  I do think running the in-repo `./scripts/test_translator.py tests/` as part of `cargo test` makes sense though, as that doesn't have many dependencies, and for ones there are, like the cross tests, we can just skip testing them if they're not available.

thedataking:
That sounds reasonable. `./scripts/test_translator.py tests/ tests/` has far fewer dependencies, so that should be okay.

Issue #474, opened 2022-06-30
Fix all `clippy` warnings


kkysen:
Fix all remaining `clippy` warnings.  Then we can test `cargo clippy` in CI to keep it that way.

* [x] vast majority of `clippy` warnings: https://github.com/immunant/c2rust/pull/460

* [x] [`clippy::duplicate_underscore_argument`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  https://github.com/immunant/c2rust/pull/497

* [x] [`clippy::large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  https://github.com/immunant/c2rust/pull/508
  https://github.com/immunant/c2rust/pull/509

* [x] [`clippy::type_complexity`](https://rust-lang.github.io/rust-clippy/master/#duplicate_underscore_argument)
  https://github.com/immunant/c2rust/pull/467

* [x] [`clippy::borrowed_box`](https://rust-lang.github.io/rust-clippy/master/#borrowed_box)
  https://github.com/immunant/c2rust/pull/476

* [x] [`clippy::vec_box`](https://rust-lang.github.io/rust-clippy/master/#vec_box)
  https://github.com/immunant/c2rust/pull/499 (locally allowed)
  Fixing this would be a massive change.  It cascades to basically remove every unneeded `Box`, which is generally good, but some of these types, `Expr`, `Type`, `Item`, are big, around 400 bytes, so the extra `memcpy`s might not be worth it, but who knows.  But also there are a ton of unnecessary allocations happening currently.  The best possible solution would probably be to re-introduce a type like `P`, which was in `libsyntax` in `rustc` I believe, and which was arena-allocated.  But `syn` uses `Box`es, so we're kinda stuck as I don't think `syn`'s types have an `A: Allocator` generic.  The other alternative is just to `#![allow(clippy::vec_box)]`, or at least do it locally on the few individual warnings.  Maybe we should do that until more refactoring of the transpiler is done to de-duplicate a lot of it and add better APIs.

* [x] [`clippy::if_same_then_else`](https://rust-lang.github.io/rust-clippy/master/#if_same_then_else)
  https://github.com/immunant/c2rust/pull/462
  For the remaining case, I'm not sure why this is the same in both branches.  Should ask about before fixing, so it remains a warning for now.
  https://github.com/immunant/c2rust/pull/498

[...]

PR #483, opened 2022-07-01
Add `--all-features`, `cargo fmt --check`, `cargo check`, `cargo test`, `cargo doc`, and `RUSTFLAGS=-Dwarnings` to CI

kkysen:
Starts to f ix https://github.com/immunant/c2rust/issues/477.

This adds `--all-features`/`--features dynamic-instrumentation`, `cargo fmt --check`, `cargo check`, `cargo test`, `cargo doc`, and `RUSTFLAGS=-Dwarnings` to CI.  Only `cargo clippy` is left.

This adds about 5 minutes to CI, going from 10-15 minutes to 15-25 minutes.  This is mainly from:

* `cargo check --all-features`: ~5 minutes
* `cargo test`: ~3 minutes (mostly pdg snapshot and static analysis tests)

On the other hand, these take practically no time:

* `cargo fmt --check`: 1 second
* `cargo doc --no-deps`: 6 seconds

Previous we only had:

* [x] `cargo build`
* [x] `./scripts/test_translator.py tests/`

Now we have added:

* [x] `cargo fmt --check`
* [x] `RUSTFLAGS='-D warnings' cargo check --all-features` 
* [x] `RUSTFLAGS='-D warnings' cargo build`
* [x] `RUSTFLAGS='-D warnings' cargo test`
* [x] `RUSTDOCFLAGS='-D warnings' cargo doc --all-features --document-private-items`

We still have these to go:

* [ ] `RUSTFLAGS='-D warnings' cargo clippy --tests --all --all-features` (includes `cargo check`): https://github.com/immunant/c2rust/issues/474

For commands that check only and do not run (`cargo check`, `cargo doc`), we specify `--all-features`, which is equivalent to `--features llvm-static`.  But for commands that do build/run (`cargo build`, `cargo test`), we don't specify `--all-features` as we don't want to test `--features llvm-static` by default, and it doesn't work on the Arch and Fedora docker images (see https://github.com/immunant/c2rust/issues/500).

Currently, `cargo test -p c2rust-analyze` is excluded as it fails on some OSes in CI (https://github.com/immunant/c2rust/issues/593).  It will be re-enabled once that's fixed, but I want to get `cargo test` into CI as soon as possible.


kkysen:
Fails on Arch and Fedora during `cargo test` for two similar reasons:

[Arch](https://dev.azure.com/immunant/c2rust/_build/results?buildId=1850&view=logs&j=f7d03aff-a499-5cb4-81d5-3cd2462ec94c&t=748661c0-d104-5e8e-02ad-4796b765f966&l=81):

```shell
     Running unittests (target/release/deps/c2rust_ast_exporter-08a638edb21ed379)
: CommandLine Error: Option 'openmp-ir-builder-optimistic-attributes' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
error: test failed, to rerun pass '-p c2rust-ast-exporter --lib'
```

[Fedora](https://dev.azure.com/immunant/c2rust/_build/results?buildId=1850&view=logs&j=1bc19d9b-b6c8-52d0-07d6-07754b7d13fb&t=0ea4c305-8a5c-5f7f-3862-8dc92e4cf4f2&l=81):

```shell
     Running unittests (target/release/deps/c2rust_ast_exporter-08a638edb21ed379)
: CommandLine Error: Option 'help-list' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
error: test failed, to rerun pass '-p c2rust-ast-exporter --lib'
```

It seems we're passing multiple of the same options to `clang` in `c2rust-ast-exporter`.  I'm not sure where or why, though, or why it's only failing on Arch and Fedora.  They must use different `clang` versions, right?

kkysen:
Neither of the errors are reproducible in local docker builds; that's weird.

kkysen:
Doesn't appear to be flaky CI either; very odd.

kkysen:
It appears the same CI errors are happening in Arch in Fedora, but in `./test_translator.py` instead of `cargo test` (which I commented out).  I have no idea what's happening.  Maybe the CI steps aren't isolated from each other as I thought?

kkysen:
The errors went away when I removed `--all-features` from `cargo build`, so I'm guessing `c2rust-instrument` links against the wrong/multiple version(s) of LLVM on Arch and Fedora maybe, and that causes the issue.  `./test_translator.py` doesn't need `c2rust-instrument` and `--all-features`, so it should be okay if we can do `cargo check --all-features` and then only `cargo build` (with no `--all-features`) for now, but this is something we should fix.

kkysen:
Currently, anything that builds with `--all-features` (either `cargo build` and subsequent tests or `cargo test`) and runs those binaries on Arch or Fedora crash.  This is due to them linking twice to/linking to the wrong version of LLVM, and I'm not sure what causes this in `--feature dynamic-instrumentation`/`c2rust-instrument`.  Clearly we should fix that, but in the meantime, I could also split Arch and Fedora to have a different pipeline, as Darwin currently does (and thus can enable `--all-features` on `cargo build` and `cargo test`).  Or I could just leave it as is for now until we solve the issue, as we still run `cargo check --all-features`, so we should catch any compile-time errors; we're just missing `cargo test` for `c2rust-instrument` really.

See https://github.com/immunant/c2rust/issues/500 for more on this.

kkysen:
This should be good to go now.  Only `clippy` is skipped for now, though that's almost fixed as well (just have a couple tweaks left).

kkysen:
Once these PRs land, we can add `clippy`:
* https://github.com/immunant/c2rust/pull/497
* https://github.com/immunant/c2rust/pull/508
* https://github.com/immunant/c2rust/pull/509

kkysen:
The PDG snapshot tests are under `cargo test`, so we'd need to run that in CI like implemented here to test that now.

kkysen:
`clippy` is ready now (https://github.com/immunant/c2rust/issues/474), so I can add that here (just uncommenting) or put it in a separate PR.

thedataking:
How does the `check.sh` and `pdg.sh` scripts relate to the CI changes. I don't see them being referenced in the azure configuration file. If they are not needed for this PR, please remove them. 

[nitpick & digression] `pdg.sh` is too terse and `check.sh` too generic IMHO. Can you use names that help new users understand the purpose of these scripts and match the names of other scripts better?

thedataking:
> `clippy` is ready now (#474), so I can add that here (just uncommenting) or put it in a separate PR.

Generally speaking, smaller and more focused PRs are better. It helps the reviewer and leads to a cleaner revision history.

kkysen:
> 1. How do the `check.sh` and `pdg.sh` scripts relate to the CI changes? I don't see them being referenced in the azure configuration file. If they are not needed for this PR, please remove them.

https://github.com/immunant/c2rust/blob/342ebc6e0cc074c48c02c80dcc42732418ae3a44/scripts/pdg.sh#L80
For `pdg.sh`, `unset RUSTFLAGS` is done because CI sets `RUSTFLAGS="-D warnings"`, which would be inherited by `c2rust instrument` for `analysis/test`, and as transpiled code, `analysis/test` has many warnings that would then turn into errors in CI.

`check.sh` is not necessary for these CI changes, but reflects the checks done in CI so that it's easy to run them locally (which is way faster than waiting to see if it passes CI).  If you don't want them in this PR, then I think makes more sense to merge `check.sh` before the CI changes, but they're both in-sync, and I think it makes sense to add them together.
 
> 2. Now that we have substantially more test steps, it would be good to capture the motivation for some of the steps. For instance, it isn't immediately clear to me why it is beneficial to first run `cargo check` and then `cargo build`. I couldn't tell what problems the first step would catch that the second wouldn't. (This may be obvious to others but the point still stands.)

Okay, I can document that in comments in the code, though I explained the reasoning in the PR description.  `cargo check` we can run with `--all-features` but `cargo build` we cannot, so we need to check that code at some point.  Furthermore, once the `clippy` PR merges, `cargo check` will be replaced by `cargo clippy`, so it doesn't really matter.

> [nitpick & digression] `pdg.sh` is too terse and `check.sh` too generic IMHO. Can you use names that help new users understand the purpose of these scripts and match the names of other scripts better? Please fix in a separate PR if these were not committed by accident.

Okay, I can rename them.  `check.sh` I can do here (unless I put `check.sh` in a PR before this one), and `pdg.sh` already exists, so I can rename that separately.  `pdg.sh` does have lengthy documentation explaining what it does and what it's for, though.

> Generally speaking, smaller and more focused PRs are better. It helps the reviewer and leads to a cleaner revision history.

Not always, as not including `clippy` here is why everyone is asking why `cargo check` is done before `cargo build`.

thedataking:
> Okay, I can document that in comments in the code, though I explained the reasoning in the PR description.  

Thanks. You did write a great PR description but the chances that folks will read it is lower than if you put breadcrumbs in the test file itself. Thanks for highlighting the planned switch from `cargo check` to `cargo clippy`.

For `check.sh`, try to see what name would be consistent with the other contents of the `scripts` folder. Maybe `run_all_test_and_checks.sh` or `do_presubmit_testing.sh`? 

My preference would be for you to pick a more verbose name for `pdg.sh` in a separate PR but admittedly I am nitpicking. 

Other than that, LGTM!

thedataking (approved):
I did this wrong, see review feedback in https://github.com/immunant/c2rust/pull/483#issuecomment-1183789962

kkysen:
Okay, I'll fix the issues that with the snapshot test that are causing issues, document the CI steps inline, rename `check.sh` to `run_ci_checks.sh` (?), and deal with renaming `pdg.sh` later.  Then once this is merged, it'll test https://github.com/immunant/c2rust/issues/518 more thoroughly, and if that passes, I'll merge that one, too.

kkysen:
`analysis/test` fails on Darwin as `reallocarray` doesn't exist on macOS, as also discovered in https://github.com/immunant/c2rust/pull/497#issuecomment-1180781612.

kkysen:
@spernsteiner, CI for `cargo test` is failing on not finding `FileCheck` in Ubuntu 18 (but not 20) and Debian 10 (but not 11).  It must not be included in LLVM installations by default in older distros (I'm guessing).  We'd really like to have `cargo test` running in CI.

It appears the flaky CI bug *may* (🤞) be gone now after the major #554 PR that improved instrumentation/test times.

kkysen:
@spernsteiner, I tried https://github.com/immunant/c2rust/pull/536#discussion_r923606625, but `apt install llvm-tools` doesn't seem to be a thing.

spernsteiner:
> I tried [#536 (comment)](https://github.com/immunant/c2rust/pull/536#discussion_r923606625), but `apt install llvm-tools` doesn't seem to be a thing.

Apparently the package name is e.g. `llvm-11-tools`.  There is no version-independent `llvm-tools` alias.  However, it looks like `llvm-dev` depends on `llvm-$version-dev`, which depends on `llvm-$version-tools`, so maybe you can pull it in that way.

kkysen:
It seems that for LLVM 6 and earlier, it's `llvm-6.0-tools` and it's not installed by default with `llvm` like it is in later versions.  I found a fix for that now, but your `FileCheck` search code isn't finding it since it doesn't look for `FileCheck-6.0`.  I think a better solution is to look for `llvm-config` in the same way `c2rust-ast-exporter` does and then use `$(llvm-config --bindir)/FileCheck`.

Issue #480, opened 2022-07-01
`c2rust transpile --emit-no-std` is not always honored


kkysen:
`c2rust transpile --emit-no-std` is not always honored, as sometimes the transpiler code hardcodes `std`.

I looked through all of the current uses of `"std"`.  It appears most can be moved to `core` or `alloc`, or can be soon:

| `std` | `core`/`alloc` | Notes |
| ----- | -------------- | ----- |
| `std::mem::transmute` | `core::mem::transmute` | |
| `std::ptr::write_bytes` | `core::ptr_write_bytes` | |
| `std::ptr::null_mut` | `core::ptr::null_mut` | |
| `std::vec::from_elem` | `alloc::vec::from_elem` | apparently it's `#[doc(hidden)]` |
| `std::ffi::CStr` | `core::ffi::CStr` | `#![feature(core_c_str)]`, being stabilized in `1.63.0` |
| `std::ffi::CString` | `alloc::ffi::CString` | `#![feature(alloc_c_string)]`, being stabilized in `1.63.0` |
| `std::process::exit` | | needs `std` |
| `std::env::args` | | needs `std` |
| `std::env::vars` | | needs `std` |

PR #484, opened 2022-07-04
translator: remove `std_or_core()` function

afetisov:
This is a redundant complication of translator source code: the `libcore` is always available,
whether the crate is `#![no_std]` or not. Using the imports from `core` or `std` is purely a
matter of code style, which should not be the responsibility of the transpiler.

If the transpiled crate should use `libstd` instead of `libcore` in imports, this can be fixed
by some refactoring tools, or possibly even `rustfmt`.

Also fixes https://github.com/immunant/c2rust/issues/480.

### DIFF
--- a/c2rust-transpile/src/convert_type.rs
+++ b/c2rust-transpile/src/convert_type.rs
@@ -132,6 +132,12 @@ pub const RESERVED_NAMES: [&str; 103] = [
 ];
 
 impl TypeConverter {
+    // We don't provide a `Default` impl to simplify future compatibility:
+    // if `TypeConverter` ever gets fields incompatible with `Default`, then
+    // cleaning out the uses of `impl Default for TypeConverter` can be a pain.
+    // More practically, there is a single use of `TypeConverter::new` and no
+    // current plans to use a `Default` impl, so providing it isn't worth the
+    // potential breakage.
     #[allow(clippy::new_without_default)]
     pub fn new() -> TypeConverter {
         TypeConverter {
--- a/c2rust-transpile/src/convert_type.rs
+++ b/c2rust-transpile/src/convert_type.rs
@@ -2,7 +2,6 @@ use crate::c_ast::CDeclId;
 use crate::c_ast::*;
 use crate::diagnostics::TranslationResult;
 use crate::renamer::*;
-use crate::translator::std_or_core;
 use c2rust_ast_builder::{mk, properties::*};
 use failure::format_err;
 use std::collections::{HashMap, HashSet};
@@ -21,7 +20,6 @@ pub struct TypeConverter {
     fields: HashMap<CDeclId, Renamer<FieldKey>>,
     suffix_names: HashMap<(CDeclId, &'static str), String>,
     features: HashSet<&'static str>,
-    pub emit_no_std: bool,
 }
 
 pub const RESERVED_NAMES: [&str; 103] = [
@@ -134,14 +132,14 @@ pub const RESERVED_NAMES: [&str; 103] = [
 ];
 
 impl TypeConverter {
-    pub fn new(emit_no_std: bool) -> TypeConverter {
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> TypeConverter {
         TypeConverter {
             translate_valist: false,
             renamer: Renamer::new(&RESERVED_NAMES),
             fields: HashMap::new(),
             suffix_names: HashMap::new(),
             features: HashSet::new(),
-            emit_no_std,
         }
     }
 
@@ -301,7 +299,7 @@ impl TypeConverter {
         ctype: CTypeId,
     ) -> TranslationResult<Box<Type>> {
         if self.translate_valist && ctxt.is_va_list(ctype) {
-            let path = vec![std_or_core(self.emit_no_std), "ffi", "VaList"];
+            let path = vec!["core", "ffi", "VaList"];
             let ty = mk().path_ty(mk().abs_path(path));
             return Ok(ty);
         }
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -280,20 +280,6 @@ pub struct Translation<'c> {
     cur_file: RefCell<Option<FileId>>,
 }
 
-pub fn std_or_core(emit_no_std: bool) -> &'static str {
-    if emit_no_std {
-        "core"
-    } else {
-        "std"
-    }
-}
-
-impl Translation<'_> {
-    pub fn std_or_core(&self) -> &'static str {
-        std_or_core(self.tcfg.emit_no_std)
-    }
-}
-
 fn simple_metaitem(name: &str) -> NestedMeta {
     let meta_item = mk().meta_path(name);
 
@@ -362,20 +348,12 @@ fn unwrap_function_pointer(ptr: Box<Expr>) -> Box<Expr> {
     mk().method_call_expr(ptr, "expect", vec![err_msg])
 }
 
-fn transmute_expr(
-    source_ty: Box<Type>,
-    target_ty: Box<Type>,
-    expr: Box<Expr>,
-    no_std: bool,
-) -> Box<Expr> {
+fn transmute_expr(source_ty: Box<Type>, target_ty: Box<Type>, expr: Box<Expr>) -> Box<Expr> {
     let type_args = match (&*source_ty, &*target_ty) {
         (Type::Infer(_), Type::Infer(_)) => Vec::new(),
         _ => vec![source_ty, target_ty],
     };
-    let mut path = vec![
-        mk().path_segment(std_or_core(no_std)),
-        mk().path_segment("mem"),
-    ];
+    let mut path = vec![mk().path_segment("core"), mk().path_segment("mem")];
 
     if type_args.is_empty() {
         path.push(mk().path_segment("transmute"));
@@ -1239,7 +1217,7 @@ impl<'c> Translation<'c> {
         main_file: &path::Path,
     ) -> Self {
         let comment_context = CommentContext::new(&mut ast_context);
-        let mut type_converter = TypeConverter::new(tcfg.emit_no_std);
+        let mut type_converter = TypeConverter::new();
 
         if tcfg.translate_valist {
             type_converter.translate_valist = true
@@ -1394,7 +1372,7 @@ impl<'c> Translation<'c> {
                     use CastKind::*;
                     match cast_kind {
                         IntegralToPointer | FunctionToPointerDecay | PointerToIntegral => {
-                            return true
+                            return true;
                         }
                         _ => {}
                     }
@@ -1805,7 +1783,7 @@ impl<'c> Translation<'c> {
                         _ => {
                             return Err(TranslationError::generic(
                                 "Found non-field in record field list",
-                            ))
+                            ));
                         }
                     }
                 }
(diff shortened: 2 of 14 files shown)


c2rust-transpile/src/convert_type.rs
```diff
@@ -134,14 +132,14 @@ pub const RESERVED_NAMES: [&str; 103] = [
...
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> TypeConverter {
         TypeConverter {
             translate_valist: false,
             renamer: Renamer::new(&RESERVED_NAMES),
             fields: HashMap::new(),
             suffix_names: HashMap::new(),
             features: HashSet::new(),
-            emit_no_std,
         }
```
kkysen, line 143: Don't suppress the warning; follow it.  We should `impl Default for TypeConverter`.  It might need an `impl Default for Renamer` that calls `Self::new(&RESERVED_NAMES)`.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1394,7 +1372,7 @@ impl<'c> Translation<'c> {
                     use CastKind::*;
                     match cast_kind {
                         IntegralToPointer | FunctionToPointerDecay | PointerToIntegral => {
-                            return true
+                            return true;
```
kkysen, line 1375: Are all the `;`s from `rustfmt` with default settings?  I thought I had run that on everything already.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3211,12 +3189,7 @@ impl<'c> Translation<'c> {
 
         let ty = self.convert_type(type_id)?;
         let tys = vec![ty];
-        let mut path = vec![];
-        if self.tcfg.emit_no_std {
-            path.push(mk().path_segment("core"));
-        } else {
-            path.push(mk().path_segment("std"));
-        }
+        let mut path = vec![mk().path_segment("core")];
```
kkysen, line 3192: Thanks for catching this other std or core logic that I missed!

pdg/src/util.rs
```diff
@@ -113,15 +113,6 @@ impl<T: Display> Display for Duplicates<T> {
...
 
-impl<T: Debug + Eq + Hash> Duplicates<T> {
-    pub fn assert_empty_debug(&self) {
-        if self.is_empty() {
-            return;
-        }
-        panic!("unexpected duplicates: {:?}", self);
-    }
-}
-
```
kkysen, line 124: Don't delete this.  It's sometimes used, and it has nothing to do with this PR.

kkysen (changes requested):
Thanks for more help, @afetisov!  This is something I had been meaning to ask about and do, as I agree it simplifies things a lot while still being fully correct in all cases, and I don't see many substantial drawbacks at all (if you do want `std` types used, you can just search and replace `core::` with `std::`).  What do others think?  Is there any advantage that @afetisov 
 and I am missing of having the option to emit `std` types vs. `core` ones and defaulting to it?

A couple of things, though:
* There are a lot of other places where we currently hardcode `std` (https://github.com/immunant/c2rust/issues/480) and those should be updated to use `std_or_core` (without this PR) or `"core"` (with this PR).  Maybe these should be added to this PR, or maybe we can keep them separate.
* `--emit-no-std` should be removed as an option as we'll no longer support it (it'll be forced on, basically)
* Hardcoding strings like `"std"`, `"core"`, `"instrinsincs"`, etc. isn't very good either.  We don't have the best test coverage so it's not the best just to leave until then to catch bugs.  I think a good option could be to have a macro like `path!(core::ptr::null_mut)` that expands to something like this:

  ```rust
  {
      use core::ptr:null_mut;
      ["core", "ptr", "null_mut"]
  }

Also, I left some other inline comments.

pdg/src/util.rs
afetisov: It issues a dead code warning. If it is occasionally used, I'll add an allow with a comment.

afetisov:
> There are a lot of other places where we currently hardcode std (https://github.com/immunant/c2rust/issues/480) and those should be updated to use std_or_core (without this PR) or "core" (with this PR).

Most of those places cannot be done with `#![no_std]`, because stuff like `ffi` or `Vec` is not available in `#![no_std]`. But indeed, there are a couple of places where there should be "core".

> --emit-no-std should be removed as an option as we'll no longer support it (it'll be forced on, basically)

Thought about it. Currently this option adds `#![no_std]` at the top level. I think it should stay that way. It is unreasonable to try translating everything to `#![no_std]` by default, and misleading having that attribute when the crate implicitly uses POSIX calls (which it does, during ordinary translation).

In the future, that option should do something more complicated, like erroring out if the translation can not be done for freestanding systems, or possibly using some polyfill logic.

> Hardcoding strings like "std", "core", "instrinsincs", etc. isn't very good either. We don't have the best test coverage so it's not the best just to leave until then to catch bugs. I think a good option could be to have a macro like path!(core::ptr::null_mut) that expands to something like this:
>
>    {
>        use core::ptr:null_mut;
>        ["core", "ptr", "null_mut"]
>    }

It is also hardcoding, just in a more pretty way. I agree that the API should be prettier, but that doesn't solve the issue with testing, and should be fixed in a different PR anyway. I'm currently doing a cleaning pass over the builder API, perhaps I'll add something like that.

I have thought about adding separate methods for stuff like `core::ptr::null_mut`, but I have decided it is useless. Most functions are used at a single place. The closer that call looks like ordinary inline Rust call, the less the possibility of errors.

Also, a mistype in a module name is pretty much the least of your worries if you have bad test coverage, so that is also something to prioritize. Currently the tests are quite burdensome to run, they have a lot of implicit dependencies, and different test cases are dumped in the same test fixtures. That's also something to be solved.

c2rust-transpile/src/convert_type.rs
kkysen: Even with `emit_no_std` as a field, it still could have been `#[derive(Default)]`ed if there is an

```rust
impl Default for Renamer {
    fn default() -> Self {
        Self::new(&RESERVED_NAMES);
    }
}
```
which seems like a good Rust default (and the only time it's used outside of tests).  Then `new` could be (before this PR):

```rust
pub fn new(emit_no_std: bool) -> Self {
    Self {
        emit_no_std,
        ..Default::default()
    }
}
```

and (after this PR):

```rust
pub fn new() -> Self {
    Self::default()
}
```

`TypeConverter` seems like a type that will always have defaults, and so I think it should `impl Default`, regardless of if `new` takes arguments or not.  That said, the point about future compatibility and re-adding parameters is a good point, though I think any future parameters/fields would have a default, like `emit_no_std` did.  And if we do go this route, you should include that above explanation in a comment above `#[allow(clippy::new_without_default)]`.

afetisov: Ok, I've added an explanation.

kkysen:
> Most of those places cannot be done with `#![no_std]`, because stuff like `ffi` or `Vec` is not available in `#![no_std]`. But indeed, there are a couple of places where there should be "core".

I looked through all of the current uses of `"std"`.  It appears most can be moved to `core` or `alloc`, or can be soon:

| `std` | `core`/`alloc` | Notes |
| ----- | -------------- | ----- |
| `std::mem::transmute` | `core::mem::transmute` | |
| `std::ptr::write_bytes` | `core::ptr_write_bytes` | |
| `std::ptr::null_mut` | `core::ptr::null_mut` | |
| `std::vec::from_elem` | `alloc::vec::from_elem` | apparently it's `#[doc(hidden)]` |
| `std::ffi::CStr` | `core::ffi::CStr` | `#![feature(core_c_str)]`, being stabilized in `1.63.0` |
| `std::ffi::CString` | `alloc::ffi::CString` | `#![feature(alloc_c_string)]`, being stabilized in `1.63.0` |
| `std::process::exit` | | need `std` |
| `std::env::args` | | need `std` |
| `std::env::vars` | | need `std` |

> Thought about it. Currently this option adds #![no_std] at the top level. I think it should stay that way. It is unreasonable to try translating everything to #![no_std] by default, and misleading having that attribute when the crate implicitly uses POSIX calls (which it does, during ordinary translation).
> 
> In the future, that option should do something more complicated, like erroring out if the translation can not be done for freestanding systems, or possibly using some polyfill logic.

I hadn't realized `--emit-no-std` also emits `#![no_std]`.  In that case, we should definitely keep it.

> It is also hardcoding, just in a more pretty way. I agree that the API should be prettier, but that doesn't solve the issue with testing, and should be fixed in a different PR anyway. I'm currently doing a cleaning pass over the builder API, perhaps I'll add something like that.

With a macro like this, `path!(core::ptr::null_mut)`:

```rust
macro_rules! path {
    ($path:path) => {
        {
            #[allow(unused_imports)]
            use $path;

            stringify!($path).split("::")
        }
    };
}
```

we do get compiler-checked paths as well as IDE goto-able paths, which I think are quite useful for such a simple macro.

pdg/src/util.rs
kkysen: We haven't fixed all the warnings yet, so can you just leave this as is?  This is in fast-changing PDG code.

c2rust-transpile/src/convert_type.rs
kkysen: I still think it should be `#[derive(Default)]`ed.

pdg/src/util.rs
```diff
@@ -113,6 +113,17 @@ impl<T: Display> Display for Duplicates<T> {
     }
 }
 
+impl<T: Debug + Eq + Hash> Duplicates<T> {
+    // This function is sometimes used during debugging of PRs.
```
kkysen, line 117: It's not for PRs.

afetisov:
Cool, I didn't know `CStr` and `CString` were subject to be moved. In that case I'll update their code.

afetisov:
> Cool, I didn't know CStr and CString were subject to be moved. In that case I'll update their code.

On second thought, I don't want to do it now. CStr isn't used in transpiler, and CString is used only when collecting arguments into a `Vec` in the transpiled `main` function. It depends on `std` anyway, and we can update the code once the feature is stabilized.

kkysen:
They're being moved to `core`/`alloc` along with the `c_*` types.

You can leave `CStr` and `CString` for now as they're still behind feature flags for now.  Maybe just add a comment saying we can update them soon once they're stabilized.

afetisov:
> With a macro like this, path!(core::ptr::null_mut) ...

I'm going to add such a macro. In fact, I plan to make it more powerful, so it will be able to directly produce Path, or Type and Expr which are just a path (i.e. wrap `path_ty` and `path_expr`). There is a lot of path building in the code. A macro like this can easily give 90% of simplifications compared to `syn::parse_quote!`, without its complexity.

kkysen:
Yeah, that would be perfect, thanks!

afetisov:
Are there any other issues with this PR? I'd like to build the other PRs on top of it.

kkysen:
I still think `TypeConverter` should be `#[derive(Default)]`ed.  And I'd like to get someone else's input on this, too, probably tomorrow as today is a holiday.

fw-immunant (approved):
I think this is a good change in general--more things have moved to libcore than once were there, and the toggle isn't really very useful as a result.

I don't think it's necessary to reconsider the API of `TypeConverter` as of this change. In general I think `Default` is useful for types created often, especially deep inside the construction of other, highly-nested structures, but we literally only call `TypeConverter::new` once in c2rust.

kkysen (approved):

Issue #489, opened 2022-07-07
`c2rust_analysis_rt` crashes when instrumenting static initializers


kkysen:
Just leaving this as a tracking issue.

Currently, `c2rust_analysis_rt` often (randomly) crashes when instrumenting static initializers, as they run before `main`, but they use Rust's/`pthread`s threads, which haven't been properly initialized yet.  This has been manually fixed in [`lighttpd-rust`](https://github.com/immunant/lighttpd-rust) by @rinon by moving the static initializers (which are run through `.init_array` I believe) to the top of the Rust `main`.

But a more general fix should be done, and it should be fairly easy.  We can simply drop `Event`s before `main`, then initialize the runtime at the top of `main`, which is the first use of threads.

PR #490, opened 2022-07-07
De-globalize `c2rust_analysis_rt`'s runtime and `Metadata`

kkysen:
This is a significant refactor of `analysis/runtime` and `pdg` to de-globalize as much as possible, namely the runtime and the `Metadata`, as well as other opportunities doing so unlocks.

## `Metadata`

More specifically, for `Metadata`, `mir_loc::set_file` and the `lazy_static!`s `MIR_LOC_FILE_PATH` and `MIR_LOCS` are removed.  Now there is no more global `Metadata`, and `fn get(index: MirLocId) -> Option<&'static MirLoc>` is now `fn Metadata::get(&self, index: MirLocId) -> &MirLoc`.  Note that the `Option` is gone now as the `&self` `Metadata` must be there now.

However, `mir_loc::get` and some other things using the global `MIR_LOCS` were using in `Debug` and `Display` `impl`s.  To still keep these, I added

```rust
pub struct WithMetadata<'a, T> {
    pub inner: &'a T,
    pub metadata: &'a Metadata,
}

pub trait IWithMetadata where Self: Sized {
    fn with_metadata<'a>(&'a self, metadata: &'a Metadata) -> WithMetadata<'a, Self> {
        WithMetadata {
            inner: self,
            metadata,
        }
    }
}
```

Then for every `T` that we previously had a `Debug` or `Display` `impl` for, we can now have an `impl IWithMetadata for T {}` and `impl {Debug, Display} for WithMetadata<'_, T>`, and then with a `t.with_metadata(metadata)`, we can get back the `Debug` or `Display` `impl`, and we can thread this `Metadata` through everything to re-create `Debug` and `Display` `impl`s for everything that depends on a `Metadata` for that.

There are a couple of things I had to get around with in using this API.  I previously had a blanket `impl` for `IWithMetadata`, but due to the orphan rule, `impl Debug for WithMetadata<'_, T>` couldn't be done in `c2rust_pdg` as `WithMetadata` is defined in `c2rust_analysis_rt`.  Thus, I added the `decl_with_metadata!()` macro, which just declares the above `WithMetadata` and `IWithMetadata` in your own crate.  This prevents me from adding a blanket `impl`, because then the blanket `impl`s conflict, and there are ambiguities (which can be disambiguated, but it's very ugly, and the manual default `impl`s are easier).

## The Runtime

The runtime/backend in `analysis/runtime/src/backend.rs` is now heavily refactored into `analysis/runtime/src/runtime/`.  It is de-globalized as much as possible such that the runtime is not initialized implicitly, only by `c2rust_analysis_rt::initialize`, which `c2rust-instrument` inserts a call to at the top of `main`.  Thus, the runtime, which is not needed by `c2rust-pdg`, can be avoided, and thus the dependencies the runtime brings, such as its initialization code and need for `$METADATA_FILE`, can be dropped.  Furthermore, the runtime now drops all `Event`s before `c2rust_analysis_rt::initialize` and after `c2rust_analysis_rt::finalize`, which are called at the start and end of `main`.  Thus, `Event`s are only processed during the lifetime of `main`, which is where Rust code is safe to execute.  This includes spawning the backend thread, which requires the initialization of Rust's thread library, usually `libpthread`, which is not ready before `main`, and led to crashes (https://github.com/immunant/c2rust/issues/489).

For an overview of how this is implemented, first the `FINISHED` `(Mutex<bool>, Condvar)` is kept as a global (though using `once_cell::sync::Lazy` instead of `lazy_static!`, and once `1.63` lands, neither) so that it can have a `'static` lifetime so it can be used safely across non-scoped threads like the backend thread.  Then the `TX: SyncSender<Event>` is encapsulated in `struct Runtime` and `Runtime::tx`, which manages its initialization and finalization.  It `impl`s `Drop`, though this is mostly for naught, as `static`s don't get their destructors run.  In `Runtime::try_init`, the backend thread and channel are spawned, as well as the backend detected with `Backend::detect`, which can either be `Debug` or `Log`.

Then there is `struct GlobalRuntime`, which contains a `OnceCell<Runtime>`.  There is one global for this, `static RUNTIME: GlobalRuntime`.  When `const`-initialized, it's an empty `OnceCell`.  Only when `c2rust_analysis_rt::initialize` is called does it get initialized, which calls `Runtime::try_init`.  Before that, `GlobalRuntime::send_event` just drops `Event`s.  Furthermore, `Runtime` also contains a `OnceCell<()>` that controls the finalizer so that it's only run once.  After it's run from `c2rust_analysis_rt::finalize`, `Event`s are dropped again.  Thus, the `Runtime`, the `Backend`, and its thread are only used during the lifetime of `main`, in which is safe to run non-`const` Rust code.

## Other Improvements

Improvements that this unlocked:
* `c2rust-pdg` no longer needs to run the runtime and there is no global `Metadata`.  Thus, `$METADATA_FILE` is no longer necessary and can be passed as a normal argument along with the event log path.
* Code is free to be parallelized now, as there is no global data lurking around anymore, except for the `RUNTIME` in instrumented code, which lives exactly as long as `main`.  This includes parallelization that `cargo test` does by default.
* It is much more easy to construct multiple PDGs sequentially as well, which could maybe/kind of be done before, but was very unwieldy with all the globals and `lazy_static!`s.
* Any `Event`s before or after `main` are dropped, instead of causing weird crashes with threads, as `libpthread` was not fully initialized before when instrumenting code in static initializers in `.init_array` (`.fini_array` would also be a problem, I believe).  Thus, this fixed https://github.com/immunant/c2rust/issues/489, which I tested (no more random crashes, yay!).
* `lazy_static!` is replaced with `once_cell:sync::{OnceCell, Lazy}`.  They are generally preferred over the `lazy_static!` macro, and are on track to be included in `libstd`.
* Many unnecessary locks are removed, especially around the global `MIR_LOC_FILE_PATH` and `MIR_LOCS`.
* Path arguments now use `PathBuf`/`Path`/`OsString`/`env::args_os`/`env::var_os` instead of `String`/`env::args`/`env::var`, allowing non-UTF-8 paths that are otherwise valid.

Some other smaller improvements that tagged along:
* The runtime code is much more well documented now.

Once this lands, I can much more easily set up the snapshot testing with `insta`.

Also waiting for this PR to land:
* https://github.com/immunant/c2rust/pull/492
* https://github.com/immunant/c2rust/pull/493
* https://github.com/immunant/c2rust/pull/494


aneksteind:
Ok, I'll review this in more depth and assume that the more-trivial but unrelated changes such as `clap` will go into their own PR. I acknowledge your point re: the decoupling of Metadata and Runtime changes, so that can stay. Would also like to have @rinon's review before this is merged.

kkysen:
I just moved https://github.com/immunant/c2rust/pull/491 with https://github.com/immunant/c2rust/pull/490/commits/d564787419699949987935ee1c8f1579c962f534 into a separate PR that should be merged before this one.

Now I'm splitting the `clap` additions into a separate PR.

For the `String` to `OsString` changes, do those have to be put in a separate PR?  They're very minor changes, just adding a few `*_os`s, but they're not separated into different commits, so it's harder to untangle them without making a mess of all of the branches branched off of [`kkysen/de-globalize-runtime-and-metadata`](https://github.com/immunant/c2rust/tree/kkysen/de-globalize-runtime-and-metadata).

analysis/runtime/src/metadata.rs
```diff
@@ -0,0 +1,109 @@
...
+    F: Copy + FnOnce(&mut Formatter) -> fmt::Result,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.0(f)
+    }
+}
+
+impl Debug for Metadata {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        // f.debug_map().entries(entries)
```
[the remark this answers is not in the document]
kkysen: Yeah, thanks, I missed that.

analysis/runtime/src/runtime/backend.rs
```diff
@@ -0,0 +1,94 @@
...
+        let (lock, cvar) = &*FINISHED;
+        let mut finished = lock.lock().unwrap();
+        self.write_all(rx);
+        *finished = true;
+        cvar.notify_one();
+    }
+}
+
+impl DebugBackend {
+    pub fn detect() -> Result<Self, AnyError> {
```
aneksteind, line 64: why isn't this part of the `IBackend` interface?

kkysen: I originally tried to put it in `IBackend`, too, but that didn't work.

analysis/runtime/src/runtime/global_runtime.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// perhaps in an `.init_array` or `.ctors` section,
+    /// and we cannot use threads yet as the Rust runtime is not fully initialized yet.
+    /// Therefore, we silently drop the [`Event`].
+    /// 
+    /// It also silently drops the [`Event`] if the [`Runtime`]
+    /// has been [`Runtime::finalize`]d/[`GlobalRuntime::finalize`]d.
+    pub fn send_event(&self, event: Event) {
+        match self.runtime.get() {
+            None => {
+                // Silently drop the [`Event`] as the [`Runtime`] isn't ready/initialized yet.
```
aneksteind, line 40: could we add some debug logs here to get an account of what isn't getting captured, and do you think that's useful?

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -58,9 +60,8 @@ impl InstrumentMemoryOps {
     pub fn finalize(&self, metadata_file_path: &Path) -> anyhow::Result<()> {
         let mut locs = self.mir_locs.lock().unwrap();
         let mut functions = self.functions.lock().unwrap();
-        let locs: Vec<MirLoc> = locs.drain(..).collect();
-        let functions: HashMap<c2rust_analysis_rt::DefPathHash, String> =
-            functions.drain().collect();
+        let locs = locs.drain(..).collect::<Vec<_>>();
```
aneksteind, line 63: is this a convention that's preferred?

kkysen: Specifying the generic directly in the call instead of as an annotation also has the added benefit that the expression is now self contained and can be freely moved or used in a larger expression without modification.  You don't get that with the type annotation.

kkysen:
Just rebased against `master` after https://github.com/immunant/c2rust/pull/491 merged so I can remove that duplicate commit.

analysis/runtime/src/runtime/global_runtime.rs
kkysen: With the old `lighttpd` that still had true pre-`main` static initializers, we skip 1801 `Event`s before `main`, and nothing after `main`.

aneksteind: can you collect them in a structure that you only output once main has started?

kkysen: Perhaps what would be safer and still useful would be to simply count the `Event`s and once `main` starts, print how many were skipped if any were.  That way you still get notified it happens and at all, don't drown in tons of `Event`s, and stay safe before `main`.

Alternatively, we could just write the pre-`main` `Event`s on the main thread, perhaps unbuffered, as `write(2)` should always be safe.  Allocation could potentially be unsafe if the allocator uses any `pthread` locks internally.  Furthermore, `std::env::{var,args}{,_os}` I believe are very unsafe pre-`main`, as Rust's runtime that it sets up before `main` explicitly initializes these.

Overall, I think it's best to do the bare minimum before `main` and it should be up to the user to move their static initializers into `main` if they want them analyzed.  I think the most we should do is notify the user that this has happened so they can fix it.

analysis/runtime/src/runtime/scoped_runtime.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// panicking if there is a [`SendError`](std::sync::mpsc::SendError).
+    pub fn send_event(&self, event: Event) {
+        match self.finalized.get() {
+            None => {
+                // `.unwrap()` as we're in no place to handle an error here,
+                // unless we should silently drop the [`Event`] instead.
+                self.tx.send(event).unwrap();
+            }
+            Some(()) => {
+                // Silently drop the [`Event`] as the [`Runtime`] has already been [`Runtime::finalize`]d.
```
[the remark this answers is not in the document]
kkysen: Noted; moving the conversation to that question.

pdg/src/main.rs
```diff
@@ -24,34 +24,43 @@ mod query;
...
-    let pdg = construct_pdg(&events);
+    let pdg = construct_pdg(&events, &metadata);
     pdg.assert_all_tests();
 
     for (graph_id, graph) in pdg.graphs.iter_enumerated() {
         let needs_write = graph
             .needs_write_permission()
             .map(|node_id| node_id.as_usize())
             .collect::<Vec<_>>();
+        let graph = graph.with_metadata(&metadata);
```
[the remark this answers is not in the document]
kkysen: That is true I think.  If we replace

```rust
pub struct Func(pub DefPathHash);
```

with

```rust
pub struct Func {
    pub def_path_hash: DefPathHash,
    pub name: String,
}
```

which we create with the `Metadata` while constructing the PDG, then I think we'll be able to get rid of the `WithMetadata` stuff for the PDG code.  We still need it in some places, but not directly in `c2rust-pdg`.

Should I make that change?  Thanks for pointing this out!

analysis/runtime/src/runtime/global_runtime.rs
kkysen: > can you collect them in a structure that you only output once main has started?

Oh, sorry, I didn't see this comment until now.

That's not the best option either for a few reasons:
* We'd have to allocate, which likely is but may not be safe.
* We'd need to collect the skipped `Event`s in a `static`, which we'd have to protect with a `Mutex` to make it thread-safe, but we can't use `Mutex`es before `main` on all OSes and Rust versions as that uses `libpthread`.
* Since this code is pre-`main`, we may be able to use `unsafe` and a `static mut` as there should be no other threads.  That still feels pretty dangerous, though.

rinon: I'm fine with the collected count and debug print. I'm not sure how useful 1800 lines of debug output really would be, and it's not that hard to go remove constructors if someone really wants to capture all these events (like I did for lighttpd).

kkysen: Yeah, that's what I was thinking.  We really just need to notify the user that this is happening so they can fix it and do things correctly.  I'll add this commit then.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/490/commits/1b72fd4ee2c716ab8f3ab9634bca778df96fb4ca.

pdg/src/main.rs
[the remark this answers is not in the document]
kkysen: Fixed in https://github.com/immunant/c2rust/pull/490/commits/eff8ff9a47459803f64c3157629602d38c67a934.  Does that look better?  I think it's a lot cleaner now.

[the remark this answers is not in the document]
aneksteind: It's a great start; it could be taken a step further where you add the function name to the `MirLoc` during instrumentation, or just before it ends, and use that in creating `Node.function` -- that would preclude the need for `WithMetadata` etc. altogether.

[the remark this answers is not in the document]
kkysen: Ultimately I do think we still need it because we don't want to put extra data in `Event` and we want to have a `Debug` `impl` for it, which needs the `Metadata`.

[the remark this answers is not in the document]
kkysen: That's basically the same logic though.  I don't think getting rid of WithMetadata is worth that much, as then we also lose the debugging convenience of `Debug` `impl`s.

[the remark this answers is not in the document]
aneksteind: Something like #496 is what I'd be willing to accept. It removes code and has the same logic.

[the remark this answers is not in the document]
kkysen: https://github.com/immunant/c2rust/pull/496 is merged now.

aneksteind (approved):
LGTM, would like to also get a look from @fw-immunant or @rinon

analysis/runtime/src/mir_loc.rs
```diff
@@ -47,39 +14,34 @@ pub enum MirProjection {
 /// See [`rustc_middle::mir::Local`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Local.html).
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub struct Local {
-    /// TODO(kkysen) change to u32 like
     /// [`rustc_middle::mir::Local`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Local.html),
     /// but need to keep bincode binary format.
```
fw-immunant, line 18: This comment doesn't make a lot of sense without its first line--remove entirely/redocument?

analysis/runtime/src/runtime/backend.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::sync::mpsc::Receiver;
+
+use bincode;
+
+use super::{AnyError, FINISHED};
+use crate::events::{Event, EventKind};
+use crate::metadata::Metadata;
+
+#[enum_dispatch]
+pub(super) trait IBackend {
```
fw-immunant, line 14: nit: can this just be `Backend`? All traits are interfaces.

analysis/runtime/src/runtime/backend.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl IBackend for LogBackend {
+    fn write(&mut self, event: Event) {
+        bincode::serialize_into(&mut self.writer, &event).unwrap();
+    }
+}
+
+#[enum_dispatch(IBackend)]
+pub enum Backend {
```
fw-immunant, line 40: This would then be `BackendKind` or such.

fw-immunant (approved):
This generally looks good. I had a few naming nits, but they can be addressed after merging if we care. It is kind of tricky to name things sanely and without conflicts with the code patterns being used for the `runtime::Backend`s.

analysis/runtime/src/runtime/backend.rs
kkysen: `BackendKind` doesn't make sense since the `enum` has data.  If it didn't and were just the kind, then it'd make sense.  Also then `{Log,Debug}Backend` should be `Kind`s, too, but that makes even less sense.

analysis/runtime/src/runtime/backend.rs
kkysen: Yeah, I had tried to make it `Backend`, too, but `trait`s and `struct`/`enum`s can't overlap.  I thought about making the `trait` `Backend` and the `enum` `BackendImpl`, but then the `{Log,Debug}Backend`s should be `Impl`s, too, which I thought made less sense than naming this `IBackend`.

analysis/runtime/src/runtime/backend.rs
fw-immunant: Yeah, I realized this right after commenting, but the `IBackend` name is still not ideal. Maybe `WriteEvent` would be a better name for the trait, as that's the only method exposed.

kkysen: Yeah, that's a better idea.  Should it be `WriteEvent` or `EventWriter`?

fw-immunant: The former--the more rusty convention is imperative, like `Read`/`Write`/`Debug`/`Serialize`.

kkysen: Ah yes, okay, I'll rename that and then merge.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/490/commits/e3098a14131f0acd39db71a5dbc3d0709fcb694f.

analysis/runtime/src/runtime/backend.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/490/commits/e3098a14131f0acd39db71a5dbc3d0709fcb694f.

PR #493, opened 2022-07-07
Add a `c2rust-pdg --print TO_PRINT` option

kkysen:
Added a `c2rust-pdg --print TO_PRINT` option that can be any of
* `graphs`
* `counts`
* `events`
* `latest-assignments`
* `write-permissions`

and can be repeated to print more of them.  This
* brings some control to all the ad-hoc things that were printed before
* makes sure we don't have code that's commented out that has to be maintained
* lets us easily change what we want to be printed without re-compiling, as @fw-immunant came across once

### DIFF
--- a/pdg/src/main.rs
+++ b/pdg/src/main.rs
@@ -71,14 +71,14 @@ fn main() -> eyre::Result<()> {
     env_logger::init();
     let args = Args::parse();
 
-    let print = args.print.iter().collect::<HashSet<_>>();
-    let print = |to_print| print.contains(&to_print);
+    let print_args = args.print.iter().collect::<HashSet<_>>();
+    let should_print = |to_print| print_args.contains(&to_print);
 
     let events = read_event_log(Path::new(&args.event_log))?;
     let metadata = read_metadata(Path::new(&args.metadata))?;
     let metadata = &metadata;
 
-    if print(ToPrint::Events) {
+    if should_print(ToPrint::Events) {
         for event in &events {
             let mir_loc = metadata.get(event.mir_loc);
             let kind = &event.kind;
@@ -89,7 +89,7 @@ fn main() -> eyre::Result<()> {
     let pdg = construct_pdg(&events, metadata);
     pdg.assert_all_tests();
 
-    if print(ToPrint::LatestAssignments) {
+    if should_print(ToPrint::LatestAssignments) {
         for ((func_hash, local), p) in &pdg.latest_assignment {
             let func = &metadata.functions[func_hash];
             println!("({func}:{local:?}) => {p:?}");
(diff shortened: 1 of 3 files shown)


pdg/src/main.rs
```diff
@@ -24,62 +24,111 @@ mod query;
...
+    print: Vec<ToPrint>,
 }
 
 fn main() -> eyre::Result<()> {
     color_eyre::install()?;
     env_logger::init();
     let args = Args::parse();
 
+    let print = args.print.iter().collect::<HashSet<_>>();
+    let print = |to_print| print.contains(&to_print);
```
aneksteind, line 76: I'm curious how the shadowing and closing over `print` works here, does it create a local copy in the lambda?

fw-immunant: it references and shadows the existing `print` (it would move if it were a `move` closure). That said, I think `print_args` and `should_print` would be more descriptive names.

kkysen: Okay, I'll change the variable names; that's clearer.

pdg/src/main.rs
```diff
@@ -24,62 +24,111 @@ mod query;
...
     #[clap(long, value_parser)]
     event_log: PathBuf,
+
     /// Path to the instrumented program's metadata generated at compile/instrumentation time.
     #[clap(long, value_parser)]
     metadata: PathBuf,
+
+    /// What to print.
+    #[clap(long, value_parser)]
+    print: Vec<ToPrint>,
```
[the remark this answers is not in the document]
kkysen: I had initially tried to do that because that is generally the default that we want, but I'm not sure how to do it with `clap` derive (`default_value_t = vec![ToPrint::Graphs]` doesn't work because `Vec` doesn't `impl Display`).  That said, thinking about it, if `--print graphs` is the default, how would you disable it?  I couldn't think of a good way to do that, so I think just specifying `--print graphs` is okay.  The `pdg.sh` script does it by default anyways.

pdg/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/493/commits/1d5baf6cbe1a7730bcfdbb03434d02e097ccfcce.

pdg/src/main.rs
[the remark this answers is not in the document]
kkysen: Fixed in https://github.com/immunant/c2rust/pull/493/commits/a4a5e5206d704af59bc7ee80d124600556b47272.

aneksteind:
LGTM after the concerns with #490 are addressed and reflected here.

kkysen:
@aneksteind, it's now merged with https://github.com/immunant/c2rust/pull/490https://github.com/immunant/c2rust/pull/490, so `WithMetadata` is removed now.

kkysen:
Merging brought in tons of other commits to this PR so I rebased instead.  Should be good to go now.

PR #494, opened 2022-07-07
Append to event log

kkysen:
Append to the event log in `c2rust_analysis_rt::runtime::backend::LogBackend` by opening the event log in append mode.  This lets us collect events from multiple, sequential runs of the same program.  This can be quite useful, as often tests, like for `lighttpd`, run the instrumented binary multiple times.  Note that parallel runs of the same program would not work at all.  For that case, each run should write to a different event log, which we can concatenate together at the end.

The event log, since it's split into separate `Graph`s for each object anyways, doesn't really matter if it represents multiple runs or not.  However, this PR also makes sure the `EventKind::Done` `Event` is written at the end, so that if we do want to split the event logs from different runs back apart, we can easily do so.

Now that we append, we need to delete the event log before running the `pdg.sh` script.

### DIFF
--- a/analysis/runtime/src/runtime/backend.rs
+++ b/analysis/runtime/src/runtime/backend.rs
@@ -77,7 +77,20 @@ impl LogBackend {
     pub fn detect() -> Result<Self, AnyError> {
         let path = env::var_os("INSTRUMENT_OUTPUT")
             .ok_or("Instrumentation requires the INSTRUMENT_OUTPUT environment variable be set")?;
-        let file = OpenOptions::new().create(true).append(true).open(&path)?;
+        let append = env::var("INSTRUMENT_OUTPUT_APPEND").ok().ok_or(
+            "Instrumentation requires the INSTRUMENT_OUTPUT_APPEND environment variable be set",
+        )?;
+        let append = match append.as_str() {
+            "true" => true,
+            "false" => false,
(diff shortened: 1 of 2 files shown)


aneksteind:
Unless I'm missing something, I think this makes it too easy for a program that has been changed or instrumented differently to append to the event log from a prior run, before those changes. Could we add something to this PR to prevent against that?

kkysen:
Yeah, that might be a good idea.  I don't think it's really possible to (efficiently) detect if the instrumented program has changed, but we could add an environment variable to control this that does not append by default, something like `$INSTRUMENT_OUTPUT_APPEND`.  How does that sound?

aneksteind:
That would be fine, and if the file already exists maybe we can prompt the user with a warning/option to overwrite/append

kkysen:
> if the file already exists maybe we can prompt the user with a warning/option to overwrite/append

I think that's too complex to implement inside of another program we're instrumenting, but I can require `$INSTRUMENT_OUTPUT_APPEND` to be set to either `true` or `false` instead of having a default so that it's always explicit.  Also, we can't have different behavior if the file exists because of TOCTOU race conditions.

aneksteind:
Okay, sounds good to me

aneksteind (approved):

PR #496, opened 2022-07-08
remove `WithMetadata`, etc.

aneksteind:


kkysen (commented):
Can we instead/on top of this simply add the function name to `DefPathHash` and remove the `Metadata::functions: HashMap<DefPathHash, String>` field entirely?  Is there any reason to store that data separately?

aneksteind:
As far as getting rid of `functions` goes, there are functions that we want to know the name for and don't instrument (callees from external libraries for example), but `MirLoc` is only generated for instrumented functions. So they need to be separate for at least that reason.

> Also, I think if impl Debug for Event just uses the impl Debug for EventKind and doesn't print the whole Event, we should just skip that impl Debug instead of pretending we're able to print it.

Feel free to commit changes to this to simplify.

kkysen:
It appears `Metadata::functions` is never actually used.  I assume it may be in the future, right?  Could I change it a `Vec<DefPathHash>` with the function name inside of `DefPathHash`, or do we need the O(1) lookup (I don't know since it's not currently used at all)?

aneksteind:
> It appears Metadata::functions is never actually used. I assume it may be in the future, right? Could I change it a Vec<DefPathHash> with the function name inside of DefPathHash, or do we need the O(1) lookup (I don't know since it's not currently used at all)?

I think that's my mistake, give me a moment and I'll push up a fix

kkysen:
I was gonna either rename what we currently have as `Fingerprint` because that's what rustc has it as a newtype of, or I could name the struct with the function name `DefPath`.

aneksteind:
I don't think we should copy the name `Fingerprint` or `DefPath` etc. if the underlying structure is not mirrored to the MIR one. To clarify what I mean, `Fingerprint` is basically a tuple, and if we add a function name it's not really a tuple anymore, but tuple + something else. We're already diverging from that in the current implementation and so not reinforcing that seems best.

kkysen:
Our definition now is the same as `Fingerprint`, not `DefPathHash`, and I thought `DefPath` would make sense since it has more of the path (the function name) instead of only the hash.

aneksteind:
Something like `FuncInfo` might be better, but I'll know more of what you had in mind after the commit. What I think is most important is getting this overall change into #490 so that snapshot testing can continue.

kkysen:
I took the `FuncInfo` approach, but that's the same as the `Func` in `c2rust_pdg`, so I just moved that `Func` into `c2rust_analysis_rt::mir_loc`.  I also cleaned up a bunch of extra/verbose conversions from `rustc` to `c2rust_analysis_rt` types.  I think this is ready to merge into https://github.com/immunant/c2rust/pull/490 now, and then that into `master`.

aneksteind:
Okay to approve and merge into #490

kkysen (approved):

PR #497, opened 2022-07-09
Fix the `clippy::duplicate_underscore_argument` warning

kkysen:
Fixed the `clippy::duplicate_underscore_argument` warning by renaming the offending argument.

@aneksteind said it was related to the handler functions (to which calls are inserted in `c2rust instrument`) having to have the same number of arguments or something, and that the 3rd argument is usually `ptr`.  I'm not sure exactly, but I just renamed it to `_third_arg_that_is_usually_a_ptr`.


fw-immunant (changes requested):
For functions in `HOOK_FUNCTIONS`, the tracing passes the return value of the traced function as the last argument to the trace hook for it; see the `if after_call` block in `apply_instrumentation` in `instrument_memory.rs`. So this third argument is the unit return value from `free`: perhaps `_free_retval` would be a better name.

kkysen:
Ah, that makes a lot more sense now.  I renamed it to `_free_ret_val`.

fw-immunant (approved):

kkysen:
@fw-immunant, is https://github.com/immunant/c2rust/pull/497/commits/cb2a40a96cb214c74358fd42433ef63d8ed01cc0 okay, or should it go in a separate PR?  Do we want something like this?  I would've found it useful while refactoring before.

kkysen:
I reverted https://github.com/immunant/c2rust/commit/cb2a40a96cb214c74358fd42433ef63d8ed01cc0 and made it just more documented instead.  In doing so, though, I realized that our hook functions only work when C `size_t` translates to `libc::c_ulong` and `libc::c_ulong` is `u64`, which is not the case on 32-bit architectures, for example.  I think we should leave them as `usize`, which is what `libc` does, as that would adjust to the architecture.

fw-immunant:
The only complaint I have here is that it's somewhat confusing that the name of the tracing function (`self::$ident`) gets used as the name of the function to trace (which is read from `HOOK_FUNCTIONS`). These do match exactly, so there's no practical problem, but it is a bit confusing because the dataflow in the program doesn't match the causal flow ("malloc" is a name determined by the C runtime, not by what we decided to name our tracing hook, and similarly, `offset` is so-named because it traces `pointer::offset`). Maybe we could clarify naming somewhere or comment to make it more obvious that the handlers here must be named identically to the functions that they hook.

fw-immunant:
I'm also wondering if our logic in inserting calls to tracing hooks is correct--will we trace a call to a user-defined `fn offset` in any context as if it were `pointer::offset` from libstd?

kkysen:
> I'm also wondering if our logic in inserting calls to tracing hooks is correct--will we trace a call to a user-defined `fn offset` in any context as if it were `pointer::offset` from libstd?

I hope not.  I'm not very familiar with that part of the code, though, so I'm not sure.  We should make sure we're not, though.

We probably should handle the `libc` hooked functions and `offset` somewhat separately, though, as `offset` needs to be the pointer method `offset`, and the `libc` ones need to match a bare, `#[no_mangle] extern "C" fn`.

kkysen:
I merged this PR and opened issues for the (possible) bugs I found rather than trying to fix them here at all, as this is mainly a `clippy` warning fix:
* https://github.com/immunant/c2rust/issues/510
* https://github.com/immunant/c2rust/issues/511

PR #498, opened 2022-07-09
Fixed the `clippy::if_same_then_else` warning

kkysen:
This fixes the `clippy::if_same_then_else` warning here:

https://github.com/immunant/c2rust/blob/2113c3e93bdcb866e975079668e4427dc520f29b/c2rust-transpile/src/translator/mod.rs#L2374-L2388

It would be a trivial fix, except there is a comment in only one of the equivalent branches, so I wasn't sure if they should be merged or not.  No one's said anything about it, so I decided to extract the branch condition with the comment into a separate `bool` variable, move the comment there, and then combine the branches.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2405,9 +2405,7 @@ impl<'c> Translation<'c> {
                 // Only add linkage attributes if the function is `extern`
                 let mut mk_ = if is_main {
                     mk()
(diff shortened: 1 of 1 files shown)


fw-immunant (approved):
This is a bit confusing, historically, but it looks like this duplication was introduced as-is in 0448665c1f67b11fa4e1372f8853ac0eb0d34f71 and isn't the result of convergent evolution, so it should be fine to merge these.

PR #506, opened 2022-07-12
Improve the `c2rust_analysis_rt` env arg parser

kkysen:
This improves the environment variable argument parsing in `c2rust_analysis_rt`.  Ideally, we could use something nice like `clap`, but that's a huge dependency (compile-time and size-wise) to add to the runtime (currently it is recompiled on every `c2rust instrument` and already takes 15 seconds on `donna`).

Instead, I added `mod parse`, which contains a manually re-implemented version of something similar to `clap::ArgEnum`, as well as a few wrappers around parsing arguments from environment variables and printing nice, standardized error messages.

Furthermore, instead of calling `GlobalRuntime::try_init().unwrap()` in `initialize`, which prints the `Debug` `impl` of the `Error` and the `.unwrap()` source location (not very useful, as it's not where the error occurs, and backtraces using something like `eyre` might be too heavy for the runtime), we print the `Display` `impl` of the `Error` only and then `exit(1)`.

### DIFF
--- a/analysis/runtime/src/parse.rs
+++ b/analysis/runtime/src/parse.rs
@@ -1,0 +1,65 @@
+//! Don't want a huge depedency on `clap`.
+
+use std::{
+    error::Error,
+    fmt::{self, Debug, Display, Formatter},
+    marker::PhantomData,
+};
+
+pub trait AsStr {
+    fn as_str(&self) -> &'static str;
+}
+
+pub trait Choices
+where
+    Self: Sized,
+{
+    fn choices() -> &'static [Self];
+}
+
+impl AsStr for bool {
+    fn as_str(&self) -> &'static str {
+        match self {
+            true => "true",
+            false => "false",
+        }
+    }
+}
+
+impl Choices for bool {
+    fn choices() -> &'static [Self] {
+        &[true, false]
+    }
+}
+
+pub struct ChoiceError<T> {
+    pub found: String,
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Choices + AsStr + 'static> Display for ChoiceError<T> {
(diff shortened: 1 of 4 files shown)


analysis/runtime/src/parse.rs
```diff
@@ -0,0 +1,97 @@
+//! Don't want a huge depedency on `clap`.
```
fw-immunant, line 1:
```suggestion
//! Don't want a huge dependency on `clap`.
```

analysis/runtime/src/parse.rs
```diff
@@ -0,0 +1,97 @@
...
+        &[true, false]
+    }
+}
+
+pub fn one_of<T: GetChoices + AsStr + 'static>(s: &OsStr) -> Result<&'static T, String> {
+    T::choices()
+        .iter()
+        .find(|choice| s == OsStr::new(choice.as_str()))
+        .ok_or_else(|| {
+            let s = s.to_string_lossy().into_owned();
```
fw-immunant, line 68: I think deref should work here instead of `into_owned`, and the same for the other instances of this pattern.

kkysen: Ah yes, I didn't realize that works.  It doesn't need ownership, but I didn't know `Cow` just works nicely like that.

fw-immunant (approved):
A couple nits inline. I think the `GetChoices` machinery is a bit baroque, but this is a nice improvement and cleanup.

kkysen:
For `GetChoices`, I think it's pretty similar to what `clap::ArgEnum` does, just more simple/fewer features and manually implemented instead of using a slow derive macro.

PR #512, opened 2022-07-12
make build.rs scripts work without rustup

fw-immunant:
Fixes https://github.com/immunant/c2rust/issues/516.

Rustup is useful to install our dependencies at the versions we need, but we don't need to explicitly depend on it; users may install dependencies other ways, too.

### DIFF
--- a/analysis/build.rs
+++ b/analysis/build.rs
@@ -3,13 +3,17 @@ use std::path::Path;
 use std::process::Command;
 use std::str;
 
-fn main() {
+fn main() -> std::io::Result<()> {
     // Add the toolchain lib/ directory to `-L`.  This fixes the linker error "cannot find
     // -lLLVM-13-rust-1.60.0-nightly".
     let out = Command::new("rustup")
         .args(&["which", "rustc"])
         .output()
-        .unwrap();
+        .or_else(|_| {
+            Command::new("which")
+            .args(&["rustc"])
+            .output()
+        })?;
     assert!(out.status.success());
     let rustc_path = Path::new(str::from_utf8(&out.stdout).unwrap().trim_end());
     let lib_dir = rustc_path.parent().unwrap().parent().unwrap().join("lib");
@@ -22,4 +26,5 @@ fn main() {
     assert!(out.status.success());
     let sysroot = str::from_utf8(&out.stdout).unwrap();
     println!("cargo:rustc-env=RUST_SYSROOT={}", sysroot)
+    Ok(())
 }
--- a/dynamic_instrumentation/build.rs
+++ b/dynamic_instrumentation/build.rs
@@ -3,13 +3,13 @@ use std::path::Path;
 use std::process::Command;
 use std::str;
 
-fn main() {
+fn main() -> std::io::Result<()> {
     // Add the toolchain lib/ directory to `-L`.  This fixes the linker error "cannot find
     // -lLLVM-13-rust-1.60.0-nightly".
     let out = Command::new("rustup")
         .args(&["which", "rustc"])
         .output()
-        .unwrap();
+        .or_else(|_| Command::new("which").args(&["rustc"]).output())?;
     assert!(out.status.success());
     let rustc_path = Path::new(str::from_utf8(&out.stdout).unwrap().trim_end());
     let lib_dir = rustc_path.parent().unwrap().parent().unwrap().join("lib");
@@ -21,5 +21,6 @@ fn main() {
         .expect("Could not invoke rustc to find rust sysroot");
     assert!(out.status.success());
     let sysroot = str::from_utf8(&out.stdout).unwrap();
-    println!("cargo:rustc-env=RUST_SYSROOT={}", sysroot)
+    println!("cargo:rustc-env=RUST_SYSROOT={}", sysroot);
+    Ok(())
 }
--- a/pdg/build.rs
+++ b/pdg/build.rs
@@ -9,7 +9,10 @@ fn main() -> eyre::Result<()> {
     color_eyre::install()?;
     // Add the toolchain lib/ directory to `-L`.  This fixes the linker error "cannot find
     // -lLLVM-13-rust-1.60.0-nightly".
-    let out = Command::new("rustup").args(&["which", "rustc"]).output()?;
+    let out = Command::new("rustup")
+        .args(&["which", "rustc"])
+        .output()
+        .or_else(|_| Command::new("which").args(&["rustc"]).output())?;
     assert!(out.status.success());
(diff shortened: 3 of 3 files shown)


kkysen (changes requested):
I'm not sure this is correct, as I believe (not sure; if it doesn't, then this is okay) `rustup`'s resolution logic uses `rust-toolchain{,.toml}`, which `which` won't.  If `rustup` is available, we should use it.  I think `which` as a backup is okay, though if we get nasty rust library version errors, those could be hard to debug.

kkysen:
```shell
kkysen in 🌐 donna in c2rust on  master via 🐍 v3.10.4 via 🦀 v1.60.0-nightly via 🐏 6GiB/39GiB | 462MiB/2GiB took 50ms
at 12:07:44 PM ❯ realpath $(dirname $(rustup which rustc))/../lib
/home/kkysen/.rustup/toolchains/nightly-2022-02-14-x86_64-unknown-linux-gnu/lib

kkysen in 🌐 donna in c2rust on  master via 🐍 v3.10.4 via 🦀 v1.60.0-nightly via 🐏 6GiB/39GiB | 462MiB/2GiB took 48ms
at 12:09:11 PM ❯ realpath $(dirname $(which rustc))/../lib
/home/kkysen/.cargo/lib

kkysen in 🌐 donna in c2rust on  master via 🐍 v3.10.4 via 🦀 v1.60.0-nightly via 🐏 6GiB/39GiB | 462MiB/2GiB took 10ms
at 12:07:24 PM ❯ ll $(dirname $(which rustc))/../lib
"/home/kkysen/.cargo/bin/../lib": No such file or directory (os error 2)

kkysen in 🌐 donna in c2rust on  master via 🐍 v3.10.4 via 🦀 v1.60.0-nightly via 🐏 6GiB/39GiB | 462MiB/2GiB took 7ms
at 12:07:39 PM ❯ ll $(dirname $(rustup which rustc))/../lib
.rw-r--r-- 101M kkysen  7 Jun 12:56 libLLVM-13-rust-1.60.0-nightly.so
.rw-r--r-- 119M kkysen  7 Jun 12:56 librustc_driver-f43b82d7671d3119.so
.rw-r--r-- 9.2M kkysen  7 Jun 12:56 libstd-fad6565a4a24ca7c.so
.rw-r--r-- 2.9M kkysen  7 Jun 12:56 libtest-b81f71b2dfdd83d8.so
drwxrwxr-x    - kkysen 28 Jun 18:02 rustlib/
```

`which rustc` resolves to the wrapper `rustc` in `~/.cargo` (by default), not the actual target and toolchain-specific `rustc` which we need.

fw-immunant:
Our existing logic works fine for the case where rustup is installed and rustc was installed through it (using `rustup which rustc` to determine the path of rustc itself). I added fallback logic that works correctly in the case that rustup is not installed and rustc is installed to the system. It may be possible to install rustc through rustup and then remove rustup itself while the existing rustc remains a wrapper script that calls into created by rustup, and you're right that in that case we would fail to find the real path of rustc.

With that said, I don't know that starting from the path of the `rustc` executable is even a sane thing to do to determine the lib dir to pass as `rustc-link-search`. `rustc` has `--print` flags for both `target-libdir` (e.g. `/usr/lib64/rustlib/x86_64-unknown-linux-gnu/lib` or `~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib`, which seem to have the mentioned `libLLVM-…-rust-…-nightly.so` in `../../../`) and `sysroot` (e.g. `/usr` or `~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu`, which both have the lib in their `lib/` directory). Is one of these a better option, and then we could just invoke `rustc --print` and add a suffix?

kkysen:
If you don't have `rustup`, I think you'll need to implement your own resolution logic.  I looked and it doesn't appear `rustc` or `cargo` itself have ways of exposing this information, even though they of course do that resolution logic internally.  You could use something hacky like this, though:

```sh
> RUSTC_WRAPPER=false strace -etrace=execve cargo rustc |& rg 'execve\("([^"]*)".*\) = 0' --only-matching --replace '$1' | tail -n 1
/home/kkysen/.rustup/toolchains/nightly-2022-07-11-x86_64-unknown-linux-gnu/bin/cargo
```

That gives `cargo`, but we just need the directory it's in, so that's okay.

fw-immunant:
I'm not sure I understand... without rustup, what resolution logic is there?

kkysen:
I think there are 2 cases here:

1. `rustup` is being used and exists on `$PATH`.  In this case, we need to use `rustup`.
2. `rustup` is not installed or being used.  I know nothing about this case, so perhaps there doesn't need to be the same kind of resolution logic.  If `$(dirname $(which rustc))/../lib` exists and contains the shared libraries, then that's okay, but I don't know if that's the case.

fw-immunant:
I agree, those are the two cases. This PR makes the latter case work:
```
$ ls $(dirname $(which rustc))/../lib/libLLVM-*-rust-*
/usr/bin/../lib/libLLVM-14-rust-1.64.0-nightly.so
```

kkysen:
Okay, then we should try `rustup which rustc` first, and if it fails on an `ENOENT`, use `which rustc` instead.

kkysen:
Also, this part of the build script should probably be refactored out into another crate instead of repeating it 3 times and having to keep those in sync.

kkysen:
I'm not sure about that, but I'll add that right now whatever we're doing here isn't working for `c2rust instrument`, as we're having to do the same thing, setting `LD_LIBRARY_PATH`.

kkysen:
I believe we may actually be able to use `$RUSTC` from `build.rs` (https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), which works as long as `cargo` is running it, works without or without `rustup`, and doesn't need to redo the lookup.

kkysen:
Using `$RUSTC` would also correctly resolve if we compile using `cargo +stable` for example, which it currently does not when using `rustup which rustc` directly.

kkysen:
Also, what is `analysis/build.rs` doing?  It's not in a crate.

kkysen:
It seems `cargo` only sets `RUSTC=rustc` normally, but it does set `CARGO` to the resolved `cargo`, which is in the same directory as the resolve `rustc`, so we can just use that instead, as we only need their directory.

kkysen:
See these PRs for an alternative to this one:
* https://github.com/immunant/c2rust/pull/515
* https://github.com/immunant/c2rust/pull/517
* https://github.com/immunant/c2rust/pull/518

PR #515, opened 2022-07-13
Delete `analyis/build.rs` as it's unused

kkysen:
I'm pretty sure this is unused as it's not in a crate, `analysis/runtime` doesn't need it, and things build and test fine with it deleted.


kkysen:
@aneksteind, it says you committed this, so I just want to make sure we're not using this somewhere before I delete it.

aneksteind (approved):
If the tests / scripts still work without it, then it's okay to delete.

Issue #504, opened 2022-07-12
Reorganize directory structure


kkysen:

thedataking:
Re. reorganizing scripts, adding subdirectories sounds like a good idea but I'm against renaming most of the scripts just to make them shorter. If anything `pdg.sh` and other terse names need to be longer and consistent with what we had before.

thedataking:
I think the focus should be on the new top level directories. (Whatever we do with scripts is a smaller issue overall.) 

The "old" directories say something about what they do "c2rust-ast-exporter", "c2rust-ast-builder", "c2rust-ast-printer", "c2rust-transpile", "c2rust-refactor".  Also, the current command line interface is `c2rust <verb>` where `c2rust <verb>` causes `c2rust-verb` to run.  E.g. `c2rust transpile ...` runs `c2rust-transpile`.

#### candidates for renaming

- `pdg`:  most folks will not realize that this stands for program dependence graph. 
- `dynamic_instrumentation`: uses underscores instead of kebab case. 
- `analysis`
- `rustc-private-link`: not sure this is c2rust specific, couldn't this be a crate on crates.io or a separate repo?

#### suggested organizing principle

We figure out what commands we plan to add to `c2rust`. If there is going to be a `c2rust lift` command, we put the related code under a top-level directory called `c2rust-lift`. Lift is just an example, the actual verbs may be `instrument`, `observe`, `analyze`, `rewrite` ...

kkysen:
Re `rustc-private-link`, that can definitely be moved to an external repo and crates.io package.  It does make sense to at least initially keep it as a workspace crate, though, as then it's easier to keep it in sync with the rest of the c2rust code and have branches and PRs be able to modify it, too.  Once it's a bit more stable, though, making it a separate repo and package is a good idea I think.  Where would I put that?  Under `immunant/`?  It seems quite small and pretty general, though, as it'd be useful to any project that's using rustc private crates, so I could also maintain it myself separately as well if you'd like.

thedataking:
> Where would I put that? Under `immunant/`? It seems quite small and pretty general, though, as it'd be useful to any project that's using rustc private crates, so I could also maintain it myself separately as well if you'd like.

Great. Either option (`immunant` or under your personal `kkysen` account) sounds good to me; your call.

PR #517, opened 2022-07-13
Refactor `build.rs`s into `rustc-private-link` crate

kkysen:
Fixes https://github.com/immunant/c2rust/issues/516.

This refactors the `build.rs`s in `dynamic-instrumentation`, `pdg`, and `c2rust` into a separate library crate, `rustc-private-link`, to avoid duplication as we're going to add a bunch of fixes/improvements to it now.

It keeps most of the same logic currently used in the `build.rs`s (no improvements yet, look for later PRs on top of this), except for two things:
* Where `color-eyre` was used in `pdg/build.rs`, it drops that as it's a bit overkill for a `build.rs`.
* `$RUSTC` is looked up using `env::var_os` instead of `env::var` and a default of `"rustc"` is used.
* To determine where the `rustc_private` crate dynamic libraries are, it uses `rustc --print sysroot` instead of using any of `rustup which rustc`, `which rustc`, or `$CARGO` and then looking up 2 directories from that.  Those are both the same directories (then the libraries are in `lib/` under the sysroot), and we already use `rustc --print sysroot` in `dynamic-instrumentation/build.rs`).  Thus, this fixes https://github.com/immunant/c2rust/issues/516.

Also, any thoughts on the naming?  Initially I chose `rustc-private-link`, though maybe `link-rustc-private` would be better, but I'm not sure if that's accurate enough, since some of the uses are more sysroot related (`set_env_{rust_sysroot,rustlib}`).

### DIFF
--- a/rustc-private-link/src/lib.rs
+++ b/rustc-private-link/src/lib.rs
@@ -13,6 +13,7 @@ impl SysRoot {
             .expect("could not invoke `rustc` to find rust sysroot");
         let path = String::from_utf8(output.stdout)
             .expect("`rustc --print sysroot` is not UTF-8")
+            .trim_end()
             .into();
         Self { path }
     }
--- a/rustc-private-link/src/lib.rs
+++ b/rustc-private-link/src/lib.rs
@@ -1,0 +1,28 @@
+use std::{env, path::PathBuf, process::Command};
+
+pub struct SysRoot {
+    path: PathBuf,
+}
+
+impl SysRoot {
+    pub fn resolve() -> Self {
+        let rustc = env::var_os("RUSTC").unwrap_or_else(|| "rustc".into());
+        let output = Command::new(rustc)
+            .args(&["--print", "sysroot"])
+            .output()
+            .expect("could not invoke `rustc` to find rust sysroot");
+        let path = String::from_utf8(output.stdout)
+            .expect("`rustc --print sysroot` is not UTF-8")
+            .into();
+        Self { path }
+    }
+
+    pub fn set_env(&self) {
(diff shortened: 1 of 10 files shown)


fw-immunant (approved):

rustc-private-link/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+        println!("cargo:rustc-env=RUST_SYSROOT={}", self.sysroot().display());
+    }
+
+    pub fn set_env_rustlib(&self) {
+        println!("cargo:rustc-env=RUSTLIB={}", self.rustlib().display());
+    }
+
+    pub fn link_rustc_private(&self) {
+        println!("cargo:rustc-link-search={}", self.lib().display());
+    }
```
fw-immunant, line 56: We really shouldn't lossily convert paths to strings with `.display()` to write them to stdout as a command-line argument/env variable. Is there a relatively easy, cross-platform-compatible way to pass these through losslessly?

If not... oh well, it was broken before too.

kkysen: I added that in https://github.com/immunant/c2rust/pull/519.  I skipped doing it for Windows, though, as that's too complicated to spend much time getting exactly right.

thedataking:
The location of this code should be reconsidered. See #504. We keep adding stuff without sufficient coordination IMHO.

kkysen:
See https://github.com/immunant/c2rust/issues/504#issuecomment-1183932776 for more, but I think it should definitely be kept within this repo while it's still very much being actively modified.  Once it's a bit more stable, then we can put it in it's own repo and separate crates.io package with its own versioning.

PR #519, opened 2022-07-13
Allow non-UTF-8 paths in `rustc-private-link`

kkysen:
I'm not sure how important this is, but this seems like an odd restriction.  I mostly left Windows alone, though, as that's too complex, so we only require UTF-8 paths there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -200,6 +200,7 @@ dependencies = [
  "git-testament",
  "log",
  "regex",
+ "rustc-private-link",
  "shlex",
 ]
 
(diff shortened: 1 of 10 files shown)


fw-immunant (approved):
LGTM, thanks!

SimonIT:
Hey, did you test it on windows? Because I get the following error:

```
error[E0433]: failed to resolve: could not find `unix` in `os`
  --> c2rust-build-paths\src\lib.rs:35:26
   |
35 |             use std::os::unix::ffi::OsStrExt;
   |                          ^^^^ could not find `unix` in `os`

error[E0599]: no function or associated item named `from_bytes` found for struct `OsStr` in the current scope
  --> c2rust-build-paths\src\lib.rs:37:20
   |
37 |             OsStr::from_bytes(path)
   |                    ^^^^^^^^^^ function or associated item not found in `OsStr`

Some errors have detailed explanations: E0433, E0599.
```

kkysen:
Thanks for finding this, @SimonIT!  I guess I misunderstood how `if cfg!` works vs. `#[cfg]`.  `c2rust` doesn't actually build easily on Windows due to LLVM issues so I never actually tested this there, but I can fix it now.

PR #521, opened 2022-07-13
PDG references: part 1

oinoom:
Adds support to track shared reference use.

Handles the following cases:

- [x] `_2 = &_1`
- [x] `_3 = _2`
- [x] `_4 = &(*_3)` i.e. trivial reborrows
- [x] shared references as arguments to functions
- [x] returning references from functions

In subsequent PRs:

- [ ] `_9 = &((*_8).field)`
- [ ] mutable references as arguments to functions
- [ ] casts to and from pointers
- [x] `_5 = &raw (*_4)` getting the address of a referenced value, treated as a copy: https://github.com/immunant/c2rust/pull/533

### DIFF
--- a/dynamic_instrumentation/src/instrument_memory.rs
+++ b/dynamic_instrumentation/src/instrument_memory.rs
@@ -338,7 +338,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
             .expect("Could not find pointer load hook");
 
         let base_ty = self.body.local_decls[place.local].ty;
-        // Instrument field projections on raw-ptr places
         if is_region_or_unsafe_ptr(base_ty) && context.is_use() && !place.projection.is_empty() {
             for (pid, (_base, elem)) in place.iter_projections().enumerate() {
                 if let PlaceElem::Field(field, _) = elem {
@@ -386,7 +385,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                 // copying address of dereferenced value does not apply, i.e. _2 = &raw (*_1)
                 && !matches!(self.assignment, Some((_, Rvalue::AddressOf(..))) if base_ty.is_region_ptr())
             {
-                // Place is loading from a raw ptr; trace the raw ptr
                 self.add_instrumentation_point(
                     location,
                     load_fn,
@@ -451,7 +449,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                 // Strip all derefs to set base_dest to the pointer that is deref'd
                 let base_dest = strip_all_deref(&dest, self.tcx);
 
-                // Storing a raw pointer in an indirect destination; trace the destination
                 let base_dest_ty = base_dest.ty(&self.body.local_decls, self.tcx).ty;
                 self.add_instrumentation_point(
                     location,
@@ -471,7 +468,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                 if is_region_or_unsafe_ptr(value_ty) {
                     let mut loc = location;
                     loc.statement_index += 1;
-                    // Stored value is a raw pointer; trace the destination
                     self.add_instrumentation_point(
                         loc,
                         store_value_fn,
@@ -494,7 +490,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                         if !p.is_indirect()
                             && is_region_or_unsafe_ptr(p.ty(&self.body.local_decls, self.tcx).ty)
                         {
-                            // Cast from raw pointer to usize; trace the pointer being casted
                             self.add_instrumentation_point(
                                 location,
                                 ptr_to_int_fn,
@@ -788,7 +783,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                         ) {
                             location.statement_index = 0;
                             location.block = destination.unwrap().1;
-                            // Non-hooked fn with raw-ptr result called; trace destination
                             self.add_instrumentation_point(
                                 location,
                                 arg_fn,
@@ -817,7 +811,6 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
             TerminatorKind::Return => {
                 let place = Place::return_place();
                 if is_region_or_unsafe_ptr(self.body.local_decls[place.local].ty) {
-                    // Function returned raw ptr; trace the return place
                     self.add_instrumentation_point(
                         location,
                         ret_fn,
@@ -982,7 +975,7 @@ fn apply_instrumentation<'tcx>(
                 args.iter_mut().for_each(|arg| *arg = arg.to_copy());
 
                 let place_ty = &place.ty(locals, tcx).ty;
-                // The return type of a hooked fn is always a raw ptr or unit
+                // The return type of a hooked fn is always a raw ptr, reference, or unit
                 if place_ty.is_unit() {
                     // It's somewhat wrong to call unit an AddressUsize, but it has the pass-through
                     // semantics we want


dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -237,15 +241,50 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
     fn should_instrument(&self, local: Local, context: PlaceContext) -> bool {
-        let is_unsafe_ptr = self.body.local_decls[local].ty.is_unsafe_ptr();
+        let is_unsafe_ptr = self.is_shared_region_or_unsafe_ptr(self.body.local_decls[local].ty);
         is_unsafe_ptr && context.is_use()
     }
 
+    fn is_unsafe_ptr(&self, ty: &TyS<'tcx>) -> bool {
+        ty.is_unsafe_ptr()
+    }
+
```
fw-immunant, line 251: This seems dead?

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -237,15 +241,50 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+    fn is_shared_region_or_unsafe_ptr(&self, ty: &TyS<'tcx>) -> bool {
```
fw-immunant, line 252: I would call this `is_shared_or_unsafe_ptr`; `region` doesn't disambiguate anything and `shared ptrs` are a perfectly valid thing to call `&T` that does still distinguish from `&mut T`.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -237,15 +241,50 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+        }
+        Some(Place {
+            local: p.local,
+            projection: tcx.intern_place_elems(&p.projection[..p.projection.len() - 1]),
+        })
+    }
+
+    // For mutable borrows (let _n = &mut place), create a parallel binding that takes
+    // a raw pointer to the same place, without involving _n
+
```
fw-immunant, line 275: This comment doesn't make sense after moving these fns out of their original context.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -465,131 +523,122 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
+                    Rvalue::AddressOf(_, p) => {
+                        // Instrument which local's address is taken
+                        location.statement_index += 1;
                         self.add_instrumentation_point(
                             location,
-                            copy_fn,
-                            vec![InstrumentationOperand::RawPtr(Operand::Copy(dest))],
+                            addr_local_fn,
+                            vec![
+                                InstrumentationOperand::from_type(Operand::Copy(dest), &dest_ty),
```
fw-immunant, line 621: The type of this destination is always a raw ptr because `AddressOf` signifies "`&v as *const _` or raw address of expressions like `&raw v` or `addr_of!(v)`", so don't use `from_type` here to dynamically dispatch on the statically-known type.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -441,19 +496,22 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                         },
                     );
                 }
-            } else if value.ty(&self.body.local_decls, self.tcx).is_integral() {
+            } else if value_ty.is_integral() {
```
fw-immunant, line 499: This isn't new with this PR, but w/r/t this being an `else if`, for code like `*x = myusize as *const _;` (where `x` is an `&mut *const something`), it seems we'll take the `then` branch and emit a `store_value` event but won't emit a `ptr_to_int` event. Do we need to emit the latter in such situations (e.g. to do something with provenance), or do we just care about capturing the pointer value (which the former does)?

aneksteind: Good point, although I think it would actually be an `int_to_ptr` event. I'll consider this a bug that should be addressed in another PR.

aneksteind: #530

kkysen:
Just general advice, I think it would be helpful if you addressed comments in separate commits, which can come with explanations for why that extra change was added.  Otherwise we lose all that history and reasoning if we amend commits and force push.  For small fixes it's fine, but especially if an explanation would be useful, I think it's better to put it in a new commit.  This is mostly new to me, though, so if there are other (dis)advantages I'm missing, let me know.

aneksteind:
> Just general advice, I think it would be helpful if you addressed comments in separate commits, which can come with explanations for why that extra change was added. Otherwise we lose all that history and reasoning if we amend commits and force push. For small fixes it's fine, but especially if an explanation would be useful, I think it's better to put it in a new commit. This is mostly new to me, though, so if there are other (dis)advantages I'm missing, let me know.

I think that's valid, will look to do that as long as no one else minds the commit history including those sort of things on the main branch.

kkysen:
I'm not sure what other people have opinions on the commit history on the main branch; there's certainly a fine line I think, but I would say if a change could use a separate explanation or cause some different behavior, it deserves a commit.  But like a typo or a small format don't need their own commit usually.  Then if anything goes wrong in the future, we can be more fine-grained in tracking it down.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,62 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+
+fn pop_last_projection<'tcx>(p: &Place<'tcx>, tcx: TyCtxt<'tcx>) -> Option<Place<'tcx>> {
+    if p.projection.is_empty() {
+        return None;
+    }
+    Some(Place {
+        local: p.local,
+        projection: tcx.intern_place_elems(&p.projection[..p.projection.len() - 1]),
+    })
+}
```
kkysen, line 261: I refactored this using `split_last`; see https://github.com/immunant/c2rust/pull/521/commits/b29e72608b2b26cceea797f963179e4cbf79cdf4.  Does that look fine?

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,60 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
 }
 
+fn has_outer_deref(p: &Place) -> bool {
+    matches!(
+        p.iter_projections().last(),
+        Some((_, ProjectionElem::Deref))
+    )
+}
+
+fn pop_last_projection<'tcx>(p: &Place<'tcx>, tcx: TyCtxt<'tcx>) -> Option<Place<'tcx>> {
```
kkysen, line 255: `pop` implies it'll return the popped element.  Is `split_last_projection` better, following `List::split_last` and other Rust collections.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,60 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+    let Place { local, projection } = *p;
+    projection.split_last().map(|(_last, rest)| Place {
+        local,
+        projection: tcx.intern_place_elems(rest),
+    })
+}
+
+/// Get the inner-most dereferenced place.
```
kkysen, line 263:
```suggestion
/// Get the inner-most dereferenced [`Place`].
```

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,60 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+                projection: tcx.intern_place_elems(cur_ref.projection),
+            };
+        }
+        place_ref = cur_ref;
+    }
+
+    base_dest
+}
+
+/// Used to strip initital deref from projection sequences
```
kkysen, line 280:
```suggestion
Used to strip initital [`Deref`](ProjectionElem::Deref) from projection sequences.
```

aneksteind: I think the comment string /// is gone in your suggested change

kkysen: whoops

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,60 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
         find_instrumentation_def(self.tcx, self.runtime_crate_did, name)
     }
 
-    fn should_instrument(&self, local: Local, context: PlaceContext) -> bool {
-        let is_unsafe_ptr = self.body.local_decls[local].ty.is_unsafe_ptr();
-        is_unsafe_ptr && context.is_use()
+    fn is_shared_or_unsafe_ptr(&self, ty: &TyS<'tcx>) -> bool {
+        ty.is_unsafe_ptr() || (ty.is_region_ptr() && !ty.is_mutable_ptr())
```
kkysen, line 240: Should we still have a `fn should_instrument` that's updated to also return `true` if it's a shared ptr?  The intent is clearer that way.

And I think `fn is_shared_or_unsafe_ptr` could be split into `fn is_shared_ptr` and `TyS::is_unsafe_ptr`.

aneksteind: i think we have too many context-dependent cases to support `should_instrument` in the form it was in

i'll add a commit to split those up

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -332,14 +380,21 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
             }
 
-            if place.is_indirect() && !context.is_mutating_use() {
+            if place.is_indirect()
+                && !context.is_mutating_use()
+                // reborrows do not apply, i.e. _2 = &(*_1)
+                && !matches!(self.assignment, Some((_, Rvalue::Ref(..))))
+                // copying address of dereferenced value does not apply, i.e. _2 = &raw (*_1)
+                && !matches!(self.assignment, Some((_, Rvalue::AddressOf(..))) if base_ty.is_region_ptr())
+            {
```
kkysen, line 389: Maybe refactor this into a separate method.

dynamic_instrumentation/src/instrument_memory.rs
aneksteind: https://github.com/immunant/c2rust/pull/521/commits/1debde1354b0bdc4c10ccb01f2d3ab16845a8469

dynamic_instrumentation/src/instrument_memory.rs
kkysen: `sans` is good.  `without` would work, too, but I agree `split` isn't the best as the other `split` methods return both parts.

dynamic_instrumentation/src/instrument_memory.rs
kkysen: Okay, if they're different in enough places, then, yeah, it makes sense not to have `should_instrument`.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,64 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
         find_instrumentation_def(self.tcx, self.runtime_crate_did, name)
     }
 
-    fn should_instrument(&self, local: Local, context: PlaceContext) -> bool {
-        let is_unsafe_ptr = self.body.local_decls[local].ty.is_unsafe_ptr();
-        is_unsafe_ptr && context.is_use()
+    fn is_shared_ptr(&self, ty: &TyS<'tcx>) -> bool {
+        ty.is_region_ptr() && !ty.is_mutable_ptr()
+    }
```
kkysen, line 241: It doesn't need `&self`, and then it could be called with `Self::is_shared_ptr`.

spernsteiner:
What's the reason for ignoring `&mut` references for now?  I'm not familiar with the details of the instrumentation, but I'd expect `&` and `&mut` to be handled almost the same.  You'd need a slightly different cast to get the address without upsetting the borrow checker (`&*r as *const _` instead of `r as *const _`), but would there be any other major differences?

fw-immunant:
> I'm not sure what other people have opinions on the commit history on the main branch; there's certainly a fine line I think, but I would say if a change could use a separate explanation or cause some different behavior, it deserves a commit. But like a typo or a small format don't need their own commit usually. Then if anything goes wrong in the future, we can be more fine-grained in tracking it down.

To hop in on the meta discussion, I personally prefer to address review comments kernel-style "v1"/"v2" etc., by incorporating fixups into new versions of each commit so each commit approaches the "ideal" way to make each change. Then the cover letter or PR description can have an explanation of the overall change and deltas between each version. To me the more interesting story is how the post-merge codebase will have changed than how the code in the PR evolved.

aneksteind:
> What's the reason for ignoring &mut references for now? I'm not familiar with the details of the instrumentation, but I'd expect & and &mut to be handled almost the same. You'd need a slightly different cast to get the address without upsetting the borrow checker (&*r as *const _ instead of r as *const _), but would there be any other major differences?

@spernsteiner  I had decided to tackle this first because inserting a statement just above the line of interest, to take the address of something that was mutably borrowed, wasn't working -- particularly with two-phase borrows. I'll make another attempt and delay special treatment of two phase borrows until later on in another PR.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -529,6 +539,43 @@ g {
...
+nodes_that_need_write = []
+
+g {
+	n[0]: &_3        _    => _5 @ bb0[13]: fn test_unique_ref;
+	n[1]: addr.store n[0] => _  @ bb0[18]: fn test_unique_ref;
+}
+nodes_that_need_write = [1, 0]
+
+g {
+	n[0]: &_1 _ => _7 @ bb0[16]: fn test_unique_ref;
```
aneksteind, line 569: these will be tied together properly once https://github.com/immunant/c2rust/pull/533 lands

dynamic_instrumentation/src/instrument_memory.rs
aneksteind: I'm going to save this for a PR for refactoring, which will follow before any new features.

aneksteind:
I've added a change that supports all mutable references except for ones that are arguments to functions: those are the most common use case for two-phase borrows. Support for two-phase borrows will be added in a subsequent PR.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -488,18 +498,21 @@ g {
...
-nodes_that_need_write = [12, 11, 10, 6, 5, 4]
+	n[6]:  copy        n[5]  => _6     @ bb1[3]:  fn fdevent_register;  
+	n[7]:  addr.store  n[5]  => _      @ bb2[0]:  fn fdevent_register;  
+	n[8]:  addr.load   n[5]  => _      @ bb2[3]:  fn fdevent_register;  
+	n[9]:  value.load  _     => _5     @ bb0[4]:  fn fdevent_unregister;
+	n[10]: offset[0]   n[9]  => _4     @ bb0[9]:  fn fdevent_unregister;
+	n[11]: addr.load   n[10] => _      @ bb1[2]:  fn fdevent_unregister;
+	n[12]: value.load  _     => _19    @ bb7[5]:  fn fdevent_unregister;
+	n[13]: offset[0]   n[12] => _18    @ bb7[10]: fn fdevent_unregister;
+	n[14]: copy        n[13] => _17    @ bb8[3]:  fn fdevent_unregister;
```
aneksteind, line 509: similar to https://github.com/immunant/c2rust/pull/521/files#r925774095

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -488,18 +498,21 @@ g {
...
-	n[11]: offset[0]   n[10] => _18    @ bb7[10]: fn fdevent_unregister;
-	n[12]: addr.store  n[11] => _      @ bb8[4]:  fn fdevent_unregister;
-	n[13]: copy        n[1]  => _20    @ bb6[7]:  fn lighttpd_test;     
-	n[14]: copy        n[13] => _19    @ bb6[8]:  fn lighttpd_test;     
-	n[15]: free        n[14] => _18    @ bb6[9]:  fn lighttpd_test;     
-}
-nodes_that_need_write = [12, 11, 10, 6, 5, 4]
+	n[6]:  copy        n[5]  => _6     @ bb1[3]:  fn fdevent_register;  
+	n[7]:  addr.store  n[5]  => _      @ bb2[0]:  fn fdevent_register;  
+	n[8]:  addr.load   n[5]  => _      @ bb2[3]:  fn fdevent_register;
```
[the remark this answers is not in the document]
aneksteind: to clarify, `fresh0` is a reference, which is why this is new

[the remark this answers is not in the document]
spernsteiner: Looking at this and the previous bit, if the code is something like

```Rust
let ref mut fresh0 = *((*ev).fdarray).offset(fd as isize);
*fresh0 = ...;   // Is this present?  bb2[0]?
let mut fdn: *mut fdnode = *fresh0;  // bb2[3]?
```

then I think `n[7]` and `n[8]` should have `n[6]` (`_6` / `fresh0`) as their source, rather than `n[5]` (`_7` / the temporary storing the result of `offset`).  Also, if the original Rust code doesn't contain a write to `*fresh0` like the line I marked `bb2[0]?`, then `n[7]` should not be present - there should only be a `StoreAddr` if `*fresh0` is actually written, and things like `&mut (*fresh0)` don't actually touch memory.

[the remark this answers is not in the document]
aneksteind: regarding the source node of `n[6]`, I think you're right and this is a bug. looking into it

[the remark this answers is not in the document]
aneksteind: Fixed in 3f4b5a8e

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,64 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+
+/// Get the inner-most dereferenced [`Place`].
+fn strip_all_deref<'tcx>(p: &Place<'tcx>, tcx: TyCtxt<'tcx>) -> Place<'tcx> {
+    let mut base_dest = *p;
+    let mut place_ref = base_dest.as_ref();
+    while let Some((cur_ref, proj)) = place_ref.last_projection() {
+        if let ProjectionElem::Deref = proj {
+            base_dest = Place {
+                local: cur_ref.local,
+                projection: tcx.intern_place_elems(cur_ref.projection),
```
rinon, line 275: Nit: This will intern multiple times if there are multiple derefs, right? Might be better to keep a ref to the inner-most deref and then intern and return that at the end?

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -236,16 +236,64 @@ impl<'a, 'tcx: 'a> CollectFunctionInstrumentationPoints<'a, 'tcx> {
...
+    base_dest
+}
+
+/// Used to strip initital deref from projection sequences
+fn remove_outer_deref<'tcx>(p: Place<'tcx>, tcx: TyCtxt<'tcx>) -> Place<'tcx> {
+    // Remove outer deref if present
+    match p.as_ref().last_projection() {
+        Some((_, ProjectionElem::Deref)) => {
+            let sans_proj =
+                sans_last_projection(&p, tcx).expect("expected but did not find deref projection");
```
rinon, line 290: You only use this function here, right? Wouldn't it be simpler to combine the two, i.e. do the split_last and just return the rest if it matches?

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -293,19 +341,19 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
             .find_instrumentation_def(Symbol::intern("ptr_load"))
             .expect("Could not find pointer load hook");
 
+        let base_ty = self.body.local_decls[place.local].ty;
         // Instrument field projections on raw-ptr places
```
rinon, line 345: Is this comment still accurate? Looks like you now instrument references here too.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -332,18 +383,25 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                 }
             }
 
-            if place.is_indirect() && !context.is_mutating_use() {
+            if place.is_indirect()
+                && !context.is_mutating_use()
+                // reborrows do not apply, i.e. _2 = &(*_1)
+                && !matches!(self.assignment, Some((_, Rvalue::Ref(..))))
```
rinon, line 389: I think this match and the following match should each get split into mini functions with a descriptive name or use bool variables with names, rather than adding comments in the middle of the conditional.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -419,19 +467,22 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
+                        source: Some(to_mir_place(&remove_outer_deref(dest, self.tcx))),
                         destination: None,
                         transfer_kind: TransferKind::None,
                     },
                 );
-                if value.ty(&self.body.local_decls, self.tcx).is_unsafe_ptr() {
+                if is_region_or_unsafe_ptr(value_ty) {
                     let mut loc = location;
                     loc.statement_index += 1;
                     // Stored value is a raw pointer; trace the destination
```
rinon, line 478: Not just raw pointers anymore

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -441,19 +492,20 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
                 }
-            } else if value.ty(&self.body.local_decls, self.tcx).is_integral() {
+            } else if value_ty.is_integral() {
                 if let Rvalue::Cast(_, op, _) = value {
                     if let Some(p) = op.place() {
                         if !p.is_indirect()
-                            && p.ty(&self.body.local_decls, self.tcx).ty.is_unsafe_ptr()
+                            && is_region_or_unsafe_ptr(p.ty(&self.body.local_decls, self.tcx).ty)
                         {
                             // Cast from raw pointer to usize; trace the pointer being casted
```
rinon, line 501: Not just raw pointers anymore

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -465,148 +517,154 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
-                            },
-                        );
+            } else if is_region_or_unsafe_ptr(value_ty) {
+                match value {
+                    Rvalue::Use(Operand::Copy(p)) | Rvalue::Use(Operand::Move(p)) => {
+                        location.statement_index += 1;
+                        if p.is_indirect() {
+                            // We're dereferencing something, the result of which is another pointer;
+                            // trace the destination
+                            self.add_instrumentation_point(
```
rinon, line 527: You could just set `source` and `instrument_fn` locals in both sides of this if and combine the instrumentation call, saving some redundant code.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -465,148 +517,154 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
-                                transfer_kind: TransferKind::None,
-                            },
-                        );
-                    } else {
-                        // Casting between pointer types; trace the destination
+                    Rvalue::Cast(_, p, _) => {
+                        location.statement_index += 1;
+                        if p.ty(&self.body.local_decls, self.tcx).is_integral() {
+                            // Casting integer to a pointer; trace the destination
+                            self.add_instrumentation_point(
```
rinon, line 565: This call could also be combined with the one below pretty cleanly.

rinon (approved):
I haven't read the snapshot output at all. This review is basically just structure and readability, I'm generally assuming it actually _does_ what we want it to (at least for what we've tested so far) although at first glance it seems to be right. I added a few comments, mostly style and comments, nothing blocking.

dynamic_instrumentation/src/instrument_memory.rs
aneksteind: this will be addressed in #552

dynamic_instrumentation/src/instrument_memory.rs
aneksteind: will do in #552

dynamic_instrumentation/src/instrument_memory.rs
aneksteind: will do in #552

PR #523, opened 2022-07-14
builder: use `Token![]` macro instead of explicit token names

afetisov:
The `Token` macro is more idiomatic, since it is well-known to anyone working with proc macros, more widespread than explicit token struct names, and directly references the familiar pieces of Rust code.


kkysen:
This looks good.  I agree it's a lot more idiomatic.  I'll merge it once it passes CI.

PR #524, opened 2022-07-14
Transpiler cleanup 15.07

afetisov:
* remove UB const DUMMY_SP, use safe `Span::call_site()` instead;
* simplify argument parsing chains `it.next().unwrap().parse().unwrap()` into a single pattern match;
* extract common logic for adding architecture-dependent item usage in simd.rs into a separate function;
* various minor cleanups.

### DIFF
--- a/c2rust-transpile/src/rust_ast/mod.rs
+++ b/c2rust-transpile/src/rust_ast/mod.rs
@@ -16,9 +16,8 @@ fn raise_span_limit(_new_limit: u32) {
     if new_limit >= limit {
         let delta = new_limit - limit;
         let s = str::repeat("        ", (delta as usize + 7) / 8);
-        use std::str::FromStr;
         /* used only for its side-effect of expanding the source map */
-        let _ = proc_macro2::TokenStream::from_str(&s);
+        let _ = s.parse::<proc_macro2::TokenStream>();
         SPAN_LIMIT.store(new_limit, Ordering::Relaxed);
     }
 }
@@ -139,9 +138,9 @@ struct SpanRepr {
 /// On the plus side, the `fallback::Span` payload is a POD pair of two u32s, so that case is trivial.
 fn validate_repr() {
     let repr: SpanRepr = unsafe { std::mem::transmute(Span::call_site()) };
-    assert!(repr.compiler_or_fallback == 1);
-    assert!(repr.lo == 0);
-    assert!(repr.hi == 0);
+    assert_eq!(repr.compiler_or_fallback, 1);
+    assert_eq!(repr.lo, 0);
+    assert_eq!(repr.hi, 0);
 }
 
 /** return a pair of mutable references to s.lo and s.hi */
--- a/c2rust-transpile/src/translator/assembly.rs
+++ b/c2rust-transpile/src/translator/assembly.rs
@@ -496,23 +496,22 @@ fn map_input_op_idx(
     if let Some(adj_idx) = idx.checked_sub(num_output_operands) {
         // will only be Some(idx) if it was an input operand
         match tied_operands.get(&(adj_idx, false)) {
-            Some(&out_idx) => {
-                return out_idx;
-            }
+            Some(&out_idx) => out_idx,
             None => {
                 // get number of tied inputs before this index
                 // TODO: can calculate this once before the call, not once for each input
                 let num_tied_before = tied_operands
                     .keys()
-                    .filter(|&&(iidx, is_out)| !is_out && iidx < adj_idx)
+                    .copied()
+                    .filter(|&(iidx, is_out)| !is_out && iidx < adj_idx)
                     .count();
                 // shift the original index by the number of tied input operands prior to the current one
-                return idx - num_tied_before;
+                idx - num_tied_before
             }
-        };
+        }
+    } else {
+        idx
     }
-
-    idx
 }
 
 /// Rewrite a LLVM inline assembly template string into an asm!-compatible one
@@ -687,13 +686,7 @@ impl<'c> Translation<'c> {
         let mut tokens: Vec<TokenTree> = vec![];
 
         let mut tied_operands = HashMap::new();
-        for (
-            input_idx,
-            &AsmOperand {
-                ref constraints, ..
-            },
-        ) in inputs.iter().enumerate()
-        {
+        for (input_idx, AsmOperand { constraints, .. }) in inputs.iter().enumerate() {
             let constraints_digits = constraints.trim_matches(|c: char| !c.is_ascii_digit());
             if let Ok(output_idx) = constraints_digits.parse::<usize>() {
                 let output_key = (output_idx, true);
@@ -722,10 +715,9 @@ impl<'c> Translation<'c> {
                 if let Ok(idx) = ref_str.parse::<usize>() {
                     outputs
                         .iter()
-                        .chain(inputs.iter())
+                        .chain(inputs)
                         .nth(idx)
-                        .map(operand_is_mem_only)
-                        .unwrap_or(false)
+                        .map_or(false, operand_is_mem_only)
                 } else {
                     false
                 }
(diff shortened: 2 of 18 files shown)


c2rust-transpile/src/translator/assembly.rs
```diff
@@ -496,23 +496,22 @@ fn map_input_op_idx(
...
-            }
+            Some(&out_idx) => out_idx,
             None => {
                 // get number of tied inputs before this index
                 // TODO: can calculate this once before the call, not once for each input
                 let num_tied_before = tied_operands
                     .keys()
-                    .filter(|&&(iidx, is_out)| !is_out && iidx < adj_idx)
+                    .copied()
+                    .filter(|&(iidx, is_out)| !is_out && iidx < adj_idx)
```
kkysen, line 506: I think the previous was cleaner.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -496,23 +496,22 @@ fn map_input_op_idx(
...
-                return idx - num_tied_before;
+                idx - num_tied_before
             }
-        };
+        }
+    } else {
+        idx
     }
-
-    idx
```
kkysen, line 515: If you're cleaning up this method, I'd do a

```rust
let adj_idx = match idx.checked_sub(num_output_operands) {
    None => return idx,
    Some(adj_idx) => adj_idx,
};
```

at the top of the method.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -722,10 +715,9 @@ impl<'c> Translation<'c> {
                 if let Ok(idx) = ref_str.parse::<usize>() {
                     outputs
                         .iter()
-                        .chain(inputs.iter())
+                        .chain(inputs)
                         .nth(idx)
-                        .map(operand_is_mem_only)
-                        .unwrap_or(false)
+                        .map_or(false, operand_is_mem_only)
```
kkysen, line 720: The previous way seems cleaner to me.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -646,16 +646,9 @@ impl<'c> Translation<'c> {
...
-                .ok_or("Missing first argument to convert_overflow_arith")?;
-            let b = args
-                .next()
-                .ok_or("Missing second argument to convert_overflow_arith")?;
-            let c = args
-                .next()
-                .ok_or("Missing third argument to convert_overflow_arith")?;
+            let [a, b, c]: [_; 3] = args
+                .try_into()
+                .expect("`convert_overflow_arith` must have exactly 3 arguments");
```
kkysen, line 651: This changes the behavior to require exactly 3 args, whereas previously it required just at least 3 args.  That's probably the behavior we want, though.  @fw-immunant, thoughts?

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -691,16 +684,9 @@ impl<'c> Translation<'c> {
...
-                .ok_or("Missing dst argument to convert_libc_fns")?;
-            let c = args
-                .next()
-                .ok_or("Missing c argument to convert_libc_fns")?;
-            let len = args
-                .next()
-                .ok_or("Missing len argument to convert_libc_fns")?;
+            let [dst, c, len]: [_; 3] = args
+                .try_into()
+                .expect("`convert_libc_fns` must have exactly 3 arguments");
```
kkysen, line 689: Same as previous comment.

c2rust-transpile/src/translator/simd.rs
```diff
@@ -428,16 +386,9 @@ impl<'c> Translation<'c> {
...
-                .ok_or("Missing first param in convert_shuffle_vector")?;
-            let second = params
-                .next()
-                .ok_or("Missing second param in convert_shuffle_vector")?;
-            let third = params
-                .next()
-                .ok_or("Missing third param in convert_shuffle_vector")?;
+            let [first, second, third]: [_; 3] = params
+                .try_into()
+                .expect("`convert_shuffle_vector` is expected to have exactly 3 parameters");
```
kkysen, line 391: Same as that other comment.

kkysen (changes requested):
Hi @afetisov, thanks for another PR!  Could you split this one up, though, as there are a few separate changes happening here?
* the `DUMMY_SP` to `Span::call_site()` change:
    * @fw-immunant is a lot more familiar with this code, but if this is valid, that would be a great fix, getting rid of the UB.  Thanks!
* the argument parsing changes, which slightly change behavior, though probably in the way that we want
* the `add_arch_use` refactor
* the pure cleanups

c2rust-transpile/src/rust_ast/mod.rs
```diff
@@ -28,14 +27,6 @@ pub fn pos_to_span(pos: BytePos) -> Span {
...
 }
 
-pub const DUMMY_SP: Span = unsafe {
-    std::mem::transmute(SpanRepr {
-        compiler_or_fallback: 1,
-        lo: 0,
-        hi: 0,
-    })
-};
-
```
kkysen, line 38: If we remove `DUMMY_SP`, do we still need `SpanRepr` (I'm not sure)?

afetisov:
If I split it up, all the comments you made here will be scrapped :-/

c2rust-transpile/src/rust_ast/mod.rs
afetisov: Yes, unfortunately. It is still used in `get_inner` and `synthesize`. It is still UB. At the moment I have no solution.

c2rust-transpile/src/translator/assembly.rs
afetisov: That's certainly an option, but it introduces annoying identifier stuttering: `adj_idx`, `adj_idx`, `adj_idx`. Personally I would prefer to leave it as a nested if until `let ... else` syntax is stabilized, at which point such nesting would be best to refactor away.

c2rust-transpile/src/translator/assembly.rs
afetisov: Double references are usually confusing.

c2rust-transpile/src/translator/assembly.rs
afetisov: It's certainly a matter of taste. However, the `map_or` function is already used 45 times in the project (10 times in transpiler, 35 in refactorer, and 1 in build script), and there is some value in shorter method chains (since there are less operations to parse in the head).

c2rust-transpile/src/rust_ast/mod.rs
kkysen: It seems `get_inner` could be implemented inefficiently and very unstably, but soundly, by parsing `Span`'s `Debug` representation.  And `get_inner_mut` is unused, but `synthesize` I'm not sure how to do without UB.  Also, `SpanRepr` should be the same source code as `Span`, which would minimize chances of it breaking in practice, as `rustc` is unlikely to compile the same source code with different ABIs.

kkysen:
None of them are too in-depth fortunately; I can re-make them.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -722,10 +715,9 @@ impl<'c> Translation<'c> {
...
                 } else {
                     false
                 }
```
kkysen, line 723: This could be rewritten as:

```rust
ref_str.parse::<usize>().ok().and_then(|idx| {
    outputs
        .iter()
        .chain(inputs)
        .nth(idx)
        .map(operand_is_mem_only)
}).unwrap_or_default()
```

c2rust-transpile/src/translator/assembly.rs
kkysen: I disagree.  You want longer method chains if each step in the chain is short.  That way you only need to keep one small step at a time in your head, instead of two at once.

PR #525, opened 2022-07-14
Print `Graph` hashes instead of indices

kkysen:
When pretty-printing a `Graph`, instead of the `GraphId` index, print the `Graph`'s hash.

The reason we print the hash instead of the index is so that `Graph`s remain stable when their position changes,
 so when changes cause a new `Graph` to be emitted,
 it doesn't cause a snapshot diff on all subsequent `Graph`s, only the new one.

As for the actual hash printed, the hash (a `u64`) is turned into a `u32` and then printed as base64 to be succinct and readable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,6 +86,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


aneksteind:
I'm not sure we get much from printing any ID / hash at all, and removing it altogether in the printed output could accomplish the same thing.

kkysen:
I thought it might be useful for identifying a graph when discussing it with someone.

aneksteind:
Typically we talk about one object at a time because they're all self-contained, and when talking about changes to a particular object, we can refer to the snapshot file differences.

PR #526, opened 2022-07-14
remove unused dependencies and features

afetisov:
Disable default features for dependencies where features are unused.
Remove `clap` and `c2rust-bitfields` from dependencies of `c2rust-transpile` (unused).

### DIFF
--- a/c2rust-ast-exporter/Cargo.toml
+++ b/c2rust-ast-exporter/Cargo.toml
@@ -20,7 +20,7 @@ serde_bytes = "0.11"
 serde_cbor = "0.11"
 
 [build-dependencies]
-bindgen = { version = "0.60", features = ["logging"] }
+bindgen = { version = "0.60", features = ["logging"], default-features = false }
 clang-sys = "1.3"
 # Pinning until https://github.com/rust-lang/cmake-rs/issues/131 is resolved
 # Fixed by https://github.com/rust-lang/cmake-rs/pull/146 on 5/12/2022; waiting for next release.
@@ -29,6 +29,5 @@ env_logger = "0.9"
 
 [features]
 default = []
-
 # Force static linking of LLVM
 llvm-static = []
--- a/c2rust-ast-builder/Cargo.toml
+++ b/c2rust-ast-builder/Cargo.toml
@@ -12,5 +12,5 @@ description = "Rust AST builder support crate for the C2Rust project"
 edition = "2021"
 
 [dependencies]
-proc-macro2 = { version = "1.0", features = ["span-locations"]}
-syn = { version = "1.0", features = ["full", "extra-traits", "printing"]}
+proc-macro2 = { version = "1.0", features = ["span-locations"], default-features = false }
+syn = { version = "1.0", features = ["full", "extra-traits", "printing"], default-features = false }
--- a/c2rust-ast-printer/Cargo.toml
+++ b/c2rust-ast-printer/Cargo.toml
@@ -9,6 +9,6 @@ description = "Customized version of libsyntax rust pretty-printer"
 
 [dependencies]
 log = "0.4"
-syn = { version = "1.0", features = ["full", "proc-macro", "printing"] }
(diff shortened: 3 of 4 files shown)


c2rust-ast-exporter/Cargo.toml
```diff
@@ -14,20 +14,23 @@ description = "Clang AST extraction API for use in the C2Rust project"
 edition = "2021"
 
 [dependencies]
+
```
kkysen, line 17: Why an empty line here?

c2rust-ast-exporter/Cargo.toml
```diff
@@ -14,20 +14,23 @@ description = "Clang AST extraction API for use in the C2Rust project"
...
 libc = "0.2"
 serde = "1.0"
 serde_bytes = "0.11"
 serde_cbor = "0.11"
 
 [build-dependencies]
-bindgen = { version = "0.60", features = ["logging"] }
+
```
kkysen, line 24: And here.

c2rust-ast-exporter/Cargo.toml
```diff
@@ -14,20 +14,23 @@ description = "Clang AST extraction API for use in the C2Rust project"
...
+bindgen = { version = "0.60", features = ["logging"], default-features = false }
 clang-sys = "1.3"
 # Pinning until https://github.com/rust-lang/cmake-rs/issues/131 is resolved
 # Fixed by https://github.com/rust-lang/cmake-rs/pull/146 on 5/12/2022; waiting for next release.
 cmake = "=0.1.45"
 env_logger = "0.9"
 
 [features]
+
```
kkysen, line 33: And here

kkysen (approved):
This looks good, just there are a couple extra empty lines that I'm not sure why they got added.

c2rust-ast-exporter/Cargo.toml
afetisov: I think it looks better.

kkysen: That's not how `Cargo.toml` and TOML is generally formatted.  Could you delete those empty lines?  Otherwise this PR is great.

kkysen:
@afetisov, this should've updated `Cargo.lock`, too, which I should've noticed.  I committed those changed to `master` in https://github.com/immunant/c2rust/commit/06d8cb67b686efad730b089cfd4eae91f96a76db.

kkysen:
This change, specifically the `default-features = false` changes, broke `c2rust-ast-builder` and `c2rust-ast-printer`.  They no longer compile with `cargo build -p c2rust-ast-{builder,printer}`.  I'm not sure why `c2rust-transpile`, which depends on these, still works, and I likely wouldn't have noticed this breakage if `rust-analyzer` didn't mysteriously start throwing errors about lacking `Clone` `impl`s.

kkysen:
This also broke `cargo test`, as `c2rust-transpile` does depend on `c2rust-bitfields` in its tests.  @afetisov, please be more careful about these kinds of changes, and I'll be more careful about approving the PRs as well.  More of these checks will be added to CI very soon, but they aren't quite ready yet.

afetisov:
Damn, I've tested them :-( The problem is that, in a workspace, if some crate enables a feature, then every crate can use it, even if that crate has all features disabled. This means that the only proper way to test this stuff is to make a clean build of each crate separately, and even then you may accidentally depend on the features in your dependencies.

I don't really know what could be done to prevent that mess in the future. Cargo check and cargo clippy also fail to find that issue. RA may help, but I'm using Intellij, and I don't know if RA checks can be automated.

Come to think of it, Intellij was also complaining about missing clone impls, but I have disregarded it as its bug. Such stuff happens.

kkysen:
This also broke macOS builds (https://github.com/immunant/c2rust/issues/569), so I've reverted everything except for the `clap` dependency deletion (https://github.com/immunant/c2rust/pull/571), as I've more thoroughly checked that that change is fine.

PR #527, opened 2022-07-15
remove `c2rust-macros` from the workspace as it's unused

afetisov:
This crate is used only in c2rust-refactor, which is itself excluded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,6 @@ members = [
     "c2rust-ast-exporter",
     "c2rust-ast-printer",
     "c2rust-bitfields",
-    "c2rust-macros",
     "c2rust-asm-casts",
     "analysis/runtime",
     "dynamic_instrumentation",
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -6,14 +6,14 @@ members = [
...
     "analysis/runtime",
     "dynamic_instrumentation",
     "pdg",
     "rustc-private-link",
 ]
 exclude = [
+    "c2rust-macros",
     "cross-checks/pointer-tracer",
     "cross-checks/zero-malloc",
     "cross-checks/rust-checks",
```
kkysen, line 19: Can do in another PR, but I think we can remove these (as they don't exist anymore) and add `c2rust-refactor` to `exclude`.

kkysen (approved):
`c2rust-macros` is only used by `c2rust-refactor`, which isn't in the workspace either, so this is good.  @rinon agreed.

Cargo.toml
kkysen: * `c2rust-refactor` excluded in b0ca01f87ac831d5fa67a561215cc10918e58122
* `cross-checks/*` removed in 3ed73b1bce62fd952e1e5d259598bc7aea50311b

PR #528, opened 2022-07-15
transpile: allow stabilized const integer arithmetic methods

afetisov:
The wrapping arithmetic methods on unsigned integers are stably const now.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -611,11 +611,6 @@ impl<'c> Translation<'c> {
             }
 
             c_ast::BinOp::Multiply if is_unsigned_integral_type => {
-                if ctx.is_const {
-                    return Err(TranslationError::generic(
-                        "Cannot use wrapping multiply in a const expression",
-                    ));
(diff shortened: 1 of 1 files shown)


kkysen (commented):
This looks good, though untested.  A translator test under `tests/` would be nice.

kkysen (approved):
On second thought, I think this is a change we want regardless of if `rustc` can compile the resultant code.  We shouldn't prematurely decide if it can or cannot, as `rustc` will likely give a much better error message anyways, and it's more future proof, as it allows `rustc` to improve separately.

kkysen:
@afetisov, I just realized now, but this missed another such error on `const` wrapping arithmetic (fixed in https://github.com/immunant/c2rust/commit/538690e05ae5bba2796eb1903d0031b036eac133) and the resulting unused `ctx: ExprContext` arguments (fixed in https://github.com/immunant/c2rust/commit/e3a3808a7a0a90a8e3fecd8ef50dffe94092b77b).  Hopefully we can get `clippy` in CI working soon so this will be caught.

PR #533, opened 2022-07-15
add support for taking addresses of referenced values

aneksteind:


pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -158,10 +158,66 @@ g {
 nodes_that_need_write = [3, 2, 1, 0]
 
 g {
-	n[0]: &_9  _    => _8 @ bb2[13]: fn exercise_allocator;
-	n[1]: copy n[0] => _7 @ bb2[14]: fn exercise_allocator;
-	n[2]: copy n[1] => _6 @ bb2[16]: fn exercise_allocator;
-	n[3]: copy n[2] => _1 @ bb0[0]:  fn printf;            
+	n[0]:  copy _     => _8  @ bb2[13]:  fn exercise_allocator;
```
aneksteind, line 161: all uses of `b"%i\n\x00"` across the program, such as:

```rust
printf(b"%i\n\x00" as *const u8 as *const libc::c_char, ...);
```

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -544,6 +528,7 @@ g {
 	n[3]: copy n[2] => _1 @ bb0[0]:  fn shared_ref_foo; 
 	n[4]: copy n[3] => _0 @ bb0[1]:  fn shared_ref_foo; 
 	n[5]: copy n[4] => _4 @ bb1[0]:  fn test_shared_ref;
+	n[6]: copy n[5] => _6 @ bb1[4]:  fn test_shared_ref;
```
aneksteind, line 531: corresponds to:

```rust
let bar = std::ptr::addr_of!(*foo);
```

where `foo` is a shared reference

fw-immunant (approved):
The functional changes look fine, but I would prefer the usage of `InstrumentationOperand::from_type` to be replaced with a more specialized method because we can't, in general, know how we want to instrument an `Operand` based solely on its type: e.g., a `Place` that holds a usize or a pointer may be one we want to log the address of or the content of, depending on how it's used rather than its type, and `::Place` is applied in cases where we do want to log the address of the `Place`. Otherwise, if we get confused about which types will be passed to `from_type`, we may log pointers to locals instead of their contents (or vice-versa) and end up with horribly confusing bugs.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
spernsteiner: This sounds like it should be `AddrOfStatic -> Copy`, but `AddrOfStatic` is not implemented yet.

aneksteind: agreed

PR #538, opened 2022-07-17
transpile: simplify dynamic argument parsing

afetisov:
Instead of `it.next().unwrap().parse().unwrap()` chains use slice patterns and a single `expect` call.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -646,16 +646,9 @@ impl<'c> Translation<'c> {
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         let args = self.convert_exprs(ctx.used(), args)?;
         args.and_then(|args| {
-            let mut args = args.into_iter();
-            let a = args
-                .next()
-                .ok_or("Missing first argument to convert_overflow_arith")?;
-            let b = args
-                .next()
-                .ok_or("Missing second argument to convert_overflow_arith")?;
-            let c = args
-                .next()
-                .ok_or("Missing third argument to convert_overflow_arith")?;
+            let [a, b, c]: [_; 3] = args
+                .try_into()
+                .map_err(|_| "`convert_overflow_arith` must have exactly 3 arguments")?;
             let overflowing = mk().method_call_expr(a, method_name, vec![b]);
             let sum_name = self.renamer.borrow_mut().fresh();
             let over_name = self.renamer.borrow_mut().fresh();
@@ -691,16 +684,9 @@ impl<'c> Translation<'c> {
         let mem = mk().path_expr(vec!["libc", name]);
         let args = self.convert_exprs(ctx.used(), args)?;
         args.and_then(|args| {
-            let mut args = args.into_iter();
-            let dst = args
-                .next()
-                .ok_or("Missing dst argument to convert_libc_fns")?;
-            let c = args
-                .next()
-                .ok_or("Missing c argument to convert_libc_fns")?;
-            let len = args
-                .next()
-                .ok_or("Missing len argument to convert_libc_fns")?;
+            let [dst, c, len]: [_; 3] = args
+                .try_into()
+                .map_err(|_| "`convert_libc_fns` must have exactly 3 arguments: [dst, c, len]")?;
             let size_t = mk().path_ty(vec!["libc", "size_t"]);
             let len1 = mk().cast_expr(len, size_t);
             let mem_expr = mk().call_expr(mem, vec![dst, c, len1]);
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/simd.rs
```diff
@@ -428,16 +428,9 @@ impl<'c> Translation<'c> {
...
-                .ok_or("Missing first param in convert_shuffle_vector")?;
-            let second = params
-                .next()
-                .ok_or("Missing second param in convert_shuffle_vector")?;
-            let third = params
-                .next()
-                .ok_or("Missing third param in convert_shuffle_vector")?;
+            let [first, second, third]: [_; 3] = params
+                .try_into()
+                .expect("`convert_shuffle_vector` is expected to have exactly 3 parameters");
```
kkysen, line 433:
```suggestion
                .expect("`convert_shuffle_vector` must have exactly 3 parameters");
```

Use the same error message as the others.  `parameters` instead of `arguments` is okay as the variable is called `params`, but the rest should be the same.

c2rust-transpile/src/translator/simd.rs
```diff
@@ -428,16 +428,9 @@ impl<'c> Translation<'c> {
         let param_translation =
             self.convert_exprs(ctx.used(), &[first_expr_id, second_expr_id, mask_expr_id])?;
```
kkysen, line 429: This change would be obviated by [`array::try_map`](https://doc.rust-lang.org/std/primitive.array.html#method.map)/[`#![feature(array_try_map)]`](https://github.com/rust-lang/rust/issues/79711), but it doesn't look like it'll land soon.  [`array::map`](https://doc.rust-lang.org/std/primitive.array.html#method.map) is stable, but that doesn't work because of the error handling.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -691,16 +684,9 @@ impl<'c> Translation<'c> {
...
-                .ok_or("Missing dst argument to convert_libc_fns")?;
-            let c = args
-                .next()
-                .ok_or("Missing c argument to convert_libc_fns")?;
-            let len = args
-                .next()
-                .ok_or("Missing len argument to convert_libc_fns")?;
+            let [dst, c, len]: [_; 3] = args
+                .try_into()
+                .expect("`convert_libc_fns` must have exactly 3 arguments");
```
kkysen, line 689:
```suggestion
                .expect("`convert_libc_fns` must have exactly 3 arguments: [dst, c, len]");
```

Unlike the other two, the error messages here specify the argument names, so I'd keep them.

c2rust-transpile/src/translator/simd.rs
kkysen, line 433:
```suggestion
                .map_err(|_| "`convert_shuffle_vector` is expected to have exactly 3 parameters")?;
```

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -646,16 +646,9 @@ impl<'c> Translation<'c> {
...
-                .ok_or("Missing first argument to convert_overflow_arith")?;
-            let b = args
-                .next()
-                .ok_or("Missing second argument to convert_overflow_arith")?;
-            let c = args
-                .next()
-                .ok_or("Missing third argument to convert_overflow_arith")?;
+            let [a, b, c]: [_; 3] = args
+                .try_into()
+                .expect("`convert_overflow_arith` must have exactly 3 arguments");
```
kkysen, line 651:
```suggestion
                .map_err(|_| "`convert_overflow_arith` must have exactly 3 arguments")?;
```

c2rust-transpile/src/translator/builtins.rs
kkysen, line 689:
```suggestion
                .map_err(|_| "`convert_libc_fns` must have exactly 3 arguments")?;
```

kkysen (changes requested):
I left some small suggestions on error messages and naming, but the main issue here is that previously bubble-up errors are now panicked, i.e. `.ok_or(...)?` was changed to `.expect(...)`.  I added suggestions to change these back to `.map_err(|_| ...)?` to get the previous behavior back.  In the `simd.rs` case, `.expect` could be okay as the array is created statically and has exactly 3 elements, but `.map_err(|_| ...)?` is still better to keep consistency with the old behavior, and avoid introducing panics when refactoring later.  Once it's changed to `.map_err(|_| ...)?`, this should be good to go.

c2rust-transpile/src/translator/simd.rs
afetisov: What change? I didn't change anything here. If you mean that the `self.convert_exprs` call can be removed, that is problematic because of the `WithStmts` type, which cannot be constructed from an array via any current means (fixed-sized iterators are not even unstable).

kkysen: Sorry, I meant this comment more as a lament that this can't be written more nicely.  It's fine the way you wrote it, modulo my other comments.

kkysen (approved):

PR #539, opened 2022-07-17
Remove DUMMY_SP and replace it with safe `Span::call_site()`

afetisov:
This removes UB in the creation of `DUMMY_SP`. Unfortunately, `SpanRepr` and the associated UB transmutes cannot be removed so easily.


c2rust-transpile/src/rust_ast/mod.rs
```diff
@@ -205,35 +196,3 @@ impl SpanExt for Span {
...
-    }
-
-    fn inner(&self) -> (u32, u32) {
-        (self.lo, self.hi)
-    }
-
-    fn inner_mut(&mut self) -> (&mut u32, &mut u32) {
-        (&mut self.lo, &mut self.hi)
-    }
-}
```
kkysen, line 239: I'd leave `MySpan` for now in this PR.  Remove it in a separate PR.  I'm not sure what `MySpan` was used for, and although it's related to the `DUMMY_SP` -> `Span::call_site()` change, neither are dependent on each other and it can be removed in a separate PR.

c2rust-transpile/src/rust_ast/mod.rs
```diff
@@ -28,14 +28,6 @@ pub fn pos_to_span(pos: BytePos) -> Span {
     SpanExt::new(pos.0, pos.0)
 }
 
-pub const DUMMY_SP: Span = unsafe {
-    std::mem::transmute(SpanRepr {
-        compiler_or_fallback: 1,
-        lo: 0,
-        hi: 0,
-    })
-};
```
kkysen, line 37: `validate_repr` asserts that these are equivalent, so using the safe `Span::call_site` instead is preferrable.

kkysen (changes requested):
`validate_repr` asserts that `DUMMY_SP` and `Span::call_site()` are equivalent, so using the safe `Span::call_site()` instead is preferrable.  So this is definitely a change we should make, as it reduces the amount of UB we're using.  Thanks for fixing this!

There are a couple of other `clippy` warnings I've added suggestions for fixing; I think you may have just missed those.  And there's one place where still using `Self` instead of `Span` (reverting the change) is fine and makes more sense.

Other than that, I think the removal of `MySpan` should be separated into another PR, as it's not necessary and/or strictly tied to this PR, and as I'm not sure what `MySpan` was used for (as it wasn't marked as unused by the compiler, but was clearly not in active use), I think it'd better to separate that change.

c2rust-transpile/src/rust_ast/mod.rs
afetisov: Do you think I have nothing better to do with my life than slicing and dicing microPRs?

`MySpan` isn't used either in the code or in the tests, it isn't marked as unused because it's `pub` and has an impl of a pub trait. As far as the compiler is concerned, it is used.

kkysen: > Do you think I have nothing better to do with my life than slicing and dicing microPRs?

No, I don't.  Thank you for all these cleanups, but we don't exactly *need* them at the moment, and it's okay if you want to leave some of these things out if you want to focus on more-involved PRs.  Hopefully someone can eventually get around to these cleanups eventually, and thank you for doing a lot of them already for us, but they aren't the most pressing for us, as we are focused more on getting unsafe-to-safe lifting working, as well as bug fixes.  But importantly, we can't let any of these cleanups break things, so we have to be careful and conservative with them, especially if they aren't in parts of the code that are well-tested.

> `MySpan` isn't used either in the code or in the tests, it isn't marked as unused because it's `pub` and has an impl of a pub trait. As far as the compiler is concerned, it is used.

It could be used sometimes for debugging things, for example.  The point is, I'm not sure how it might be used, so I think it'd be better to wait and ask, as not deleting `MySpan` isn't blocking anything else.  @fw-immunant, who added this code, is now on vacation, so I think it'd be best to hold off on this part of the change.

kkysen: If you'd like, I can split this into another PR for you as this one's quite simple.  I'll force push to your branch to remove that commit and then cherry pick it to a new branch.

kkysen: Split into https://github.com/immunant/c2rust/pull/541.

kkysen: I forced push to remove the `MySpan` removal commit and I'll merge this PR now.

kkysen (approved):

PR #540, opened 2022-07-17
translator: refactor `simd.rs`; extract arch-dependent item use into a function

afetisov:


c2rust-transpile/src/translator/simd.rs
```diff
@@ -67,6 +67,23 @@ static SIMD_X86_64_ONLY: &[&str] = &[
...
+            mk().meta_list(
+                "cfg",
+                vec![NestedMeta::Meta(
+                    mk().meta_namevalue("target_arch", arch_name),
+                )],
+            ),
+        )
+        .pub_(),
+    );
+}
```
kkysen, line 85: I would add this as a method on `ItemStore` instead.  It's one call to `ItemStore::add_use_with_attr`, so it makes sense to add `ItemStore::add_arch_use` as well.

kkysen (changes requested):
I think it would make more sense to make `add_arch_use` as a method of `ItemStore` (see the comment).  Other than that, this looks great!  Thanks!

c2rust-transpile/src/translator/simd.rs
afetisov: That would be a leaky abstraction, since the code within is really specific to the simd intrinsics. It wouldn't be used outside of `simd.rs`, and it can't reliably be used since there is no guarantee that all subitems of `core::arch` exist at the top level rather than in some submodule `core:;arch::foobar`, or that architecture-dependent stuff always lives in `core::arch` and not e.g. `std::arch` or some other module.

kkysen: I was under the impression that

```rust
impl ItemStore {
    fn add_arch_use(&mut self, arch_name: &str, item_name: &str) { ... }
```

in `simd.rs` is private within `mod simd`.  I may be wrong about that, though, and if I am, then your free-function way is better, but if it can be made a method without leaking abstractions, I think we should.

afetisov: Hm, yes, that would work. I still don't think it should be done. It is quite confusing when inherent impls are smeared across many modules, doubly so when they aren't even a submodule of a common ancestor. It makes the code harder to navigate, even though I often rely on IDE. It also introduces some very unfortunate tight coupling between the modules. E.g. I can't split some subtree of modules into a different crate, since it would contain foreign inherent impls, and it's harder to understand the visibility of methods.

The free function does its job well and to the point, and doesn't cause any overarching issues for abstraction and architecture.

kkysen: Hmm, I don't think there's much advantage to either way, so we can just leave it as a free function for now.  If anyone else feels strongly later, it can be changed.

kkysen (approved):

PR #542, opened 2022-07-18
`cargo fix` on `c2rust-analyze`

kkysen:


spernsteiner:
I did this cleanup manually.  I normally check all the warnings myself because there's often a bug or two that `cargo fix` would miss.  In this case, there was a variable that was unused because I had used the wrong variable later in the function, along with some minor style issues like `cargo fix` suggesting `for (_i, x) in foo.iter().enumerate()` where `for x in foo.iter()` would make more sense.

PR #545, opened 2022-07-18
Improve `c2rust-analyze`'s `.gitignore`

kkysen:
Improve `c2rust-analyze`'s `.gitignore` to:
* ignore `*.rlib`, which are produced by `cargo test`
* don't redundantly ignore `/target/`, as that's already ignored by the root `.gitignore`


spernsteiner:
Cherry-picked ab2a96f5d9f1d1920031bec7e1ea89164b5f7b8d; the other change I already made locally.

Issue #522, opened 2022-07-14
`reallocarray` doesn't exist on macOS, but is used in `analysis/test/src/pointers.rs`


kkysen:
`reallocarray` doesn't exist on macOS, but is used in `analysis/test/src/pointers.rs`, causing `analysis/test` to not work on macOS.  I tried fixing this by simply implementing `reallocarray` in terms of `realloc`, but I think this causes two allocations to be registered, counting both `reallocarray` and the internal `realloc`.

I'm not sure what the best way to fix this is.

aneksteind:
I'm ok with either commenting it out with a note or separating it into its own test program. I would say we could conditionally compile it out but that would change the snapshot on OSX platforms and fail the build. We shouldn't let an involved fix for this block the streamlined testing in #483 IMO

kkysen:
I'm getting some weird results with `reallocarray` when implementing it in terms of `realloc`:

```diff
diff --git a/pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap b/pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
index e91ede1b..58ef52e6 100644
--- a/pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
+++ b/pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
@@ -171,7 +170,7 @@ g[/9qaCQ] {
 }
 nodes_that_need_write = []

-g[EvDfig] {
+g[Yw+erQ] {
        n[0]:  malloc(n = 1) _     => _11 @ bb5[2]:   fn exercise_allocator;
        n[1]:  copy          n[0]  => _1  @ bb6[3]:   fn exercise_allocator;
        n[2]:  copy          n[1]  => _19 @ bb6[7]:   fn exercise_allocator;
@@ -192,6 +191,8 @@ g[EvDfig] {
        n[17]: addr.load     n[16] => _   @ bb13[2]:  fn exercise_allocator;
        n[18]: copy          n[1]  => _43 @ bb21[7]:  fn exercise_allocator;
        n[19]: copy          n[18] => _42 @ bb21[8]:  fn exercise_allocator;
+       n[20]: copy          n[1]  => _6  @ bb0[7]:   fn reallocarray;
+       n[21]: free          n[20] => _0  @ bb1[3]:   fn reallocarray;
 }
[...]

PR #547, opened 2022-07-18
Polyfill `reallocarray`

kkysen:
Fixes https://github.com/immunant/c2rust/issues/522.

This polyfills `reallocarray` using `realloc`, as `reallocarray` doesn't exist on macOS.

The polyfilling is done through a `const` function pointer so that the instrumentation does not see the `realloc` and generate an extra `free` and `malloc` event for it (though the instrumentation probably should see it; when this breaks, we'll know instrumentation is smart enough to see through things like a `const` function pointer).

Normally we'd only polyfill it on macOS, but then we'd need a different snapshot file for macOS, as polyfilling results in a couple of extra copies.  Thus, we just polyfill always.

### DIFF
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -13,12 +13,22 @@ extern "C" {
     #[no_mangle]
     fn realloc(_: *mut libc::c_void, _: libc::c_ulong) -> *mut libc::c_void;
     #[no_mangle]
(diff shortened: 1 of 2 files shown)


kkysen:
Merging this would fix https://github.com/immunant/c2rust/issues/522 and thus allow https://github.com/immunant/c2rust/pull/483 to work and then merge, which would've been quite useful in catching some breakages.

aneksteind (approved):

PR #548, opened 2022-07-18
transpile: remove `trait LitStringable`, `impl Make<T> for &T`

afetisov:
The `LitStringable` `trait` is deprecated and doesn't do anything useful in the current codebase (it is only implemented for `&str`, which is much easier to take directly).

The `impl Make<T> for &T` is redundant and causes issues with type inference. It is also more clear and easier to optimize if the clones are explicit at use site.

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -166,12 +166,6 @@ impl<T> Make<T> for T {
     }
 }
 
-impl<'a, T: Clone> Make<T> for &'a T {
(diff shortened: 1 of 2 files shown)


kkysen (approved):
@fw-immunant already said these were good cleanups to make (https://github.com/immunant/c2rust/pull/485#issuecomment-1180605997).

I also added https://github.com/immunant/c2rust/pull/548/commits/4b3be46d4d8208605acd157e592fb07b01d981a0, which fixes the redundant clones surfaced by your refactors.

PR #549, opened 2022-07-18
Specialize builder functions where unambiguous

afetisov:
Remove generic parameters which are actually used only with a single concrete type, substituting that
type instead. 

E.g. a function `fn foo<E: Make<Box<Expr>>, T: Make<Box<Type>>(e: E, t: T)` is turned into a function
`fn foo(e: Box<Expr>, t: Box<Type>)`.

This doesn't restrict the usages of the function in any way, since the only type which is `Make<Box<Expr>>`
is `Box<Expr>` itself. These generic parameters are strictly redundant. They produce generic bloat, harm
compile times, complicate the API, and prevent type inference for function arguments (i.e. if `foo(f1(), f2())`
and either `f1` or `f2` is generic in its return type, then type inference will fail).

Using the concrete types is strictly better, and opens the possibility of more ergonomic API with better type
inference.

Also remove the explicit type casts which were required only due to the overly generic API, and can now be safely elided.

This PR was split from #485. As far as I can tell from the discussion in #485, this change is uncontroversial.

This PR builds on the branch of #548 and should be merged after it to avoid merge conflicts.

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -518,8 +518,7 @@ impl Builder {
         Builder { span, ..self }
     }
 
-    pub fn generic_over<P: Make<GenericParam>>(mut self, param: P) -> Self {
-        let param = param.make(&self);
+    pub fn generic_over(mut self, param: GenericParam) -> Self {
         self.generics.params.push(param);
         self
     }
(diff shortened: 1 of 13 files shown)


kkysen:
Rebased against `master` after merging https://github.com/immunant/c2rust/pull/548 to `master` so I can review a cleaner diff.

kkysen (approved):
As you noted, @fw-immunant already reviewed and approved these changes in https://github.com/immunant/c2rust/pull/485#issuecomment-1180605997, and I agree they are very helpful in reducing unnecessary generics and improving inference.

This does cause some `clippy::boxed_local` warnings, but I'll fix them separately, as it's not a tiny fix.

PR #550, opened 2022-07-19
Fixed `clippy::boxed_local` warnings surfaced by de-genericizing PR

kkysen:
Fixed the `clippy::boxed_local` warnings surfaced by the de-genericizing in https://github.com/immunant/c2rust/pull/549, as well as the cascading de-`Box`ing.  This gets us back to `clippy` warning-free code.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -369,12 +369,12 @@ fn vec_expr(val: Box<Expr>, count: Box<Expr>) -> Box<Expr> {
     mk().call_expr(from_elem, vec![val, count])
 }
 
-pub fn stmts_block(mut stmts: Vec<Stmt>) -> Box<Block> {
+pub fn stmts_block(mut stmts: Vec<Stmt>) -> Block {
     match stmts.pop() {
         None => {}
         Some(Stmt::Expr(Expr::Block(ExprBlock {
             block, label: None, ..
-        }))) if stmts.is_empty() => return Box::new(block),
+        }))) if stmts.is_empty() => return block,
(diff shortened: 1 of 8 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -3785,12 +3785,11 @@ impl<'c> Translation<'c> {
                                 Type::Tuple(TypeTuple { elems: ref v, .. }) if v.is_empty() => ReturnType::Default,
                                 _ => ReturnType::Type(Default::default(), ret_ty),
                             };
-                            let bare_ty: (Vec<BareFnArg>, Option<Variadic>, ReturnType) = (
+                            let bare_ty = (
                                 vec![mk().bare_arg(mk().infer_ty(), None::<Box<Ident>>); args.len()],
                                 None,
                                 ret_ty
                             );
```
fw-immunant, line 3792: The type annotation here previously served as documentation for what `None` was doing here (specifying that the `fn` is not variadic), so it would be nice to replace this type with a comment or type annotation on the `None`.

fw-immunant (approved):
One change (inline) would be nice for clarity, but otherwise good to go.

c2rust-transpile/src/translator/mod.rs
kkysen: Okay, I'll do `None::<Variadic>`.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/550/commits/d3da84c875c6ff1c2b93a17055457c090a12408a.

PR #552, opened 2022-07-20
Refactor dynamic analysis

aneksteind:
No functional changes, just a code refactor to simplify control flow etc.

### DIFF
--- a/dynamic_instrumentation/src/instrument_memory.rs
+++ b/dynamic_instrumentation/src/instrument_memory.rs
@@ -379,14 +379,12 @@ impl Source for u32 {
 
 trait InstrumentationState {}
 
-struct Uninitialized {}
 struct NeedsLoc {}
 struct ReadyToInstrument<'tcx> {
     point: InstrumentationPoint<'tcx>,
 }
 impl InstrumentationState for NeedsLoc {}
 impl<'tcx> InstrumentationState for ReadyToInstrument<'tcx> {}
-impl InstrumentationState for Uninitialized {}
 
 trait IntoOperand<'tcx> {
     fn op(self, tcx: TyCtxt<'tcx>) -> Operand<'tcx>;
--- a/dynamic_instrumentation/src/instrument_memory.rs
+++ b/dynamic_instrumentation/src/instrument_memory.rs
@@ -341,7 +341,7 @@ trait Source {
     fn source(&self) -> Option<MirPlace>;
 }
 
-impl<'tcx> Source for Place<'tcx> {
+impl Source for Place<'_> {
     fn source(&self) -> Option<MirPlace> {
         Some(to_mir_place(self))
     }
--- a/dynamic_instrumentation/src/instrument_memory.rs
+++ b/dynamic_instrumentation/src/instrument_memory.rs
@@ -53,6 +53,7 @@ impl Convert<mir_loc::DefPathHash> for DefPathHash {
     }
 }
 
+#[derive(Default)]
 pub struct InstrumentMemoryOps {
     mir_locs: Mutex<IndexSet<MirLoc>>,
     functions: Mutex<HashMap<mir_loc::DefPathHash, String>>,
@@ -61,14 +62,11 @@ pub struct InstrumentMemoryOps {
 impl InstrumentMemoryOps {
     /// Create a new instrumentation object.
     ///
-    /// A single `InstrumentMemoryOps` instance should be shared across the
+    /// A single [`InstrumentMemoryOps`] instance should be shared across the
     /// entire crate being instrumented, as the indexed source locations are
     /// shared and should be global.
     pub fn new() -> Self {
-        Self {
-            mir_locs: Mutex::new(IndexSet::new()),
-            functions: Mutex::new(HashMap::new()),
-        }
+        Self::default()
     }
 
     /// Instrument memory operations in-place in the function `body`.
--- a/dynamic_instrumentation/src/instrument_memory.rs
+++ b/dynamic_instrumentation/src/instrument_memory.rs
@@ -245,6 +245,11 @@ impl<'a, 'tcx: 'a> InstrumentationAdder<'a, 'tcx> {
         find_instrumentation_def(self.tcx, self.runtime_crate_did, name)
     }
 
+    fn find_hook(&self, name: &str) -> DefId {
+        self.find_instrumentation_def(Symbol::intern(name))
+            .unwrap_or_else(|| panic!("could not find `{name}` hook"))
+    }
+
     fn func_hash(&self) -> mir_loc::DefPathHash {
         self.tcx.def_path_hash(self.body.source.def_id()).convert()
     }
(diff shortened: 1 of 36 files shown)


dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -327,84 +300,164 @@ fn rv_place<'tcx>(rv: &'tcx Rvalue) -> Option<Place<'tcx>> {
...
+
+impl<'tcx> Source for Place<'tcx> {
+    fn source(&self) -> Option<MirPlace> {
+        Some(to_mir_place(self))
+    }
+}
+
+impl<'tcx> Source for Operand<'tcx> {
+    fn source(&self) -> Option<MirPlace> {
+        self.place().as_ref().map(to_mir_place)
```
kkysen, line 315:
```suggestion
        self.place().as_ref().and_then(Place::source)
```

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -327,84 +300,164 @@ fn rv_place<'tcx>(rv: &'tcx Rvalue) -> Option<Place<'tcx>> {
...
+    }
+}
+
+impl<'tcx> Source for Vec<Operand<'tcx>> {
+    fn source(&self) -> Option<MirPlace> {
+        // TODO: have hook-specific sources
+        self.first()
+            .and_then(|op| op.place().as_ref().map(to_mir_place))
```
kkysen, line 323:
```suggestion
        self.first().and_then(Operand::source)
```

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -327,84 +300,164 @@ fn rv_place<'tcx>(rv: &'tcx Rvalue) -> Option<Place<'tcx>> {
...
+    }
+}
+
+impl<'tcx> Source for Rvalue<'tcx> {
+    fn source(&self) -> Option<MirPlace> {
+        rv_place(self).as_ref().map(to_mir_place)
```
kkysen, line 329:
```suggestion
        rv_place(self).as_ref().and_then(Place::source)
```

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -327,84 +300,164 @@ fn rv_place<'tcx>(rv: &'tcx Rvalue) -> Option<Place<'tcx>> {
...
+    }
+}
+
+impl Source for u32 {
+    fn source(&self) -> Option<MirPlace> {
+        Place::from(Local::from_u32(*self)).source()
+    }
+}
```
kkysen, line 337:
```suggestion
impl Source for Local {
    fn source(&self) -> Option<MirPlace> {
        Place::from(*self).source()
    }
}

impl Source for u32 {
    fn source(&self) -> Option<MirPlace> {
        Local::from_u32(*self).source()
    }
}
```

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -690,138 +678,85 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
+                                        &place_ty,
+                                    ))
+                                    .source(&place)
+                                    .dest(&callee_arg)
+                                    .transfer(transfer_kind)
+                                    .add(self);
                             }
                         }
-                        arg_local.index += 1;
+                        callee_arg = Local::from(callee_arg.local.as_u32() + 1).into();
```
kkysen, line 703:
```suggestion
                        callee_arg.local.increment_by(1);
```

Requires `use rustc_index::vec::Idx` (separate suggestion).

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -681,7 +669,7 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
                 destination,
                 ..
             } => {
-                let mut arg_local = mir_loc::Local { index: 1 };
+                let mut callee_arg: Place = Local::from_u32(1).into();
```
kkysen, line 672:
```suggestion
                let mut callee_arg: Place = Local::new(1).into();
```

Requires `use rustc_index::vec::Idx` (separate suggestion).

I think it makes more sense to use the `Idx` methods, as we're using this as an index and we don't have to hardcode the `u32` here.  The `Idx::increment_by` method makes more sense than the previous code, so I think it'd be better if we stick to the `Idx` methods for `callee_arg`.

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -6,8 +6,10 @@ use c2rust_analysis_rt::mir_loc::{
 use c2rust_analysis_rt::HOOK_FUNCTIONS;
 use indexmap::IndexSet;
 use log::debug;
+use rustc_ast::ptr::P;
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_index::vec::IndexVec;
```
kkysen, line 11:
```suggestion
use rustc_index::vec::{Idx, IndexVec};
```

2 other suggestions need this.

kkysen (changes requested):
Overall this looks fine; there's just some small suggestions I made, as well as a slightly larger one in a PR (https://github.com/immunant/c2rust/pull/555).

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -426,240 +479,175 @@ impl<'a, 'tcx: 'a> Visitor<'tcx> for CollectFunctionInstrumentationPoints<'a, 't
...
+            {
+                if is_region_or_unsafe_ptr(place_ty(p)) {
+                    InstrumentationPoint::new(location, ptr_to_int_fn)
+                        .arg(InstrumentationArg::from_type(
+                            Operand::Copy(p.local.into()),
+                            &local_ty(p),
+                        ))
+                        .source(p)
+                        .add(self);
                 }
```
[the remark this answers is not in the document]
aneksteind: This is technically a functional change, because if the first two clauses are true but the last is false, then the `Rvalue::Cast(_, op, _)` expression that's matched later on will get hit instead of it happening here. It doesn't make a difference in the snapshot currently, but I want to avoid making that change in this PR.

kkysen (changes requested):
## Builder Questions

Could you explain the changes to the builder?
* What is `InstrumentationState`?
* Why is `InstrumentationBuilder` now generic over the state `S`?
* What is `NeedsLoc`?
## `arg_var` and `IntoOperand` uses

For `InstrumentationBuilder::arg_var`, many of its uses use a part of a type that already `impl`s `IntoOperand`, such as `Place` (`place.local`) and `Local` (`local.as_u32()`):

```shell
dynamic_instrumentation/src/instrument_memory.rs
520:                        .arg_var(place.local)
521:                        .arg_var(field.as_u32())
561:                .arg_var(p.local)
601:                        .arg_var(p.local)
612:                    .arg_var(p.local.as_u32())
673:                        .arg_var(p.local.as_u32())
682:                        .arg_var(p.local.as_u32())
```

Are these redundant, and if not, why don't the `impl IntoOperand`s cover this?  If they don't, what's the difference between the `impl IntoOperand` behavior and the expressions passed to `.arg_var`?

## Builder Improvements

Also, I think the builder around `InstrumentationPoint` can be cleaned up into a separate builder type.  That is, `InstrumentationPoint` should be the final stored type, but there should be an `impl Default` type that is used in the builder, as `InstrumentationPoint::{id,loc,func}` aren't used until the end.

If we have

```rust
#[derive(Default)]
struct InstrumentationPointBuilder<'tcx> {
	args: Vec<InstrumentationArg<'tcx>>,
	is_cleanup: bool,
	after_call: bool,
	metadata: EventMetadata,
}
```

that gets transformed into a full `InstrumentationPoint` with the `loc`, `func`, and `id` fields in `add_to`, I think that would be a lot cleaner.  This way you also have the option of doing this, for example:

```rust
InstrumentationPointBuilder {
	args: vec![arg],
	..Default::default()
}
```

I can do that in PR if you'd like.

## Other Changes

Other than that, I have some more minor suggestions, as well as a PR (https://github.com/immunant/c2rust/pull/557) of other improvements that are a bit bigger/don't fit into a single suggestion.  A couple of the changes are unrelated to your existing changes, but since this is a refactoring PR, I thought this is a good place to put them.

Also, this is merging into `feat.pdg.ref`, not `master`?

kkysen (approved):

dynamic_instrumentation/src/instrument_memory.rs
```diff
@@ -180,29 +191,47 @@ struct InstrumentationPoint<'tcx> {
     metadata: EventMetadata,
 }
 
-struct CollectFunctionInstrumentationPoints<'a, 'tcx: 'a> {
+struct InstrumentationAdder<'a, 'tcx: 'a> {
```
fw-immunant, line 199: NAK on this rename. This visitor collects a set of instrumentation points that we want to add to the program. It doesn't itself add instrumentation to the program, and the distinction between "adding" and "applying" instrumentation points is not clear to a reader.

kkysen: I do think it should be a noun not a verb, though.

fw-immunant: As verbose as it is, I like `CollectInstrumentationPointsVisitor`--mentioning that it is a visitor explains why the struct exists in the first place (to hang the Visitor impl and intermediate state off of), and the "CollectInstrumentationPoints" portion explains what its specific task is.

Issue #488, opened 2022-07-07
`libiconv` necessary on mac, doesn't link


boyland-pf:
The mac version of c2rust with dynamic instrumentation enabled only requires libiconv at some point in the build, resulting in the following error.

```shell
  = note: Undefined symbols for architecture x86_64:
            "_libiconv", referenced from:
                _git_fs_path_iconv in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
            "_libiconv_open", referenced from:
                _git_fs_path_direach in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
                _git_fs_path_iconv_init_precompose in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
                _git_fs_path_diriter_init in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
            "_libiconv_close", referenced from:
                _git_fs_path_direach in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
                _git_fs_path_iconv_clear in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
                _git_fs_path_diriter_free in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
                _git_fs_path_dirload in liblibgit2_sys-4e72aab137059590.rlib(fs_path.o)
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This happens on my mac (a Catalina), which has iconv, and won't link libiconv (installed via brew) because it already has iconv which does the same thing.

This post on stackoverflow (https://stackoverflow.com/questions/57734434/libiconv-or-iconv-undefined-symbol-on-mac-osx) has the best description of the problem that I've found.

Because the compilation is handled by cargo, I don't know how to find a list of all the compilation steps to add `--with-libiconv-prefix=<dir>` to, or whether this would work.

fw-immunant:
This seems to impact us because this crate depends on the `cargo` crate which (transitively) requires the `libgit2-sys` crate. There are [a number of upstream reports](https://github.com/rust-lang/git2-rs/search?q=iconv&type=issues) related to this issue, but this recent one seems potentially helpful: https://github.com/rust-lang/git2-rs/issues/830

[...]

Issue #448, opened 2022-06-15
`c2rust instrument` should apply extra `cargo` args


kkysen:
Currently, `c2rust instrument` takes extra `cargo` args to apply when instrumenting a crate it's building.  However, in `c2rust_dynamic_instrumentation::instrument`, these args are unused, and the compilation is a default `cargo` debug build.  For example, if you try to do a release build with `--release`, you can't.  Running the instrumented binary in release mode seems like a fairly important thing to be able to do for performance reasons.  The issue goes a lot deeper than just not using these extra arguments, however.

In `c2rust_dynamic_instrumentation`, we use the [`cargo`](https://docs.rs/cargo/) crate and its APIs directly.  There are APIs for replicating the passed `cargo` args, such as [`BuildConfig::requested_profile`](https://docs.rs/cargo/latest/cargo/core/compiler/struct.BuildConfig.html#structfield.requested_profile).  However, all of `cargo`'s logic for parsing the args, applying them, and creating a [`CompileOptions`](https://docs.rs/cargo/latest/cargo/ops/struct.CompileOptions.html) are in [`src/bin/cargo/`](https://github.com/rust-lang/cargo/tree/master/src/bin/cargo).  That is, they are within `cargo`'s binary crates, and as such, are private, so we cannot use them at all.  So if we want to support extra `cargo` args in the same way `cargo` handles them, we would have to duplicate `cargo`'s arg parsing code, and we would have to always keep it in sync with changes to `cargo`.  This seems quite brittle.

[...]

Issue #395, opened 2022-05-10
Update to clap3


rinon:
Clap3 deprecated YAML support, so we need to rewrite command-line parsing to use the new APIs.

kkysen:
Did `clap` 3 actually deprecate yaml support?  #702 suggests it didn't, and from `clap`'s documentation (v3 and v4) it seems like yaml is still supported.  If that's the case, then we can just upgrade `clap` to 3 and even 4 without changes.  I think using the derive API is a lot cleaner and type-safe, but that can be done separately from updating the `clap` dependency.

rinon:
deprecated != removed (https://github.com/clap-rs/clap/issues/3087). This is very low priority though.

kkysen:
Ah, okay.  I missed that.  Though can't we still upgrade to `clap` 3, though, so we only have one `clap` version and faster builds.  Then leave this issue for moving away from the yaml API?  I get that it's not very important, though, but #702 was trying to do it but without removing the yaml.

marcograss:
#703 now

fw-immunant:
Fixed by #703.

kkysen:
Not really fixed by #703.  Actually will be fixed by #777, so we'll close this when that merges.

PR #554, opened 2022-07-22
Instrument using `$RUSTC_WRAPPER` and `cargo` as a subcommand

kkysen:
Fixes https://github.com/immunant/c2rust/issues/448 and should fix https://github.com/immunant/c2rust/issues/488.

## Overview

This reimplements `c2rust-instrument` by invoking `cargo` as a subcommand, overriding `$RUSTC_WRAPPER`.

## Implementation

More specifically, instead of using `cargo` as a dependency, which is enormous (very long compile times) and causes some compilation issues (https://github.com/immunant/c2rust/issues/488), we invoke `cargo` as a subcommand.  This also has the advantage of using the correct `rustup` `cargo` wrapper so that it resolves correctly.  See https://github.com/immunant/c2rust/issues/448 for a more detailed explanation of why this is a better approach and why other tools like `clippy` and `miri` do it this way.

Then we set `$RUSTC_WRAPPER` to our own binary.  We use `$RUSTC_WRAPPER` to detect if we're being invoked as a `cargo` or `rustc` wrapper.

If we're the `cargo` wrapper, we:
* parse args using `clap` derive (https://github.com/immunant/c2rust/issues/395)
* run `cargo` as a subcommand, passing:
    * the `cargo` args as is (so all normal `cargo` invocations all work)
    * our own executable as `$RUSTC_WRAPPER`
    * the metadata path in `$C2RUST_INSTRUMENT_METADATA`

If we're the `rustc` wrapper, we:
* determine if we should instrument the current `rustc` command if `$CARGO_PRIMARY_PACKAGE` is set
* run `RunCompiler::new(...).run()` with our `MirTransformCallbacks`, or `TimePassesCallbacks::default()` (`rustc`'s default) if not instrumenting
* then save the metadata to the metadata file

Note that we have to pass the `--sysroot` to `RunCompiler` as normally `rustc` looks this up relative to its executable's location, which works when the `rustup` `rustc` wrapper resolves to the toolchain-specific `rustc`, but since we're using `RunCompiler` directly and being `rustc`, we need to explicitly set the sysroot.

## Usage Changes

Instead of passing `c2rust_analysis_rt` as an `--extern` and running `cargo` separately to compile it, we just add it as a normal, optional dependency to the crate we're instrumenting.  In this way, it's similar to `libc`, which can be `extern`ed as it's in the sysroot,
but the preferred way is to specify it as a normal dependency.  Thus, we can also drop the runtime argument that seemed odd to have to specify to `c2rust instrument`.

The existing `c2rust-instrument` binary in the `c2rust` package is now gone (along with its `--features dynamic-instrumentation`), replaced by the new `c2rust-instrument` binary in `dynamic-instrumentation/src/bin.rs` (the crate has also been renamed to `c2rust-instrument`).  Now, `c2rust_instrument` no longer exposes a library crate, and is purely a binary crate.  This is because there's no way to use `c2rust-instrument` not as a binary, as it invokes itself and thus needs control over `main` to function correctly.

This does cause a slight change in that `c2rust` no longer sees the `instrument.yaml`; only `c2rust-instrument` does.  This could potentially hurt shell autocompletion in the future (not currently implemented), but the drawbacks seems slim.  Instead, we don't need to remember `--features dynamic-instrumentation` as a simple `cargo build` builds everything.  And we've upgraded to a superior `clap` 3 derive parser, not the deprecated YAML parser.

## Fast!

Furthermore, now that we use `cargo` normally and include `c2rust_analysis_rt` as a normal dependency, we can delete only the main binary (with `cargo clean --package`), forcing it to be rebuilt and re-instrumented.  Most of the time building used to be in building `c2rust_analysis_rt`, and on the fast `donna`, was ~15 seconds for `c2rust instrument`.  Now it takes only ~0.5 seconds, 30x faster.

Also enabling this is using a separate `$CARGO_TARGET_DIR`: `instrument.target` instead of the default `target`.  This allows builds not to conflict with normal, non-instrumented builds using `cargo build`, so we don't need to `cargo clean` everything every time.

Thus, we now also unconditionally instrument (in `pdg.sh`).  Before, we only did when the `c2rust-instrument` binary changed, but this missed changes to the instrumented crate.  Now that it's fast enough, we just always instrument.

Previously, I was working on getting incremental instrumentation to work, but that proved harder than I thought for the corner cases (which did pop up).  With this change, we remove the largest block to fast iteration times, so I think we don't need incremental instrumentation at all until we get to iterating on small edits in large instrumented crates.

## Fixes and Improvements

* https://github.com/immunant/c2rust/issues/448: the primary fix.  Now we can specify any `cargo` args, such as `build`, `run`, `test`, `--release`, etc.
* https://github.com/immunant/c2rust/issues/488: should be fixed as the offending `cargo` dependency is removed, but needs to be tested by @boyland-pf.
* We can remove `pretty-instrument-err.mjs` and `get-binary-names-from-cargo-metadata.mjs`, the temporary `node` scripts.   
 Thus, we can remove all `node` dependencies.
* With enough polishing, we may be able to entirely get rid of `pdg.sh`.
* The `cargo` dependency is dropped, saving a lot on compile time, as it is a huge dependency.
* The `--features dynamic-instrumentation` is removed, so a simple `cargo build` works for everything now.


dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+    if wrapping_rustc {
+        let sysroot = get_sysroot_fast()
+            .ok_or(())
+            .or_else(|()| get_sysroot_slow())?;
+        let mut at_args = env::args().skip(1).collect::<Vec<_>>();
+        let crate_target = CrateTarget::from_rustc_args(&at_args)?;
+        let info = env::var(instrument_info_var)?;
+        let info = serde_json::from_str::<InstrumentInfo>(&info)?;
+        let should_instrument = info.crate_targets.contains(&crate_target);
+        if !should_instrument {
```
aneksteind, line 283: nit: put the `should_instrument` path first for style

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -77,12 +77,12 @@ g {
 nodes_that_need_write = []
 
 g {
-	n[0]: copy _ => _46 @ bb31[6]: fn main;
+	n[0]: copy _ => _45 @ bb30[7]: fn main;
```
aneksteind, line 80: why has the snapshot changed?

scripts/get-binary-names-from-cargo-metadata.mjs
```diff
@@ -1,43 +0,0 @@
-#!/usr/bin/env node
```
aneksteind, line 1: does this mean we can get rid of the node dependency?

aneksteind: ah, i finally made it to the bottom of your description -- is that something that we remove in another PR?

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
aneksteind: i see you mentioned this in your description as well -- i'd like to get to the bottom of that

aneksteind: my initial guess is that something has changed in the way we compile the program we want to instrument, just need to identify what that is if it's the case

scripts/get-binary-names-from-cargo-metadata.mjs
kkysen: The `node` dependency was implicit, so it's gone once we stop using it (as done here).

dynamic_instrumentation/src/main.rs
kkysen: I did it this way since I was going to skip the `else`, as this first branch is shorter and ends with an `exit_with_status`, which returns `!`, but that's unstable, so I can't mark it as such.  I'm probably gonna end up refactoring things more so this might change anyways.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
kkysen: It seems like most of the changes are moved one `Local` and one basic block back.  I'm still not sure what I changed to cause these changes.  Perhaps `c2rust_analysis_rt` being a normal dependency and not an `extern`?  That might change the order of crate compilation, as previously we forced `c2rust_analysis_rt` to be compiled first, and now `cargo` just figures out the order it wants.

dynamic_instrumentation/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/554/commits/b6c98aca1e19cb5e03f9b071dd0ce13256a36c82.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
kkysen: The difference is that now the `--release` arg is now honored, and in the release build, some things are optimized out, resulting in the slightly lower local and basic block numbers.  Now I've switched back to explicitly specifying `PROFILE=debug` so that there's no diff (https://github.com/immunant/c2rust/pull/554/commits/0a46ad82bb4b4fc0a237ad33ddde3adc6f156bc9).  Going forward, I can add a `PROFILE=release` one, too, to track both.

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+    path::{Path, PathBuf},
+    process::{self, Command, ExitStatus},
+    str::FromStr,
+};
+
+use c2rust_dynamic_instrumentation::{MirTransformCallbacks, INSTRUMENTER};
+use camino::Utf8Path;
+use cargo_metadata::{Metadata, MetadataCommand, Package, Target};
+use clap::Parser;
+use color_eyre::eyre;
```
rinon, line 20: I don't think we should be mixing anyhow and eyre just to get console colors. If we want to switch to eyre from anyhow, that should be a separate change.

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+    // not at compile-time in the `build.rs`,
+    // as the toolchain locations could be different
+    // from where this binary was compiled and where it is running
+    // (it could be on a different machine with a different `$RUSTUP_HOME`).
+    let sysroot: &Utf8Path = sysroot.try_into()?;
+    at_args.extend(["--sysroot".into(), sysroot.as_str().into()]);
+    RunCompiler::new(&at_args, &mut MirTransformCallbacks)
+        .run()
+        .map_err(|_| eyre!("`rustc` failed"))?;
+    INSTRUMENTER.finalize(metadata).map_err(|e| eyre!(e))?;
```
rinon, line 300: You shouldn't need to `map_err()` here, `?` should be able to convert anything that implements the `Error` trait.

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    let rustc_wrapper_var = "RUSTC_WRAPPER";
+    let instrument_info_var = "C2RUST_INSTRUMENT_INFO";
+
+    let own_exe = env::current_exe()?;
+
+    let wrapping_rustc = env::var_os(rustc_wrapper_var).as_deref() == Some(own_exe.as_os_str());
+    if wrapping_rustc {
+        let sysroot = get_sysroot()?;
+        let at_args = env::args().skip(1).collect::<Vec<_>>();
```
rinon, line 374: Cargo passes rustc as the first arg. Is that the full path to rustc? If so, could we use that to get the sysroot and not rely on rustup at all?

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+
+        // We could binary encode this, but it's likely very short,
+        // so just json encode it, so it's also human readable and inspectable.
+        let info = serde_json::to_string(&info)?;
+
+        let cargo = env::var_os("CARGO").unwrap_or_else(|| "cargo".into());
+        let status = Command::new(cargo)
+            .args(cargo_args)
+            .env(rustc_wrapper_var, &own_exe)
+            .env(instrument_info_var, info)
```
rinon, line 408: Environment variables have max lengths, so I'm concerned about this string being too long to fit here. Ideally we can avoid passing this information entirely.

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+        } = Args::parse();
+
+        let cargo_metadata = MetadataCommand::new().exec()?;
+        let crate_targets = CrateTarget::from_metadata(&cargo_metadata)?;
+        let info = InstrumentInfo {
+            crate_targets,
+            metadata,
+        };
+
+        delete_metadata_and_dependencies(&info, &cargo_metadata)?;
```
rinon, line 398: Instead of deleting files ourselves, could you use `cargo clean --package crate_name` instead?

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+    let own_exe = env::current_exe()?;
+
+    let wrapping_rustc = env::var_os(rustc_wrapper_var).as_deref() == Some(own_exe.as_os_str());
+    if wrapping_rustc {
+        let sysroot = get_sysroot()?;
+        let at_args = env::args().skip(1).collect::<Vec<_>>();
+        let crate_target = CrateTarget::from_rustc_args(&at_args)?;
+        let info = env::var(instrument_info_var)?;
+        let info = serde_json::from_str::<InstrumentInfo>(&info)?;
+        let should_instrument = info.crate_targets.contains(&crate_target);
```
rinon, line 378: Cargo sets `CARGO_PRIMARY_PACKAGE` when compiling the primary package (i.e. not dependencies). Could we use this instead of passing info and parsing the command line?

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+fn get_sysroot() -> eyre::Result<PathBuf> {
+    get_sysroot_fast()
+        .ok_or(())
+        .or_else(|()| get_sysroot_slow())
+}
+
+fn passthrough_rustc(at_args: &[String], sysroot: &Path) -> eyre::Result<ExitStatus> {
+    // We can skip a `rustup` `rustc` -> real `rustc` resolution by just invoking the real `rustc` ourselves.
+    // TODO(kkysen) this might not work without `rustup`
+    let rustc = sysroot.join("bin/rustc");
```
rinon, line 273: See other comment, but if cargo passes us a valid rustc path we should just use that rather than try to find the right one ourselves.

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+    let status = Command::new(rustc).args(at_args.iter().skip(1)).status()?;
+    Ok(status)
+}
+
+fn instrument_rustc(mut at_args: Vec<String>, sysroot: &Path, metadata: &Path) -> eyre::Result<()> {
```
rinon, line 278: This should be moved into the library so we can avoid exposing `MirTransformCallbacks` and `INSTRUMENTER` as a public API.

rinon (changes requested):
Overall, I think this is the right direction. Hopefully we can fix this up to not require serializing and passing the cargo target metadata to the subprocess, that's pretty cumbersome if it can be avoided (see comments).

I do have one major concern with this approach... Previously we would rebuild everything using whatever version of rustc was embedded in the dynamic instrumenter. Now, we rely on the `rustc` that this wrapper calls to be identical to (or at least closely compatible with) the version we linked against. We should enforce this and warn if there is a mismatch.

analysis/test/Cargo.toml
```diff
@@ -7,3 +7,4 @@ edition = "2021"
 
 [dependencies]
 libc = "0.2"
+c2rust-analysis-rt = { path = "../../analysis/runtime"}
```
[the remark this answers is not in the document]
kkysen: One idea I had was to make it an optional dependency, and then we can only enable that feature when instrumenting.  I think that's definitely preferable to the current solution.  We could also `cargo add` that dependency ourselves so that the user doesn't have to do it manually.  Another solution I was thinking of was to copy the existing `Cargo.toml`, add the `c2rust-analysis-rt` dependency to the copy, and then specify it with `--manifest-path` (I don't see an environment variable for that), though not quite all `cargo` commands allow `--manifest-path` (though I think all the ones we care about do).

dynamic_instrumentation/src/main.rs
```diff
@@ -0,0 +1,413 @@
...
+    // Also note that this sysroot lookup should be done at runtime,
+    // not at compile-time in the `build.rs`,
+    // as the toolchain locations could be different
+    // from where this binary was compiled and where it is running
+    // (it could be on a different machine with a different `$RUSTUP_HOME`).
+    let sysroot: &Utf8Path = sysroot.try_into()?;
+    at_args.extend(["--sysroot".into(), sysroot.as_str().into()]);
+    RunCompiler::new(&at_args, &mut MirTransformCallbacks)
+        .run()
+        .map_err(|_| eyre!("`rustc` failed"))?;
```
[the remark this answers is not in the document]
kkysen: The error returned by `RunCompiler::run` is `ErrorGuaranteed`, and it doesn't `impl Error` as it means the error(s) have already been reported to the user, so the only thing we need to do is exit with a non-zero exit code.  This is how it was handled previously, too, though with `anyhow`.

dynamic_instrumentation/src/main.rs
kkysen: Okay, I can switch things back to `anyhow` for now and do the change to `eyre` separately.  `eyre` I find useful mostly for the backtraces, though the colors are also very nice.  It's a shame it's not compatible with `anyhow`, as it is a fork of `anyhow` and very similar.

dynamic_instrumentation/src/main.rs
kkysen: Yeah, that was due to the `anyhow` to `eyre` change (it didn't work without this conversion), but I can revert back to `anyhow` and later change to `eyre`, but change throughout the whole crate so there are no weird conversions like this.

dynamic_instrumentation/src/main.rs
kkysen: Unfortunately, it's not, or at least, it's not usually (it probably can be).  It's normally just `rustc`, and `$RUSTC` is only set to `rustc`, too (I think there was a change to this in the latest `rustup` ([`CHANGELOG.md`](https://github.com/rust-lang/rustup/blob/master/CHANGELOG.md#1251---2022-07-12)) where the full path was specified in `$RUSTC` as an optimization, but it got reverted due to some people depended on it).

kkysen: Note that we can look up the sysroot without explicitly depending on `rustup` with `rustc --print sysroot` (I haven't implemented that part yet), the current way just uses a `rustup`-dependent fast-path.

dynamic_instrumentation/src/main.rs
kkysen: I think in practice it should be fine, because we only pass the metadata path and the targets of the main package (i.e. the `lib`, `bin`, etc. crates), which there usually aren't very many of.  But if there's a way we can avoid this, that'd be better, though doing the `cargo metadata` lookup in each `rustc` invocation is not the best option, as that repeats a lot of work.

dynamic_instrumentation/src/main.rs
kkysen: Ah, I didn't know that's a thing.  In that case, that should work perfectly.  Thanks!

dynamic_instrumentation/src/main.rs
kkysen: I don't remember seeing this environment variable (I dumped and looked through all of the ones passed to the `rustc` wrapper), but I can check.  If it is, I think we can use it instead of the current way I have of checking if we should instrument.

dynamic_instrumentation/src/main.rs
kkysen: Unfortunately not.  We can use `rustc --print sysroot`, though that's slower (subprocess) and harder to get right with non-UTF-8 paths.

dynamic_instrumentation/src/main.rs
kkysen: I think that can be done, but I was thinking we shouldn't have a `lib.rs` for this crate at all, as it doesn't make sense to use it when we don't have control of the binary and `main`.

kkysen:
> Overall, I think this is the right direction. Hopefully we can fix this up to not require serializing and passing the cargo target metadata to the subprocess, that's pretty cumbersome if it can be avoided (see comments).

If the `CARGO_PRIMARY_PACKAGE` way works, that'd be great, and we can get rid of all that code (just need to pass the metadata path then).

> I do have one major concern with this approach... Previously we would rebuild everything using whatever version of rustc was embedded in the dynamic instrumenter. Now, we rely on the `rustc` that this wrapper calls to be identical to (or at least closely compatible with) the version we linked against. We should enforce this and warn if there is a mismatch.

That's a good point.  One thing to note, though, is that we *must* depend on the system's `rustc` sysroot, as we don't ship one.  So I think either way, we need `rustc` versions to match.  I could change the code to use `RunCompiler` with `NullCallbacks`, instead, but we'll have the `rustc` version issue whichever way we do things.  I think we should also set `$RUSTUP_TOOLCHAIN`, but still also check versions.  Thinking about it more, I think if we use `RunCompiler` with `NullCallbacks`, the hashes in dynamic libraries should enforce the version check.  The reason I used (the real) `rustc` as a subcommand instead was so that we get exactly correct handling, as there's a bit more code in the `rustc` binary than just `RunCompiler::new(...).run()`.

This is also why I think that, in hindsight, the rpath embedding is suboptimal, as it embeds an absolute path in a binary that should be portable.  It's not the worst, as we just get the old behavior if that absolute rpath doesn't exist, and if it does exist, it's highly likely it's correct, because it's a toolchain/version-specific path, and the dynamic libraries include their hash.  Instead, I think it's better that all `c2rust` binaries that operate on a Rust crate and `rustc` private crates (i.e. `c2rust-instrument`, `c2rust-pdg`, `c2rust-analyze`, but not `c2rust-transpile`), should be run through `cargo` as a `cargo-*` plugin.

rinon:
> > I do have one major concern with this approach... Previously we would rebuild everything using whatever version of rustc was embedded in the dynamic instrumenter. Now, we rely on the `rustc` that this wrapper calls to be identical to (or at least closely compatible with) the version we linked against. We should enforce this and warn if there is a mismatch.
> 
> That's a good point. One thing to note, though, is that we _must_ depend on the system's `rustc` sysroot, as we don't ship one. So I think either way, we need `rustc` versions to match. I could change the code to use `RunCompiler` with `NullCallbacks`, instead, but we'll have the `rustc` version issue whichever way we do things. I think we should also set `$RUSTUP_TOOLCHAIN`, but still also check versions. Thinking about it more, I think if we use `RunCompiler` with `NullCallbacks`, the hashes in dynamic libraries should enforce the version check. The reason I used (the real) `rustc` as a subcommand instead was so that we get exactly correct handling, as there's a bit more code in the `rustc` binary than just `RunCompiler::new(...).run()`.

I was more thinking about the situation where we compile, e.g. the target crate with the stable compiler (so all its deps are built with the stable compiler), then we `cargo clean` just the primary crate, then rebuild with the nightly we linked against. The dependencies won't match up with the primary crate. That was why I rebuilt all crates every time - to avoid this mismatch. We can do better, but we need to make sure that cargo is aware that the compiler version has changed. I think it checks this already, but we will just pass that through to the `rustc` binary, right? I think we should validate that `rustc` and our wrapper report the same version.

kkysen:
Hmm, that's a good point and more likely to happen (though usually we need `nightly` for the transpiled crate).  How about just having instrumentation use a different `$CARGO_TARGET_DIR`?

kkysen:
I had thought crates were rebuilt so that up-to-date metadata would be created.

dynamic_instrumentation/src/main.rs
kkysen: `$CARGO_PRIMARY_PACKAGE` works, which allows things to be simplified a ton.  Thanks!

Fixed in:
* https://github.com/immunant/c2rust/pull/554/commits/a1fdf172fd7abaaf387c69a949748d68f2c22cba
* https://github.com/immunant/c2rust/pull/554/commits/d28775ae0f5a1404ee8f24e2d570e3a37d25fc6c
* https://github.com/immunant/c2rust/pull/554/commits/8c5568d3e1d2c0a2136306a0a81df4dd11663e89

dynamic_instrumentation/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/554/commits/035670ec7b8eb95592ce9f28aed720956e583b8d.  That's a lot cleaner and reliable, too!

dynamic_instrumentation/src/main.rs
kkysen: Not an issue anymore, as now we only pass the metadata path, which shouldn't (can't?) be too long.

dynamic_instrumentation/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/554/commits/d8c0b4922fc5f7a38e3a81381eb83bf4fd8afab1 after reverting back to `anyhow`.

dynamic_instrumentation/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/554/commits/045c62d7ada40b73fa67aa6521e8d62dc1185073.

kkysen:
I've updated things to use `RunCompiler::new(...).run()` instead of invoking an external `rustc` so that we don't get version mismatches (in a single compiler) (https://github.com/immunant/c2rust/pull/554/commits/c035e75e88465371541e2398436a050abee714c5).  Also, this now uses `TimePassesCallbacks::default()` instead of `NullCallbacks`, as that's what `rustc` itself uses and it already exists so we don't have to define `NullCallbacks` (https://github.com/immunant/c2rust/pull/554/commits/8e9e94e7e1742983a38343925a1f847aff1a7ba9).

kkysen:
I also switched to using `$CARGO_PRIMARY_PACKAGE` instead of passing the crate target info, which allows me to get rid of all that code.  Now only the metadata path is passed as its own environment variable, so we don't need `serde` either.

I also switched back to `anyhow` from `eyre`; we can fix that in a later change that changes it everywhere if we want (I do think it's a lot better for debugging).

I also am now using `cargo clean --package` instead of doing it manually, which is a lot more reliable (a bit slower, but I think it's worth it to do it right).  However, we still need `cargo metadata` for this to resolve the package name, as we need to run this in the `cargo` wrapper, not the `rustc` wrapper (or else the `rustc` wrapper won't even be run).

dynamic_instrumentation/src/main.rs
kkysen: I don't think there's a difference between exposing `MirTransformCallbacks` and `fn instrument`; neither should be called.

kkysen:
I just added new changes that:
* https://github.com/immunant/c2rust/pull/554/commits/0ab0a3084626c238f20227e459ededb1d6c3fa78: removes `cargo` as a dependency
* https://github.com/immunant/c2rust/pull/554/commits/0ab0a3084626c238f20227e459ededb1d6c3fa78: removes `c2rust-instrument` as a binary from the `c2rust` package
* https://github.com/immunant/c2rust/pull/554/commits/e32281cf9a178f29fd4d9aa2df87a62da1695bf3: replaces `c2rust-instrument` by renaming `c2rust-dynamic-instrumentation` to `c2rust-instrument` so that the `bin.rs` is now `c2rust-instrument`
* https://github.com/immunant/c2rust/pull/554/commits/676dee2df9e3e2dc1151a1f7c333b69f4d9a191f: removes `dynamic-instrumentation/src/lib.rs` entirely to make the new `c2rust-instrument` package have only a binary crate so that it cannot be used without going through the `c2rust-instrument` binary

kkysen:
> I was more thinking about the situation where we compile, e.g. the target crate with the stable compiler (so all its deps are built with the stable compiler), then we `cargo clean` just the primary crate, then rebuild with the nightly we linked against. The dependencies won't match up with the primary crate. That was why I rebuilt all crates every time - to avoid this mismatch. We can do better, but we need to make sure that cargo is aware that the compiler version has changed. I think it checks this already, but we will just pass that through to the `rustc` binary, right? I think we should validate that `rustc` and our wrapper report the same version.

Now `CARGO_TARGET_DIR=instrument.target` is set, so nothing conflicts with normal `cargo build`s (https://github.com/immunant/c2rust/pull/554/commits/6595ad8503db4fcf6e17f3d104ebbd92825b4753).  I also tried `CARGO_TARGET_DIR=target/instrument`, but then certain commands like `cargo clean` conflict.

analysis/test/Cargo.toml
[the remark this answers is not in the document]
kkysen: It's now optional in https://github.com/immunant/c2rust/pull/554/commits/de3fd2a1aa1b9b1970871f8aefd556831faf25a1.

@rinon, let me know if you want me to implement the other suggestions:
* automatically `cargo add /path/to/runtime --optional`, but otherwise the same
* create a shadow copy of the `Cargo.toml` at `Cargo.instrument.toml`, `cargo add /path/to/runtime` to it, run using `--manifest-path Cargo.instrument.toml`, and then delete it

dynamic_instrumentation/src/main.rs
kkysen: Not an issue anymore as `lib.rs` no longer exists.

dynamic_instrumentation/src/main.rs
kkysen: `get_sysroot_slow` is now implemented in https://github.com/immunant/c2rust/pull/554/commits/628266b0ca0cb7347ff981d6e806d36db41f347c.

kkysen:
Currently, we still need to set the path to find `rustc` private crates in `c2rust-instrument`.  This is necessary for the initial `cargo` wrapper invocation, but isn't necessary for the `rustc` wrapper invocations, as those are subcommands of `cargo` and `cargo` correctly sets the library path.  For the `cargo` wrapper part, we don't need any of the `rustc` private crates, so we could split this into two binaries, but that does have a few drawbacks.  @rinon, @aneksteind, do you think I should do this?  The alternative, which I think is superior, is to make `c2rust-instrument` (and eventually `c2rust-pdg` and `c2rust-analyzer`, too) subcommands of `cargo`.

analysis/test/Cargo.toml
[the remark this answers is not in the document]
kkysen: The `cargo add` approach is done in https://github.com/immunant/c2rust/pull/562, but is blocked on https://github.com/immunant/c2rust/pull/513, which upgrades our pinned `nightly` to one that supports `cargo add` natively.

kkysen:
Problem: `$CARGO_PRIMARY_PACKAGE` is set for build scripts as well, but `c2rust_analysis_rt` is not a dependency for those, causing it to fail.  I need a way to reliably detect when the `build.rs` script is compiling, which I believe is not *always* `build.rs` (and the binary named `build_script_build`).

kkysen:
I'm now (https://github.com/immunant/c2rust/pull/554/commits/26609468d6d4c8102139210fdc94891f59b02eca) detecting build scripts by checking if it is `--crate-type bin` and does not set `$CARGO_BIN_NAME`, because `$CARGO_BIN_NAME` is set for normal binary crates but not for build scripts.  I'm not sure if this is the best way to detect build scripts, but I couldn't find a better way.

kkysen:
However, now I'm getting an error in the actual instrumentation code, which wasn't happening previously, but this is probably from other changes and refactors to the instrumentation code in other PRs, as it's not in the code changes from this PR.

`./scripts/pdg.sh ../lighttpd/rust/`:

```shell
thread 'rustc' panicked at 'operand is not of integer-castable type: _4', dynamic_instrumentation/src/arg.rs:29:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/library/core/src/panicking.rs:143:14
   2: c2rust_instrument::point::build::InstrumentationBuilder::arg_var
   3: c2rust_instrument::instrument::<impl rustc_middle::mir::visit::Visitor for c2rust_instrument::point::InstrumentationAdder>::visit_assign::{{closure}}
   4: rustc_middle::mir::visit::Visitor::super_basic_block_data
   5: rustc_middle::mir::visit::Visitor::visit_body
   6: c2rust_instrument::instrument::Instrumenter::instrument_fn
   7: core::ops::function::FnOnce::call_once
   8: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<rustc_middle::ty::WithOptConstParam<rustc_span::def_id::LocalDefId>, &rustc_data_structures::steal::Steal<rustc_middle::mir::Body>>>
   9: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::mir_built
  10: rustc_mir_transform::check_unsafety::unsafety_check_result
  11: <rustc_mir_transform::check_unsafety::provide::{closure#0} as core::ops::function::FnOnce<(rustc_middle::ty::context::TyCtxt, rustc_span::def_id::LocalDefId)>>::call_once
  12: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<rustc_span::def_id::LocalDefId, &rustc_middle::mir::query::UnsafetyCheckResult>>
  13: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::unsafety_check_result
  14: rustc_mir_transform::mir_const
  15: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<rustc_middle::ty::WithOptConstParam<rustc_span::def_id::LocalDefId>, &rustc_data_structures::steal::Steal<rustc_middle::mir::Body>>>
  16: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::mir_const
  17: rustc_mir_transform::mir_promoted
  18: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<rustc_middle::ty::WithOptConstParam<rustc_span::def_id::LocalDefId>, (&rustc_data_structures::steal::Steal<rustc_middle::mir::Body>, &rustc_data_structures::steal::Steal<rustc_index::vec::IndexVec<rustc_middle::mir::Promoted, rustc_middle::mir::Body>>)>>
  19: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::mir_promoted
  20: <rustc_borrowck::provide::{closure#0} as core::ops::function::FnOnce<(rustc_middle::ty::context::TyCtxt, rustc_span::def_id::LocalDefId)>>::call_once
  21: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<rustc_span::def_id::LocalDefId, &rustc_middle::mir::query::BorrowCheckResult>>
  22: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::mir_borrowck
  23: <rustc_session::session::Session>::time::<(), rustc_interface::passes::analysis::{closure#2}>
  24: rustc_interface::passes::analysis
  25: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<(), core::result::Result<(), rustc_errors::ErrorReported>>>
  26: rustc_query_system::query::plumbing::get_query::<rustc_query_impl::queries::analysis, rustc_query_impl::plumbing::QueryCtxt>
  27: <rustc_interface::passes::QueryContext>::enter::<rustc_driver::run_compiler::{closure#1}::{closure#2}::{closure#3}, core::result::Result<(), rustc_errors::ErrorReported>>
  28: rustc_interface::interface::create_compiler_and_run::<core::result::Result<(), rustc_errors::ErrorReported>, rustc_driver::run_compiler::{closure#1}>
```

This is the code that panics: https://github.com/immunant/c2rust/blob/26609468d6d4c8102139210fdc94891f59b02eca/dynamic_instrumentation/src/arg.rs#L20-L31

@aneksteind, any thoughts?  I'm not too familiar with this part of the code other than refactors.

kkysen:
The above `lighttpd` `Adt` error happens on `master` as well, so it's definitely not caused by changes from this PR.  I think there was a change in a previous PR that landed that @fw-immunant had mentioned might not handle all cases, but I'm not sure where that is.

kkysen:
Filed an issue for it: https://github.com/immunant/c2rust/issues/570, so it shouldn't block this PR.

aneksteind (approved):

rinon (commented):
This PR has grown somewhat beyond it's original goal, which was just to build via the `cargo` cli tool instead of the library. I'd like to see this get split into a few different changes. At the very least, move the script changes into a separate PR?

I put a few more comments, but this is really too large to properly review as it is now.

dynamic_instrumentation/src/arg.rs
```diff
@@ -26,7 +26,8 @@ impl<'tcx> ArgKind<'tcx> {
         } else if ty.is_integral() {
             AddressUsize
         } else {
-            panic!("operand is not of integer-castable type: {:?}", op)
+            let kind = ty.kind();
+            panic!("operand is not of integer-castable type: op = {op:?}, ty = {ty:?}, ty.kind(): {kind:?}")
```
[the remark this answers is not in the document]
kkysen: No, I added it cause I thought the bug was in this PR, but it's not.  I've since cherry-picked this to `master` and opened an issue for it.

c2rust/src/main.rs
```diff
@@ -7,7 +7,7 @@ use std::process::{exit, Command};
 git_testament!(TESTAMENT);
 
 fn main() {
-    let subcommand_yamls = [load_yaml!("transpile.yaml"), load_yaml!("instrument.yaml")];
```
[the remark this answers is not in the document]
kkysen: The `c2rust instrument` binary now lives in `c2rust-instrument` rather than `c2rust`.  Shoehorning it back into `c2rust` is very unwieldy and would need extra code to do.  The deprecated clap YAML way is not used anymore, so it's more difficult.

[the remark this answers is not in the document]
kkysen: If `c2rust/arc/main.rs` didn't do the argument checks itself, it would still work, as would `c2rust-analyze` and `c2rust-pdg`.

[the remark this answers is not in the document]
kkysen: I do think the subcommand way needs improving, so I didn't think extra code was worth it here if it's not going to be eventually used, and it would further complicate this PR.

dynamic_instrumentation/src/lib.rs
```diff
@@ -1,223 +0,0 @@
-#![feature(rustc_private)]
```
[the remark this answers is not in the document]
kkysen: I didn't use `git mv` directly, though it seems it usually autodetects moves for me.  But I think there's a bit too many changes for it to see it as a `git mv`.  Basically the new changes in `main.rs` replaced the old `cargo` dependency stuff in `lib.rs`, and the rest was moved as is.

kkysen:
I can move the script deletions to a separate PR, but the `pdg.sh` changes are in sync with the changes to the other code such that `cargo test` should always work.

dynamic_instrumentation/src/arg.rs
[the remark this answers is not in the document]
kkysen: Should I rebase?  Doing so would remove the order of all the comments and the commits as is, which I find useful, so I was hesitant to do so for this small change.

dynamic_instrumentation/src/lib.rs
```diff
@@ -1,223 +0,0 @@
...
-use std::sync::{Arc, Mutex};
-
-lazy_static! {
-    /// TODO(kkysen) can be made non-lazy when `Mutex::new` is `const` in rust 1.63
-    static ref INSTRUMENTER: Instrumenter = Instrumenter::new();
-}
-
-struct NullCallbacks;
-
-impl rustc_driver::Callbacks for NullCallbacks {}
```
kkysen, line 65: This was deleted as now `TimePassesCallbacks::default()` is used instead, as that's `rustc`'s default (https://github.com/immunant/c2rust/pull/554/commits/8e9e94e7e1742983a38343925a1f847aff1a7ba9).

dynamic_instrumentation/src/lib.rs
```diff
@@ -1,223 +0,0 @@
...
-            rustc_driver::RunCompiler::new(&args, &mut callbacks)
-                .run()
-                .map_err(|_| anyhow!("Compilation failed"))
-        }
-    }
-
-    fn force_rebuild(&self, unit: &Unit) -> bool {
-        self.building_rt.load(Ordering::Relaxed) || self.default.force_rebuild(unit)
-    }
-}
```
kkysen, line 223: This was deleted as it is the `cargo`-as-a-library code that was deleted.  The rest of `lib.rs` was moved to the end of `main.rs` as-is, with only imports merged.

kkysen:
> This PR has grown somewhat beyond it's original goal, which was just to build via the `cargo` cli tool instead of the library. I'd like to see this get split into a few different changes. At the very least, move the script changes into a separate PR?
> 
> I put a few more comments, but this is really too large to properly review as it is now.

Is it really too large of a change to review?  There are ~160 new lines of code in `main.rs`, ~100 lines of documentation in `main.rs`, a few tiny additions in other files, and then a lot of code that was able to be deleted now.  I don't see a good way of breaking it up, and breaking it up by putting deleted files in another PR doesn't seem very helpful, as those are just dead code and (I don't think) require significant cognitive complexity to review.

It's unfortunate that the `lib.rs` to `main.rs` move with some changes is hard to see in the diff, but I just went through and tried to highlight the differences with comments on the code.  `NullCallbacks` and the `cargo`-as-a-library code was all deleted, and then the rest was moved as-is to `main.rs`.

Significant development-time savings are waiting on this PR, as well as the changes I'm trying to make off of this, which are hard to do while this one still changes.  Splitting this PR up into more is both difficult as I'm not sure how to do so effectively, might introduce bugs that I've already fixed, and make everything waiting for this PR take much longer.

There's been a lot of development on this PR, so there are a lot of commits, but a lot of those went towards simplifying things and fixing bugs.  For example, the `$CARGO_PRIMARY_PACKAGE` change allowed a lot of code to be removed.  Again, it's very unfortunate that the `lib.rs` to `main.rs` move couldn't be properly modeled by `git`, but there aren't actually so many additions in this PR, just a lot of dead code removal.

fw-immunant:
W/r/t the `lib.rs` -> `main.rs` move, is there a way to put that code in a separate (non-`lib.rs`) file with `pub(crate)` visibility to simplify the diff?

kkysen:
Oh, that's a good idea.  Let me see if that works.

dynamic_instrumentation/src/arg.rs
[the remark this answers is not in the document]
kkysen: I removed it from this PR now as we don't need to fix the bug here.  It's in `master`.

kkysen:
@fw-immunant, I tried moving that previously `lib.rs` code to `callbacks.rs`, but `git` still doesn't recognize it as a moved file.  Do you know how to get `git` to recognize that after the fact?  (I could create `callbacks.rs` on `master` with a similar refactor, but I'm not sure if I should do that.)

c2rust/src/main.rs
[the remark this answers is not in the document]
rinon: Sounds good

rinon (approved):

fw-immunant:
> @fw-immunant, I tried moving that previously `lib.rs` code to `callbacks.rs`, but `git` still doesn't recognize it as a moved file. Do you know how to get `git` to recognize that after the fact? (I could create `callbacks.rs` on `master` with a similar refactor, but I'm not sure if I should do that.)

git only detects moves/renames within a single commit, so the commits moving code from `lib.rs` into `main.rs` and moving back out from `main.rs` into `callbacks.rs` need to be squashed together. I tried doing this with
```
git show 'HEAD^{/Removed `InstrumentationExecutor::default: DefaultExecutor`}'`
```
but this change to the rebase plan
```
pick 676dee2d Deleted `c2rust-dynamic-instrumentation`'s `lib.rs` and moved it all to `bin.rs`, b/c it should only have a CLI interface.
squash d99b9a8d Move the previous `lib.rs` code in `main.rs` to `callbacks.rs` for a (hopefully) cleaner diff.
```
ran into a bunch of conflicts, presumably because of the merge commits between the two.

I don't really know how to work with interactive rebase in the presence of various merge commits.

kkysen:
@fw-immunant, I think it's too hard to rebase that far back now, as doing so would require a lot of manual merges.  @rinon approved now, so I think I'll just merge this as is, and then fix up the `c2rust` subcommand logic so `c2rust instrument` works the same again (as well as the other subcommands).

PR #555, opened 2022-07-25
Simplify the `InstrumentationPoint` builder

kkysen:
Simplify the `InstrumentationPoint` builder:
* pass `self` by `mut self` rather than `&mut Self`, which is more flexible
* remove an unnecessary `.clone()`
* elide unnecessary lifetimes


aneksteind:
Addressed in #552

PR #559, opened 2022-07-27
Modularize instrumentation code

kkysen:
`dynamic_instrumentation/instrument_memory.rs` is really large (> 1000 lines).  This splits it up into more sane modules.  There's further modularization coming, but that needs some refactoring as well, so it'll be in a separate PR.


dynamic_instrumentation/src/cast.rs
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/// Cast an argument from pointer to `usize`, if needed.
+///
+/// Casts `arg` to `usize` if needed, returning the cast statement and new,
+/// `usize`-typed operand if the cast was needed. This cast statement must be
+/// inserted into the function's body before the new operand is used.
+/// `arg` will be used as a copy in the new statement,
+/// so this statement must be inserted in a position where `arg` is alive.
+pub fn cast_ptr_to_usize<'tcx>(
```
aneksteind, line 23: feels odd to have a module for one function, and I don't anticipate needing to add more

dynamic_instrumentation/src/into_operand.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+impl<'tcx> IntoOperand<'tcx> for Operand<'tcx> {
+    fn op(self, _tcx: TyCtxt<'tcx>) -> Self {
+        self
+    }
+}
+
+fn make_const(tcx: TyCtxt, idx: u32) -> Operand {
```
aneksteind, line 36: i think functions like these and most of the functions in `deref` should be combined into a `mir_utils` module

dynamic_instrumentation/src/point.rs
```diff
@@ -0,0 +1,199 @@
...
+use rustc_middle::{
+    mir::{Body, Location, Place, Rvalue},
+    ty::TyCtxt,
+};
+use rustc_span::{def_id::DefId, Symbol};
+
+use crate::{
+    arg::{ArgKind, InstrumentationArg},
+    instrument::find_instrumentation_def,
+    into_operand::IntoOperand,
```
aneksteind, line 12: `into_operand` can be folded into here, it's not used elsewhere and is specific to instrumentation points

dynamic_instrumentation/src/source.rs
```diff
@@ -0,0 +1,67 @@
+use c2rust_analysis_rt::mir_loc::MirPlace;
+use rustc_middle::mir::{Local, Operand, Place, Rvalue};
+
+use crate::util::Convert;
+
+pub trait Source {
```
aneksteind, line 6: i think this module should also be folded into `point`, for the same reasons `into_operand` should

dynamic_instrumentation/src/source.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl Source for u32 {
+    fn source(&self) -> Option<MirPlace> {
+        Local::from_u32(*self).source()
+    }
+}
+
+/// Get the one and only input [`Place`], if applicable.
+fn rv_place<'tcx>(rv: &'tcx Rvalue) -> Option<Place<'tcx>> {
```
aneksteind, line 48: `mir_utils`

kkysen: `rv_value` is only used here by a `Source` `impl`, so it's private and encapsulated.  I think it should stay together.

dynamic_instrumentation/src/into_operand.rs
kkysen: This can be encapsulated by this module and made private, which is why I put it here.

dynamic_instrumentation/src/cast.rs
kkysen: I don't think it's that odd.  Would it make more sense in `mir_utils`?

kkysen:
I think we should be focusing on smaller modules.  The existing ones were way too big.  I can move some of the modules into `mod point` as you noted, but I'll put them as submodules.  Does that work?

dynamic_instrumentation/src/source.rs
aneksteind: you're right it's only used here, but that's coincidental -- we have a bunch of MIR-specific utilities and this can be one of them, there is no issue making it public (it's similar to `Operand::place` anyhow)

dynamic_instrumentation/src/into_operand.rs
aneksteind: no need for it to be private, it can be useful in future contexts

aneksteind:
I don't disagree, but some are broken up too much, in my opinion. If we folded `source` and `into_operand` into `point`, the modules would still be broken up and provide the much-needed improvement.

dynamic_instrumentation/src/cast.rs
aneksteind: i don't think it belongs in `mir_utils` because it references non-mir types, i.e. InstrumentationArg. It should go into `points`, if anything, and some of the internals could be broken up into mir-util-specific behavior and factored out in another PR

dynamic_instrumentation/src/source.rs
kkysen: Okay.

dynamic_instrumentation/src/into_operand.rs
kkysen: Unlike `rv_place`, I don't agree.  The `IntoOperand` `impl` does exactly the same, and I think it's bad to have two exposed functions that do the exact same thing, which causes people to end up using both randomly.

dynamic_instrumentation/src/cast.rs
kkysen: Can we leave it in it's own module for now and then refactor it into pieces in other modules later?

aneksteind:
I think the modularization in #560 splits up `points` in a more logical fashion than separating `cast`, `into_operand`, and `source` from it and does the job for `points` well-enough on its own.

dynamic_instrumentation/src/into_operand.rs
aneksteind: ok, on second thought I think you're right and `into_operand` can stay its own module

aneksteind:
I've come around on `into_operand`, I think that can stay as-is

aneksteind (changes requested):
I'm okay with `cast`, `source` being their own submodules of `point`. `into_operand` can stay as-is. I'd prefer if someone were using `cast_ptr_to_usize` that they import it with  `use crate::point::cast_ptr_to_usize`

kkysen:
Okay, that sounds good.  Do you have other comments for the following PR?  If you don't, it might be easier for me to make those above changes on top of both of these PRs.

aneksteind:
Nothing substantial re: #560. It's okay to add it there.

dynamic_instrumentation/src/cast.rs
kkysen: Moved to a submodule of `mod point` in https://github.com/immunant/c2rust/pull/559/commits/8e2e95f0e639c467ca23f1b1a4e19c57fea5c749 and re-exported from `mod point` in https://github.com/immunant/c2rust/pull/559/commits/1982e8d040d9a2b68da5557817793ad373ea2d01.

dynamic_instrumentation/src/point/cast.rs
```diff
@@ -10,7 +10,7 @@ use rustc_span::DUMMY_SP;
 
 use crate::{
     arg::{ArgKind, InstrumentationArg},
-    deref::remove_outer_deref,
```
aneksteind: would like a similar re-export here to have `mir_utils::remove_outer_deref`

kkysen: Fixed in https://github.com/immunant/c2rust/pull/559/commits/50d305f1f2510161d8326d472dd561c0d87fa9e6.

aneksteind (approved):

PR #560, opened 2022-07-27
Further modularize and refactor instrumentation code

kkysen:
This further modularizes the instrumentation code, but adds some refactoring as well to make APIs better (less interdependent) and allow for certain APIs (like `Hooks`) to be used more broadly (allowing some previous manual inlining to be refactored).


dynamic_instrumentation/src/point/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+        Self {
+            hooks,
+            body,
+            instrumentation_points: Default::default(),
+            assignment: Default::default(),
+        }
+    }
+
+    pub fn tcx(&self) -> TyCtxt<'tcx> {
+        self.hooks.tcx()
```
aneksteind, line 41: why does `hooks` need a `tcx`?

kkysen: It needs it to look through module items for the hooks.  I wasn't sure if I should put the `TyCtxt` in `Hooks`, but I tried other places and it didn't make as much sense or led to a bunch more code.

aneksteind: i don't think `InstrumentationAdder` should rely on the `tcx` that `hooks` has. If we didn't have hooks, we'd still want `InstrumentationAdder` to have access to a `tcx`. In other words I think it should just get its own.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -7,23 +7,23 @@ use log::debug;
...
 use std::path::Path;
 use std::sync::Mutex;
 
 use crate::arg::{ArgKind, InstrumentationArg};
 use crate::cast::cast_ptr_to_usize;
 use crate::deref::{has_outer_deref, remove_outer_deref, strip_all_deref};
-use crate::into_operand::IntoOperand;
-use crate::point::{InstrumentationAdder, InstrumentationPoint};
+use crate::hooks::Hooks;
+use crate::point::apply::InstrumentationApplier;
```
aneksteind, line 25: can `point` re-export `apply` to have `crate::point::InstrumentApplier`, which matches `crate::point::InstrumentationAdder`?

kkysen: Sure, but I don't think it's too important.  They're usually just auto-imported.

dynamic_instrumentation/src/point/mod.rs
kkysen: I can do that.  That makes the struct bigger to store duplicate `TyCtxt`s, but it shouldn't matter much.

dynamic_instrumentation/src/instrument.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/560/commits/1e7f2a09b4a18319d50e668a97d1d46f612f0099.

dynamic_instrumentation/src/point/mod.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/560/commits/2d2813d49195fd8b378b4eeda6abb057b1b96ef9.

aneksteind (approved):

PR #562, opened 2022-07-30
Run `cargo add` in `c2rust-instrument --set-runtime` to add `c2rust-analysis-rt` automatically

kkysen:
Currently, in #554, the `c2rust-analysis-rt` runtime must be added as an (optional) dependency in the instrumented crate.  This runs `cargo add c2rust-analysis-rt --optional` so that it is automatically added and up-to-date when instrumenting.  This also adds an optional `--runtime ${runtime}` argument to `c2rust-instrument`.  If it is given, then `cargo add c2rust-analysis-rt --optional --path ${runtime} --offline` is run.  Thus, it works well both when `c2rust-analysis-rt` is already on your machine and you're developing locally, as well as a user who would download `c2rust-analysis-rt` from [crates.io](https://crates.io).  Note that `cargo add` is only run when `--set-runtime` is passed, so this is off by default, and must be opted-into.

Another, more complex, option is to copy the `Cargo.toml` to a `Cargo.instrument.toml`, add the `c2rust-analysis-rt` dependency there, and the run with `--manifest-path Cargo.instrument.toml`.  However, that's more complex, and not all `cargo` commands uniformly take a `--manifest-path` argument, and there's no environment variable for that, which would be seamlessly inherited.

### DIFF
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -316,7 +316,7 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
         // it usually isn't that slow.
         cmd.args(&["clean", "--package", root_package.name.as_str()]);
     })?;
-
+    
     // TODO(kkysen) Once we upgrade to 1.62, we can just use `cargo` and not specify `+stable`.
     // The problem currently is that `+stable` doesn't work on a resolved `$CARGO`,
     // which happens when this runs inside of `cargo test`,
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -51,7 +51,7 @@ struct Args {
     /// Path to the `c2rust-analysis-rt` crate if you want to use a local version of it (vs. the crates.io one).
     /// This is not used unless `--set-runtime` is also passed.
     #[clap(long, value_parser)]
-    runtime: Option<PathBuf>,
+    runtime_path: Option<PathBuf>,
 
     /// Add the runtime as an optional dependency to the instrumented crate using `cargo add`.
     #[clap(long)]
@@ -295,7 +295,7 @@ fn set_rust_toolchain() -> anyhow::Result<()> {
 fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
     let Args {
         metadata,
-        runtime,
+        runtime_path,
         set_runtime,
         mut cargo_args,
     } = Args::parse();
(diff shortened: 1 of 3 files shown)


rinon:
I don't think we should automatically run `cargo add` when instrumenting, that seems too magic to me. We should just document this requirement in a README. The most automatic thing I think we should possibly do is diagnose this error when instrumenting and present a friendly version.

kkysen:
It's off by default now and enabled with `--set-runtime`.  I still think this is good for testing especially, since it ensures the runtime crate is up-to-date and points to the correct version.  This allows us to avoid keeping the runtime dependency in `Cargo.toml` checked into version control, too.

dynamic_instrumentation/src/main.rs
```diff
@@ -48,6 +48,15 @@ struct Args {
     #[clap(long, value_parser)]
     metadata: PathBuf,
 
+    /// Path to the `c2rust-analysis-rt` crate if you want to use a local version of it (vs. the crates.io one).
+    /// This is not used unless `--set-runtime` is also passed.
+    #[clap(long, value_parser)]
+    runtime: Option<PathBuf>,
```
rinon, line 54: I would call this `runtime_path` or something.

dynamic_instrumentation/src/main.rs
```diff
@@ -48,6 +48,15 @@ struct Args {
...
+
+    /// Add the runtime as an optional dependency to the instrumented crate using `cargo add`.
+    #[clap(long)]
+    set_runtime: bool,
```
rinon, line 58: What about `add_runtime`?

dynamic_instrumentation/src/main.rs
```diff
@@ -312,6 +323,17 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
         cmd.args(&["clean", "--package", root_package.name.as_str()]);
     })?;
 
+    if set_runtime {
+        cargo.run(|cmd| {
```
rinon, line 327: What happens when the crate is already added? Does this exit with a non-zero status?

kkysen: No, it just overwrites it or does nothing, which is why I thought `--set-runtime` was clearer.

dynamic_instrumentation/src/main.rs
kkysen: I had named it `--runtime` since it was previously named that, before the `cargo`/`$RUSTC_WRAPPER` refactor.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/562/commits/9d89818f44016348ff96a554c646f6ba53a03198.

dynamic_instrumentation/src/main.rs
kkysen: I thought `--set-runtime` was clearer since it overwrites the previous value, and this will presumably be called many times, especially in testing.

rinon (approved):

PR #566, opened 2022-08-01
Snapshot debug representation improvements

aneksteind:


Cargo.lock
```diff
@@ -287,6 +287,7 @@ dependencies = [
  "cargo",
  "cargo-util",
  "indexmap",
+ "itertools",
```
kkysen, line 290: I added `itertools` cause it greatly simplifies the `args` debug printing, and it's already used in the workspace.

dynamic_instrumentation/Cargo.toml
```diff
@@ -10,6 +10,7 @@ c2rust-analysis-rt = { path = "../analysis/runtime"}
 cargo = "0.62"
 cargo-util = "0.1"
 indexmap = "1.8"
+itertools = "0.10"
```
kkysen, line 13: I added `itertools` because it greatly simplifies the `args` debug printing, and it's already used in the workspace.

dynamic_instrumentation/src/point/build.rs
```diff
@@ -139,33 +140,33 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
...
+                destination: Some((destination, _)),
+                func,
+                ..
+            } => {
+                let func_name = if let &ty::FnDef(def_id, _) = func.ty(self.body, self.tcx).kind() {
+                    let name = self.tcx.item_name(def_id);
+                    format!("{name}")
+                } else {
+                    format!("{func:?}")
+                };
```
kkysen, line 160: I couldn't see an easy way to avoid the extra `String` allocation here without making the code a lot harder to read, but it shouldn't be too important.

dynamic_instrumentation/src/point/build.rs
```diff
@@ -139,33 +140,33 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
...
-                format!("{:?}", block.statements[loc.statement_index])
+    fn debug_mir_to_string(&self, loc: Location) -> String {
+        let block = &self.body.basic_blocks()[loc.block];
+        if loc.statement_index != block.statements.len() {
+            return format!("{:?}", block.statements[loc.statement_index]);
+        }
+        match &block.terminator().kind {
+            TerminatorKind::Call {
+                args,
+                destination: Some((destination, _)),
```
kkysen, line 151: This was an easy fix compared to the `match ... if destination.is_some()` and `destination.unwrap()`.

dynamic_instrumentation/src/point/build.rs
```diff
@@ -139,33 +140,33 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
...
+                    let name = self.tcx.item_name(def_id);
+                    format!("{name}")
+                } else {
+                    format!("{func:?}")
+                };
+                let args = args.iter().format(", ");
+                format!("{destination:?} = {func_name}({args:?})")
             }
-        };
+            _ => String::from(""),
```
kkysen, line 164: Should we really print nothing here?

dynamic_instrumentation/src/point/build.rs
```diff
@@ -139,33 +140,33 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
...
+        }
+    }
+
+    pub fn debug_mir(mut self, loc: Location) -> Self {
+        self.point.metadata.debug_info = self.debug_mir_to_string(loc);
```
kkysen, line 169: The changes here didn't change the snapshot at all.  I'm pretty sure I didn't change anything.

dynamic_instrumentation/src/point/build.rs
```diff
@@ -139,33 +140,33 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
...
+                    let name = self.tcx.item_name(def_id);
+                    format!("{name}")
+                } else {
+                    format!("{func:?}")
+                };
+                let args = args.iter().format(", ");
+                format!("{destination:?} = {func_name}({args:?})")
             }
-        };
+            _ => "".into(),
```
kkysen, line 164: Should we really print nothing here?

kkysen (approved):

kkysen:
@aneksteind, can you check my changes and comments?

dynamic_instrumentation/src/point/build.rs
aneksteind: I think it's fine for now

dynamic_instrumentation/src/point/build.rs
```diff
@@ -150,7 +150,13 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
...
                         ..
                     } if destination.is_some() => {
-                        let mut s = format!("{:?} = {:?}(", destination.unwrap().0, func);
+                        let mut s = format!("{:?} = ", destination.unwrap().0);
+                        if let ty::FnDef(def_id, _) = func.ty(self.body, self.tcx).kind() {
+                            let name = self.tcx.item_name(*def_id);
+                            s.push_str(&format!("{}(", name));
+                        } else {
+                            s.push_str(&format!("{:?}(", func));
+                        };
```
kkysen: I hadn't realized this changed, which changes things like this:

```rust
std::vec::Vec::<*mut i8>::push(move _29, move _30);
```

to this:

```rust
push(move _29, move _30);
```

which is IMO a lot less readable.  Sorry, I just realized this on some other merge.

What we want is

```rust
Vec::<*mut i8>::push(move _29, move _30);
```

since `Vec` is in the `std` prelude.

If that's not easy, I think we should revert this change.  I'll open a PR to do so.

PR #567, opened 2022-08-01
add test for fields of references

aneksteind:


pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -687,6 +687,22 @@ g {
...
+	n[0]:  alloc      _     => _2  @ bb1[2]:  fn test_ref_field; _2 = pointers::calloc(move _3, move _4)  
+	n[1]:  copy       n[0]  => _1  @ bb2[3]:  fn test_ref_field; _1 = move _2 as *mut pointers::S (Misc)  
+	n[2]:  copy       n[1]  => _6  @ bb2[6]:  fn test_ref_field; _6 = &mut (*_1)                          
+	n[3]:  field.3    n[2]  => _   @ bb2[9]:  fn test_ref_field; _7 = (((*_6).3: pointers::T).3: i32)     
+	n[4]:  field.3    n[3]  => _7  @ bb2[9]:  fn test_ref_field; _7 = (((*_6).3: pointers::T).3: i32)     
+	n[5]:  addr.load  n[4]  => _   @ bb2[9]:  fn test_ref_field; _7 = (((*_6).3: pointers::T).3: i32)     
+	n[6]:  field.3    n[2]  => _   @ bb2[10]: fn test_ref_field; (((*_6).3: pointers::T).3: i32) = move _7
+	n[7]:  field.3    n[6]  => _   @ bb2[10]: fn test_ref_field; (((*_6).3: pointers::T).3: i32) = move _7
+	n[8]:  addr.store n[7]  => _   @ bb2[10]: fn test_ref_field; (((*_6).3: pointers::T).3: i32) = move _7
+	n[9]:  copy       n[0]  => _10 @ bb2[16]: fn test_ref_field; _10 = &raw mut (*_6)
```
spernsteiner, line 700: This has source `n[0]`, but I think it should be `n[2]`, since that's the last node that assigned to `_6`.

aneksteind: Fixed in 4ca53bd3

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -687,6 +687,22 @@ g {
 }
 nodes_that_need_write = []
 
+g {
+	n[0]:  alloc      _     => _2  @ bb1[2]:  fn test_ref_field; _2 = pointers::calloc(move _3, move _4)  
+	n[1]:  copy       n[0]  => _1  @ bb2[3]:  fn test_ref_field; _1 = move _2 as *mut pointers::S (Misc)
```
kkysen, line 692: What does `(Misc)` mean?

analysis/test/src/pointers.rs
```diff
@@ -504,6 +514,7 @@ unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> lib
     test_arg_rec();
     test_shared_ref();
     test_unique_ref();
+    test_ref_field();
```
kkysen, line 517: Could you put the new `test_ref_field()` at the end so it doesn't disturb the snapshot as much?  Or is there a reason it's in that position?

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -165,59 +165,59 @@ g {
 	n[1]:  copy n[0]  => _7  @ bb2[14]:  fn exercise_allocator;      _7 = move _8 as *const u8 (Pointer(ArrayToPointer))  
 	n[2]:  copy n[1]  => _6  @ bb2[16]:  fn exercise_allocator;      _6 = move _7 as *const i8 (Misc)                     
 	n[3]:  copy n[2]  => _1  @ bb0[0]:   fn printf;                  _5 = pointers::printf(move _6, move _10)             
-	n[4]:  copy n[0]  => _30 @ bb11[7]:  fn exercise_allocator;      _30 = &raw const (*_31)                              
+	n[4]:  copy _     => _30 @ bb11[7]:  fn exercise_allocator;      _30 = &raw const (*_31)
```
kkysen, line 168: Does this have no source because it's a copy of a string literal?

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
kkysen: And in general the other things in the parentheses after the type.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -241,10 +241,12 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
                     statement_index: location.statement_index + 1,
                     ..location
                 };
+
+                let source = remove_outer_deref(*p, self.tcx());
```
kkysen, line 245: `remove_outer_deref(*p, self.tcx())` is already computed above on line 238:

```rust
                    && place_ty(&remove_outer_deref(*p, self.tcx())).is_region_ptr() =>
```

Could you de-duplicate this?

kkysen: Maybe a `try_remove_outer_deref` would be useful here:

```rust
/// Try to strip the initital [`Deref`](ProjectionElem::Deref)
/// from a [`projection`](PlaceRef::projection) sequence.
pub fn try_remove_outer_deref<'tcx>(p: Place<'tcx>, tcx: TyCtxt<'tcx>) -> Option<Place<'tcx>> {
    // Remove outer deref if present
    match p.as_ref() {
        PlaceRef {
            local,
            projection: &[ref base @ .., ProjectionElem::Deref],
        } => Some(Place {
            local,
            projection: tcx.intern_place_elems(base),
        }),
        _ => None,
    }
}

/// Strip the initital [`Deref`](ProjectionElem::Deref)
/// from a [`projection`](PlaceRef::projection) sequence
/// if there is one.
pub fn remove_outer_deref<'tcx>(p: Place<'tcx>, tcx: TyCtxt<'tcx>) -> Place<'tcx> {
    try_remove_outer_deref(p, tcx).unwrap_or(p)
}
```

kkysen: Could do

```rust
            Rvalue::AddressOf(_, p)
                if let Some(source) = try_remove_outer_deref(*p, self.tcx())
                && place_ty(source).is_region_ptr() =>
            {
                let instrumentation_location = Location {
                    statement_index: location.statement_index + 1,
                    ..location
                };
                // Instrument which local's address is taken
                self.loc(instrumentation_location, copy_fn)
                    .arg_var(dest)
                    .source(source)
                    .dest(&dest)
                    .debug_mir(location)
                    .add_to(self);
            }
```

but that requires `#![feature(if_let_guard, let_chains)]`.  It might be too clunky without those, I think.

kkysen: How about the changes in https://github.com/immunant/c2rust/pull/573?

kkysen: I think https://github.com/immunant/c2rust/pull/573 is probably too big for the scope of this PR, but keep it in my mind for after merging this PR.

analysis/test/src/pointers.rs
kkysen: If you want to, it's done in https://github.com/immunant/c2rust/pull/574.  The diff with `master` is cleaner.

kkysen (approved):
See my improvement PRs.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
spernsteiner: That's the [`CastKind`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/enum.CastKind.html).  There are a few kinds of pointer coercions that have specific `CastKind`s, and everything else is `CastKind::Misc`.

analysis/test/src/pointers.rs
```diff
@@ -376,6 +376,16 @@ pub unsafe extern "C" fn test_unique_ref() {
...
 }
 #[no_mangle]
+pub unsafe extern "C" fn test_ref_field() {
+    let s = calloc(
+        0i32 as libc::c_ulong,
+        ::std::mem::size_of::<S>() as libc::c_ulong,
+    ) as *mut S;
+    let s_ref = &mut *s;
+    (*s_ref).field4.field4 = (*s_ref).field4.field4;
+    free(s_ref as *mut S as *mut libc::c_void);
```
fw-immunant, line 386: This pattern (`free` on a ptr-casted ref) is surprising to me--I don't think the transpiler emits this sort of code (it would pass the original raw ptr to free), and neither idiomatic nor partially-refactored, transpiled rust seems likely to behave this way either (the latter would presumably not make `s_ref` a mutable reference if it saw it being passed to `free`). This doesn't mean this isn't a reasonable way to test accesses to fields of references for instrumentation, but it did raise some eyebrows. Are we avoiding using a stack `S` here because `AddrOfLocal` support is still WIP?

fw-immunant: `calloc(0, ...)` here is not sane since we do read/write into the allocation.

fw-immunant (approved):
Approved conditional on fixing the `calloc(0, ...)` in the test case.

analysis/test/src/pointers.rs
kkysen: It should be `1`, right?  How come that doesn't segfault on a null pointer then?

analysis/test/src/pointers.rs
```diff
@@ -376,6 +376,16 @@ pub unsafe extern "C" fn test_unique_ref() {
     *fresh1 = &mut x as *mut i32;
 }
 #[no_mangle]
+pub unsafe extern "C" fn test_ref_field() {
+    let s = calloc(
+        0i32 as libc::c_ulong,
```
kkysen, line 381:
```suggestion
        1i32 as libc::c_ulong,
```

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
aneksteind: it's because the rvalue of the statement that assigns to `_31` isn't yet instrumented, i.e. `Operand::Constant`. I'll make an issue for this

aneksteind: https://github.com/immunant/c2rust/issues/577

Issue #569, opened 2022-08-01
cargo build fails on MacOS


thedataking:
`cargo build` on `master` fails on MacOS (tested on 12.5). Error

```shell
error: failed to run custom build command for `c2rust-ast-exporter v0.16.0 (/Users/perl/Work/c2rust/c2rust-ast-exporter)`

Caused by:
  process didn't exit successfully: `/Users/perl/Work/c2rust/target/debug/build/c2rust-ast-exporter-134ca328f5ca9f46/build-script-build` (signal: 6, SIGABRT: process abort signal)
  --- stderr
  dyld[18823]: Library not loaded: '@rpath/libclang.dylib'
    Referenced from: '/Users/perl/Work/c2rust/target/debug/build/c2rust-ast-exporter-134ca328f5ca9f46/build-script-build'
    Reason: tried: '/Users/perl/Work/c2rust/target/debug/deps/libclang.dylib' (no such file), '/Users/perl/Work/c2rust/target/debug/libclang.dylib' (no such file), '/Users/perl/.rustup/toolchains/nightly-2022-02-14-x86_64-apple-darwin/lib/rustlib/x86_64-apple-darwin/lib/libclang.dylib' (no such file), '/Users/perl/.rustup/toolchains/nightly-2022-02-14-x86_64-apple-darwin/lib/libclang.dylib' (no such file), '/Users/perl/lib/libclang.dylib' (no such file), '/usr/local/lib/libclang.dylib' (no such file), '/usr/lib/libclang.dylib' (no such file)
```

```shell
$ brew info llvm
llvm: stable 14.0.6 (bottled), HEAD [keg-only]
Next-gen compiler infrastructure
https://llvm.org/
/usr/local/Cellar/llvm/14.0.6_1 (5,851 files, 988.2MB)
  Poured from bottle on 2022-07-23 at 18:59:22
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/llvm.rb
License: Apache-2.0 with LLVM-exception
```

kkysen:
Hmm, that's weird.  The Darwin CI has been fine, and the rpath change (https://github.com/immunant/c2rust/pull/518) you tested (https://github.com/immunant/c2rust/pull/518#issuecomment-1183893828).  Plus, this is coming from `c2rust-ast-exporter`, which I don't think uses rpath.

kkysen:
[...]

PR #571, opened 2022-08-02
Fix macOS build crash in #569 by reverting #526

kkysen:
Fixes https://github.com/immunant/c2rust/issues/569.

This reverts the `default-features = false` changes from https://github.com/immunant/c2rust/commit/b971ab5d3cadaaa112bb4b51a9e22fa1504352fc (part of https://github.com/immunant/c2rust/pull/526).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -321,7 +321,6 @@ dependencies = [
  "c2rust-ast-exporter",
  "c2rust-ast-printer",
  "c2rust-bitfields",
- "clap 2.34.0",
  "colored 2.0.0",
  "dtoa",
  "failure",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,6 +103,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
+ "clap 3.2.8",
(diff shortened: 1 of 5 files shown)


thedataking (approved):
This LGTM insofar that it will fix the problem but I think you made a good point that it might be safer to revert the problematic commit entirely as we don't know that we found all issues.

kkysen:
I effectively reverted all the `default-features = false` changes from https://github.com/immunant/c2rust/commit/b971ab5d3cadaaa112bb4b51a9e22fa1504352fc.  I had already reverted the `c2rust-bitfields` deletion in https://github.com/immunant/c2rust/commit/5170f64abff8c795e7eb100e82017c05a06b659d.  The remaining change, the `clap` dependency deletion from `c2rust-transpile` should be fine, because the only instance of `clap` in `c2rust-transpile` (checked with `rg`) is a comment saying how `c2rust_transpile::transpile` should be used, and it is not an intra-doc link.  `clap` is also a large dependency, so this change is actually quite useful.  Thus, I re-deleted `clap` in a separate commit.

PR #573, opened 2022-08-02
Kkysen/tests.refs.fields/remove outer deref refactor

kkysen:
This refactors `mod deref`, implementing most of the functions there in terms of one `try_remove_outer_deref_as_ref` so that we know they're all in sync.  Also, `try_remove_outer_deref` is added so that you don't have to do separate `has_outer_deref` and `remove_outer_deref` calls.

Then I used this to de-duplicate the 1 `has_outer_ref` and 2 `remove_outer_deref` calls in a `RValue::AddressOf` `match` arm by combining those 2 arms and using `try_remove_outer_deref`..

Then I used [`apply`](https://docs.rs/apply) to simplify the conditional builder logic.  It's quite a small but useful dependency for functional, chaining code.

### DIFF
--- a/dynamic_instrumentation/src/instrument.rs
+++ b/dynamic_instrumentation/src/instrument.rs
@@ -20,7 +20,9 @@ use std::sync::Mutex;
 
 use crate::arg::{ArgKind, InstrumentationArg};
 use crate::hooks::Hooks;
-use crate::mir_utils::{has_outer_deref, remove_outer_deref, strip_all_deref};
+use crate::mir_utils::{
+    has_outer_deref, remove_outer_deref, strip_all_deref, try_remove_outer_deref,
+};
 use crate::point::cast_ptr_to_usize;
 use crate::point::InstrumentationAdder;
 use crate::point::InstrumentationApplier;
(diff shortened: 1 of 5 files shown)


dynamic_instrumentation/src/instrument.rs
```diff
@@ -229,25 +230,21 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
...
-                    .arg_var(dest)
-                    .source(&source)
-                    .dest(&dest)
-                    .debug_mir(location)
-                    .add_to(self);
-            }
             Rvalue::AddressOf(_, p) => {
                 // Instrument which local's address is taken
-                self.loc(location.successor_within_block(), addr_local_fn)
+                let source = try_remove_outer_deref(*p, self.tcx())
```
aneksteind, line 235: i don't want to combine these, there are more complicated cases incoming and it'll be much clearer if they're separate

dynamic_instrumentation/src/instrument.rs
```diff
@@ -281,49 +278,30 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
                     .debug_mir(location)
                     .add_to(self);
             }
-            Rvalue::Ref(_, bkind, p) if has_outer_deref(p) => {
+            Rvalue::Ref(_, bkind, p) => {
                 // this is a reborrow or field reference, i.e. _2 = &(*_1)
-                let source = remove_outer_deref(*p, self.tcx());
-                if let BorrowKind::Mut { .. } = bkind {
+                let source = try_remove_outer_deref(*p, self.tcx());
```
aneksteind, line 283: i don't want to combine these, there are more complicated cases incoming and it'll be much clearer if they're separate

aneksteind (changes requested):
I'm not in favor of combining the match arms. I do however like your original idea outlined [here](https://github.com/immunant/c2rust/pull/567#discussion_r935134219) in order to bind `source` in the match arm for use in the body and would support that change.

kkysen:
> I'm not in favor of combining the match arms. I do however like your original idea outlined [here](https://github.com/immunant/c2rust/pull/567#discussion_r935134219) in order to bind `source` in the match arm for use in the body and would support that change.

My original idea uses `#![feature(if_let_guard, let_chains)]`, though.  It is a lot nicer, though, I think, but I don't want to add unstable features.

kkysen:
`#![feature(let_chains)]` is already merged and stabilizing in Rust 1.64 (https://github.com/rust-lang/rust/pull/94927), so since we're pinned to a `nightly`, too, I think using that should be fine, but `#![feature(if_let_guard)]` is not close to stabilization (https://github.com/rust-lang/rust/issues/51114).

kkysen:
I understand there might be more complex logic coming, so I shouldn't combine the logic in the way I did, but I don't think combining `match` arms into one per `Rvalue` variant is bad.  We can then do more complex branching within each `Rvalue` `match` arm rather than be limited to only guards.

PR #574, opened 2022-08-02
Move the new test to the end so the snapshot isn't disturbed as much

kkysen:
The snapshot diff with `master` is now cleaner, as it doesn't change `fn main_0` as much.


aneksteind:
I personally favor having tests that test similar things in proximity to one another.

PR #579, opened 2022-08-04
Make `c2rust` just forward args to discovered subcommands

kkysen:
This changes `c2rust` to be a simpler wrapper around the other `c2rust-*` subcommand.

Instead of `c2rust` having to know about all the subcommands and their arguments upfront, `c2rust $subcommand $args` just runs `c2rust-$subcommand $args`.  This allows `c2rust instrument` to work as before (before #554), while also enabling `c2rust analyze` and `c2rust pdg` in the same way.  The `clap` help messages are still preserved for the most part, except for the short help messages for the subcommands.  Otherwise, `c2rust --help` works as before (while also suggesting the new subcommands), and `c2rust $subcommand --help` works by running `c2rust-$subcommand --help` (instead of `clap` intercepting the `--help`).

The way this is implemented is, first the `c2rust` binary's directory is searched for executables named `c2rust-*` to discover subcommands.  This is combined with the simple list of known subcommands (`["transpile", "instrument", "pdg", "analyze"]`) in case they're not discovered properly and we still want to suggest them.  Then we check if the first argument is one of these subcommands.  If it exists, we invoke it.  If it doesn't exist, but is known, we suggest building it, and it doesn't exist and isn't known (or there was no subcommand given), then we run the `clap` parser and let it handle arg parsing and nice error/help messages.  The reason we don't have everything go through `clap` is that I couldn't figure out a way to have `clap` just forward all arguments, even ones like `--metadata` with hyphens (`Arg::allow_hyphen_values` didn't work), without requiring a leading `--` argument.

### DIFF
--- a/c2rust/src/main.rs
+++ b/c2rust/src/main.rs
@@ -42,25 +42,22 @@ impl SubCommand {
         for entry in dir.read_dir()? {
             let entry = entry?;
             let file_type = entry.file_type()?;
-            if !(file_type.is_file() || file_type.is_symlink()) {
-                continue;
-            }
-            let file_name = entry.file_name();
-            let name = match Path::new(&file_name)
-                .to_str()
+            let path = entry.path();
+            let name = match path.file_name()
+                .and_then(|name| name.to_str())
                 .and_then(|name| name.strip_prefix(c2rust_name))
                 .and_then(|name| name.strip_prefix('-'))
+                .map(|name| name.to_owned())
+                .map(Cow::from)
+                .filter(|_| file_type.is_file() || file_type.is_symlink())
+                .filter(|_| path.is_executable())
             {
                 Some(name) => name,
                 None => continue,
             };
-            let path = entry.path();
-            if !path.is_executable() {
-                continue;
-            }
             sub_commands.push(Self {
                 path: Some(path),
-                name: name.to_owned().into(),
+                name,
             });
         }
         Ok(sub_commands)
--- a/c2rust/src/main.rs
+++ b/c2rust/src/main.rs
@@ -43,22 +43,21 @@ impl SubCommand {
             let entry = entry?;
             let file_type = entry.file_type()?;
             let path = entry.path();
-            let name = match path.file_name()
+            let name = path
+                .file_name()
                 .and_then(|name| name.to_str())
                 .and_then(|name| name.strip_prefix(c2rust_name))
                 .and_then(|name| name.strip_prefix('-'))
                 .map(|name| name.to_owned())
                 .map(Cow::from)
                 .filter(|_| file_type.is_file() || file_type.is_symlink())
-                .filter(|_| path.is_executable())
-            {
-                Some(name) => name,
-                None => continue,
-            };
-            sub_commands.push(Self {
-                path: Some(path),
-                name,
-            });
+                .filter(|_| path.is_executable());
+            if let Some(name) = name {
+                sub_commands.push(Self {
+                    path: Some(path),
+                    name,
+                });
+            }
         }
         Ok(sub_commands)
     }
--- a/c2rust/src/main.rs
+++ b/c2rust/src/main.rs
@@ -16,7 +16,9 @@ git_testament!(TESTAMENT);
 
 /// A `c2rust` sub-command.
 struct SubCommand {
-    /// The path to the [`SubCommand`]'s executable.
+    /// The path to the [`SubCommand`]'s executable,
+    /// if it was found (see [`Self::find_all`]).
+    /// Otherwise [`None`] if it is a known [`SubCommand`] (see [`Self::known`]).
     path: Option<PathBuf>,
     /// The name of the [`SubCommand`], i.e. in `c2rust-{name}`.
     name: Cow<'static, str>,
@@ -112,6 +114,8 @@ fn main() -> anyhow::Result<()> {
     // like the ones with hyphens like `--metadata`,
     // even though I've set [`Arg::allow_hyphen_values`].
     // This is faster anyways.
+    // I also tried a single "subcommand" argument with [`Arg::possible_values`],
+    // but that had the same problem passing through all arguments as well.
     //
     // Furthermore, doing it this way correctly forwards `--help` through to the subcommand
     // instead of `clap` intercepting it and displaying the top-level `--help`.
@@ -128,9 +132,6 @@ fn main() -> anyhow::Result<()> {
     }
 
     // If we didn't get a subcommand, then use `clap` for parsing and error/help messages.
-
-    // For some stupid reason, [`Arg::possible_values`] requires a slice, not an iterator.
-    // let sub_command_names = sub_commands.keys().copied().collect::<Vec<_>>();
     let matches = App::new("C2Rust")
         .version(&*render_testament!(TESTAMENT))
         .author(crate_authors!(", "))
@@ -145,11 +146,9 @@ fn main() -> anyhow::Result<()> {
                     .allow_hyphen_values(true),
             )
         }))
-        // .arg(Arg::with_name("subcommand").possible_values(sub_command_names.as_slice()))
         .get_matches();
     let sub_command_name = matches
         .subcommand_name()
-        // .value_of("subcommand")
         .ok_or_else(|| anyhow!("no subcommand"))?;
     sub_commands[sub_command_name].invoke(args)
 }
--- a/c2rust/src/main.rs
+++ b/c2rust/src/main.rs
@@ -46,13 +46,11 @@ impl SubCommand {
                 continue;
             }
             let file_name = entry.file_name();
-            let file_name: &Utf8Path = Path::new(&file_name).try_into()?;
-            let file_name = file_name.as_str();
-            let name = match file_name.strip_prefix(c2rust_name) {
-                Some(name) => name,
-                None => continue,
-            };
-            let name = match name.strip_prefix('-') {
+            let name = match Path::new(&file_name)
+                .to_str()
+                .and_then(|name| name.strip_prefix(c2rust_name))
+                .and_then(|name| name.strip_prefix('-'))
+            {
                 Some(name) => name,
                 None => continue,
             };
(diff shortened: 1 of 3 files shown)


c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
-            .args(args)
-            .status()
-            .expect("SubCommand failed to start")
-            .code()
-            .unwrap_or(-1),
-    );
+fn main() -> anyhow::Result<()> {
+    let sub_commands = SubCommand::all()?.collect::<Vec<_>>();
+    let sub_commands = sub_commands
+        .iter()
```
rinon, line 107: Why make a vec then immediately iterate? Can't you just do `all()?.map(...)`?

c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
+    // like the ones with hyphens like `--metadata`,
+    // even though I've set [`Arg::allow_hyphen_values`].
+    // This is faster anyways.
+    // I also tried a single "subcommand" argument with [`Arg::possible_values`],
+    // but that had the same problem passing through all arguments as well.
+    //
+    // Furthermore, doing it this way correctly forwards `--help` through to the subcommand
+    // instead of `clap` intercepting it and displaying the top-level `--help`.
+    let mut args = env::args_os();
+    args.next(); // skip exe name
```
rinon, line 123: You probably want to skip the second `args.next()` if the first one was `None`. Technically I think this is possible if you `fork()` and `exec()` manually, passing an empty `argv` array. Not realistic but good practice to handle that regardless. Probably can just skip straight to the clap arg parsing then.

c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
+        .and_then(|arg| arg.to_str())
+        .and_then(|name| sub_commands.get(name));
+
+    if let Some(sub_command) = sub_command {
+        return sub_command.invoke(args);
+    }
+
+    // If we didn't get a subcommand, then use `clap` for parsing and error/help messages.
+    let matches = App::new("C2Rust")
+        .version(&*render_testament!(TESTAMENT))
```
rinon, line 136: Not for this change, but maybe we should add this version thing to the other subcommands. I don't see it for c2rust-transpile at least.

c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
+            let name = match file_name.strip_prefix(c2rust_name) {
+                Some(name) => name,
+                None => continue,
+            };
+            let name = match name.strip_prefix('-') {
+                Some(name) => name,
+                None => continue,
+            };
+            let path = entry.path();
+            if !path.is_executable() {
```
rinon, line 60: Is this worth adding a whole dependency? We will get an error if the subcommand isn't executable anyway.

c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
+        let mut sub_commands = Vec::new();
+        for entry in dir.read_dir()? {
+            let entry = entry?;
+            let file_type = entry.file_type()?;
+            if !(file_type.is_file() || file_type.is_symlink()) {
+                continue;
+            }
+            let file_name = entry.file_name();
+            let file_name: &Utf8Path = Path::new(&file_name).try_into()?;
+            let file_name = file_name.as_str();
```
rinon, line 50: Do you need to round-trip through both a `Path` and `Utf8Path` here? Can't you just do `.to_str();`

c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
-        }
-        _ => {
-            eprintln!("{:?}", arg_name);
-            panic!("Could not match subcommand");
+impl SubCommand {
+    /// Find all [`SubCommand`]s adjacent to the current (`c2rust`) executable.
+    /// They are of the form `c2rust-{name}`.
+    pub fn find_all() -> anyhow::Result<Vec<Self>> {
+        let c2rust = env::current_exe()?;
+        let c2rust_name: &Utf8Path = c2rust
```
rinon, line 32: What purpose is the Utf8Path serving here?

c2rust/src/main.rs
```diff
@@ -1,56 +1,154 @@
...
+            let file_name: &Utf8Path = Path::new(&file_name).try_into()?;
+            let file_name = file_name.as_str();
+            let name = match file_name.strip_prefix(c2rust_name) {
+                Some(name) => name,
+                None => continue,
+            };
+            let name = match name.strip_prefix('-') {
+                Some(name) => name,
+                None => continue,
+            };
```
rinon, line 58: Rather than doing two strips, I think it would be clearer to append `-` to the name and strip that prefix once. Maybe a bit less efficient in memory, but we don't care here.

rinon (commented):
This looks good. Can you paste an example of what a bare `c2rust` invocation with no args looks like now? Would save me pulling this down and building it, thanks.

c2rust/src/main.rs
kkysen: No they have to be stored somewhere.  Afterwards they're references to the ones in the `Vec`.  I'm pretty sure I tried your suggestion already and it yelled at me.

kkysen:
```shell
❯ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/c2rust`
C2Rust v0.16.0+904 (7bab0a01b 2022-08-04) dirty 2 modifications
The C2Rust Project Developers <c2rust@immunant.com>

USAGE:
    c2rust <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    analyze
    help          Prints this message or the help of the given subcommand(s)
    instrument
    pdg
    transpile
```

c2rust/src/main.rs
rinon: Hm, I figured the HashMap would do that, but I guess if you're storing the command name as a ref then that won't work. Maybe store the key as a String?

c2rust/src/main.rs
kkysen: If we don't do this check, we'll suggest `c2rust-transpile.d`, etc. as subcommands.

c2rust/src/main.rs
kkysen: I think it's just simpler this way.  Other ways I was running into self-referential data.

c2rust/src/main.rs
kkysen: Yeah, that might be a good idea.

c2rust/src/main.rs
kkysen: I think I used `camino` and `Utf8Path` instead of just `Path` because `Path` and `OsStr` lack a lot of methods, like `strip_prefix` and `starts_with`, and it's kind of impossible to have that functionality in a cross-platform way.  These names are very unlikely not to be UTF-8, so it's just easier to enforce that with `camino::Utf8Path`.

c2rust/src/main.rs
kkysen: `Path::to_str` returns an `Option<&str>`, not a `&str` like `Utf8Path::to_str` does.

c2rust/src/main.rs
kkysen: You're right.  I'll change it to use one `args.nth(1)`, which propagates the first `args.next()` properly.

c2rust/src/main.rs
kkysen: Actually, maybe just skipping the path when `Path::to_str` returns `None` is better than an `Err`, since we're just not interested in the non-UTF8 paths but don't necessarily want to error on one.

c2rust/src/main.rs
kkysen: How about this instead?  https://github.com/immunant/c2rust/blob/5c4867130b5430ee7a343965e07c1375af48e59e/c2rust/src/main.rs#L42-L60

c2rust/src/main.rs
kkysen: How about this instead? https://github.com/immunant/c2rust/blob/5c4867130b5430ee7a343965e07c1375af48e59e/c2rust/src/main.rs#L42-L60

c2rust/src/main.rs
rinon: Fair enough. It's a small thing anyway.

c2rust/src/main.rs
rinon: ahhhh yes, alright.

c2rust/src/main.rs
rinon: Yeah, that looks fine.

c2rust/src/main.rs
rinon: Sure, but if the file is a valid Utf8Path then the it's a valid String, no? Just seems like overkill. It's not necessarily wrong, just fancier than I would expect we would need.

rinon: My reasoning here is that we shouldn't add dependencies for simple things unless they are _extremely_ convenient or we use it a lot. Managing dependencies, keeping them up-to-date, making sure they are maintained and get security fixes, even just downloading and building an extra dep has a cost. Not necessarily a huge deal for this project, but something to weigh against.

Seems we already use camino though? So we can just keep using it here I guess, even though I'd prefer not to.

rinon (approved):

c2rust/src/main.rs
fw-immunant: I agree with the above--I'm not a huge fan of more dependencies but it shouldn't block this PR especially as we are already using `camino` in other c2rust crates. Someone could probably remove all usages of it in without too much work in an independent PR if they wanted.

fw-immunant (approved):

c2rust/src/main.rs
kkysen: Okay, I can see about removing it, here and in other places.  I think I had initially added it since I was using it much more thoroughly, but this one use here can definitely be removed.

kkysen: I'll remove it in another PR, along with the other locations.

kkysen: @fw-immunant, @rinon, this (removing `camino`) is fixed in #644.

Issue #578, opened 2022-08-03
Dynamic instrumentation does not support intermediate `LoadValue` and `LoadAddr` events


aneksteind:
We'd like to treat the following MIR:

```rust
_13 = &raw const (*((*((*_8).2: *mut pointers::S)).2: *mut pointers::S))
```

as something roughly equal to:

```rust
1: p = (*_8).2: *mut pointers::S
2: q = (*p).2: *mut pointers::S
3: _13 = &raw const (*q)
```

implying that there `LoadAddr` and `LoadValue` events for `(*_8).2` and `(*p).2`

When generating these events, the value of each respective pointer will need to be passed to the instrumentation in order for the corresponding object to be detected. This means that the source needs to be instrumented with assignment statements to locals
that can be passed. For example:

```rust
1: _21 =  (*_8).2
2: field_hook(/* ... info about line 1 ... */)
3: load_addr_hook(/* ... info about line 1 ... */)
4: load_value_hook(/* ... info about line 1 ... */)
5: _22 = (*_21).2: *mut pointers::S
6: field_hook(/* ... info about line 5 ... */)
7: load_addr_hook(/* ... info about line 5 ... */)
8: load_value_hook(/* ... info about line 5 ... */)
9: _13 = &raw const (*((*((*_8).2: *mut pointers::S)).2: *mut pointers::S))
10: copy_addr_hook(/* ... info about line 9 ... */)
```

aneksteind:
There is an additional challenge to this: instrumentation points are created before statements are inserted. This means that `... info about line 1 ...` is (at that point), impossible to determine at point-building time (it doesn't exist yet)

aneksteind:
I've added test cases that demonstrate the issue in #568. The following PDG node is generated there:

```rust
n[0]: &_8 _ => _14 @ bb0[39]: fn test_ref_field_addr; _14 = &raw const ((*((*_8).2: *mut pointers::S)).2: *mut pointers::S);
```

This is a both a bug (`AddrOf` node is incorrect) and does not produce all the events listed above

Issue #570, opened 2022-08-02
`lighttpd` instrumentation crashes on `Adt`


kkysen:
`master` is failing to instrument [`lighttpd`](https://github.com/immunant/lighttpd-rust), panicking because `TyKind::Adt` is not handled in `ArgKind::from_type`/`ArgKind::from_type` is used in contexts it's not meant to handle.

```shell
> ./scripts/pdg.sh ../lighttpd/rust/
...
thread 'rustc' panicked at 'operand is not of integer-castable type: op = _4, ty = src::configfile::config_t, ty.kind(): Adt(src::configfile::config_t, [])', dynamic_instrumentation/src/arg.rs:30:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/library/core/src/panicking.rs:143:14
   2: c2rust_instrument::arg::ArgKind::from_type
             at /home/kkysen/work/rust/c2rust/dynamic_instrumentation/src/arg.rs:30:13
   3: c2rust_instrument::point::build::InstrumentationBuilder::arg_var
             at /home/kkysen/work/rust/c2rust/dynamic_instrumentation/src/point/build.rs:83:42
   4: c2rust_instrument::instrument::<impl rustc_middle::mir::visit::Visitor for c2rust_instrument::point::InstrumentationAdder>::visit_assign::{{closure}}
             at /home/kkysen/work/rust/c2rust/dynamic_instrumentation/src/instrument.rs:181:13
   5: c2rust_instrument::instrument::<impl rustc_middle::mir::visit::Visitor for c2rust_instrument::point::InstrumentationAdder>::visit_assign
             at /home/kkysen/work/rust/c2rust/dynamic_instrumentation/src/instrument.rs:190:17
   6: rustc_middle::mir::visit::Visitor::super_statement
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/compiler/rustc_middle/src/mir/visit.rs:381:25
   7: rustc_middle::mir::visit::Visitor::visit_statement
             at /rustc/1e12aef3fab243407f9d71ba9956cb2a1bf105d5/compiler/rustc_middle/src/mir/visit.rs:95:17
[...]

PR #582, opened 2022-08-04
check dereferenced type before passing to load-addr hook

aneksteind:
This is insufficient for reasons outlined [here](https://github.com/immunant/c2rust/issues/570#issuecomment-1205834908) (addressed via #578), but should hopefully let us fix regressions.


kkysen (approved):
Looks good for a temporary fix.

PR #583, opened 2022-08-05
feed original MIR location to PDG node

### LINKED ISSUE #581: PDG references location (block id and statement id) of instrumentation, but should be using the location of original MIR
```
n[15]: copy        n[3]  => _14    @ bb4[25]: fn simple; _14 = &raw const (*_1);                  
n[16]: field.2     n[3]  => _      @ bb4[25]: fn simple; ((*_1).2: *const pointers::S) = move _14;
```
Followed by the same statement at two different positions:
```
n[17]: addr.store  n[16] => _      @ bb4[25]: fn simple; ((*_1).2: *const pointers::S) = move _14;
n[18]: value.store n[15] => _1.*.2 @ bb4[26]: fn simple; ((*_1).2: *const pointers::S) = move _14;
```

In the example above, there are statements that are different but show the same location, and identical statements that show different locations. This is due to the fact that the PDG is keeping track of where the instrumentation was supposed to go, and not where the original MIR was.

aneksteind:
Fixes #581 by adding the original MIR location to the instrumentation point build process. Extracts that location at instrumentation time when creating MIR metadata in `get_mir_loc_idx`


dynamic_instrumentation/src/instrument.rs
```diff
@@ -22,8 +22,8 @@ use crate::arg::{ArgKind, InstrumentationArg};
 use crate::hooks::Hooks;
 use crate::mir_utils::{has_outer_deref, remove_outer_deref, strip_all_deref};
 use crate::point::cast_ptr_to_usize;
-use crate::point::InstrumentationAdder;
 use crate::point::InstrumentationApplier;
+use crate::point::InstrumentationPointCollectorVisitor;
```
fw-immunant, line 26: I hate to nitpick when I'd already commented on this point, but `InstrumentationPointCollectorVisitor` stutters in a way that makes it sound like it's a visitor of `InstrumentationPointCollector`s. I think a verb-like name is ideal, as a visitor class/struct exists for the behavior (verb) in its `Visitor` impl. So I would recommend `CollectInstrumentationPoints` or `CollectInstrumentationPointsVisitor`.

A high-level perspective on this is that the whole visitor is an alternative way to write a `collect_instrumentation_points` function.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -428,7 +424,7 @@ fn instrument_body<'a, 'tcx>(
     body_did: DefId,
 ) {
     let hooks = Hooks::new(tcx);
-    let mut adder = InstrumentationAdder::new(tcx, hooks, body);
+    let mut adder = InstrumentationPointCollectorVisitor::new(tcx, hooks, body);
```
fw-immunant, line 427: This variable name should also be updated, perhaps to `visitor`, `collector`, or `collect_points`.

fw-immunant (changes requested):
Location changes look good, but some additional comments on naming that I'd like addressed.

fw-immunant (approved):

kkysen (changes requested):
A few comments:

* If this fixes #581, then it needs to say `fixes #581` or one of the other closing keywords, not `addresses`.  I fixed that here and in another PR.
* The renaming changes should be in a separate PR.  As far as I can tell, they have nothing to do with the functional changes in this PR.
* From what I can tell, `debug_mir` isn't necessary anymore (or it could take no argument) as that information is already there in the `original_location`.
* It'd be helpful if you could give a bit of a description for the PRs.  If the the issue is a description of the bug, then the PR should have a description of how this fixes it, things considered, any other changes, etc.

dynamic_instrumentation/src/point/build.rs
```diff
@@ -24,19 +24,20 @@ struct InstrumentationPointBuilder<'tcx> {
     pub metadata: EventMetadata,
 }
 
-impl<'tcx> InstrumentationAdder<'_, 'tcx> {
-    fn add(&mut self, point: InstrumentationPointBuilder<'tcx>, loc: Location, func: DefId) {
+impl<'a, 'tcx> InstrumentationAdder<'a, 'tcx> {
+    fn add(&mut self, builder: InstrumentationBuilder<'a, 'tcx>) {
```
kkysen, line 28: I don't think `add` should take a whole `InstrumentationBuilder`, as that passes extra things that `add` doesn't need, like some fields and a lifetime.  The lifetime isn't an issue now, but I think minimizing lifetimes where possible is generally very useful, and once things get hairy with lifetimes, it gets hard.  Instead, we can just change the `loc: Location` arg to `original_location: Location, instrumentation_location: Location`.

kkysen (changes requested):
Looks all good now except for the `fn add` comment.

kkysen (approved):

PR #584, opened 2022-08-05
track assignment of references to constants and pointers

### LINKED ISSUE #577: Dynamic instrumentation does not track `Assignment` with `Rvalue::Use(Operand::Constant(..))`
in cases where we have MIR such as
```rust
1: let _5: &[u8; 4]; 
2:    bb0: {
3:        _5 = const b"%i\n\x00";
4:        _4 = &raw const (*_5);
```

We want to treat the operation on line 4 as a copy of a reference, but the source of that event isn't captured because line 3 isn't instrumented.

aneksteind:
Fixes #577.

### DIFF
--- a/dynamic_instrumentation/src/instrument.rs
+++ b/dynamic_instrumentation/src/instrument.rs
@@ -257,6 +257,8 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
                     .add_to(self);
             }
             Rvalue::Use(Operand::Constant(..))=> {
+                // Track (as copies) assignments that give local names to constants so that code
+                // taking references to said constants can refer to these assignments as sources
                 self.loc(location.successor_within_block(), copy_fn)
                     .arg_var(dest)
                     .dest(&dest)
--- a/dynamic_instrumentation/src/instrument.rs
+++ b/dynamic_instrumentation/src/instrument.rs
@@ -256,6 +256,13 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
                     .debug_mir(location)
                     .add_to(self);
             }
+            Rvalue::Use(Operand::Constant(..))=> {
+                self.loc(location.successor_within_block(), copy_fn)
+                    .arg_var(dest)
+                    .dest(&dest)
+                    .debug_mir(location)
+                    .add_to(self);
+            }
             Rvalue::Use(Operand::Copy(p) | Operand::Move(p)) => {
                 self.loc(location.successor_within_block(), copy_fn)
                     .arg_var(dest)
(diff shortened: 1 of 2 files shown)


dynamic_instrumentation/src/instrument.rs
```diff
@@ -256,6 +256,13 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
                     .debug_mir(location)
                     .add_to(self);
             }
+            Rvalue::Use(Operand::Constant(..))=> {
+                self.loc(location.successor_within_block(), copy_fn)
```
aneksteind, line 260: this will be replaced by `AddrOfStatic` when we begin support for it

dynamic_instrumentation/src/instrument.rs
```diff
@@ -256,6 +256,13 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
                     .debug_mir(location)
                     .add_to(self);
             }
+            Rvalue::Use(Operand::Constant(..))=> {
```
fw-immunant, line 259:
```suggestion
            Rvalue::Use(Operand::Constant(..))=> {
                // Track (as copies) assignments that give local names to constants so that code
                // taking references to said constants can refer to these assignments as sources
```

fw-immunant: I think this is necessary to explain why this change is correct--otherwise, it isn't very clear why we would instrument such statically-known behavior. Is this explanation sensical?

spernsteiner: Based on the other comment on the diff, I think this is for future `AddrOfStatic` support.  In the MIR, taking the address of a static looks like `_1 = const {alloc1: &usize};`, which is `Use(Constant(..))` with a constant of type `&'static usize` whose value is a pointer to a static allocation.

spernsteiner:
Snapshot changes look okay to me.

dynamic_instrumentation/src/instrument.rs
fw-immunant: OK, then let's have a comment to that effect.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -256,6 +256,16 @@ impl<'tcx> Visitor<'tcx> for InstrumentationAdder<'_, 'tcx> {
...
```
kkysen, line 259:
```suggestion
            Rvalue::Use(Operand::Constant(..)) => {
```

`cargo fmt --check` is on in CI now.

kkysen: If you merge/rebase to `master`, you can then run `./scripts/run_ci_checks.sh` to check the same things locally.

kkysen (approved):
This looks great!  The pdg is way easier to read and reason about now that the `const`s are there.

There's just the little formatting issue; once it passes CI, this should be good to go.

PR #585, opened 2022-08-05
add pointer-to-pointer test cases

aneksteind:


analysis/test/src/pointers.rs
```diff
@@ -536,6 +561,20 @@ unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> lib
...
+
+    let mut x: i32 = 0; 
+    let mut p_x = std::ptr::addr_of_mut!(x);
+    let pp_x: *mut *mut i32 = std::ptr::addr_of_mut!(p_x);
+
+    let mut y: i32 = 0; 
+    let mut p_y = std::ptr::addr_of_mut!(y);
+    let pp_y: *mut *mut i32 = std::ptr::addr_of_mut!(p_y);
+
+    ptrptr1_backward(true, pp_x, pp_y);
```
aneksteind, line 573: @spernsteiner let me know if it's better to have a unique set of pointers for each call

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -959,6 +959,50 @@ g {
 }
 nodes_that_need_write = [75, 74, 73, 66, 65, 64, 63, 62, 61, 54, 53, 52, 45, 44, 43, 33, 32, 31, 27, 26, 25, 15, 14, 13, 6, 5, 4, 0]
 
+g {
+	n[0]: &_43 _ => _44 @ bb32[9]: fn main_0; _44 = &raw mut _43;
+}
```
spernsteiner, line 964: Since `_44` is address-taken, this graph should look like `AddrOfLocal(_43) -> StoreValue`, and there should be a corresponding `StoreAddr` in the graph below that starts with `AddrOfLocal(_44)`.

The `p = *_4; *p = 1` code should also produce a `LoadValue -> p -> StoreAddr` chain somewhere, possibly in this graph (since `p` will concretely equal `&_43`).

analysis/test/src/pointers.rs
spernsteiner: Shouldn't make much difference, I think.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -959,6 +959,50 @@ g {
...
+nodes_that_need_write = []
+
+g {
+	n[0]:  &_44       _     => _45 @ bb32[12]: fn main_0;           _45 = &raw mut _44;                                     
+	n[1]:  copy       n[0]  => _50 @ bb32[28]: fn main_0;           _50 = _45;                                              
+	n[2]:  copy       n[1]  => _2  @ bb0[0]:   fn ptrptr1_backward; _49 = ptrptr1_backward(const true, move _50, move _51); 
+	n[3]:  copy       n[2]  => _4  @ bb1[1]:   fn ptrptr1_backward; _4 = _2;                                                
+	n[4]:  addr.store n[0]  => _   @ bb4[2]:   fn ptrptr1_backward; (*(*_4)) = const 1_i32;
```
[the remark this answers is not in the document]
aneksteind: thanks, linking #578 which will fix this once addressed

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -959,6 +959,50 @@ g {
...
+	n[4]:  addr.store n[0]  => _   @ bb4[2]:   fn ptrptr1_backward; (*(*_4)) = const 1_i32;                                 
+	n[5]:  copy       n[0]  => _53 @ bb33[6]:  fn main_0;           _53 = _45;                                              
+	n[6]:  copy       n[5]  => _2  @ bb0[0]:   fn ptrptr1_backward; _52 = ptrptr1_backward(const false, move _53, move _54);
+	n[7]:  copy       n[0]  => _56 @ bb34[6]:  fn main_0;           _56 = _45;                                              
+	n[8]:  copy       n[7]  => _2  @ bb0[0]:   fn ptrptr1_bidir;    _55 = ptrptr1_bidir(const true, move _56, move _57);    
+	n[9]:  copy       n[8]  => _4  @ bb1[1]:   fn ptrptr1_bidir;    _4 = _2;                                                
+	n[10]: addr.store n[0]  => _   @ bb4[2]:   fn ptrptr1_bidir;    (*(*_2)) = const 1_i32;                                 
+	n[11]: copy       n[0]  => _59 @ bb35[6]:  fn main_0;           _59 = _45;                                              
+	n[12]: copy       n[11] => _2  @ bb0[0]:   fn ptrptr1_bidir;    _58 = ptrptr1_bidir(const false, move _59, move _60);   
+	n[13]: addr.store n[0]  => _   @ bb4[2]:   fn ptrptr1_bidir;    (*(*_2)) = const 1_i32;
```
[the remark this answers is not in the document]
aneksteind: should this be any different than [here](https://github.com/immunant/c2rust/pull/585/files#r939210844)?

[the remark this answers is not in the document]
spernsteiner: Both places have both problems (wrong `source` and `AddrStore` where it should be `AddrLoad`)

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
aneksteind: Which statement would the `StoreValue` result from? From my understanding it would need a store on the LHS, and a RHS that evaluates to the value `_43`, but I don't see where that is. My guess is that it must be `*p = 1`?

spernsteiner: It should come from the statement `_44 = &raw mut _43`.  Since `_44` is address-taken, we treat this like

```Rust
a = &mut _44;
*a = &raw mut _43;
```

Which produces the chains `AddrOfLocal(_44) -> a -> StoreAddr` and `AddrOfLocal(_43) -> StoreValue`.

aneksteind: That makes sense, thank you

PR #587, opened 2022-08-06
Add info needed for static analysis to PDG `Node`s

boyland-pf:
Checks whether nodes flow to loads/stores/offsets, and whether they are unique.

Uniqueness of node X is determined based on whether there is a node Y which

- is X's ancestor through copies, fields, and offsets
- is a node Z's ancestor through copies, fields, offsets, where the fields are the same between X and Z
- Z is chronologically between X and X's last descendent


pdg/src/main.rs
```diff
@@ -1,6 +1,8 @@
 #![feature(min_specialization)]
 #![feature(rustc_private)]
 #![feature(map_try_insert)]
+#![feature(type_ascription)]
```
kkysen, line 4: Where is this used?  If we don't need it, which we probably don't, we should avoid it.

pdg/src/info.rs
```diff
@@ -0,0 +1,204 @@
+use std::collections::HashMap;
+use crate::graph::{Graphs,GraphId,Graph,NodeId,Node,NodeKind};
+use rustc_index::vec::IndexVec;
+use rustc_middle::mir::Field;
+use std::collections::HashSet;
+use std::fmt::{self,Debug, Formatter, Display};
+
+#[derive(Debug, Eq, PartialEq, Hash, Clone)]
+pub struct NodeInfo {
```
kkysen, line 9: Is `NodeInfo` a descriptive enough name for this?  Did you have any other ideas in mind?  What do others think?

kkysen (changes requested):
Could you first run `cargo fmt`?  I have some other feedback but it'll be cleaner if I comment all of that after it's reformatted.

Also, can you update the PDG snapshot so we can see what running this looks like?  If you run `cargo test` that should fail on the updated snapshot, and then you run `cargo insta review` (need `cargo install cargo-insta` for that) and approve the changes, and then check in the changes to git.

pdg/src/main.rs
```diff
@@ -107,7 +112,7 @@ fn main() -> eyre::Result<()> {
             .map(|node_id| node_id.as_usize())
             .collect::<Vec<_>>();
         if should_print(ToPrint::Graphs) {
-            println!("{graph}");
+            println!("Graph:\n {graph}");
```
kkysen, line 115:
```suggestion
            println!("{graph}");
```

Do we need this?  The `g {` printed is meant to mean it's a `Graph`, like how `n[id]` is for a `NodeId`.  Originally we'd print the `GraphId` similarly, too, but that caused unnecessarily noisy diffs and we didn't really need it so we removed that.

kkysen:
Could you also add your PR description as inline documentation, including doc comments as fitting?

pdg/src/graph.rs
```diff
@@ -119,6 +120,8 @@ pub struct Node {
     pub kind: NodeKind,
     /// The `Node` that produced the input to this operation.
     pub source: Option<NodeId>,
+    /// Additional information generated from the graphs
+    pub node_info: Option<NodeInfo>,
```
kkysen, line 124: I'm not sure this needs to be an `Option` as it's never `None` by the end.  Maybe `augment_with_info` should just return a something new where `NodeInfo` is unconditionally set.  I'll think more about a good way to arrange that.

pdg/src/main.rs
boyland-pf: Sounds good, I think it was just used in a previous version, and I ended up not using it. Will remove.

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::fmt::{self, Debug, Display, Formatter};
+
+///The information checked in this struct is whether nodes flow to loads, stores, and offsets (pos
+///and neg), as well as whether they are unique.
+///Uniqueness of node X is determined based on whether there is a node Y which is X's ancestor
+///through copies, fields, and offsets, also is a node Z's ancestor through copies, fields, offsets,
+///where the fields are the same between X and Z, and where Z is chronologically between X and X's last descendent.
+///If such a node exists, X is not unique.
```
kkysen, line 13:
```suggestion
/// The information checked in this struct is whether nodes flow to loads, stores, and offsets (pos
/// and neg), as well as whether they are unique.
/// Uniqueness of node X is determined based on whether there is a node Y which is X's ancestor
/// through copies, fields, and offsets, also is a node Z's ancestor through copies, fields, offsets,
/// where the fields are the same between X and Z, and where Z is chronologically between X and X's last descendent.
/// If such a node exists, X is not unique.
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+            }
+            add_children_to_vec(g, &desc_seen, &mut to_view);
+        }
+    }
+    return greatest_node_idx;
+}
+
+///Finds the highest-up ancestor of the given node n in g which is reached through Copy, Field, and
+///Offset, and returns its index as well as the Fields through which it is reached, in order
+///(the final element is the Field closest to the returned idx)
```
kkysen, line 130:
```suggestion
/// Finds the highest-up ancestor of the given node n in g which is reached through Copy, Field, and
/// Offset, and returns its index as well as the Fields through which it is reached, in order
/// (the final element is the Field closest to the returned idx)
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+            }
+            _ => break,
+        };
+    }
+    return (n_idx, lineage);
+}
+
+///Looks for a node which proves that the given node n is not unique. If any is found, it's
+///immediately returned (no guarantee of which is returned if multiple violate the uniqueness conditions);
+///otherwise None is returned.
```
kkysen, line 155:
```suggestion
/// Looks for a node which proves that the given node n is not unique. If any is found, it's
/// immediately returned (no guarantee of which is returned if multiple violate the uniqueness conditions);
/// otherwise None is returned.
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+        };
+        write!(f, "{fm},{fl},{fpo},{fno},{funi}")
+    }
+}
+
+fn node_does_mutation(n: &Node) -> bool {
+    match n.kind {
+        NodeKind::StoreAddr | NodeKind::StoreValue => true,
+        _ => false,
+    }
```
kkysen, line 53:
```suggestion
    matches!(n.kind, NodeKind::StoreAddr | NodeKind::StoreValue)
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn node_does_load(n: &Node) -> bool {
+    match n.kind {
+        NodeKind::LoadAddr | NodeKind::LoadValue => true,
+        _ => false,
+    }
```
kkysen, line 60:
```suggestion
    matches!(n.kind, NodeKind::LoadAddr | NodeKind::LoadValue)
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn node_does_pos_offset(n: &Node) -> bool {
+    match n.kind {
+        NodeKind::Offset(x) => x > 0,
+        _ => false,
+    }
```
kkysen, line 67:
```suggestion
    matches!(n.kind, NodeKind::Offset(x) if x > 0)
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn node_does_neg_offset(n: &Node) -> bool {
+    match n.kind {
+        NodeKind::Offset(x) => x < 0,
+        _ => false,
+    }
```
kkysen, line 74:
```suggestion
    matches!(n.kind, NodeKind::Offset(x) if x < 0)
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn add_children_to_vec(g: &Graph, parents: &HashSet<NodeId>, v: &mut Vec<NodeId>) {
+    for (possible_child_id, possible_child_node) in g.nodes.iter_enumerated() {
+        if let Some(srcidx) = possible_child_node.source {
+            if parents.contains(&srcidx) {
+                v.push(possible_child_id);
+            }
+        }
+    }
```
kkysen, line 84:
```suggestion
    v.extend(
        g.nodes
            .iter_enumerated()
            .filter_map(|(id, node)| Some((id, node.source?)))
            .filter(|(_, src_idx)| parents.contains(src_idx))
            .map(|(id, _)| id),
    );
```

pdg/src/graph.rs
```diff
@@ -156,9 +160,13 @@ impl Node {
             statement_idx,
         };
         let fn_ = function;
+        let info = match node_info {
+            None => format!(""),
+            Some(i) => format!("{i}"),
+        };
```
kkysen, line 166: let info = node_info
            .as_ref()
            .map(|i| i.to_string())
            .unwrap_or_default();

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn check_flows_to_node_kind(
+    g: &Graph,
+    n: &NodeId,
+    node_check: fn(&Node) -> bool,
+) -> Option<NodeId> {
+    let mut seen = HashSet::new();
+    let mut to_view = Vec::new();
+    to_view.push(*n);
```
kkysen, line 94:
```suggestion
    let mut to_view = vec![*n];
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+            seen.insert(node_id_to_check);
+            let node_to_check: &Node = g.nodes.get(node_id_to_check).unwrap();
+            if node_check(node_to_check) {
+                return Some(node_id_to_check);
+            } else {
+                add_children_to_vec(g, &seen, &mut to_view);
+            }
+        }
+    }
+    return None;
```
kkysen, line 106:
```suggestion
    None
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+        if !desc_seen.contains(&node_id_to_check) {
+            desc_seen.insert(node_id_to_check);
+            if node_id_to_check.index() > greatest_index {
+                greatest_index = node_id_to_check.index();
+                greatest_node_idx = node_id_to_check;
+            }
+            add_children_to_vec(g, &desc_seen, &mut to_view);
+        }
+    }
+    return greatest_node_idx;
```
kkysen, line 125:
```suggestion
    greatest_node_idx
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+    return None;
+}
+
+fn greatest_desc(g: &Graph, n: &NodeId) -> NodeId {
+    let mut desc_seen = HashSet::<NodeId>::new();
+    let mut to_view = Vec::new();
+    let mut greatest_index = n.index();
+    let mut greatest_node_idx: NodeId = n.clone();
```
kkysen, line 113:
```suggestion
    let mut greatest_node_idx: NodeId = *n;
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+    to_view.push(*n);
```
kkysen, line 114:
```suggestion
    let mut to_view = vec![*n];
    let mut greatest_index = n.index();
    let mut greatest_node_idx: NodeId = n.clone();
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+    return greatest_node_idx;
+}
+
+///Finds the highest-up ancestor of the given node n in g which is reached through Copy, Field, and
+///Offset, and returns its index as well as the Fields through which it is reached, in order
+///(the final element is the Field closest to the returned idx)
+fn calc_lineage(g: &Graph, n: &NodeId) -> (NodeId, Vec<Field>) {
+    let mut lineage = Vec::new();
+    let mut n_idx = n.clone();
```
kkysen, line 133:
```suggestion
    let mut n_idx = *n;
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+            NodeKind::Offset(_) => parent,
+            NodeKind::Copy => parent,
+            NodeKind::Field(f) => {
+                lineage.push(f);
+                parent
+            }
+            _ => break,
+        };
+    }
+    return (n_idx, lineage);
```
kkysen, line 150:
```suggestion
    (n_idx, lineage)
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+///Looks for a node which proves that the given node n is not unique. If any is found, it's
+///immediately returned (no guarantee of which is returned if multiple violate the uniqueness conditions);
+///otherwise None is returned.
+pub fn check_whether_rules_obeyed(g: &Graph, n: &NodeId) -> Option<NodeId> {
+    let (oldest_ancestor, oldest_lineage) = calc_lineage(g, n);
+    let youngest_descendent = greatest_desc(g, n);
+    let mut to_view = Vec::new();
+    to_view.push((oldest_ancestor, oldest_lineage));
```
kkysen, line 160:
```suggestion
    let mut to_view = vec![(oldest_ancestor, oldest_lineage)];
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+        }
+        for (possible_child_id, possible_child_node) in g.nodes.iter_enumerated() {
+            if let Some(srcidx) = possible_child_node.source {
+                if cur_node_id == srcidx {
+                    to_view.push((possible_child_id, lineage.clone()));
+                }
+            }
+        }
+    }
+    return None;
```
kkysen, line 189:
```suggestion
    None
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+pub fn augment_with_info(pdg: &mut Graphs) {
+    let gs: &mut IndexVec<GraphId, Graph> = &mut pdg.graphs;
+    for g in gs.iter_mut() {
+        let mut idx_flow_to_mut = HashMap::new();
+        let mut idx_flow_to_use = HashMap::new();
+        let mut idx_flow_to_pos_offset = HashMap::new();
+        let mut idx_flow_to_neg_offset = HashMap::new();
+        let mut idx_non_unique = HashMap::new();
+        for (idx, _) in g.nodes.iter_enumerated() {
+            if let Some(descmutidx) = check_flows_to_node_kind(&g, &idx, node_does_mutation) {
```
kkysen, line 203:
```suggestion
            if let Some(descmutidx) = check_flows_to_node_kind(g, &idx, node_does_mutation) {
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+                idx_flow_to_mut.insert(idx, descmutidx);
+            }
+            if let Some(descuseidx) = check_flows_to_node_kind(&g, &idx, node_does_load) {
```
kkysen, line 206:
```suggestion
            if let Some(descuseidx) = check_flows_to_node_kind(g, &idx, node_does_load) {
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+                idx_flow_to_use.insert(idx, descuseidx);
+            }
+            if let Some(descposoidx) = check_flows_to_node_kind(&g, &idx, node_does_pos_offset) {
```
kkysen, line 209:
```suggestion
            if let Some(descposoidx) = check_flows_to_node_kind(g, &idx, node_does_pos_offset) {
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+                idx_flow_to_pos_offset.insert(idx, descposoidx);
+            }
+            if let Some(descnegoidx) = check_flows_to_node_kind(&g, &idx, node_does_neg_offset) {
```
kkysen, line 212:
```suggestion
            if let Some(descnegoidx) = check_flows_to_node_kind(g, &idx, node_does_neg_offset) {
```

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+                idx_flow_to_neg_offset.insert(idx, descnegoidx);
+            }
+            if let Some(non_unique_idx) = check_whether_rules_obeyed(&g, &idx) {
```
kkysen, line 215:
```suggestion
            if let Some(non_unique_idx) = check_whether_rules_obeyed(g, &idx) {
```

kkysen:
If the suggestions look good, I can organize them into commits if you want.

kkysen:
The snapshot you checked is not quite the right one.  That's the `*.snap.new` file that's generated when you run `cargo test` and the snapshot fails, the new snapshot. But then you need to `cargo insta review` and accept the changes so that it saves it to the primary `*.snap` snapshot.  When you check it in, you should be able to see the dif against the old `*.snap` file rather than an entirely new file; otherwise, we can't see the diff in GitHub.

Also, try to separate different changes into separate commits.  That's fine even if they're small, as long as they're separate concerns and leave the code in a working state (usually) in between.

boyland-pf:
The changes are helpful, I'll commit them all grouped.

pdg/src/info.rs
```diff
@@ -0,0 +1,229 @@
...
+        };
+        let fno = match self.flows_to_neg_offset {
+            None => "no flow to neg_offset".to_string(),
+            Some(_) => "does flow to neg_offset".to_string(),
+        };
+        let funi = match self.non_unique {
+            None => "unique".to_string(),
+            Some(x) => format!("non unique proven by {}", x),
+        };
+        write!(f, "{fm},{fl},{fpo},{fno},{funi}")
```
kkysen, line 45: This generates too much noise I think.  I think it's better if we only print the rarer cases (i.e. the `Some` cases).

kkysen:
```suggestion
let s = [
            ("mut", self.flows_to_mutation),
            ("load", self.flows_to_load),
            ("+offset", self.flows_to_neg_offset),
            ("-offset", self.flows_to_neg_offset),
            ("non unique by", self.non_unique),
        ]
        .into_iter()
        .filter_map(|(name, node)| Some((name, node?)))
        .format_with(", ", |(name, node), f| f(&format_args!("{name} {node}")));
        write!(f, "{}", s)
```

How about this?

kkysen: Fix the snapshot before accepting this change, as it'll change the snapshot again and I want to see the diff.

boyland-pf: Sounds good. Do you think it should be abbreviated?

spernsteiner:
The analysis `augment_with_info` uses for read/write/offset does `O(N^3)` work on each graph (where `N` is the size of the graph), which will be very slow for reasonably sized inputs like N = 10000.  Specifically, `augment_with_info` visits each node in the graph and calls `check_flows_to_node_kind` on it, `check_flows_to_node_kind` visits each descendant of the input node and calls `add_children_to_vec` on it, and `add_children_to_vec` visits every node in the graph.  I think this analysis can instead be done in `O(N)` time by iterating backwards over the nodes of each graph and propagating the read/write/offset info from each node to its `source` (assuming each node's `source` has a lower node ID than the node itself).

The uniqueness check is also `O(N^3)` (via `augment_with_info` -> `calc_lineage` -> `add_children_to_vec`).  You can at least get this down to `O(N^2)` (or so) by building a map from parent `NodeId` to a list of all its children and using that for `add_children_to_vec`.  And I suspect it's possible to get this down to `O(N)` (or maybe something like `O(V + E)`) with a different design - maybe something involving propagating ranges of earliest/latest use up the tree, and checking for overlaps at each node with multiple children.

boyland-pf:
I'll change it tonight to be O(N), I wanted to first get out a simple version to make sure I'm computing the right thing.

spernsteiner:
For the uniqueness check, I think you've correctly implemented the `UNIQUE` property I described on slack, but the property I described isn't what the static analysis actually needs.  The property computed here, which I wrongly called `UNIQUE`, should instead be called something like `NON_CONFLICTING`.  Then we redefine `UNIQUE` by saying that a node is `UNIQUE` if all its ancestors are `UNIQUE` (or it has no ancestors) and all its children are `NON_CONFLICTING`.  The reason for the ancestor rule is that if the node's `source` isn't `&mut`/`UNIQUE`, then the node itself can't be `UNIQUE`, since you can't take an `&mut` borrow unless the source of the borrow is also `&mut`.  As for the child rule, if we have two conflicting `&mut` borrows, we turn them into `&Cell<T>` instead; however, the conversion from `&mut` to `&Cell<T>` takes a mutable borrow of the parent, so we can't use it on the edges from the parent to the conflicting children - the parent must already itself be `&Cell<T>` (non-`UNIQUE`).

I think this still isn't quite right.  First, I haven't put any thought into handling fields.  Second, I think we need a slightly more precise notion of `NON_CONFLICTING` to handle immutable borrows, which can overlap with each other (but not with mutable borrows) without creating a conflict.

I pushed a few test cases for the uniqueness check on the branch `sp/extra-info-test-cases`.  They're currently failing, but I think should pass under the new definition of `UNIQUE`.  If you can, please add some more tests like this with fields and with immutable borrows, so we can get a clearer idea of how the uniqueness check compares to the actual borrow checker.

kkysen:
@spernsteiner, to clarify, is the `UNIQUE` property supposed to be exactly what can be made `&mut` and satisfy the borrow checker?  And if it does not satisfy the `UNIQUE` property, would it make sense to say it aliases?

Also, instead of marking things as `mut` if they eventually go through a store, can we mark them as `store` similarly to how we mark things as `load` if they go through a load?  Having `load` and `store` makes more sense to me than `load` and `mut`.

Re. `Cell`, what about `RefCell`?  Doesn't `Cell` require you to have a replacement value to use it (or have it be `Copy` or `Default`), and isn't that not always the case?  And if we use `RefCell`, for now, we're ignoring more complex borrowing/ownership schemes that are safe but would panic with a `RefCell`, right?

Since this part of the PR seems more complex and still under work and testing to get things right, how about we remove that from this PR so that we can merge it separately from the `UNIQUE` property code?
And for [`sp/extra-info-test-cases`](https://github.com/immunant/c2rust/compare/pfb/extra-info...sp/extra-info-test-cases), should we open a PR for that into this PR/branch?

spernsteiner:
> to clarify, is the `UNIQUE` property supposed to be exactly what can be made `&mut` and satisfy the borrow checker?

Yes, that's right.

> How does this handle interior mutability, such as atomics? Or is that (atomic accesses) just not handled at all at the moment?

Types like `AtomicUsize` are no different from `Cell<T>` for our purposes.  We don't have any special handling for interior mutability in the input program.  On the static analysis side, we suggest a rewrite using `Cell<T>` in cases where a pointer is `WRITE` but not `UNIQUE`, which means it's the target of a store but can't be made `&mut`.

> Re. `Cell`, what about `RefCell`? Doesn't `Cell` require you to have a replacement value to use it (or have it be `Copy` or `Default`), and isn't that not always the case?

All C types are `Copy`.  The only non-`Copy` types we'll be introducing are `&mut T` and `Box<T>`, and for those I think we can use something like `Cell<Option<Box<T>>>` to give it a `Default` (where "borrows" actually `take()` the `Box<T>`, and concurrent accesses will panic, similar to `RefCell` but without the layout change).

> Since this part of the PR seems more complex and still under work and testing to get things right, how about we remove that from this PR so that we can merge it separately from the `UNIQUE` property code?

Seems reasonable to me.

> And for [`sp/extra-info-test-cases`](https://github.com/immunant/c2rust/compare/pfb/extra-info...sp/extra-info-test-cases), should we open a PR for that into this PR/branch?

I guess that would work, but since it's only a single commit, I figured it would be simpler to just `git fetch && git cherry-pick origin/sp/extra-info-test-cases`

kkysen:
> Types like `AtomicUsize` are no different from `Cell<T>` for our purposes. We don't have any special handling for interior mutability in the input program. On the static analysis side, we suggest a rewrite using `Cell<T>` in cases where a pointer is `WRITE` but not `UNIQUE`, which means it's the target of a store but can't be made `&mut`.
> 
> > Re. `Cell`, what about `RefCell`? Doesn't `Cell` require you to have a replacement value to use it (or have it be `Copy` or `Default`), and isn't that not always the case?
> 
> All C types are `Copy`. The only non-`Copy` types we'll be introducing are `&mut T` and `Box<T>`, and for those I think we can use something like `Cell<Option<Box<T>>>` to give it a `Default` (where "borrows" actually `take()` the `Box<T>`, and concurrent accesses will panic, similar to `RefCell` but without the layout change).

All C types are `Copy`, but that doesn't mean they should be copied as they could be quite large, and wouldn't have been made `Copy` if they were written in Rust.

Don't we also need to account for partially human-refactored code?  We also **definitely** don't want to make transformations that make human refactoring harder either, as that's still always going to be needed and is still the primary use case (dtolnay emphasized this being important, too).

> > And for [`sp/extra-info-test-cases`](https://github.com/immunant/c2rust/compare/pfb/extra-info...sp/extra-info-test-cases), should we open a PR for that into this PR/branch?
> 
> I guess that would work, but since it's only a single commit, I figured it would be simpler to just `git fetch && git cherry-pick origin/sp/extra-info-test-cases`

I just find being able to see and comment on it on GitHub generally easier.

spernsteiner:
> All C types are `Copy`, but that doesn't mean they should be copied as they could be quite large, and wouldn't have been made `Copy` if they were written in Rust.

There are some tricks we can use there, like replacing `&Cell<Foo>` with `&FooCell`, where `FooCell` is like `Foo` but each field is wrapped in `Cell`.  But the important point is that it's semantically legal to do those copies; we can add optimizations beyond that later as needed.  (Note LLVM can already see through `Cell::get` and `memcpy` - for example, the optimized LLVM IR for [this code](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=e1f03db9d95fad39d0107e7e126f14c9) has only one 64-bit load and one 8-bit load.)

>Don't we also need to account for partially human-refactored code?

I think we should avoid breaking human-edited code, but otherwise shouldn't promise that the tool will actually do anything useful on that code.  The analysis and refactoring problems here are already hard enough.  If the user manually converts a struct field to `Vec<T>` or something like that, I think it's reasonable to require them to also update all uses of that field, or maybe even all uses of the struct in general.

kkysen:
> I think we should avoid breaking human-edited code, but otherwise shouldn't promise that the tool will actually do anything useful on that code. The analysis and refactoring problems here are already hard enough. If the user manually converts a struct field to `Vec<T>` or something like that, I think it's reasonable to require them to also update all uses of that field, or maybe even all uses of the struct in general.

I was more thinking about human refactoring after doing this partial lifting.  If it would make more sense to a human to turn something into a `RefCell` vs. a `Cell` and we turn it into a `Cell`, I want to make sure it's not harder to refactor `Cell` to `RefCell` than original unsafe code to `RefCell`.

boyland-pf:
I found It's easier to efficiently calculate uniqueness as defined directly. The fact that if A and B conflict you no longer have to check the descendants of A and B allows for a merely linear number of comparisons of ranges (though it's still O(nlogn)). Once I'm done with the slide presentation I'll be adding the additional tests to demonstrate how it's currently dealing with fields.

boyland-pf:
A high level explanation is here: https://docs.google.com/document/d/1gPkUOLT4I_thFb6LRXSAOpUIVqDQU1MXuYWLMyy50tc/edit?usp=sharing.

kkysen:
> A high level explanation is here: https://docs.google.com/document/d/1gPkUOLT4I_thFb6LRXSAOpUIVqDQU1MXuYWLMyy50tc/edit?usp=sharing.

Can we inline that here?

boyland-pf:
The goal is to determine nodes which cannot be mutable references.

As shown below, a node cannot be a mutable reference if it:

1. has a conflict with another node
2. is a descendent of a node with a conflict
3. is the direct parent of a node with a conflict.

A node A has a conflict with another node B if:

- they may point to the same address with the same type (assuming, like the rust compiler, we can’t do range analysis) 
- A is not a descendent of B or vice versa
- B is used while A is still live

If A has a conflict with B, they are distinct and live at the same time, so neither of them can be a mutable reference.
Mutable references cannot be safely created from unsafe pointers, so none of the descendents of A or B can be mutable references.

Reasoning:

- If A has a descendant B, then A lives at least as long as B, and comes into existence earlier than B.
- Two nodes that conflict must have a common ancestor.
- If A has a child B which has a conflict with C, then either A is the closest common ancestor of B and C or not
- If it is, then A can’t be a mutable reference because it’s used both for the B and C branches.
- If it isn’t, then A can’t be a mutable reference it itself will conflict with C


Therefore parents of conflicting nodes cannot be mutable references.

To find all pointers that cannot be mutable references, the code:

* Groups up nodes by which of them can alias to each other, and organizes each group into shrubs.
  - If A is descended from B but isn’t the same address (or an offset), no descendent of it can be.
  - So the nodes of any individual alias set can be organized into a number of shrubs which are all cousins of each other.
* Sorts the shrub roots by onset, and compares each root to the latest descendent of all previous roots, keeping note of any conflicts seen.
  - If a node X in shrub A has a conflict with a node Y in shrub B, then because A lasts longer than X and Y starts earlier than B, A will also have a conflict with B.
  - If A has a conflict with B>A, then the latest descendent when B is examined must be at least A’s latest descendent, so B will have a conflict with something, and will correctly be labeled as a conflicting node.
* All descendents of these nodes are labeled as impossible to be mutable references.
  - If A has a conflict with B, then a descendent X of A may or may not have a conflict with any descendent Y of B, but by rule 2 it cannot be a mutable reference.
  - By rule 3, all parents of nodes that are known to not be mutable references are also labeled as such.
  - There are no false positives because the only non-conflicting nodes which are labeled as not-mutable-references are from step 3, and so necessarily only have parents who are already ruled out of being mutable references.

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
+use crate::graph::{Graph, GraphId, Graphs, Node, NodeId, NodeKind};
+use itertools::Itertools;
+use rustc_index::vec::IndexVec;
+use rustc_middle::mir::Field;
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::cmp;
+use std::fmt::{self, Debug, Display, Formatter};
+
+/// The information checked in this struct is whether nodes flow to loads, stores, and offsets (pos
```
aneksteind, line 10: Is one node "flowing" to another node semantically the same as there being a directional path from one node to another?

aneksteind: Also curious if a node flows to another if it's only a descendant, only and ancestor, or either

boyland-pf: Yes, node A flows to node B if there's a path from B to A along "source" edges.

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+fn create_flow_info (g: &Graph) -> HashMap<NodeId,GraphTraverseInfo> {
+    let mut f = HashMap::from_iter(
+        g.nodes.iter_enumerated()
+        .map(|(idx,node)| (idx,init_traverse_info(idx,node)))
+    );
+    for (n_id,node) in g.nodes.iter_enumerated().rev(){ 
+        let cur : GraphTraverseInfo = *(f.get(&n_id).unwrap());
+        if let Some(p_id) = node.source {
+            let parent = f.get_mut(&p_id).unwrap();
+            parent.last_descendent = cmp::max(cur.last_descendent,parent.last_descendent);
```
aneksteind, line 58: Correct me if I'm missing something... since you're going in reverse order which is node id descending, I think it's enough to set this once iff it's not set, rather than doing a comparison

aneksteind: it's also not a big deal if the above is true, what you have is simpler and gets the point across

aneksteind: on the other hand, if you didn't go in reverse order, you could just set `parent.last_descendent = cur.last_descendent` without any checking, but that might require changes in the flow logic below. why do you traverse in reverse?

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+    f
+}
+
+fn collect_children (g: &Graph) -> HashMap<NodeId,Vec<NodeId>> {
+    let mut m = HashMap::new();
+    for (par,chi) in g.nodes.iter_enumerated().filter_map(|(idx,node)| Some((node.source?,idx))){
+        m.entry(par).or_insert_with(Vec::new).push(chi)
+    };
+    for (par,chi) in g.nodes.iter_enumerated(){
+        m.try_insert(par,Vec::new());
```
aneksteind, line 74: i think this loop should happen first, and the second loop should just push to the existing vector

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+    };
+    m
+}
+
+fn partition_into_alias_sets (g: &Graph) -> HashMap<(NodeId,Vec<Field>),Vec<NodeId>> {
+    let mut store_seen : HashMap<NodeId,(NodeId,Vec<Field>)> = HashMap::new();
+    let mut store_roots : HashMap<(NodeId,Vec<Field>),Vec<NodeId>> = HashMap::new();
+
+    for n_id in g.nodes.indices(){
+        let node = g.nodes.get(n_id).unwrap();
```
aneksteind, line 84: prefer to iterate over the collection that's enumerated, so that you don't need to unwrap here

pdg/src/info.rs
boyland-pf: > Correct me if I'm missing something... since you're going in reverse order which is node id descending, I think it's enough to set this once iff it's not set, rather than doing a comparison

If we have A, B, C, D, E, with A->B, A->C, B->E, C->D, then C will be encountered before B, and its last descendent is D, but the correct last descendent of A is E.

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+    for (par,chi) in g.nodes.iter_enumerated().filter_map(|(idx,node)| Some((node.source?,idx))){
+        m.entry(par).or_insert_with(Vec::new).push(chi)
+    };
+    for (par,chi) in g.nodes.iter_enumerated(){
+        m.try_insert(par,Vec::new());
+    };
+    m
+}
+
+fn partition_into_alias_sets (g: &Graph) -> HashMap<(NodeId,Vec<Field>),Vec<NodeId>> {
```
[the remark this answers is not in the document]
aneksteind: i'm actually not totally sure what this is doing

pdg/src/info.rs
aneksteind: ah, thanks for the counter-example

pdg/src/info.rs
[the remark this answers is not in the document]
aneksteind: is it gathering all field projections seen from each node and its descendants?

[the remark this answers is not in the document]
boyland-pf: Yes. But it turns out this is not needed, so my most recent version doesn't use it.

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+
+fn check_sibling_conflict(siblings: &mut Vec<NodeId>, flow_info: &HashMap<NodeId,GraphTraverseInfo>, conflict_result: &mut HashMap<NodeId,NodeId>){
+    let mut max_desc : NodeId = *siblings.get(0).unwrap();
+    let mut max_desc_parent : NodeId = *siblings.get(0).unwrap();
+    for id in siblings {
+        if *id < max_desc {
+            conflict_result.insert(max_desc_parent,*id);
+            conflict_result.insert(*id,max_desc_parent);
+        }
+        if flow_info.get(&id).unwrap().last_descendent > max_desc {
```
aneksteind, line 115: `flow_info.get(&id).unwrap().last_descendent` can be put into a variable rather than calling twice

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+        };
+    };
+    store_roots
+}
+
+fn check_sibling_conflict(siblings: &mut Vec<NodeId>, flow_info: &HashMap<NodeId,GraphTraverseInfo>, conflict_result: &mut HashMap<NodeId,NodeId>){
+    let mut max_desc : NodeId = *siblings.get(0).unwrap();
+    let mut max_desc_parent : NodeId = *siblings.get(0).unwrap();
+    for id in siblings {
+        if *id < max_desc {
```
aneksteind, line 111: it seems like this assumes some invariant for `siblings`. It would be good to document what that is, add an assertion, or encode it in the type of `siblings` so that the user doesn't accidentally call this function with input that's semantically incorrect.

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+                    },
+                    _ => store_seen.insert(n_id,(parent_root, parent_fields)),
+                }
+            }
+        };
+    };
+    store_roots
+}
+
+fn check_sibling_conflict(siblings: &mut Vec<NodeId>, flow_info: &HashMap<NodeId,GraphTraverseInfo>, conflict_result: &mut HashMap<NodeId,NodeId>){
```
aneksteind, line 107: also needs documentation -- what's a conflict? I see it in the PR description but it would be useful here as well

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+/// the correct flow information.
+pub fn augment_with_info(pdg: &mut Graphs) {
+    let gs: &mut IndexVec<GraphId, Graph> = &mut pdg.graphs;
+    for g in gs {
+        let flow_info = create_flow_info(g);
+        let mut conflicting = determine_non_conflicting(&g,&collect_children(&g),&flow_info);
+        for (idx,node) in g.nodes.iter_enumerated_mut(){
+            let node_flow = flow_info.get(&idx).unwrap();
+            node.node_info = Some(NodeInfo {
+                flows_to_mutation: node_flow.flows_to_store,
```
aneksteind, line 315: why the difference in naming (mutation vs. store)?

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+                flows_to_load: node_flow.flows_to_load,
+                flows_to_pos_offset: node_flow.flows_to_pos_offset,
```
aneksteind, line 317: should these 4 attributes be a part of their own `FlowInfo` struct?

pdg/src/info.rs
boyland-pf: Okay, will add that in.

pdg/src/info.rs
boyland-pf: This is an inconsistency that's being removed (to just be "store")

pdg/src/info.rs
boyland-pf: @kkysen mentioned I ought to do that, removes some code redundancy.

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+            node_info: None,
+            debug_info: "".into(),
+        })
+    }
+
+    fn mk_addr_of_local(g: &mut Graph, local: impl Into<Local>) -> NodeId {
+        mk_node(g, NodeKind::AddrOfLocal(local.into()), None)
+    }
+
+    fn mk_copy(g: &mut Graph, source: NodeId) -> NodeId {
```
aneksteind, line 352: these functions might be useful to split into a separate module

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
```diff
@@ -3,970 +3,970 @@ source: pdg/src/main.rs
...
-	n[72]: addr.load  n[71] => _   @ bb12[2]:  fn insertion_sort; _20 = (*_21);                                        
-	n[73]: copy       n[6]  => _47 @ bb24[8]:  fn insertion_sort; _47 = _2;                                            
-	n[74]: offset[4]  n[73] => _46 @ bb24[13]: fn insertion_sort; _46 = offset(move _47, move _48);                    
-	n[75]: addr.store n[74] => _   @ bb25[2]:  fn insertion_sort; (*_46) = move _45;                                   
+	n[0]:  &_35       _     => _34 @ bb30[4]:  fn main_0;         (mut n[75], load n[72]) _34 = &mut _35                                      
+	n[1]:  copy       n[0]  => _40 @ bb30[12]: fn main_0;         ()                      _40 = &(*_34)                                       
+	n[2]:  copy       n[1]  => _39 @ bb30[13]: fn main_0;         ()                      _39 = move _40 as &[i32] (Pointer(Unsize))          
+	n[3]:  copy       n[2]  => _1  @ bb0[0]:   fn len;            ()                      _38 = len(move _39)                                 
+	n[4]:  copy       n[0]  => _42 @ bb31[6]:  fn main_0;         (mut n[75], load n[72]) _42 = &raw mut (*_34)
```
aneksteind, line 887: i feel like the surrounding `n[_]` might be a little too verbose and redundant -- the node id's should be enough

pdg/src/info.rs
```diff
@@ -0,0 +1,515 @@
...
+    None
+}
+
+/// Takes a list of graphs, creates a NodeInfo object for each node in each graph, filling it with
+/// the correct flow information.
+pub fn augment_with_info(pdg: &mut Graphs) {
+    let gs: &mut IndexVec<GraphId, Graph> = &mut pdg.graphs;
+    for g in gs {
+        let flow_info = create_flow_info(g);
+        let mut conflicting = determine_non_conflicting(&g,&collect_children(&g),&flow_info);
```
aneksteind, line 311: `conflicting = determine_non_conflicting` is definitely confusing to me. is one of them misnamed?

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot.snap
boyland-pf: Sounds good.

kkysen: If we change that, I think we should change it everywhere we use `n[_]`.

pdg/src/info.rs
kkysen: Yeah, the 4 flow properties are gonna be put in their own struct and be part of a separate PR (separate from the unique/aliasing one).

PR #589, opened 2022-08-08
Add pointer-to-pointer equivalence class analysis

spernsteiner:
This modifies the static analysis component to compute equivalence classes of pointers that are required to have the same type.  For example, given this code:

```Rust
let x: *mut *mut i32 = ...;
let mut y: *mut *mut i32 = ...;
y = x;
```

The assignment `y = x` requires that the inner pointer types be identical, as there is no safe cast that can turn e.g. `&mut &mut i32` into `&&i32`.  The outer pointer types can still vary (with `y` being a supertype of `x`), so the assignment `x: &mut &mut i32`, `y: &&mut i32` would be valid.

The general strategy is to assign `PointerId`s to all pointer types, compute equivalence classes over `PointerId`s (based on assignments like the one above), and finally reassign `PointerId`s so that pointer types in the same equivalence class use the same `PointerId`.  For example, the above code would first be numbered with unique `PointerId`s like `x: *mut /*1*/ *mut /*2*/ i32`, `y: *mut /*3*/ *mut /*4*/ i32`, then after equivalence classes are computed, it would be renumbered with `x: *mut /*5*/ *mut /*6*/ i32`, `y: *mut /*7*/ *mut /*6*/ i32` - note that both inner pointers now use the same `PointerId`.  The remaining passes can then operate unchanged, assigning permission bits and new types to each `PointerId`, and these results will naturally be shared between all pointer types that use the same `PointerId`.

This branch also splits `PointerId`s into separate local and global namespaces.  The local namespace is for pointer types that are relevant only within a single function body.  The global namespace is for pointer types that may be used in multiple functions, such as struct fields or function argument/return types.  Currently, no global pointers are ever defined, but having this distinction affects the design of the equivalence class data structure.

### DIFF
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -2,6 +2,7 @@ use self::atoms::{AllFacts, AtomMaps, Loan, Origin, Output, Path, SubPoint};
 use crate::context::{AnalysisCtxt, PermissionSet};
 use crate::dataflow::DataflowConstraints;
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
+use crate::pointer_id::PointerTableMut;
 use crate::util::{describe_rvalue, RvalueDesc};
 use rustc_middle::mir::{Body, BorrowKind, Local, LocalKind, Place, StatementKind, START_BLOCK};
 use rustc_middle::ty::{List, TyKind};
@@ -25,7 +26,7 @@ pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label>;
 pub fn borrowck_mir<'tcx>(
     acx: &AnalysisCtxt<'_, 'tcx>,
     dataflow: &DataflowConstraints,
-    hypothesis: &mut [PermissionSet],
+    hypothesis: &mut PointerTableMut<PermissionSet>,
     name: &str,
     mir: &Body<'tcx>,
 ) {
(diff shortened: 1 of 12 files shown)


fw-immunant (approved):
I appreciate how particularly well-factored and readable this PR is--thanks! <3

One place where I'd appreciate rationale comment explaining is on `GlobalAnalysisCtxt::remap_pointers`, explaining why it doesn't need to consult the map.

Also, a couple naming nits:
- `RawPointerTable` is a bit confusing, as it has nothing to do with raw pointers--`PointerTableInner` might be more clear. `LocalPointerTable` and `GlobalPointerTable` don't have this problem as they are both local/global tables and tables pertaining to local/global pointers.
- The names of `enter_function` and `enter_function_with_data` make it seem like there might be some kind of cleanup/`leave` necessary after using the returned `AnalysisCtxt`. As this is not actually a stateful API, `function_context`/`function_context_with_data` or similar would be preferable.

Otherwise, LGTM.

spernsteiner:
Thanks! I made all the changes suggested above.

> The names of `enter_function` and `enter_function_with_data` make it seem like there might be some kind of cleanup/`leave` necessary after using the returned `AnalysisCtxt`. As this is not actually a stateful API, `function_context`/`function_context_with_data` or similar would be preferable.

The usual pattern inside rustc is for `enter` methods to take a closure and do cleanup after the closure returns, so there is no explicit "leave" for the caller to do.  Example: [`QueryContext::enter`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_interface/passes/struct.QueryContext.html#method.enter) (which is used elsewhere in this code).  `enter_function` doesn't fit that pattern, so I renamed it to `function_context` as you suggested.

I'm planning to merge this once the CI passes.

fw-immunant:
> The usual pattern inside rustc is for `enter` methods to take a closure and do cleanup after the closure returns, so there is no explicit "leave" for the caller to do. Example: [`QueryContext::enter`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_interface/passes/struct.QueryContext.html#method.enter) (which is used elsewhere in this code). `enter_function` doesn't fit that pattern, so I renamed it to `function_context` as you suggested.

That makes sense. Thanks for the cleanups; merge away!

PR #592, opened 2022-08-09
Detect installed targets by existence of `rustc --print target-libdir`, not checking unstable error messages

kkysen:
We should detect installed targets by checking for the existence of the directory `$(rustc --print target-libdir)` rather than checking unstable error messages from `rustc /dev/null`.

This still uses `rustc`, so it works whether or not `rustc` is really `rustup` or not and if `rustup` was used for installation.


fw-immunant (approved):
This is much more robust, thanks for digging into this and finding something that works.

PR #594, opened 2022-08-09
Turn on `clippy` in CI

### LINKED ISSUE #477: Catch more compile-time errors in CI

kkysen:
Fixes #477.

### DIFF
--- a/scripts/run_ci_checks.sh
+++ b/scripts/run_ci_checks.sh
@@ -12,12 +12,11 @@ fmt() {
     cargo fmt --check
 }
 
-# Soon to be superceded by the commented out `cargo clippy` below.
 # This is different from `cargo build`
-# as this uses `--all-features` to check everything.
+# as this uses `--all-features` to check everything,
+# and runs extra `clippy` checks.
 check() {
-    cargo check --tests --all-features
-    # cargo clippy --tests --all-features
+    cargo clippy --tests --all-features
 }
 
 doc() {


aneksteind:
I'm conflicted about having this without also having pre-commit hooks, to avoid only finding out about clippy failures during CI, but I think those two together might be overkill at this stage of the project. So, I think that we should instead have only have commit hooks that the developer may selectively ignore locally rather than make continual exceptions for in CI.

kkysen:
How do pre-commit hooks work?  I would like to use those, especially for `cargo fmt`.

aneksteind:
https://pre-commit.com/

kkysen:
> https://pre-commit.com/

I'll look into them more, but that seems to be Python-specific pre-commit hooks.

Can pre-commit hooks be pushed to the repo and by default have everyone use them?

aneksteind:
> I'll look into them more, but that seems to be Python-specific pre-commit hooks.

Yep, this is just to demonstrate the concept. I'm sure an equivalent for our purposes can be found.

> Can pre-commit hooks be pushed to the repo and by default have everyone use them?

That's been my experience in the past, yes

PR #597, opened 2022-08-10
Renamed infos for clarity, consistency, and succinctness

kkysen:
Renamed infos for clarity, consistency, and succinctness.  Specifically,
* `mut` renamed to `store` to match `load`
* `non_unique` renamed to `alias`, as this more closely follows Rust's terminology for aliasing, and is also more succinct


spernsteiner:
`store`/`load` changes look good.  But I think I prefer `non_unique` over `alias`, due to the connection to c2rust-analyze's `PermissionSet::UNIQUE` flag.

kkysen:
Isn't aliases the opposite of unique?  I much prefer `alias` because it's shorter, it's a verb like the others, and it's saying what it does, not what it doesn't.

spernsteiner:
You can have references that alias (in that they point to the same memory location) but are still unique / `&mut`.  For example:

```Rust
let x: &mut i32 = ...;
let y = &mut *x;
f(y);
f(x);
```

`y` aliases `x` and has overlapping lifetime, but both can still be `&mut`, since the compiler can statically track the relationship between the two.

kkysen:
If they alias, then they are not unique, by the common definitions of those words.  I think in the Rust sense, it's understood that you can't actively use mutably aliased references or pointers in that the uniqueness or aliasing property is determined at use time.  Unique and alias are pretty antonyms to me, so I don't think there's any incompatibility between them.  The issue you highlighted above is just as much a problem for calling things unique.  And alias is quite common terminology used in Rust, so I think it makes a lot of sense to use as an antonym to unique, in the aliased xor mutable (unique) sense.

PR #600, opened 2022-08-10
call `debug_mir` once

aneksteind:
`debug_mir` is used everywhere, so call once when constructing the `InstrumentationBuilder`


kkysen (approved):
Looks good, but let's merge once other things land in `master`.

PR #601, opened 2022-08-10
Better `impl`s for some `c2rust-analysis-rt` types

kkysen:
This adds better `impl`s for some `c2rust-analysis-rt` types.  Namely, we
* https://github.com/immunant/c2rust/commit/8230808bc590a45a492312cf8052f944d8132ba6: match `rustc`'s derive `impl`s for mirrored types like `Fingerprint` and `DefPathHash`
* https://github.com/immunant/c2rust/commit/2f596ae50aac109b1eeaa15eb60e2d2265b810e4: extend those same `impl`s to other types `impl`ed based on `rustc`-mirrored types, like `Func` (based on `DefPathHash`)
* https://github.com/immunant/c2rust/commit/9fac146fb2a77d7e862ecb7581041ecfbf7c8722: omit `EventMetadata::debug_info` from the `Eq` and `Hash` `impl`s, as it's just extra info

### DIFF
--- a/analysis/runtime/src/mir_loc.rs
+++ b/analysis/runtime/src/mir_loc.rs
@@ -186,7 +186,7 @@ impl Default for TransferKind {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone, Default)]
+#[derive(Debug, Serialize, Deserialize, Eq, Clone, Default)]
 pub struct EventMetadata {
     /// Input [`Local`]s for an [`Event`](crate::events::Event).
     pub source: Option<MirPlace>,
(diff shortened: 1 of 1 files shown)


aneksteind:
for the types that are meant to be mirrored, what about calling the mirrored types equality etc. impls after calling .into() for self and other?

kkysen:
> for the types that are meant to be mirrored, what about calling the mirrored types equality etc. impls after calling .into() for self and other?

We could do that, but that's 1) a separate change than the ones here, and 2) equivalent to what we already have, so I'm not sure if it's worth it.

fw-immunant:
I was looking at this and would like some clarification on intent/consequences here: are any of these fixes for correctness, or do they just bring things more into alignment with what `rustc` does? Do we use these `impl`s now, or is it just likely that we'll want them in the future?

kkysen:
> I was looking at this and would like some clarification on intent/consequences here: are any of these fixes for correctness, or do they just bring things more into alignment with what `rustc` does? Do we use these `impl`s now, or is it just likely that we'll want them in the future?

These aren't correctness fixes.  They more closely match what `rustc` `impl`s, and for some structs, it's an optimization (e.x. only comparing the one field that matters, not the bigger field that doesn't).  I do think it's likely we'll need some of these, like the `Ord` `impl`s.  I was working on something before where I needed an `Ord` `impl`, which is what prompted these changes, but I since figured out a different way to do that.  That said, it's probable that we'll need an `Ord` `impl` in the future again.

fw-immunant (approved):

PR #605, opened 2022-08-12
Append `Metadata` to metadata file using a file lock

### LINKED ISSUE #586: `rustc_wrapper` is invoked twice during dynamic instrumentation

kkysen:
Fixes #586.

This uses [`fs2`](https://docs.rs/fs2) file locking to append serialized `Metadata`s to the metadata file without race conditions.  Thanks for the suggestion, @ahomescu!

I'm unsure how to test for any race conditions, though.

### DIFF
--- a/dynamic_instrumentation/src/instrument.rs
+++ b/dynamic_instrumentation/src/instrument.rs
@@ -2,6 +2,7 @@ use anyhow::Context;
 use c2rust_analysis_rt::metadata::Metadata;
 use c2rust_analysis_rt::mir_loc::{self, EventMetadata, Func, MirLoc, MirLocId, TransferKind};
 use c2rust_analysis_rt::HOOK_FUNCTIONS;
+use fs2::FileExt;
 use fs_err::OpenOptions;
 use indexmap::IndexSet;
 use log::debug;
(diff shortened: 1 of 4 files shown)


dynamic_instrumentation/src/instrument.rs
```diff
@@ -62,14 +65,22 @@ impl Instrumenter {
...
-        fs_err::write(metadata_file_path, &bytes).context("Could not open metadata file")?;
+        let mut file = OpenOptions::new()
+            .append(true)
+            .write(true)
+            .open(metadata_path)
+            .context("Could not open metadata file")?;
+        file.file().lock_exclusive()?;
+        let e = file.write_all(&bytes);
+        file.file().unlock()?;
+        e?;
```
kkysen, line 83: I wasn't sure how exactly to do this.  Am I correct in not returning the error until after unlocking the file so that we don't leave the file locked if there is a write error?  And is there a reason there's not a lock guard or something similar returned from locking?

aneksteind: I looks right to me to unlock before returning the error. I'd be curious if `write_all` could panic even though it returns a result. My guess about the locking guard is to match what the underlying sys crates are doing.

aneksteind (approved):

dynamic_instrumentation/src/instrument.rs
kkysen: `write_all` shouldn't panic unless something is seriously wrong with the `write` syscall.

rinon (approved):

PR #609, opened 2022-08-12
WIP: casting fat ptrs to non-fat ptrs and then to `usize`

kkysen:
I'm not exactly sure how to do the actual casting.


kkysen:
Fixed in https://github.com/immunant/c2rust/pull/595/commits/906968bffba98ecf95487bec9c6e64d374b176d0 instead.

PR #612, opened 2022-08-14
Support `va_copy` with a member of a struct pointer

Hakuyume:
This PR add support for `va_copy` which argument is a member of a struct point.

Example C code.
```c
#include <stdarg.h>

struct S {
  va_list args;
};

void foo(va_list args) {
  struct S *s;
  va_copy(s->args, args);
}
```

Emitted Rust code.
```rust
#![register_tool(c2rust)]
#![feature(register_tool)]
pub type __builtin_va_list = [__va_list_tag; 1];
#[derive(Copy, Clone)]
#[repr(C)]
pub struct __va_list_tag {
    pub gp_offset: libc::c_uint,
    pub fp_offset: libc::c_uint,
    pub overflow_arg_area: *mut libc::c_void,
    pub reg_save_area: *mut libc::c_void,
}
pub type va_list = __builtin_va_list;
#[derive()]
#[repr(C)]
pub struct S<'a> {
    pub args: ::core::ffi::VaListImpl::<'a>,
}
#[no_mangle]
pub unsafe extern "C" fn foo(mut args: ::core::ffi::VaList) {
    let mut s: *mut S = 0 as *mut S;
    (*s).args = args.clone();
}
```

### DIFF
--- a/c2rust-transpile/src/translator/variadic.rs
+++ b/c2rust-transpile/src/translator/variadic.rs
@@ -58,6 +58,22 @@ impl<'c> Translation<'c> {
             Some(va_id)
         }
 
+        // struct-based va_list (e.g. x86_64) where va_list is accessed as a member of a struct pointer
+        fn match_vastart_struct_pointer_member(
+            ast_context: &TypedAstContext,
+            expr: CExprId,
+        ) -> Option<CDeclId> {
+            match_or! { [ast_context[expr].kind]
+            CExprKind::ImplicitCast(_, me, _, _, _) => me }
+            match_or! { [ast_context[me].kind]
+            CExprKind::Member(_, ie, _, _, _) => ie }
+            match_or! { [ast_context[ie].kind]
+            CExprKind::ImplicitCast(_, e, _, _, _) => e }
+            match_or! { [ast_context[e].kind]
+            CExprKind::DeclRef(_, va_id, _) => va_id }
+            Some(va_id)
+        }
+
         // char pointer-based va_list (e.g. x86)
         fn match_vastart_pointer(ast_context: &TypedAstContext, expr: CExprId) -> Option<CDeclId> {
             match_or! { [ast_context[expr].kind]
@@ -68,6 +84,7 @@ impl<'c> Translation<'c> {
         match_vastart_struct(&self.ast_context, expr)
             .or_else(|| match_vastart_pointer(&self.ast_context, expr))
             .or_else(|| match_vastart_struct_member(&self.ast_context, expr))
+            .or_else(|| match_vastart_struct_pointer_member(&self.ast_context, expr))
     }
 
     pub fn match_vaend(&self, expr: CExprId) -> Option<CDeclId> {
--- a/c2rust-transpile/src/translator/variadic.rs
+++ b/c2rust-transpile/src/translator/variadic.rs
@@ -59,6 +59,7 @@ impl<'c> Translation<'c> {
         }
 
         // struct-based va_list (e.g. x86_64) where va_list is accessed as a member of a struct pointer
+        // supporting this pattern is necessary to transpile [graphviz](https://gitlab.com/graphviz/graphviz/-/blob/5.0.0/lib/sfio/sftable.c#L321)
         fn match_vastart_struct_pointer_member(
             ast_context: &TypedAstContext,
             expr: CExprId,
--- a/tests/items/src/varargs.c
+++ b/tests/items/src/varargs.c
@@ -80,6 +80,16 @@ void valist_struct_member(const char *fmt, ...) {
   va_end(thestruct.args);
 }
 
+// pattern first seen in graphviz (sftable.c)
+void valist_struct_pointer_member(const char *fmt, ...) {
+  struct vastruct thestruct;
+  struct vastruct *pointer = &thestruct;
+
+  va_start(pointer->args, fmt);
+  vprintf(fmt, pointer->args);
+  va_end(pointer->args);
+}
+
 // mirrors pattern from json-c's sprintbuf
 void restart_valist(const char *fmt, ...) {
     va_list ap;
--- a/tests/items/src/varargs.c
+++ b/tests/items/src/varargs.c
@@ -73,21 +73,27 @@ struct vastruct {
 
 // pattern first seen in apache (util_script.c)
 void valist_struct_member(const char *fmt, ...) {
-  struct vastruct thestruct;
+  struct vastruct a, b;
 
-  va_start(thestruct.args, fmt);
-  vprintf(fmt, thestruct.args);
-  va_end(thestruct.args);
+  va_start(a.args, fmt);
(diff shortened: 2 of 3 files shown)


kkysen:
Hi @Hakuyume!  Thanks for the help!

I haven't looked at this part of the transpiler before, so could you tell me what the previous behavior was (an error, mis-transpilation, etc.)?  Also, what exactly was the motivating use case for this?  The example you gave,

```c
void foo(va_list args) {
  struct S *s;
  va_copy(s->args, args);
}
```

dereferences an uninitialized pointer.  I assume that's just an overly simplified example, but if you could give some more context on what you need this for, that would be great!

That said, I'm not sure why the transpiler only works on only pointers and struct members and not just general expressions.  Adding more support would be helpful, but ideally, it'd just work on any expression.

Hakuyume:
> so could you tell me what the previous behavior was (an error, mis-transpilation, etc.)?

c2rust ignores (skips) the function that contains `va_copy` with a member of struct pointer. In the example above, the function `foo` is entirely skipped. 

> I assume that's just an overly simplified example, but if you could give some more context on what you need this for, that would be great!

Yes. As you mentioned, my example is too simplified.
Actually, I would like to transpile graphviz.
graphviz contains `va_copy` with such arguments.
https://gitlab.com/graphviz/graphviz/-/blob/5.0.0/lib/sfio/sftable.c#L321

I think my patch is ad-hoc. If there is more general way, it would be nice.

kkysen:
Thanks for the further context!  I'll look into it and see if there's a reason it's not more generic.  @thedataking, any thoughts, as it says you wrote most of the `match_va*` code in #92?

kkysen:
@thedataking, why not do something like this?

Define

```rust
// Or maybe call this `__builtin_va_copy`.
fn va_copy(dest: &mut VaListImpl, src: &VaListImpl) {
    dest.clone_from(src);
}
```

and then transform

```c
va_copy(e1, e2)
```

into

```rust
va_copy(&mut transpile(e1), &transpile(e2));
```

Why is the specific matching on arguments necessary?  Also, I think a mis-translation of invalid C code into similarly invalid Rust code is better than entirely skipping a function, which makes no sense.

thedataking:
It has been more than three years since I worked on #92 so I'm fuzzy on some of the details. As far as I remember, the challenge is that some uses of `va_copy` would require (arbitrarily) complex static analysis. I think Andrei and I discovered this while looking at some uses in Apache apr but it might have been something else.

> Also, I think a mis-translation of invalid C code into similarly invalid Rust code is better than entirely skipping a function, which makes no sense.

The overall operation of the transpiler is that it either translates the input into Rust that (modulo bugs) is functionally equivalent to the input or bails out with an error. Not doing so could introduce some *very* subtle and hard to find bugs.

A few other notes:
- I don't see tests for this feature, this PR should test the new code paths. `test_varargs.rs` might be a good place to do so.
- Handling of variadic functions vary from 32-bit code and 64-bit code and between different architectures. Be very careful.

kkysen:
Thanks for context, @thedataking!  In that case, @Hakuyume, I think an ad-hoc fix like this is okay if we can't easily get a general solution.  Please add a test in `tests/items/src/test_varargs.rs` and `tests/items/src/varargs.c` that tests this `va_copy` with a pointer in a struct case, simplified from the graphviz source but containing any UB.  Then it will likely be good to merge after I review it more carefully.

tests/items/src/varargs.c
```diff
@@ -80,6 +80,16 @@ void valist_struct_member(const char *fmt, ...) {
...
 }
 
+// pattern first seen in graphviz (sftable.c)
+void valist_struct_pointer_member(const char *fmt, ...) {
+  struct vastruct thestruct;
+  struct vastruct *pointer = &thestruct;
+
+  va_start(pointer->args, fmt);
+  vprintf(fmt, pointer->args);
+  va_end(pointer->args);
```
kkysen, line 90: It's not testing `va_copy`.  That might be handled similarly to `va_start` and `va_end`, but I think it should still be tested as well.  Can you add that?  Might as well add the same to the similar `valist_struct_member` function above.

kkysen (changes requested):
See the comment; just add a `va_copy` in the test and then the test should be all good.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -58,6 +58,22 @@ impl<'c> Translation<'c> {
             Some(va_id)
         }
 
+        // struct-based va_list (e.g. x86_64) where va_list is accessed as a member of a struct pointer
```
kkysen, line 61:
```suggestion
        // struct-based va_list (e.g. x86_64) where va_list is accessed as a member of a struct pointer
        // supporting this pattern is necessary to transpile [graphviz](https://gitlab.com/graphviz/graphviz/-/blob/5.0.0/lib/sfio/sftable.c#L321)
```

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -58,6 +58,22 @@ impl<'c> Translation<'c> {
...
+        ) -> Option<CDeclId> {
+            match_or! { [ast_context[expr].kind]
+            CExprKind::ImplicitCast(_, me, _, _, _) => me }
+            match_or! { [ast_context[me].kind]
+            CExprKind::Member(_, ie, _, _, _) => ie }
+            match_or! { [ast_context[ie].kind]
+            CExprKind::ImplicitCast(_, e, _, _, _) => e }
+            match_or! { [ast_context[e].kind]
+            CExprKind::DeclRef(_, va_id, _) => va_id }
+            Some(va_id)
```
kkysen, line 74: What are `me`, `ie`, and `e`?  I'm not sure what the `me` and `e` are in `match_vastart_struct_member` either.

Hakuyume: I copied `me` and `e` from `match_vastart_struct_member`. I don't know their meaning. I guessed that `e` stands for "expr" and `me` stands for "Member expr". I used `ie` for `ImplicitCast expr` in the same way.

Hakuyume:
@kkysen Thank you for reviewing. I updated test cases. Some of CIs failed but the failure does not seem related to my patch.

c2rust-transpile/src/translator/variadic.rs
kkysen: Hmm, I think it's okay for now.  Probably it should be given a more descriptive name, but so should `match_vastart_struct_member`.  I may clean up some of this code, especially the `match_or!` macro, is it's quite un-idiomatic Rust and contains hidden non-local control flow and adds bindings.

kkysen (approved):
I'm just going to rebase and rename that suggestion commit, but otherwise, it's all good and I'll merge this once CI passes.  Thank you!

kkysen:
The CI failure is due to some other flaky tests that we're trying to debug.  I re-ran the CI and it should be fine.

PR #614, opened 2022-08-15
Warn if any `--binary`s are not transpiled

### LINKED ISSUE #611: Transpiling doesn't raise any error if the `--binary parameter` is incorrect
It'd be helpful if `c2rust transpile --binary` raised an error, if the parameter provided is incorrect.

Currently, the binary functionality fails silently, and if the user accidentally typoes the name, the result is very confusing.

kkysen:
Fixes #611.

Blocked on #615.

For the example given by @64kramsystem in #611,

```sh
gh repo clone Blzut3/CatacombSDL
cd CatacombSDL
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=1
c2rust transpile --binary pizzarulez compile_commands.json
```

this is now the output (once https://github.com/immunant/c2rust/pull/421/commits/6a6c0d8bb6a6a6f5bc5ab18233c622102e5f9291 is fixed):

```shell
warning: binary not used: pizzarulez
info: candidate modules for binaries are: pcrlib_a, catasm, catacomb, objects, cpanel, cat_play, pcrlib_c, rleasm
```

@64kramsystem, if you could review this PR, at least the output, that would be great!

### DIFF
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -18,6 +18,7 @@ use std::path::{Path, PathBuf};
 use std::process;
 
 use failure::Error;
+use itertools::Itertools;
 use log::{info, warn};
 use regex::Regex;
 use serde_derive::Serialize;
@@ -110,11 +111,12 @@ impl TranspilerConfig {
         for binary in &self.binaries {
             if !module_names.contains(binary) {
                 ok = false;
-                warn!("binary not used: {binary:?}");
+                warn!("binary not used: {binary}");
             }
         }
         if !ok {
-            info!("candidate modules for binaries are: {module_names:#?}");
+            let module_names = module_names.iter().format(", ");
+            info!("candidate modules for binaries are: {module_names}");
         }
         ok
     }
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -88,10 +88,32 @@ pub struct TranspilerConfig {
 }
 
 impl TranspilerConfig {
-    fn is_binary(&self, file: &Path) -> bool {
+    fn binary_name_from_path(file: &Path) -> String {
         let file = Path::new(file.file_stem().unwrap());
-        let name = get_module_name(file, false, false, false).unwrap();
-        self.binaries.contains(&name)
+        get_module_name(file, false, false, false).unwrap()
+    }
+
+    fn is_binary(&self, file: &Path) -> bool {
+        let module_name = Self::binary_name_from_path(file);
+        self.binaries.contains(&module_name)
+    }
+
+    fn check_binaries(&self, modules: impl IntoIterator<Item = impl AsRef<Path>>) -> bool {
+        let module_names = modules
+            .into_iter()
+            .map(|module| Self::binary_name_from_path(module.as_ref()))
+            .collect::<HashSet<_>>();
+        let mut ok = true;
+        for binary in &self.binaries {
+            if !module_names.contains(binary) {
+                ok = false;
+                warn!("binary not used: {binary:?}");
+            }
+        }
+        if !ok {
+            warn!("candidate modules for binaries are: {module_names:#?}");
+        }
+        ok
     }
 
     fn crate_name(&self) -> String {
@@ -216,6 +238,7 @@ pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]
     let mut top_level_ccfg = None;
     let mut workspace_members = vec![];
     let mut num_transpiled_files = 0;
+    let mut all_modules = Vec::new();
     let build_dir = get_build_dir(&tcfg, cc_db);
     for lcmd in &lcmds {
         let cmds = &lcmd.cmd_inputs;
@@ -297,6 +320,8 @@ pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]
         pragmas.sort();
         crates.sort();
 
+        all_modules.extend(modules.iter().cloned());
+
         if tcfg.emit_build_files {
             if modules_skipped {
                 // If we skipped a file, we may not have collected all required pragmas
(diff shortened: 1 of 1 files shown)


kkysen:
@64kramsystem, I think you should be able to officially review the PR with only read access.  Let me know if you can't though, as I had thought anyone can.

64kramsystem (approved):
Reviewed and verified. Looks good!

c2rust-transpile/src/lib.rs
```diff
@@ -88,10 +88,32 @@ pub struct TranspilerConfig {
...
-        self.binaries.contains(&name)
+        get_module_name(file, false, false, false).unwrap()
+    }
+
+    fn is_binary(&self, file: &Path) -> bool {
+        let module_name = Self::binary_name_from_path(file);
+        self.binaries.contains(&module_name)
+    }
+
+    fn check_binaries(&self, modules: impl IntoIterator<Item = impl AsRef<Path>>) -> bool {
```
fw-immunant, line 101:
```suggestion
    fn check_all_binaries_used(&self, modules_to_transpile: impl IntoIterator<Item = impl AsRef<Path>>) -> bool {
```

I think these names make it clear what the function will do without reading its body.

c2rust-transpile/src/lib.rs
```diff
@@ -88,10 +88,32 @@ pub struct TranspilerConfig {
...
+            .collect::<HashSet<_>>();
+        let mut ok = true;
+        for binary in &self.binaries {
+            if !module_names.contains(binary) {
+                ok = false;
+                warn!("binary not used: {binary:?}");
+            }
+        }
+        if !ok {
+            warn!("candidate modules for binaries are: {module_names:#?}");
```
fw-immunant, line 114: It would be nice for these warnings to specify the binary names in the same format that the user would use as an input (i.e., an `OsStr` string that can be copy-pasted into the user's command line). The Rust-style `Debug` syntax isn't particularly friendly as user-directed output here.

fw-immunant (approved):
LGTM after inline comments are addressed.

c2rust-transpile/src/lib.rs
kkysen: I agree, though I think `check_if_all_binaries_used` is a bit clearer still.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/614/commits/fe5c78369af8d3e04ae242b15c9fabfe54fb15e7.

c2rust-transpile/src/lib.rs
kkysen: `binary` is a `String` and `module_names` is a `HashSet<String>`, so the `Debug` output is just the `Display` output but quoted.  They should be fully copy-pastable.  I thought the quotes would clarify exactly what is part of the module names, and the `:#?` extended `Debug` formatting is so it's printed across multiple lines and is clearer.

fw-immunant: The format printed by `Debug` is basically Rust source code, which overlaps partially but (in edge cases) confusingly with what shells understand, so I think the right compromise here is to simply treat it as "plain" text (i.e., non-adversarial strings in the filesystem encoding). We don't need to be secure against malicious inputs here because this is just a debugging hint, and I think it's more human-friendly to just print the strings as-is than in their `Debug` representation. I think the indentation provided by `:#?` is good, but we can also provide indentation ourselves.

Here's example of Rust debug output that differs from what a shell would understand:
```rust
fn main(){
    let mut x = std::collections::HashSet::new();
    x.insert("a\u{1c}b");
    // prints an escape sequence (`\u`) that shell interprets differently from Rust
    println!("{:#?}", x);
    for i in x {
        // just prints the string itself, raw
        println!("\t{}", i);
    }
}
```

Output:
```
{
    "a\u{1c}b",
}
	ab
```

kkysen: The latter looks way less readable, understandable, and copyable to me.  Plus, these have to be valid Rust module names (and thus an identifier), right, so they can't be too crazy?

fw-immunant: My point is that valid Rust module names don't benefit from being printed with quotes, commas, and escaping, and that furthermore the format we print isn't the same format as the user needs (though that, as you mention, shouldn't really matter for module names).

kkysen: Names like module names are often quoted in `'`, `"`, or `` ` ``.  Is it that big of a deal?

kkysen: This has readability benefits, too.  For example, syntax highlighting works on the module names here:

```shell
warning: binary not used: "pizzarulez"
warning: candidate modules for binaries are: {
    "objects",
    "cpanel",
    "cat_play",
    "pcrlib_c",
    "rleasm",
    "catacomb",
    "catasm",
    "pcrlib_a",
}
```

thedataking: @kkysen 

Frances makes a good point:
> It would be nice for these warnings to specify the binary names in the same format that the user would use as an input (i.e., an OsStr string that can be copy-pasted into the user's command line). 

Further, it is not necessary to print out candidate modules so we can shortcut any further discussion about the format by simply omitting it. I would just print out the first line (with the changes Frances suggested.) Please consider that someone will have to maintain any features that you add within a very limited budget.

Regardless of whose suggestions you take, please wrap up this PR since it does not support our lifting goal. Frances has already been generous with his time , has offered helpful feedback, and we cannot justify spending more time here.

kkysen: Switched to `Display` in https://github.com/immunant/c2rust/pull/614/commits/6a6fd8bf52f345e499e872dbf01b74692c2066ca, and I switched to `info!` for the suggestion.  I do think it's quite useful and is like 1 line of code.  I'll merge this now unless there are further comments.  I was just trying to fix a pretty simple bug someone found.

thedataking (approved):

PR #615, opened 2022-08-16
Fix logging regression (no logs)

kkysen:
Fix logging regression (no logs) from https://github.com/immunant/c2rust/pull/421/commits/6a6c0d8bb6a6a6f5bc5ab18233c622102e5f9291.  I tested this combined with #614.

### DIFF
--- a/c2rust-transpile/src/diagnostics.rs
+++ b/c2rust-transpile/src/diagnostics.rs
@@ -32,7 +32,7 @@ pub fn init(mut enabled_warnings: HashSet<Diagnostic>, log_level: log::LevelFilt
     enabled_warnings.extend(DEFAULT_WARNINGS.iter().cloned());
 
     let colors = ColoredLevelConfig::new();
-    let (_log_level, logger) = fern::Dispatch::new()
+    let (max_level, logger) = fern::Dispatch::new()
         .format(move |out, message, record| {
             let level_label = match record.level() {
                 Level::Error => "error",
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/diagnostics.rs
```diff
@@ -64,7 +64,9 @@ pub fn init(mut enabled_warnings: HashSet<Diagnostic>, log_level: log::LevelFilt
         })
         .chain(io::stderr())
         .into_log();
+    let _ = log_reroute::init(); // Ignore the [`SetLoggerError`] b/c we just want to make sure it's set at least once.
```
rinon, line 67: I think you should make this explicitly `let _ignored: Result<(), SetLoggerError> = ...` so that if log_reroute ever returns a new error, we think about whether we want to check for it.

rinon (approved):
One thing you should make explicit, but LGTM other than that.

c2rust-transpile/src/diagnostics.rs
kkysen:
```suggestion
    // Ignore the [`SetLoggerError`] b/c we just want to make sure it's set at least once.
    let _: Result<(), SetLoggerError> = log_reroute::init();
```

How about this?  `_` vs. `_ignored` have different drop semantics, but putting the type annotation there ensures it won't change without a compile error.

rinon: Yep, that looks good. I didn't actually know you could attach a type annotation to `_`. Makes sense, but TIL.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/615/commits/d7508f925555d72eccba536307b0943769c0fe15.

Issue #625, opened 2022-08-18
`c2rust-instrument` does not correctly forward applicable `cargo` args to `cargo` subcommands besides the primary one


kkysen:
See https://github.com/immunant/c2rust/issues/624#issuecomment-1219217877.

After #626 lands, most of this will be fixed, except for forwarding `--manifest-path` to `cargo add`.

PR #626, opened 2022-08-18
Disable incremental instrumentation and remove problematic non-primary `cargo` subcommands

### LINKED ISSUE #624: `c2rust-instrument` fails with a non `debug`/`dev` profile (and other various `cargo` args)
`c2rust-instrument` runs `cargo clean --package {primary_package}`, but doesn't forward the other `cargo` arguments passed to `c2rust-instrument`.  This breaks things when `--release` is set, for example, as then the right package isn't cleaned.  However, we can't just blindly forward all the arguments, as some don't apply to `cargo clean`.  So we need to detect which arguments to forward.  We should also be careful about this with the other `cargo` subcommands beyond the main one, such as `cargo metadata` and `cargo add`.  In particular, I know at least `--manifest-path` needs to be forwarded.

kkysen:
Fixes #624 and f_ixes most of #625.

This disables incremental compilation in the `rustc_wrapper` in `MirTransformCallbacks`, thus disabling incremental instrumentation while still allowing incremental compilation for other `rustc` calls.  Thus, we do not need to `cargo clean --package {root_package}` anymore, fixing #624.

Without `cargo clean --package {root_package}`, `cargo` will not re-run `rustc` if no inputs change, so we also have to account for this case.  We don't have to actually do anything in this case, since nothing has changed, and so we don't need to touch the metadata file at all.  However, we have to avoid truncating the metadata file in this case, as that was the previous behavior.  Since we don't truncate the metadata file at the beginning, we end up with an ever-growing metadata file as subsequent `cargo-instrument`s keep appending new `Metadata`s to the metadata file.  This actually seems to work when they're merged, but is obviously undesirable.  To fix this, we check after if the metadata file has been appended to.  If it previously had `Metadata` and now has more `Metadata`, then we delete the old `Metadata` from the file (in practice, read the new `Metadata`, write it to the beginning, and then truncate the file).

Without having to run `cargo clean --package {root_package}` anymore, we also don't need to run `cargo metadata` either, as we used that to lookup the root package (note that this was incorrect if `c2rust-instrument ... --package {package}` was specified, for example, and we now handle such cases).  Thus, we can also drop the `cargo_metadata` dependency.

The only other `cargo` subcommand we invoke is `cargo add` (and only when `--set-runtime` is explicitly set).  Unlike `cargo clean` and many other `cargo` subcommands, `cargo add`'s arguments are largely orthogonal to other `cargo` args and thus we don't need to forward very many args to it (which is incredibly hard to do correctly).  The only argument that I see that we need to forward is `--manifest-path`, and it should be simple enough to detect that by itself.

Also, this approach is also a bit faster (50-100% faster when there are minimal/no changes), as we don't need a separate `cargo clean` and in the case of no inputs having changed, we don't need to do anything.

### DIFF
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -24,8 +24,10 @@ mod util;
 use crate::callbacks::{MirTransformCallbacks, INSTRUMENTER};
 
 use std::{
+    cmp::Ordering,
     env,
     ffi::{OsStr, OsString},
+    io::{Read, Seek, Write},
     iter,
     path::{Path, PathBuf},
     process::{self, Command, ExitStatus},
@@ -34,8 +36,9 @@ use std::{
 use fs_err::OpenOptions;
 use rustc_driver::{RunCompiler, TimePassesCallbacks};
 use rustc_session::config::CrateType;
+use std::io::SeekFrom;
 
-use anyhow::{anyhow, ensure, Context};
+use anyhow::{anyhow, bail, ensure, Context};
 use camino::Utf8Path;
 use cargo_metadata::MetadataCommand;
 use clap::Parser;
@@ -313,16 +316,6 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
         .root_package()
         .ok_or_else(|| anyhow!("no root package found by `cargo`"))?;
 
-    cargo.run(|cmd| {
-        // Clean the primary package so that we always rebuild it
-        // and get up-to-date, complete instrumentation [`Metadata`].
-        // Incremental instrumentation is very tricky,
-        // so don't try that yet,
-        // and if we only need to rebuild the primary package,
-        // it usually isn't that slow.
-        cmd.args(&["clean", "--package", root_package.name.as_str()]);
-    })?;
-
     if set_runtime {
         cargo.run(|cmd| {
             cmd.args(&["add", "--optional", "c2rust-analysis-rt"]);
@@ -334,12 +327,16 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
         })?;
     }
 
-    // Create and truncate the metadata file for the [`rustc_wrapper`]s to append to.
-    OpenOptions::new()
+    // Create the metadata file for the [`rustc_wrapper`]s to append to.
+    // Note that we can't truncate it because in the case that no inputs have changed,
+    // `cargo` won't recompile, so a new metadata file won't be regenerated.
+    // We should keep the old one in that case.
+    let mut metadata_file = OpenOptions::new()
         .create(true)
-        .write(true) // need write for truncate
-        .truncate(true)
+        .truncate(false)
         .open(&metadata_path)?;
+
+    let old_len = metadata_file.metadata()?.len();
 
     cargo.run(|cmd| {
         // Enable the runtime dependency.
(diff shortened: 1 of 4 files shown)


rinon:
What about doing something a bit different and I think simpler... passing a unique temporary metadata path (either using tempfile or, even simpler, just passing metadata_path.PID where PID comes from `std::process::id`). That guarantees that we won't get conflicts with another concurrent invocation of this process. Then we can move that file to `metadata_path` if it exists (and therefore got written to) at the end of the operation, and otherwise we won't overwrite anything.

Something to watch out for here is what happens when some but not all the targets of the crate (e.g. one binary but not another) needs recompilation. Do we end up recompiling all, or does cargo short-circuit this? If we don't recompile everything, that seems broken and we might need per-output binary metadata files.

kkysen:
> What about doing something a bit different and I think simpler... passing a unique temporary metadata path (either using tempfile or, even simpler, just passing metadata_path.PID where PID comes from `std::process::id`). That guarantees that we won't get conflicts with another concurrent invocation of this process. Then we can move that file to `metadata_path` if it exists (and therefore got written to) at the end of the operation, and otherwise we won't overwrite anything.

That's a very good idea, which I think can eliminate any copies and simplify things.  I'll move the old metadata to `metadata.old.bc` or something similar based on the PID or tempfile.  Then if a new metadata file is created, I delete the old one, and if there is no new metadata file, I move it back.

> Something to watch out for here is what happens when some but not all the targets of the crate (e.g. one binary but not another) needs recompilation. Do we end up recompiling all, or does cargo short-circuit this? If we don't recompile everything, that seems broken and we might need per-output binary metadata files.

That's a good catch.  I'm still not totally sure about `cargo`'s behavior, but I'm assuming that if only the binary target (and not the library target) need to be rebuilt, `cargo` will only invoke `rustc` for the binary target, so we have to avoid deleting the library target metadata in that case.  Perhaps the best way to do that is indeed separate files, but I want to address (as in decide how to handle it) #632 more thoroughly before trying to do that here at all (and would like to fix it in a separate PR anyways if I need to).

kkysen:
I switched to the temp file and renaming solution in https://github.com/immunant/c2rust/pull/626/commits/6b085c24d3e9e5e10728083fb6df1cf73e5ddfbc, though I still need to clean it up a bit.  It's also been tested upstream in #637 and still works there.

aneksteind (approved):

kkysen:
I added the TODO here: https://github.com/immunant/c2rust/pull/626/commits/0a5026b7ef7e95cf750530e7a54ff8943a334518.

dynamic_instrumentation/src/main.rs
```diff
@@ -291,10 +284,134 @@ fn set_rust_toolchain() -> anyhow::Result<()> {
...
+}
+
+impl MetadataFile {
+    /// The old, original, intended [`Path`] for the [`Metadata`].
+    ///
+    /// This is the location that existing [`Metadata`] may be at,
+    /// and where the [`Metadata`] will end after this program exits.
+    ///
+    /// [`Metadata`]: c2rust_analysis_rt::metadata::Metadata
+    pub fn old_path(&self) -> &Path {
```
rinon, line 304: small nitpick, I would call this `final_path` and the other `temp_path` but I'm not going to block this PR on that change. It's fine to merge as is.

rinon (approved):

dynamic_instrumentation/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/626/commits/1dfcd2a966583147ea496c9f6ff0c3ef58ac2b35.

PR #627, opened 2022-08-18
Forward `--manifest-path` to all `cargo` subcommands in `c2rust-instrument`

### LINKED ISSUE #625: `c2rust-instrument` does not correctly forward applicable `cargo` args to `cargo` subcommands besides the primary one

kkysen:
Fixes #625.

This uses a separate `clap::Parser` (`cargo` also uses `clap` 3) to parse part of the forwarded `cargo` args (to the primary `cargo` subcommand), in particular, `--manifest-path`, which is the only general `cargo` arg that we need to forward to the only non-primary `cargo` subcommand remaining, `cargo add`.

This also correctly resolves the `$CARGO_TARGET_DIR` and metadata path relative to a `--manifest-path`.

With `--manifest-path` now working, we can do things like this, all in one command, making `pdg.sh` less and less necessary:

```shell
INSTRUMENT_BACKEND=log INSTRUMENT_OUTPUT=log.bc INSTRUMENT_OUTPUT_APPEND=false cargo run --bin c2rust-instrument -- --metadata analysis/test/metadata.bc -- run --manifest-path ../lighttpd/rust/Cargo.toml -- -D -f ~/work/rust/lighttpd-1.4.64/mylighttpd.conf -i 5 1> /dev/null
```

Note that `pdg.sh` has also been updated and simplified with the new improvements using `--manifest-path`.

### DIFF
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -450,7 +450,7 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
             cmd.args(&["add", "--optional", "c2rust-analysis-rt"]);
             if let Some(mut runtime) = runtime_path {
                 if manifest_dir.is_some() {
-                    runtime = runtime.canonicalize()?;
+                    runtime = fs_err::canonicalize(runtime)?;
                 }
                 // Since it's a local path, we don't need the internet,
                 // and running it offline saves a slow index sync.
@@ -473,7 +473,7 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
 
         // The [`rustc_wrapper`] might run in a different working directory if `--manifest-path` was passed.
         let metadata_path = metadata_file.temp_path();
-        let abs_metadata_path = metadata_path.canonicalize()?;
+        let abs_metadata_path = fs_err::canonicalize(metadata_path)?;
         let metadata_path = match manifest_dir {
             Some(_) => abs_metadata_path.as_path(),
             None => metadata_path,
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -60,6 +60,13 @@ struct Args {
     cargo_args: Vec<OsString>,
 }
 
+/// `cargo` args that we intercept.
+#[derive(Debug, Parser)]
+struct InterceptedCargoArgs {
+    #[clap(long, value_parser)]
+    manifest_path: Option<PathBuf>,
+}
+
 fn exit_with_status(status: ExitStatus) {
     process::exit(status.code().unwrap_or(1))
 }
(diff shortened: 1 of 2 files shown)


dynamic_instrumentation/src/main.rs
```diff
@@ -63,6 +63,17 @@ struct Args {
...
+/// `cargo` args that we intercept.
+#[derive(Debug, Parser)]
+struct InterceptedCargoArgs {
+    #[clap(long, value_parser)]
+    manifest_path: Option<PathBuf>,
+
+    /// Need this so `--` is allowed.
+    /// Not actually used.
+    extra_args: Vec<OsString>,
+}
```
kkysen, line 75: I actually need to fix this to allow arbitrary `cargo` args to pass through without error.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/627/commits/561a2701d7a29e45314411a16a2933f9641d1fb5.

dynamic_instrumentation/src/main.rs
```diff
@@ -402,23 +423,43 @@ fn cargo_wrapper(rustc_wrapper: &Path) -> anyhow::Result<()> {
...
     }
 
-    let mut metadata_file = MetadataFile::open(&metadata_path)?;
+    let mut metadata_file = MetadataFile::open(metadata_path)?;
 
     cargo.run(|cmd| {
         // Enable the runtime dependency.
+        let cargo_target_dir = manifest_dir
+            .unwrap_or_else(|| Path::new(""))
+            .join("instrument.target");
```
fw-immunant, line 447: If the user has `CARGO_TARGET_DIR` set already but does not specify a manifest path, this will overwrite the user's `CARGO_TARGET_DIR` with an empty string, which is probably unexpected.

The comment above here is also confusing now--I believe it should instead stay with the below `add_feature` line.

kkysen: No, this will overwrite it with `instrument.target`, which was my intended (and previous before this PR) behavior.

kkysen: Yeah, the comment is in the wrong place now, good catch.

fw-immunant: > No, this will overwrite it with `instrument.target`, which was my intended (and previous before this PR) behavior.

Ohh, you're right. :+1:

kkysen: Comment fixed in https://github.com/immunant/c2rust/pull/627/commits/ca434f24e5dd19f308b4d5ccfdcaa95f2cf4e7de.

Also, in https://github.com/immunant/c2rust/pull/627/commits/3d2e8c0da5189df31929f0e82ac3a86ebacf1b50, I switched from `""` to `"."`, which I think is a bit clearer.

aneksteind (approved):
lgtm once tested with lighttpd or similar

kkysen:
It's been tested on `lighttpd`; I developed it testing on `lighttpd`.

PR #629, opened 2022-08-19
Add `c2rust-instrument --rustflags`

kkysen:
Add a `--rustflags` argument to `c2rust-instrument`, as it much more easily allows you specify `$RUSTFLAGS` for the instrumentation only (and not all the other `cargo`s).  Otherwise, you have to split up the `cargo` commands and run `c2rust-instrument` directly, which can be a real pain during development.

I'm also not sure why `cargo` doesn't have this, too.

Perhaps this should only apply to `rustc_wrapper`s where we actually do instrumentation.  What do others think?

### DIFF
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -60,18 +60,20 @@ struct Args {
     #[clap(long)]
     set_runtime: bool,
 
-    /// `$RUSTFLAGS`, but as a CLI argument, not an inheritable environment variable.
-    /// These flags are appended (space-separated) to the existing `$RUSTFLAGS`, if it exists.
+    /// Set `$RUSTFLAGS` for the instrumented `cargo`.
     ///
-    /// The reason this exists is twofold:
-    ///
-    /// 1. It would be convenient if `cargo` itself had such a `--rustflags` argument,
-    ///    so at least we can recreate it here ourselves.
-    ///
-    /// 2. If this binary is invoked by something like `cargo run --bin c2rust-instrument`,
-    ///    then it's impossible to set `$RUSTFLAGS` only for the inner `cargo` that you want to add them to
-    ///    without also adding them to the outer `cargo run` `cargo`.
-    ///    `--rustflags` lets you do that easily.
+    /// This allows setting `$RUSTFLAGS` for the inner `cargo` when `c2rust-instrument` is invoked via `cargo run`, for example.
+    /// If `$RUSTFLAGS` is already set, these `--rustflags` are appended with a space.
+    //
+    // The reason this exists is twofold:
+    //
+    // 1. It would be convenient if `cargo` itself had such a `--rustflags` argument,
+    //    so at least we can recreate it here ourselves.
+    //
+    // 2. If this binary is invoked by something like `cargo run --bin c2rust-instrument`,
+    //    then it's impossible to set `$RUSTFLAGS` only for the inner `cargo` that you want to add them to
+    //    without also adding them to the outer `cargo run` `cargo`.
+    //    `--rustflags` lets you do that easily.
     #[clap(long)]
     rustflags: Option<OsString>,
 
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -24,6 +24,7 @@ mod util;
 use crate::callbacks::{MirTransformCallbacks, INSTRUMENTER};
 
 use std::{
+    borrow::Borrow,
     cmp::Ordering,
     env,
     ffi::{OsStr, OsString},
@@ -58,6 +59,10 @@ struct Args {
     /// Add the runtime as an optional dependency to the instrumented crate using `cargo add`.
     #[clap(long)]
     set_runtime: bool,
+
+    /// `$RUSTFLAGS`, but as a CLI argument, not an inheritable environment variable.
+    #[clap(long)]
+    rustflags: Option<OsString>,
 
     /// `cargo` args.
     cargo_args: Vec<OsString>,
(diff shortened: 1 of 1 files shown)


rinon:
What are the cargo commands we need to exclude `RUSTFLAGS` from? I could see this being useful if we only apply it to the crates we are instrumenting, but if we don't do that, what is the use case?

kkysen:
We often run things like this:

```sh
cargo run --release --bin c2rust-instrument -- --metadata ../lighttpd/rust/metadata.bc --rustflags="-Z time-passes" -- build --manifest-path ../lighttpd/rust/Cargo.toml
```

so it's very convenient to pass `$RUSTFLAGS` only to the `cargo` inside of `c2rust-instrument`, and not the outer one.  It's not strictly necessary, but it's much easier to use than separate `cargo build --bin c2rust-instrument` and `./target/$profile/c2rust-instrument` steps.  Accidentally messing up triggers a full `c2rust` rebuild, as changing `$RUSTFLAGS` flushes all the caches.

Furthermore, I've also wished `cargo` had a `--rustflags` arg, so here we're able to add one.

I do think it might be more useful to only apply it to the instrumented `rustc`s, but I'll have to see how that works and if that's a good idea (sometimes it can still be useful to add `$RUSTFLAGS` to the dependencies, too).  However that goes, I think it should be something built on top of this.

dynamic_instrumentation/src/main.rs
```diff
@@ -59,6 +60,10 @@ struct Args {
     #[clap(long)]
     set_runtime: bool,
 
+    /// `$RUSTFLAGS`, but as a CLI argument, not an inheritable environment variable.
```
fw-immunant, line 63: This should explain the use case for this option, not just what it is. This is doubly confusing because the implementation actually is setting an inheritable environment variable.

fw-immunant: This should also document that the argument to this CLI flag is appended (after a space) to the value of `RUSTFLAGS` if it is set in the environment.

rinon:
> We often run things like this:

ahhhhh I never even considered that, because I never use `cargo run`. I guess I'm old :older_man: 

Sure, this makes sense then, although please improve the help text as Frances mentioned.

dynamic_instrumentation/src/main.rs
kkysen: Okay, I'll add much more thorough documentation.

kkysen:
Sometimes that `cargo run` is inside of a `cargo test`, too.

dynamic_instrumentation/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/629/commits/aebde0a35bc7236e4223d44f7028178b139da089.

kkysen:
@fw-immunant, @rinon, does it look good now?

dynamic_instrumentation/src/main.rs
```diff
@@ -61,6 +61,17 @@ struct Args {
...
+    ///
+    /// The reason this exists is twofold:
+    ///
+    /// 1. It would be convenient if `cargo` itself had such a `--rustflags` argument,
+    ///    so at least we can recreate it here ourselves.
+    ///
+    /// 2. If this binary is invoked by something like `cargo run --bin c2rust-instrument`,
+    ///    then it's impossible to set `$RUSTFLAGS` only for the inner `cargo` that you want to add them to
+    ///    without also adding them to the outer `cargo run` `cargo`.
+    ///    `--rustflags` lets you do that easily.
```
fw-immunant, line 74:
```suggestion
    /// Set the `RUSTFLAGS` environment variable for the instrumented Cargo invocation.
    ///
    /// This allows setting `RUSTFLAGS` for the inner `cargo` when c2rust-instrument is invoked via
    /// `cargo run`. If `RUSTFLAGS` is already set in the environment, these flags are appended to
    /// the old value following a space character.
```

Thanks for the more detailed documentation. Would you be opposed to something more concise, like the above?

kkysen: More concise, yes, but I'm not sure your suggestion captures the full intent.  Is there a reason the verbosity is bad?

```suggestion
    /// `$RUSTFLAGS`, but as a CLI argument, not an inheritable environment variable.
    /// These flags are appended (space-separated) to the existing `$RUSTFLAGS`, if it exists.
    ///
    /// The reason this exists is twofold:
    ///
    /// 1. It would be convenient if `cargo` itself had such a `--rustflags` argument,
    ///    so at least we can recreate it here ourselves.
    ///
    /// 2. If this binary is invoked by something like `cargo run --bin c2rust-instrument`,
    ///    then it's impossible to set `$RUSTFLAGS` only for the inner `cargo` that you want to add them to
    ///    without also adding them to the outer `cargo run` `cargo`.
    ///    `--rustflags` lets you do that easily.
```

rinon: I prefer Frances wording as well, it communicates the information needed by users. We don't really need to document why the flag exists in the help text, that's more the place of comments in sources and commit messages. I appreciate the intent, and we should put this somewhere, I also just don't think it belongs here.

kkysen: It is also simultaneously the `rustdoc` documentation for that field, so I'm not sure where else to put it.

kkysen: How about this?

```rust
    /// Set `$RUSTFLAGS` for the instrumented `cargo`.
    ///
    /// This allows setting `$RUSTFLAGS` for the inner `cargo` when `c2rust-instrument` is invoked via `cargo run`, for example.
    /// If `$RUSTFLAGS` is already set, these `--rustflags` are appended with a space.
    //
    // The reason this exists is twofold:
    //
    // 1. It would be convenient if `cargo` itself had such a `--rustflags` argument,
    //    so at least we can recreate it here ourselves.
    //
    // 2. If this binary is invoked by something like `cargo run --bin c2rust-instrument`,
    //    then it's impossible to set `$RUSTFLAGS` only for the inner `cargo` that you want to add them to
    //    without also adding them to the outer `cargo run` `cargo`.
    //    `--rustflags` lets you do that easily.
```

Note that the bottom part use `//` not `///`.

rinon: That works!

kkysen: Fixed in https://github.com/immunant/c2rust/pull/629/commits/7f608b9105b9978fac7a3f9c228b60ad6479da42.

fw-immunant:
> @fw-immunant, @rinon, does it look good now?

Yes, I'll merge.

kkysen:
Reopened in #639.  It was meant to be merged into `master` once the 2 PRs in front of it merged.

PR #631, opened 2022-08-19
Convert `println!`s in `c2rust-instrument` to `log::trace!`s

kkysen:
The `println!`s really clutter up the output, especially for larger crates like `lighttpd`.  This emits them using an actual logger instead so that we can toggle them with just an environment variable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,6 +238,7 @@ dependencies = [
  "camino",
  "cargo_metadata",
  "clap 3.2.16",
+ "env_logger",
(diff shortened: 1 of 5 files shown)


rinon (approved):
These two logs could even be `trace` level IMO.

kkysen:
Switched from `info!` to `trace!` in https://github.com/immunant/c2rust/pull/631/commits/180a6c9ce76a64f55b305a586ef25d7f430c4c37, as I agree with @rinon about that.

PR #636, opened 2022-08-23
Inline `pdg.sh` into `c2rust-pdg` tests

kkysen:
This inlines `scripts/pdg.sh` into `c2rust-pdg`'s tests, which accomplishes:
* letting us test using the same `c2rust-pdg` test binary instead of running a second one in `pdg.sh`
* reducing our dependency on temporary bash scripts like `pdg.sh`
* making the test code much more generic, allowing it to be more easily extended

This is done by running, as a subcommand, `cargo run --bin c2rust-instrument ...`.  This is necessary as `c2rust-instrument` must be used a binary since it invokes itself through `cargo` and `$RUSTC_WRAPPER`.  Note that it detects what profile (e.x. `release`, `debug`) `c2rust-pdg` was compiled with and compiles `c2rust-instrument` with the same (`pdg.sh` did not do exactly this).  

Then the `c2rust-pdg`'s `main.rs` code was refactored so that the printing logic in `main` was moved to an `impl Display for PdgRepr` so that we can either `println!` it for `main` or `.to_string()` it for `insta::assert_display_snapshot!` (note that calls to that must be in `#[test]` functions so that `insta` can correctly detect the right test name.

#635 should be merged into `main` before this and then merged/rebased into this, but since that branch and the branch this is merging into have diverged, I can't express that correctly in GitHub.

Furthermore, I haven't deleted `pdg.sh` yet, as it still provides an interface for testing other crates (like `lighttpd`).  I intend to next add that functionality through `c2rust-pdg`'s `main` so that we can entirely get rid of `pdg.sh`.


aneksteind (changes requested):
would prefer all changes related to printing be in their own PR

kkysen:
Which printing-related changes?  This one: https://github.com/immunant/c2rust/pull/636/commits/d16b4c3fea3e49f7d646594728c23afc1d0f6d2a?  The other changes are refactors and don't change the output at all.

kkysen:
The refactor to `impl Display for PdgRepr` was split into #640.  This PR now is just the `pdg.sh` inlining changes.

pdg/src/main.rs
```diff
@@ -175,21 +175,134 @@ fn main() -> eyre::Result<()> {
...
+        let event_log_path = exe_dir.join("event.log.bc");
+
+        let mut cmd = Command::new("cargo");
+        cmd.current_dir(repo_dir()?)
+            .args(&[
+                "run",
+                "--bin",
+                "c2rust-instrument",
+                "--profile",
+                // Compile `c2rust-instrument` with the same profile us, `c2rust-pdg`, was compiled in.
```
fw-immunant, line 263:
```suggestion
                // Compile `c2rust-instrument` with the same profile as `c2rust-pdg` was.
```
Just a grammar fix.

pdg/src/main.rs
```diff
@@ -175,21 +175,134 @@ fn main() -> eyre::Result<()> {
...
+
+    pub fn repo_dir() -> eyre::Result<PathBuf> {
+        let crate_dir = env::var("CARGO_MANIFEST_DIR")?;
+        let repo_dir = Path::new(&crate_dir)
+            .parent()
+            .ok_or_else(|| eyre!("`$CARGO_MANIFEST_DIR` should have a parent"))?;
+        Ok(repo_dir.to_owned())
+    }
+
+    fn pdg_snapshot_inner(
```
fw-immunant, line 244: This function does several things without explaining beforehand or inline what it's trying to do, and its name isn't enough to explain what exactly it should do. Let's have a doc comment on it along the lines of the comment in `pdg.sh`.

fw-immunant (changes requested):
Requested comment changes.

pdg/src/main.rs
kkysen: Okay, I'll add that.  Should it be on `pdg_snapshot_inner` or on `pdg_snapshot`, as `pdg_snapshot` is meant to be the one you call (the inner one is so the long function doesn't have to be generic)?

fw-immunant: As `pdg_snapshot_inner` is where the nontrivial logic is implemented, the logic should be explained there. A high-level doc comment on the wrapper too wouldn't hurt, of course.

fw-immunant (approved):

aneksteind (approved):

pdg/src/main.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/636/commits/84bb03335cf512509a3eaeba5fc71c3a2ebcd57d.

PR #638, opened 2022-08-23
Refactor variadics

kkysen:
These are various refactors of the variadics code I made when reviewing #612:

Notably, I felt `match_or!` was quite un-idiomatic as it contained hidden non-local control flow (`return None`) and defined variables in a non-intuitive way.  Thus, I removed these.  Using `?` makes the control flow not any longer but much clearer, and now we have to use actual `let`s, but this improves readability a lot and allows inlay type hints to work, which are very useful.

It might be easier to review by reviewing each commit individually.

### DIFF
--- a/c2rust-transpile/src/translator/variadic.rs
+++ b/c2rust-transpile/src/translator/variadic.rs
@@ -107,12 +107,13 @@ impl<'c> Translation<'c> {
         CExprKind::DeclRef(_, decl_id, _) => decl_id }
         match_or! { [self.ast_context[decl_id].kind]
         CDeclKind::Function { ref name, .. } => name }
-        match (name.as_str(), args.as_slice()) {
-            ("__builtin_va_start", &[expr, _]) => self.match_vastart(expr).map(VaPart::Start),
-            ("__builtin_va_copy", &[dst_expr, src_expr]) => self
+        let name = name.strip_prefix("__builtin_va_")?;
+        match (name, args.as_slice()) {
+            ("start", &[expr, _]) => self.match_vastart(expr).map(VaPart::Start),
+            ("copy", &[dst_expr, src_expr]) => self
                 .match_vacopy(dst_expr, src_expr)
                 .map(|(did, sid)| VaPart::Copy(did, sid)),
-            ("__builtin_va_end", &[expr]) => self.match_vaend(expr).map(VaPart::End),
+            ("end", &[expr]) => self.match_vaend(expr).map(VaPart::End),
             _ => None,
         }
     }
--- a/c2rust-transpile/src/translator/variadic.rs
+++ b/c2rust-transpile/src/translator/variadic.rs
@@ -34,13 +34,13 @@ impl<'c> Translation<'c> {
     }
 
     pub fn match_vastart(&self, expr: CExprId) -> Option<CDeclId> {
-        let ast_context = &self.ast_context;
+        let ast = &self.ast_context;
 
         // `struct`-based `va_list` (e.g. x86_64).
         let match_vastart_struct = || {
-            match_or! { [ast_context[expr].kind]
+            match_or! { [ast[expr].kind]
             CExprKind::ImplicitCast(_, e, _, _, _) => e }
-            match_or! { [ast_context[e].kind]
+            match_or! { [ast[e].kind]
             CExprKind::DeclRef(_, va_id, _) => va_id }
             Some(va_id)
         };
@@ -49,11 +49,11 @@ impl<'c> Translation<'c> {
         //
         // Supporting this pattern is necessary to transpile apache httpd.
         let match_vastart_struct_member = || {
-            match_or! { [ast_context[expr].kind]
+            match_or! { [ast[expr].kind]
             CExprKind::ImplicitCast(_, me, _, _, _) => me }
-            match_or! { [ast_context[me].kind]
+            match_or! { [ast[me].kind]
             CExprKind::Member(_, e, _, _, _) => e }
-            match_or! { [ast_context[e].kind]
+            match_or! { [ast[e].kind]
             CExprKind::DeclRef(_, va_id, _) => va_id }
             Some(va_id)
         };
@@ -63,20 +63,20 @@ impl<'c> Translation<'c> {
         // Supporting this pattern is necessary to transpile
         // [graphviz](https://gitlab.com/graphviz/graphviz/-/blob/5.0.0/lib/sfio/sftable.c#L321).
         let match_vastart_struct_pointer_member = || {
-            match_or! { [ast_context[expr].kind]
+            match_or! { [ast[expr].kind]
             CExprKind::ImplicitCast(_, me, _, _, _) => me }
-            match_or! { [ast_context[me].kind]
+            match_or! { [ast[me].kind]
             CExprKind::Member(_, ie, _, _, _) => ie }
-            match_or! { [ast_context[ie].kind]
+            match_or! { [ast[ie].kind]
             CExprKind::ImplicitCast(_, e, _, _, _) => e }
-            match_or! { [ast_context[e].kind]
+            match_or! { [ast[e].kind]
             CExprKind::DeclRef(_, va_id, _) => va_id }
             Some(va_id)
         };
 
         // `char *` pointer-based `va_list` (e.g. x86).
         let match_vastart_pointer = || {
-            match_or! { [ast_context[expr].kind]
+            match_or! { [ast[expr].kind]
             CExprKind::DeclRef(_, va_id, _) => va_id }
             Some(va_id)
         };
--- a/c2rust-transpile/src/translator/variadic.rs
+++ b/c2rust-transpile/src/translator/variadic.rs
@@ -108,19 +108,11 @@ impl<'c> Translation<'c> {
         match_or! { [self.ast_context[decl_id].kind]
         CDeclKind::Function { ref name, .. } => name }
         match (name.as_str(), args.as_slice()) {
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/translator/variadic.rs
```diff
@@ -9,130 +9,110 @@ pub enum VaPart {
...
-            }
-
+        let (func, args) = match_or! { [self.ast_context[expr].kind]
+        CExprKind::Call(_, func, ref args) => (func, args) }?;
+        let fexp = match_or! { [self.ast_context[func].kind]
+        CExprKind::ImplicitCast(_, fexp, CastKind::BuiltinFnToFnPtr, _, _) => fexp }?;
+        let decl_id = match_or! { [self.ast_context[fexp].kind]
+        CExprKind::DeclRef(_, decl_id, _) => decl_id }?;
+        let name = match_or! { [self.ast_context[decl_id].kind]
+        CDeclKind::Function { ref name, .. } => name }?;
```
fw-immunant, line 108: Indent the line continuations within the `match_or` macro invocations here to make it clearer that they are not their own expressions at the same nesting level as the `let`s.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -9,130 +9,110 @@ pub enum VaPart {
...
+        let name = name.strip_prefix("__builtin_va_")?;
+        match (name, args.as_slice()) {
+            ("start", &[expr, _]) => self.match_vastart(expr).map(VaPart::Start),
+            ("copy", &[dst_expr, src_expr]) => self
+                .match_vacopy(dst_expr, src_expr)
+                .map(|(did, sid)| VaPart::Copy(did, sid)),
+            ("end", &[expr]) => self.match_vaend(expr).map(VaPart::End),
```
fw-immunant, line 115: NAK on factoring out the `"__builtin_va_"` prefix, because these are magic strings that are good to be able to grep.

c2rust-transpile/src/translator/variadic.rs
kkysen: I'm pretty sure I tried that but `cargo fmt` changed it back to this.  Let me check, though.  The macro would have to be re-written to work better with `rustfmt`.

c2rust-transpile/src/translator/variadic.rs
kkysen: What do you mean by NAK?

I agree with your point, though.  I can remove this commit.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -9,130 +9,110 @@ pub enum VaPart {
...
-        ) -> Option<CDeclId> {
-            match_or! { [ast_context[expr].kind]
-            CExprKind::ImplicitCast(_, me, _, _, _) => me }
-            match_or! { [ast_context[me].kind]
-            CExprKind::Member(_, ie, _, _, _) => ie }
-            match_or! { [ast_context[ie].kind]
-            CExprKind::ImplicitCast(_, e, _, _, _) => e }
-            match_or! { [ast_context[e].kind]
-            CExprKind::DeclRef(_, va_id, _) => va_id }
+        // `struct`-based `va_list` (e.g. x86_64) where `va_list` is accessed as a member of a `struct *` pointer.
```
fw-immunant, line 61: "`struct *`" or "`struct` pointer" are fine ways to describe this pattern, but "`struct *` pointer" makes it sound like a double indirection could be present, which is not the case.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -9,130 +9,110 @@ pub enum VaPart {
...
+            CExprKind::DeclRef(_, va_id, _) => va_id }?;
             Some(va_id)
-        }
+        };
 
-        // char pointer-based va_list (e.g. x86)
-        fn match_vastart_pointer(ast_context: &TypedAstContext, expr: CExprId) -> Option<CDeclId> {
-            match_or! { [ast_context[expr].kind]
-            CExprKind::DeclRef(_, va_id, _) => va_id }
+        // `char *` pointer-based `va_list` (e.g. x86).
```
fw-immunant, line 77: Same here, pick one of "`char *`" or "`char` pointer", to match convention in other comment.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -9,130 +9,110 @@ pub enum VaPart {
...
-            decls.contains(&decl_id)
-        } else {
-            false
-        }
+        self.function_context
+            .borrow()
+            .va_list_decl_ids
+            .as_ref()
+            .map(|decls| decls.contains(&decl_id))
+            .unwrap_or_default()
```
fw-immunant, line 33: `Default` for `bool` returning false is pretty arbitrary and relying on it is less clear here than `.unwrap_or(false)` would be.

fw-immunant (changes requested):
This looks better than the code did before, thanks.

Various requested changes described inline. Be sure to indent continued lines within all `match_or!` macro invocations.

c2rust-transpile/src/translator/variadic.rs
kkysen: True, I'll just use `struct *` then.

c2rust-transpile/src/translator/variadic.rs
kkysen: I can't indent them according to `cargo fmt`, but I can make each `match_or!` on 1 line only.

c2rust-transpile/src/translator/variadic.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/638/commits/b4428a365092f531b1c034b69e0d038ad404a355.  Explanation in the revert commit.

c2rust-transpile/src/translator/variadic.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/638/commits/8fae26f66fa8e21116cb3f8c8e646a50846152de.

c2rust-transpile/src/translator/variadic.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/638/commits/8fae26f66fa8e21116cb3f8c8e646a50846152de.

c2rust-transpile/src/translator/variadic.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/638/commits/eaccb5cfa2440f79eb0c350f34a40b92b6ad0346.

c2rust-transpile/src/translator/variadic.rs
kkysen: Would this look better?

```rust
        let (func, args) = match_or! { [ast[expr].kind] CExprKind::Call(_, func, ref args) => (func, args) }?;
        let fexp = match_or! { [ast[func].kind] CExprKind::ImplicitCast(_, fexp, CastKind::BuiltinFnToFnPtr, _, _) => fexp }?;
        let decl_id = match_or! { [ast[fexp].kind] CExprKind::DeclRef(_, decl_id, _) => decl_id }?;
        let name = match_or! { [ast[decl_id].kind] CDeclKind::Function { ref name, .. } => name }?;
```

c2rust-transpile/src/translator/variadic.rs
thedataking: I would prefer you revert commit 3866a89 altogether. The existing code was good enough and your rewritten code needed improving (I had the same thought as Frances that using `.unwap_or_default()` is less clear so). Please avoid code churn where the effort to review it and the risk of introducing bugs doesn't outweigh the benefits.

c2rust-transpile/src/translator/variadic.rs
thedataking: What we had before was perfectly fine. For instance
>  // `struct`-based `va_list` (e.g. x86_64) where `va_list` is accessed as a member of a `struct *`.

is not an improvement over the previous comment

>  // struct-based va_list (e.g. x86_64) where va_list is accessed as a member of a struct pointer

I'd prefer it if you can just revert to the old comments.

thedataking (changes requested):
There are some improvements on this PR but in several places (see comments) I think you should have left things as is vs. making changes that needs reviewing and risks breaking stuff. Please make it a higher priority to avoid code churn going forward.

c2rust-transpile/src/translator/variadic.rs
kkysen: Okay, sorry about that, I can revert those two commits.

c2rust-transpile/src/translator/variadic.rs
kkysen: I thought the `` ` ``s increased readability; they definitely do to me.  And there's no real risk of churn here, as it's just a comment.

c2rust-transpile/src/translator/variadic.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/638/commits/b31901c159df78f5142e2a02430527f946025b81 assuming that looks good.

c2rust-transpile/src/translator/variadic.rs
kkysen: Reverted in https://github.com/immunant/c2rust/pull/638/commits/1a5085a40e5f7f61a510869a8a069fab14a0c36f.

PR #639, opened 2022-08-24
Add `c2rust-instrument --rustflags`

kkysen:
Reopening #629.  It was merged early.  It was done and ready to merge, but I was waiting to merge it directly into `master` once the 2 PRs in front of it merged.

Add a `--rustflags` argument to `c2rust-instrument`, as it much more easily allows you specify `$RUSTFLAGS` for the instrumentation only (and not all the other `cargo`s).  Otherwise, you have to split up the `cargo` commands and run `c2rust-instrument` directly, which can be a real pain during development.

I'm also not sure why `cargo` doesn't have this, too.

Perhaps this should only apply to `rustc_wrapper`s where we actually do instrumentation.  What do others think?

### DIFF
--- a/dynamic_instrumentation/src/main.rs
+++ b/dynamic_instrumentation/src/main.rs
@@ -58,18 +58,20 @@ struct Args {
     #[clap(long)]
     set_runtime: bool,
 
-    /// `$RUSTFLAGS`, but as a CLI argument, not an inheritable environment variable.
-    /// These flags are appended (space-separated) to the existing `$RUSTFLAGS`, if it exists.
+    /// Set `$RUSTFLAGS` for the instrumented `cargo`.
     ///
-    /// The reason this exists is twofold:
-    ///
-    /// 1. It would be convenient if `cargo` itself had such a `--rustflags` argument,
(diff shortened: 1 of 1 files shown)


fw-immunant (approved):
My r+ from before stands (sorry for prematurely merging!). On the topic of whether we should limit the scope of the `--rustflags` flags to `rustc_wrapper`s where we actually do instrumentation: if the primary use case benefits from it, sure--but I'm not really sure which this use-case is and I'm fine with this landing unchanged. We can tweak behavior later separately if we want.

kkysen:
> My r+ from before stands (sorry for prematurely merging!). On the topic of whether we should limit the scope of the `--rustflags` flags to `rustc_wrapper`s where we actually do instrumentation: if the primary use case benefits from it, sure--but I'm not really sure which this use-case is and I'm fine with this landing unchanged. We can tweak behavior later separately if we want.

Yeah, there seem to be some benefits to applying rustflags to all of them and some benefits to just the instrumented crates, but we can improve this as we come across more pressing use cases.

rinon (approved):

PR #640, opened 2022-08-24
`impl Display` for pdg

kkysen:
This refactors `c2rust-pdg`'s `main.rs` code so that the printing logic in `main` is moved to an `impl Display for PdgRepr`.  Thus, we can either `println!` it for `main` or `.to_string()` it for `insta::assert_display_snapshot!` (not yet implemented in this PR; see #636).

### DIFF
--- a/pdg/src/main.rs
+++ b/pdg/src/main.rs
@@ -24,10 +24,11 @@ mod query;
 mod util;
 
 use builder::{construct_pdg, read_event_log};
+use c2rust_analysis_rt::{events::Event, metadata::Metadata};
 use clap::{Parser, ValueEnum};
 use color_eyre::eyre;
+use graph::Graphs;
 use std::{
-    collections::HashSet,
     fmt::{self, Display, Formatter},
     path::{Path, PathBuf},
     sync::Once,
@@ -36,7 +37,7 @@ use std::{
 use crate::builder::read_metadata;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, ValueEnum)]
-enum ToPrint {
+pub enum ToPrint {
     Graphs,
     Counts,
     Events,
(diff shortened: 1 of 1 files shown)


pdg/src/main.rs
```diff
@@ -50,10 +51,104 @@ impl Display for ToPrint {
...
+        let metadata = read_metadata(metadata_path)?;
+        let graphs = construct_pdg(&events, &metadata);
+        Ok(Self {
+            events,
+            metadata,
+            graphs,
+        })
+    }
+
+    pub fn repr<'a>(&'a self, to_print: &'a [ToPrint]) -> PdgRepr<'a> {
```
aneksteind, line 72: I think `PdgRepr` should be constructed from a `Pdg`, not the other way around

aneksteind: i.e. `PdgRepr::new` should do this IMO

kkysen: Can `PdgRepr::new` be the real implementation and then still have a `Pdg::repr` method?

pdg/src/main.rs
```diff
@@ -50,10 +51,104 @@ impl Display for ToPrint {
...
+                .collect::<Vec<_>>();
+            if should_print(ToPrint::Graphs) {
+                writeln!(f, "{graph}")?;
+            }
+            if should_print(ToPrint::WritePermissions) {
+                writeln!(f, "nodes_that_need_write = {needs_write:?}")?;
+            }
+            if should_print(ToPrint::Graphs) || should_print(ToPrint::WritePermissions) {
+                writeln!(f)?;
+            }
```
fw-immunant, line 130: I would move this `if` to surround the entire `for graph...` loop, and then make this `writeln!` unconditional. That would clarify that the `writeln!(f)` terminates each graph printed and would avoid iterating if we don't need to look at graphs.

pdg/src/main.rs
```diff
@@ -50,10 +51,104 @@ impl Display for ToPrint {
...
+                let func = &metadata.functions[func_hash];
+                writeln!(f, "({func}:{local:?}) => {p:?}")?;
+            }
+        }
+
+        for graph in &graphs.graphs {
+            let needs_write = graph
+                .needs_write_permission()
+                .map(|node_id| node_id.as_usize())
+                .collect::<Vec<_>>();
```
fw-immunant, line 121: This can be moved inside of `if should_print(ToPrint::WritePermissions) { ... }`.

fw-immunant (approved):
A couple refactors suggested, but not blocking merge. LGTM overall.

pdg/src/main.rs
kkysen: Actually, the fields of `PdgRepr` are `pub`, so it can be constructed on its own.  A `PdgRepr::new` wouldn't really offer any advantage, and I'd like to keep `Pdg::repr` for ergonomics.

pdg/src/main.rs
kkysen: I'd like to change this in a separate PR, since this right now is just moving code, not changing anything.

pdg/src/main.rs
kkysen: (Same with this) I'd like to change this in a separate PR, since this right now is just moving code, not changing anything.

pdg/src/main.rs
kkysen: Wait, this is not quite equivalent.  If `--print write-permissions` is specified, but no `--print graphs`, the write permissions and no graphs should be printed.

fw-immunant: A separate PR for these changes is fine.

I think what I requested is equivalent, though--the three `writeln!`s in the loop are guarded by `should_print(ToPrint::Graphs)`, `should_print(ToPrint::WritePermissions)`, and their disjunction, respectively. Making the whole loop conditional on the disjunction doesn't impact the first two conditionals as they're refinements of it, but it makes the third fully redundant, so it can be made unconditonal:
```rust
        if should_print(ToPrint::Graphs) || should_print(ToPrint::WritePermissions) {
            for graph in &graphs.graphs {
                let needs_write = graph
                    .needs_write_permission()
                    .collect::<Vec<_>>();
                if should_print(ToPrint::Graphs) {
                    writeln!(f, "{graph}")?;
                }
                if should_print(ToPrint::WritePermissions) {
                    writeln!(f, "nodes_that_need_write = {needs_write:?}")?;
                }
                writeln!(f)?;
            }
        }
```

kkysen: Oh like that you mean.  I'm not sure that's much of an advantage, since it's just hosting the `if` around the `writeln!(f)` to outside the loop.

fw-immunant: The concrete advantage is that it skips iterating over the graphs if nothing about them needs to be printed. It's not a huge deal, but is a small "pay for what you use".

pdg/src/main.rs
aneksteind: I'm not sold on the utility of `Pdg` knowing anything about what a `PdgRepr` is, that's my main concern. I think it should be constructed on its own, how is less important.

kkysen: I see `PdgRepr` more as an `impl Display` for `Pdg`, but with `&[ToPrint]`s, too.  I'm not sure why `Pdg` can't know about `PdgRepr`, especially as its such a simple and transparent struct.

kkysen: I'm going to merge this as is if that's okay.  We can easily change it back afterward if we need to as it's only a couple of lines of code.

pdg/src/main.rs
kkysen: Fixed in #645.

PR #618, opened 2022-08-16
Extract `find_llvm_config` so it's reusable

kkysen:
Extract `find_llvm_config` from `c2rust-ast-exporter/build.rs` to `build-utils` so that it can be reused elsewhere (like `FileCheck` resolving in `c2rust-analyze`'s tests).

PR #642, opened 2022-08-24
Extract `find_llvm_config` so it's reusable

kkysen:
Extract `find_llvm_config` from `c2rust-ast-exporter/build.rs` to `build-utils` so that it can be reused elsewhere (like `FileCheck` resolving in `c2rust-analyze`'s tests).

I accidentally messed up the branches in #618 (I messed things up, then fixed it, but pushed them in the wrong order, so GitHub thought I merged the PR with 0 commits).  This is the same as #618.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,6 +192,7 @@ name = "c2rust-ast-exporter"
 version = "0.16.0"
 dependencies = [
  "bindgen",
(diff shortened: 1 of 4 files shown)


fw-immunant (approved):
Trivial code motion, LGTM.

PR #644, opened 2022-08-25
Remove `camino`

kkysen:
`camino` is now only used in 2 places and is not very necessary anymore.  This replaces the `&camino::Utf8Path` usages with `&str` and custom error messages, and then removes `camino` as a dependency from the whole workspace.

### DIFF
--- a/c2rust/src/main.rs
+++ b/c2rust/src/main.rs
@@ -1,5 +1,4 @@
 use anyhow::anyhow;
-use camino::Utf8Path;
 use clap::{crate_authors, App, AppSettings, Arg};
 use git_testament::{git_testament, render_testament};
 use is_executable::IsExecutable;
@@ -7,7 +6,7 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::env;
 use std::ffi::OsStr;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::process;
 use std::process::Command;
 use std::str;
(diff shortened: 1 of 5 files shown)


c2rust/src/main.rs
```diff
@@ -29,12 +28,11 @@ impl SubCommand {
...
     pub fn find_all() -> anyhow::Result<Vec<Self>> {
         let c2rust = env::current_exe()?;
-        let c2rust_name: &Utf8Path = c2rust
+        let c2rust_name = c2rust
             .file_name()
-            .map(Path::new)
-            .ok_or_else(|| anyhow!("no file name: {}", c2rust.display()))?
-            .try_into()?;
-        let c2rust_name = c2rust_name.as_str();
+            .ok_or_else(|| anyhow!("no file name for c2rust: {}", c2rust.display()))?
```
rinon, line 33: Is there a reason you didn't use `.with_context()` here?

rinon (approved):

rinon:
LGTM with a minor nitpick that you might want to fix

c2rust/src/main.rs
kkysen: Ah, I didn't know `Option` `impl`s `Context`.  I can change that.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/644/commits/0a449f71e10952eb0f02822e1cc2ee04ed1806b3.

kkysen: Actually, I reverted that.  There are tons of `.ok_or_else(|| anyhow!(...))`s everywhere, so I think if we want to change it, we should just change it everywhere.

PR #645, opened 2022-08-26
Skip some calculations if not being printed in a snapshot

kkysen:
Addresses https://github.com/immunant/c2rust/pull/640#discussion_r954446415.


fw-immunant (approved):
LGTM.

PR #647, opened 2022-08-26
Add an `args` arg to `pdg_snapshot` so args can be passed to the instrumented binary

kkysen:
Add an `args` arg to `pdg_snapshot` so args can be passed to the instrumented binary.

This isn't needed for `analysis/test` but is for `lighttpd`, for example.

Note that `fn pdg_snapshot` can be used to run `lighttpd` with args now, but it's not set up do that from the CLI yet, so just keep using `pdg.sh` for that in the meantime.

### DIFF
--- a/pdg/src/main.rs
+++ b/pdg/src/main.rs
@@ -194,6 +194,7 @@ fn main() -> eyre::Result<()> {
 mod tests {
     use std::{
         env,
+        ffi::OsStr,
         fmt::Display,
         path::{Path, PathBuf},
         process::Command,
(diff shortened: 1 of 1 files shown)


aneksteind (approved):
this LGTM but please add a comment to the PR or description showing the new usage for something like lighttpd

kkysen:
To the PR description, or inline in docs?  Also, it's not quite ready to be used with `lighttpd` yet.  For example, the `lighttpd` PDG is never deterministic so snapshotting isn't really helping.  Keep using the old `pdg.sh` way for now.  I'll add code showing how to use it for `lighttpd` once it's ready for that.

aneksteind:
PR description or comment here is what I had in mind, because I anticipate this changing again and we can avoid the need to update (probably unnecessary) inline documentation that way

kkysen:
Okay, I can update that then.

PR #650, opened 2022-08-26
UB-free `analysis/test` WIP

kkysen:
All UB in `analysis/test` should be gone according to `miri`.  Now I can't reproduce the failures locally.  Let's see about CI.

WIP fix for:
* #680
* #681
* #682

Fixes completed in:
* #653
* #683
* #684
* #686
* #698

### DIFF
--- a/analysis/test/src/main.rs
+++ b/analysis/test/src/main.rs
@@ -1,6 +1,7 @@
 #![feature(extern_types)]
 #![feature(label_break_value)]
 #![feature(rustc_private)]
+// #![feature(c_variadic)]
 
 #![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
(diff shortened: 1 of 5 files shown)


kkysen:
Well, this is still failing CI, welp.

Some ideas for investigating further that we discussed:
* Trying to replicate the CI environment locally as much as possible, like running with reduced core count and under heavy load.
* Check to see if `rustc` and our instrumentation is being deterministic.
* Check to see if any events are being lost.  Perhaps try processing the events on the main thread to keep things single-threaded, as perhaps there's a bug in there, especially around finalization when the main thread finishes first.

aneksteind:
What's the high level difference between this and #683 ?

kkysen:
Closing as this was a WIP and is now complete in:
* #653
* #683
* #684
* #686
* #698

PR #651, opened 2022-08-26
Pfb/extra info correct

boyland-pf:
The definition of unique has been changed to better match the real rust rules for mutable references. The algorithm was simplified to merely check all sibling relationships between nodes. It continues to be O(n log n).

- If a node and its sibling overlap and are not both field nodes of different fields, they are guaranteed to conflict, which makes their descendents and parent non-unique.
- If two nodes conflict (i.e. they overlap, are aliases of one another, and neither is a descendant of the other) then their ancestor nodes which are children of the two nodes' common ancestor will overlap and will not be field nodes of different fields.

### DIFF
--- a/pdg/src/info.rs
+++ b/pdg/src/info.rs
@@ -2,9 +2,8 @@ use crate::graph::{Graph, GraphId, Graphs, Node, NodeId, NodeKind};
 use itertools::Itertools;
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::Field;
-use std::collections::HashMap;
-use std::collections::HashSet;
 use std::cmp;
+use std::collections::{HashMap, HashSet};
 use std::fmt::{self, Debug, Display, Formatter};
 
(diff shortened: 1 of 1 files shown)


boyland-pf:
Following up from https://github.com/immunant/c2rust/pull/587

pdg/src/info.rs
```diff
@@ -52,152 +183,413 @@ fn node_does_neg_offset(n: &Node) -> bool {
...
+    #[test]
+    fn unique_interleave() {
+        let mut g = Graph::default();
+        // let mut a = 0;   // A
+        // let b = &mut a;  // B1
+        // *b = 0;          // B2
+        // let c = &mut a;  // C1
+        // *c = 0;          // C2
+        // *b = 0;          // B3
+        // *c = 0;          // C3
```
kkysen, line 270: You could put this in a doc comment for the function and have it run as a doc test.  Same for all of these test functions.

Issue #649, opened 2022-08-26
Generated wrapper `main` leaks memory of `args`


kkysen:
This leaks the memory of `args`:

```rust
pub fn main() {
    let mut args: Vec<*mut libc::c_char> = Vec::new();
    for arg in ::std::env::args() {
        println!("{:?}", arg);
        args.push(
            ::std::ffi::CString::new(arg)
                .expect("Failed to convert argument into CString.")
                .into_raw(),
        );
    }
    args.push(::std::ptr::null_mut());
    unsafe {
        main_0(
            (args.len() - 1) as libc::c_int,
            args.as_mut_ptr() as *mut *mut libc::c_char,
        );
    }
}
```

We should either free it at the end by using `CString::from_raw` and dropping it, or by making `args: Vec<CString>` and then another `args: Vec<*mut libc::c_char>` that points to the memory owned by the first `args`.  I think the latter approach is better.

PR #653, opened 2022-08-26
Fix leaking `args` in `analysis/test`'s `main`

kkysen:
This fixes the memory leakage of `args` in `main` in `analysis/test` by not `.into_raw()`ing the `CString`s, but storing them in a `Vec` and using `.as_ptr()`.


fw-immunant (approved):

aneksteind:
I'm not sure if this portion of the code was output of the transpiler or not. We should check that this behavior you've fixed isn't present generally.

kkysen:
I think (?) it is.  I opened #649 for that, but wanted to fix it here first since it's simpler.

aneksteind:
Ah, ok -- missed #649. That sounds good.

PR #637, opened 2022-08-23
Add an `analysis/test` snapshot test for `--release`

kkysen:
This adds a snapshot test for `analysis/test`, but compiled in `--release` mode.  This generates very different code for `analysis/test`, so it tests different instrumentation and runtime paths, and allows us to see how other PRs change the snapshot for both `debug` and `release` modes.

PR #654, opened 2022-08-29
Add address-taken local support in dynamic analysis

aneksteind:
We want to track usage of the memory of address-taken locals in dynamic analysis and in the pointer derivation graph for static analysis. This adds support for that. This also tracks operations on the local's memory even before the local becomes address-taken. Additionally, support for address-taken arguments have been added. Taking references of locals with `&` and `&mut` are also considered taking their address. 

Support for address-taken locals is done in three stages: identifying where address-taken locals get their address taken, a rewriting stage, and post-processing. The first two stages are accomplished in dynamic analysis with the introduction of two new MIR visitors and the last occurs during construction of the pointer derivation graph. 

The first traverses the AST and identifies where locals have their addresses taken and stores that information in a data structure for reference in the next stage. The next stage is another MIR visit that rewrites all address-taken locals in terms of their address. For example, if `_x` is an address-taken local and `_y` stores its address, `_x` is rewritten as `*_y`. These two stages are conducted before instrumentation points are collected in the third and final MIR visitor.

The final stage ensures that all duplicate instances of taking the address of a particular local are treated as copies of its address.

Below are a few examples of what is now supported:

```rust
pub fn test_addr_taken() {
    let x = 2;
    let y = 2 + x;
    let px = std::ptr::addr_of!(x);
    let z = x + y;
}

pub fn test_addr_taken_cond(cond: bool) {
    // let x = if cond { ... } else { ... }; ... &x ...
    let x = if cond { 1 } else { 2 };
    let y = &x;
    let a = x;
}

pub fn test_addr_taken_init_cond(cond: bool) {
    // let mut x; if cond { x = 1; ... &x ... }; x = 2; ... &x ...
    let mut x;
    let mut y;
    if cond {
        x = 1;
        y = &x;
    }
    x = 2;
    let z = x;
}

#[no_mangle]
pub fn test_addr_taken_loop() {
    // loop { let x = ...; ... &x ... }
    let mut count = 0;
    loop {
        let x = 2;
        let z = if count % 2 == 0 { &x } else { &1 };
        if count >= 3 {
            break;
        }
        count += *z;
    }
}

#[no_mangle]
pub fn test_addr_taken_arg(mut t: T) {
    t.field3 = 0 as *const S;
    let z = &t;
}
```

### DIFF
--- a/analysis/runtime/src/events.rs
+++ b/analysis/runtime/src/events.rs
@@ -6,7 +6,7 @@ use std::fmt::Formatter;
 
 pub type Pointer = usize;
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct Event {
     pub mir_loc: MirLocId,
     pub kind: EventKind,
@@ -55,6 +55,12 @@ pub enum EventKind {
     /// The pointer appears as the address of a store operation.
     StoreAddr(Pointer),
 
+    /// An address-taken local is assigned to, which is semantically the
+    /// same as [`StoreAddr`](Self::StoreAddr), but needs to be distinguished because
+    /// storing to an address-taken local does not imply write permissions
+    /// are necessary for its underlying address.
+    StoreAddrTaken(Pointer),
+
     /// The pointer that appears as the address result of addr_of(Local)
     AddrOfLocal(Pointer, Local),
 
@@ -69,6 +75,12 @@ pub enum EventKind {
     StoreValue(Pointer),
 
     Offset(Pointer, isize, Pointer),
+
+    /// Marks the start of events in a new function body.
+    /// Used to distinguish address-taken locals that are treated
+    /// as copies and ones that aren't; all but the first [`AddrOfLocal`](Self::AddrOfLocal)
+    /// events after a [`BeginFuncBody`](Self::BeginFuncBody) event are treated as copies.
+    BeginFuncBody,
 
     Done,
 }
@@ -90,8 +102,10 @@ impl Debug for EventKind {
             } => write!(f, "realloc(0x{:x}, {}) -> 0x{:x}", old_ptr, size, new_ptr),
             Ret(ptr) => write!(f, "ret(0x{:x})", ptr),
             Done => write!(f, "done"),
+            BeginFuncBody => write!(f, "begin func body"),
             LoadAddr(ptr) => write!(f, "load(0x{:x})", ptr),
             StoreAddr(ptr) => write!(f, "store(0x{:x})", ptr),
+            StoreAddrTaken(ptr) => write!(f, "store(0x{:x})", ptr),
             CopyRef => write!(f, "copy_ref"),
             AddrOfLocal(ptr, _) => write!(f, "addr_of_local = 0x{:x}", ptr),
             ToInt(ptr) => write!(f, "to_int(0x{:x})", ptr),
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -436,6 +436,11 @@ pub fn shared_ref_foo(x: &u8) -> &u8 {
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn test_addr_taken_arg(mut t: T) {
+    t.field3 = 0 as *const S;
+    let z = &t;
+}
+#[no_mangle]
 pub unsafe extern "C" fn test_shared_ref() {
     let x = 2;
     let y = &x;
@@ -471,6 +476,46 @@ pub unsafe extern "C" fn test_ref_field() {
     s.field4.field4 = s.field4.field4;
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn test_addr_taken() {
+    let x = 2;
+    let y = 2 + x;
+    let px = std::ptr::addr_of!(x);
+    let z = x + y;
+}
+#[no_mangle]
+pub fn test_addr_taken_cond(cond: bool) {
+    // let x = if cond { ... } else { ... }; ... &x ...
+    let x = if cond { 1 } else { 2 };
+
+    let y = &x;
+    let a = x;
+}
+#[no_mangle]
+pub fn test_addr_taken_init_cond(cond: bool) {
+    // let mut x; if cond { x = 1; ... &x ... }; x = 2; ... &x ...
+    let mut x;
+    let mut y;
+    if cond {
+        x = 1;
+        y = &x;
+    }
+    x = 2;
+    let z = x;
+}
+#[no_mangle]
+pub unsafe fn test_addr_taken_loop() {
+    // loop { let x = ...; ... &x ... }
+    let mut count = 0;
+    loop {
+        let x = 2;
+        let z = if count % 2 == 0 { &x } else { &1 };
+        if count >= 3 {
+            break;
+        }
+        count += *z;
+    }
+}
 #[no_mangle]
 pub unsafe extern "C" fn test_realloc_reassign() {
     let mut s = malloc(::std::mem::size_of::<S>() as libc::c_ulong);
@@ -597,6 +642,69 @@ pub unsafe extern "C" fn insertion_sort(n: libc::c_int, p: *mut libc::c_int) {
     }
 }
 
+/// This is a minimal breaking example from `lighttpd` where instrumented code failed to compile.
+#[no_mangle]
+#[cold]
+unsafe extern "C" fn log_error_va_list_impl(
+    mut errh: *const libc::c_void,
+    filename: *const libc::c_char,
+    line: libc::c_uint,
+    fmt: *const libc::c_char,
+    mut ap: ::std::ffi::VaList,
+    perr: libc::c_int,
+) {
+    let mut prefix: [libc::c_char; 40] = [0; 40];
+    vsprintf(prefix.as_mut_ptr(), fmt, ap.as_va_list());
+}
+pub unsafe extern "C" fn log_error(
+    errh: *mut libc::c_void,
+    filename: *const libc::c_char,
+    line: libc::c_uint,
+    mut fmt: *const libc::c_char,
+    mut args: ...
+) {
+    let mut ap: ::std::ffi::VaListImpl;
+    ap = args.clone();
+    log_error_va_list_impl(errh, filename, line, fmt, ap.as_va_list(), 0 as libc::c_int);
+}
+
+/*
+    This is a minimal breaking example from lighttpd where instrumented code failed
+    to compile.
+*/
+fn vsprintf(_: *mut libc::c_char, _: *const libc::c_char, _: ::std::ffi::VaList) -> libc::c_int {
+    0
+}
+#[no_mangle]
+pub unsafe extern "C" fn ErrorMsg(
+    mut filename: *const libc::c_char,
+    mut lineno: libc::c_int,
+    mut format: *const libc::c_char,
+    mut args: ...
+) {
+    let mut errmsg: [libc::c_char; 10000] = [0; 10000];
+    let mut prefix: [libc::c_char; 40] = [0; 40];
+    let mut ap: ::std::ffi::VaListImpl;
+    ap = args.clone();
+    if lineno > 0 as libc::c_int {
+        sprintf(
+            prefix.as_mut_ptr(),
+            b"%.*s:%d: \0" as *const u8 as *const libc::c_char,
+            30 as libc::c_int - 10 as libc::c_int,
+            filename,
+            lineno,
+        );
+    } else {
+        sprintf(
+            prefix.as_mut_ptr(),
+            b"%.*s: \0" as *const u8 as *const libc::c_char,
+            30 as libc::c_int - 10 as libc::c_int,
+            filename,
+        );
+    }
+    vsprintf(errmsg.as_mut_ptr(), format, ap.as_va_list());
+}
+
 unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> libc::c_int {
     simple();
     exercise_allocator();
@@ -621,6 +729,7 @@ unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> lib
     test_realloc_reassign();
     test_realloc_fresh();
     test_load_addr();
+    test_load_addr();
     test_overwrite();
     test_store_addr();
     test_load_other_store_self();
@@ -634,7 +743,26 @@ unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> lib
     let nums = &mut [2i32, 5i32, 3i32, 1i32, 6i32];
     insertion_sort(nums.len() as libc::c_int, nums as *mut libc::c_int);
     test_ref_field();
+    test_addr_taken();
+    let mut t = T {
+        field: 0i32,
+        field2: 0u64,
+        field3: 0 as *const S,
+        field4: 0i32,
+    };
+    test_addr_taken_arg(t);
+    test_addr_taken_loop();
+    test_addr_taken_cond(true);
+    test_addr_taken_cond(false);
+    test_addr_taken_init_cond(true);
+    test_addr_taken_init_cond(false);
     return 0i32;
+}
+
+/// This is a minimal breaking example where instrumented code failed to compile.
+pub fn instrument_args_loop() {
+    let mut args: Vec<*mut libc::c_char> = Vec::new();
+    for arg in ::std::env::args() {}
 }
 
 pub fn main() {
--- a/dynamic_instrumentation/src/instrument.rs
+++ b/dynamic_instrumentation/src/instrument.rs
@@ -6,11 +6,13 @@ use fs2::FileExt;
 use fs_err::OpenOptions;
 use indexmap::IndexSet;
 use log::{debug, trace};
+use rustc_ast::Mutability;
 use rustc_index::vec::Idx;
-use rustc_middle::mir::visit::{MutatingUseContext, PlaceContext, Visitor};
+use rustc_middle::mir::visit::{MutVisitor, MutatingUseContext, PlaceContext, Visitor};
 use rustc_middle::mir::{
-    BasicBlock, BasicBlockData, Body, BorrowKind, HasLocalDecls, Local, LocalDecl, Location,
-    Operand, Place, PlaceElem, Rvalue, SourceInfo, Terminator, TerminatorKind, START_BLOCK,
+    BasicBlock, BasicBlockData, Body, BorrowKind, ClearCrossCrate, HasLocalDecls, Local, LocalDecl,
+    Location, Operand, Place, PlaceElem, ProjectionElem, Rvalue, Safety, SourceInfo, SourceScope,
+    SourceScopeData, Statement, StatementKind, Terminator, TerminatorKind, START_BLOCK,
 };
 use rustc_middle::ty::{self, Ty, TyCtxt};
 use rustc_span::def_id::{DefId, DefPathHash};
@@ -23,9 +25,11 @@ use std::sync::Mutex;
 use crate::arg::{ArgKind, InstrumentationArg};
 use crate::hooks::Hooks;
 use crate::mir_utils::{has_outer_deref, remove_outer_deref, strip_all_deref};
-use crate::point::CollectInstrumentationPoints;
 use crate::point::InstrumentationApplier;
 use crate::point::{cast_ptr_to_usize, InstrumentationPriority};
+use crate::point::{
+    CollectAddressTakenLocals, CollectInstrumentationPoints, RewriteAddressTakenLocals,
+};
 use crate::util::Convert;
 
 #[derive(Default)]
@@ -117,6 +121,224 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
     ty.is_unsafe_ptr() || ty.is_region_ptr()
 }
 
+impl<'tcx> Visitor<'tcx> for CollectAddressTakenLocals<'_, 'tcx> {
+    /// Checks the right hand side of each MIR assignment statement for taking the
+    /// address of a local, which occurs only in [`Rvalue`]s. This may
+    /// happen explicitly with something like [`std::ptr::addr_of`]`!` or
+    /// implicitly by taking a reference `&` of a local. Instances/locations
+    /// of address-taking are saved for a later walk over the MIR AST with
+    /// [`RewriteAddressTakenLocals`].
+    fn visit_assign(&mut self, place: &Place<'tcx>, rvalue: &Rvalue<'tcx>, location: Location) {
+        self.super_assign(place, rvalue, location);
+        let p = match rvalue {
+            Rvalue::AddressOf(_, p) | Rvalue::Ref(_, _, p) => p,
+            _ => return,
+        };
+        let value_ty = rvalue.ty(self, self.tcx());
+        if is_region_or_unsafe_ptr(value_ty) && p.projection.is_empty() {
+            self.address_taken.insert(p.local);
+        }
+    }
+}
+
+impl<'tcx> MutVisitor<'tcx> for RewriteAddressTakenLocals<'tcx> {
+    /// Rewrites an address-taken local in terms of its underlying address.
+    fn visit_place(
+        &mut self,
+        mut place: &mut Place<'tcx>,
+        context: PlaceContext,
+        location: Location,
+    ) {
+        // If we have found an address-taken local `_x`, substitute with `(*_y)` where `_y`
+        // is the address of `_x`.
+        if let Some(substitute) = self.local_to_address.get(&place.local) {
+            // We only want to rewrite address-taken locals that are not assigned to in
+            // a MIR assignment statement, unless the assignment is to one of that local's
+            // projections, such as one of its fields. The reason for this is that the local
+            // needs to be initialized at least once to make the compiler happy and not throw
+            // an error claiming use without initialization (such as when taking its address)
+            let is_assignment_to_local_projection = context.is_place_assignment()
+                && !place.is_indirect()
+                && !place.projection.is_empty();
+            let is_non_assignment_use = !context.is_place_assignment();
+
+            if context.is_use()
+                && (is_non_assignment_use || is_assignment_to_local_projection)
+                // maintain drop semantics for original address-taken local -- the liveness
+                // properties of its address are not necessarily the same, and dropping
+                // `(*_y)` is undesireable
+                && !context.is_drop()
+            {
+                // add deref
+                let projection = {
+                    let v = [ProjectionElem::Deref]
+                        .into_iter()
+                        .chain(place.projection)
+                        .collect::<Vec<_>>();
+                    self.tcx().intern_place_elems(&v)
+                };
+                // replace `_x` with `_y`
+                place.local = *substitute;
+                place.projection = projection;
+            }
+        }
+
+        self.super_place(place, context, location)
+    }
+
+    fn tcx<'a>(&'a self) -> TyCtxt<'tcx> {
+        self.tcx()
+    }
+
+    fn visit_body(&mut self, body: &mut Body<'tcx>) {
+        // for each address-taken local, push a new local that will get assigned
+        // its address
+        for local in self.address_taken.iter().copied() {
+            let arg_ty = Place::from(local).ty(&body.local_decls, self.tcx()).ty;
+            let inner_ty = ty::TypeAndMut {
+                ty: arg_ty,
+                mutbl: Mutability::Mut,
+            };
+
+            let raw_ptr_ty = self.tcx().mk_ptr(inner_ty);
+            let raw_ptr_local = body.local_decls.push(LocalDecl::new(raw_ptr_ty, DUMMY_SP));
+
+            self.local_to_address.insert(local, raw_ptr_local);
+
+            // set original local as mutable, because it's mutably borrowed when it has
+            // its address taken
+            body.local_decls.get_mut(local).unwrap().mutability = Mutability::Mut;
+        }
+
+        // START ADDRESS-TAKING STATEMENTS
+        //
+        // The code below appends context-dependent MIR locations to a collection
+        // that will later be used as reference for where to insert the first instance
+        // of taking the address of an already-determined-to-be address-taken local.
+        // E.g. if `_x` is an address-taken local, below is what determines where the
+        // statement `_y = &raw _x` will be placed.
+
+        let mut local_address_statements = Vec::new();
+
+        // assign insertion location of address-taken arguments, skipping over
+        // return local 0
+        for arg_idx in 1..=body.arg_count {
+            let arg_local = Local::from_usize(arg_idx);
+            if self.local_to_address.get(&arg_local).is_some() {
+                // put into first block, first statement
+                local_address_statements.push((Place::from(arg_local), Location::START));
+            }
+        }
+
+        // when the address-taken local is assigned to for the first time, we know it's active,
+        // so insert `_y = &raw x` just below that assignment, which is neccessary because
+        // otherwise the address-taking statement would be taking the address of an uninitialized
+        // variable. For assignment statements, place `_y = &raw _x` one statement below. For
+        // terminators with a destination to the address-taken local, or drop and replace
+        // statements, put the address-taking statement in the first statement of the successor
+        // block
+        for (bid, block) in body.basic_blocks().iter_enumerated().rev() {
+            for (sid, statement) in block.statements.iter().enumerate().rev() {
+                if let StatementKind::Assign(stmt) = &statement.kind {
+                    let (ref place, _) = **stmt;
+                    if place
+                        .as_local()
+                        .filter(|local| self.address_taken.contains(local))
+                        .is_some()
+                    {
+                        // put just below first assignment
+                        local_address_statements.push((
+                            *place,
+                            Location {
+                                block: bid,
+                                statement_index: sid + 1,
+                            },
+                        ));
+                    }
+                }
+            }
+            if let Some(term) = &block.terminator {
+                match &term.kind {
+                    TerminatorKind::Call {
+                        func: _,
+                        args: _,
+                        destination,
+                        target,
+                        ..
+                    } => {
+                        if destination
+                            .as_local()
+                            .filter(|l| self.address_taken.contains(l))
+                            .is_some()
+                        {
+                            if let Some(next_block) = target {
+                                // put into first statement of following block
+                                local_address_statements.push((
+                                    *destination,
+                                    Location {
+                                        block: *next_block,
+                                        statement_index: 0,
+                                    },
+                                ));
+                            }
+                        }
+                    }
+                    TerminatorKind::DropAndReplace {
+                        place,
+                        value,
+                        target,
+                        unwind: _,
+                    } if value.place().is_some() => {
+                        if let Some(local_to_address) =
+                            place.as_local().and_then(|l| self.local_to_address.get(&l))
+                        {
+                            // put into first statement of following block
+                            local_address_statements.push((
+                                *place,
+                                Location {
+                                    block: *target,
+                                    statement_index: 0,
+                                },
+                            ));
+                            self.local_to_address.insert(place.local, *local_to_address);
+                        }
+                    }
+                    _ => (),
+                }
+            }
+        }
+
+        // END ADDRESS-TAKING STATEMENTS
+        // visit places first before inserting `_y = &raw _x`, so that we don't accidentally
+        // rewrite `_y = &raw _x` into `y = &raw (*_y)`
+        self.super_body(body);
+
+        // sort the insertion locations by statement id descending so that statement insertion
+        // for statements with indices N and M, where N < M, does not shift/invalidate index M
+        local_address_statements.sort_by_key(|(_, loc)| loc.statement_index);
+        let local_address_statements = local_address_statements.into_iter().rev();
+
+        // Add `_y = &raw _x` for each address-taken local _x
+        for (addr_taken_local, loc) in local_address_statements {
+            let addr_of_stmt = Statement {
+                source_info: SourceInfo::outermost(DUMMY_SP),
+                kind: StatementKind::Assign(Box::new((
+                    self.local_to_address
+                        .get(&addr_taken_local.local)
+                        .copied()
+                        .unwrap()
+                        .into(),
+                    Rvalue::AddressOf(Mutability::Mut, addr_taken_local),
+                ))),
+            };
+
+            body.basic_blocks_mut()[loc.block]
+                .statements
+                .insert(loc.statement_index, addr_of_stmt);
+        }
+    }
+}
+
 impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
     fn visit_place(&mut self, place: &Place<'tcx>, context: PlaceContext, location: Location) {
         self.super_place(place, context, location);
@@ -156,6 +378,7 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
         let load_value_fn = self.hooks().find("load_value");
         let store_value_fn = self.hooks().find("store_value");
         let store_fn = self.hooks().find("ptr_store");
+        let store_addr_taken_fn = self.hooks().find("ptr_store_addr_taken");
         let load_fn = self.hooks().find("ptr_load");
 
         let dest = *dest;
@@ -176,6 +399,33 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
             PlaceContext::MutatingUse(MutatingUseContext::Store),
             location,
         );
+
+        // Instrument assignment to a local that is address-taken, even
+        // if its address was taken after this assignment statement
+        match self.addr_taken_local_addresses.get(&dest.local) {
+            Some(address) if dest.projection.is_empty() => {
+                let addr_of_local = Place::from(*address);
+                // TODO: this is a hack that places the store_addr_taken_fn
+                // after the instrumentation for taking the address of that local,
+                // which must be in place prior to this instrumentation.
+                let num_statements = self.body.basic_blocks()[location.block].statements.len();
+                let store_addr_taken_loc = Location {
+                    block: location.block,
+                    // +1 to ensure `dest` is in scope
+                    // +1 to be placed after address-taking statement
+                    statement_index: std::cmp::min(num_statements, location.statement_index + 2),
+                };
+                self.loc(
+                    location,
+                    store_addr_taken_loc, // to be placed after address-of-local instrumentation
+                    store_addr_taken_fn,
+                )
+                .arg_addr_of(dest)
+                .source(&addr_of_local)
+                .add_to(self);
+            }
+            _ => (),
+        }
 
         let mut add_load_instr = |p: &Place<'tcx>| {
             self.loc(location, location, load_fn)
@@ -281,7 +531,10 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
                     .dest(&dest)
                     .add_to(self);
             }
-            Rvalue::Ref(_, bkind, p) if has_outer_deref(p) => {
+            Rvalue::Ref(_, bkind, p)
+                if has_outer_deref(p)
+                    && is_region_or_unsafe_ptr(place_ty(&remove_outer_deref(*p, self.tcx()))) =>
+            {
                 // this is a reborrow or field reference, i.e. _2 = &(*_1)
                 let source = remove_outer_deref(*p, self.tcx());
                 if let BorrowKind::Mut { .. } = bkind {
@@ -290,7 +543,6 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
                         .arg_addr_of(*p)
                         .source(&source)
                         .dest(&dest)
-                        .instrumentation_priority(InstrumentationPriority::Early)
                         .add_to(self);
                 } else {
                     // Instrument immutable borrows by tracing the reference itself
@@ -298,12 +550,10 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
                         .arg_var(dest)
                         .source(&source)
                         .dest(&dest)
-                        .instrumentation_priority(InstrumentationPriority::Early)
                         .add_to(self);
                 };
             }
             Rvalue::Ref(_, bkind, p) if !p.is_indirect() => {
-                // this is a reborrow or field reference, i.e. _2 = &(*_1)
                 let source = remove_outer_deref(*p, self.tcx());
                 if let BorrowKind::Mut { .. } = bkind {
                     // Instrument which local's address is taken
@@ -312,6 +562,7 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
                         .arg_index_of(p.local)
                         .source(&source)
                         .dest(&dest)
+                        .instrumentation_priority(InstrumentationPriority::Early)
                         .add_to(self);
                 } else {
                     // Instrument immutable borrows by tracing the reference itself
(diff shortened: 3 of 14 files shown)


spernsteiner:
I checked most of the larger graphs in the snapshot diff and they all look good to me.

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot_debug.snap
```diff
@@ -3,36 +3,36 @@ source: pdg/src/main.rs
...
+	n[0]: copy        _    => _14    @ bb6[4]:  fn main;                 _14 = null_mut();
+	n[1]: copy        n[0] => _1     @ bb0[0]:  fn once;                 _13 = once(move _14);
+	n[2]: value.store _    => _20.*  @ bb4[7]:  fn invalid;              (*_20) = const 0_usize as *mut pointers::S (PointerFromExposedAddress);
+	n[3]: value.store _    => _17.*  @ bb8[4]:  fn fdevent_unregister;   (*_17) = const 0_usize as *mut pointers::fdnode_st (PointerFromExposedAddress);
+	n[4]: int_to_ptr  _    => _2     @ bb0[2]:  fn test_ref_field;       _2 = const 0_usize as *const pointers::S (PointerFromExposedAddress);
+	n[5]: int_to_ptr  _    => _5     @ bb0[8]:  fn test_ref_field;       _5 = const 0_usize as *const pointers::S (PointerFromExposedAddress);
+	n[6]: int_to_ptr  _    => _46    @ bb34[3]: fn main_0;               _46 = const 0_usize as *const pointers::S (PointerFromExposedAddress);
+	n[7]: value.store _    => _3.*.2 @ bb0[1]:  fn test_addr_taken_arg;  ((*_3).2: *const pointers::S) = const 0_usize as *const pointers::S (PointerFromExposedAddress);
```
spernsteiner, line 27: I guess this is the graph for the null pointer?  Seems kind of weird to have such a thing - maybe it should be filtered out.  But any such changes can be left to a separate PR (and we should probably think more about `IntToPtr` as part of that).

spernsteiner (approved):
`test_addr_taken` and `test_addr_taken_arg` snapshot results look good to me

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot_debug.snap
```diff
@@ -821,112 +835,134 @@ nodes_that_need_write = [3, 2, 1, 0]
...
+	n[3]: addr.store n[1] => _  @ bb2[8]:  fn test_load_value_store_value;  (*_4) = move _5;
+	n[4]: addr.load  n[1] => _  @ bb2[12]: fn test_load_value_store_value;  _7 = (*_4);
+}
+nodes_that_need_write = [3, 1, 0]
+
+g {
+	n[0]:  &_35       _     => _   @ bb30[4]:  fn main_0;          _49 = &raw mut _35;
+	n[1]:  copy       n[0]  => _34 @ bb30[5]:  fn main_0;          _34 = &mut (*_49);
+	n[2]:  addr.store n[0]  => _   @ bb30[3]:  fn main_0;          _35 = [const 2_i32, const 5_i32, const 3_i32, const 1_i32, const 6_i32];
+	n[3]:  copy       n[1]  => _40 @ bb30[12]: fn main_0;          _40 = &(*_34);
```
spernsteiner, line 858: The ordering here seems off - I think nodes 1 and 2 are swapped.  Usually for initialization of an address-taken local, you'll see something like `&_1 @ bb0[2]; addr.store @ bb0[1]`, where the store that initializes the local comes first in the MIR but second in the PDG.  Here, there's a third node in between (`n[1]`).  Also, this ordering introduces a spurious conflict: the pointer produced by `n[1]` is live across the store in `n[2]` (note there are things below with `source: n[1]`).

spernsteiner:
Have you tested the new `AddrOfLocal` handling on code with loops or other interesting control flow?  The current strategy involves special handling for the first initialization of a variable, and that makes me a bit nervous.  Some things I'd particularly like to see:

* A variable declared and initialized inside a loop: `loop { let x = ...; ... &x ... }`
* A variable initialized with a conditional: `let x = if cond { ... } else { ... }; ... &x ...`
* A variable that's only conditionally initialized: `let mut x;  if cond { x = 1; ... &x ... }; x = 2; ... &x ...`

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot_debug.snap
aneksteind: You're right, fixed in https://github.com/immunant/c2rust/pull/654/commits/b86ccc7e9154870bf0c0abf0dc681407e10d96f8 which includes the fix to the snapshot.

kkysen:
What does this PR do? Could you update PR title and add a brief description?

aneksteind:
> Have you tested the new AddrOfLocal handling on code with loops or other interesting control flow?

No, I will add more snapshot tests for those cases and update you when they are ready.

aneksteind:
> What does this PR do? Could you update PR title and add a brief description?

Will do soon, will definitely have it by the time I mark it ready for review for you and others.

kkysen:
> Will do soon, will definitely have it by the time I mark it ready for review for you and others.

Okay, great.  I wasn't sure since @spernsteiner already approved.

aneksteind:
> Okay, great. I wasn't sure since @spernsteiner already approved.

Yeah, I won't be merging until I have approval for the PR generally. That was just for the snapshot.

aneksteind:
> Have you tested the new AddrOfLocal handling on code with loops or other interesting control flow? The current strategy involves special handling for the first initialization of a variable, and that makes me a bit nervous. Some things I'd particularly like to see...

@spernsteiner would you mind taking a look at the [test cases](https://github.com/immunant/c2rust/commit/b01d886ef72e7a3bd987f5a680b805db3584c63e#diff-1988564ffd77b2441b1bad4177de849772d566f32b6affb4baf467fd822137d3) and [snapshot](https://github.com/immunant/c2rust/commit/b01d886ef72e7a3bd987f5a680b805db3584c63e#diff-2c6c58999046b7f42ba24e5494cd3f8b90744ba025a061308a6455912c90b5ba) added in https://github.com/immunant/c2rust/pull/654/commits/b01d886ef72e7a3bd987f5a680b805db3584c63e?

pdg/src/snapshots/c2rust_pdg__tests__analysis_test_pdg_snapshot_debug.snap
```diff
@@ -963,6 +959,47 @@ g {
...
+	n[10]: copy       n[8] => _4  @ bb5[0]:  fn test_addr_taken_loop;  _4 = &(*_20);
+}
+nodes_that_need_write = [9, 8, 5, 4, 1, 0]
+
+g {
+	n[0]: &_2        _    => _  @ bb1[1]: fn test_addr_taken_cond;  _6 = &raw mut _2;
+	n[1]: addr.store n[0] => _  @ bb1[0]: fn test_addr_taken_cond;  _2 = const 1_i32;
+	n[2]: copy       n[0] => _4 @ bb4[3]: fn test_addr_taken_cond;  _4 = &(*_6);
+	n[3]: addr.load  n[0] => _  @ bb4[6]: fn test_addr_taken_cond;  _5 = (*_6);
+	n[4]: copy       n[0] => _6 @ bb3[1]: fn test_addr_taken_cond;  _6 = &raw mut _2;
```
spernsteiner, line 982: It seems a bit odd to claim that `_6` in the second call is initialized by copying from `_6` in the first call, given none of the first call's locals exist any more.  Though I'm not sure if this will actually cause us any problems.

However, when I tried adding a test case to break this, I got some really odd results.  Here's the function I added to `pointers.rs`:

```Rust
#[no_mangle]
pub fn test_addr_taken_rec(n: usize) {
    if n == 0 {
        return;
    }

    let mut x = 1;
    let y = &mut x;
    test_addr_taken_rec(n - 1);
    *y = 2;
}
```

I got one graph that looks like this, which is fine:

```
g {
	n[0]: &_6        _    => _  @ bb6[4]: fn test_addr_taken_rec;  _11 = &raw mut _6;
	n[1]: addr.store n[0] => _  @ bb6[3]: fn test_addr_taken_rec;  _6 = const 1_i32;
	n[2]: copy       n[0] => _7 @ bb6[7]: fn test_addr_taken_rec;  _7 = &mut (*_11);
	n[3]: addr.store n[2] => _  @ bb7[2]: fn test_addr_taken_rec;  (*_7) = const 2_i32;
}
```

But I also got another graph that looks like this:

```
g {
	n[0]:  &_1         _    => _    @ bb0[2]:  fn test_unique_ref;      _8 = &raw mut _1;
	n[1]:  addr.store  n[0] => _    @ bb0[1]:  fn test_unique_ref;      _1 = const 10_i32;
	n[2]:  copy        n[0] => _4   @ bb0[9]:  fn test_unique_ref;      _4 = &mut (*_8);
	n[3]:  copy        n[2] => _3   @ bb0[10]: fn test_unique_ref;      _3 = &raw mut (*_4);
	n[4]:  copy        n[0] => _7   @ bb0[18]: fn test_unique_ref;      _7 = &mut (*_8);
	n[5]:  copy        n[4] => _6   @ bb0[19]: fn test_unique_ref;      _6 = &raw mut (*_7);
	n[6]:  value.store n[5] => _5.* @ bb0[20]: fn test_unique_ref;      (*_5) = move _6;
	n[7]:  copy        n[5] => _11  @ bb6[4]:  fn test_addr_taken_rec;  _11 = &raw mut _6;
	n[8]:  addr.store  n[7] => _    @ bb6[3]:  fn test_addr_taken_rec;  _6 = const 1_i32;
	n[9]:  copy        n[7] => _7   @ bb6[7]:  fn test_addr_taken_rec;  _7 = &mut (*_11);
	n[10]: addr.store  n[9] => _    @ bb7[2]:  fn test_addr_taken_rec;  (*_7) = const 2_i32;
}
```

This is clearly bogus - no pointers flow from `test_unique_ref` to `test_addr_taken_rec` or vice versa.  My guess is that `test_addr_taken_local`'s `_6` (in the recursive call) happened to land at the same address on the stack as `test_unique_ref`'s `_1`, and the latter was still considered live (wrongly), so the initialization of `_11 = &raw mut _6` was treated as a copy of `test_unique_ref`'s `_1`.

I think the instrumentation and/or PDG builder needs to mark addr-taken locals as dead at some point (the same way heap allocations are marked as dead when passed to `free`, so no more nodes can be added to their graphs).  One option is to do this when returning from the function, but that might be a pain.  Another option might be to instrument `StorageLive` for addr-taken locals to mark all objects that overlap the local as dead.

aneksteind: I'm aware this is a bug and am working on it

analysis/runtime/src/events.rs
```diff
@@ -70,6 +76,11 @@ pub enum EventKind {
 
     Offset(Pointer, isize, Pointer),
 
+    /// Marks the start of events in a new function body.
+    /// Used to distinguish address-taken locals that treated
+    /// as copies and ones that aren't.
+    BeginBody,
```
kkysen, line 82: Would `BeginFunc` make more sense?  `Body` is not as clear without `Func`, and with `Func`, `Body` is kind of redundant.

analysis/runtime/src/events.rs
```diff
@@ -90,8 +101,10 @@ impl Debug for EventKind {
             } => write!(f, "realloc(0x{:x}, {}) -> 0x{:x}", old_ptr, size, new_ptr),
             Ret(ptr) => write!(f, "ret(0x{:x})", ptr),
             Done => write!(f, "done"),
+            BeginBody => write!(f, "begin body"),
             LoadAddr(ptr) => write!(f, "load(0x{:x})", ptr),
             StoreAddr(ptr) => write!(f, "store(0x{:x})", ptr),
+            StoreAddrTaken(ptr) => write!(f, "store(0x{:x})", ptr),
```
kkysen, line 107: Should these print the same?  I would think we'd want to distinguish them.

analysis/test/src/pointers.rs
```diff
@@ -540,6 +589,73 @@ pub unsafe extern "C" fn insertion_sort(n: libc::c_int, p: *mut libc::c_int) {
...
+    let mut prefix: [libc::c_char; 40] = [0; 40];
+    vsprintf(prefix.as_mut_ptr(), fmt, ap.as_va_list());
+}
+pub unsafe extern "C" fn log_error(
+    errh: *mut libc::c_void,
+    filename: *const libc::c_char,
+    line: libc::c_uint,
+    mut fmt: *const libc::c_char,
+    mut args: ...
+) {
```
kkysen, line 616: Is a variadic function necessary?  `miri` can't run variadic functions, so if we don't need to test variadic functions, we shouldn't add them.  I don't want UB creeping back into this file.

If we do want to test variadics themselves, we could do that separately (where we're more sure there's no UB by human inspection), and have a `cfg!(feature = "miri")` around calling it.  If necessary for this test, we could also use that strategy here.

pdg/src/builder.rs
```diff
@@ -49,23 +61,15 @@ impl EventKindExt for EventKind {
...
-
-    fn parent(&self, obj: (GraphId, NodeId)) -> Option<(GraphId, NodeId)> {
-        self.has_parent().then_some(obj)
-    }
-
-    fn to_node_kind(&self) -> Option<NodeKind> {
+    fn to_node_kind(
+        &self,
+        func: Func,
+        address_taken: &mut IndexSet<(Func, u32)>,
```
kkysen, line 71: Can the `u32` just be `Local`?  If not, we should at least use a type alias for clarity.

kkysen:
Is there a way to split this PR up more?  It's getting quite big.  The formatting changes in `pointers.rs` can at least be done directly to `master`; I've been meaning to do that.

analysis/test/src/pointers.rs
aneksteind: this doesn't run / get called anyway, and yes it was necessary to reproduce the instrumentation error I encountered

aneksteind:
> Is there a way to split this PR up more? It's getting quite big. The formatting changes in pointers.rs can at least be done directly to master; I've been meaning to do that.

@kkysen I'd rather not, the parts that could be split out are pretty small in comparison to the one big feature. If that were broken up, it would mean merging incomplete functionality / incorrect PDG objects, and so I think it makes sense to land all at once.

analysis/test/src/pointers.rs
kkysen: Okay, if it's not actually run, then it's fine with `miri`.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+        // START ADDRESS-TAKING STATEMENTS
+        //
+        // The code below appends context-dependent MIR locations to a collection
+        // that will later be used as reference for where to insert the first instance
+        // of taking the address of an already-determined-to-be address-taken local.
+        // E.g. if `_x` is an address-taken local, below is what determines where the
+        // statement `_y = &raw _x` will be placed.
+        // assign insertion location of address-taken arguments
+        for l in 1..body.arg_count + 1 {
+            let local = Local::from_u32(l as u32);
```
kkysen, line 226:
```suggestion
            let local = Local::from_usize(l);
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+            if self.local_substitute.get(&local).is_some() {
+                // put into first block, first statement
+                local_address_statements.push((Place::from(local), BasicBlock::from_u32(0), 0));
```
kkysen, line 229:
```suggestion
                local_address_statements.push((Place::from(local), START_BLOCK, 0));
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+
+        self.super_place(place, context, location)
+    }
+
+    fn tcx<'a>(&'a self) -> TyCtxt<'tcx> {
+        self.tcx()
+    }
+
+    fn visit_body(&mut self, body: &mut Body<'tcx>) {
+        let mut local_address_statements = Vec::new();
```
kkysen, line 197: Move this to near the `START ADDRESS-TAKING STATEMENTS` comment where it's used.

kkysen, line 197: I think it would be helpful to use a `struct` with named fields instead of the tuple `(Place, BasicBlock, usize)`, which is confusing, especially for the `usize`, which is a statement index/id from what I can tell.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+        // terminators with a destination to the address-taken local, or drop and replace
+        // statements, put the address-taking statement in the first statement of the successor
+        // block
+        for (bid, block) in body.basic_blocks().iter_enumerated().rev() {
+            for (sid, statement) in block.statements.iter().enumerate().rev() {
+                if let StatementKind::Assign(stmt) = &statement.kind {
+                    let (ref place, _) = **stmt;
+                    if place
+                        .as_local()
+                        .filter(|l| self.address_taken.contains(l))
```
kkysen, line 246:
```suggestion
                        .filter(|local| self.address_taken.contains(local))
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+                        // put just below first assignment
+                        local_address_statements.push((*place, bid, sid + 1));
+                    }
+                }
+            }
+            if let Some(term) = &block.terminator {
+                match &term.kind {
+                    TerminatorKind::Call {
+                        func: _func,
+                        args: _args,
```
kkysen, line 258:
```suggestion
                        func: _,
                        args: _,
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+        let local_address_statements = local_address_statements.into_iter().rev();
+
+        // Add `_y = &raw _x` for each address-taken local _x
+        for (addr_taken_local, bid, sid) in local_address_statements {
+            let addr_of_stmt = Statement {
+                source_info: SourceInfo::outermost(DUMMY_SP),
+                kind: StatementKind::Assign(Box::new((
+                    self.local_substitute
+                        .get(&addr_taken_local.local)
+                        .cloned()
```
kkysen, line 310:
```suggestion
                        .copied()
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -323,13 +553,27 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+        // adds an instrumentation to mark the start of the body
+        self.super_body(body);
+        if self.instrumentation_points.len() > 0 {
+            let body_begin_func = self.hooks().find("mark_begin_body");
+            let start_loc = Location {
+                block: BasicBlock::from_u32(0),
+                statement_index: 0,
+            };
+            self.loc(start_loc, start_loc, body_begin_func).add_to(self);
+        }
```
kkysen, line 574:
```suggestion
        
        if self.instrumentation_points.is_empty() {
            return;
        }
        let body_begin_func = self.hooks().find("mark_begin_body");
        let start_loc = Location::START;
        self.loc(start_loc, start_loc, body_begin_func).add_to(self);
```

`clippy::len_zero` warning.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -323,13 +553,27 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+                        .instrumentation_priority(0)
                         .add_to(self);
                 };
             }
             _ => (),
         }
     }
 
+    fn visit_body(&mut self, body: &Body<'tcx>) {
+        // adds an instrumentation to mark the start of the body
```
kkysen, line 565:
```suggestion
    /// Adds an instrumentation to mark the start of the body.
    fn visit_body(&mut self, body: &Body<'tcx>) {
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
     ty.is_unsafe_ptr() || ty.is_region_ptr()
 }
 
+impl<'tcx> Visitor<'tcx> for CheckAddressTakenLocals<'_, 'tcx> {
+    /// Checks the right hand side of each assignment statement for taking the
+    /// address of a local. This may happen explicitly with something like
+    /// std::ptr::addr_of! or implicitly by taking a reference & of a local.
+    /// Instances/locations of address-taking are saved for a later walk over the MIR
+    /// AST.
```
fw-immunant, line 137: It would be good to note here that address-taking occurs only in `Rvalue`s, not in `Operand`s or `Place`s, so that looking at all MIR assignments covers all places where an address might be taken (including the surface-language constructs you call out by name). Maybe change "assignment statement" to "MIR assignment statement" to disambiguate that this isn't referring to surface-language assignment statements.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+            };
+
+            let raw_ptr_ty = self.tcx().mk_ptr(inner_ty);
+            let raw_ptr_local = body.local_decls.push(LocalDecl::new(raw_ptr_ty, DUMMY_SP));
+
+            self.local_substitute.insert(*local, raw_ptr_local);
+
+            // set original local as mutable
+            body.local_decls.get_mut(*local).unwrap().mutability = Mutability::Mut;
+        }
```
kkysen, line 215:
```suggestion
        for local in self.address_taken.iter().copied() {
            let arg_ty = Place::from(local).ty(&body.local_decls, self.tcx()).ty;
            let inner_ty = ty::TypeAndMut {
                ty: arg_ty,
                mutbl: Mutability::Mut,
            };

            let raw_ptr_ty = self.tcx().mk_ptr(inner_ty);
            let raw_ptr_local = body.local_decls.push(LocalDecl::new(raw_ptr_ty, DUMMY_SP));

            self.local_substitute.insert(local, raw_ptr_local);

            // set original local as mutable
            body.local_decls.get_mut(local).unwrap().mutability = Mutability::Mut;
        }
```

kkysen (changes requested):

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+                ty: arg_ty,
+                mutbl: Mutability::Mut,
+            };
+
+            let raw_ptr_ty = self.tcx().mk_ptr(inner_ty);
+            let raw_ptr_local = body.local_decls.push(LocalDecl::new(raw_ptr_ty, DUMMY_SP));
+
+            self.local_substitute.insert(*local, raw_ptr_local);
+
+            // set original local as mutable
```
fw-immunant, line 213: Why?

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+
+        // START ADDRESS-TAKING STATEMENTS
+        //
+        // The code below appends context-dependent MIR locations to a collection
+        // that will later be used as reference for where to insert the first instance
+        // of taking the address of an already-determined-to-be address-taken local.
+        // E.g. if `_x` is an address-taken local, below is what determines where the
+        // statement `_y = &raw _x` will be placed.
+        // assign insertion location of address-taken arguments
+        for l in 1..body.arg_count + 1 {
```
fw-immunant, line 225:
```suggestion
        for l in 1..=body.arg_count {
```

dynamic_instrumentation/src/instrument.rs
kkysen: See #705.

analysis/test/src/pointers.rs
```diff
@@ -597,6 +642,72 @@ pub unsafe extern "C" fn insertion_sort(n: libc::c_int, p: *mut libc::c_int) {
...
+    let mut ap: ::std::ffi::VaListImpl;
+    ap = args.clone();
+    log_error_va_list_impl(errh, filename, line, fmt, ap.as_va_list(), 0 as libc::c_int);
+}
+
+/*
+    This is a minimal breaking example from lighttpd where instrumented code failed
+    to compile.
+*/
```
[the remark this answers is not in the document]
kkysen: See #705.

analysis/test/src/pointers.rs
```diff
@@ -597,6 +642,72 @@ pub unsafe extern "C" fn insertion_sort(n: libc::c_int, p: *mut libc::c_int) {
     }
 }
 
+/*
+    This is a minimal breaking example from lighttpd where instrumented code failed
+    to compile.
+*/
```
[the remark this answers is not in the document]
kkysen: See #705.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -426,9 +670,32 @@ fn instrument_body<'a, 'tcx>(
...
+    local_rewriter.visit_body(body);
+
+    // The local rewriter above rewrites address-taken locals with (*_x) where _x
+    // is a pointer, resulting in an unsafe operation. To allow this, set all scopes
+    // as unsafe
+    for scope in &mut body.source_scopes {
+        if let ClearCrossCrate::Set(data) = &mut scope.local_data {
+            data.safety = Safety::BuiltinUnsafe;
+        }
+    }
```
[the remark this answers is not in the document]
fw-immunant: AIUI, setting everything as unsafe just lets the code get accepted by later validation passes that try to make sure safe contexts don't contain unsafe operations.

pdg/src/main.rs
```diff
@@ -79,6 +80,23 @@ impl Pdg {
...
     }
+
+    pub fn remove_sources(&mut self) {
+        /*
+            The destination node of AddrOfLocal is always Some(local)
+            and is used in determining the sources of subsequent PDG
+            nodes. However, for the purposes of static analysis, it's
+            undesired and therefore destinations of AddrOfLocalNodes
+            are removed.
+        */
```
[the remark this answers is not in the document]
kkysen: See #705 for just moving the comment to a doc comment.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -123,6 +129,198 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+
+        // END ADDRESS-TAKING STATEMENTS
+        // visit places first before inserting `_y = &raw _x`, so that we don't accidentally
+        // rewrite `_y = &raw _x` into `y = &raw (*_y)`
+        self.super_body(body);
+
+        // sort the insertion locations by statement id descending so that statement insertion
+        // for statements with indices N and M, where N < M, does not shift/invalidate index M
+        local_address_statements.sort_by_key(|(_, _bid, sid)| *sid);
+        let local_address_statements = local_address_statements.into_iter().rev();
```
fw-immunant, line 301: It would be cleaner and faster to sort by `-*sid` here, or (if this tuple becomes a struct) define `PartialOrd` accordingly.

kkysen: What do you mean exactly?  By the faster part?  The sort needs to be done in place, right?

pdg/src/builder.rs
```diff
@@ -22,11 +23,21 @@ pub fn read_metadata(path: &Path) -> eyre::Result<Metadata> {
...
 pub trait EventKindExt {
     fn ptr(&self, metadata: &EventMetadata) -> Option<Pointer>;
-    fn has_parent(&self) -> bool;
-    fn parent(&self, obj: (GraphId, NodeId)) -> Option<(GraphId, NodeId)>;
-    fn to_node_kind(&self) -> Option<NodeKind>;
+    fn to_node_kind(
+        &self,
+        func: Func,
+        address_taken: &mut IndexSet<(Func, Local)>,
+    ) -> Option<NodeKind>;
```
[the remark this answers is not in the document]
kkysen: See #707, which adds the `FuncId`, though as a newtype, not just a type alias.

kkysen:
Also, after #637 landed, we have a `release` snapshot in addition to the existing `debug` one, so can you add the updates to that one here?

dynamic_instrumentation/src/instrument.rs
fw-immunant: Ah, yeah, never mind the faster bit--I forgot we were just reversing an iterator, because `local_address_statements` changes from a concrete collection into an iterator with this rebinding.

kkysen: Yeah, the `.rev()` should be zero-cost, and if `sid` is unsigned, `-sid` is a problem.  There's `Ordering::reverse`, but that would only work with `.sort_by()`, not `.sort_by_key()`.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -286,7 +513,10 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
                     .dest(&dest)
                     .add_to(self);
             }
-            Rvalue::Ref(_, bkind, p) if has_outer_deref(p) => {
+            Rvalue::Ref(_, bkind, p)
+                if has_outer_deref(p)
+                    && is_region_or_unsafe_ptr(place_ty(&remove_outer_deref(*p, self.tcx()))) =>
+            {
```
fw-immunant, line 519: I can't seem to find which commit added this change. `git log -p dynamic_instrumentation/src/instrument.rs` makes it look like it comes from `7675d10cdf0d3885ab5656d3882b111548ac0407`, but that commit isn't listed as part of this PR... I don't understand what's happening here--maybe merge weirdness?

The history here is really convoluted and I would love to see a simple sequence of commits without merges.

fw-immunant: Anyhow, the reason I wanted to know where this line came from is because I'd like to understand better exactly which bug is being fixed by this change, which should ideally be in the commit message of a small commit that does only this.

dynamic_instrumentation/src/instrument.rs
fw-immunant, line 565:
```suggestion
        // add an instrumentation point to mark the start of the body
```

dynamic_instrumentation/src/point/build.rs
```diff
@@ -140,6 +151,15 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
         self
     }
 
+    /// Sets the priority of this instrumentation. Lower value
+    /// signifies higher priority and implies that an instrumentation
+    /// with higher priority will be placed before one with lower
+    /// priority.
+    pub fn instrumentation_priority(mut self, p: usize) -> Self {
+        self.instrumentation_priority = p;
+        self
```
fw-immunant, line 160: It's a bit more obvious to write self-with-a-modification using record update syntax:
```suggestion
        Self { instrumentation_priority: p, ..self }
```

dynamic_instrumentation/src/point/cast.rs
```diff
@@ -76,7 +73,6 @@ pub fn cast_ptr_to_usize<'tcx>(
         // create a raw ptr with `addr_of!`.
         InstrumentationArg::AddrOf(arg) => {
             let arg_place = arg.place().expect("Can't get the address of a constant");
-            let arg_place = remove_outer_deref(arg_place, tcx);
```
fw-immunant, line 79: This appears to have been changed in a merge commit (c32d9b67fa9728b244c1d290266afee41959f9dd). Why?

dynamic_instrumentation/src/point/mod.rs
```diff
@@ -23,25 +24,87 @@ pub struct InstrumentationPoint<'tcx> {
...
     pub after_call: bool,
+    pub instrumentation_priority: usize,
     pub metadata: EventMetadata,
 }
 
+/// Gathers a set of all address-taken locals in a function
+/// body.
+///
+/// The set may include address-taken arguments.
+pub struct CheckAddressTakenLocals<'a, 'tcx: 'a> {
```
fw-immunant, line 35: How about `GatherAddressTakenLocals` or `CollectAddressTakenLocals`, in accord with the comment explaining what the visitor does.

dynamic_instrumentation/src/point/build.rs
kkysen: True, but it gets more complicated with something like https://github.com/immunant/c2rust/blob/a35136ec9d962e64f57980c2ef5ea3bbee0f40bc/dynamic_instrumentation/src/point/build.rs#L149-L152

it doesn't work as well, as this keeps the same style for all of these builder methods.

dynamic_instrumentation/src/point/mod.rs
kkysen: Is checking done?

dynamic_instrumentation/src/instrument.rs
kkysen: > The history here is really convoluted and I would love to see a simple sequence of commits without merges.

Yeah, that gets hard on these very large PRs that are developed for months.  I think any opportunity we see to split something out into a separate, more targeted PR, we should do.

dynamic_instrumentation/src/point/mod.rs
```diff
@@ -23,25 +24,87 @@ pub struct InstrumentationPoint<'tcx> {
...
+    }
+}
+
 pub struct CollectInstrumentationPoints<'a, 'tcx: 'a> {
     tcx: TyCtxt<'tcx>,
     hooks: Hooks<'tcx>,
-    body: &'a Body<'tcx>,
-    instrumentation_points: Vec<InstrumentationPoint<'tcx>>,
+    pub body: &'a Body<'tcx>,
+    pub instrumentation_points: Vec<InstrumentationPoint<'tcx>>,
```
fw-immunant, line 89: What foreign context now consumes these?

dynamic_instrumentation/src/instrument.rs
fw-immunant: I think rebasing is the only thing that makes sense for long-living PRs. Addressed comments going stale isn't a huge loss, and the ability for reviewers to figure out what comes from where when reviewing code is critical.

fw-immunant (changes requested):
Requested changes commented inline. These also came up inline, but as top-level comments:
- It would be better to use an enum of priorities than magic numbers.
- It's really hard to figure out which commits introduced individual changes in the presence of merge commits here. I would advocate a rebasing workflow over a merge-based one in the future.

spernsteiner:
Snapshot changes look good to me

analysis/runtime/src/events.rs
aneksteind: MIR refers to it as a Body, which is why it's not Func here.

pdg/src/builder.rs
```diff
@@ -148,11 +171,12 @@ pub fn add_node(
...
     }
 
-    let head = event
+    let provenance = event
         .kind
         .ptr(event_metadata)
         .and_then(|ptr| provenances.get(&ptr).cloned());
-    let ptr = head.and_then(|(gid, _last_nid_ref)| {
+
+    let direct_source = provenance.and_then(|(gid, _last_nid_ref)| {
```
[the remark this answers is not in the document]
kkysen: See #706.

analysis/runtime/src/events.rs
kkysen: Can we name it `BeginFuncBody` then?

pdg/src/builder.rs
```diff
@@ -75,9 +79,27 @@ impl EventKindExt for EventKind {
             Field(_, field) => NodeKind::Field(field.into()),
             LoadAddr(..) => NodeKind::LoadAddr,
             StoreAddr(..) => NodeKind::StoreAddr,
+            StoreAddrTaken(..) => NodeKind::StoreAddr,
```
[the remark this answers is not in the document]
aneksteind: i disagree, these are not meant to be treated the same in all cases and i don't want to imply they should be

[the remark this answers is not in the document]
kkysen: Okay, if that's the intent then, it's fine to leave it as is.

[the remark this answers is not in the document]
kkysen: Okay, if that's the intent then, it's fine to leave it as is.

aneksteind:
> Also, after https://github.com/immunant/c2rust/pull/637 landed, we have a release snapshot in addition to the existing debug one, so can you add the updates to that one here?

Done

dynamic_instrumentation/src/point/mod.rs
aneksteind: `CollectInstrumentationPoints::visit_assign` on line 389:

```rust
let num_statements = self.body.basic_blocks()[location.block].statements.len();
```

aneksteind: also line 564 in `visit_body`:

```rust
        if self.instrumentation_points.is_empty() {
            return;
        }
```

dynamic_instrumentation/src/point/cast.rs
aneksteind: it was a mistake to include it in that commit, that removal was fixing a bug

dynamic_instrumentation/src/point/mod.rs
fw-immunant: Ah, it's because we implement the visitor trait in terms of the struct's internals in a separate module from where we define the visitor struct itself.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -183,6 +382,33 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+                let addr_of_local = Place::from(*address);
+                // TODO: this is a hack that places the store_addr_taken_fn
+                // after the instrumentation for taking the address of that local,
+                // which must be in place prior to this instrumentation.
+                let num_statements = self.body.basic_blocks()[location.block].statements.len();
+                let store_addr_taken_loc = Location {
+                    block: location.block,
+                    // +1 to ensure `dest` is in scope
+                    // +1 to be placed after address-taking statement
+                    statement_index: std::cmp::min(num_statements, location.statement_index + 2),
```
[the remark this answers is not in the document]
kkysen: I just thought its cleaner and shows the intent to just update one field, `statement_index`.  It's not too big of an issue, though.

pdg/src/main.rs
```diff
@@ -387,6 +387,14 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn analysis_test_pdg_snapshot_release() -> eyre::Result<()> {
+        init();
+        let pdg = analysis_test_pdg_snapshot(Profile::Release)?;
+        insta::assert_display_snapshot!(pdg);
+        Ok(())
+    }
```
kkysen, line 404: Why's this in this commit?

dynamic_instrumentation/src/instrument.rs
aneksteind: used `Location` to help with that understanding

dynamic_instrumentation/src/instrument.rs
```diff
@@ -426,9 +670,32 @@ fn instrument_body<'a, 'tcx>(
...
+
+    // collect instrumentation points
+    let mut collector =
+        CollectInstrumentationPoints::new(tcx, hooks, body, local_rewriter.local_substitute);
     collector.visit_body(body);
     let points = collector.into_instrumentation_points();
+
+    // insert instrumentation
     let mut applier = InstrumentationApplier::new(state, tcx, body, body_did);
     applier.apply_points(&points);
```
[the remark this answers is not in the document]
aneksteind: I've done that for `points`, but am not really sure what that would look like with `applier`

dynamic_instrumentation/src/instrument.rs
aneksteind: the local's address is taken as `std::ptr::addr_of_mut`, and so is mutably borrowed. added a comment

analysis/runtime/src/events.rs
```diff
@@ -90,8 +102,10 @@ impl Debug for EventKind {
             } => write!(f, "realloc(0x{:x}, {}) -> 0x{:x}", old_ptr, size, new_ptr),
             Ret(ptr) => write!(f, "ret(0x{:x})", ptr),
             Done => write!(f, "done"),
+            BeginFuncBody => write!(f, "begin body"),
```
kkysen, line 105:
```suggestion
            BeginFuncBody => write!(f, "begin func body"),
```

analysis/runtime/src/events.rs
```diff
@@ -90,8 +102,10 @@ impl Debug for EventKind {
...
             LoadAddr(ptr) => write!(f, "load(0x{:x})", ptr),
             StoreAddr(ptr) => write!(f, "store(0x{:x})", ptr),
+            StoreAddrTaken(ptr) => write!(f, "store(0x{:x})", ptr),
```
kkysen, line 108: Are these meant to be printed the same?  I may have asked this before.

dynamic_instrumentation/src/instrument.rs
aneksteind: `0` is the return local so we want to ignore it, i'll take your suggestion

pdg/src/builder.rs
```diff
@@ -49,23 +61,15 @@ impl EventKindExt for EventKind {
...
-
-    fn parent(&self, obj: (GraphId, NodeId)) -> Option<(GraphId, NodeId)> {
-        self.has_parent().then_some(obj)
-    }
-
-    fn to_node_kind(&self) -> Option<NodeKind> {
+    fn to_node_kind(
+        &self,
+        func: FuncId,
+        address_taken: &mut IndexSet<(FuncId, Local)>,
```
kkysen, line 71: Can we add a `type AddressTaken = IndexSet<(FuncId, Local)>;`

dynamic_instrumentation/src/point/build.rs
aneksteind: opting to keep as-is

dynamic_instrumentation/src/instrument.rs
```diff
@@ -117,6 +121,239 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
     ty.is_unsafe_ptr() || ty.is_region_ptr()
 }
 
+impl<'tcx> Visitor<'tcx> for CollectAddressTakenLocals<'_, 'tcx> {
+    /// Checks the right hand side of each MIR assignment statement for taking the
+    /// address of a local, which occurs only in [`Rvalues`](`Rvalue`). This may
```
kkysen, line 126:
```suggestion
    /// address of a local, which occurs only in [`Rvalue`]s. This may
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -117,6 +121,239 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+        self.super_place(place, context, location)
+    }
+
+    fn tcx<'a>(&'a self) -> TyCtxt<'tcx> {
+        self.tcx()
+    }
+
+    fn visit_body(&mut self, body: &mut Body<'tcx>) {
+        // for each address-taken local, push a new local that will get assigned
+        // its address
```
kkysen, line 195: Should this be a doc comment?

kkysen: No maybe not, this function is pretty long.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -117,6 +121,239 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+        // for statements with indices N and M, where N < M, does not shift/invalidate index M
+        local_address_statements.sort_by_key(
+            |(
+                _,
+                Location {
+                    block: _bid,
+                    statement_index: sid,
+                },
+            )| *sid,
+        );
```
kkysen, line 326:
```suggestion
        local_address_statements.sort_by_key(|(_, loc)| *loc.statement_index);
```

dynamic_instrumentation/src/instrument.rs
```diff
@@ -117,6 +121,239 @@ fn is_region_or_unsafe_ptr(ty: Ty) -> bool {
...
+        let local_address_statements = local_address_statements.into_iter().rev();
+
+        // Add `_y = &raw _x` for each address-taken local _x
+        for (
+            addr_taken_local,
+            Location {
+                block: bid,
+                statement_index: sid,
+            },
```
kkysen, line 335: I'd just leave this as `loc` and use `loc.block` and `loc.statement_index`, as those are also clearer to understand than `bid` and `sid`.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -592,8 +592,8 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
         }
     }
 
+    /// Adds an instrumentation to mark the start of the body.
     fn visit_body(&mut self, body: &Body<'tcx>) {
-        // add an instrumentation point to mark the start of the body
```
kkysen: Should stay as `Add` not `Adds`.

dynamic_instrumentation/src/point/mod.rs
```diff
@@ -86,23 +86,23 @@ pub struct CollectInstrumentationPoints<'a, 'tcx: 'a> {
     pub body: &'a Body<'tcx>,
     pub instrumentation_points: Vec<InstrumentationPoint<'tcx>>,
     assignment: Option<(Place<'tcx>, Rvalue<'tcx>)>,
-    pub addr_taken_local_substitutions: IndexMap<Local, Local>,
+    pub local_to_address: IndexMap<Local, Local>,
```
kkysen, line 89: I think the `addr_taken` part of the name was useful here since this is in `CollectInstrumentationPoints`, which doesn't already mention address taken like the other ones.

dynamic_instrumentation/src/instrument.rs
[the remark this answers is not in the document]
kkysen: I guess it's not too important.  I was just thinking

```rust
{
    let mut applier = InstrumentationApplier::new(state, tcx, body, body_did);
    applier.apply_points(&points);
}
```

analysis/runtime/src/events.rs
kkysen: Yeah I already did.

aneksteind:
@fw-immunant @kkysen i've squashed things into one commit and will split it up from here, that seems easier than approaching it from the opposite direction

kkysen:
That seems like a reasonable idea.

dynamic_instrumentation/src/point/build.rs
```diff
@@ -31,6 +31,7 @@ impl<'tcx> CollectInstrumentationPoints<'_, 'tcx> {
         original_location: Location,
         instrumentation_location: Location,
         func: DefId,
+        instrumentation_priority: usize,
```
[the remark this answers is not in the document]
aneksteind: #710

pdg/src/main.rs
```diff
@@ -79,6 +80,23 @@ impl Pdg {
             to_print,
         }
     }
+
+    pub fn remove_sources(&mut self) {
```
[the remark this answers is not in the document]
aneksteind: I've just left it as a method in `Graphs` for now, but have at least moved the call to `Pdg::new`

fw-immunant (approved):
Comments have been addressed.

analysis/runtime/src/events.rs
aneksteind: We only want to distinguish them to treat of kinds of instrumentation points differently, this is treated as a store in the same way `StoreAddr` is

pdg/src/builder.rs
```diff
@@ -249,8 +264,15 @@ pub fn add_node(
 pub fn construct_pdg(events: &[Event], metadata: &Metadata) -> Graphs {
     let mut graphs = Graphs::new();
     let mut provenances = HashMap::new();
+    let mut address_taken = IndexSet::new();
```
kkysen, line 267:
```suggestion
    let mut address_taken = AddressTaken::new();
```

pdg/src/builder.rs
```diff
@@ -123,18 +138,18 @@ fn update_provenance(
 pub fn add_node(
     graphs: &mut Graphs,
     provenances: &mut HashMap<Pointer, (GraphId, NodeId)>,
+    address_taken: &mut IndexSet<(FuncId, Local)>,
```
kkysen, line 141:
```suggestion
    address_taken: &mut AddressTaken,
```

kkysen (approved):
There's just a couple remaining `AddressTaken` uses I think you missed that I marked, but I think it looks much better now.

PR #655, opened 2022-09-01
Initial function call support in static analysis

spernsteiner:
This branch adds basic interprocedural analysis for pointer permissions.  It propagates the "easy"/dataflow permissions, such as `WRITE` and `OFFSET_ADD`, across function calls and returns.  For example, given `fn f(x: *mut i32) { *x = 1; }`, it will correctly infer that a call `f(p)` means that `p` must have `WRITE` permission, even if there is no explicit write in the calling function.

Handling of the `UNIQUE` permission is much more complex and is left for a separate branch.  In short, at each function call, the borrow checker needs to know how the lifetimes of the various pointers in the arguments and return values are related so that it knows how the call affects borrows within the caller.  For example, given `fn g(x: *mut i32) -> *mut i32 { x }`, it needs to know that the returned pointer borrows from the argument so it can properly remove `UNIQUE` for code like `let y = g(x); let z = g(x); *y = 1;`

Support for struct fields is also left for later, though that should be straightforward - pointers in field types simply get global `PointerId`s like function arguments do and participate in the dataflow analysis the same way.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -119,6 +119,9 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         self.next_ptr_id.num_pointers()
     }
 
+    /// Update all `PointerId`s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
+    /// `PointerId`" counter to `counter`.  `map` and `counter` are usually computed together via
+    /// `GlobalEquivSet::renumber`.
     pub fn remap_pointers(
         &mut self,
         map: &GlobalPointerTable<PointerId>,
@@ -237,6 +240,9 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
 }
 
 impl<'tcx> AnalysisCtxtData<'tcx> {
+    /// Update all `PointerId`s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
+    /// `PointerId`" counter to `counter`.  `map` and `counter` are usually computed together via
+    /// `LocalEquivSet::renumber`.
     pub fn remap_pointers(
         &mut self,
         lcx: LTyCtxt<'tcx>,
@@ -267,6 +273,8 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
     }
 }
 
+/// For every `PointerId` `p` that appears in `lty`, replace `p` with `map[p]` (except that
+/// `PointerId::NONE` is left unchanged) and return the updated `LTy`.
 fn remap_lty_pointers<'tcx, T>(lcx: LTyCtxt<'tcx>, map: &T, lty: LTy<'tcx>) -> LTy<'tcx>
 where
     T: Index<PointerId, Output = PointerId>,
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -58,6 +58,7 @@ pub use crate::pointer_id::PointerId;
 pub type LTy<'tcx> = LabeledTy<'tcx, PointerId>;
 pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, PointerId>;
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub struct LFnSig<'tcx> {
     pub inputs: &'tcx [LTy<'tcx>],
     pub output: LTy<'tcx>,
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -261,6 +261,10 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
 
         *next_ptr_id = counter;
     }
+
+    pub fn num_pointers(&self) -> usize {
+        self.next_ptr_id.num_pointers()
+    }
 }
 
 fn remap_lty_pointers<'tcx, T>(lcx: LTyCtxt<'tcx>, map: &T, lty: LTy<'tcx>) -> LTy<'tcx>
@@ -430,6 +434,7 @@ fn label_no_pointers<'tcx>(acx: &AnalysisCtxt<'_, 'tcx>, ty: Ty<'tcx>) -> LTy<'t
     })
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub struct GlobalAssignment {
     pub perms: GlobalPointerTable<PermissionSet>,
     pub flags: GlobalPointerTable<FlagSet>,
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -5,12 +5,15 @@ use crate::pointer_id::{
 };
 use crate::util::{describe_rvalue, RvalueDesc};
 use bitflags::bitflags;
+use rustc_hir::def_id::DefId;
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::{
     Body, HasLocalDecls, Local, LocalDecls, Operand, Place, PlaceElem, PlaceRef, ProjectionElem,
     Rvalue,
 };
 use rustc_middle::ty::{Ty, TyCtxt, TyKind};
+use std::collections::HashMap;
+use std::ops::Index;
 
 bitflags! {
     #[derive(Default)]
@@ -55,9 +58,16 @@ pub use crate::pointer_id::PointerId;
 pub type LTy<'tcx> = LabeledTy<'tcx, PointerId>;
 pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, PointerId>;
 
+pub struct LFnSig<'tcx> {
+    pub inputs: &'tcx [LTy<'tcx>],
+    pub output: LTy<'tcx>,
+}
+
 pub struct GlobalAnalysisCtxt<'tcx> {
     pub tcx: TyCtxt<'tcx>,
     pub lcx: LTyCtxt<'tcx>,
+
+    pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
     next_ptr_id: NextGlobalPointerId,
 }
@@ -83,6 +93,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         GlobalAnalysisCtxt {
             tcx,
             lcx: LabeledTyCtxt::new(tcx),
+            fn_sigs: HashMap::new(),
             next_ptr_id: NextGlobalPointerId::new(),
         }
     }
@@ -99,8 +110,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         AnalysisCtxt::from_data(self, mir, data)
     }
 
-    #[allow(dead_code)]
-    pub fn _new_pointer(&mut self) -> PointerId {
+    pub fn new_pointer(&mut self) -> PointerId {
         self.next_ptr_id.next()
     }
 
@@ -110,16 +120,25 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
 
     pub fn remap_pointers(
         &mut self,
-        _map: &GlobalPointerTable<PointerId>,
+        map: &GlobalPointerTable<PointerId>,
         counter: NextGlobalPointerId,
     ) {
         let GlobalAnalysisCtxt {
             tcx: _,
-            lcx: _,
+            lcx,
+            ref mut fn_sigs,
             ref mut next_ptr_id,
         } = *self;
 
-        // `GlobalAnalysisCtxt` doesn't yet have any fields that need remapping.
+        for sig in fn_sigs.values_mut() {
+            sig.inputs = lcx.mk_slice(
+                &sig.inputs
+                    .iter()
+                    .map(|&lty| remap_lty_pointers(lcx, map, lty))
+                    .collect::<Vec<_>>(),
+            );
+            sig.output = remap_lty_pointers(lcx, map, sig.output);
+        }
 
         *next_ptr_id = counter;
     }
@@ -230,13 +249,7 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
         } = *self;
 
         for lty in local_tys {
-            *lty = lcx.relabel(*lty, &mut |inner_lty| {
-                if inner_lty.label.is_none() {
-                    PointerId::NONE
-                } else {
-                    map[inner_lty.label]
-                }
-            });
+            *lty = remap_lty_pointers(lcx, &map, *lty);
         }
 
         for ptr in addr_of_local {
@@ -247,6 +260,19 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
 
         *next_ptr_id = counter;
     }
+}
+
+fn remap_lty_pointers<'tcx, T>(lcx: LTyCtxt<'tcx>, map: &T, lty: LTy<'tcx>) -> LTy<'tcx>
+where
+    T: Index<PointerId, Output = PointerId>,
+{
+    lcx.relabel(lty, &mut |inner_lty| {
+        if inner_lty.label.is_none() {
+            PointerId::NONE
+        } else {
+            map[inner_lty.label]
+        }
+    })
 }
 
 impl<'tcx> HasLocalDecls<'tcx> for AnalysisCtxt<'_, 'tcx> {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -119,9 +119,9 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         self.next_ptr_id.num_pointers()
     }
 
-    /// Update all `PointerId`s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
-    /// `PointerId`" counter to `counter`.  `map` and `counter` are usually computed together via
-    /// `GlobalEquivSet::renumber`.
+    /// Update all [`PointerId`]s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
+    /// [`PointerId`]" counter to `counter`.  `map` and `counter` are usually computed together via
+    /// [`GlobalEquivSet::renumber`][crate::equiv::GlobalEquivSet::renumber].
     pub fn remap_pointers(
         &mut self,
         map: &GlobalPointerTable<PointerId>,
@@ -240,9 +240,9 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
 }
 
 impl<'tcx> AnalysisCtxtData<'tcx> {
-    /// Update all `PointerId`s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
-    /// `PointerId`" counter to `counter`.  `map` and `counter` are usually computed together via
-    /// `LocalEquivSet::renumber`.
+    /// Update all [`PointerId`]s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
+    /// [`PointerId`]" counter to `counter`.  `map` and `counter` are usually computed together via
+    /// [`LocalEquivSet::renumber`][crate::equiv::LocalEquivSet::renumber].
     pub fn remap_pointers(
         &mut self,
         lcx: LTyCtxt<'tcx>,
@@ -273,8 +273,8 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
     }
 }
 
-/// For every `PointerId` `p` that appears in `lty`, replace `p` with `map[p]` (except that
-/// `PointerId::NONE` is left unchanged) and return the updated `LTy`.
+/// For every [`PointerId`] `p` that appears in `lty`, replace `p` with `map[p]` (except that
+/// [`PointerId::NONE`] is left unchanged) and return the updated `LTy`.
 fn remap_lty_pointers<'tcx, T>(lcx: LTyCtxt<'tcx>, map: &T, lty: LTy<'tcx>) -> LTy<'tcx>
 where
     T: Index<PointerId, Output = PointerId>,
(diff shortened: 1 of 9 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -55,10 +58,18 @@ pub use crate::pointer_id::PointerId;
 pub type LTy<'tcx> = LabeledTy<'tcx, PointerId>;
 pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, PointerId>;
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct LFnSig<'tcx> {
```
kkysen, line 62: Just `L` for `Labeled` seems a bit short here.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -209,16 +213,50 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+        substs: SubstsRef<'tcx>,
+        args: &[Operand<'tcx>],
+        dest: Place<'tcx>,
+    ) {
+        let sig = match self.acx.gacx.fn_sigs.get(&def_id) {
+            Some(&x) => x,
+            None => todo!("call to unknown function {:?}", def_id),
+        };
+        if substs.non_erasable_generics().next().is_some() {
+            todo!("call to generic function {:?} {:?}", def_id, substs);
```
kkysen, line 243:
```suggestion
            todo!("call to generic function {def_id:?} {substs:?}");
```

c2rust-analyze/src/main.rs
```diff
@@ -35,29 +41,113 @@ mod pointer_id;
...
+        MaybeUnset(None)
+    }
+}
+
+impl<T> MaybeUnset<T> {
+    pub fn set(&mut self, x: T) {
+        if self.0.is_some() {
+            panic!("value is already set");
+        }
+        self.0 = Some(x);
```
kkysen, line 61:
```suggestion
        if self.0.replace(x).is_some() {
            panic!("value is already set");
        }
```

c2rust-analyze/src/main.rs
```diff
@@ -35,29 +41,113 @@ mod pointer_id;
...
+    }
+
+    pub fn clear(&mut self) {
+        if self.0.is_none() {
+            panic!("value is already cleared");
+        }
+        self.0 = None;
```
kkysen, line 68:
```suggestion
        if self.0.take().is_none() {
            panic!("value is already cleared");
        }
```

c2rust-analyze/src/main.rs
```diff
@@ -35,29 +41,113 @@ mod pointer_id;
...
 fn run(tcx: TyCtxt) {
     let mut gacx = GlobalAnalysisCtxt::new(tcx);
-    let mut func_info = Vec::new();
+    let mut func_info = HashMap::new();
+
+    #[derive(Default)]
+    struct FuncInfo<'tcx> {
+        acx_data: MaybeUnset<AnalysisCtxtData<'tcx>>,
+        dataflow: MaybeUnset<DataflowConstraints>,
+        l_equiv: MaybeUnset<LocalEquivSet>,
```
kkysen, line 105: Previously `l` meant `Labeled`.  Here I assume it means `Local`.  I wouldn't use `l`/`L` for both of them.

c2rust-analyze/src/main.rs
```diff
@@ -35,29 +41,113 @@ mod pointer_id;
...
 
         // Assign PointerIds to local types
         assert!(acx.local_tys.is_empty());
         acx.local_tys = IndexVec::with_capacity(mir.local_decls.len());
         for (local, decl) in mir.local_decls.iter_enumerated() {
-            let lty = assign_pointer_ids(&mut acx, decl.ty);
+            let lty = match mir.local_kind(local) {
+                LocalKind::Var | LocalKind::Temp => acx.assign_pointer_ids(decl.ty),
+                LocalKind::Arg => {
+                    debug_assert!(local.as_usize() >= 1 && local.as_usize() <= mir.arg_count);
```
kkysen, line 146: Why a `debug_assert!` and not just an `assert!`?  If the assert fails but it's in release mode, this will cause an overflow, right?

c2rust-analyze/src/main.rs
```diff
@@ -73,40 +163,85 @@ fn run(tcx: TyCtxt) {
...
-        let mir = tcx.mir_built(ldid_const);
-        let mir = mir.borrow();
 
-        let (mut data, mut dataflow, l_equiv) = info;
-        // Remap pointers based on equivalence classes, so all members of an equivalence class now
-        // use the same `PointerId`.
-        let (l_counter, l_equiv_map) = l_equiv.renumber(&g_equiv_map);
+    for ldid in tcx.hir().body_owners() {
+        let info = func_info.get_mut(&ldid).unwrap();
+        let (l_counter, l_equiv_map) = info.l_equiv.renumber(&g_equiv_map);
```
kkysen, line 181: Why is `l` sometimes used without a separator (like `l*`) and sometimes with a `_` (like `l_*`)?

c2rust-analyze/src/main.rs
```diff
@@ -73,40 +163,85 @@ fn run(tcx: TyCtxt) {
...
         eprintln!("l_equiv_map = {:?}", l_equiv_map);
```
kkysen, line 182:
```suggestion
        eprintln!("l_equiv_map = {l_equiv_map:?}");
```

c2rust-analyze/src/main.rs
```diff
@@ -73,40 +163,85 @@ fn run(tcx: TyCtxt) {
...
+        info.lasn.set(lasn);
+    }
 
-        let mut lasn =
-            LocalAssignment::new(acx.num_pointers(), PermissionSet::UNIQUE, FlagSet::empty());
-        let mut asn = gasn.and(&mut lasn);
+    let order = walk_callgraph(tcx);
+    eprintln!("callgraph traversal order:");
+    for &ldid in &order {
+        eprintln!("  {:?}", ldid);
```
kkysen, line 201:
```suggestion
        eprintln!("  {ldid:?}");
```

c2rust-analyze/src/main.rs
```diff
@@ -180,11 +315,36 @@ fn run(tcx: TyCtxt) {
...
 }
 
-fn assign_pointer_ids<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, ty: Ty<'tcx>) -> LTy<'tcx> {
-    acx.lcx().label(ty, &mut |ty| match ty.kind() {
-        TyKind::Ref(_, _, _) | TyKind::RawPtr(_) => acx.new_pointer(),
-        _ => PointerId::NONE,
-    })
+trait AssignPointerIds<'tcx> {
+    fn lcx(&self) -> LTyCtxt<'tcx>;
+    fn new_pointer(&mut self) -> PointerId;
```
kkysen, line 320:
```suggestion
    fn lcx(&self) -> LTyCtxt<'tcx>;
    
    fn new_pointer(&mut self) -> PointerId;
```

c2rust-analyze/src/main.rs
```diff
@@ -222,6 +382,68 @@ fn describe_span(tcx: TyCtxt, span: Span) -> String {
...
+    impl<'tcx, F: FnMut(LocalDefId)> Visitor<'tcx> for CalleeVisitor<'_, 'tcx, F> {
+        fn visit_operand(&mut self, operand: &Operand<'tcx>, _location: Location) {
+            let ty = operand.ty(self.mir, self.tcx);
+            if let Some(Callee::Other { def_id, .. }) = util::ty_callee(self.tcx, ty) {
+                if let Some(ldid) = def_id.as_local() {
+                    (self.f)(ldid);
+                }
+            }
+        }
+    }
```
kkysen, line 442: I didn't realize you can `impl` inside of a function.  Neat!

c2rust-analyze/src/main.rs
```diff
@@ -222,6 +382,68 @@ fn describe_span(tcx: TyCtxt, span: Span) -> String {
...
+    let mut stack = Vec::new();
+
+    for root_ldid in tcx.hir().body_owners() {
+        if seen.contains(&root_ldid) {
+            continue;
+        }
+        stack.push(Visit::Pre(root_ldid));
+        while let Some(x) = stack.pop() {
+            match x {
```
kkysen, line 402:
```suggestion
        while let Some(ldid) = stack.pop() {
            match ldid {
```

c2rust-analyze/src/main.rs
```diff
@@ -222,6 +382,68 @@ fn describe_span(tcx: TyCtxt, span: Span) -> String {
     format!("{}: {}{}{}", line, src1, src2, src3)
 }
 
+/// Return all `body_owners`, ordered according to a postorder traversal of the graph of references
+/// between bodies.
+fn walk_callgraph(tcx: TyCtxt) -> Vec<LocalDefId> {
```
kkysen, line 387: Could be an iterator, external like `Iterator` or internal like `for_each_callee`.

c2rust-analyze/tests/filecheck/call1.rs
```diff
@@ -0,0 +1,48 @@
+// With a naive iteration order, reaching a fixpoint used to take 3 iterations.  By following a
+// postorder traversal of the callgraph, we reduce that to 2 iterations: the first iteration
+// computes all the right permissions, and the second checks that we've actually reached a
+// fixpoint.
```
kkysen, line 4: Does it make sense for these to be doc comments?

c2rust-analyze/src/context.rs
```diff
@@ -247,6 +261,23 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
 
         *next_ptr_id = counter;
     }
+
+    pub fn num_pointers(&self) -> usize {
+        self.next_ptr_id.num_pointers()
+    }
+}
+
+fn remap_lty_pointers<'tcx, T>(lcx: LTyCtxt<'tcx>, map: &T, lty: LTy<'tcx>) -> LTy<'tcx>
```
aneksteind, line 270: would you mind adding a docstring here explaining the use case

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -48,16 +49,25 @@ impl<'tcx> TypeChecker<'tcx, '_> {
     }
 
     pub fn visit_place_ref(&mut self, pl: PlaceRef<'tcx>, ctx: PlaceContext) -> LTy<'tcx> {
+        let mutbl = match ctx {
```
aneksteind, line 52: do you think this would be useful as its own utility function?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -48,16 +49,25 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+
         let mut lty = self.acx.local_tys[pl.local];
         let mut prev_deref_ptr = None;
 
         for proj in pl.projection {
             match proj {
                 ProjectionElem::Deref => {
                     // All derefs except the last are loads, to retrieve the pointer for the next
-                    // deref.  The last deref is either a load or a store, depending on `ctx`.
+                    // deref.  However, if the last deref is `&mut` (and is used mutably), then the
```
aneksteind, line 65: is the last deref being &mut and used mutably determined exclusively by `mutbl`?

aneksteind (approved):
a general comment -- i think more documentation and longer variable names would have helped my understanding without having to scroll up and down for more context. any additions along those lines would be welcome

c2rust-analyze/src/main.rs
spernsteiner: In the future (in code I've already written based on top of this branch), it seems like we'll need to iterate over the ordering multiple times, so I think leaving it as `Vec` is best.

c2rust-analyze/tests/filecheck/call1.rs
spernsteiner: This comment is describing reasoning for the `// CHECK` below, not a function or module.  Plus, I can't imagine anyone ever running `rustdoc` on these files.

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: It turns out the use of `PlaceContext` was a holdover from previous code that was implemented as a `mir::Visitor`.  I changed these functions to take `mutbl: Mutability` directly instead.

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: Yes - I updated the comment slightly

c2rust-analyze/src/main.rs
spernsteiner: OK, I renamed all the `l_`/`g_`s to `local_`/`global_`

c2rust-analyze/src/main.rs
spernsteiner: `x` is a wrapper and not (just) a `LocalDefId`, so naming it `ldid` would be misleading IMO

c2rust-analyze/src/main.rs
spernsteiner: `ldid` is (roughly) an initialism for `LocalDefId`, similar to `did` for `DefId`.  The latter appears pervasively within rustc.

c2rust-analyze/src/main.rs
spernsteiner: IMO this makes the logic harder to follow

c2rust-analyze/src/main.rs
spernsteiner: IMO this makes the logic harder to follow

c2rust-analyze/src/context.rs
spernsteiner: `LTy` is used pervasively throughout this codebase, so I don't think it needs a long name.  And `LFnSig` is named by analogy with rustc's ordinary `FnSig` and `Ty`.

c2rust-analyze/src/main.rs
spernsteiner: If this condition doesn't hold, the next line will panic with index out of bounds (possibly after underflowing to `usize::MAX`).  The assertion here is technically redundant - its only purpose is to express the expected property in a more straightforward way.

spernsteiner:
I think I've addressed all the previous comments.  I used the "resolve conversation" button liberally to help keep track of which threads I'd already handled - feel free to unresolve if there's more to discuss.

c2rust-analyze/tests/filecheck/call1.rs
kkysen: Okay, that's reasonable.

c2rust-analyze/src/main.rs
kkysen: I think `walk` isn't the best name then.  How about `collect` or `walk_and_collect`?  `walk` sounds like an iterator (internal or external).

c2rust-analyze/src/main.rs
kkysen: I think `ldid` makes more sense than `x`.  It always only contains an `ldid`, it's just either `Pre` or `Post`.

c2rust-analyze/src/main.rs
kkysen: Why was this resolved?  There's spacing between the other methods in this trait.

kkysen: Oh wait, I think I was looking at an outdated version.

c2rust-analyze/src/main.rs
spernsteiner: OK, I've renamed it to `body_owners_postorder` (since it returns the same sequence as `tcx.body_owners()`, but sorted according to a postorder traversal)

c2rust-analyze/src/main.rs
spernsteiner: This function takes a recursive algorithm (postorder graph traversal) and implements it iteratively.  `Visit` is best thought of not as "a `LocalDefId` plus an extra bit" (even though concretely that's exactly what it is), but as "a step of the original recursive algorithm".

kkysen: Can it be named `visit` then?

spernsteiner: OK, renamed to `visit`

c2rust-analyze/src/context.rs
kkysen: Ok, I think it's fine as long as `L` only means `Labeled`, but if it means `locals` or anything else elsewhere, then it gets very confusing.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -117,14 +112,8 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-            Operand::Copy(pl) => {
-                let ctx = PlaceContext::NonMutatingUse(NonMutatingUseContext::Copy);
-                self.visit_place(pl, ctx)
-            }
-            Operand::Move(pl) => {
-                let ctx = PlaceContext::NonMutatingUse(NonMutatingUseContext::Move);
-                self.visit_place(pl, ctx)
-            }
+            Operand::Copy(pl) => self.visit_place(pl, Mutability::Not),
+            Operand::Move(pl) => self.visit_place(pl, Mutability::Not),
```
kkysen, line 116: Could these match arms be combined?

c2rust-analyze/src/main.rs
kkysen: I disagree.  The `Option` methods are designed for exactly this and are idiomatic.

c2rust-analyze/src/main.rs
```diff
@@ -66,47 +169,107 @@ fn run(tcx: TyCtxt) {
...
+        info.dataflow.set(dataflow);
+        info.local_equiv.set(local_equiv);
+        func_info.insert(ldid, info);
     }
 
-    // Compute permission and flag assignments.
+    // Remap pointers based on equivalence classes, so all members of an equivalence class now use
+    // the same `PointerId`.
+    let (global_counter, global_equiv_map) = global_equiv.renumber();
+    eprintln!("global_equiv_map = {:?}", global_equiv_map);
```
kkysen, line 190:
```suggestion
    eprintln!("global_equiv_map = {global_equiv_map:?}", );
```

kkysen (changes requested):
Can you make the referenced items in doc comments into real intra-doc links?  It helps a lot with navigating documentation and ensuring they stay up-to-datr through refactors.

Also, I think you should use the existing, idiomatic `Option` methods when the logic is exactly what those methods do.  If not using them was not as clear, then why are they in `core`?

There's also a few other small suggestions I made.

c2rust-analyze/src/main.rs
aneksteind: fwiw I agree with Stuart that this is harder to follow -- for me it's because I'm less familiar with `replace`

kkysen: There's also `std::mem::{replace,take}`.  It's an idiom Rust uses in multiple places, so I do think it's still worth it to use.  You could use `match` instead of `map` or `expect`, too, as the logic is easier for a beginner to follow, but you would probably encourage people to use the dedicated method and have them learn it.

c2rust-analyze/src/main.rs
kkysen: I think we also get a better, more standard, more recognizable and related to `Option` error message here with `expect`.

c2rust-analyze/src/main.rs
spernsteiner: The connotation of `replace` to me is that there are two interesting values, a new one and (possibly) an old one.  Here, we don't care about the old value at all (and in fact we're asserting that there isn't one), so `replace` seems like an unusual choice.

Similarly, `take` is about extracting the value out of an option, so using it when we don't care about the value is a bit weird.  I would guess that this `x.take().unwrap();` type of construct (where the result is unused) is almost unheard of in Rust code.

Overall, I think the original versions of these functions express the high-level intent about as clearly and directly as they possibly could.  Switching to `replace`/`take` brings in concepts of old/new values which are not actually helpful for understanding.

c2rust-analyze/src/main.rs
spernsteiner: I hadn't considered the panic message.  IMO, the fact that the message would mention `Option` is a (very minor) downside of the `expect` version.  The user of `MaybeUnset<T>` should not be thinking of it as an `Option<T>`.  They should think of it as a `T`, with the minor caveat that Rust's normal static checks for variable initialization checks don't apply.

c2rust-analyze/src/main.rs
kkysen: [`Cell::set` is implemented](https://doc.rust-lang.org/src/core/cell.rs.html#362) using `Cell::replace`.  That's exactly what we're doing here, using replace to set a new value.  Except unlike `Cell`, we don't care about propagating the old value to the user, but we want to make sure we didn't actually replace a value, so we panic instead.

c2rust-analyze/src/main.rs
kkysen: Here I see `take` as in it's taking away the current value, the same as clearing it.

spernsteiner: This simply is not an idiomatic use of `take`.  The primary purpose of `take` is to get the old value.  This is reflected in the docs ("Takes the value out of the option") and in common usage.  I did a quick grep of rust-lang/rust, and every instance of `take()`, `take().unwrap()`, or `take().expect()` that I found (outside of test code) was actually using the extracted value for something afterwards.

spernsteiner:
I've addressed the remaining comments.  I added doc links only in the comments that were added or modified in this PR, to avoid polluting the diff with unrelated changes - the rest can be done in a separate PR if needed.

PR #656, opened 2022-09-01
flow data structure, creation

boyland-pf:
The flow data structure is created, and is filled with the correct information for 4 of the 6 flags needed for `c2rust-analyze`:

- `READ`
- `WRITE`
- `OFFSET_ADD`
- `OFFSET_SUB`

The assumption made is that nodes that can alias each other form a tree in which the source fields are the edges.

### DIFF
--- a/pdg/src/info.rs
+++ b/pdg/src/info.rs
@@ -45,10 +45,7 @@ fn init_flows(n_id: NodeId, n: &Node) -> Flows {
 /// Gathers information from a [`Graph`] (assumed to be acyclic and topologically sorted but not
 /// necessarily connected) for each [`Node`] in it whether there is a path following 'source' edges
 /// from any [`Node`] with a given property to the [`Node`] in question.
-///
-/// [`Node`]: crate::graph::Node
-/// [`Graph`]: graph::graph::Graph
-fn create_flow_info(g: &Graph) -> HashMap<NodeId, Flows> {
+fn set_flow_info(g: &mut Graph) -> HashMap<NodeId, Flows> {
     let mut f = HashMap::from_iter(
         g.nodes
             .iter_enumerated()
@@ -64,20 +61,14 @@ fn create_flow_info(g: &Graph) -> HashMap<NodeId, Flows> {
             parent.neg_offset = parent.neg_offset.or(cur.neg_offset);
         }
     }
-    f
+    node.info = Some(NodeInfo { flows_to: flows.remove(&n_id).unwrap()});
 }
 
 /// Initialize [`Node::info`] for each [`Node`].
 ///
 /// This includes all of the information answering questions of the form "is there a [`Node`] that this is an ancestor of with trait X".
-///
-/// [`Node`]: crate::graph::Node
-/// [`Node::info`]: crate::graph::Node::info
 pub fn add_info(pdg: &mut Graphs) {
     for g in &mut pdg.graphs {
-        let mut flows = create_flow_info(&g);
-        for (n_id,mut node) in g.nodes.iter_enumerated_mut() {
-            node.info = Some(NodeInfo { flows_to: flows.remove(&n_id).unwrap()});
-        }
+        set_flow_info(&mut g);
     }
 }
--- a/pdg/src/info.rs
+++ b/pdg/src/info.rs
@@ -1,7 +1,7 @@
+use crate::graph::{Graph, NodeId, NodeKind};
 use crate::Graphs;
-use crate::graph::{Graph,NodeId,NodeKind};
-use std::fmt::{self, Debug, Display, Formatter};
 use std::collections::HashMap;
+use std::fmt::{self, Debug, Display, Formatter};
 
 /// Information generated from the PDG proper that is queried by static analysis.
 ///
@@ -15,10 +15,10 @@ pub struct NodeInfo {
 }
 
 /// Contains information about what kinds of [`Node`]s a [`Node`] flows to.
-/// Load and store kinds contain both Load/Store-Value and Load/Store-Addr. 
+/// Load and store kinds contain both Load/Store-Value and Load/Store-Addr.
 ///
 /// [`Node`]: crate::graph::Node
-#[derive(Debug, Hash, Clone, Copy,PartialEq,Default)]
+#[derive(Debug, Hash, Clone, Copy, PartialEq, Default)]
 pub struct FlowInfo {
     load: Option<NodeId>,
     store: Option<NodeId>,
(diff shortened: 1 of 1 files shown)


pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
+    neg_offset: Option<NodeId>,
+}
 
 impl Display for NodeInfo {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "")
     }
 }
 
+fn init_flows(n_id: NodeId, n: &Node) -> Flows {
```
aneksteind, line 35: I think this would be better off as a method in `Flows` like `Flows::new`

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
+pub struct NodeInfo {
+    flows_to: Flows,
+}
+
+/// Contains information about what kinds of [`Node`]s the [`Node`] flows to.
+/// Load and store kinds contain both 
+///
+/// [`Node`]: crate::graph::Node.
+#[derive(Debug, Hash, Clone, Copy,PartialEq,Default)]
+pub struct Flows {
```
aneksteind, line 22: I think a name like `FlowInfo` would be better

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
 ///
 /// [`Node`]: crate::graph::Node
-#[derive(Hash, Clone, PartialEq, Eq, Debug)]
-pub struct NodeInfo {}
+#[derive(Hash, Clone, PartialEq, Debug)]
+pub struct NodeInfo {
+    flows_to: Flows,
+}
+
+/// Contains information about what kinds of [`Node`]s the [`Node`] flows to.
```
aneksteind, line 17: I think the definition of "flows" needs to be explained here in this documentation

aneksteind, line 17:
```suggestion
/// Contains information about what kinds of [`Node`]s a [`Node`] flows to.
```

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
+    }
+}
+
+
+/// Gathers information from a [`Graph`] (assumed to be acyclic and topologically sorted but not
+/// necessarily connected) for each [`Node`] in it whether there is a path following 'source' edges
+/// from any [`Node`] with a given property to the [`Node`] in question.
+///
+/// [`Node`]: crate::graph::Node
+/// [`Graph`]: graph::graph::Graph
```
aneksteind, line 50: are these necessary?

aneksteind: i'm just not sure what it does

pdg/src/info.rs
kkysen: Also, does it need the whole `Node` or just the `NodeKind`?

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
 ///
 /// [`Node`]: crate::graph::Node
 /// [`Node::info`]: crate::graph::Node::info
 pub fn add_info(pdg: &mut Graphs) {
     for g in &mut pdg.graphs {
-        for mut node in &mut g.nodes {
-            node.info = Some(NodeInfo {});
+        let mut flows = create_flow_info(&g);
+        for (n_id,mut node) in g.nodes.iter_enumerated_mut() {
+            node.info = Some(NodeInfo { flows_to: flows.remove(&n_id).unwrap()});
```
aneksteind, line 80: i think this `node.info = ...` line can be moved into `create_flow_info` after [line 65](https://github.com/immunant/c2rust/pull/656/files#diff-653d59c0200a732696efee707278862a58af520ec8115a110028bde9aae403ffR65), and can be renamed `set_flow_info` without returning the hashmap. then this for loop can go away

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
+/// [`Node`]: crate::graph::Node
+/// [`Graph`]: graph::graph::Graph
+fn create_flow_info(g: &Graph) -> HashMap<NodeId, Flows> {
+    let mut f = HashMap::from_iter(
+        g.nodes
+            .iter_enumerated()
+            .map(|(idx, node)| (idx, init_flows(idx, node))),
+    );
+    for (n_id, node) in g.nodes.iter_enumerated().rev() {
+        let cur: Flows = *(f.get(&n_id).unwrap());
```
aneksteind, line 58:
```suggestion
        let node_flow_info: Flows = *(flow_map.get(&n_id).unwrap());
```

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
+
+
+/// Gathers information from a [`Graph`] (assumed to be acyclic and topologically sorted but not
+/// necessarily connected) for each [`Node`] in it whether there is a path following 'source' edges
+/// from any [`Node`] with a given property to the [`Node`] in question.
+///
+/// [`Node`]: crate::graph::Node
+/// [`Graph`]: graph::graph::Graph
+fn create_flow_info(g: &Graph) -> HashMap<NodeId, Flows> {
+    let mut f = HashMap::from_iter(
```
aneksteind, line 52:
```suggestion
    let mut flow_map = HashMap::from_iter(
```

pdg/src/info.rs
boyland-pf: Thumbs up. I will need to use the flow later, but I guess I can just get that from the info field later on.

boyland-pf: This should be fixed now.

pdg/src/info.rs
boyland-pf: It's named cur because there's both its and its parent's node_info being manipulated in the loop. Can I rename it to cur_node_flow_info instead?

pdg/src/info.rs
boyland-pf: Changed on both counts.

pdg/src/info.rs
aneksteind: sure

aneksteind (approved):

kkysen:
I still have some comments, I'll review it this weekend.

kkysen (changes requested):

pdg/src/info.rs
```diff
@@ -1,31 +1,83 @@
...
+pub struct Flows {
+    load: Option<NodeId>,
+    store: Option<NodeId>,
+    pos_offset: Option<NodeId>,
+    neg_offset: Option<NodeId>,
+}
 
 impl Display for NodeInfo {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "")
```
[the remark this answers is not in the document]
kkysen: I think I'll actually just do this in a separate PR.

pdg/src/info.rs
```diff
@@ -1,31 +1,76 @@
...
+pub struct FlowInfo {
+    load: Option<NodeId>,
+    store: Option<NodeId>,
+    pos_offset: Option<NodeId>,
+    neg_offset: Option<NodeId>,
+}
+
+impl FlowInfo {
+    //initializing flow information based on a node's kind
+    fn new(n_id: NodeId, k: &NodeKind) -> FlowInfo {
```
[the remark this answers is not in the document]
kkysen: It is in `master` now after #657.  I was planning on a further refactor PR for this code after this PR lands, so I can just add that there.

kkysen (approved):

PR #658, opened 2022-09-03
Improve/update docs for `NodeKind`

kkysen:
It might be easier to review each commit individually so that it's easier to see the differences.

### DIFF
--- a/pdg/src/graph.rs
+++ b/pdg/src/graph.rs
@@ -15,7 +15,7 @@ use crate::util::ShortOption;
 
 #[derive(Debug, Eq, PartialEq, Hash, Clone)]
 pub enum NodeKind {
-    /// A [`Copy`](Self::Copy) from one [`Local`] to another.
+    /// A copy from one [`Local`] to another.
     ///
     /// This also covers casts such as `&mut T` to `&T` or `&T` to `*const T`
     /// that don't change the type or value of the pointer.
--- a/pdg/src/graph.rs
+++ b/pdg/src/graph.rs
@@ -22,6 +22,7 @@ pub enum NodeKind {
     /// Field projection.  Used for operations like `_2 = &(*_1).0`.  Nested field accesses like
     /// `_4 = &(*_1).x.y.z` are broken into multiple `Node`s, each covering one level.
     Field(Field),
+
     /// Pointer arithmetic.  The `isize` is the concrete offset distance.  We use this to detect
     /// when two pointers always refer to different indices.
     Offset(isize),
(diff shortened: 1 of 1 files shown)


pdg/src/graph.rs
```diff
@@ -15,47 +15,85 @@ use crate::util::ShortOption;
...
-    /// Pointer arithmetic.  The `isize` is the concrete offset distance.  We use this to detect
-    /// when two pointers always refer to different indices.
+
+    /// Pointer arithmetic.
+    ///
+    /// The [`isize`] is the concrete offset distance.
+    /// We use this to detect when two pointers always refer to different indices.
     Offset(isize),
 
     // Operations that can't have a `source`.
```
kkysen, line 37: What's this referring to?

pdg/src/graph.rs
```diff
@@ -15,47 +15,85 @@ use crate::util::ShortOption;
...
+    /// Details TBD.
     IntToPtr,
-    /// The result of loading a value through some other pointer.  Details TBD.
+
+    /// The result of loading a value through some other pointer.
+    ///
+    /// Details TBD.
     LoadValue,
 
     // Operations that can't be the `source` of any other operation.
```
kkysen, line 74: What's this referring to?

pdg/src/graph.rs
```diff
@@ -15,47 +15,85 @@ use crate::util::ShortOption;
 
 #[derive(Debug, Eq, PartialEq, Hash, Clone)]
 pub enum NodeKind {
-    /// A copy from one local to another.  This also covers casts such as `&mut T` to `&T` or `&T`
-    /// to `*const T` that don't change the type or value of the pointer.
+    /// A [`Copy`](Self::Copy) from one [`Local`] to another.
```
aneksteind, line 18: i think it's fine to refer to "copy" colloquially here, no need to self-reference

pdg/src/graph.rs
aneksteind: it refers to `Node::source`, which will be `None` for all `Node`s with any of the following `NodeKind`s

pdg/src/graph.rs
aneksteind: Like above, refers to `Node::source`, which can never point to a `Node` (via `Node::dest`) which has any of the following `NodeKind`s

pdg/src/graph.rs
kkysen: I meant what is the comment and operation referring to.  Which variant?

aneksteind (approved):
I imagine you'll want to update the documentation in response to my comments and I trust what you put will be sufficient, so LGTM

pdg/src/graph.rs
aneksteind: all of them that follow until the next similar comment

pdg/src/graph.rs
kkysen: Okay that makes sense.  I'll change that back.

pdg/src/graph.rs
kkysen: Ahh, got it.  I'll make that clearer.

kkysen: I'll put that in the docs for each variant it applies to, as I think doing it as is is a bit brittle if the variants are ever moved around, and it wasn't that clear to begin with.

pdg/src/graph.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/658/commits/40f6ecfd73752c804da9be32c67d79acc36dcade.

pdg/src/graph.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/658/commits/dc29cfaabeb0dbc8764247e84decdad82ad3f3e8.

pdg/src/graph.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/658/commits/dc29cfaabeb0dbc8764247e84decdad82ad3f3e8.

PR #660, opened 2022-09-12
Add uniqueness (`&mut`-ability) to `NodeInfo`

boyland-pf:
Finished generating info for the uniqueness "query". 10 unit tests. The one thing I didn't include is formatting because that would require a new snapshot.


pdg/src/info.rs
```diff
@@ -1,24 +1,32 @@
...
 ///
 /// Includes information about what kinds of [`Node`]s the [`Node`] flows to,
-/// and eventually will also include its ability to be used as a `&mut`.
+/// as well as its ability to be used as a `&mut`.
 ///
 /// [`Node`]: crate::graph::Node
 #[derive(Hash, Clone, PartialEq, Eq, Debug)]
 pub struct NodeInfo {
     flows_to: FlowInfo,
+    unique: bool,
```
aneksteind, line 18: documenting `unique` and what it means would be helpful

boyland-pf: what if I add in the parenthetical
`as well as its ability to be used as a `&mut` (signified by 'unique')`?

pdg/src/info.rs
```diff
@@ -67,18 +67,486 @@ fn set_flow_info(g: &mut Graph) {
...
+fn collect_children(g: &Graph) -> HashMap<NodeId, Vec<NodeId>> {
+    let mut m = HashMap::new();
+    for (par, chi) in g
+        .nodes
+        .iter_enumerated()
+        .filter_map(|(idx, node)| Some((node.source?, idx)))
+    {
+        m.entry(par).or_insert_with(Vec::new).push(chi)
+    }
+    for par in g.nodes.indices() {
```
aneksteind, line 101: this for loop can go above the first, the other can just push to the existing vector without `or_insert_with` or `new`

pdg/src/info.rs
```diff
@@ -67,18 +67,486 @@ fn set_flow_info(g: &mut Graph) {
...
+        let _ = m.try_insert(par, Vec::new());
+    }
+    m
+}
+
+///Given a list of nodes of the same parent and information about them,
+///determines if any have conflicts with any of the others.
+///Children which are not a field cannot be live at the same time as any other child.
+///Children which are a field cannot be live at the same time as any other one of the same field.
```
aneksteind, line 110:
```suggestion
/// Given a list of nodes of the same parent and information about them,
/// determines if any have conflicts with any of the others.
/// Children which are not a field cannot be live at the same time as any other child.
/// Children which are a field cannot be live at the same time as any other one of the same field.
```

pdg/src/info.rs
```diff
@@ -67,18 +67,486 @@ fn set_flow_info(g: &mut Graph) {
...
+    children: &HashMap<NodeId, Vec<NodeId>>,
+    descs: &HashMap<NodeId, NodeId>,
+) -> bool {
+    let mut max_descs: HashMap<Option<Field>, NodeId> = HashMap::new();
+    for id in children.get(n_id).unwrap() {
+        let sib_node: &Node = g.nodes.get(*id).unwrap();
+        let my_last_desc = descs.get(&id).unwrap().clone();
+        print!("{}\n", my_last_desc);
+        if matches!(max_descs.get(&None), Some(max_desc) if max_desc > id)
+            || matches!(sib_node.kind,NodeKind::Field(f) if matches!(max_descs.get(&Some(f)),Some(max_desc_field) if max_desc_field > id))
```
aneksteind, line 123: have you run `rustfmt`?

pdg/src/info.rs
```diff
@@ -67,18 +67,486 @@ fn set_flow_info(g: &mut Graph) {
...
+    n_id: &NodeId,
+    children: &HashMap<NodeId, Vec<NodeId>>,
+    descs: &HashMap<NodeId, NodeId>,
+) -> bool {
+    let mut max_descs: HashMap<Option<Field>, NodeId> = HashMap::new();
+    for id in children.get(n_id).unwrap() {
+        let sib_node: &Node = g.nodes.get(*id).unwrap();
+        let my_last_desc = descs.get(&id).unwrap().clone();
+        print!("{}\n", my_last_desc);
+        if matches!(max_descs.get(&None), Some(max_desc) if max_desc > id)
```
aneksteind, line 122: i think this deserves a comment explaining what you're matching for here and what makes it a conflict

pdg/src/info.rs
aneksteind: i think it should be added above the attribute itself, not in the general struct documentation

pdg/src/info.rs
boyland-pf: This part isn't changed by it.

pdg/src/info.rs
```diff
@@ -91,15 +91,15 @@ fn get_last_desc(g: &mut Graph) -> HashMap<NodeId, NodeId> {
...
     for (par, chi) in g
         .nodes
         .iter_enumerated()
         .filter_map(|(idx, node)| Some((node.source?, idx)))
     {
-        m.entry(par).or_insert_with(Vec::new).push(chi)
-    }
-    for par in g.nodes.indices() {
-        let _ = m.try_insert(par, Vec::new());
+        m.get_mut(&par).unwrap().push(chi);
```
aneksteind, line 102: i think you can `map` here instead of unwrap

pdg/src/info.rs
```diff
@@ -67,18 +67,481 @@ fn set_flow_info(g: &mut Graph) {
...
+    }
+    false
+}
+
+fn set_uniqueness(g: &mut Graph) {
+    let children = collect_children(g);
+    let last_descs = get_last_desc(g);
+    let mut nonuniqueness: HashSet<NodeId> = HashSet::new();
+    for (n_id, node) in g.nodes.iter_enumerated() {
+        if matches!(node.source, Some(p_id) if nonuniqueness.contains(&p_id))
```
aneksteind, line 143: does this imply that the whole connected tree is non-unique if any nodes contained in it are?

pdg/src/info.rs
aneksteind: i would add some spacing after the comments and see if that changes, makes things more readable IMO

pdg/src/info.rs
boyland-pf: I've added a comment describing what the two cases are

pdg/src/info.rs
boyland-pf: There are test cases in which the other direction doesn't hold - if there's a conflict between A's child B and A's child C, A can't be unique, but its parent Z can still be.

pdg/src/info.rs
boyland-pf: we're guaranteed the unwrap will work, so isn't unwrap better here because it's simpler?

pdg/src/info.rs
aneksteind: the reason I asked is because I can't tell whether or not the order of traversal in nodes matters or not, but it sounds like it doesn't because it's strictly limited to immediate parent and children. is that right?

pdg/src/info.rs
aneksteind: i think `map` is better because it decouples this line of code from the knowledge of what the code above it does

pdg/src/info.rs
```diff
@@ -67,18 +67,483 @@ fn set_flow_info(g: &mut Graph) {
...
+        let b1 = mk_copy(&mut g, a);
+        let b2 = mk_store_addr(&mut g, b1);
+        let c1 = mk_copy(&mut g, b1);
+        let c2 = mk_store_addr(&mut g, c1);
+        let b3 = mk_store_addr(&mut g, b1);
+        let c3 = mk_store_addr(&mut g, c1);
+
+        let pdg = build_pdg(g);
+        assert!(info(&pdg, a).unique);
+        assert!(!info(&pdg, b1).unique);
```
aneksteind, line 349: i'm trying to make sense of these, why should the ancestors of `a` be non-unique if they're written to?

pdg/src/info.rs
```diff
@@ -67,18 +67,483 @@ fn set_flow_info(g: &mut Graph) {
...
+    #[test]
+    fn okay_use_different_fields() {
+        let mut g = Graph::default();
+        //let mut a = Point {x: 0, y:0}; // A
+        //let b = &mut a.x;              // B1
+        //let c = &mut a.y;              // C1
+        //*b = 1;                        // B2
+        //*c = 2;                        // C2
+        //
+        //A
```
aneksteind, line 366: spaces after `//` would help readability

pdg/src/info.rs
```diff
@@ -67,18 +67,483 @@ fn set_flow_info(g: &mut Graph) {
...
+            desc_map.insert(p_id, std::cmp::max(cur_node_last_desc, parent_last_desc));
+        }
+    }
+    desc_map
+}
+
+/// Finds the inverse of a [`Graph`], each [`Node`] mapping to a list of its children.
+fn collect_children(g: &Graph) -> HashMap<NodeId, Vec<NodeId>> {
+    let mut m = HashMap::new();
+    for par in g.nodes.indices() {
```
kkysen, line 94: What's `par`?

pdg/src/info.rs
kkysen: How would `.map` be used here?

pdg/src/info.rs
boyland-pf: each possible parent

pdg/src/info.rs
aneksteind: after our conversation offline, the test case makes sense to me

pdg/src/info.rs
```diff
@@ -67,18 +67,483 @@ fn set_flow_info(g: &mut Graph) {
...
+        //a.z.r = 200;                                                              //A2
+        //*b = 4;                                                                   //B2
+        //*c = 2;                                                                   //C2
+        //let d = &mut a;                                                           //D1
+        //*d = ColorPoint {x: 0, y: 0, z: Color {r: 20, g:200, b: 20}};             //D2
+        //let e = &mut a.z;                                                         //E
+        //let f = &mut e.g;                                                         //F1
+        //let g = &mut e.g;                                                         //G
+        //*f = 3;                                                                   //F2
+        //a.z.r = 100;                                                              //A3
```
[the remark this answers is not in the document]
aneksteind: and please check the rest for parity

pdg/src/info.rs
```diff
@@ -67,18 +67,483 @@ fn set_flow_info(g: &mut Graph) {
...
+
+        let a = mk_addr_of_local(&mut g, 0_u32);
+        let b1 = mk_field(&mut g, a, 0_u32);
+        let c1 = mk_field(&mut g, a, 1_u32);
```
aneksteind, line 513: i think it would be helpful to have variables such as
```rust
let (x, y, z) = (0_u32, 1_u32, 2_u32);
```
to correspond to the comments more accurately

pdg/src/info.rs
```diff
@@ -67,18 +67,483 @@ fn set_flow_info(g: &mut Graph) {
...
+        assert!(!info(&pdg, c2).unique);
+        assert!(!info(&pdg, b2).unique);
+    }
+
+    #[test]
+    fn lots_of_siblings() {
+        let mut g = Graph::default();
+        //let mut a = ColorPoint {x: 0, y: 0, z: Color{ r: 100, g: 100, b: 100}};   //A1
+        //let b = &mut a.x;                                                         //B1
+        //let c = &mut a.y;                                                         //C1
```
aneksteind, line 499: thoughts on interleaving these comments above each respective line, rather than have them all at the top?

pdg/src/info.rs
aneksteind: would `m.get_mut(&par).map(|ref mut v| v.push(chi); v)` or not work? but on second thought maybe `unwrap` is fine because the map option relies on side affects

kkysen: Yeah, `map` is generally used for mapping to a value, not running a side effect.  The equivalent to the `map` but more idiomatic would be an `if let Some`.  But we also want to assert here that it's `Some`, right?  At the very least, `unwrap` serves as documentation that it will always be `Some`.

pdg/src/info.rs
kkysen: I think it should be in doc comments for the function so that it's run and tested.  It's easy to make a small mistake as is.

pdg/src/info.rs
boyland-pf: Order of traversal does matter, because any descendant of a non-unique node is non-unique. My assumption which I believe is listed in the comments is that the nodes are ordered.

pdg/src/info.rs
aneksteind: that sounds reasonable

pdg/src/info.rs
aneksteind: i don't see how these comments can be tested. considering that this function is the test itself, what do you envision?

kkysen: I mean we can test that they're well-formed Rust.

boyland-pf: I believe I have tested them all, but that would make it easier to be sure, and also if there's changes...

aneksteind: my primary concern is attributing lines here to lines of test code below. i still would consider interleaving them for that purpose because i think having them spatially local helps with that. testing that it's well formed rust seems orthogonal to that

pdg/src/info.rs
boyland-pf: I've done that, but it hasn't added any newlines, if you want me to add one I'm not sure where to put it.

aneksteind: it's fine, not the biggest deal

pdg/src/info.rs
```diff
@@ -28,7 +36,7 @@ pub struct FlowInfo {
 }
 
 impl FlowInfo {
-    //initializing flow information based on a node's kind
+    ///Initializes a [`FlowInfo`] based on a node's [`NodeKind`]
```
aneksteind, line 39:
```suggestion
    /// Initializes a [`FlowInfo`] based on a node's [`NodeKind`]
```

pdg/src/info.rs
```diff
@@ -67,18 +67,486 @@ fn set_flow_info(g: &mut Graph) {
...
+    n_id: &NodeId,
+    children: &HashMap<NodeId, Vec<NodeId>>,
+    descs: &HashMap<NodeId, NodeId>,
+) -> bool {
+    let mut max_descs: HashMap<Option<Field>, NodeId> = HashMap::new();
+    for id in children.get(n_id).unwrap() {
+        let sib_node: &Node = g.nodes.get(*id).unwrap();
+        let my_last_desc = descs.get(&id).unwrap().clone();
+        //if the first below matches, then two siblings, neither a field, conflict
+        //if the second matches, then two siblings of the same field conflict
```
aneksteind, line 122:
```suggestion
        // if the first below matches, then two siblings, neither a field, conflict
        // if the second matches, then two siblings of the same field conflict
```

pdg/src/info.rs
```diff
@@ -67,18 +67,486 @@ fn set_flow_info(g: &mut Graph) {
...
+    }
+    false
+}
+
+fn set_uniqueness(g: &mut Graph) {
+    let children = collect_children(g);
+    let last_descs = get_last_desc(g);
+    let mut nonuniqueness: HashSet<NodeId> = HashSet::new();
+    for (n_id, node) in g.nodes.iter_enumerated() {
+        if matches!(node.source, Some(p_id) if nonuniqueness.contains(&p_id))
```
aneksteind, line 145: I think we talked about adding comments to explain the relationship between uniqueness and parentage, etc.

pdg/src/info.rs
```diff
@@ -28,7 +38,7 @@ pub struct FlowInfo {
 }
 
 impl FlowInfo {
-    //initializing flow information based on a node's kind
+    /// Initializes a [`FlowInfo`] based on a node's [`NodeKind`]
```
kkysen, line 41:
```suggestion
    /// Initializes a [`FlowInfo`] based on a [`Node`]'s [`NodeKind`].
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        .nodes
+        .iter_enumerated()
+        .filter_map(|(idx, node)| Some((node.source?, idx)))
+    {
+        m.get_mut(&par).unwrap().push(chi);
+    }
+    m
+}
+
+/// Given a list of nodes of the same parent and information about them,
```
kkysen, line 109:
```suggestion
/// Given a list of [`Node`]s of the same parent and information about them,
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        assert!(!info(&pdg, b1).unique);
+        assert!(!info(&pdg, b2).unique);
+        assert!(!info(&pdg, b3).unique);
+        assert!(!info(&pdg, c1).unique);
+        assert!(!info(&pdg, c2).unique);
+        assert!(!info(&pdg, c3).unique);
+    }
+
+    #[test]
+    /// let mut a = Point {x: 0, y:0};
```
kkysen, line 385:
```suggestion
    /// let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        let mut g = Graph::default();
+        // A
+        // |-------
+        // |x     |
+        // B1     |y
+        // |      C1
+        // B2     |
+        //        C2
+        
+        // let mut a = Point {x: 0, y:0};
```
kkysen, line 400:
```suggestion
        // let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        let pdg = build_pdg(g);
+        assert!(info(&pdg, a).unique);
+        assert!(info(&pdg, b1).unique);
+        assert!(info(&pdg, b2).unique);
+        assert!(info(&pdg, c1).unique);
+        assert!(info(&pdg, c2).unique);
+    }
+
+    #[test]
+    /// let mut a = Point {x: 0, y:0};
```
kkysen, line 422:
```suggestion
    /// let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        // |-------   |
+        // |x     |   |
+        // B1     |x  |
+        // |      C1  |y
+        // |      |   |
+        // B2     |   |
+        //        C2  |
+        //            D1
+        
+        // let mut a = Point {x: 0, y:0};
```
kkysen, line 443:
```suggestion
        // let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        assert!(!info(&pdg, j).unique);
+        assert!(!info(&pdg, b1).unique);
+        assert!(!info(&pdg, b2).unique);
+        assert!(!info(&pdg, c1).unique);
+        assert!(!info(&pdg, c2).unique);
+        assert!(info(&pdg, d2).unique);
+    }
+
+    #[test]
+    /// let mut a = Point {x: 0, y:0};
```
kkysen, line 472:
```suggestion
    /// let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        let mut g = Graph::default();
+        // A
+        // |-------
+        // |      |
+        // B1     |y
+        // |      C1
+        // |      C2
+        // B2
+
+        // let mut a = Point {x: 0, y:0};
```
kkysen, line 487:
```suggestion
        // let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        let pdg = build_pdg(g);
+        assert!(!info(&pdg, a).unique);
+        assert!(!info(&pdg, b1).unique);
+        assert!(!info(&pdg, b2).unique);
+        assert!(!info(&pdg, c1).unique);
+        assert!(!info(&pdg, c2).unique);
+    }
+
+    #[test]
+    /// let mut a = Point {x: 0, y:0};
```
kkysen, line 508:
```suggestion
    /// let mut a = Point {x: 0, y: 0};
```

pdg/src/info.rs
```diff
@@ -67,18 +69,549 @@ fn set_flow_info(g: &mut Graph) {
...
+        let mut g = Graph::default();
+        // A
+        // |-------
+        // |      |
+        // B1     |y
+        // |      C1
+        // |y
+        // B2
+
+        // let mut a = Point {x: 0, y:0};
```
kkysen, line 524:
```suggestion
        // let mut a = Point {x: 0, y: 0};
```

aneksteind (approved):
this LGTM after @spernsteiner confirms that it's functionally correct

kkysen (dismissed):
I'd like to merge #666 into this before merging, as well as another PR for cleaning up some stuff.  I'll try to get that next PR done soon, so if I take longer, we can merge this PR and I'll open those subsequent PRs later.  #666 should merge first, though.

spernsteiner:
Sorry for the delay on this.  I poked at this code a bit and found some bugs.  I've pushed a commit with some additional tests to demonstrate.

I think the general issue is that `check_children_conflict` only looks at the immediate children of a node, but in fact it needs to look at entire chains of `Field` nodes to figure out what really conflicts.  `Field(0)` conflicts with `Field(0)`, but `Field(0) -> Field(0)` does not conflict with `Field(0) -> Field(1)`.  Concretely, you can have `let p = &mut x.0.0; let q = &mut x.0.1;` and overlapping uses of `p` and `q` do not cause a borrow checker error (the test `nested_field_no_conflict` shows an example of this).

boyland-pf:
Thanks, that example makes sense. Can you explain, though, why "offset_field_no_conflict" should actually find a conflict? Rust is willing to compile it, at least on my machine.

pdg/src/info.rs
kkysen: `par` is confusing.  Can we just call it `parent`?  `par` is more often `parallel`.  Same with `chi` and `child`.

pdg/src/info.rs
```diff
@@ -67,18 +67,940 @@ fn set_flow_info(g: &mut Graph) {
...
+        .nodes
+        .iter_enumerated()
+        .filter_map(|(idx, node)| Some((node.source?, idx)))
+    {
+        m.get_mut(&par).unwrap().push(chi);
+    }
+    m
+}
+
+/// Given a list of [`Node`]s of the same parent and information about them,
```
kkysen, line 107: Why do we need two passes over the nodes?

kkysen:
Can all the `HashMap<NodeId, T>`s be `IndexVec<NodeId, T>`s?

pdg/src/info.rs
kkysen: Hmm, with an `InddxVec` (basically just a `Vec`, the first pass should just be a `memset`, so I'll wait until we convert the `HashMap`s to `IndexVec`s.

boyland-pf:
On the subject of a[1].0 not being able to be live at the same time as a[1].1 (even though rustc allows it), I’m assuming that the modification needed to the algorithm output is that any descendent of an offset is required to not be live at the same time as any other descendent of an offset of the same variable.

If so, will push a fix in a couple hours both for this and the more obvious field one.

spernsteiner:
My original thinking was that offset nodes don't require any special handling - only fields require you to look through several levels before deciding whether there's a conflict.  But I haven't put much thought into the details, so that could be wrong.

boyland-pf:
The code should now be correct, though somewhat messier again. In particular, it is currently somewhat inefficient; however, the inefficiency is entirely within a hashmap that can be refactored into a trie, which would be efficient.

Some test cases were changed because the code that was in their comments had `let x = a.0;` as merely a field node instead of a field followed by a copy as it should be. I added a complement to the `nested_field_no_conflict` test case showing the case where storing an intermediate field in a variable changes the &mut-ability.

pdg/src/info.rs
```diff
@@ -786,6 +824,54 @@ mod test {
...
+    /// ```
+    ///
+    /// ```text
+    /// A--------
+    /// |       |
+    /// X1(f0)  Y1(f0)
+    /// X2      Y2
+    /// X3(f0)  Y3(f1)
+    /// X4      Y4
+    ///
```
kkysen, line 842:
```suggestion
    /// ```
```

pdg/src/info.rs
```diff
@@ -96,52 +97,84 @@ fn get_last_desc(g: &mut Graph) -> HashMap<NodeId, NodeId> {
...
+fn prefix(fs1: &Vec<Field>, fs2: &Vec<Field>) -> bool {
+    let len1 = fs1.len();
+    let len2 = fs2.len();
+    for i in 0..min(len1, len2) {
+        if fs1[len1 - 1 - i] != fs2[len2 - 1 - i] {
+            return false;
+        }
+    }
+    true
+}
```
kkysen, line 149: Is this checking if one of `fs1` and `fs2` is a prefix of the other?

boyland-pf: Yes, and that's a job only efficient to do with prefix trees or some other data structure intended for it.

kkysen: This doesn't seem very inefficient?  It should be able to be compiled to just a `memcmp`.  And if performance is important, we can use `tinyvec` for the `Vec<Field>`s and the `memcmp` should be short and cache-resident, usually.  There shouldn't be very many `Field`s in a `Vec<Field>`, right?

Also, I can make this more idiomatic.

boyland-pf: The inefficient part isn't this, but having to run this for every single child. Though I suppose this is only inefficient in theory, because maybe there won't ever be any nodes with more than a few dozen immediate children.

kkysen: Ah, got it.  Well, let's save any optimizations until later when we know it's actually a bottleneck, or if a better trie solution is simpler.

boyland-pf:
Outside of an efficiency issue that only appears if no maximum degree of field children is assumed, this appears done to me.

I would appreciate in particular my change to the `nested_field_no_conflict` to be looked at, standing in for the changes made throughout to bring the test pdg in line with what the pdg of the example rust code should actually be. The contrast between `nested_field_no_conflict` and `diff_field_conflict` may be interesting.

boyland-pf:
@spernsteiner Could you take a look at this?

spernsteiner:
>I would appreciate in particular my change to the `nested_field_no_conflict` test to be looked at, standing in for the changes made throughout to bring the test pdg in line with what the pdg of the example rust code should actually be.

The only change in `nested_field_no_conflict` is the addition of the two `Copy` nodes, right?  That seems fine.

>The contrast between `nested_field_no_conflict` and `diff_field_conflict` may be interesting.

That is interesting!  I hadn't considered that there would be a difference between `&mut a.0.0` and `&mut (*&mut a.0).0`.  I was thinking of `Copy` as a no-op that could potentially be elided if it didn't impact the graph structure, but I see now that's not the case.

spernsteiner (approved):
I haven't looked at the implementation, but the changes to the tests look good to me.

pdg/src/info.rs
```diff
@@ -67,18 +67,597 @@ fn set_flow_info(g: &mut Graph) {
...
+    }
+
+    /// ```rust
+    /// let mut a = Point { x: 0, y: 0 };
+    /// let j = &mut a;
+    /// let b = &mut j.x;
+    /// let c = &mut j.x;
+    /// *b = 1;
+    /// *c = 2;
+    /// *(a.y) = 3;
```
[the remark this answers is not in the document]
spernsteiner: There are similar issues on several other tests.  In general, accessing a field behind a reference should be either `(*a).y` or just `a.y`, rather than `*(a.y)` (which indicates dereferencing a reference stored in `a.y` instead).

Issue #665, opened 2022-09-15
Make `c2rust-pdg` also have a library target


kkysen:
`c2rust-pdg` is meant to eventually be used by `c2rust-analyze`, so it needs a library target (`lib.rs`).  Plus, this also will allow doc tests to work (they only run in library targets).  We'll keep the binary target, though.

PR #666, opened 2022-09-15
Fix a bunch of comments, mostly in the pdg info tests

kkysen:
Note that the doc tests don't actually run since `c2rust-pdg` only has a binary and not library target (see #665).


boyland-pf (approved):
Thanks for getting the one test I missed! I've looked through, it looks good.

PR #667, opened 2022-09-16
`check_unique` refactor

kkysen:
Refactored all the unique and non-unique `assert!`ions into `check_unique`, which is much more concise and avoids the ton of repetition there was previously.  It also adds more context to the error message when it fails.


spernsteiner:
One downside I see to this approach is that in some cases it forces the node names in the assertion to be out of order compared to the graph, which makes it slightly harder to match the two up.  I think we could get the best of both worlds with a macro along the lines of `assert_unique!(pdg, [a, !j, !b1, !b2, !c1, !c2, d2]);`, or something like this:
```Rust
let [a, j, b1, b2, c1, c2, d2] = unique_flags([a, j, b1, b2, c1, c2, d2]);
assert!(a && !j && !b1 && !b2 && !c1 && !c2 && d2);
```

kkysen:
> One downside I see to this approach is that in some cases it forces the node names in the assertion to be out of order compared to the graph, which makes it slightly harder to match the two up. I think we could get the best of both worlds with a macro along the lines of `assert_unique!(pdg, [a, !j, !b1, !b2, !c1, !c2, d2]);`, or something like this:
> 
> ```rust
> let [a, j, b1, b2, c1, c2, d2] = unique_flags([a, j, b1, b2, c1, c2, d2]);
> assert!(a && !j && !b1 && !b2 && !c1 && !c2 && d2);
> ```

See #668.  That solves this issue I think by putting the expected `unique` value where we create the `Node`.

pdg/src/info.rs
```diff
@@ -308,12 +316,7 @@ mod test {
         let b3 = mk_store_addr(&mut g, b1);
 
         let pdg = build_pdg(g);
-        assert!(!info(&pdg, a).unique);
-        assert!(!info(&pdg, b1).unique);
-        assert!(!info(&pdg, b2).unique);
-        assert!(!info(&pdg, b3).unique);
-        assert!(!info(&pdg, c1).unique);
-        assert!(!info(&pdg, c2).unique);
+        check_unique(&pdg, &[], &[a, b1, b2, b3, c1, c2]);
```
aneksteind, line 319: having these two arguments really doesn't make clear that you're checking for non-uniqueness here, in addition to the function name. i suggest having one function for each

aneksteind (changes requested):
see comment

pdg/src/info.rs
kkysen: Yeah, that might be better.  Might not be worth changing, though, given #668.

kkysen:
> I think we could get the best of both worlds with a macro along the lines of `assert_unique!(pdg, [a, !j, !b1, !b2, !c1, !c2, d2]);`
I don't think we should add a macro if we don't need to.

> or something like this:
> 
> ```rust
> let [a, j, b1, b2, c1, c2, d2] = unique_flags([a, j, b1, b2, c1, c2, d2]);
> assert!(a && !j && !b1 && !b2 && !c1 && !c2 && d2);
> ```

This wouldn't give good error messages, though.

spernsteiner:
This whole PR is superseded by the approach in #668, right?

kkysen:
> This whole PR is superseded by the approach in #668, right?

Yeah, pretty much.

PR #668, opened 2022-09-16
Refactor `check_unique` to `Checked{Graph,Node}`

kkysen:
This refactors `check_unique` to `Checked{Graph,Node}`.

`CheckedNode` stores not just the `NodeId` like as passed to the previous `check_unique`, but also if it should be unique (thus we ensure we don't forget to check any `Node`s (we did)), as well as the Rust source code statement for that `Node` as debug info.

Then `CheckedGraph` stores a `Graph` and a `Vec<CheckedNode>`.  When `.mk_*` methods are called on it, it adds the `CheckedNode` to the `Vec`, and then when `.check_unique()` is called at the end, it uses all these `CheckedNode`s.

The benefits of this are:
* The expected `unique` value is right next to the call that creates that `Node`
* All `Node`s have to be marked `unique` or not, so we don't forget any (we did)
* The Rust source code statements are passed as strings for debug info instead of stored as comments.  Thus, when an `assert!`ion fails and we print the `NodeId` and `Graphs`, we also print the Rust source code as the debug info (previously this was left blank).

### DIFF
--- a/pdg/src/info.rs
+++ b/pdg/src/info.rs
@@ -175,65 +175,97 @@ mod test {
     use c2rust_analysis_rt::mir_loc::Func;
     use rustc_middle::mir::Local;
 
-    fn mk_node(g: &mut Graph, kind: NodeKind, source: Option<NodeId>) -> NodeId {
-        g.nodes.push(Node {
-            function: Func {
-                def_path_hash: (1, 2).into(),
-                name: "fake_function".into(),
-            },
-            block: 0_u32.into(),
-            statement_idx: 0,
-            dest: None,
-            kind,
-            source,
-            info: None,
-            debug_info: "".into(),
-        })
+    #[derive(Debug)]
+    struct CheckedNode {
+        id: NodeId,
+        unique: bool,
+        stmt: &'static str,
     }
 
-    fn mk_addr_of_local(g: &mut Graph, local: impl Into<Local>) -> NodeId {
-        mk_node(g, NodeKind::AddrOfLocal(local.into()), None)
+    #[derive(Debug, Default)]
+    struct CheckedGraph {
+        graph: Graph,
+        checked_nodes: Vec<CheckedNode>,
     }
 
-    fn mk_copy(g: &mut Graph, source: NodeId) -> NodeId {
-        mk_node(g, NodeKind::Copy, Some(source))
-    }
-
-    fn mk_store_addr(g: &mut Graph, source: NodeId) -> NodeId {
-        mk_node(g, NodeKind::StoreAddr, Some(source))
-    }
-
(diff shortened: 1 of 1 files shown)


pdg/src/info.rs
```diff
@@ -259,25 +291,17 @@ mod test {
...
-        // *c = 0;
-        let c3 = mk_store_addr(&mut g, c1);
-
-        let pdg = build_pdg(g);
-        check_unique(&pdg, &[], &[a, b1, b2, b3, c1, c2, c3]);
+        let mut g = CheckedGraph::default();
+
+        let a = g.mk_addr_of_local("let mut a = 0", false, 0_u32);
+        let b1 = g.mk_copy("let b = &mut a", false, a);
+        g.mk_store_addr("*b = 0", false, b1);
```
spernsteiner, line 298: How about `let _b2 = g.mk_store_addr(...)`, so the name `b2` is visible here for matching with the graph above?

spernsteiner (approved):
Not sure this is really necessary, but it certainly doesn't hurt

pdg/src/info.rs
kkysen: Sure, I can add them back and just `_`-prepend them.

pdg/src/info.rs
```diff
@@ -175,65 +175,97 @@ mod test {
...
+    struct CheckedGraph {
+        graph: Graph,
+        checked_nodes: Vec<CheckedNode>,
+    }
+
+    impl CheckedGraph {
+        fn mk_node(
+            &mut self,
+            stmt: &'static str,
+            unique: bool,
```
aneksteind, line 195: nit pick: i think i'd rather this be an enum than a bool

aneksteind: optional change, just my two cents

kkysen: What's the advantage of an `enum`?  If we're calling them `Unique` and `NonUnique`, I'm not sure there's an advantage over a `bool`.  One is just non the other.  Plus, then we should just change the `unique` field in `NodeInfo::unique` to the same `enum`.

aneksteind: > Plus, then we should just change the unique field in NodeInfo::unique to the same enum.

I agree that should be done. My preference is because I don't like relying on IDEs to tell me what the bool represents. A type `Unique` or `NonUnique` is very clear, despite it being semantically the same as a `bool`. In cases where functions take more than one bool (not the case here), it prevents mixing argument order by making things type safe.

kkysen: Okay, that makes sense.  Plus we could change it to `Unique` and `Aliases` if we want, too.  I do think we should save that for a later PR, though.

aneksteind: that sounds okay

PR #670, opened 2022-09-16
Refactor `fn check_children_conflicts`

kkysen:
See the individual commits for the series of refactors.

### DIFF
--- a/pdg/src/info.rs
+++ b/pdg/src/info.rs
@@ -125,11 +125,12 @@ fn check_children_conflict(
         {
             return true;
         }
-        let my_entry: Entry<_, _> = if let NodeKind::Field(f) = sib_node.kind {
-            max_descs.entry(Some(f))
+        let my_key = if let NodeKind::Field(f) = sib_node.kind {
+            Some(f)
         } else {
-            max_descs.entry(None)
+            None
         };
+        let my_entry: Entry<_, _> = max_descs.entry(my_key);
         my_entry
             .and_modify(|past_last_desc| *past_last_desc = max(*past_last_desc, my_last_desc))
             .or_insert(my_last_desc);
--- a/pdg/src/info.rs
+++ b/pdg/src/info.rs
@@ -2,7 +2,6 @@ use crate::graph::{Graph, Node, NodeId, NodeKind};
 use crate::Graphs;
 use rustc_middle::mir::Field;
 use std::cmp::max;
-use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{self, Debug, Display, Formatter};
 
@@ -125,13 +124,12 @@ fn check_children_conflict(
         {
             return true;
         }
-        let my_key = if let NodeKind::Field(f) = sib_node.kind {
-            Some(f)
-        } else {
-            None
-        };
-        let my_entry: Entry<_, _> = max_descs.entry(my_key);
-        my_entry
+        max_descs
+            .entry(if let NodeKind::Field(f) = sib_node.kind {
+                Some(f)
+            } else {
+                None
+            })
             .and_modify(|past_last_desc| *past_last_desc = max(*past_last_desc, my_last_desc))
             .or_insert(my_last_desc);
     }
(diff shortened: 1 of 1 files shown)


pdg/src/info.rs
```diff
@@ -1,11 +1,12 @@
 use crate::graph::{Graph, Node, NodeId, NodeKind};
 use crate::Graphs;
-use rustc_middle::mir::Field;
 use std::cmp::max;
-use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{self, Debug, Display, Formatter};
 
+/// Force an import of [`Node`] just for docs.
+const _: Option<Node> = None;
```
aneksteind, line 8: this can't be resolved in the docstring itself?

pdg/src/info.rs
```diff
@@ -114,25 +115,26 @@ fn check_children_conflict(
...
-    for id in children.get(n_id).unwrap() {
-        let sib_node: &Node = g.nodes.get(*id).unwrap();
-        let my_last_desc = *descs.get(id).unwrap();
-        // if the first below matches, then two siblings, neither a field, conflict
-        // if the second matches, then two siblings of the same field conflict
-        if matches!(max_descs.get(&None), Some(max_desc) if max_desc > id)
-            || matches!(sib_node.kind, NodeKind::Field(f) if matches!(max_descs.get(&Some(f)), Some(max_desc_field) if max_desc_field > id))
-        {
+    let mut max_descs = HashMap::new();
+    for id in &children[n_id] {
```
aneksteind, line 119: in another PR you made `id` clearer by renaming to `child_id`. I think similar changes should happen here

pdg/src/info.rs
kkysen: Yeah, but then we have to add a `[`Node`]: crate::graph::Node` to almost every doc comment.  This does it globally.  The other way works fine, too, but it just adds a ton of extra lines vs. this one line.

pdg/src/info.rs
kkysen: Okay, I'll try to make them all consistent.  I've been using `child` and `parent` for the `NodeId`s, and `child_node` for the `Node` (though there are a couple of exceptions, which I'll fix).  Using the `NodeId` is much more common so I'm using the shorter name for them.

aneksteind (changes requested):
i'm not sure this refactor should go in given the comments at https://github.com/immunant/c2rust/pull/660#issuecomment-1248764136. let me know if you disagree

kkysen:
> i'm not sure this refactor should go in given the comments at [#660 (comment)](https://github.com/immunant/c2rust/pull/660#issuecomment-1248764136). let me know if you disagree

I think it's still worth it.  It'll help keep the style better when `check_children_conflicts` is fixed.

aneksteind:
I think you should check with @boyland-pf to make sure he hasn't already started making changes. he'd have to rebase, that takes time, and i think it's important that his feature lands before his last day

kkysen:
Okay, that's a good point.  Let me check.

pdg/src/info.rs
aneksteind: is there a module-level doc that it could be snuck into, thereby making it available elsewhere, or is that not how rustdoc works?

kkysen: Yeah, unfortunately, `rustdoc` doesn't work that way.  I've tried `#[cfg(doc)] use ...` in the past for doc-only imports, but it didn't work either.  Then I just thought of this as another workaround.

boyland-pf:
This should be good.

kkysen:
> This should be good.

Okay, great!

pdg/src/info.rs
kkysen: Made consistent in:
* https://github.com/immunant/c2rust/pull/669/commits/a1dfd02f40a1c501a9342460914aa96cf5c7d728
* https://github.com/immunant/c2rust/pull/671/commits/ec0e70a9285f6c6d452e03b18bbc2a4e0dc27d2f

PR #674, opened 2022-09-21
Add python `compiledb` example

### LINKED ISSUE #659: Documentation request: Add python `compiledb` as a way to generate `compile-commands.json` from a `Makefile`
I was using c2rust and I was trying to translate this project called [light](https://github.com/haikarainen/light).
~~None of the methods from [the docs](https://c2rust.com/manual/quickstart.html?highlight=intercept#-with--intercept-build) to generate a compile-commands.json worked.~~ 
Turns out the other tools are just a bit more finicky (at least for the project I was working on).

[`compiledb`](https://pypi.org/project/compiledb/) works great.
All I had to do was `compiledb make` (after running `./autogen.sh && ./configure --with-udev` of course).

Could you please add this to the docs?

wildwestrom:
Fixes #659.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,24 @@ it can be used similarly to `intercept-build`:
 bear <build command>
 ```
 
+#### ... with `compiledb`
+
+The `compiledb` package can also be used for makefile projects if the other tools don't work.
+Install via `pip` with:
+
+```sh
+pip install compiledb
+```
+
+Usage:
+
+```sh
+# After running
+./autogen.sh && ./configure # etc.
+# Run
+compiledb make
+```
+
 <!-- ANCHOR_END: translating-c-to-rust -->
 
 ## Contact


thedataking (approved):
LGTM, thanks!

kkysen (changes requested):
I think we should clarify that `compiledb` works without having to recompile everything, which `bear` and `intercept-build` require.

thedataking:
> I think we should clarify that `compiledb` works without having to recompile everything, which `bear` and `intercept-build` require.

How the tools work isn't terribly important. `bear` and `intercept-build` do not require recompilation. They require a full (non-incremental) build.

kkysen:
> How the tools work isn't terribly important. `bear` and `intercept-build` do not require recompilation. They require a full (non-incremental) build.

That's what I mean.  I don't think `compiledb` requires a full build.

That's where this issue came from.  `bear` was being used in an incremental build so it wasn't working, but `compiledb` did.

thedataking:
> That's what I mean. I don't think compiledb requires a full build.

Correct. However, it is incumbent on the user to learn how to use each tool using the relevant documentation. Our README simply provides pointers to each tool that may or may not be useful.

kkysen:
> Correct. However, it is incumbent on the user to learn how to use each tool using the relevant documentation. Our README simply provides pointers to each tool that may or may not be useful.

That's true, but they often don't, and a lot of the opened issues end up being about not able to create the compilation database correctly.  An extra sentence of clarification won't hurt.

PR #677, opened 2022-09-23
Add support for LLVM 15

### LINKED ISSUE #676: `c2rust` fails to build with LLVM 15
```shell
 --- stderr
  CMake Warning:
    Manually-specified variables were not used by the project:

      CMAKE_ASM_COMPILER
      CMAKE_ASM_FLAGS


  * Closing connection 0
  * Closing connection 1
  /Users/jrmuizel/.cargo/git/checkouts/c2rust-2a3ef4c32beab81a/1fddfc4/c2rust-ast-exporter/src/AstExporter.cpp:2270:52: error: no member named 'Ascii' in 'clang::StringLiteral::StringKind'
              case clang::StringLiteral::StringKind::Ascii:
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
  1 error generated.
  make[3]: *** [CMakeFiles/clangAstExporter.dir/AstExporter.cpp.o] Error 1
  make[2]: *** [CMakeFiles/clangAstExporter.dir/all] Error 2
  make[1]: *** [CMakeFiles/clangAstExporter.dir/rule] Error 2
```

thedataking:
`clang::StringLiteral::StringKind::Ascii` was renamed to `clang::StringLiteral::StringKind::Ordinary`. Add release keys and update `build_translator.py`. Closes #676.


chrysn:
<details><summary>Unrelated trouble with earlier Debian versions that's not this PR's fault</summary>

Testers on not-perfectly-recent Debian beware that the [change to LLVM 15 that requires this](https://reviews.llvm.org/D128762) (the StringLiteral::StringKind renaming) came in quite late in the LLVM release process. Users of Debian experimental will find that the new version will *not* build on some versions of llvm-15-dev, but that the current master does work. This is no fault of this PR, but of the Debian package claiming something to be -15 when it doesn't have all of -15's changes (but to be fair, it *is* an experimental pre-release, and needs to be opted in to).

</details>

With the very latest experimental release (llvm-15 1:15.0.1-1~exp2 from Debian, including mlir-15-tools, and a bunch of packages the build process will conveniently ask for), this passes compilation, but fails at link time:

```
  = note: /usr/bin/ld: /usr/lib/llvm-15/lib/libclangSema.a(SemaRISCVVectorLookup.cpp.o): in function `clang::CreateRISCVIntrinsicManager(clang::Sema&)':
          (.text._ZN5clang27CreateRISCVIntrinsicManagerERNS_4SemaE+0x1b8): undefined reference to `clang::RISCV::RVVIntrinsic::computeBuiltinTypes(llvm::ArrayRef<clang::RISCV::PrototypeDescriptor>, bool, bool, bool, unsigned int)'
          /usr/bin/ld: (.text._ZN5clang27CreateRISCVIntrinsicManagerERNS_4SemaE+0x1f9): undefined reference to `clang::RISCV::RVVIntrinsic::computeBuiltinTypes(llvm::ArrayRef<clang::RISCV::PrototypeDescriptor>, bool, bool, bool, unsigned int)'
```

I think this would be resolved by not only linking in libclangSema.a but also libclangSupport.a, but I don't know where that's stated (yet?).

thedataking:
Thanks @chrysn, I've incorporated your suggestion. Let me know if you hit any other issues.

c2rust-ast-exporter/build.rs
```diff
@@ -181,7 +184,14 @@ fn build_native(llvm_info: &LLVMInfo) {
             "clangRewrite",
             "clangLex",
             "clangBasic",
-        ] {
+        ];
+        if llvm_info.llvm_major_version >= 15 {
+            // insert after clangSema
```
kkysen, line 189: Is it order sensitive?

chrysn: Generally, library sequences do matter to linkers, even though modern linkers sometimes tolerate arbitrary sequences. It might not in this special case, but would take some effort to find out, and more to be sure it stays that way. https://www.linuxtopia.org/online_books/an_introduction_to_gcc/gccintro_18.html has a good summary

thedataking: Thanks for weighing in @chrysn! I thought I had provided an answer earlier (that order indeed matters since linkers resolve symbols in libraries from left to right) but I must have messed up somehow since I don't see my reply.

scripts/common.py
```diff
@@ -109,16 +109,24 @@ class Config:
     Reflect changes to all configuration variables that depend on LLVM_VER
     """
     def _init_llvm_ver_deps(self):
+        def llvm_major_ver_ge(test_ver) -> bool:
+            (major, _, _) = self.LLVM_VER.split(".")
+            return int(major) >= test_ver
```
kkysen, line 114: I think `llvm_major_ver() -> int` would be better, and do the `>=` at call site.  It's much easier to read and more flexible, and `test_ver` isn't otherwise used in `llvm_major_ver_ge`.

c2rust-ast-exporter/build.rs
```diff
@@ -181,7 +189,14 @@ fn build_native(llvm_info: &LLVMInfo) {
...
+            let sema_pos = clang_libs.iter().position(|&r| r == "clangSema").unwrap();
+            clang_libs.insert(sema_pos + 1, "clangSupport");
+        }
```
kkysen, line 197: I would put this inline as something like `if llvm_info.llvm_major_version >= 15 { "clangSupport" } else { "" }` and then filter out `""`.  It's much cleaner, easier to understand the order of libraries, and more extensible if we need to add conditional libraries where order matters.

c2rust-ast-exporter/build.rs
```diff
@@ -151,7 +151,15 @@ fn build_native(llvm_info: &LLVMInfo) {
...
     let use_libclang = if cfg!(target_os = "macos") {
-        false
+        // We hit an issue linking against the shared libraries for the homebrew
+        // version of LLVM 15 because they use a feature (opaque pointers) which
+        // are not understood by earlier versions of LLVM so we link against
+        // libclang unless static linking has been explicitly requested.
+        if cfg!(feature = "llvm-static") {
+            false
+        } else {
+            true
```
kkysen, line 161: Do we not need to do a similar `libclang_path.exists()` check for macOS like we do for linux?

PR #683, opened 2022-09-25
Fix UB in `analysis/test`

### LINKED ISSUE #680: Fix UB in `analysis/test`
There's a bunch of UB in `analysis/test`, which causes flakiness in the pdg snapshot test.  We should fix all of the UB detected by `cargo miri run`, though there may be some more as well (it remains flaky in CI).

#650 is a WIP version that fixes the UB and most non-intentional memory leaks.

kkysen:
Fixes #680.

See the commits for the individual fixes.  The fixes are:
* 2 use after frees: 429962d193b81ccc638924657bbfea66a9fabea9, 95e483d77d1d293daf1e01760b6bf423b4492041
* 2 uninitialized reads: 50b1bb4c9b0d322a73c75015336a50abd3f5fe4d, f6e5bce29673ff7170a9cec704fcbce925a32f2a
* 2 nullptr derefs: f96bcb23dcd2a5cc42e1d5948399203d10293325, 69de3c3145ec2bb8d1f4dff9e6dc1a3eb94d73e8

Now `cd analysis/test && cargo miri run` reports no UB.  However, `miri` can't run variadic functions like `printf`, so running this requires the monomorphic `printf` shim in #686.

### DIFF
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -417,18 +417,7 @@ pub unsafe extern "C" fn test_realloc_fresh() {
 }
 #[no_mangle]
 pub unsafe extern "C" fn test_load_addr() {
-    let s = malloc(::std::mem::size_of::<S>() as libc::c_ulong) as *mut S;
-    *s = S {
-        field: 0i32,
-        field2: 0u64,
-        field3: 0 as *const S,
-        field4: T {
-            field: 0i32,
-            field2: 0u64,
-            field3: 0 as *const S,
-            field4: 0i32,
-        },
-    };
+    let s = calloc(1, ::std::mem::size_of::<S>() as libc::c_ulong) as *mut S;
     let x = (*s);
     free(s as *mut libc::c_void);
 }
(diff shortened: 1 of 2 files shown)


aneksteind:
Can you remind me whether or not this UB is the fault of the transpiler? This is where much of the code in the test file has come from. It would be great to treat the issue at the source, if so.

aneksteind (approved):

kkysen:
Yeah, none of the UB fixes here are in functions that are in `pointers.c`.  it's pretty out-of-date.

aneksteind:
Ok, that sounds fine then

kkysen:
There was one transpiled leak in `main` for arg conversion, but I already opened an issue for that in #649.  It was fixed here manually in #653.

analysis/test/src/pointers.rs
```diff
@@ -413,6 +413,17 @@ pub unsafe extern "C" fn test_realloc_fresh() {
...
+        field: 0i32,
+        field2: 0u64,
+        field3: 0 as *const S,
+        field4: T {
+            field: 0i32,
+            field2: 0u64,
+            field3: 0 as *const S,
+            field4: 0i32,
+        },
+    };
```
kkysen: See comment by @aneksteind: https://github.com/immunant/c2rust/commit/f6e5bce29673ff7170a9cec704fcbce925a32f2a#r85154562.

kkysen: `calloc` works, so fixed in https://github.com/immunant/c2rust/pull/683/commits/f21764b3b4a82cbf9da04cc088cba6a5a6b2a5c2.

PR #684, opened 2022-09-25
Fix memory leaks in `analysis/test`

### LINKED ISSUE #681: Fix non-intentional memory leaks in `analysis/test`
This is related to #680 (fixing UB).  There are also a bunch of non-intentional memory leaks in `analysis/test`.  These are not UB, so they shouldn't be causing any unexpected behavior, but there's no good reason to have them, and they clutter up the `miri` output when looking for UB.  We should only keep the intentional leaks that are there for testing how we handle leaks.

kkysen:
Fixes #681.

This fixes all of the memory leaks in `analysis/test`.

There was one that was intentional (`no_owner(0)`), but I freed that in `main_0` immediately after the `no_owner(0)` call instead.

### DIFF
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -516,6 +516,7 @@ unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> lib
     analysis2();
     inter_function_analysis();
     no_owner(0i32);
+    free(global as *mut libc::c_void);
     no_owner(1i32);
     invalid();
     testing();
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -179,6 +179,7 @@ pub unsafe extern "C" fn connection_accepted(
 unsafe extern "C" fn connection_close(mut srv: *mut server, mut con: *mut connection) {
     fdevent_fdnode_event_del((*srv).ev, (*con).fdn);
     fdevent_unregister((*srv).ev, (*con).fd);
+    free(con as *mut libc::c_void);
 }
 #[no_mangle]
 pub unsafe extern "C" fn fdevent_fdnode_event_del(mut ev: *mut fdevents, mut fdn: *mut fdnode) {
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -418,9 +418,11 @@ pub unsafe extern "C" fn test_load_addr() {
 #[no_mangle]
 pub unsafe extern "C" fn test_overwrite() {
     let mut s = malloc(::std::mem::size_of::<S>() as libc::c_ulong);
+    let s2 = s;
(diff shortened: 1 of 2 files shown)


analysis/test/src/pointers.rs
```diff
@@ -508,6 +516,7 @@ unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> lib
     analysis2();
     inter_function_analysis();
     no_owner(0i32);
+    free(global as *mut libc::c_void);
```
kkysen, line 519: The leak in `no_owner` when called with `no_owner(0)` is intentional.  Should it be freed in `main_0`, or is it intentional to completely leak it?

analysis/test/src/pointers.rs
```diff
@@ -340,6 +343,7 @@ pub unsafe extern "C" fn test_arg() {
     let mut s = malloc(::std::mem::size_of::<S>() as libc::c_ulong);
     foo(s);
     let t = s;
+    free(s as *mut libc::c_void);
```
kkysen, line 346: Should `t` be freed instead of `s`?

analysis/test/src/pointers.rs
```diff
@@ -355,6 +359,7 @@ pub unsafe extern "C" fn foo_rec(n: i32, bar: *mut libc::c_void) -> *mut libc::c
 pub unsafe extern "C" fn test_arg_rec() {
     let mut s = malloc(::std::mem::size_of::<S>() as libc::c_ulong);
     let t = foo_rec(3, s);
+    free(s as *mut libc::c_void);
```
kkysen, line 362: Should `t` be freed instead of `s`?

analysis/test/src/pointers.rs
```diff
@@ -79,6 +80,7 @@ pub unsafe extern "C" fn simple() {
     recur(3, x);
     let s = *y;
     *x = s;
+    free(x2 as *mut libc::c_void);
```
aneksteind, line 83: why is this necessary as opposed to freeing `x`?

analysis/test/src/pointers.rs
aneksteind: if they're the same address, does it make a difference?

analysis/test/src/pointers.rs
aneksteind: oh, i see, because x is reassigned

analysis/test/src/pointers.rs
aneksteind: i think what you have here is fine

analysis/test/src/pointers.rs
aneksteind: there was no particular intention, so this is fine

analysis/test/src/pointers.rs
aneksteind: this is fine

aneksteind (approved):

PR #686, opened 2022-09-25
Make `analysis/test` miri-compatible and add a `miri` test for it

### LINKED ISSUE #680: Fix UB in `analysis/test`

### LINKED ISSUE #681: Fix non-intentional memory leaks in `analysis/test`

### LINKED ISSUE #682: Fix int-to-ptr casts in `analysis/test`
`analysis/test` contains many int-to-ptr casts for null pointers, like `0 as *const _` and `0 as *mut _`.  These use `ptr::from_exposed_addr` as part of strict provenance.  Since we don't actually need to use `ptr::from_exposed_addr` here, we should switch to using `str::ptr::null{,_mut}()` instead.  `miri` warns on these, so it's easier to find and fix UB bugs if these aren't in the way, and it's best practice to do this as well.

See the `miri` warning:
```shell
warning: integer-to-pointer cast
   --> src/pointers.rs:82:17
    |
82  |         field3: 0 as *const S,
    |                 ^^^^^^^^^^^^^ integer-to-pointer cast
    |
    = help: This program is using integer-to-pointer casts or (equivalently) `ptr::from_exposed_addr`,
    = help: which means that Miri might miss pointer bugs in this program.
    = help: See https://doc.rust-lang.org/nightly/std/ptr/fn.from_exposed_addr.html for more details on that operation.
    = help: To ensure that Miri does not miss bugs in your program, use Strict Provenance APIs (https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance, https://crates.io/crates/sptr) instead.
    = help: You can then pass the `-Zmiri-strict-provenance` flag to Miri, to ensure you are not relying on `from_exposed_addr` semantics.
    = help: Alternatively, the `-Zmiri-permissive-provenance` flag disables this warning.
    = note: backtrace:
    = note: inside `pointers::simple` at src/pointers.rs:82:17
```

### LINKED ISSUE #685: Add a `miri` test for `analysis/test`
We should add a `miri` (`cargo miri run`) test for `analysis/test` to ensure we avoid re-introducing any UB into it.

The only change we need to add for this is adding a monomorphic `printf` shim since `miri` doesn't work with variadic functions.  This should be feature-gated, since a monomorphic `printf` shim isn't generic enough for general use, so we only want it for `miri` testing.

kkysen:
Fixes #685.

This makes `analysis/test` `miri`-compatible when running with `--features miri`.  This is done by using a monomorphic `printf` shim, since `miri` can't handle variadic functions like `printf`.  Since all uses of `printf` in `analysis/test` are monomorphic (they all have the same format string), we can substitute a `fn printf` that is non-`extern "C"`, non-variadic (and thus `miri`-compatible), and that still has the same behavior for its call sites.

Then we add a test in `c2rust-pdg` that runs `miri` on `analysis/test` to ensure it stays UB-free.

However, we don't yet run this test by default (it's `#[ignore]`d for now) as there are issues with running `miri` in CI (it installs `xargo` every time and I'm getting a permission denied error (not sure from quite what exactly), and it'd be better to install `xargo` upfront, not on every run).  Thus, I'm `#[ignore]`ing it for now in https://github.com/immunant/c2rust/pull/686/commits/4152d34d1f7140f28b346f4be52f0500aa7d7bc4.  We can get it to run in CI correctly later in another PR, but I want to merge this now and avoid over-complicating it here.  The test can still be manually run with `cargo test -p c2rust-pdg -- --ignored analysis_test_miri`.  See #698 for the tracking issue to re-enable this test by default.

The new test passing is blocked on:
* #683 (f ixes #680)
* #684 (f ixes #681)

It would also be nice to f ix #682, but that's not completely necessary for this (though it would create a much less noisy output).

### DIFF
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -21,6 +21,18 @@ extern "C" {
     fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
 }
 
+/// `miri` does not support calling variadic functions like [`printf`],
+/// but we want to test for UB, leaks, etc. using `cargo miri run`.
+///
+/// Luckily, all [`printf`] calls in this module are monomorphic,
+/// in that they all have the same format string and same call signature,
+/// so we can replace it with a [`printf`] shim that preserves the behavior
+/// only for the exact monomorphic usages in this module.
+///
+/// Note that there is no way to detect `miri` is running,
+/// so we have to put this under a separate `miri` feature
+/// that should be enabled when testing under `miri` with
+/// `cargo miri run --features miri`.
 #[cfg(feature = "miri")]
 fn printf(fmt: *const libc::c_char, i: i32) -> libc::c_int {
     use std::ffi::CStr;
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -14,7 +14,14 @@ extern "C" {
     fn calloc(_: libc::c_ulong, _: libc::c_ulong) -> *mut libc::c_void;
     fn realloc(_: *mut libc::c_void, _: libc::c_ulong) -> *mut libc::c_void;
     fn free(__ptr: *mut libc::c_void);
-    fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
+}
+
+fn printf(fmt: *const libc::c_char, i: i32) -> libc::c_int {
+    use std::ffi::CStr;
+    assert_eq!(unsafe { CStr::from_ptr(fmt) }, CStr::from_bytes_with_nul(b"%i\n\x00").unwrap());
+    let s = format!("{i}\n");
+    print!("{s}");
+    s.len() as libc::c_int
 }
 
 /// Hidden from instrumentation so that we can polyfill [`reallocarray`] with it.
--- a/analysis/test/src/pointers.rs
+++ b/analysis/test/src/pointers.rs
@@ -16,6 +16,12 @@ extern "C" {
     fn free(__ptr: *mut libc::c_void);
 }
 
+#[cfg(not(feature = "miri"))]
+extern "C" {
+    fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
+}
+
+#[cfg(feature = "miri")]
 fn printf(fmt: *const libc::c_char, i: i32) -> libc::c_int {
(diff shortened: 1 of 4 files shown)


analysis/test/src/pointers.rs
```diff
@@ -14,9 +14,22 @@ extern "C" {
...
     fn free(__ptr: *mut libc::c_void);
+}
+
+#[cfg(not(feature = "miri"))]
+extern "C" {
     fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
 }
 
+#[cfg(feature = "miri")]
+fn printf(fmt: *const libc::c_char, i: i32) -> libc::c_int {
```
aneksteind, line 25: please add a comment here similar to the PR description explaining why this is done

aneksteind (changes requested):

analysis/test/src/pointers.rs
kkysen: Fixed in https://github.com/immunant/c2rust/pull/686/commits/c02f3e4f0e9170de80facee19cf3a960e7a27ab8.

aneksteind (approved):

kkysen:
There are issues with running `miri` in CI (it installs `xargo` every time and I'm getting a permission denied error (not sure from quite what exactly), and it'd be better to install `xargo` upfront, not on every run).  Thus, I'm `#[ignore]`ing it for now in https://github.com/immunant/c2rust/pull/686/commits/4152d34d1f7140f28b346f4be52f0500aa7d7bc4.  We can get it to run in CI correctly later in another PR, but I want to merge this now and avoid overcomplicating it here.  The test can still be manually run with `cargo test -p c2rust-pdg -- --ignored analysis_test_miri`.

Issue #613, opened 2022-08-15
Flaky CI in `cargo test -p c2rust-pdg` (snapshot)


kkysen:
The `cargo test` step of CI is occasionally flaky due to the snapshot test in `cargo test -p c2rust-pdg`.  Usually it can be resolved by just re-running CI.

The error shown is a difference in the snapshot, but I believe it's caused by the instrumented binary being terminated early (and silently), as the difference is always purely in the trace being cut off early.

kkysen:
I believe this is caused by UB in `analysis/test`.  When I reduced `main_0` to just

```rust
unsafe fn main_0(mut argc: libc::c_int, mut argv: *mut *mut libc::c_char) -> libc::c_int {
    simple();
    exercise_allocator();
    simple_analysis();
}
```

it was reliably crashing the snapshot tests on every other run or so, locally, not just in CI.

When I fixed the UB in that (checked with `cargo miri run`), the crashes went away.

For example, there is a lot of UB in `simple`.

kkysen:
The significantly rarer race condition in `cargo add` is a separate bug.  It is much more rare, though, and we should be able to fix it by adding locking at some level (perhaps in the test code, and don't run `cargo add` in `c2rust-instrument`).

kkysen:
Fixed by:
* #683
* #691

Issue #687, opened 2022-09-25
Add a single-threaded instrumentation runtime


kkysen:
We should add a single-threaded instrumentation runtime.  This could help with various things, such as:
* simplifying the logic and avoiding threading, which helps diagnose bugs like #613
* allowing instrumented code to run in single-threaded environments, which could perhaps be useful in the future for constrained programs

I am thinking we can have multiple runtimes that can be dynamically selected through `$INSTRUMENT_RUNTIME`, similar to how `$INSTRUMENT_BACKEND` works at the moment.

PR #691, opened 2022-09-26
Flush the `Backend` buffer

### LINKED ISSUE #689: Flush the `LogBackend` `BufWriter`
Normally the `BufWriter` in `LogBackend` is `flush`ed in `drop`, but since all of this runs from a static or a background thread, no destructors are run (reliably at least; not sure about the background thread case).  Thus, we need to explicitly `flush` the buffer.  I believe this may be the remaining cause of #613.

I discovered this while implementing #687, as the missing `flush` is reliably detected when running the `Backend` on the main thread.

kkysen:
Fixes #689.

Normally the `BufWriter` in `LogBackend` is `flush`ed in `drop`, but since all of this runs from a static or a background thread, no destructors are run (reliably at least; not sure about the background thread case).  Thus, we need to explicitly `flush` the buffer.  I believe this may be the remaining cause of #613.

This fix adds a `flush` method to `trait WriteEvent`, which is called after all `Event`s have been written so that we never lose any of the `Event`s.  It's implemented by delegating to `Write::flush`, both for `BufWriter` in `LogBackend` and `stderr()` in `DebugBackend` (though that one is usually unbuffered).

I discovered this while implementing #687, as the missing `flush` is reliably detected when running the `Backend` on the main thread.

### DIFF
--- a/analysis/runtime/src/runtime/backend.rs
+++ b/analysis/runtime/src/runtime/backend.rs
@@ -1,7 +1,7 @@
 use enum_dispatch::enum_dispatch;
 use fs_err::{File, OpenOptions};
 use std::fmt::Debug;
-use std::io::{BufWriter, Write};
+use std::io::{stderr, BufWriter, Write};
 use std::sync::mpsc::Receiver;
 
 use bincode;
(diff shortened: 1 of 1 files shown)


aneksteind (approved):

kkysen:
I believe this (and #683) have finally fixed #613.  I haven't noticed anymore flaky CI failures since.  @aneksteind, have you noticed any?  If not, I'll finally close that issue.

Issue #692, opened 2022-09-27
RISC-V vector instructions pulled in through LLVM cause build failures


chrysn:
Minimal example:

```shell
$ mkdir /tmp/test
$ cd /tmp/test
$ touch empty.h
$ cat > compile_commands <<EOF
[
  {
    "arguments": [
      "any-cc",
      "-target",
      "riscv32",
      "empty.h"
    ],
    "directory": "/tmp/test",
    "file": "empty.h"
  }
]
EOF
$ c2rust transpile compile_commands.json
warning: c2rust: Encountered unsupported BuiltinType kind 99 for type __rvv_int8mf8_t
warning: c2rust: Encountered unsupported BuiltinType kind 100 for type __rvv_int8mf4_t
warning: c2rust: Encountered unsupported BuiltinType kind 101 for type __rvv_int8mf2_t
warning: c2rust: Encountered unsupported BuiltinType kind 102 for type __rvv_int8m1_t
warning: c2rust: Encountered unsupported BuiltinType kind 103 for type __rvv_int8m2_t
warning: c2rust: Encountered unsupported BuiltinType kind 104 for type __rvv_int8m4_t
warning: c2rust: Encountered unsupported BuiltinType kind 105 for type __rvv_int8m8_t
warning: c2rust: Encountered unsupported BuiltinType kind 106 for type __rvv_uint8mf8_t
warning: c2rust: Encountered unsupported BuiltinType kind 107 for type __rvv_uint8mf4_t
warning: c2rust: Encountered unsupported BuiltinType kind 108 for type __rvv_uint8mf2_t
warning: c2rust: Encountered unsupported BuiltinType kind 109 for type __rvv_uint8m1_t
warning: c2rust: Encountered unsupported BuiltinType kind 110 for type __rvv_uint8m2_t
warning: c2rust: Encountered unsupported BuiltinType kind 111 for type __rvv_uint8m4_t
warning: c2rust: Encountered unsupported BuiltinType kind 112 for type __rvv_uint8m8_t
warning: c2rust: Encountered unsupported BuiltinType kind 113 for type __rvv_int16mf4_t
warning: c2rust: Encountered unsupported BuiltinType kind 114 for type __rvv_int16mf2_t
warning: c2rust: Encountered unsupported BuiltinType kind 115 for type __rvv_int16m1_t
warning: c2rust: Encountered unsupported BuiltinType kind 116 for type __rvv_int16m2_t
[...]

PR #693, opened 2022-09-27
Tolerate RISC-V vector types

chrysn:
Closes: #692

Note that I have no clue how these vector types would be used -- I don't even have a processor that supports them, but their mere presence in LLVM's builtins caused the above-mentioned issue.

This follows along the paths of existing vector types, reusing the code already established for SVE. As this produces a `Float16` rather than a `Half` builtin down the road, support for that is added by mapping to Half. That seems to be OK because, unlike C's `int` size mess, `half` is commonly understood to mean an IEEE 16-bit float.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -400,6 +400,9 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
             case BuiltinType::UInt: return TagUInt;
             case BuiltinType::ULong: return TagULong;
             case BuiltinType::ULongLong: return TagULongLong;
+            // Constructed as a consequence of the conversion of
+            // built-in to normal vector types.
+            case BuiltinType::Float16: return TagHalf;
             case BuiltinType::Half: return TagHalf;     
             #if CLANG_VERSION_MAJOR >= 11
             case BuiltinType::BFloat16: return TagBFloat16;
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -323,7 +323,12 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 
 #if CLANG_VERSION_MAJOR >= 10
         // Handle built-in vector types as if they're normal vector types
-        if (kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool) {
+        if (kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool
+#if CLANG_VERSION_MAJOR >= 13
+            /* RISC-V vector types */
+            || kind >= BuiltinType::RvvInt8mf8 && kind <= BuiltinType::RvvBool64
+#endif // CLANG_VERSION_MAJOR >= 13
+            ) {
 // Declare ElemType and ElemCount as needed by various Clang versions
 #if CLANG_VERSION_MAJOR >= 11
             auto Info = Context->getBuiltinVectorTypeInfo(T);
@@ -338,6 +343,8 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 #else // CLANG_VERSION_MAJOR >= 11
             auto &Ctx = *Context;
             // Copy-pasted from Type::getSveEltType introduced after Clang 10:
+            // (Not extended for RISCV types
+            // as they are not available in that version anyway).
             auto ElemType = [&] {
                 switch (kind) {
                 default: llvm_unreachable("Unknown builtin SVE type!");


chrysn:
The CI failures indicate that the "|| it's one of the riscv vector types" will needs its own ifdef around it.

Before I start digging into LLVM to find the precise version to test for, is the PR in general acceptable? (there's already half a day's work in it, I'd prefer not to add more without a rough thumbs-up-or-thumbs-down.)

kkysen:
Thanks for the PR!  I think it's generally good.  We added similar code in #441 to handle the SVE types, so if it follows that, it should be generally good.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -323,7 +323,10 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 
 #if CLANG_VERSION_MAJOR >= 10
         // Handle built-in vector types as if they're normal vector types
-        if (kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool) {
+        if (kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool ||
+            /* RISC-V vector types */
+            kind >= clang::BuiltinType::RvvInt8mf8 && kind <= clang::BuiltinType::RvvBool64
```
kkysen, line 328:
```suggestion
            kind >= BuiltinType::RvvInt8mf8 && kind <= BuiltinType::RvvBool64
```

chrysn: Fixed. (I could have looked around before copy-pasting values from gdb...).

chrysn:
Now it seems to pass everything that doesn't need extra secrets.

Do you want commits neatly packed into the two it semantically is, or would you squash on merge anyway?

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -323,12 +323,22 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
...
 // Declare ElemType and ElemCount as needed by various Clang versions
 #if CLANG_VERSION_MAJOR >= 11
             auto Info = Context->getBuiltinVectorTypeInfo(T);
             auto ElemType = Info.ElementType;
 
+//             // Brutal fix around _Float16 being an unsupported BuiltinType on RISCV
+//             if (ElemType == Context->Float16Ty) {
+//               ElemType = Context->Char16Ty;
+//             }
+
```
kkysen: Is this still needed?

thedataking:
Given the size of this chang, I think a squash merge is fine – but if you have the time to clean up the commit history, that would be nice too. I noticed you have some commented out code in there – can we get rid of that before merging?

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -338,6 +348,8 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 #else // CLANG_VERSION_MAJOR >= 11
             auto &Ctx = *Context;
             // Copy-pasted from Type::getSveEltType introduced after Clang 10:
+            // (Not extended for RISCV types as they were not available in that
+            // version anyway)
```
kkysen, line 347:
```suggestion
            // (Not extended for RISCV types
            // as they are not available in that version anyway).
```

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -393,6 +405,9 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
             case BuiltinType::UInt: return TagUInt;
             case BuiltinType::ULong: return TagULong;
             case BuiltinType::ULongLong: return TagULongLong;
+            // Constructed as consequence of the conversion of built-in vector
+            // types to normal vector types
```
kkysen, line 404:
```suggestion
            // Constructed as a consequence of the conversion of
            // built-in to normal vector types.
```

chrysn:
Oups, thanks for spotting that, sorry this slipped in.

Squashing...

c2rust-ast-exporter/src/AstExporter.cpp
kkysen: I prefer wrapping on semantic breaks like punctuation, clauses, etc.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -323,7 +323,12 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 
 #if CLANG_VERSION_MAJOR >= 10
         // Handle built-in vector types as if they're normal vector types
-        if (kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool) {
+        if (kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool
+#if CLANG_VERSION_MAJOR >= 13
+            /* RISC-V vector types */
+            || kind >= BuiltinType::RvvInt8mf8 && kind <= BuiltinType::RvvBool64
+#endif // CLANG_VERSION_MAJOR >= 13
+            ) {
```
kkysen, line 331:
```suggestion
        const bool is_sve = kind >= BuiltinType::SveInt8 && kind <= BuiltinType::SveBool;
#else
        const bool is_sve = false;
#endif // CLANG_VERSION_MAJOR >= 10

#if CLANG_VERSION_MAJOR >= 13
        // RISC-V vector types
        const bool is_rvv = kind >= BuiltinType::RvvInt8mf8 && kind <= BuiltinType::RvvBool64;
#else
        const bool is_rvv = false;
#endif // CLANG_VERSION_MAJOR >= 13

#if CLANG_VERSION_MAJOR >= 10
            if (is_sve || is_rvv) {
```

I think this is cleaner and more symmetrical.

I don't *think* that last `#if CLANG_VERSION_MAJOR >= 10` is needed, so if it's not, you can remove it and its `#endif`.

This change I think should go in a separate commit.

kkysen (changes requested):
Some small changes I suggested, but generally it looks very good.

kkysen:
> Do you want commits neatly packed into the two it semantically is, or would you squash on merge anyway?

I think rebasing to 2 semantic commits would be nice, plus possibly this one (https://github.com/immunant/c2rust/pull/693#discussion_r982042196) as a 3rd.

c2rust-ast-exporter/src/AstExporter.cpp
chrysn: It's already removed in the squash.

c2rust-ast-exporter/src/AstExporter.cpp
chrysn: The "if >= 10" is still needed for switch-cases that refer to `builtinType::Sve*`. I don't think of this as strictly better, but it's not worse either, so as you prefer it, I'm working it in.

chrysn: The diff algorithm made 3001e7b77b957903bef237fc6e44d8ff7880d693 remarkably pretty :-)

chrysn: (Not *that* slim any more as 0feb8f3b091711a2897c1261eeaee73e1007ac6d now that I adjusted the comments, but still looks fine.)

kkysen: This looks much better I think!  Narrowing the scope of the `#ifdef`s helps a lot in understanding the code.  And the comments at the top are rearranged better now, too.

chrysn: This has been causing all kinds of other trouble, compared to the original proposal.

I'm splitting that out into another PR (https://github.com/immunant/c2rust/pull/694) for the record, but think this one is better off without them. It is rolled back to only the original changes, with all the smaller fixes, but without this thread's refactoring.

kkysen: Thanks, I agree your way is better.

chrysn:
Has the Darwin pipeline been updated subtly? One failed job [shows](https://dev.azure.com/immunant/c2rust/_build/results?buildId=2661&view=logs&jobId=02bad918-85c8-5a4b-30eb-93cb7f58f78e&j=02bad918-85c8-5a4b-30eb-93cb7f58f78e&t=92312585-ff34-5e62-11b6-5ebd1acc3131)

```
  * Closing connection 1
  /Users/runner/work/1/s/c2rust-ast-exporter/src/AstExporter.cpp:2280:52: error: no member named 'Ascii' in 'clang::StringLiteral::StringKind'
              case clang::StringLiteral::StringKind::Ascii:
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
  1 error generated.
```

which is the symptom of https://github.com/immunant/c2rust/issues/676 (indicating that the image is using LLVM 15 already).

kkysen:
> Has the Darwin pipeline been updated subtly?

#677 is not merged yet, so I don't think so.  Darwin CI is solely based on `azure-pipelines.yml`, unlike the others which are based on Docker images and have to be updated separately from GitHub.  `azure-pipelines.yml` does say it uses `macOS-latest`, so perhaps that's been updated?  @thedataking, thoughts?

chrysn:
> `azure-pipelines.yml` does say it uses `macOS-latest`, so perhaps that's been updated?

Rerunning the GitHub actions on the master branch might be helpful here.

kkysen:
You're right; I missed that `master` just started failing.

This is weird: https://github.com/immunant/c2rust/commit/7e79f918a081b54d70deac367612f644cd8233ee was the first to fail, starting at 11:06 pm, but then https://github.com/immunant/c2rust/commit/ec2c291dfc57df8fbb36ae64359005f284373aa3 right after succeeded, starting at 11:12 pm.

kkysen:
LLVM 15 was indeed used in the [failing Darwin CI run](https://dev.azure.com/immunant/c2rust/_build/results?buildId=2644&view=logs&j=02bad918-85c8-5a4b-30eb-93cb7f58f78e&t=92312585-ff34-5e62-11b6-5ebd1acc3131&l=316).  I can't tell what the successful Darwin CI run used, though, as the log on error wasn't printed, but I assume it wasn't LLVM 15.  Nothing should be different between those commits, so I'm guessing it's an Azure macOS-latest issue.

kkysen (approved):
I think it looks all good now.  The Darwin CI failure appears to be totally separate and a coincidence.

PR #694, opened 2022-09-28
Refactoring around vector type recognition

chrysn:
This is based on https://github.com/immunant/c2rust/pull/693.

The changes were suggested by @kkysen in https://github.com/immunant/c2rust/pull/693#discussion_r982042196 but caused larger fall-out and should IMO be discussed separately.


kkysen:
I agree this is a better approach to do it as a separate PR.  Can you make the base branch #693?

PR #696, opened 2022-10-09
Inline steps from `run_ci_checks.sh` into `azure-pipelines.yml`

kkysen:
This inlines steps from `run_ci_checks.sh` into `azure-pipelines.yml`.

This is a semantic revert of 6593b6c53c6b7c9517aa863f14358a7c65b65c4b.

Previously (in 6593b6c53c6b7c9517aa863f14358a7c65b65c4b), I had deduplicated the code (and documentation) from `run_ci_checks.sh` and `azure-pipelines.yml` (both for the Linux and macOS steps, which were duplicated) so that the CI steps were always in sync.  I had run into a couple of bugs where they were subtly out of sync, so I thought it'd be best to deduplicate them and have the single source of truth in the cross-platform `run_ci_checks.sh`.

However, @thedataking thinks it's more important to keep the code in `azure-pipelines.yml` completely inlined so that it's clear what exactly is being run without needing to consult any other scripts or files, so I'm opening this PR to revert* 6593b6c53c6b7c9517aa863f14358a7c65b65c4b and re-inline all the steps into both the Linux and macOS CI steps.

\* There was another change in 6593b6c53c6b7c9517aa863f14358a7c65b65c4b: putting each step in a separate function and allowing them to be easily called on the command line.  I use this feature often for running CI steps locally, and it is independent from @thedataking's concern about inlining, so I'm leaving it as is instead of doing an automatic `git revert 6593b6c53c6b7c9517aa863f14358a7c65b65c4b`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,42 +43,66 @@ jobs:
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh fmt
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      cargo fmt --check
     displayName: 'cargo fmt --check'
   
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh check
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # Differs from `cargo build` since we can use `--all-features` here.
+      # We plan to eventually replace `cargo check` with `cargo clippy`.
+      cargo check --tests --all-features
     displayName: 'cargo check'
 
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh doc
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      cargo doc --all-features --document-private-items --no-deps
     displayName: 'cargo doc'
 
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh build
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # Don't build with `--all-features` as `--all-features` includes `--features llvm-static`,
+      # which we don't want to test here (it doesn't work out of the box on Arch and Fedora;
+      # see https://github.com/immunant/c2rust/issues/500).
+      cargo build --release
     displayName: 'cargo build against host clang/LLVM (fast build)'
 
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh test
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      cargo test --release --exclude c2rust-analyze --workspace
     displayName: 'cargo test'
   
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh test-translator
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # `test_translator.py` compiles translated code,
+      # which has tons of warnings.
+      # `RUSTFLAGS="-D warnings"` would be inherited by that,
+      # causing tons of errors, so unset that.
+      # `test_translator.py` does not rebuild,
+      # so changing `RUSTFLAGS` will not trigger a full rebuild.
+      unset RUSTFLAGS
+      ./scripts/test_translator.py tests/
     displayName: 'Test translator (fast build)'
 
 - job: Darwin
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -43,42 +43,66 @@ jobs:
...
     displayName: 'cargo test'
   
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh test-translator
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # `test_translatory.py` compiles translated code,
```
thedataking: typo: translatory

azure-pipelines.yml
```diff
@@ -43,42 +43,66 @@ jobs:
...
+      # which has tons of warnings.
+      # `RUSTFLAGS="-D warnings"` would be inherited by that,
+      # causing tons of errors, so unset that.
+      unset RUSTFLAGS
```
thedataking, line 104: Remove `unset RUSTFLAGS` and `export RUSTFLAGS...` above?

azure-pipelines.yml
```diff
@@ -103,29 +127,53 @@ jobs:
...
 
   - script: |
-      ./scripts/run_ci_checks.sh test-translator
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # `test_translator.py` compiles translated code,
+      # which has tons of warnings.
+      # `RUSTFLAGS="-D warnings"` would be inherited by that,
+      # causing tons of errors, so unset that.
+      unset RUSTFLAGS
```
thedataking, line 177: See previous comment re `unset RUSTFLAGS`

azure-pipelines.yml
```diff
@@ -103,29 +127,53 @@ jobs:
...
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # Soon to be superceded by the commented out `cargo clippy` below.
+      # This is different from `cargo build`
+      # as this uses `--all-features` to check everything.
+      cargo check --tests --all-features
+      # cargo clippy --tests --all-features
     displayName: 'cargo check'
 
   - script: |
```
thedataking, line 146: I think it is safe to skip rustdoc on Darwin since we're mostly interested in detecting whether we broke the build or any tests.

azure-pipelines.yml
```diff
@@ -103,29 +127,53 @@ jobs:
...
+      export RUSTDOCFLAGS="-D warnings"
+      cargo doc --all-features --document-private-items --no-deps
     displayName: 'cargo doc'
 
   - script: |
       export LLVM_CONFIG_PATH=$(brew --prefix llvm)/bin/llvm-config
-      ./scripts/run_ci_checks.sh build
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # Don't build with `--all-features` as `--all-features` includes `--features llvm-static`,
```
thedataking, line 157: nitpick: You can cut down on duplication by saying something like 

```
# Don't build with `all-features`, see fast build step for Linux.
```

azure-pipelines.yml
```diff
@@ -43,42 +43,66 @@ jobs:
...
     displayName: 'cargo fmt --check'
   
   - script: |
       export PATH="/home/docker/.cargo/bin:$PATH"
       export RUSTUP_HOME=/home/docker/.rustup
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
-      ./scripts/run_ci_checks.sh check
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      # Soon to be superceded by the commented out `cargo clippy` below.
```
thedataking, line 57: Why not simply say something like:

```
 # Cargo check step differs from cargo build step since we can use `--all-features` here.
 # We plan to replace `cargo check` with `cargo clippy` (while keeping flags the same). 
 cargo check --tests --all-features
```

azure-pipelines.yml
```diff
@@ -103,29 +127,53 @@ jobs:
 
   - script: |
       export LLVM_CONFIG_PATH=$(brew --prefix llvm)/bin/llvm-config
-      ./scripts/run_ci_checks.sh fmt
+      export RUSTFLAGS="-D warnings"
+      export RUSTDOCFLAGS="-D warnings"
+      cargo fmt --check
     displayName: 'cargo fmt --check'
   
   - script: |
```
thedataking, line 135: I would skip this step on Darwin since we're mostly interested in detecting broken builds. That would also cut down on duplication. If you insist on keeping it, I would refer to the comment for cargo check above rather than duplicating the comment.

azure-pipelines.yml
kkysen: I fixed it in https://github.com/immunant/c2rust/commit/a6a989c51c4c5a892ff5acfcbeeaf00403be16bb and rebased that change here.

kkysen:
Re-duplicating checks on Darwin and Linux, I thought about it a bit more carefully.  `cargo fmt` takes about ~2s and `cargo doc` about ~10s, so there's not too much benefit in removing those.

The `cargo fmt` check is fully redundant, as it should be totally platform-independent, but it only takes a couple of seconds, so I can remove it from Darwin, but there's little value in doing so.

`cargo doc` is not quite fully redundant, as it could differ between targets.  Given it only takes a few seconds, I don't know if removing it is useful enough.

`cargo check --all-features` is the extra step that takes some time, but this does catch things that `cargo build` doesn't and that may be different on different targets.

I can refer to the previous comments for these steps, though.

azure-pipelines.yml
kkysen: See https://github.com/immunant/c2rust/pull/696#issuecomment-1272655223.

azure-pipelines.yml
kkysen: See https://github.com/immunant/c2rust/pull/696#issuecomment-1272655223.

azure-pipelines.yml
kkysen: Fixed in https://github.com/immunant/c2rust/commit/5bb01cb32657d0ca1cdadd0a6bbe4866b885b028 (on `master`).

azure-pipelines.yml
kkysen: Fixed in https://github.com/immunant/c2rust/pull/696/commits/d67e9621ce03b879e65bea485d5a7a887b82d706.

azure-pipelines.yml
kkysen: Fixed in https://github.com/immunant/c2rust/pull/696/commits/d67e9621ce03b879e65bea485d5a7a887b82d706.

azure-pipelines.yml
kkysen: Fixed in https://github.com/immunant/c2rust/pull/696/commits/da006571f57674fef171f13142881a4caf489c30.  I also made the same referring to the Linux comment for details change for the test translator step, too.

azure-pipelines.yml
thedataking: This is not what I suggested. I suggested you improve the comment in this script because it (and not `run_ci_checks.sh`) is the source of truth w.r.t. what is happening in azure pipelines.

azure-pipelines.yml
thedataking: > Re-duplicating checks on Darwin and Linux, I thought about it a bit more carefully. cargo fmt takes about ~2s and cargo doc about ~10s, so there's not too much benefit in removing those.

The benefit is in making this script shorter. As I've said, we are mostly interested in catching broken builds and tests. Remember that when reviewer and author disagree, break in favor of reviewer.

thedataking (approved):
Approved on the condition that you address the two remaining comments. I'd like to avoid spending more time on azure pipelines and your script to run commands locally since the long term plan is to move to GitHub Actions.

azure-pipelines.yml
kkysen: Oh, oops, I meant to rebase that change in here so it's updated in both places.  I'll fix that.

azure-pipelines.yml
kkysen: Okay, I'll remove it.

kkysen: Correctly rebased now.

azure-pipelines.yml
kkysen: Rebased correctly now.

PR #697, opened 2022-10-10
Added big endian feature for little endian CPU

xtmono:
Handle big endian data on little endian CPU.
Useful for dealing with network data.

### DIFF
--- a/c2rust-bitfields-derive/src/lib.rs
+++ b/c2rust-bitfields-derive/src/lib.rs
@@ -21,6 +21,7 @@ struct BFFieldAttr {
     name: String,
     ty: String,
     bits: (String, proc_macro2::Span),
+    endian: Option<String>,
 }
 
 fn parse_bitfield_attr(
@@ -31,6 +32,7 @@ fn parse_bitfield_attr(
     let mut ty = None;
     let mut bits = None;
     let mut bits_span = None;
+    let mut endian = None;
 
     if let Meta::List(meta_list) = attr.parse_meta()? {
         for nested_meta in meta_list.nested {
@@ -53,6 +55,7 @@ fn parse_bitfield_attr(
                             bits = Some(rhs_string);
                             bits_span = Some(meta_name_value.path.span());
                         }
+                        "endian" => endian = Some(rhs_string),
                         // This one shouldn't ever occur here,
                         // but we're handling it just to be safe
                         "padding" => {
@@ -97,6 +100,7 @@ fn parse_bitfield_attr(
         name: name.unwrap(),
         ty: ty.unwrap(),
         bits: (bits.unwrap(), bits_span.unwrap()),
+        endian: endian,
     }))
 }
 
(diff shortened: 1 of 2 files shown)


c2rust-bitfields-derive/src/lib.rs
```diff
@@ -97,6 +100,7 @@ fn parse_bitfield_attr(
         name: name.unwrap(),
         ty: ty.unwrap(),
         bits: (bits.unwrap(), bits_span.unwrap()),
+        endian: endian,
```
kkysen, line 103:
```suggestion
        endian,
```

c2rust-bitfields/src/lib.rs
```diff
@@ -28,7 +28,11 @@ pub trait FieldType: Sized {
         let (lhs_bit, rhs_bit) = bit_range;
 
         for (i, bit_index) in (lhs_bit..=rhs_bit).enumerate() {
-            let byte_index = bit_index / 8;
+            let byte_index = if big_endian {
+                field.len() - 1 - (bit_index / 8)
+            } else {
+                bit_index / 8
+            };
```
kkysen, line 35:
```suggestion
            let byte_index = bit_index / 8;
            let byte_index = if big_endian {
                field.len() - 1 - byte_index
            } else {
                byte_index
            };
```

kkysen: Also, can you refactor all 3 (the 2 below changes, too) of these same computations into some helper function?

c2rust-bitfields-derive/src/lib.rs
```diff
@@ -228,6 +232,17 @@ fn bitfield_struct_impl(struct_item: ItemStruct) -> Result<TokenStream, Error> {
     let field_bit_info = field_bit_info?;
     let field_bit_info_setters = &field_bit_info;
     let field_bit_info_getters = &field_bit_info;
+    let field_endians: Vec<_> = bitfields
+        .iter()
+        .map(|field| {
+            field
+                .endian
+                .as_ref()
+                .map_or(false, |endian| endian == "big")
```
kkysen, line 241: Should it only check for `big` or should it have to be `big`, `little`, or nothing?

c2rust-bitfields-derive/src/lib.rs
```diff
@@ -228,6 +232,17 @@ fn bitfield_struct_impl(struct_item: ItemStruct) -> Result<TokenStream, Error> {
...
+        })
+        .collect();
```
kkysen, line 243:
```suggestion
    let field_endians = bitfields
        .iter()
        .map(|field| {
            field
                .endian
                .as_ref()
                .map_or(false, |endian| endian == "big")
        })
        .collect::<Vec<_>>();
```

kkysen (changes requested):
Can you add a test showing how (and testing that) this works and would be used?  You can put them alongside the other bitfields tests in `tests/`.  Also, can you add documentation somewhere detailing how this can be used?  Thanks!

c2rust-bitfields-derive/src/lib.rs
chrysn: endian could be an enum, and if it had a default, this could be `.unwrap_or_default()`. (I lack the context here to understand whether a default endianness makes sense at all; gut says no).

PR #700, opened 2022-10-10
Add a single-threaded instrumentation runtime

### LINKED ISSUE #687: Add a single-threaded instrumentation runtime

kkysen:
Fixes #687.

This adds a single (main) thread instrumentation runtime in addition to the existing background thread runtime.

The single-threaded, main thread runtime greatly simplifies the logic, helping to easily and quickly expose bugs, like #613, which I solved in #691 as soon as I added the single-threaded runtime.  This, I think, is the main motivation for adding this, though it may also be faster in certain cases, and would allow instrumented code to run in single-threaded environments in the future.

The runtime, like the backend, is dynamically selected through `$INSTRUMENT_RUNTIME`, which must be either `fg` for the main thread runtime, or `bg` for the background thread runtime.  This is exactly how `$INSTRUMENT_BACKEND` currently works for the backend.

### DIFF
--- a/pdg/src/main.rs
+++ b/pdg/src/main.rs
@@ -370,7 +370,7 @@ mod tests {
     #[test]
     fn analysis_test_pdg_snapshot_debug() -> eyre::Result<()> {
         init();
-        let pdg = analysis_test_pdg_snapshot(Profile::Debug, RuntimeKind::BackgroundThread)?;
+        let pdg = analysis_test_pdg_snapshot(Profile::Debug, Default::default())?;
         insta::assert_display_snapshot!(pdg);
         Ok(())
     }
--- a/pdg/src/main.rs
+++ b/pdg/src/main.rs
@@ -332,6 +332,7 @@ mod tests {
             .arg("--")
             .args(args)
             .env("METADATA_FILE", &metadata_path)
+            .env("INSTRUMENT_RUNTIME", "bg")
             .env("INSTRUMENT_BACKEND", "log")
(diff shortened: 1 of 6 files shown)


pdg/src/main.rs
```diff
@@ -361,7 +370,7 @@ mod tests {
     #[test]
     fn analysis_test_pdg_snapshot_debug() -> eyre::Result<()> {
         init();
-        let pdg = analysis_test_pdg_snapshot(Profile::Debug)?;
+        let pdg = analysis_test_pdg_snapshot(Profile::Debug, RuntimeKind::BackgroundThread)?;
```
aneksteind, line 373: worth using `default()` instead?

aneksteind (approved):

pdg/src/main.rs
kkysen: Yeah, good idea.

kkysen: Fixed in https://github.com/immunant/c2rust/pull/700/commits/7efe3f8cb2338a64712b66bded268c31de4397c2.

PR #702, opened 2022-10-13
upgrade the dependencies

marcograss:
Implements #395 , but upgrade other dependencies as well.

I might port the code to clap 4 also in the future since it will have to happen sooner or later

### DIFF
--- a/c2rust-analyze/build.rs
+++ b/c2rust-analyze/build.rs
@@ -5,5 +5,5 @@ fn main() {
     sysroot.link_rustc_private();
 
     print!("cargo:rustc-env=C2RUST_TARGET_LIB_DIR=");
-    print_bytes::println_bytes(&sysroot.rustlib());
+    print_bytes::println_lossy(&sysroot.rustlib());
 }
--- a/c2rust-ast-exporter/Cargo.toml
+++ b/c2rust-ast-exporter/Cargo.toml
@@ -24,7 +24,7 @@ bindgen = { version = "0.60", features = ["logging"] }
 clang-sys = "1.4"
 # Pinning until https://github.com/rust-lang/cmake-rs/issues/131 is resolved
 # Fixed by https://github.com/rust-lang/cmake-rs/pull/146 on 5/12/2022; waiting for next release.
-cmake = "=0.1.45"
+cmake = "=0.1.48"
 env_logger = "0.9"
 c2rust-build-paths = { path = "../c2rust-build-paths" }
 
--- a/c2rust-ast-exporter/Cargo.toml
+++ b/c2rust-ast-exporter/Cargo.toml
@@ -21,7 +21,7 @@ serde_cbor = "0.11"
 
 [build-dependencies]
 bindgen = { version = "0.60", features = ["logging"] }
-clang-sys = "1.3"
+clang-sys = "1.4"
 # Pinning until https://github.com/rust-lang/cmake-rs/issues/131 is resolved
 # Fixed by https://github.com/rust-lang/cmake-rs/pull/146 on 5/12/2022; waiting for next release.
 cmake = "=0.1.45"
--- a/c2rust/Cargo.toml
+++ b/c2rust/Cargo.toml
@@ -24,7 +24,7 @@ env_logger = "0.9"
 git-testament = "0.2.1"
 is_executable = "1.0"
 log = "0.4"
-regex = "1.3"
+regex = "1.6"
 shlex = "1.1"
 
 [build-dependencies]
--- a/c2rust/Cargo.toml
+++ b/c2rust/Cargo.toml
@@ -1,34 +1,34 @@
 [package]
-name = "c2rust"
-description = "C to Rust translation, refactoring, and cross-checking"
-version = "0.16.0"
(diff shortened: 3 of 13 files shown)


marcograss:
any hint on how to run locally the testsuite so I can see what's going on that makes the CI fail?

kkysen:
> any hint on how to run locally the testsuite so I can see what's going on that makes the CI fail?

You can run `./scripts/run_ci_checks.sh`, which is what's run in CI for the most part.

kkysen:
Specifically, `./scripts/test_translator.py tests/` is failing.  You can run that and see it's `--help` for some debug output options.  And `cargo build --release` should be run before that.

c2rust-ast-exporter/Cargo.toml
```diff
@@ -21,10 +21,10 @@ serde_cbor = "0.11"
 
 [build-dependencies]
 bindgen = { version = "0.60", features = ["logging"] }
-clang-sys = "1.3"
+clang-sys = "1.4"
 # Pinning until https://github.com/rust-lang/cmake-rs/issues/131 is resolved
 # Fixed by https://github.com/rust-lang/cmake-rs/pull/146 on 5/12/2022; waiting for next release.
-cmake = "=0.1.45"
+cmake = "=0.1.48"
```
kkysen, line 27: The comment and pinning are there for a reason.  We can't upgrade until the next `cmake` version comes out, as there hasn't been a release since https://github.com/rust-lang/cmake-rs/pull/146 merged, fixing https://github.com/rust-lang/cmake-rs/issues/131 in `master`.

The `cmake` on Ubuntu 18 is an older version that doesn't support `--parallel`, so it fails there.  See the [CI failure](https://dev.azure.com/immunant/c2rust/_build/results?buildId=2693&view=logs&j=ae6377c6-aed8-50bb-2c16-24013b3441a6&t=19222e3a-3fea-5cc3-b08a-6873fb35b846&l=338).

c2rust-analyze/build.rs
```diff
@@ -5,5 +5,5 @@ fn main() {
     sysroot.link_rustc_private();
 
     print!("cargo:rustc-env=C2RUST_TARGET_LIB_DIR=");
-    print_bytes::println_bytes(&sysroot.rustlib());
+    print_bytes::println_lossy(&sysroot.rustlib());
```
kkysen, line 8: Since this is a breaking change, I think it should be changed in a separate PR.

c2rust/Cargo.toml
```diff
@@ -1,34 +1,34 @@
...
+description = "C to Rust translation, refactoring, and cross-checking"
+edition = "2021"
+homepage = "https://c2rust.com/"
 keywords = ["transpiler", "migration", "translation", "c"]
+license = "BSD-3-Clause"
+name = "c2rust"
 readme = "README.md"
-default-run = "c2rust"
+repository = "https://github.com/immunant/c2rust"
+version = "0.16.0"
```
kkysen, line 13: Please don't re-order these.  While a standard order would be nice, alphabetical is not the right one, as `name = "c2rust"` makes sense to be first.  And if we did, it should be in a separate PR.

c2rust/Cargo.toml
```diff
@@ -1,34 +1,34 @@
...
 
 [badges]
-travis-ci = { repository = "immunant/c2rust" }
-azure-devops = { project = "immunant/c2rust", pipeline = "immunant.c2rust", build="1" }
+azure-devops = {project = "immunant/c2rust", pipeline = "immunant.c2rust", build = "1"}
+travis-ci = {repository = "immunant/c2rust"}
```
kkysen, line 17: Please also don't reformat, especially because `{ key` is more idiomatic than `{key` in `Cargo.toml`s.

c2rust/Cargo.toml
```diff
@@ -1,34 +1,34 @@
...
 
 [dependencies]
 anyhow = "1.0"
-clap = { version = "2.34", features = ["yaml"] }
+c2rust-transpile = {version = "0.16.0", path = "../c2rust-transpile"}
+clap = {version = "3", features = ["yaml", "cargo"]}
```
kkysen, line 22: What is the `cargo` feature needed for?

c2rust/src/transpile.yaml
```diff
@@ -82,7 +82,6 @@ args:
   - COMPILE_COMMANDS:
       help: Input compile_commands.json file
       required: true
-      index: 1
```
kkysen, line 85: Is this due to a `clap` 2 vs. 3 change?

kkysen (changes requested):
Thanks for the PR, though this PR would need to be split up into separate changes.  And please don't reformat things that don't need to be and don't have standard formatters like `rustfmt`, especially not in the same PR.

Also, #395 is supposed to move away from the yaml API in `clap`, which is now deprecated, and ideally to the derive API.  We could still update the dependency (see https://github.com/immunant/c2rust/issues/395#issuecomment-1278207231), but that's not what that issue is really for.

marcograss:
thanks for the feedback, I will close this and rework it accordingly

kkysen:
Superseded by:
* #703
* #704

PR #703, opened 2022-10-14
Update to `clap` 3

marcograss:
Some comments on the PR:

The `cargo` feature is required in `clap` 3 to use some functions that before in `clap` 2 were enabled by default instead, such as `crate_version` and similar.

`transpile.yaml` has been modified because of an incompatibility with `clap` 3 where the `compile_commands.json` argument wasn't properly setup as it was with `clap` 2.

About `transpile.yaml`, in the near future it must be removed anyway; `clap` 4 has removed the yaml command descriptions, and even in `clap` 3, although it's still supported, it's not advertised.

### DIFF
--- a/c2rust/Cargo.toml
+++ b/c2rust/Cargo.toml
@@ -18,7 +18,7 @@ azure-devops = { project = "immunant/c2rust", pipeline = "immunant.c2rust", buil
 
 [dependencies]
 anyhow = "1.0"
-clap = { version = "3", features = ["yaml", "cargo"] }
+clap = { version = "3.2.22", features = ["yaml", "cargo"] }
 env_logger = "0.10"
 git-testament = "0.2.1"
 is_executable = "1.0"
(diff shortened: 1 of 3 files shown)


c2rust/Cargo.toml
```diff
@@ -18,7 +18,7 @@ azure-devops = { project = "immunant/c2rust", pipeline = "immunant.c2rust", buil
 
 [dependencies]
 anyhow = "1.0"
-clap = { version = "2.34", features = ["yaml"] }
+clap = { version = "3", features = ["yaml", "cargo"] }
```
kkysen, line 21: Can you specify the minor version, too?

marcograss: done

kkysen (approved):
We can update to `clap` 3 even with yaml deprecated, which remains a separate issue to fix (#395) for `clap` 4, which fully removes it.  This moves us to only having one `clap` version in the workspace, which is good for compile times (`clap` is slow).  This is a minor change, so I'm approving it, but it's not very important to get merged.

PR #704, opened 2022-10-15
Update `print_bytes` crate version

marcograss:
This PR updates the dependency `print_bytes` to `0.7.0`, which has a breaking change, renaming `println_bytes` to `println_lossy`.  The actual functionality doesn't change, but since the function does sometimes replace non-encodable characters with the Unicode replacement character, lossy is a better name.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1016,9 +1016,9 @@ dependencies = [
 
 [[package]]
 name = "print_bytes"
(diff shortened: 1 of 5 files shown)


kkysen (approved):
I've updated the PR description explaining the reason for the non-functional breaking change.  Should be all fine to merge, as no functionality is changed, just the function name.

PR #707, opened 2022-10-17
Add a newtype `struct FuncId(pub DefPathHash)`

kkysen:
Added a newtype `struct FuncId(pub DefPathHash)` to make a lot of uses much clearer that they're referring to functions only.

This also will help clean up some things in #654, and in general, I think makes a lot of uses of `FuncId`s easier to understand/read.


fw-immunant (approved):
Seems like a reasonable clarification.

PR #710, opened 2022-10-21
Introduce instrumentation priority.

aneksteind:
Sometimes, two instrumentation points will get get placed into the same MIR block and statement. When this occurs, whichever instrumentation appeared first when iterating over the list of points would get placed in location x, and the second would get inserted just before, pushing the first instrumentation to location x+1.

This feature ensures that, if instrumentation A has High priority and instrumentation B has Low priority, A will always be placed before B in the instrumented block.

### DIFF
--- a/dynamic_instrumentation/src/point/build.rs
+++ b/dynamic_instrumentation/src/point/build.rs
@@ -14,7 +14,7 @@ use crate::{
     util::Convert,
 };
 
-use super::{CollectInstrumentationPoints, InstrumentationPoint};
+use super::{CollectInstrumentationPoints, InstrumentationPoint, InstrumentationPriority};
 
 #[derive(Default)]
 struct InstrumentationPointBuilder<'tcx> {
@@ -31,6 +31,7 @@ impl<'tcx> CollectInstrumentationPoints<'_, 'tcx> {
         original_location: Location,
         instrumentation_location: Location,
         func: DefId,
+        instrumentation_priority: InstrumentationPriority,
     ) {
         let id = self.instrumentation_points.len();
         let InstrumentationPointBuilder {
@@ -47,6 +48,7 @@ impl<'tcx> CollectInstrumentationPoints<'_, 'tcx> {
             args,
             is_cleanup,
             after_call,
+            instrumentation_priority,
             metadata,
         });
     }
@@ -58,6 +60,7 @@ pub struct InstrumentationBuilder<'a, 'tcx: 'a> {
     original_location: Location,
     instrumentation_location: Location,
     func: DefId,
+    instrumentation_priority: InstrumentationPriority,
     point: InstrumentationPointBuilder<'tcx>,
 }
 
@@ -75,6 +78,7 @@ impl<'a, 'tcx: 'a> CollectInstrumentationPoints<'a, 'tcx> {
             instrumentation_location,
             func,
             point: Default::default(),
+            instrumentation_priority: Default::default(),
         }
         .debug_mir()
     }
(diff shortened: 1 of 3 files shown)


dynamic_instrumentation/src/point/build.rs
```diff
@@ -140,6 +151,15 @@ impl<'tcx> InstrumentationBuilder<'_, 'tcx> {
         self
     }
 
+    /// Sets the priority of this instrumentation. A lower value
+    /// signifies higher priority and implies that an instrumentation
+    /// with higher priority will be placed before one with lower
+    /// priority.
```
kkysen, line 157: Maybe put this documentation on `InstrumentationPriority`?  I'm not sure, but I feel like something should be there.

kkysen (approved):

dynamic_instrumentation/src/point/mod.rs
```diff
@@ -15,6 +15,17 @@ use crate::{arg::InstrumentationArg, hooks::Hooks, util::Convert};
...
+/// signifies higher priority and implies that an instrumentation
+/// with higher priority will be placed before one with lower
+/// priority.
+#[derive(PartialEq, Eq, PartialOrd, Ord, Default, Copy, Clone)]
+pub enum InstrumentationPriority {
+    High,
+    #[default]
+    Low,
+}
+
```
fw-immunant, line 28: The sorting order of the enum as defined is the opposite of what the names suggest.

E.g., the below assertion fails:
```
fn main() {
	#[derive(PartialEq, Eq, PartialOrd, Ord, Default, Copy, Clone)]
	pub enum Priority {
		High,
		#[default]
		Low,
	}

	assert!(Priority::High > Priority::Low)
}
```

kkysen: Yeah, high priority comes first.

kkysen:
Wait, what was wrong with the `#[derive(PartialOrd)]`?  I think we should keep it.

aneksteind:
> Wait, what was wrong with the #[derive(PartialOrd)]? I think we should keep it.

Talked with @rinon who recommended it for clarity

rinon:
It's less clear what the ordering is here, because it depends on another property (enum ordering)

kkysen:
What about deriving it but adding specific numbers to the variants so the ordering is clear?  The manual impl is much harder to read and doesn't scale if we want to add any other priorities.

aneksteind:
I think in the short term it's unlikely we'll need to add different priorities, but you have a good point. If that seems like it's going to happen, it's a quicj change to take your suggestion. As far as clarity goes, I believe this is clear enough to serve our purposes.

kkysen:
Okay, I guess, but I approved the previous code, not the manual impl.  Manual impls for `PartialOrd` like that are not very idiomatic and it's not clear from a glance what the order actually is.  It's much less clear than the previous code.

rinon:
> Okay, I guess, but I approved the previous code, not the manual impl.  Manual impls for `PartialOrd` like that are not very idiomatic and it's not clear from a glance what the order actually is.  It's much less clear than the previous code.

My whole point is to make this ordering extremely explicit. I glanced at the enum and auto-derives, and it wasn't clear what the ordering relationship was. We WANT to make it non-trivial to add more elements, so the programmer is forced to consider the ordering instead of just adding it down at the bottom.

kkysen:
> My whole point is to make this ordering extremely explicit. I glanced at the enum and auto-derives, and it wasn't clear what the ordering relationship was. We WANT to make it non-trivial to add more elements, so the programmer is forced to consider the ordering instead of just adding it down at the bottom.

That's a good point.  I do still think a private/internal `order` function like this would be useful and much more readable, while also forcing you to purposefully update the `match` if you add a new variant, but I agree it's important to force new variants to be purposefully added by using an exhaustive `match`.

```rust
fn order(&self) -> u8 {
    use InstrumentationPriority::*;
    match self {
        Early => 0,
        Unspecified => 1,
    }
}

fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
    self.order().partial_cmp(other.order())
}
```

PR #711, opened 2022-10-23
Update paths to homebrew LLVM for M1 macs

LegNeato:
`/opt/homebrew/` is used by `hombrew` on Apple Silicon Macs, but `/usr/local/` is still used on Intel Macs, and we want to support both.  This lists both paths in the `README.md`, clarifying which is for Intel vs. Apple Silicon Macs.

These paths are also hardcoded in `c2rust-build-paths/src/lib.rs` and `c2rust-analyze/tests/filecheck.rs`.  The latter is meant to be consolidated under `c2rust-build-paths`, but we'll keep them as is for now.  We need to also update these to include the Apple Silicon `homebrew` paths.

Now, `c2rust` builds fine on Intel Macs (CI) and Apple Silicon Macs (@LegNeato's M1).

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ For example, the official LLVM packages from [apt.llvm.org](https://apt.llvm.org
     nix-shell
     ```
 
-- **OS X:** XCode command-line tools and recent LLVM (we recommend the Homebrew version) are required.
+- **macOS:** Xcode command-line tools and recent LLVM (we recommend the Homebrew version) are required.
 
     ```sh
     xcode-select --install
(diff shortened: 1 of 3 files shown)


kkysen:
Is this the M1 path?  Maybe we should just list both paths, and in `c2rust-ast-exporter`, too, where it looks in a bunch of paths for `llvm-config`.

kkysen (changes requested):
`/opt/homebrew` is used by `hombrew` on M1 macs, but `/usr/local/` is still used on Intel macs, and we want to support both.  So we should list both paths in the `README.md`, and potentially also say one is for Intel macs and one is for M1 macs (M2, etc., too).

These paths are also hardcoded in `c2rust-build-paths/src/lib.rs` and `c2rust-analyze/tests/filecheck.rs`.  The latter is meant to be consolidated under `c2rust-build-paths`.  We need to also update these to include the M1 `homebrew` paths.

kkysen (approved):
We'll probably want to eventually generate all these different but similar paths to search, or search by glob/regex or something, but for now, this looks great and updates things in all the right places I believe.

kkysen:
By the way, @fw-immunant, I think we should merge this before publishing a new `c2rust` version.

fw-immunant:
> By the way, @fw-immunant, I think we should merge this before publishing a new `c2rust` version.

I agree. This looks good to me, but I'm not familiar with Homebrew/macOS. Should we wait for @thedataking to review, or just go ahead and merge?

kkysen:
@thedataking doesn't have an M1, right?  We can wait for him to take a quick look, but I'm not sure he'll have any extra insight.  I think if it works on both on @LegNeato's M1 mac and @thedataking's Intel mac it should be good.

thedataking (approved):
LGTM. Note that the CI workflow already tests Intel Macs. 

nit: since this patch also fixes capitalization, it should use ARM, not Arm, since it is an acronym. We could also just say Intel Macs and Apple Silicon Macs. No big deal though.

PR #712, opened 2022-10-24
Do not emit `#[register_tool(...)]` unless necessary

rinon:
The register_tool feature is nightly-only, so we don't want to include it if we aren't emitting anything that requires it. This change makes register_tool optional and only adds the feature if we really need to use it.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -796,12 +796,16 @@ pub fn translate(
         // Header Reorganization: Submodule Item Stores
         for (file_id, ref mut mod_item_store) in t.items.borrow_mut().iter_mut() {
             if *file_id != t.main_file {
+                if tcfg.reorganize_definitions {
+                    t.use_feature("register_tool");
+                }
                 let mut submodule = make_submodule(
                     &t.ast_context,
                     mod_item_store,
                     *file_id,
                     &mut new_uses,
                     &t.mod_names,
+                    tcfg.reorganize_definitions,
                 );
                 let comments = t.comment_context.get_remaining_comments(*file_id);
                 submodule.set_span(match t.comment_store.borrow_mut().add_comments(&comments) {
@@ -974,6 +978,7 @@ fn make_submodule(
     file_id: FileId,
     use_item_store: &mut ItemStore,
     mod_names: &RefCell<IndexMap<String, PathBuf>>,
+    reorganize_definitions: bool,
 ) -> Box<Item> {
     let (mut items, foreign_items, uses) = item_store.drain();
     let file_path = ast_context.get_file_path(file_id);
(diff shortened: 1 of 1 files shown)


kkysen:
Ooh, this is nice!  This was always annoying.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -796,12 +796,16 @@ pub fn translate(
         // Header Reorganization: Submodule Item Stores
         for (file_id, ref mut mod_item_store) in t.items.borrow_mut().iter_mut() {
             if *file_id != t.main_file {
+                if tcfg.reorganize_definitions {
```
kkysen, line 799: Is `reorganize_definitions` still used?  What was the original use case for `register_tool`?

kkysen (approved):
If `tcfg.reorganize_definitions` is the only thing that uses `register_tool`, then I think the code changes look good, but I didn't actually run it and see how it works and stuff.  I think @fw-immunant should still take a quick look, too.

c2rust-transpile/src/translator/mod.rs
rinon: It is currently broken, due to a lack of a refactoring tool, but I very much want to use it again as soon as we can. reorg collapses all of the duplicated header declarations in a project down to a single (rust) module. We add some annotations about where declarations came from using a custom "c2rust" attribute to do this.

kkysen: > We add some annotations about where declarations came from using a custom "c2rust" attribute to do this.

Oh yeah, that could be very useful generally.

kkysen: Maybe we should have a separate flag for `uses_c2rust_attr` for adding `register_tool` or not, and then `reorganize_definitions` is the only thing that turns on `uses_c2rust_attr` for now, as I could see a `c2rust` attribute being useful for more than just `reorganize_definitions`.  But I think this PR as is, is good for now.  If we wanted to change this, it could be done in a separate PR, or when we actually add something that would need it.

kkysen (changes requested):
Actually, I ran `test_translator.py tests/` and it seems like these changes don't fix everything.  `register_tool` isn't used anymore in the transpiled modules (like `asm.c` => `asm.rs`), but it still appears in the generated `main.rs`.

rinon:
> Actually, I ran `test_translator.py tests/` and it seems like these changes don't fix everything. `register_tool` isn't used anymore in the transpiled modules (like `asm.c` => `asm.rs`), but it still appears in the generated `main.rs`.

That appears to be an artifact of the test script itself, which is fine IMO: https://github.com/immunant/c2rust/blob/master/scripts/test_translator.py#L388

kkysen:
> That appears to be an artifact of the test script itself, which is fine IMO: https://github.com/immunant/c2rust/blob/master/scripts/test_translator.py#L388

Ah, okay.  `main.rs` is only produced by `test_translator.py` and not `c2rust transpile` at all?

rinon:
The pragmas for `main.rs`, `lib.rs`, etc. when auto-generated are set by the code I changed.

kkysen (approved):
> The pragmas for `main.rs`, `lib.rs`, etc. when auto-generated are set by the code I changed.

Okay then the PR should be all good then.

fw-immunant (approved):
LGTM.

PR #714, opened 2022-11-02
handle unary operators in static analysis

aneksteind:
both `UnOp::Neg` and `UnOp::Not` are invalid on pointer types


spernsteiner (approved):
Seems reasonable - there aren't any `UnOp`s that return pointers, so no need to label the result.

PR #715, opened 2022-11-02
add label support for struct fields

aneksteind:


aneksteind:
having arguments for the ADT doesn't make sense for its fields, only for the generic types that struct accepts. instead another approach where labels for each ADT field will be applied globally will be tried

PR #719, opened 2022-11-07
c2rust-analyze: misc features needed for insertion sort driver

spernsteiner:
This branch adds support for various features that were needed for the insertion sort test driver:

* Array aggregate and `Rvalue::Len` expressions
* Array/slice indexing `Place` projections
* Slice `as_ptr` and `as_mut_ptr` methods
* `std::process::exit` and `std::process::abort`

Other misc improvements:

* Only process functions (avoids a crash when encountering `AnonConst` MIR bodies)
* Skip assigning new types for variables that are already safe references (these occur in some transpiled code)
* Add `RewriteKind::RemoveAsPtr` for expressions where an `as_ptr()` method call is no longer needed

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -4,8 +4,8 @@ use crate::context::PermissionSet;
 use crate::util::{self, Callee};
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::{
-    BinOp, Body, BorrowKind, Local, LocalDecl, Location, Operand, Place, Rvalue, Statement,
-    StatementKind, Terminator, TerminatorKind,
+    AggregateKind, BinOp, Body, BorrowKind, Local, LocalDecl, Location, Operand, Place, Rvalue,
+    Statement, StatementKind, Terminator, TerminatorKind,
 };
 use rustc_middle::ty::{TyCtxt, TyKind};
 use std::collections::HashMap;
@@ -156,6 +156,20 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 );
                 Label::default()
             }),
+
+            Rvalue::Aggregate(ref kind, ref _ops) => match **kind {
+                AggregateKind::Array(..) => {
+                    // TODO
+                    let ty = rv.ty(self.local_decls, *self.ltcx);
+                    self.ltcx.label(ty, &mut |_ty| Label::default())
+                }
+                _ => panic!("unsupported rvalue AggregateKind {:?}", kind),
+            },
+
+            Rvalue::Len(..) => {
+                let ty = rv.ty(self.local_decls, *self.ltcx);
+                self.ltcx.label(ty, &mut |_| Label::default())
+            }
 
             ref rv => panic!("unsupported rvalue {:?}", rv),
         }
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -162,8 +162,9 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
             Rvalue::Aggregate(ref kind, ref _ops) => match **kind {
                 AggregateKind::Array(..) => {
-                    // TODO
                     let ty = rv.ty(self.local_decls, *self.ltcx);
+                    // TODO: create fresh origins for all pointers in `ty`, and generate subset
+                    // relations between the regions of the array and the regions of its elements
                     self.ltcx.label(ty, &mut |_ty| Label::default())
                 }
                 _ => panic!("unsupported rvalue AggregateKind {:?}", kind),
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -149,11 +149,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 })
             }
 
-            Rvalue::Cast(_, _, ty) => self.ltcx.label(ty, &mut |ty| {
+            Rvalue::Cast(_, _, ty) => self.ltcx.label(ty, &mut |_ty| {
+                // TODO: handle Unsize casts at minimum
+                /*
                 assert!(
                     !matches!(ty.kind(), TyKind::RawPtr(..) | TyKind::Ref(..)),
                     "pointer Cast NYI"
                 );
+                */
                 Label::default()
             }),
 
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -209,7 +209,11 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                         let rv_lty = self.visit_operand(&args[0]);
                         self.do_assign(pl_lty, rv_lty);
                     }
-                    _ => {}
+                    Some(Callee::MiscBuiltin) => {}
+                    Some(Callee::Other { .. }) => {
+                        // TODO
+                    }
+                    None => {}
                 }
             }
             // TODO(spernsteiner): handle other `TerminatorKind`s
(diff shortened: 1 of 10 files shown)


c2rust-analyze/src/main.rs
```diff
@@ -291,9 +293,14 @@ fn run(tcx: TyCtxt) {
...
 
-    // Print results for each function.
+    // Print results for each function.  We use declaration order (as reported by `body_owners`)
+    // since this makes FileCheck tests easier to write.
     for ldid in tcx.hir().body_owners() {
-        let info = func_info.get_mut(&ldid).unwrap();
+        // Skip any body owners that aren't listed in `func_info`.
+        let info = match func_info.get_mut(&ldid) {
+            Some(x) => x,
+            None => continue,
```
aneksteind, line 302: does this avoid foreign functions declared with `extern "C"` for example?

spernsteiner: Functions declared in `extern` blocks aren't body owners, so this loop won't visit them in the first place.

I updated the comment on the loop header as follows:
```Rust
    // Print results for each function in `all_fn_ldids`, going in declaration order.  Concretely,
    // we iterate over `body_owners()`, which is a superset of `all_fn_ldids`, and filter based on
    // membership in `func_info`, which contains an entry for each ID in `all_fn_ldids`.
```
I think this better explains the logic here - let me know if it's still unclear.

aneksteind: It looks like those functions count as body owners somehow, or something odd is happening. If `calloc` is declared `extern "C"` in a contrived lighttpd example:

```
error: internal compiler error: compiler/rustc_typeck/src/check/mod.rs:364:9: can't type-check body of DefId(0:13 ~ lighttpd[c91b]::{extern#0}::calloc)
  --> ../analysis/test/src/lighttpd.rs:11:5
   |
11 |     fn calloc(_: libc::c_ulong, _: libc::c_ulong) -> *mut libc::c_void;
   |
```

which happens in the `mir_built` call in `for_each_callee`

c2rust-analyze/src/main.rs
```diff
@@ -457,6 +465,16 @@ fn body_owners_postorder(tcx: TyCtxt) -> Vec<LocalDefId> {
         if seen.contains(&root_ldid) {
             continue;
         }
+
+        match tcx.def_kind(root_ldid) {
+            DefKind::Fn => {}
+            DefKind::AnonConst => continue,
+            dk => panic!(
```
aneksteind, line 472: A panic I notice when testing a `lighttpd` example is:

```
thread 'rustc' panicked at 'unexpected def_kind AssocFn for body_owner DefId(0:40 ~ lighttpd[c91b]::{impl#1}::clone)', c2rust-analyze/src/main.rs:472:19
```

so it seems like `AssocFn` should be handled here as well

c2rust-analyze/src/main.rs
aneksteind: ah, I see the problem, we don't filter by body owners when they are callees:

```rust
if let Some(Callee::Other { def_id, .. }) = util::ty_callee(self.tcx, ty) {
  if let Some(ldid) = def_id.as_local() {
    (self.f)(ldid);
  }
}
```

aneksteind: another follow-up, it would appear that `as_local` should return `None` for something declared `extern "C"` but for the `calloc` example it returns `Some(DefId(0:13 ~ lighttpd[c91b]::{extern#0}::calloc))`

spernsteiner: The `as_local` behavior here is correct.  The Rust declaration of `calloc` as an extern fn is local to the crate, even if the actual code for that function is not.

I'll add a test case using extern functions and investigate further what's going wrong here.

c2rust-analyze/src/main.rs
spernsteiner: Interesting - I think this should be trivial to fix, but I'll add a test case using `#[derive(Clone)]` just to be sure.

c2rust-analyze/src/main.rs
spernsteiner: Added a test case that shows the boundary of what's currently supported for foreign fns.  I fixed the immediate panic, so you can mention a foreign function (e.g. `let f = malloc;`), but you still can't call them, the same way you can't call unsupported library functions.

c2rust-analyze/src/main.rs
spernsteiner: The trivial fix worked - the analysis no longer panics on seeing `impl Clone for ...` blocks.  However, field accesses on ADTs are still unsupported, so deriving `Clone` without `Copy` on a non-empty struct will cause a different panic.  (Deriving Clone and Copy together is fine, since in that case the generated Clone impl just does a copy rather than cloning field by field.)

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -149,14 +149,31 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                     !matches!(ty.kind(), TyKind::RawPtr(..) | TyKind::Ref(..)),
                     "pointer Cast NYI"
                 );
+                */
                 Label::default()
             }),
 
+            Rvalue::Aggregate(ref kind, ref _ops) => match **kind {
+                AggregateKind::Array(..) => {
+                    // TODO
```
aneksteind, line 165: could you add more specifics to this TODO?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -127,28 +149,33 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-                assert!(rv_sub_lty.label != PointerId::NONE);
-                self.add_equiv(pl_sub_lty.label, rv_sub_lty.label);
+    fn do_unify(&mut self, lty1: LTy<'tcx>, lty2: LTy<'tcx>) {
+        assert_eq!(lty1.ty, lty2.ty);
+        for (sub_lty1, sub_lty2) in lty1.iter().zip(lty2.iter()) {
+            eprintln!("equate {:?} = {:?}", sub_lty1, sub_lty2);
+            if sub_lty1.label != PointerId::NONE || sub_lty2.label != PointerId::NONE {
+                assert!(sub_lty1.label != PointerId::NONE);
+                assert!(sub_lty2.label != PointerId::NONE);
+                self.add_equiv(sub_lty1.label, sub_lty2.label);
```
aneksteind, line 178: any documentation for what equivalence or unification means?

c2rust-analyze/src/expr_rewrite.rs
```diff
@@ -321,6 +327,20 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+        let op_lty = self.acx.type_of(op);
+        let op_ptr = op_lty.label;
+        let result_ptr = result_lty.label;
+
+        let (op_own, op_qty) = type_desc::perms_to_desc(self.perms[op_ptr], self.flags[op_ptr]);
+        let (result_own, result_qty) =
+            type_desc::perms_to_desc(self.perms[result_ptr], self.flags[result_ptr]);
+
+        if op_own == result_own && op_qty == result_qty {
+            self.emit(RewriteKind::RemoveAsPtr);
```
aneksteind, line 340: is removing an `as_ptr` call desired because you are keeping the value a slice?

aneksteind:
a few questions and requests for documentation, none blocking

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -127,28 +149,33 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
 
-        // Add a dataflow edge indicating that `rv` flows into `pl`.
-        assert!(pl_lty.label != PointerId::NONE);
-        assert!(rv_lty.label != PointerId::NONE);
-        self.add_edge(rv_lty.label, pl_lty.label);
+    fn do_assign_pointer_ids(&mut self, pl_ptr: PointerId, rv_ptr: PointerId) {
+        if pl_ptr != PointerId::NONE || rv_ptr != PointerId::NONE {
+            assert!(pl_ptr != PointerId::NONE);
+            assert!(rv_ptr != PointerId::NONE);
+            self.add_edge(rv_ptr, pl_ptr);
```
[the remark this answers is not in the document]
spernsteiner: I tried documenting this a bit more in https://github.com/immunant/c2rust/pull/719/commits/ef87583f576417860bdce18929c5dcaa744b722a - let me know if this helps

aneksteind (approved):

PR #726, opened 2022-11-15
label ADT struct fields in static dataflow analysis

aneksteind:
Struct fields in `TyKind::Adt` were previously ignored in labeling with `PointerId`s. This PR adds support for that.

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -34,7 +34,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
     pub fn visit_place(&mut self, pl: Place<'tcx>) -> LTy<'tcx> {
         let mut lty = self.local_ltys[pl.local.index()];
         for proj in pl.projection {
-            lty = util::lty_project(lty, &proj);
+            lty = util::lty_project(lty, &proj, |_, _| panic!("Adt not supported"));
         }
         lty
     }
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -13,7 +13,6 @@ use rustc_middle::mir::{
 };
 use rustc_middle::ty::adjustment::PointerCast;
 use rustc_middle::ty::{AdtDef, Ty, TyCtxt, TyKind};
-use rustc_target::abi::VariantIdx;
 use std::collections::HashMap;
 use std::ops::Index;
 
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -72,7 +72,7 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
-    pub field_tys: HashMap<(DefId, Field, VariantIdx), LTy<'tcx>>,
+    pub field_tys: HashMap<DefId, LTy<'tcx>>,
 
     next_ptr_id: NextGlobalPointerId,
 }
@@ -368,7 +368,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             let res = *self
                 .gacx
                 .field_tys
-                .get(&(adtdef.did(), field, VariantIdx::from_usize(0)))
+                .get(&(adtdef.all_fields().nth(usize::from(field)).unwrap().did))
                 .unwrap_or_else(|| {
                     panic!("Could not find {adtdef:?}.{fielddef_name:?} in field type map")
                 });
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -362,7 +362,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
         let adt_func = |adtdef: AdtDef, field: Field| {
-            let field_def = adtdef.all_fields().nth(usize::from(field)).unwrap();
+            let field_def = &adtdef.non_enum_variant().fields[field.index()];
             let field_def_name = field_def.name;
             eprintln!("projecting into {adtdef:?}.{field_def_name:}");
             let res = *self.gacx.field_tys.get(&field_def.did).unwrap_or_else(|| {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -4,6 +4,7 @@ use crate::pointer_id::{
     PointerTableMut,
 };
 use crate::util::{self, describe_rvalue, RvalueDesc};
+use crate::AssignPointerIds;
 use bitflags::bitflags;
 use rustc_hir::def_id::DefId;
 use rustc_index::vec::IndexVec;
@@ -12,7 +13,7 @@ use rustc_middle::mir::{
     PlaceRef, Rvalue,
 };
 use rustc_middle::ty::adjustment::PointerCast;
-use rustc_middle::ty::{AdtDef, Ty, TyCtxt, TyKind};
+use rustc_middle::ty::{AdtDef, FieldDef, Ty, TyCtxt, TyKind};
 use std::collections::HashMap;
 use std::ops::Index;
 
@@ -155,6 +156,17 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         }
 
         *next_ptr_id = counter;
+    }
+
+    pub fn label_field(&mut self, field: &FieldDef) {
+        let lty = self.assign_pointer_ids(self.tcx.type_of(field.did));
+        self.field_tys.insert(field.did, lty);
+    }
+
+    pub fn label_struct_fields(&mut self, did: DefId) {
+        for field in self.tcx.adt_def(did).all_fields() {
+            self.label_field(field);
+        }
     }
 }
 
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -8,11 +8,12 @@ use bitflags::bitflags;
 use rustc_hir::def_id::DefId;
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::{
-    Body, CastKind, HasLocalDecls, Local, LocalDecls, Location, Operand, Place, PlaceElem,
+    Body, CastKind, Field, HasLocalDecls, Local, LocalDecls, Location, Operand, Place, PlaceElem,
     PlaceRef, Rvalue,
 };
 use rustc_middle::ty::adjustment::PointerCast;
-use rustc_middle::ty::{Ty, TyCtxt, TyKind};
+use rustc_middle::ty::{AdtDef, Ty, TyCtxt, TyKind};
+use rustc_target::abi::VariantIdx;
 use std::collections::HashMap;
 use std::ops::Index;
 
@@ -71,6 +72,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
+    pub field_tys: HashMap<(AdtDef<'tcx>, Field, VariantIdx), LTy<'tcx>>,
+
     next_ptr_id: NextGlobalPointerId,
 }
 
@@ -101,6 +104,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx,
             lcx: LabeledTyCtxt::new(tcx),
             fn_sigs: HashMap::new(),
+            field_tys: HashMap::new(),
             next_ptr_id: NextGlobalPointerId::new(),
         }
     }
@@ -137,6 +141,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx: _,
             lcx,
             ref mut fn_sigs,
+            field_tys: _,
             ref mut next_ptr_id,
         } = *self;
 
@@ -357,7 +362,19 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
     }
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
-        util::lty_project(lty, proj)
+        let adt_func = |adtdef: AdtDef, field: Field| {
+            let fielddef_name = adtdef.all_fields().collect::<Vec<_>>()[usize::from(field)].name;
+            eprintln!("projecting into {adtdef:?}.{fielddef_name:}");
+            let res = *self
+                .gacx
+                .field_tys
+                .get(&(adtdef, field, VariantIdx::from_usize(0)))
+                .unwrap_or_else(|| {
+                    panic!("Could not find {adtdef:?}.{fielddef_name:?} in field type map")
+                });
+            res
+        };
+        util::lty_project(lty, proj, adt_func)
     }
 }
 
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -72,7 +72,7 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
-    pub field_tys: HashMap<(AdtDef<'tcx>, Field, VariantIdx), LTy<'tcx>>,
+    pub field_tys: HashMap<(DefId, Field, VariantIdx), LTy<'tcx>>,
 
     next_ptr_id: NextGlobalPointerId,
 }
@@ -368,7 +368,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             let res = *self
                 .gacx
                 .field_tys
-                .get(&(adtdef, field, VariantIdx::from_usize(0)))
+                .get(&(adtdef.did(), field, VariantIdx::from_usize(0)))
                 .unwrap_or_else(|| {
                     panic!("Could not find {adtdef:?}.{fielddef_name:?} in field type map")
                 });
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -362,11 +362,11 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
         let adt_func = |adtdef: AdtDef, field: Field| {
-            let fielddef = adtdef.all_fields().nth(usize::from(field)).unwrap();
-            let fielddef_name = fielddef.name;
-            eprintln!("projecting into {adtdef:?}.{fielddef_name:}");
-            let res = *self.gacx.field_tys.get(&fielddef.did).unwrap_or_else(|| {
-                panic!("Could not find {adtdef:?}.{fielddef_name:?} in field type map")
+            let field_def = adtdef.all_fields().nth(usize::from(field)).unwrap();
+            let field_def_name = field_def.name;
+            eprintln!("projecting into {adtdef:?}.{field_def_name:}");
+            let res = *self.gacx.field_tys.get(&field_def.did).unwrap_or_else(|| {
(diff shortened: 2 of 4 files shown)


c2rust-analyze/src/main.rs
```diff
@@ -150,6 +151,23 @@ fn run(tcx: TyCtxt) {
         gacx.fn_sigs.insert(ldid.to_def_id(), lsig);
     }
 
+    // Label the field types of each struct.
+    for defldid in tcx.hir_crate_items(()).definitions() {
+        let defdid = defldid.to_def_id();
+        if let DefKind::Struct = tcx.def_kind(defdid) {
+            let defty = tcx.type_of(defdid).kind();
+            if let TyKind::Adt(adtdef, _) = defty {
```
spernsteiner, line 159: You can get the `AdtDef` directly by calling `tcx.adt_def(defdid)`, instead of going through `type_of` + pattern matching.

c2rust-analyze/src/util.rs
```diff
@@ -182,6 +184,7 @@ fn builtin_callee<'tcx>(
 pub fn lty_project<'tcx, L: Debug>(
     lty: LabeledTy<'tcx, L>,
     proj: &PlaceElem<'tcx>,
+    adt_func: impl Fn(AdtDef<'tcx>, Field) -> LabeledTy<'tcx, L>,
```
spernsteiner, line 187: This closure should receive the `VariantIdx` too (assuming you decide to keep the `VariantIdx` as part of the key), so that the closure can essentially just be a lookup function and doesn't need to contain any interesting logic.

kkysen (commented):
Nit: Unless this is already a naming convention in `rustc`, can you separate parts of variables with `_`?  Like `adt_func`, but unlike:
* `fielddef_name` => `field_def_name`
* `adtdef` => `adt_def`
* `defldid` => `def_ldid` (but listen to @spernsteiner's advice first)
* `defdid` => `def_did` (but listen to @spernsteiner's advice first)
* `defty` => `def_ty`

c2rust-analyze/src/context.rs
```diff
@@ -357,7 +362,19 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
     }
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
-        util::lty_project(lty, proj)
+        let adt_func = |adtdef: AdtDef, field: Field| {
+            let fielddef_name = adtdef.all_fields().collect::<Vec<_>>()[usize::from(field)].name;
```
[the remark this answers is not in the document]
spernsteiner: Or look it up directly by variant and field indices: `adtdef.variant(variant_idx).fields[...].name`

c2rust-analyze/src/context.rs
```diff
@@ -71,6 +72,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
+    pub field_tys: HashMap<(AdtDef<'tcx>, Field, VariantIdx), LTy<'tcx>>,
```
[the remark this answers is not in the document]
aneksteind: I used the `AdtDef` `DefId` in 1d10264e.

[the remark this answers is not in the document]
aneksteind: ... and then switched to just the Field DefId in 51239c3b. Let me know which you prefer

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
aneksteind: switched to `nth` in 49721a4d

c2rust-analyze/src/util.rs
aneksteind: `VariantIdx` is no longer part of the key

c2rust-analyze/src/context.rs
```diff
@@ -357,7 +361,16 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
...
-        util::lty_project(lty, proj)
+        let adt_func = |adtdef: AdtDef, field: Field| {
+            let fielddef = adtdef.all_fields().nth(usize::from(field)).unwrap();
+            let fielddef_name = fielddef.name;
+            eprintln!("projecting into {adtdef:?}.{fielddef_name:}");
+            let res = *self.gacx.field_tys.get(&fielddef.did).unwrap_or_else(|| {
+                panic!("Could not find {adtdef:?}.{fielddef_name:?} in field type map")
+            });
+            res
+        };
```
kkysen, line 372:
```suggestion
        let adt_func = |adt_def: AdtDef, field: Field| {
            let field_def = adt_def.all_fields().nth(usize::from(field)).unwrap();
            let field_def_name = field_def.name;
            eprintln!("projecting into {adt_def:?}.{field_def_name:}");
            let res = *self.gacx.field_tys.get(&field_def.did).unwrap_or_else(|| {
                panic!("Could not find {adt_def:?}.{field_def_name:?} in field type map")
            });
            res
        };
```

See comment in https://github.com/immunant/c2rust/pull/726#pullrequestreview-1182354527.

c2rust-analyze/src/context.rs
```diff
@@ -357,7 +361,16 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
     }
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
-        util::lty_project(lty, proj)
+        let adt_func = |adtdef: AdtDef, field: Field| {
+            let fielddef = adtdef.all_fields().nth(usize::from(field)).unwrap();
+            let fielddef_name = fielddef.name;
+            eprintln!("projecting into {adtdef:?}.{fielddef_name:}");
```
kkysen, line 367: Is there any logging we should be using or is `eprintln!` fine for now?

c2rust-analyze/src/main.rs
```diff
@@ -150,6 +150,18 @@ fn run(tcx: TyCtxt) {
...
+    for ldid in tcx.hir_crate_items(()).definitions() {
+        let did = ldid.to_def_id();
+        if tcx.def_kind(did) != DefKind::Struct {
+            continue;
+        }
+        for field in tcx.adt_def(did).all_fields() {
+            let lty = gacx.assign_pointer_ids(tcx.type_of(field.did));
+            gacx.field_tys.insert(field.did, lty);
+        }
+    }
```
kkysen, line 163: Could do this (just a suggestion):

```suggestion
    gacx.field_tys = tcx
        .hir_crate_items(())
        .definitions()
        .map(|ldid| ldid.to_def_id())
        .filter(|did| tcx.def_kind(did) == DefKind::Struct)
        .flat_map(|did| tcx.adt_def(did).all_fields())
        .map(|field| {
            let lty = gacx.assign_pointer_ids(tcx.type_of(field.did));
            (field.did, lty)
        })
        .collect();
```

c2rust-analyze/src/main.rs
```diff
@@ -150,6 +150,18 @@ fn run(tcx: TyCtxt) {
...
+    // Label the field types of each struct.
+    for ldid in tcx.hir_crate_items(()).definitions() {
+        let did = ldid.to_def_id();
+        if tcx.def_kind(did) != DefKind::Struct {
+            continue;
+        }
+        for field in tcx.adt_def(did).all_fields() {
+            let lty = gacx.assign_pointer_ids(tcx.type_of(field.did));
+            gacx.field_tys.insert(field.did, lty);
+        }
```
kkysen, line 162: Could you extract out a `gacx.label_field(field)` and `gacx.label_struct_fields(did)` method?  Not sure if it's that important, but I find it cleaner and easier to read/understand later.  Though this `run` function is quite long and I'd have the same comment for much of it, but I don't think it's worth it to change yet.

c2rust-analyze/src/util.rs
kkysen: Is this because `ProjectionElem::Field` can't be called on an `enum`, only a `struct`?

c2rust-analyze/src/context.rs
```diff
@@ -357,7 +361,16 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
     }
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
-        util::lty_project(lty, proj)
+        let adt_func = |adtdef: AdtDef, field: Field| {
+            let fielddef = adtdef.all_fields().nth(usize::from(field)).unwrap();
```
kkysen, line 365:
```suggestion
            let fielddef = &adtdef.variant(0usize.into()).fields[field.index()];
```

If `adt_func` knows it's being called on a `struct` and not an `enum`, then this is clearer and more efficient.

kkysen:
How are fields within a variant handled?  Both in those code and in things like `ProjectionElem` and MIR in general?

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -34,7 +34,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
     pub fn visit_place(&mut self, pl: Place<'tcx>) -> LTy<'tcx> {
         let mut lty = self.local_ltys[pl.local.index()];
         for proj in pl.projection {
-            lty = util::lty_project(lty, &proj);
+            lty = util::lty_project(lty, &proj, |_, _| panic!("Adt not supported"));
```
kkysen, line 37: Why is it not supported here?

c2rust-analyze/src/main.rs
```diff
@@ -150,6 +150,18 @@ fn run(tcx: TyCtxt) {
         gacx.fn_sigs.insert(ldid.to_def_id(), lsig);
     }
 
+    // Label the field types of each struct.
+    for ldid in tcx.hir_crate_items(()).definitions() {
+        let did = ldid.to_def_id();
+        if tcx.def_kind(did) != DefKind::Struct {
+            continue;
+        }
+        for field in tcx.adt_def(did).all_fields() {
```
kkysen, line 159: Where's the definition of `tcx.adt_def`?  I can't find it buried within the query system.  Why does it not return an `Option`?

c2rust-analyze/src/main.rs
```diff
@@ -150,6 +150,18 @@ fn run(tcx: TyCtxt) {
         gacx.fn_sigs.insert(ldid.to_def_id(), lsig);
     }
 
+    // Label the field types of each struct.
+    for ldid in tcx.hir_crate_items(()).definitions() {
+        let did = ldid.to_def_id();
+        if tcx.def_kind(did) != DefKind::Struct {
+            continue;
+        }
```
kkysen, line 158: Maybe just remove this check and always use `.all_fields()` (superseding my previous comments if we do it this way).  That way anything should hopefully be handled, like variant fields.

kkysen: If we don't do it this way, then to keep parity with `adt_func`, I'd use `.variant(0usize.into()).fields` instead of `.all_fields()` again here.

kkysen (commented):
See my comments mainly about variants.

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
spernsteiner: I think field `DefId` is the right way to go.

c2rust-analyze/src/util.rs
spernsteiner: `ProjectionElem::Field` can be called after using `ProjectionElem::Downcast` to convert an enum to a specific variant of that enum.  For example, the MIR for this code:
```Rust
fn f(x: Option<i32>) -> i32 {
    match x {
        Some(y) => y,
        None => 0,
    }
}
```
includes this line in the `Some` case:
```Rust
_3 = ((_1 as Some).0: i32);
```
The `Place` on the RHS consists of the local `_1`, a `ProjectionElem::Downcast` (`... as Some`), and a `ProjectionElem::Field` (`(...).0`).

c2rust-analyze/src/context.rs
spernsteiner: I've just been using `eprintln!` for now, with the assumption that we'll someday set up a proper logging framework and do a big search-and-replace to convert them all to `debug!`/`trace!`.

c2rust-analyze/src/main.rs
spernsteiner: Looks like the real definition is [here](https://doc.rust-lang.org/stable/nightly-rustc/rustc_typeck/collect/fn.adt_def.html).  I found this by searching the docs for things named `adt_def` and scrolling down until I found a free function by that name (usually query implementations are defined as free functions with the same name as the query).  It happens to be pretty far down the list in this case.

>Why does it not return an Option?

It panics on failure instead.  You're supposed to ensure the precondition some other way before calling (as we do by checking `tcx.def_kind(did)`).

c2rust-analyze/src/main.rs
spernsteiner: I think we do need this check so we don't call `adt_def` on non-ADTs (I don't know of a way to enumerate only ADTs in the crate, only a way to enumerate all items).

c2rust-analyze/src/context.rs
```diff
@@ -137,6 +140,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx: _,
             lcx,
             ref mut fn_sigs,
+            field_tys: _,
```
spernsteiner, line 143: `field_tys` need to be relabeled like `fn_sigs` here.

spernsteiner:
There's a general question here of how much enum support should be included in this PR and how much should be left to a future branch.  Since we'll almost certainly need enum support eventually (`c2rust-transpile` uses `Option` for function pointers, at the very least), my inclination personally would be to include it anywhere that it's easy to do, but I think it's also fine to leave it out (subject to the usual requirement that if the analysis sees something unsupported, it should panic rather than trying to muddle through and potentially producing weird results).

Also, please be careful about the use of `Field` indexes.  That type specifically represents a field index within some variant, not within the whole ADT, so indexing into `all_fields()` using a `Field` index (e.g. `all_fields().nth(field.as_usize())`) is nonsensical.  If the idea is to support only structs for now, use `adt_def.non_enum_variant().fields[field.as_usize()]`, which asserts that there is only one variant and then indexes into that variant's field list.

c2rust-analyze/src/main.rs
kkysen: Is there a way to expand this check to all ADTs, not just structs?

kkysen:
> There's a general question here of how much enum support should be included in this PR and how much should be left to a future branch.

I think where only supporting structs and not enums is more work, we should just support both.  I'm still not sure to what extent things are like that, but since `AdtDef` seems oriented toward handling them all at once, I feel like it may just be simpler to handle them all at once.  If not, that's fine, too, though.

c2rust-analyze/src/util.rs
kkysen: In that above case, what would the `f` be in the `ProjectionElem::Field(f, _)`?

c2rust-analyze/src/context.rs
kkysen: Actually, use `.non_enum_variant()` here as @spernsteiner said.  I didn't realize that that was a method.

spernsteiner:
>since `AdtDef` seems oriented toward handling them all at once, I feel like it may just be simpler to handle them all at once

The trouble with enums is `ProjectionElem::Downcast`, which is annoying to handle.  It introduces a notion of a "variant type" (e.g. the type of `Option`s that are statically known to be `Some`), but variant types don't actually exist in the type system (not even in rustc's internal representation of it), so as a workaround you have to carry around both a type and a variant index if it's been restricted this way.  This is why the [`Place::ty` method](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/struct.Place.html#method.ty) returns a `PlaceTy` (`Ty` + `Option<VariantIdx>`) instead of a regular `Ty`.  Handling enums fully would require some refactoring - we would need something like an `LPlaceTy`, a pair of an `LTy` and `Option<VariantIdx>`.

c2rust-analyze/src/main.rs
```diff
@@ -150,6 +151,23 @@ fn run(tcx: TyCtxt) {
...
+    // Label the field types of each struct.
+    for defldid in tcx.hir_crate_items(()).definitions() {
+        let defdid = defldid.to_def_id();
+        if let DefKind::Struct = tcx.def_kind(defdid) {
+            let defty = tcx.type_of(defdid).kind();
+            if let TyKind::Adt(adtdef, _) = defty {
+                for (fid, field) in adtdef.all_fields().enumerate() {
+                    let lty = gacx.assign_pointer_ids(tcx.type_of(field.did));
+                    gacx.field_tys.insert(
+                        (*adtdef, Field::from_usize(fid), VariantIdx::from_usize(0)),
```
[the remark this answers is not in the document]
aneksteind: This is overcome by events with the new key types with Field `DefId`

c2rust-analyze/src/main.rs
aneksteind: the parts of me that like functional programming like this, but i feel like the way it is now is more scrutable

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: It wasn't supported before this PR either. I was planning on adding support for this either here or in another PR

c2rust-analyze/src/main.rs
aneksteind: does cc7d9c5c cover all possibilities?

c2rust-analyze/src/util.rs
spernsteiner: 0, matching the `.0` in the pretty-printed version.

This is more interesting with `Result` the two cases you get are these:
```Rust
// Ok case
_3 = ((_1 as Ok).0: i32);
// Err case
_4 = ((_1 as Err).0: u8);
```
Each variant has only one field, so both `Field`s use index 0.  They're distinguished instead by the `VariantIdx` used in the `ProjectionElem::Downcast`.

c2rust-analyze/src/main.rs
spernsteiner: I don't think `Variant`s have their own `AdtDef`s - they only exist as part of an enum.  If you run this on code that defines an enum, then I'd expect it to panic on the `adt_def` call below.  Only `Struct`, `Enum`, and `Union` should be accepted here.

aneksteind: Ok, fixed in https://github.com/immunant/c2rust/commit/3e040e817d91b8ab14f4247f5f54e431e9ce8411

c2rust-analyze/src/context.rs
aneksteind: let me know if 7bf10839 doesn't address this correctly

aneksteind:
> but I think it's also fine to leave it out (subject to the usual requirement that if the analysis sees something unsupported, it should panic rather than trying to muddle through and potentially producing weird results)

@spernsteiner I'd like to go the route of deferring support and creating an issue for it, and focus on the aspects of `lighttpd` that we find need support (if this is one of them then so be it, but would still like to address in another PR)

c2rust-analyze/src/context.rs
spernsteiner: The `fn_sigs` case looks like this:
```Rust
sig.output = remap_lty_pointers(lcx, map, sig.output);
```
So I think `remap_lty_pointers` just returns the remapped `LTy`, and you have to assign it back into `*labeled_field`.

spernsteiner:
>I'd like to go the route of deferring support and creating an issue for it

Seems reasonable to me.

c2rust-analyze/src/context.rs
aneksteind: addressed in 6b91f4bf

spernsteiner (approved):

c2rust-analyze/src/context.rs
```diff
@@ -357,7 +377,16 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
     }
 
     pub fn project(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
-        util::lty_project(lty, proj)
+        let adt_func = |adtdef: AdtDef, field: Field| {
```
kkysen, line 380: nit: there's still `adtdef` => `adt_def`.

kkysen (approved):

PR #727, opened 2022-11-15
Provide an initial stubbed `lighttpd` connection path

aneksteind:
Provide an initial stubbed `lighttpd` connection path.

This will continue to be split out into multiple versions as described in #728.  It'll also be moved to a separate crate (not the existing `analysis/test`), tested for no UB with `miri`, and a script added to run `c2rust-analyze` on it/them).

### DIFF
--- a/analysis/test/src/lighttpd.rs
+++ b/analysis/test/src/lighttpd.rs
@@ -1,0 +1,149 @@
+extern crate libc;
+
+use libc::*;
+use std::mem;
+pub type size_t = libc::c_ulong;
+
+extern "C" {
+    fn malloc(_: libc::c_ulong) -> *mut libc::c_void;
+    fn realloc(_: *mut libc::c_void, _: libc::c_ulong) -> *mut libc::c_void;
+    fn free(__ptr: *mut libc::c_void);
+    fn calloc(_: libc::c_ulong, _: libc::c_ulong) -> *mut libc::c_void;
+}
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct connection {
+    pub fd: libc::c_int,
+    pub fdn: *mut fdnode,
+}
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct server {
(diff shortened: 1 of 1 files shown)


spernsteiner:
This looks like a reasonable next step for the static analysis.  The one thing I'd suggest changing is removing the `handler` and `ctx` fields, at least for now.  That will let us focus on structs and malloc/free first, and once that's working, we can move on to the function pointer and `void*` cast support that's required for `handler` and `ctx`.

kkysen:
How do we test static analysis against this?

aneksteind:
@kkysen currently:
```
$ cd c2rust/analysis
$ cargo run --bin c2rust-analyze -- ../analysis/test/src/lighttpd.rs -L "$(rustc --print sysroot)/lib/rustlib/x86_64-unknown-linux-gnu/lib" --crate-type rlib
```

kkysen (changes requested):
I think this looks good.  A couple of comments:

* We need to make sure no UB was introduced, like had been in `pointers.rs` when manually adding in code, including from `lighttpd`, especially if that UB is not present in the original C code or transpiled Rust, but is manually introduced by mistake.
* Can we add that command to run static analysis on this as a bash script?
  * This is the only (simple) change for this PR I think we need now.
* I'll file an issue to track splitting this into a working (for static analysis) subset, the original `lighttpd` version, and this version that I'll gradually add more to from the `lighttpd` code.

kkysen (approved):
I'm going to just merge this as is, and then open other PRs to move things around, add the script, split it into multiple versions, etc.

kkysen:
> This looks like a reasonable next step for the static analysis. The one thing I'd suggest changing is removing the `handler` and `ctx` fields, at least for now. That will let us focus on structs and malloc/free first, and once that's working, we can move on to the function pointer and `void*` cast support that's required for `handler` and `ctx`.

@spernsteiner, we'll address this in a later PR once we split things up as described in #728.

Issue #728, opened 2022-11-16
Split the `lighttpd.rs` in #727 into a currently working and full version


kkysen:
Split the `lighttpd.rs` in #727 into a currently working and full version.

That is, we should have 3 versions:
* the subset of `lighttpd.rs` that currently works with `c2rust-analyze`.  This will be updated to include parts that start work as fixes are made to `c2rust-analyze`.
* the maximal `lighttpd.rs` that is taken directly from `lighttpd` and contains all of code (for the function(s) we're looking at for now), nothing stripped down and pared yet.
* the current `lighttpd.rs`, which we can use for experimentation.

aneksteind:
Could you help me understand why the third option is useful? Are you envisioning a workflow where we take the most recent currently-working file and add features from the version containing all the code to play with locally, without accidentally committing changes to the former?

kkysen:
We can see as we go, but I imagined a 3rd version that could be anything in-between could be useful.  If ends up not being, we can remove it.  I do think we need it at least at the very beginning, though, as I want to split out the min version from it, and split out the max version from it, but without changing itself at all.  So that'll leave us with 3 versions, and if we realize the 3rd middle one isn't useful at all, we can get rid of it.

PR #729, opened 2022-11-20
Move `analysis/test` to `analysis/tests/misc` to prepare for more tests

kkysen:
This moves `analysis/test` to `analysis/tests/misc` to prepare for more tests.  This is in preparation for #728, where I'll add separate `lighttpd` tests.

In order to stay as close as possible to the original transpiled `lighttpd` code, we'll need to put duplicate `#[no_mangle]` functions in separate crates, which will be created as directories in `analysis/tests`.  I think putting them in separate crates is a good idea regardless of the `#[no_mangle]` issue, as when we later want to test everything with dynamic analysis (`c2rust-instrument`), too, it operates on crates, and eventually, I think we'd like for `c2rust-analyze` to do so, too.


kkysen:
Is `misc` a good name for the current test?  Other tests will be `lighttpd` (split into things like `lighttpd.passing` and `lighttpd.original`, or whatever we want to name them, for #728 ).  The current test seems like a bunch of things thrown together so I thought `misc` could be a good name.

aneksteind:
I'm not sure how I feel about the `misc` naming, but I'll defer to @spernsteiner

aneksteind (approved):
we can figure out naming later

PR #732, opened 2022-11-20
Remove `#![feature(rustc_private)]` from `analysis/tests`

kkysen:
This removes `#![feature(rustc_private)]` from `analysis/tests` as there's no reason to use rustc internals in test code.  I'm not sure why it was added in the first place in f983821e0fdc9ba80a6a3787680857b2f68a7074, as things build perfectly fine without it, and I'm not sure why we would need rustc internals in the test code.


kkysen:
Merging this as #750, a superset, has been approved.

PR #741, opened 2022-11-23
introduce minimal lighttpd example that does not break static analysis

aneksteind:
As can be seen, not much works at the moment. But this gives us a starting point.

### DIFF
--- a/analysis/tests/lighttpd/src/lighttpd-minimal.rs
+++ b/analysis/tests/lighttpd/src/lighttpd-minimal.rs
@@ -1,0 +1,67 @@
+#![feature(rustc_private)]
+
+extern crate libc;
+
+use libc::*;
+use std::mem;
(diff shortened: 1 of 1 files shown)


kkysen:
I was going to put the different versions in separate crates.

aneksteind:
@kkysen feel free to commit to this PR, my primary concern was putting up a minimal working example

kkysen:
Okay, I'll move things around a bit and then merge it.

kkysen (approved):
Merging this now as is, then fixing things.

analysis/tests/lighttpd/src/lighttpd-minimal.rs
```diff
@@ -0,0 +1,67 @@
+#![feature(rustc_private)]
```
kkysen, line 1: Why is this needed?  Can I delete it?

PR #746, opened 2022-11-27
Add static analysis support for limited set of previously-unknown functions

### LINKED ISSUE #734: static analysis does not support `std::mem::size_of` or other unknown functions
In the following example:

```rust
unsafe extern "C" fn fdnode_init() -> *mut fdnode {
    let fdn: *mut fdnode = calloc(
        1 as libc::c_int as libc::c_ulong,
        ::std::mem::size_of::<fdnode>() as libc::c_ulong,
    ) as *mut fdnode;
    if fdn.is_null() {
        println!("It's null");
    }
    let x = fdevents {
        fdarray: 0 as *mut *mut fdnode,
    };
    return fdn;
}
```

the static analyzer panics:

```
visit_terminator(_7 = std::mem::size_of::<fdnode_st>() -> [return: bb1, unwind: bb14])
callee = Some(Other { def_id: DefId(2:2156 ~ core[480a]::mem::size_of), substs: [fdnode_st] })
thread 'rustc' panicked at 'not yet implemented: call to unknown function DefId(2:2156 ~ core[480a]::mem::size_of)', c2rust-analyze/src/dataflow/type_check.rs:308:21
```

aneksteind:
Fixes #734.

Makes static analysis aware of the following: `malloc`, `calloc`, `realloc`, `free` `ptr.is_null`.

Also applies pointer flow rules to those functions.

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -175,9 +175,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 self.ltcx.label(ty, &mut |_| Label::default())
             }
 
-            Rvalue::UnaryOp(_, ref op) => {
-                self.visit_operand(op)
-            }
+            Rvalue::UnaryOp(_, ref op) => self.visit_operand(op),
 
             ref rv => panic!("unsupported rvalue {:?}", rv),
         }
@@ -247,18 +245,22 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                     Some(Callee::Realloc) => {
                         // We handle this like a pointer assignment.
                         let pl_lty = self.visit_place(destination);
-                        assert!(args.len() == 2);
-                        let rv_lty = self.visit_operand(&args[0]);
+                        let rv_lty = assert_matches!(&args[..], [p, _] => {
+                            self.visit_operand(p)
+                        });
+
                         self.do_assign(pl_lty, rv_lty);
                     }
                     Some(Callee::Free) => {
                         let _pl_lty = self.visit_place(destination);
-                        assert!(args.len() == 1);
-                        let _rv_lty = self.visit_operand(&args[0]);
+                        let _rv_lty = assert_matches!(&args[..], [p] => {
+                            self.visit_operand(p)
+                        });
                     }
                     Some(Callee::IsNull) => {
-                        assert!(args.len() == 1);
-                        let _rv_lty = self.visit_operand(&args[0]);
+                        let _rv_lty = assert_matches!(&args[..], [p] => {
+                            self.visit_operand(p)
+                        });
                     }
                     None => {}
                 }
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -2,6 +2,7 @@ use crate::borrowck::atoms::{AllFacts, AtomMaps, Loan, Origin, Path, Point, SubP
 use crate::borrowck::{LTy, LTyCtxt, Label};
 use crate::context::PermissionSet;
 use crate::util::{self, Callee};
+use assert_matches::assert_matches;
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::{
     AggregateKind, BinOp, Body, BorrowKind, Local, LocalDecl, Location, Operand, Place, Rvalue,
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -175,6 +175,10 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 self.ltcx.label(ty, &mut |_| Label::default())
             }
 
+            Rvalue::UnaryOp(_, ref op) => {
+                self.visit_operand(op)
+            }
+
             ref rv => panic!("unsupported rvalue {:?}", rv),
         }
     }
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -238,6 +238,28 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                     Some(Callee::Other { .. }) => {
                         // TODO
                     }
+                    Some(Callee::Malloc) => {
+                        // TODO
+                    }
+                    Some(Callee::Calloc) => {
+                        // TODO
+                    }
+                    Some(Callee::Realloc) => {
+                        // We handle this like a pointer assignment.
+                        let pl_lty = self.visit_place(destination);
+                        assert!(args.len() == 2);
+                        let rv_lty = self.visit_operand(&args[0]);
+                        self.do_assign(pl_lty, rv_lty);
+                    }
+                    Some(Callee::Free) => {
+                        let _pl_lty = self.visit_place(destination);
+                        assert!(args.len() == 1);
+                        let _rv_lty = self.visit_operand(&args[0]);
+                    }
+                    Some(Callee::IsNull) => {
+                        assert!(args.len() == 1);
+                        let _rv_lty = self.visit_operand(&args[0]);
+                    }
                     None => {}
                 }
             }
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -170,16 +170,20 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         }
     }
 
-    fn do_assign(&mut self, pl_lty: LTy<'tcx>, rv_lty: LTy<'tcx>) {
-        // If the top-level types are pointers, add a dataflow edge indicating that `rv` flows into
-        // `pl`.
-        self.do_assign_pointer_ids(pl_lty.label, rv_lty.label);
-
+    fn do_equivalence_nested(&mut self, pl_lty: LTy<'tcx>, rv_lty: LTy<'tcx>) {
         // Add equivalence constraints for all nested pointers beyond the top level.
         assert_eq!(pl_lty.ty, rv_lty.ty);
         for (&pl_sub_lty, &rv_sub_lty) in pl_lty.args.iter().zip(rv_lty.args.iter()) {
             self.do_unify(pl_sub_lty, rv_sub_lty);
         }
+    }
+
+    fn do_assign(&mut self, pl_lty: LTy<'tcx>, rv_lty: LTy<'tcx>) {
+        // If the top-level types are pointers, add a dataflow edge indicating that `rv` flows into
+        // `pl`.
+        self.do_assign_pointer_ids(pl_lty.label, rv_lty.label);
+
+        self.do_equivalence_nested(pl_lty, rv_lty);
     }
 
     /// Add a dataflow edge indicating that `rv_ptr` flows into `pl_ptr`.  If both `PointerId`s are
@@ -303,15 +307,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                         self.constraints.add_all_perms(rv_lty.label, perms);
 
                         // unify inner-most pointer types
-                        let mut rv_iter = rv_lty.iter();
-                        let mut pl_iter = pl_lty.iter();
-                        let _outermost_rv = rv_iter.next();
-                        let _outerpost_pl = pl_iter.next();
-                        // TODO: handle return types that are different:
-                        // realloc may return a *const void but take in *const *mut ...
-                        for (rv_inner, pl_inner) in rv_iter.zip(pl_iter) {
-                            self.do_unify(rv_inner, pl_inner);
-                        }
+                        self.do_equivalence_nested(pl_lty, rv_lty);
                     }
                     Some(Callee::Free) => {
                         self.visit_place(destination, Mutability::Mut);
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -284,6 +284,35 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
                     Some(Callee::MiscBuiltin) => {}
 
+                    Some(Callee::Malloc) => {
+                        self.visit_place(destination, Mutability::Mut);
+                    }
+
+                    Some(Callee::Calloc) => {
+                        self.visit_place(destination, Mutability::Mut);
+                    }
+                    Some(Callee::Realloc) => {
+                        // We handle this like a pointer assignment.
+                        self.visit_place(destination, Mutability::Mut);
+                        let pl_lty = self.acx.type_of(destination);
+                        assert!(args.len() == 2);
+                        self.visit_operand(&args[0]);
+                        let rv_lty = self.acx.type_of(&args[0]);
+                        self.do_assign(pl_lty, rv_lty);
+                        let perms = PermissionSet::OFFSET_ADD | PermissionSet::OFFSET_SUB;
+                        self.constraints.add_all_perms(rv_lty.label, perms);
+                    }
+                    Some(Callee::Free) => {
+                        self.visit_place(destination, Mutability::Mut);
+                        assert!(args.len() == 1);
+                        self.visit_operand(&args[0]);
+                    }
+
+                    Some(Callee::IsNull) => {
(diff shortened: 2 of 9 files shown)


c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -234,6 +238,28 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        // TODO
+                    }
+                    Some(Callee::Calloc) => {
+                        // TODO
+                    }
+                    Some(Callee::Realloc) => {
+                        // We handle this like a pointer assignment.
+                        let pl_lty = self.visit_place(destination);
+                        assert!(args.len() == 2);
+                        let rv_lty = self.visit_operand(&args[0]);
```
kkysen, line 251: Could we use [`assert_matches!`](https://docs.rs/assert_matches/latest/assert_matches/macro.assert_matches.html) to keep this in sync (the `.len()` `assert!` and the indexing)?

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -234,6 +238,28 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                         // TODO
                     }
+                    Some(Callee::Malloc) => {
+                        // TODO
+                    }
+                    Some(Callee::Calloc) => {
+                        // TODO
+                    }
+                    Some(Callee::Realloc) => {
+                        // We handle this like a pointer assignment.
```
kkysen, line 248: What does this mean?

c2rust-analyze/src/util.rs
```diff
@@ -85,6 +85,16 @@ pub enum Callee<'tcx> {
     },
     /// A built-in or standard library function that requires no special handling.
     MiscBuiltin,
+    /// libc::malloc
+    Malloc,
+    /// libc::calloc
+    Calloc,
```
kkysen, line 91: Eventually, I think we'd want to just include everything with `__attribute__ ((__malloc__))`.

c2rust-analyze/src/util.rs
```diff
@@ -175,7 +185,22 @@ fn builtin_callee<'tcx>(
...
+
+        "malloc" => Some(Callee::Malloc),
+
+        "calloc" => Some(Callee::Calloc),
+
+        "realloc" => Some(Callee::Realloc),
+
+        "free" => Some(Callee::Free),
+
+        "is_null" => Some(Callee::IsNull),
```
kkysen, line 198: Why aren't we matching on canonical paths?

analysis/tests/lighttpd/src/lighttpd-minimal.rs
```diff
@@ -50,18 +50,55 @@ pub struct fdnode_st {
 
 pub type fdnode = fdnode_st;
 
-unsafe extern "C" fn connection_close(mut srv: *mut server, mut con: *mut connection) {}
+unsafe extern "C" fn fdnode_init() -> *mut libc::c_void /*TODO: handle *mut fdnode */ {
```
kkysen, line 53: For all of these changes, I want to land a `lighttpd` minimal test for `c2rust-analyze` in CI first, just checking that it doesn't crash on this as input.  We can merge this earlier if it's blocking anything and I don't get to adding the test in time, but hopefully I'll be able to.

kkysen:
Should be rebased onto #749, when that lands or now if you want.

c2rust-analyze/src/util.rs
aneksteind: can you explain what you mean?

kkysen: Why just `size_of` instead of `::core::mem::size_of`?  This would break if the user defined some other `size_of` function, right?

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: the permissions of the pointer used as an argument to `realloc` are also applied to its return value. although, @spernsteiner, is this the correct behavior?

c2rust-analyze/src/util.rs
aneksteind: I'm not exactly sure what you mean by that

c2rust-analyze/src/borrowck/type_check.rs
kkysen: Ah, that makes more sense.  Though I would re-phrase the comment as "We handle the pointer arg and returned pointer as a pointer assignment." or something similar.

c2rust-analyze/src/util.rs
kkysen: Eventually, I'd imagine that we would propagate C attributes like `__attribute__((__malloc__))` onto the transpiled Rust, which gives us more information about the meaning of these functions.  Any other `__attribute__((__malloc__))` function should be treated like `malloc`.  This is not really important at the moment, though, but I just wanted to point it out for the eventual future.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: added in 19fcf8ff

c2rust-analyze/src/main.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(rustc_private)]
+#[macro_use] extern crate assert_matches;
```
kkysen, line 2: I don't believe either `#[macro_use]` or `extern crate` are needed anymore.  I think `extern crate` is only needed for `rustc_private` crates.

aneksteind: i'm not sure what you're suggesting, to use `use assert_matches`? `std::assert_matches::assert_matches` doesn't have the same functionality of this crate and the [docs](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) have it being used in this way.

kkysen: `#[macro_use]` is for Rust 2015.  It's not needed at all in Rust 2018 and 2021.  That just means the docs are very old.

kkysen: `extern crate` is also not needed anymore in most cases (except for things like `rustc_private`) in Rust 2018 and later.

c2rust-analyze/src/borrowck/type_check.rs
spernsteiner: >is this the correct behavior?

`realloc` potentially invalidates the incoming pointer, so I think we need to treat it as a combined `free` and `malloc`.  Since the data inside the allocation is either preserved / copied over, we also need to unify non-outermost pointer types, similar to an assignment.

For a `realloc` that takes a `*mut /*1*/ *mut /*2*/ i32` and returns a `*mut /*3*/ *mut /*4*/ i32`, `PointerId(2)` and `PointerId(4)` (the non-outermost pointer types) should be unified, `PointerId(1)` should be treated like an argument to `free` (probably requiring `FREE` permission), and `PointerId(3)` should be treated like the return value of a `malloc` (which I don't think requires any special handling).

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -284,6 +284,35 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
 
+                    Some(Callee::Malloc) => {
+                        self.visit_place(destination, Mutability::Mut);
+                    }
+
+                    Some(Callee::Calloc) => {
+                        self.visit_place(destination, Mutability::Mut);
+                    }
+                    Some(Callee::Realloc) => {
+                        // We handle this like a pointer assignment.
```
spernsteiner, line 295: `realloc` potentially invalidates the incoming pointer, so I think we need to treat it as a combined `free` and `malloc`.  Since the data inside the allocation is either preserved / copied over, we also need to unify non-outermost pointer types, similar to an assignment.

For a `realloc` that takes a `*mut /*1*/ *mut /*2*/ i32` and returns a `*mut /*3*/ *mut /*4*/ i32`, `PointerId(2)` and `PointerId(4)` (the non-outermost pointer types) should be unified, `PointerId(1)` should be treated like an argument to `free` (probably requiring `FREE` permission), and `PointerId(3)` should be treated like the return value of a `malloc` (which I don't think requires any special handling).

c2rust-analyze/src/borrowck/type_check.rs
spernsteiner: For borrowck purposes, I think `realloc` should be treated as a combined `free` and `malloc`.  I don't think `malloc` needs any special handling (it returns a fresh `UNIQUE` pointer that doesn't alias anything).  And I'm currently thinking that `free` doesn't need any special handling either (we could have it require `UNIQUE` / treat it as a mutable access, but I think `UNIQUE` is not a strong enough property to be useful here).

I'm not totally sure about this approach, so please add a few test cases that use malloc/realloc/free (with assertions about the pointer permissions, along the lines of `tests/filecheck/alias1.rs`), and we can check if they produce reasonable results.  I'd particularly like to see how it works on this "growable vector" pattern:
```C
for (...) {
    if (len == capacity) {
        capacity *= 2;
        ptr = realloc(ptr, capacity);
    }
    // Add a new element; increase len
}
```

spernsteiner: IIRC, casts to and from `void*` are still not well supported, so for testing purposes, I think it would be reasonable to declare some specialized `malloc` functions in the test case:
```Rust
extern "C" {
    fn c2rust_test_typed_malloc(len: usize) -> *mut i32;
    fn c2rust_test_typed_realloc(ptr: *mut i32, len: usize) -> *mut i32;
    fn c2rust_test_typed_free(ptr: *mut i32);
}
```
And modify `util.rs` to recognize these as `Callee::Malloc` and such.

kkysen: Note that `realloc(NULL, size)` is also pretty much equivalent to `malloc(size)`, so we need to handle that case, too.

c2rust-analyze/src/util.rs
aneksteind: I tested grabbing the `link_name` via `tcx.get_attr(did, rustc_span::sym::link_name)` and it's only `Some(Attribute)` if it's explicitly placed, e.g.
```rust
extern "C" {
    #[link_name = "malloc"]
    fn malloc(_: libc::c_ulong) -> *mut libc::c_void;
...
```

aneksteind: Also, when checking if the `DefId` of `malloc` for example is local, `did.is_local()` returns true and `DefKind` is `Fn` and not `ForeignTy`, which seems counterintuitive for a declaration in an extern block. How else can I discover that `malloc` is an extern fn here?

spernsteiner: >I tested grabbing the `link_name` via `tcx.get_attr(did, rustc_span::sym::link_name)` and it's only `Some(Attribute)` if it's explicitly placed

If `link_name` is not present, then the symbol name is the name of the `fn` itself.

Also, there's a `tcx.codegen_fn_attrs` query that might be more convenient for getting the `link_name` (you don't have to parse the attribute tokens yourself that way).

>`DefKind` is `Fn` and not `ForeignTy`

`ForeignTy` is for extern types, e.g. `extern "C" { type Foo; }`.  If extern fns don't have a separate `DefKind`, you might have to check the `DefKind` of the parent item instead to see if it's an `extern` block.

aneksteind: Let me know if 85f32f91 captures what's desired

gstrauss:
Fun!  I am a lighttpd developer and while I do not know Rust, if you have any questions about lighttpd code, or have suggestions for C idioms that better translate to Rust, please tag me.  Thanks!

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: is something like 4ddbd096 what you had in mind?

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: @spernsteiner would you mind letting me know if these new tests in d5cda038 capture what you expect?

analysis/tests/lighttpd/src/lighttpd-minimal.rs
aneksteind: I believe #757 does this now

c2rust-analyze/src/borrowck/type_check.rs
spernsteiner: Yes, those tests look right to me.  I'm not sure why the `realloc` test has `i` as non-`UNIQUE`, but we can investigate and fix that separately.

c2rust-analyze/src/util.rs
spernsteiner: That seems fine for now.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -284,6 +284,50 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        // unify inner-most pointer types
+                        let mut rv_iter = rv_lty.iter();
+                        let mut pl_iter = pl_lty.iter();
+                        let _outermost_rv = rv_iter.next();
+                        let _outerpost_pl = pl_iter.next();
+                        // TODO: handle return types that are different:
+                        // realloc may return a *const void but take in *const *mut ...
+                        for (rv_inner, pl_inner) in rv_iter.zip(pl_iter) {
+                            self.do_unify(rv_inner, pl_inner);
+                        }
```
spernsteiner, line 314: This works, though the version in `do_assign` is more concise (and could probably be pulled out into a helper function).

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -284,6 +284,50 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        let perms = PermissionSet::FREE;
+                        self.constraints.add_all_perms(rv_lty.label, perms);
+
+                        // unify inner-most pointer types
+                        let mut rv_iter = rv_lty.iter();
+                        let mut pl_iter = pl_lty.iter();
+                        let _outermost_rv = rv_iter.next();
+                        let _outerpost_pl = pl_iter.next();
+                        // TODO: handle return types that are different:
+                        // realloc may return a *const void but take in *const *mut ...
```
spernsteiner, line 311: I don't think this comment applies.  C `realloc` takes and returns `*mut c_void` - if the C code passes a non-`void*` pointer, the Rust code will have explicit casts before and after the call.  And our custom `typed_realloc` for testing should similarly take and return pointers of the same type.

c2rust-analyze/src/util.rs
```diff
@@ -177,7 +187,68 @@ fn builtin_callee<'tcx>(
...
+
+        "free" | "c2rust_test_typed_free" => {
+            if matches!(tcx.def_kind(tcx.parent(did)), DefKind::ForeignMod) {
+                return Some(Callee::Free);
+            }
+            None
+        }
+
+        "is_null" => {
+            // `core::ptr::is_null`
```
spernsteiner, line 234: This case should use the same logic as for `offset`, checking that the function is an inherent method of a raw pointer type.  The actual path of `is_null` is not `core::ptr::is_null` but rather something like `core::ptr::const_ptr::{{impl}}::is_null`.  Checking the parent impl's self type is more precise and robust against standard library changes.

spernsteiner (approved):
Looks good overall.  I left a few comments on minor things.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: added in 1d0ae163

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: removed in 1d0ae163

c2rust-analyze/src/util.rs
aneksteind: changed in 197ffe22

analysis/tests/lighttpd/src/lighttpd-minimal.rs
kkysen: I'll add that test now (it should be pretty simple).  #757 doesn't quite do it, because then it runs `FileCheck`, which we don't need to run.  `FileCheck` isn't set up to run in CI yet, so avoiding that if we don't need it is helpful.

PR #750, opened 2022-11-28
Remove unused attrs in `analysis/tests`

kkysen:
This removes unused attrs in `analysis/tests`.


kkysen:
Note that these deletions have to be rebased into #749 when either lands in `master`.

aneksteind (approved):

kkysen:
Also, this is a superset of #732, so I'd ideally like to get approval on that first before merging this.

PR #757, opened 2022-12-02
create symlink to static analysis test

aneksteind:


kkysen (changes requested):
Why do we need to move this?  If we do, can we do it the other way around?  The crate itself should contain the actual file I think.

kkysen:
Also, note that `FileCheck` tests don't currently run in CI since there were issues uniformly installing it on all distros.  That can be fixed, and I think I have a branch or stash somewhere for it, but if we just need to run `c2rust-analyze` without `FileCheck`, it'll work in CI, but it'll be slightly different from the current `filecheck.rs` test.

kkysen:
See the reasons above.  I'll create a new, more-specific test for this.  It should be pretty simple.

kkysen:
See the reasons above. I'll create a new, more-specific test for this. It should be pretty simple.

kkysen:
See the reasons above. I'll create a new, more-specific test for this. It should be pretty simple.

PR #767, opened 2023-01-06
Add more comprehensive searching for `FileCheck` using `$(llvm_config --bindir)/FileCheck`

### LINKED ISSUE #593: `cargo test -p c2rust-analyze` fails CI due to no `FileCheck`
On Ubuntu 18 and Debian 10, `FileCheck` is not installed by default with `llvm`, so `cargo test -p c2rust-analyze` fails.

We can either upgrade `llvm` versions on those distros (and stop supporting LLVM 6) or just install `llvm-${major}.${minor}-tools` on those old distros, which installs `FileCheck`.  I think the latter is preferable as otherwise, we have to fix the `provision_deb.sh` script to sometimes install `llvm` and sometimes `llvm-${version}`.  Furthermore, installing `llvm-${major}-tools` on newer distros is probably smarter to ensure `FileCheck` is always there.

The other problem is that `c2rust-analyze/tests/filecheck.rs` looks for `FileCheck` independently, and doesn't currently work for `FileCheck-6.0`.  Instead, we should match the `llvm-config` resolution in `c2rust-ast-exporter/build.rs` and then use `$(llvm-config --bindir)/FileCheck`, which is more reliable and always matches in behavior.

To do this, I want to move `find_llvm_config` to `rustc-private-link` or `build-utils` (whatever it's renamed to in https://github.com/immunant/c2rust/pull/591).

Blocked on:
* #641
* #618

ivyraine:
- F_ixes part of #593, using `$(llvm-config --bindir)/FileCheck` to use the same resolving logic for `llvm-config` and `FileCheck`.

### DIFF
--- a/c2rust-analyze/tests/filecheck.rs
+++ b/c2rust-analyze/tests/filecheck.rs
@@ -1,9 +1,11 @@
 use std::env;
 use std::fs;
 use std::os::unix::io::{AsRawFd, FromRawFd};
+use std::path::Path;
+use std::path::PathBuf;
 use std::process::{Command, Stdio};
 
-fn detect_filecheck() -> Option<&'static str> {
+fn detect_filecheck() -> Option<&'static Path> {
     let candidates = [
         "FileCheck",
         // Intel macOS homebrew location.
@@ -44,7 +46,7 @@ fn detect_filecheck() -> Option<&'static str> {
             .stderr(Stdio::null())
             .status();
         if result.is_ok() {
-            return Some(filecheck);
+            return Some(&Path::new(filecheck));
         }
     }
     None
@@ -55,11 +57,11 @@ fn filecheck() {
     let lib_dir = env::var("C2RUST_TARGET_LIB_DIR").unwrap();
     let lib_dir = &lib_dir;
 
-    let filecheck_bin = env::var("FILECHECK")
-        .ok()
-        .or_else(|| detect_filecheck().map(|s| s.to_owned()))
+    let filecheck_bin = env::var_os("FILECHECK")
+        .map(PathBuf::from)
+        .or_else(|| detect_filecheck().map(|it| it.to_owned()))
         .unwrap_or_else(|| panic!("FileCheck not found - set FILECHECK=/path/to/FileCheck"));
-    eprintln!("detected FILECHECK={}", filecheck_bin);
+    eprintln!("detected FILECHECK={}", filecheck_bin.display());
 
     for entry in fs::read_dir("tests/filecheck").unwrap() {
         let entry = entry.unwrap();
--- a/c2rust-analyze/tests/filecheck.rs
+++ b/c2rust-analyze/tests/filecheck.rs
@@ -59,7 +59,6 @@ fn filecheck() {
 
     let filecheck_bin = env::var_os("FILECHECK")
         .map(PathBuf::from)
-        .or_else(|| detect_filecheck().map(|it| it.to_owned()))
         .unwrap_or_else(|| {
             PathBuf::from(
                 invoke_command(find_llvm_config().as_deref(), &["--bindir"])
--- a/c2rust-analyze/tests/filecheck.rs
+++ b/c2rust-analyze/tests/filecheck.rs
@@ -1,7 +1,7 @@
+use c2rust_build_paths::{find_llvm_config, invoke_command};
 use std::env;
 use std::fs;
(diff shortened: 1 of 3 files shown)


c2rust-analyze/tests/filecheck.rs
```diff
@@ -60,7 +60,13 @@ fn filecheck() {
     let filecheck_bin = env::var_os("FILECHECK")
         .map(PathBuf::from)
         .or_else(|| detect_filecheck().map(|it| it.to_owned()))
```
kkysen, line 62: `$(llvm-config --bindir)/FileCheck` should replace `detect_filecheck`, so we no longer need to use or have `detect_filecheck`.

c2rust-analyze/tests/filecheck.rs
```diff
@@ -60,7 +60,13 @@ fn filecheck() {
...
-        .unwrap_or_else(|| panic!("FileCheck not found - set FILECHECK=/path/to/FileCheck"));
+        .unwrap_or_else(|| {
+            PathBuf::from(
+                invoke_command(find_llvm_config().as_deref(), &["--bindir"])
+                    .expect("llvm-config not found"),
+            )
```
kkysen, line 67: Need to append `/FileCheck`.

kkysen (changes requested):
Can you also remove the `--exclude c2rust-analyze` to see if this works in CI or is only a partial solution?

kkysen:
This still doesn't fix finding `FileCheck` on older LLVMs, as I thought, so revert the commits re-enabling the tests.  We'll fix that part of the issue in a separate PR.

c2rust-analyze/tests/filecheck.rs
ivyraine: fixed in [fb5cc98](https://github.com/immunant/c2rust/pull/767/commits/fb5cc98850db169a791ab7cd43d0aa66f42af379) and  [6b6c038](https://github.com/immunant/c2rust/pull/767/commits/6b6c03833edb9c177e83eed836265fa2a4376d52)

c2rust-analyze/tests/filecheck.rs
ivyraine: fixed in [7706f76](https://github.com/immunant/c2rust/pull/767/commits/7706f76352351523abae3a26f9f58859e882cd6f)

c2rust-build-paths/src/lib.rs
```diff
@@ -128,3 +128,17 @@ pub fn find_llvm_config() -> Option<PathBuf> {
...
+    S: AsRef<OsStr>,
+{
+    let command = command?;
+    let output = Command::new(command).args(args).output().ok()?;
+    if output.status.success() {
+        Some(String::from_utf8(output.stdout).unwrap().trim().to_owned())
+    } else {
+        None
+    }
+}
```
kkysen, line 144: I would just inline this.

kkysen:
```rust
let llvm_config = find_llvm_config().expect("llvm-config not found");
let output = Command::new(llvm_config)
    .args(&["--bindir"])
    .output()
    .ok()
    .filter(|output| output.status.success())
    .expect("llvm-config error");
String::from_utf8(output.stdout).unwrap().trim().to_owned()
```

ivyraine: fixed in [828243d](https://github.com/immunant/c2rust/pull/767/commits/828243d339c6993bc0274865ccb3a512af765212)

c2rust-analyze/tests/filecheck.rs
```diff
@@ -1,65 +1,30 @@
...
+            let llvm_config = find_llvm_config().expect("llvm-config not found");
+            let output = Command::new(llvm_config)
+                .args(&["--bindir"])
+                .output()
+                .ok()
+                .filter(|output| output.status.success())
+                .expect("llvm-config error");
+            PathBuf::from(String::from_utf8(output.stdout).unwrap().trim().to_owned())
+        })
+        .join("FileCheck");
```
kkysen, line 25:
```suggestion
            PathBuf::from(String::from_utf8(output.stdout).unwrap().trim().to_owned()).join("FileCheck")
        });
```

kkysen (approved):
LGTM now.  PR to fully enable `FileCheck` and restore testing is incoming.

PR #768, opened 2023-01-06
 Support struct fields in static analysis

aneksteind:
This adds support for generating concrete `Origin`s to pass to Polonius for each ADT lifetime parameter and resolving lifetime parameters to those `Origin`s when visiting the MIR body. It also adds support for ADT field pointer permissions.

There are three stages to making this possible:
1. gather hypothetical (i.e. pointer-based) and actual (reference-based) lifetimes
2. generating concrete origins
3. resolving lifetime parameters when visiting the MIR body

Stage 1 runs in a loop and gathers ADT metadata. The loop terminates when a fixed point is reached, as determined by changes in the ADT metadata. This approach is to handle ADTs that are recursive and/or mutually recursive.

Stage 2 generates one concrete `Origin` per ADT lifetime parameter, per MIR local. When grabbing the `Origin` of a given struct field, even nested ones, it is one of these concrete origins that will be returned.

Stage 3 traces the concrete `Origin`s generated for each ADT lifetime parameter all the way to the last projection / field access. It does this by updating a mapping for each projection element visited.

### DIFF
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -44,8 +44,7 @@ pub struct FieldMetadata<'tcx> {
     pub lifetime_params: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
 }
 
-/// Metadata describing the lifetime parameters and fields
-/// of a struct.
+/// Metadata describing the lifetime parameters and fields of a `struct`.
 #[derive(Clone, PartialEq, Eq)]
 pub struct AdtMetadata<'tcx> {
     /// The lifetime parameters of a structure, including
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -301,7 +301,7 @@ fn run_polonius<'tcx>(
 
     let mut loans = HashMap::new();
     // Populate `loan_issued_at` and `loans`.
-    type_check::visit(
+    type_check::visit_body(
         tcx,
         ltcx,
         &mut facts,
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -25,7 +25,7 @@ pub struct Label<'tcx> {
     pub origin: Option<Origin>,
     /// The [`Origin`]s associated with each lifetime
     /// parameter of this type, if applicable
-    pub origin_params: Option<&'tcx [(OriginParam, Origin)]>,
+    pub origin_params: &'tcx [(OriginParam, Origin)],
     pub perm: PermissionSet,
 }
 
@@ -337,35 +337,34 @@ fn assign_origins<'tcx>(
             hypothesis[lty.label]
         };
 
-        let construct_adt_origins = |ty: &Ty, amaps: &mut AtomMaps| -> Option<&_> {
-            let adt_def = ty.ty_adt_def()?;
+        let construct_adt_origins = |ty: &Ty, amaps: &mut AtomMaps| -> &[_] {
+            let adt_def = ty.ty_adt_def().unwrap();
 
             // create a concrete origin for each actual or hypothetical
             // lifetime parameter in this ADT
             let origins: Vec<_> = adt_metadata
                 .table
-                .get(&adt_def.did())?
-                .lifetime_params
-                .iter()
-                .map(|o| {
-                    let pairing = (*o, amaps.origin());
-                    eprintln!("pairing lifetime parameter with origin: {pairing:?}");
-                    pairing
+                .get(&adt_def.did())
+                .map(|adt| {
+                    adt.lifetime_params
+                        .iter()
+                        .map(|o| {
+                            let pairing = (*o, amaps.origin());
+                            eprintln!("pairing lifetime parameter with origin: {pairing:?}");
+                            pairing
+                        })
+                        .collect()
                 })
-                .collect();
+                .unwrap_or_default();
 
-            if origins.is_empty() {
-                return None;
-            }
-
-            Some(ltcx.arena().alloc_slice(&origins[..]))
+            ltcx.arena().alloc_from_iter(origins)
         };
         match lty.ty.kind() {
             TyKind::Ref(_, _, _) | TyKind::RawPtr(_) => {
                 let origin = Some(maps.origin());
                 Label {
                     origin,
-                    origin_params: None,
+                    origin_params: &[],
                     perm,
                 }
             }
@@ -377,14 +376,11 @@ fn assign_origins<'tcx>(
                     perm,
                 }
             }
-            _ => {
-                // let origin_params = construct_adt_origins(&lty.ty, maps);
-                Label {
-                    origin: None,
-                    origin_params: None,
-                    perm,
-                }
-            }
+            _ => Label {
+                origin: None,
+                origin_params: &[],
+                perm,
+            },
         }
     })
 }
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -78,7 +78,7 @@ pub enum OriginParam {
 #[derive(Hash, PartialEq, Eq, Clone, Copy)]
 pub enum OriginArg<'tcx> {
     /// An existing [`Region`], i.e. `'a` in `&'a foo`.
-    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind::ReEarlyBound`) 
+    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind::ReEarlyBound`)
     /// or [RegionKind::ReStatic](`rustc_type_ir::RegionKind::ReStatic`)
     Actual(Region<'tcx>),
     /// A hypothesized region derived from a pointer type
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -82,7 +82,8 @@ pub enum OriginParam {
 #[derive(Hash, PartialEq, Eq, Clone, Copy)]
 pub enum OriginArg<'tcx> {
     /// An existing [`Region`], i.e. `'a` in `&'a foo`.
-    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind`) or [RegionKind::ReStatic](`rustc_type_ir::RegionKind`)
+    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind::ReEarlyBound`) 
+    /// or [RegionKind::ReStatic](`rustc_type_ir::RegionKind::ReStatic`)
     Actual(Region<'tcx>),
     /// A hypothesized region derived from a pointer type
     /// e.g. `'h0` derived from the pointer in `*mut foo`
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -86,7 +86,7 @@ pub enum OriginArg<'tcx> {
     Hypothetical(i64),
 }
 
-impl<'tcx> std::fmt::Debug for OriginArg<'tcx> {
+impl<'tcx> Debug for OriginArg<'tcx> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
             OriginArg::Actual(r) => write!(f, "{:}", r),
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -36,16 +36,13 @@ pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label<'tcx>>;
 /// of a struct field.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FieldMetadata<'tcx> {
-    /// The lifetime of the field, e.g. `*mut &'a mut foo_type`
-    /// would have an index set of {'h0, 'a}
-    pub lifetime: IndexSet<OriginArg<'tcx>>,
     /// The lifetime parameters of a field, e.g. if a struct
     /// `foo<'a, 'b>` is a field of `bar<'c, 'd>` as field: `foo<'c, 'd>`,
-    /// the lifetime params would be a set {'c, 'd}
+    /// the lifetime params would be a set {'c, 'd}. For a reference such
+    /// as &'r foo<'c, 'd>, the lifetime params in the label would be
+    /// {'r}, and {'c, 'd} would be the label of the sole argument
+    /// of the labeled reference type
     pub lifetime_params: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
-    /// The type of the field when fully dereferenced, e.g.
-    /// `&mut &mut foo_type` would have a type of `foo_type`
-    pub fully_derefed_ty: Option<Ty<'tcx>>,
 }
 
 /// Metadata describing the lifetime parameters and fields
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -45,21 +45,12 @@ pub struct FieldMetadata<'tcx> {
 }
 
 /// Metadata describing the lifetime parameters and fields of a `struct`.
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq, Default)]
 pub struct AdtMetadata<'tcx> {
     /// The lifetime parameters of a structure, including
     /// hypothetical lifetimes derived from pointer fields.
     pub lifetime_params: IndexSet<OriginParam>,
     pub field_info: IndexMap<DefId, FieldMetadata<'tcx>>,
-}
-
-impl Default for AdtMetadata<'_> {
-    fn default() -> Self {
-        Self {
-            lifetime_params: IndexSet::new(),
-            field_info: IndexMap::new(),
-        }
-    }
 }
 
 /// An origin parameter of a type to resolve in a MIR body
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -32,7 +32,8 @@ pub struct Label<'tcx> {
 pub type LTy<'tcx> = LabeledTy<'tcx, Label<'tcx>>;
 pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label<'tcx>>;
 
-/// Metadata describing lifetimes and lifetime parameters of a `struct` field.
+/// Metadata describing lifetimes and lifetime parameters of a
+/// [TyKind::Adt](`rustc_type_ir::TyKind::Adt`) field.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FieldMetadata<'tcx> {
     /// The lifetime parameters of a field, e.g. if a struct
@@ -44,7 +45,8 @@ pub struct FieldMetadata<'tcx> {
     pub lifetime_params: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
 }
 
-/// Metadata describing the lifetime parameters and fields of a `struct`.
+/// Metadata describing the lifetime parameters and fields of a
+/// [TyKind::Adt](`rustc_type_ir::TyKind::Adt`) field.
 #[derive(Clone, PartialEq, Eq, Default)]
 pub struct AdtMetadata<'tcx> {
     /// The lifetime parameters of a structure, including
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -42,20 +42,10 @@ pub struct FieldMetadata<'tcx> {
     /// The lifetime parameters of a field, e.g. if a struct
     /// `foo<'a, 'b>` is a field of `bar<'c, 'd>` as field: `foo<'c, 'd>`,
     /// the lifetime params would be a set {'c, 'd}
-    pub lifetime_params: IndexSet<OriginArg<'tcx>>,
+    pub lifetime_params: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
     /// The type of the field when fully dereferenced, e.g.
     /// `&mut &mut foo_type` would have a type of `foo_type`
     pub fully_derefed_ty: Option<Ty<'tcx>>,
-}
-
-impl Default for FieldMetadata<'_> {
-    fn default() -> Self {
-        Self {
-            lifetime: IndexSet::new(),
-            lifetime_params: IndexSet::new(),
-            fully_derefed_ty: None,
-        }
-    }
 }
 
 /// Metadata describing the lifetime parameters and fields
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -76,7 +76,7 @@ pub enum OriginParam {
     Hypothetical(i64),
 }
 
-/// An origin parameter of a field type resolve in a MIR body
+/// An origin arg of a field type resolve in a MIR body
 /// that will get mapped to a concrete Origin to
 /// provide to polonius.
 #[derive(Hash, PartialEq, Eq, Clone, Copy)]
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -8,7 +8,8 @@ use crate::AdtMetadataTable;
 use indexmap::{IndexMap, IndexSet};
 use rustc_hir::def_id::DefId;
 use rustc_middle::mir::{Body, BorrowKind, Local, LocalKind, Place, StatementKind, START_BLOCK};
-use rustc_middle::ty::{List, Region, Ty, TyKind};
+use rustc_middle::ty::{EarlyBoundRegion, List, Region, Ty, TyKind};
+use rustc_type_ir::RegionKind::ReEarlyBound;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::hash::Hash;
@@ -24,7 +25,7 @@ pub struct Label<'tcx> {
     pub origin: Option<Origin>,
     /// The [Origins](`Origin`) associated with each lifetime
     /// parameter of this type, if applicable
-    pub origin_params: Option<&'tcx [(OriginKind<'tcx>, Origin)]>,
+    pub origin_params: Option<&'tcx [(OriginParam, Origin)]>,
     pub perm: PermissionSet,
 }
 
@@ -37,11 +38,11 @@ pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label<'tcx>>;
 pub struct FieldMetadata<'tcx> {
     /// The lifetime of the field, e.g. `*mut &'a mut foo_type`
     /// would have an index set of {'h0, 'a}
-    pub lifetime: IndexSet<OriginKind<'tcx>>,
+    pub lifetime: IndexSet<OriginArg<'tcx>>,
     /// The lifetime parameters of a field, e.g. if a struct
     /// `foo<'a, 'b>` is a field of `bar<'c, 'd>` as field: `foo<'c, 'd>`,
     /// the lifetime params would be a set {'c, 'd}
-    pub lifetime_params: IndexSet<OriginKind<'tcx>>,
+    pub lifetime_params: IndexSet<OriginArg<'tcx>>,
     /// The type of the field when fully dereferenced, e.g.
     /// `&mut &mut foo_type` would have a type of `foo_type`
     pub fully_derefed_ty: Option<Ty<'tcx>>,
@@ -63,7 +64,7 @@ impl Default for FieldMetadata<'_> {
 pub struct AdtMetadata<'tcx> {
     /// The lifetime parameters of a structure, including
     /// hypothetical lifetimes derived from pointer fields.
-    pub lifetime_params: IndexSet<OriginKind<'tcx>>,
+    pub lifetime_params: IndexSet<OriginParam>,
     pub field_info: IndexMap<DefId, FieldMetadata<'tcx>>,
 }
 
@@ -76,23 +77,61 @@ impl Default for AdtMetadata<'_> {
     }
 }
 
-/// An origin parameter to resolve in a MIR body
+/// An origin parameter of a type to resolve in a MIR body
 /// that will get mapped to a concrete Origin to
 /// provide to polonius.
 #[derive(Hash, PartialEq, Eq, Clone, Copy)]
-pub enum OriginKind<'tcx> {
-    /// An existing region, i.e. `'a` in `&'a foo`
+pub enum OriginParam {
+    /// An existing [`EarlyBoundRegion`], i.e. `'a` in `struct A<'a>`
+    Actual(EarlyBoundRegion),
+    /// A hypothesized region derived from a pointer type
+    /// e.g. `'h0` derived from the pointer in `*mut foo`
+    Hypothetical(i64),
+}
+
+/// An origin parameter of a field type resolve in a MIR body
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum OriginArg<'tcx> {
+    /// An existing [`Region`], i.e. `'a` in `&'a foo`.
+    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind`) or [RegionKind::ReStatic](`rustc_type_ir::RegionKind`)
     Actual(Region<'tcx>),
     /// A hypothesized region derived from a pointer type
     /// e.g. `'h0` derived from the pointer in `*mut foo`
     Hypothetical(i64),
 }
 
-impl<'tcx> std::fmt::Debug for OriginKind<'tcx> {
+impl<'tcx> std::fmt::Debug for OriginArg<'tcx> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
-            OriginKind::Actual(r) => write!(f, "{:}", r),
-            OriginKind::Hypothetical(h) => write!(f, "'h{h:?}"),
+            OriginArg::Actual(r) => write!(f, "{:}", r),
+            OriginArg::Hypothetical(h) => write!(f, "'h{h:?}"),
+        }
+    }
+}
+
+impl TryFrom<&OriginArg<'_>> for OriginParam {
+    type Error = ();
+    fn try_from(value: &OriginArg<'_>) -> Result<Self, Self::Error> {
+        Ok(match value {
+            OriginArg::Hypothetical(h) => OriginParam::Hypothetical(*h),
+            OriginArg::Actual(r) => {
+                if let ReEarlyBound(eb) = r.kind() {
+                    OriginParam::Actual(eb)
+                } else {
+                    return Err(());
+                }
+            }
+        })
+    }
+}
+
+impl std::fmt::Debug for OriginParam {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self {
+            OriginParam::Actual(r) => write!(f, "{:}", r.name),
+            OriginParam::Hypothetical(h) => write!(f, "'h{h:?}"),
         }
     }
 }
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -65,8 +65,7 @@ impl Default for AdtMetadata<'_> {
 }
 
 /// An origin parameter of a type to resolve in a MIR body
-/// that will get mapped to a concrete Origin to
-/// provide to polonius.
+/// that will get mapped to a concrete [`Origin`] to provide to polonius.
 #[derive(Hash, PartialEq, Eq, Clone, Copy)]
 pub enum OriginParam {
     /// An existing [`EarlyBoundRegion`], i.e. `'a` in `struct A<'a>`
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -87,7 +87,7 @@ pub enum OriginArg<'tcx> {
 }
 
 impl<'tcx> Debug for OriginArg<'tcx> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match &self {
             OriginArg::Actual(r) => write!(f, "{:}", r),
             OriginArg::Hypothetical(h) => write!(f, "'h{h:?}"),
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -1,4 +1,4 @@
-use self::atoms::{AllFacts, AtomMaps, Loan, Origin, Output, Path, SubPoint};
+use self::atoms::{AllFacts, AtomMaps, Origin, Output, SubPoint};
 use crate::context::{AnalysisCtxt, PermissionSet};
 use crate::dataflow::DataflowConstraints;
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
@@ -7,11 +7,11 @@ use crate::util::{describe_rvalue, RvalueDesc};
 use crate::AdtMetadataTable;
 use indexmap::{IndexMap, IndexSet};
 use rustc_hir::def_id::DefId;
-use rustc_middle::mir::{Body, BorrowKind, Local, LocalKind, Place, StatementKind, START_BLOCK};
+use rustc_middle::mir::{Body, LocalKind, Place, StatementKind, START_BLOCK};
 use rustc_middle::ty::{EarlyBoundRegion, List, Region, Ty, TyKind};
 use rustc_type_ir::RegionKind::ReEarlyBound;
 use std::collections::HashMap;
-use std::fmt::Debug;
+use std::fmt::{Debug, Formatter};
 use std::hash::Hash;
 
 mod atoms;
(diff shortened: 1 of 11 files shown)


c2rust-analyze/tests/filecheck/fields.rs
```diff
@@ -0,0 +1,45 @@
...
+// CHECK-DAG: ([[#@LINE+1]]: ppd): &mut &mut Data
+unsafe fn _field_access<'d>(ra: &'d mut A<'d>, ppd: *mut *mut Data<'d>) {
+    // CHECK-DAG: ([[#@LINE+2]]: rd): addr_of = UNIQUE, type = READ | UNIQUE
+    // CHECK-DAG: ([[#@LINE+1]]: rd): &Data
+    let rd = (*(**ppd).a.pra).rd;
+
+    // CHECK-DAG: ([[#@LINE+2]]: pi): addr_of = UNIQUE, type = READ | WRITE | UNIQUE
+    // CHECK-DAG: ([[#@LINE+1]]: pi): &mut i32
+    let pi = rd.pi;
+    *pi = 3;
```
aneksteind, line 40: @spernsteiner it appears that this would not be labeled as `&mut` unless it's explicitly written to, even though `Data::pi` is a `*mut` field. Is that expected? Is it because `c2rust` wouldn't translate something to `*mut` unless it was definitely written to?

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -229,12 +322,68 @@ fn assign_origins<'tcx>(
         } else {
             hypothesis[lty.label]
         };
+
+        let construct_adt_origins = |ty: &Ty, amaps: &mut AtomMaps| -> Option<&_> {
+            let mut fully_derefed_ty = ty;
+            loop {
+                match fully_derefed_ty.kind() {
+                    TyKind::RawPtr(ty) => {
```
spernsteiner, line 330: I don't think this traversal of `ty` is necessary.  `ltcx.relabel` already recurses through all the nodes in the type - for example, if the input is `*mut MyStruct`, the callback will run once for `MyStruct` and again for `*mut MyStruct`, attaching the two labels returned by the callback to their respective nodes.  With the traversal logic here, if `MyStruct` has origin parameters, then I think it will generate two sets of concrete origins, one for the `MyStruct` type itself and another copy attached to the outer `RawPtr`, something like `*mut<'h2, 'h3> MyStruct<'h0, 'h1>`.  But it doesn't make sense to have extra origins at the `*mut` level like this.  I think `assign_origins` should have basically the same structure as on master, but with an additional `TyKind::Adt` case that assigns extra origins only on the struct type itself.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -33,10 +37,92 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
 
     pub fn visit_place(&mut self, pl: Place<'tcx>) -> LTy<'tcx> {
-        let mut lty = self.local_ltys[pl.local.index()];
+        let mut lty: LTy = self.local_ltys[pl.local.index()];
+        let map: &mut HashMap<OriginKind, Origin> = &mut lty
+            .label
+            .origin_params
+            .map(|params| HashMap::from_iter(params.to_vec()))
+            .unwrap_or_default();
+        let mut adt_func = |base_adt_def: AdtDef, field: Field, field_ty: Ty<'tcx>| {
```
spernsteiner, line 47: This strategy of keeping a persistent map that gets updated over time seems potentially fragile, particularly since we eventually need to support recursive or mutually-recursive types.  But I think the actual operation being performed here is essentially just a standard type substitution, e.g. replace `'b` with `'0` in the declared field type `Foo<'b>`, where the mapping `'b -> '0` is obtained by matching up the formal parameters of `struct Bar<'b>` with the actual arguments of the base type `Bar<'0>`.  I think if you modify `util::lty_project` to pass through the base `LTy` into `adt_func` (since the base `LTy` is the one that contains the concrete origin `'0`), then you can do the whole substitution inside `adt_func` without carrying any mutable state.  (You already have `self.adt_metadata`, from which you can obtain the formal parameters of the ADT and the declared field type.)  The one quirk here is that the "declared field type" will have to be labeled with `OriginKind`s rather than `Label`s/`Origin`s, since it needs to refer to the origin parameters rather than concrete origins.

c2rust-analyze/src/context.rs
```diff
@@ -164,14 +166,46 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
 
-    pub fn label_field(&mut self, field: &FieldDef) {
-        let lty = self.assign_pointer_ids(self.tcx.type_of(field.did));
+    pub fn assn_ptr_to_field(&mut self, field: &FieldDef) {
+        let mut lty = self
+            .lcx
+            .label(self.tcx.type_of(field.did), &mut |_| PointerId::NONE);
+
+        // If this is an Adt, erase any lifetime bounds because they will
+        // be erased in the MIR body when visiting.
```
spernsteiner, line 175: Why is this necessary?  I think we mostly ignore regions appearing in the MIR or elsewhere.

Also, `TyCtxt` has an `erase_regions` method, which you might be able to use instead of the `label` + `rewrite_unlabeled` logic here.

c2rust-analyze/tests/filecheck/fields.rs
spernsteiner: That's the expected behavior.  It only uses `&mut` for pointers with the `WRITE` permission, and pointers only get `WRITE` permission if the pointer or something derived from it is written to.  Nothing in the analysis pays attention to whether the original C pointer was `const` or not (a lot of C code is not const-correct - it's common to just use `foo*` everywhere, even when `const foo*` would make more sense).

c2rust-analyze/src/context.rs
aneksteind: the region types aren't ignored when checking the types of e.g. an assignment lhs and rhs. i'll look into the `erase_regions` method

spernsteiner: >the region types aren't ignored when checking the types of e.g. an assignment lhs and rhs in the assertions made.

I think those are just debug assertions.  We could presumably change the assertions to ignore regions when checking for equality, though erasing regions up front is probably easier (I couldn't find an existing "compare ignoring regions" in the rustc docs).

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +19,98 @@ mod dump;
...
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum OriginKind<'tcx> {
+    /// An existing region, i.e. `'a` in `&'a foo`
+    Actual(Region<'tcx>),
+    /// A hypothesized region derived from a pointer type
+    /// e.g. `'h0` derived from the pointer in `*mut foo`
+    Hypothetical(i64),
+}
```
spernsteiner, line 89: I think this type is used in two different ways (let me know if I have this right):

1. In `AdtMetadata` for listing all the (real and hypothetical) origin parameters of a type.  In that case, `Actual` is only used with `ReEarlyBound`.
2. In `FieldMetadata` for listing the origin parameters of the field type.  There, `Actual` can either be `ReEarlyBound` or `ReStatic`, and both the `ReEarlyBound` and `Hypothetical` cases can only refer to the parameters listed in the parent `AdtMetadata`.

I would be in favor of splitting this into two separate types, maybe named `OriginParam` for use 1 and `OriginArg` for use 2.  I think this let us be clearer in docs and type signatures about whether we're talking about a declaration or a use of an origin parameter and which `RegionKind` variants could reasonably appear there (in fact, `OriginParam` could probably use `EarlyBoundRegion` directly, if `ReEarlyBound` is really the only legal variant to use with `Actual`).

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: addressed in edff552b

c2rust-analyze/src/borrowck/mod.rs
aneksteind: addressed in edff552b

c2rust-analyze/src/context.rs
aneksteind: addressed in 04656e0d

c2rust-analyze/tests/filecheck/fields.rs
```diff
@@ -0,0 +1,45 @@
+pub struct Data<'d> {
+    // final lifetime parameters: Data<'d, 'h0, 'h1, 'h2>
+
+    // 1: hypothetical pointer field lifetime 'h0 -> Data<'d, 'h0>
+    pub pi: *mut i32, // &'h0 i32
+
+    // 2: hypothetical pointer field lifetime 'h1 -> Data<'d, 'h0, 'h1>
+    // 3: A has no new parameters to bubble up
+    // 8: A has new lifetime param 'h2 -> Data<'d, 'h0, 'h1, 'h2>
+    pub pa: *mut A<'d>, // &'h1 A<'a, 'h0, 'h1, 'h2>
```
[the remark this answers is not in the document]
aneksteind: added in d04dbede

c2rust-analyze/src/borrowck/mod.rs
aneksteind: Let me know if 8f5fec6c accomplishes this

c2rust-analyze/src/context.rs
spernsteiner: Looks good.  I wasn't sure about the performance implications of calling `tcx.erase_regions` in so many places, but apparently the results of that operation are cached.

c2rust-analyze/src/borrowck/mod.rs
spernsteiner: That looks about right.  I have a few more comments on parts of it, which I'll add separately.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -33,10 +37,114 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                    .lifetime_params
+                    .iter()
+                    .zip(field_adt_metadata.lifetime_params.iter())
+                {
+                    let field_lifetime_param =
+                        if let Ok(param) = OriginParam::try_from(field_lifetime_arg) {
+                            param
+                        } else {
+                            continue;
+                        };
```
spernsteiner, line 95: I think what we're doing in this loop is essentially evaluating `field_lifetime_arg`, which is a sort of "origin expression", down to a concrete `Origin` that we can store in `field_origin_param_map` under the key `field_struct_lifetime_param`.  The evaluator has to handle all three kinds of "expressions": `ReEarlyBound` (resolved by looking in `base_origin_param_map`), `Hypothetical` (also resolved by looking in `base_origin_param_map`), and `ReStatic` (currently ignored - the `continue` here).  I'm not sure how `'static` is represented in Polonius, but we should either do it properly or complain that it's unsupported.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,136 @@ mod dump;
...
+/// of a struct field.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FieldMetadata<'tcx> {
+    /// The lifetime of the field, e.g. `*mut &'a mut foo_type`
+    /// would have an index set of {'h0, 'a}
+    pub lifetime: IndexSet<OriginArg<'tcx>>,
+    /// The lifetime parameters of a field, e.g. if a struct
+    /// `foo<'a, 'b>` is a field of `bar<'c, 'd>` as field: `foo<'c, 'd>`,
+    /// the lifetime params would be a set {'c, 'd}
+    pub lifetime_params: IndexSet<OriginArg<'tcx>>,
```
spernsteiner, line 45: If I'm understanding this correctly, then a  type like `*mut &'a Foo<'b>` (where `Foo` contains a pointer and therefore has a hypothetical lifetime parameter) would have metadata like this:
```
lifetime = ['h0, 'a]
lifetime_params = ['b, 'h1]
```
But this representation only works when the field type has a specific shape (some number of pointers/refs, with the final pointee type possibly being an ADT).  Would it be reasonable to instead use something like `LabeledTy<&[OriginArg]>`, where each type is labeled with a list of `OriginArg`s, similar to the `origin_params` field of `Label`?  (For ref/ptr, you could use a list of length one.)  With this representation, we could handle (to some extent) arbitrary types appearing in partially-refactored code, like `field1: (*mut i32, *mut i32)`  or `field2: Vec<MyAdt>`.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: It's represented as a universal region. I've put in panics  for now in 6465c22c, and would like to add support for this in a follow-on PR.

spernsteiner: Sounds good.

c2rust-analyze/src/main.rs
```diff
@@ -99,6 +104,250 @@ impl<T> DerefMut for MaybeUnset<T> {
...
+                                        .entry(field.did)
+                                        .or_default()
+                                        .lifetime
+                                        .insert(origin_arg);
+                                });
+                            fully_derefed_field_ty = ty.ty;
+                        }
+                        TyKind::Ref(reg, ty, _mutability) => {
+                            eprintln!("\t\tfound reference field lifetime: {reg:}");
+                            let origin_arg = OriginArg::Actual(*reg);
```
spernsteiner, line 213: Can you add assertions here and in the ADT case to make sure that `ReEarlyBound` and `ReStatic` are the only things that can be placed in the `OriginArg::Actual` variant?  I'm pretty sure those are the only two `RegionKind`s that can appear inside an `AdtDef`, but it would be good to have a check in case I'm wrong or in case rustc's behavior ever changes.

c2rust-analyze/src/borrowck/mod.rs
aneksteind: I think that's reasonable and a good idea. I've put some changes in 1e9a0fb6. Would you mind taking a look?

c2rust-analyze/src/main.rs
aneksteind: added in d8e813cb, let me know if I've missed any cases to check

c2rust-analyze/src/borrowck/mod.rs
spernsteiner: It sounds like this handles `*mut (ADT, ADT)` but not `(*mut ADT, *mut ADT)` - is that correct?

In general, I think we should avoid making unnecessary distinctions between `Ref`/`RawPtr` and `Adt` types.  Given `struct MyRef<'a>(&'a i32)`, fields `x: &'b i32` and `y: MyRef<'b>` should be treated almost identically in most of the field handling code.  In particular, I don't think it's necessary to have this split between `lifetime` (`OriginArg`s for outer `Ref`/`RawPtr` types) and `lifetime_params` (`OriginArg`s for everything else).  `&'b i32` and `MyRef<'b>` each have a single `OriginArg::Actual`, which gets substituted the same way when the field is accessed - why does one have its `OriginArg` stored in `lifetime` while the other's goes in `lifetime_params`?

Specifically, I think both `lifetime` and `lifetime_params` can be combined into a single `LabeledTy` field like the one you added in that commit.  This should allow ref/ptr and ADT types to be handled more uniformly.

c2rust-analyze/src/main.rs
spernsteiner: Looks good - those seem to be the only places where `OriginArg::Actual` is constructed.

c2rust-analyze/src/borrowck/mod.rs
aneksteind: Yes, you're right. I've made the changes in 9f5ad68b

aneksteind: also added test case in fae34e90

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+    /// as &'r foo<'c, 'd>, the lifetime params in the label would be
+    /// {'r}, and {'c, 'd} would be the label of the sole argument
+    /// of the labeled reference type
+    pub lifetime_params: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
+}
+
+/// Metadata describing the lifetime parameters and fields
+/// of a struct.
+#[derive(Clone, PartialEq, Eq)]
+pub struct AdtMetadata<'tcx> {
```
kkysen, line 51: Is this for `struct`s only?  Why call it `AdtMetadata` then?  Also, in general, how does this PR interact with non-`struct` ADTs (`enum`s)?

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum OriginParam {
+    /// An existing [`EarlyBoundRegion`], i.e. `'a` in `struct A<'a>`
+    Actual(EarlyBoundRegion),
+    /// A hypothesized region derived from a pointer type
+    /// e.g. `'h0` derived from the pointer in `*mut foo`
+    Hypothetical(i64),
+}
+
+/// An origin parameter of a field type resolve in a MIR body
```
kkysen, line 79:
```suggestion
/// An origin arg of a field type resolve in a MIR body
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum OriginArg<'tcx> {
+    /// An existing [`Region`], i.e. `'a` in `&'a foo`.
+    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind`) or [RegionKind::ReStatic](`rustc_type_ir::RegionKind`)
```
kkysen, line 85:
```suggestion
    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind::ReEarlyBound`) 
    /// or [RegionKind::ReStatic](`rustc_type_ir::RegionKind::ReStatic`)
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+}
+
+/// An origin parameter of a type to resolve in a MIR body
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum OriginParam {
+    /// An existing [`EarlyBoundRegion`], i.e. `'a` in `struct A<'a>`
+    Actual(EarlyBoundRegion),
+    /// A hypothesized region derived from a pointer type
```
kkysen, line 74:
```suggestion
    /// A hypothesized region derived from a pointer type,
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+    /// e.g. `'h0` derived from the pointer in `*mut foo`
+    Hypothetical(i64),
+}
+
+/// An origin parameter of a field type resolve in a MIR body
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
```
kkysen, line 81:
```suggestion
/// that will get mapped to a concrete [`Origin`] to provide to polonius.
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+        Self {
+            lifetime_params: IndexSet::new(),
+            field_info: IndexMap::new(),
+        }
+    }
+}
+
+/// An origin parameter of a type to resolve in a MIR body
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
```
kkysen, line 69:
```suggestion
/// that will get mapped to a concrete [`Origin`] to provide to polonius.
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
     pub perm: PermissionSet,
 }
 
-pub type LTy<'tcx> = LabeledTy<'tcx, Label>;
-pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label>;
+pub type LTy<'tcx> = LabeledTy<'tcx, Label<'tcx>>;
+pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label<'tcx>>;
+
+/// Metadata describing lifetimes and lifetime parameters
+/// of a struct field.
```
kkysen, line 36:
```suggestion
/// Metadata describing lifetimes and lifetime parameters of a `struct` field.
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+    /// `foo<'a, 'b>` is a field of `bar<'c, 'd>` as field: `foo<'c, 'd>`,
+    /// the lifetime params would be a set {'c, 'd}. For a reference such
+    /// as &'r foo<'c, 'd>, the lifetime params in the label would be
+    /// {'r}, and {'c, 'd} would be the label of the sole argument
+    /// of the labeled reference type
+    pub lifetime_params: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
+}
+
+/// Metadata describing the lifetime parameters and fields
+/// of a struct.
```
kkysen, line 49:
```suggestion
/// Metadata describing the lifetime parameters and fields of a `struct`.
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+#[derive(Clone, PartialEq, Eq)]
+pub struct AdtMetadata<'tcx> {
+    /// The lifetime parameters of a structure, including
+    /// hypothetical lifetimes derived from pointer fields.
```
kkysen, line 53:
```suggestion
    /// The lifetime parameters of a `struct`,
    /// including hypothetical lifetimes derived from pointer fields.
```

I think it's helpful to be deliberate about the difference from a Rust `struct` or the more generic word "structure".  I assume here it's referring to a Rust `struct`.

c2rust-analyze/src/context.rs
```diff
@@ -164,14 +164,14 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         *next_ptr_id = counter;
     }
 
-    pub fn label_field(&mut self, field: &FieldDef) {
+    pub fn assn_ptr_to_field(&mut self, field: &FieldDef) {
```
kkysen, line 167: What does `assn` mean?  `assign`?  Is that already a standard naming convention?  `assn` is only 2 characters shorter than `assign`.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+pub enum OriginArg<'tcx> {
+    /// An existing [`Region`], i.e. `'a` in `&'a foo`.
+    /// Can be [RegionKind::ReEarlyBound](`rustc_type_ir::RegionKind`) or [RegionKind::ReStatic](`rustc_type_ir::RegionKind`)
+    Actual(Region<'tcx>),
+    /// A hypothesized region derived from a pointer type
+    /// e.g. `'h0` derived from the pointer in `*mut foo`
+    Hypothetical(i64),
+}
+
+impl<'tcx> std::fmt::Debug for OriginArg<'tcx> {
```
kkysen, line 92:
```suggestion
impl<'tcx> Debug for OriginArg<'tcx> {
```

Can you `use std::fmt::Debug`?

kkysen: Elsewhere, too.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
kkysen, line 93:
```suggestion
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
```

Can you `use std::fmt::Formatter` and `use std::fmt`?

Elsewhere, too.  Note that `<'_>` isn't needed but is often auto-filled by rust-analyzer.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+}
 
 pub fn borrowck_mir<'tcx>(
     acx: &AnalysisCtxt<'_, 'tcx>,
     dataflow: &DataflowConstraints,
     hypothesis: &mut PointerTableMut<PermissionSet>,
     name: &str,
     mir: &Body<'tcx>,
+    adt_metadata: &AdtMetadataTable<'tcx>,
+    field_tys: HashMap<DefId, crate::LTy<'tcx>>,
```
kkysen, line 133: Is it useful enough to name both `LabeledTy<'tcx, PointerId>` and `LabeledTy<'tcx, Label<'tcx>>` as `LTy<'tcx>`?

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -190,6 +296,17 @@ fn run_polonius<'tcx>(
...
+    // Gather field permissions
+    let mut field_permissions: HashMap<DefId, PermissionSet> = HashMap::new();
+    for (did, lty) in field_tys {
+        let perm = if lty.label.is_none() {
+            PermissionSet::empty()
+        } else {
+            hypothesis[lty.label]
+        };
+        field_permissions.insert(*did, perm);
+    }
```
kkysen, line 308:
```suggestion
    let field_permissions = field_tys.iter().map(|(did, lty)| {
        let perm = if lty.label.is_none() {
            PermissionSet::empty()
        } else {
            hypothesis[lty.label]
        };
        (*did, perm)
    }).collect::<HashMap<_, _>>();
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -190,6 +296,17 @@ fn run_polonius<'tcx>(
...
+
     let mut loans = HashMap::<Local, Vec<(Path, Loan, BorrowKind)>>::new();
```
kkysen, line 310:
```suggestion
    let mut loans = HashMap::new();
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -229,12 +349,55 @@ fn assign_origins<'tcx>(
...
+                    eprintln!("pairing lifetime parameter with origin: {pairing:?}");
+                    pairing
+                })
+                .collect();
+
+            if origins.is_empty() {
+                return None;
+            }
+
+            Some(ltcx.arena().alloc_slice(&origins[..]))
```
kkysen, line 374:
```suggestion
            let origin_iter = adt_metadata
                .table
                .get(&adt_def.did())?
                .lifetime_params
                .iter()
                .map(|o| {
                    let pairing = (*o, amaps.origin());
                    eprintln!("pairing lifetime parameter with origin: {pairing:?}");
                    pairing
                });
            let origins: &[_] = ltcx.arena().alloc_from_iter(origin_iter);
            Some(origins).filter(|origins| !origins.is_empty())
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+            field_info: IndexMap::new(),
+        }
+    }
+}
+
+/// An origin parameter of a type to resolve in a MIR body
+/// that will get mapped to a concrete Origin to
+/// provide to polonius.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub enum OriginParam {
```
kkysen, line 71: What's the difference between an `OriginParam` and `OriginArg` exactly?

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
...
+}
+
+impl Default for AdtMetadata<'_> {
+    fn default() -> Self {
+        Self {
+            lifetime_params: IndexSet::new(),
+            field_info: IndexMap::new(),
+        }
+    }
+}
```
kkysen, line 65:
```suggestion
#[derive(Clone, PartialEq, Eq, Default)]
pub struct AdtMetadata<'tcx> {
    /// The lifetime parameters of a structure, including
    /// hypothetical lifetimes derived from pointer fields.
    pub lifetime_params: IndexSet<OriginParam>,
    pub field_info: IndexMap<DefId, FieldMetadata<'tcx>>,
}
```

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,123 @@ mod dump;
 mod type_check;
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, Default)]
-pub struct Label {
+pub struct Label<'tcx> {
+    /// The [`Origin`] of this type
     pub origin: Option<Origin>,
+    /// The [Origins](`Origin`) associated with each lifetime
```
kkysen, line 26:
```suggestion
    /// The [`Origin`]s associated with each lifetime
```

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -191,6 +310,24 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                for (pl_origin_param, rv_origin_param) in
+                    pl_origin_params.iter().zip(rv_origin_params.iter())
+                {
+                    let point = self.current_point(SubPoint::Mid);
+                    self.facts
+                        .subset_base
+                        .push((rv_origin_param.1, pl_origin_param.1, point));
+                }
+            }
+        }
```
kkysen, line 330:
```suggestion
        let mut add_mid_point = |pl: Origin, rv: Origin| {
            let point = self.current_point(SubPoint::Mid);
            self.facts.subset_base.push((rv, pl, point));
        };

        if let (Some(pl), Some(rv)) = (pl_lty.label.origin, rv_lty.label.origin) {
            add_mid_point(pl, rv);
        }

        for (pl_lty, rv_lty) in zip(pl_lty.iter(), rv_lty.iter()) {
            if let (Some(pl), Some(rv)) = (pl_lty.label.origin_params, rv_lty.label.origin_params) {
                assert_eq!(pl.len(), rv.len());
                for (&(_, pl_origin), &(_, rv_origin)) in zip(pl, rv) {
                    add_mid_point(rv_origin, pl_origin);
                }
            }
        }
```

This should also replace the previous part of the function, too, but GitHub won't let me highlight those lines in the review.  This seems easier to read/understand to me, with
- the refactored out `add_mid_point`
- the free `zip` function rather than the method
- more concise variable names since they're repeated so often in close proximity here

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -279,7 +416,9 @@ pub fn visit<'tcx>(
     maps: &mut AtomMaps<'tcx>,
     loans: &mut HashMap<Local, Vec<(Path, Loan, BorrowKind)>>,
     local_ltys: &[LTy<'tcx>],
+    field_permissions: &HashMap<DefId, PermissionSet>,
     mir: &Body<'tcx>,
+    adt_metadata: &AdtMetadataTable<'tcx>,
```
kkysen, line 421: `clippy` warns here that there are too many arguments to the function, and there are a lot.  Could some be refactored out similar to `TypeChecker`, which contains most of them already?

Also, what does `visit` do exactly?  What does it visit?  If it visits an entire MIR body, perhaps it can be `visit_body`.

c2rust-analyze/src/borrowck/mod.rs
spernsteiner: `alloc_from_iter` I think is a good change - the iterator is an `ExactSizeIterator`, so it should be able to preallocate the right amount of space in the arena.

I'm not a fan of `Some(x).filter(...)`, but actually I wonder if the `Option` is needed here at all.  Currently the `Label::origin_params` field has two different "empty" values: `None` and `Some(&[])`.  Is it actually necessary to distinguish between those cases, or could we remove the `Option` and just have `&[]` as the only "empty" state?  (I also wonder if `origin` and `origin_params` could be folded into a single `&[(OriginParam, Origin)]` field for uniformity, though this would need a new `OriginParam` variant for the built-in origin parameter of `TyKind::Ref` and `TyKind::RawPtr`.)

c2rust-analyze/src/borrowck/mod.rs
spernsteiner: When you declare `fn f(x: i32)` and call `f(123)`, `x` is the "formal parameter" and `123` is the "actual argument".  Similarly, when you declare `struct MyStruct<'a>` and use `MyStruct<'static>`, `'a` is an `OriginParam` and `'static` is an `OriginArg`.

c2rust-analyze/src/borrowck/type_check.rs
spernsteiner: Looks good, with two nitpicks:
1. `add_mid_point` should be called something like `add_subset_base` - the fact that it uses `SubPoint::Mid` is not the most important thing about it
2. Currently, the first two variables named `pl` (and `rv`) have the same type (`Origin`), but the third has a different type (`&[(OriginParam, Origin)]`).  I think the second loop should name the three levels of variables something like `pl_lty`, `pl_origins`, `pl` so that the use of `pl` remains consistent within this function.

c2rust-analyze/src/borrowck/type_check.rs
spernsteiner: The `visit` function is the entry point and the only public definition in this module.  I don't see much difference between passing 9 arguments to a function vs building a struct with 9 fields and then calling a method on it, so I prefer the simpler approach of just calling the function.  The function is also called in only one place IIRC.

The only visitable thing in the argument list is indeed the MIR `Body`.

c2rust-analyze/src/borrowck/mod.rs
aneksteind: for now, it's tailored to `structs` and `unions`. `enums` are not yet supported in full. essentially, anything that can be coerced via `non_enum_variant` is supported and will otherwise panic

c2rust-analyze/src/borrowck/mod.rs
aneksteind: technically we're supporting more than just `struct`, at least eventually. it's meant more colloquially than specifically

c2rust-analyze/src/context.rs
aneksteind: certain MIR portions receive a pointer id that is unique to them or their usage so that dataflow and permissions properties can be tracked properly, see the `AssignPointerIds` trait

aneksteind: expanded names in c0f91f45

c2rust-analyze/src/borrowck/mod.rs
aneksteind: What do you mean by useful enough?

c2rust-analyze/src/borrowck/mod.rs
aneksteind: address in 34bc309f except for consolidating `origin` and `origin_param`; that's a larger change involving new code for `Aggregate` and `Cast` handling, and I'd like to do that in another PR.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: addressed in 356d7b18

c2rust-analyze/src/borrowck/mod.rs
aneksteind: altered in 6aa326c0

c2rust-analyze/src/borrowck/mod.rs
kkysen: Sorry, that was less clear than in my head.  I mean, naming them both `LTy` is pretty confusing, but if `LTy` is used so ubiquitously in its own module that the short name is quite useful, then it could be a better choice still.

c2rust-analyze/src/borrowck/type_check.rs
kkysen: > I don't see much difference between passing 9 arguments to a function vs building a struct with 9 fields and then calling a method on it

The difference is unless an IDE inserts the parameter names (unlike on GitHub, for example), it's much harder to read, as the fields/parameters are unnamed at the call site.  `clippy` has the warning for a reason.  That said, if you don't agree, we should at least add an `#[allow(clippy::too_many_arguments)]`.

> The only visitable thing in the argument list is indeed the MIR `Body`.

Could we rename it `visit_body` then?  That could be done in a separate PR, though, as it's not related to this one.  `visit_body` is a lot clearer to me, though.

spernsteiner: Presumably the reason this lint is in clippy and not in rustc is that it produces too many false positives - I think this is an example of one.  The function has only a [single call site](https://github.com/immunant/c2rust/blob/f3d2e044fed855a17f7a0601de279f94ae35ecae/c2rust-analyze/src/borrowck/mod.rs#L312) (and isn't likely to gain more), and at that call site, each argument expression uses a local variable with the same name as the parameter.  The struct version would initialize its fields with `tcx: tcx`, `facts: &mut facts`, `local_ltys: &local_ltys`, etc, which isn't really adding information for the reader.

>we should at least add an `#[allow(clippy::too_many_arguments)]`
>
>Could we rename it `visit_body` then?

These both sound good to me.

c2rust-analyze/src/borrowck/mod.rs
spernsteiner: The `LabeledTy<Label>` type is used pervasively in the `borrowck` module, so it should have a short name, but I think it would be a good idea to rename it (in a separate PR) to something like `BTy` ("borrowck type") so it can't be confused with the `LTy = LabeledTy<PointerId>` type.

c2rust-analyze/src/main.rs
```diff
@@ -99,6 +107,254 @@ impl<T> DerefMut for MaybeUnset<T> {
...
+        let mut next_hypo_origin_id = 0;
+
+        // for each struct, gather lifetime information (actual and hypothetical)
+        for struct_did in &adt_metadata_table.struct_dids {
+            let adt_def = tcx.adt_def(struct_did);
+            eprintln!("gathering lifetimes and lifetime parameters for {adt_def:?}");
+            for field in adt_def.all_fields() {
+                let field_ty: Ty = tcx.type_of(field.did);
+                eprintln!("\t{adt_def:?}.{:}", field.name);
+                let mut fully_derefed_field_ty = None;
```
spernsteiner, line 180: I think `fully_derefed_field_ty` is no longer used, since the changes to how field types are labeled.

spernsteiner (approved):
Looks good overall - only a few minor comments remaining.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: addressed in 2e2b2f3e

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -15,25 +20,113 @@ mod dump;
...
+    pub origin_params: &'tcx [(OriginParam, Origin)],
     pub perm: PermissionSet,
 }
 
-pub type LTy<'tcx> = LabeledTy<'tcx, Label>;
-pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label>;
+pub type LTy<'tcx> = LabeledTy<'tcx, Label<'tcx>>;
+pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, Label<'tcx>>;
+
+/// Metadata describing lifetimes and lifetime parameters of a
```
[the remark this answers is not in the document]
aneksteind: address in 20442559

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -33,10 +37,115 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                            if let Some((base_lifetime_param, og)) =
+                                base_origin_param_map.get_key_value(&field_lifetime_param)
+                            {
+                                eprintln!(
+                                        "mapping {base_adt_def:?} lifetime parameter {base_lifetime_param:?} to \
+                                        {base_adt_def:?}.{:} struct definition lifetime parameter {field_struct_lifetime_param:?}, \
+                                        corresponding to its lifetime parameter {field_lifetime_param:?} within {base_adt_def:?}",
+                                        field_def.name
+                                    );
+                                field_origin_param_map.insert(*field_struct_lifetime_param, *og);
```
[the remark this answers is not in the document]
aneksteind: addressed in 6544b0db

c2rust-analyze/src/main.rs
aneksteind: 89af0a5f

c2rust-analyze/src/main.rs
```diff
@@ -99,6 +107,254 @@ impl<T> DerefMut for MaybeUnset<T> {
...
+                                        }
+                                    });
+                                }
+                            }
+                        }
+                        _ => (),
+                    }
+
+                    if field_lifetime_params.is_empty() {
+                        return &mut [];
```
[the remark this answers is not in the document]
aneksteind: 87795b6a

c2rust-analyze/src/main.rs
```diff
@@ -99,6 +107,254 @@ impl<T> DerefMut for MaybeUnset<T> {
...
+    let mut loop_count = 0;
+    loop {
+        /*
+            This loop iterates over all structs and gathers metadata for each.
+            If there were no recursive or mutually-recursive data structures,
+            this loop would only need one iteration to complete. To support
+            recursive and mutually-recursive structs, the loop iterates until
+            the metadata gathered for each struct reaches a fixed point.
+        */
+        loop_count += 1;
```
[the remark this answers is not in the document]
aneksteind: i'm afraid that won't scale to larger codebases, but i think a higher number would quell that a bit. so i've set it to 1000 instead 5ff01c01

c2rust-analyze/src/main.rs
```diff
@@ -99,6 +107,254 @@ impl<T> DerefMut for MaybeUnset<T> {
...
+            eprintln!("reached a fixed point in struct lifetime reconciliation\n");
+            break;
+        }
+    }
+
+    adt_metadata_table
+}
+
+pub struct AdtMetadataTable<'tcx> {
+    pub tcx: TyCtxt<'tcx>,
```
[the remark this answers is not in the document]
aneksteind: 1ebead71

PR #770, opened 2023-01-09
Install `FileCheck` for older (`apt`-based) distros

### LINKED ISSUE #477: Catch more compile-time errors in CI

### LINKED ISSUE #593: `cargo test -p c2rust-analyze` fails CI due to no `FileCheck`

kkysen:
- Fixes #593, along with the changes just merged in #767.

This installs `FileCheck` for older (`apt`-based) distros so that we can use `FileCheck` in testing without having to disable it anywhere.  Notably, we can use `FileCheck` for `c2rust-analyze` tests.

These older distros/default LLVM versions require the `llvm-${major}-tools` or `llvm-${major}.${minor}-tools` packages, while newer ones install `FileCheck` automatically, so this adds the installation of the older versions if `FileCheck` is not found in the expected location, `$(llvm-config --bindir)/FileCheck`, using the same `llvm-config` resolution used elsewhere.

This changes the docker builds, so we'll need a `./scripts/docker_build.sh push-all`.

This also adds an array of packages in the script, which makes adding to it easier, as well as allowing for inline comments (which don't work with `\` line continuation).

### DIFF
--- a/scripts/provision_deb.sh
+++ b/scripts/provision_deb.sh
@@ -56,6 +56,16 @@ packages=(
     zlib1g-dev
 )
 
+if ! [[ -x "$(llvm-config --bindir)/FileCheck" ]]; then
+	IFS="." read -r major minor patch <<< "$(llvm-config --version)"
+    if [[ ${major} -gt 6 ]]; then
+        tools="llvm-${major}-tools"
+    else
+        tools="llvm-${major}.${minor}-tools"
+    fi
+    packages+=("${tools}")
+fi
+
 apt-get install -qq "${packages[@]}"
 
 apt-get clean # clear apt-caches to reduce image size
--- a/scripts/provision_deb.sh
+++ b/scripts/provision_deb.sh
@@ -30,31 +30,33 @@ apt-get install -y --no-install-recommends \
     apt-utils \
     apt-transport-https \
     ca-certificates
-# gnupg2: required for gnupg2 key retrieval
-# llvm: required for llvm-config
-apt-get install -qq \
-    clang \
-    cmake \
-    curl \
-    dirmngr \
-    git \
-    gnupg2 \
-    gperf \
-    htop \
-    libclang-dev \
-    libssl-dev \
-    llvm \
-    ninja-build \
-    pkg-config \
-    python-dev \
-    python3-pip \
-    python3-setuptools \
-    software-properties-common \
-    strace \
-    unzip \
-    libncurses5-dev \
-    luarocks \
+
+packages=(
+    clang
+    cmake
+    curl
+    dirmngr
+    git
+    gnupg2 # required for gnupg2 key retrieval
+    gperf
+    htop
+    libclang-dev
+    libssl-dev
+    llvm # required for llvm-config
+    ninja-build
+    pkg-config
+    python-dev
+    python3-pip
+    python3-setuptools
+    software-properties-common
+    strace
+    unzip
+    libncurses5-dev
+    luarocks
     zlib1g-dev
+)
+
+apt-get install -qq "${packages[@]}"
 
 apt-get clean # clear apt-caches to reduce image size
 


aneksteind (approved):

scripts/provision_deb.sh
```diff
@@ -56,6 +56,16 @@ apt-get install -qq \
...
     zlib1g-dev
 
+if ! [[ -x "$(llvm-config --bindir)/FileCheck" ]]; then
+	IFS="." read -r major minor patch <<< "$(llvm-config --version)"
+    if [[ ${major} -gt 6 ]]; then
+        tools="llvm-${major}-tools"
+    else
+        tools="llvm-${major}.${minor}-tools"
+    fi
+    apt install "${tools}"
```
thedataking, line 66: Why does this not use -qq like the other `apt install` invocation?

This would not have been a problem if you had re-used the existing `apt install` invocation. You could define a variable `packages` and simply add to that if `FileCheck` is not installed.

Also, would dropping support for LLVM 6 (I think that's safe to do by now) allow you to just add a static package name? If so, that would be better.

Side notes: 
- Can you remove the retry loop for `apt update -qq` in a separate commit? We don't need that anymore. I think I put it in but I don't think we run this script in CI anymore so it is confusing to not just `apt update -qq`.
- Do you need to update the GitHub Actions provisioning code too?

thedataking (changes requested):
I'd like to know if dropping LLVM 6 support would simplify the provisioning change you are proposing.

scripts/provision_deb.sh
thedataking: This does not seem to work for `build_docker.sh debian:buster`. I see this on Donna:

```
The following additional packages will be installed:
  python-chardet python-pkg-resources python-pygments python-yaml
Suggested packages:
  python-setuptools python-pygments-doc ttf-bitstream-vera
The following NEW packages will be installed:
  llvm-7-tools python-chardet python-pkg-resources python-pygments python-yaml
0 upgraded, 5 newly installed, 0 to remove and 1 not upgraded.
Need to get 1225 kB of archives.
After this operation, 5753 kB of additional disk space will be used.
Do you want to continue? [Y/n] Abort.
The command '/bin/sh -c chmod +x /tmp/provision*.sh && /tmp/provision.sh' returned a non-zero code: 1
```

kkysen: > Why does this not use -qq like the other `apt install` invocation?

Oops, I missed that.  By the way, is there a reason why things like `-qq` are used?  I usually have to turn them all off when I'm debugging anything.  Also, is there a reason `apt-get install` is used instead of `apt install`?

> This would not have been a problem if you had re-used the existing `apt install` invocation. You could define a variable `packages` and simply add to that if `FileCheck` is not installed.

I could do it that way instead.  I thought this way was more minimally invasive.

> Also, would dropping support for LLVM 6 (I think that's safe to do by now) allow you to just add a static package name? If so, that would be better.

We'd need to drop Ubuntu 18 and Debian 10 support, as it's the default LLVM version on those platforms and that's where the problem arises.

> * Can you remove the retry loop for `apt update -qq` in a separate commit? We don't need that anymore. I think I put it in but I don't think we run this script in CI anymore so it is confusing to not just `apt update -qq`.

Sure.

> * Do you need to update the GitHub Actions provisioning code too?

I wasn't sure if that needed to be.  The package list there is a bit different, so I wasn't sure if I'm supposed to add to both.  I don't think the `c2rust-testsuite` action we have runs `cargo test` and thus doesn't need `FileCheck`.

kkysen: > This does not seem to work for `build_docker.sh debian:buster`. I see this on `donna`:
> 
> ```
> The following additional packages will be installed:
>   python-chardet python-pkg-resources python-pygments python-yaml
> Suggested packages:
>   python-setuptools python-pygments-doc ttf-bitstream-vera
> The following NEW packages will be installed:
>   llvm-7-tools python-chardet python-pkg-resources python-pygments python-yaml
> 0 upgraded, 5 newly installed, 0 to remove and 1 not upgraded.
> Need to get 1225 kB of archives.
> After this operation, 5753 kB of additional disk space will be used.
> Do you want to continue? [Y/n] Abort.
> The command '/bin/sh -c chmod +x /tmp/provision*.sh && /tmp/provision.sh' returned a non-zero code: 1
> ```

Oh, I thought I had tested building all of them.  Let me check again.

kkysen: > You could define a variable `packages` and simply add to that if FileCheck is not installed.

Fixed in 44b4c0eb99177601df8429a908e48b36e0f27bec and 1e80811a474eb45bfdbdfad1142edf3047f9bc9f.

> Why does this not use -qq like the other apt install invocation?

> This would not have been a problem if you had re-used the existing apt install invocation.

Fixed in 1e80811a474eb45bfdbdfad1142edf3047f9bc9f.

> Can you remove the retry loop for apt update -qq in a separate commit? We don't need that anymore. I think I put it in but I don't think we run this script in CI anymore so it is confusing to not just apt update -qq.

I'll do that separately at some point.

kkysen: @thedataking, I think it should be ready to go now.

I think doing it this way is simpler than dropping LLVM 6 and Ubuntu 18/Debian 10 support and making sure nothing breaks.  Making sure we don't break anything seems much more difficult.

> Can you remove the retry loop for `apt update -qq` in a separate commit? We don't need that anymore. I think I put it in but I don't think we run this script in CI anymore so it is confusing to not just `apt update -qq`.

This is done separately in #781.

kkysen:
@thedataking said it looks better, so I'll merge now.

scripts/provision_deb.sh
kkysen: > You could define a variable `packages` and simply add to that if `FileCheck` is not installed.

I'm going to switch back to the previous way, as this logic needs `llvm-config`, and thus needs `llvm` to be installed first.  So I need separate `apt install` commands.

PR #771, opened 2023-01-09
Re-enable `c2rust-analyze` tests in CI

### LINKED ISSUE #608: Run `cargo test -p c2rust-analyze` in CI
`cargo test --package c2rust-analyze` is currently not run in CI as it needs `FileCheck` and `FileCheck` is not available by default on Ubuntu 18 and Debian 10.

Blocked on:
* #641: I plan to move `find_llvm_config` there and the use `$(llvm-config --bindir)/FileCheck` as the `FileCheck` path.
* #593: the main issue for failing due to `FileCheck`.

kkysen:
- Fixes #608.

Depends on:
- #770
- #799

Once #770 lands, `FileCheck` will be installed everywhere, so we can finally re-enable the `c2rust-analyze` tests, which use `FileCheck`, in CI.  This is getting increasingly important as we more rapidly work on `c2rust-analyze`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -86,7 +86,7 @@ jobs:
       export CARGO_HOME=$AGENT_TEMPDIRECTORY/.cargo
       export RUSTFLAGS="-D warnings"
       export RUSTDOCFLAGS="-D warnings"
-      cargo test --release --exclude c2rust-analyze --workspace
+      cargo test --release --workspace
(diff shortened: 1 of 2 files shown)


aneksteind (approved):

kkysen:
This currently fails on Ubuntu 18 and Debian 10 (LLVM 7) due to #771, but the fix is ready in #801, so I'll merge this first and then that one.

PR #772, opened 2023-01-09
Add a `c2rust-analyze` test for `lighttpd-minimal`

kkysen:
This adds a `c2rust-analyze` test for `lighttpd-minimal` that just checks that `c2rust-analyze` runs without error on it.

### DIFF
--- a/c2rust-analyze/tests/lighttpd.rs
+++ b/c2rust-analyze/tests/lighttpd.rs
@@ -1,0 +1,23 @@
+use std::process::Command;
+
+#[test]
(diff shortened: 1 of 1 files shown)


kkysen:
Note that this still isn't being tested in CI until #771 lands, so we want to land that soon.  We can enable this in CI without #771, but it's more work than merging #770 and #771.  Just make just to test locally until that lands.  @aneksteind

aneksteind:
Understood, will try to expedite the docker image changes needed shortly after #770 lands

PR #780, opened 2023-01-13
Fix types in Python scripts

fw-immunant:
Some of these type errors are associated with real bugs, and partially this is just fighting bitrot. These scripts are supposed to be used for our release process, so cleaning them up is one step towards actually getting a new release out.

It would be nice to run mypy (`mypy --ignore-missing-imports --disallow-untyped-defs scripts/*.py`) in CI, but that requires dealing with python packaging for distros, which is a nightmare and imo not worth holding up real fixes for.

### DIFF
--- a/scripts/build_translator.py
+++ b/scripts/build_translator.py
@@ -25,12 +25,14 @@ from common import (
 )
 
 
-def download_llvm_sources():
+def download_llvm_sources() -> None:
     tar = get_cmd_or_die("tar")
 
     if not c.LLVM_SKIP_SIGNATURE_CHECKS:
         # make sure we have the gpg public key installed first
         install_sig(c.LLVM_PUBKEY)
+
+    assert c.LLVM_ARCHIVE_URLS is not None  # for mypy
 
     with pb.local.cwd(c.BUILD_DIR):
         # download archives and signatures
@@ -41,7 +43,7 @@ def download_llvm_sources():
                 c.LLVM_ARCHIVE_DIRS):
 
             if c.LLVM_SKIP_SIGNATURE_CHECKS:
-                asig = None
+                asig = None  # type: ignore
             # download archive and (by default) its signature
             download_archive(aurl, afile, asig)
 
@@ -77,7 +79,7 @@ def download_llvm_sources():
                 "--directory", cmake_modules_dir)
 
 
-def configure_and_build_llvm(args) -> None:
+def configure_and_build_llvm(args: argparse.Namespace) -> None:
     """
     run cmake as needed to generate ninja buildfiles. then run ninja.
     """
@@ -165,8 +167,8 @@ def configure_and_build_llvm(args) -> None:
             'llvm-config',
             'install-clang-headers', 'install-compiler-rt-headers',
             'FileCheck', 'count', 'not']
-        (major, _minor, _point) = c.LLVM_VER.split(".")
-        major = int(major)
+        (major_str, _minor, _point) = c.LLVM_VER.split(".")
+        major = int(major_str)
         if major >= 7 and major < 10:
             nice_args += [
                 'LLVMDebugInfoMSF',
@@ -188,7 +190,7 @@ def configure_and_build_llvm(args) -> None:
         os.makedirs(os.path.join(c.LLVM_INSTALL, 'bin'), exist_ok=True)
 
 
-def need_cargo_clean(args) -> bool:
+def need_cargo_clean(args: argparse.Namespace) -> bool:
     """
     Cargo may not pick up changes in c.BUILD_DIR that would require
     a rebuild. This function tries to detect when we need to clean.
@@ -213,7 +215,7 @@ def need_cargo_clean(args) -> bool:
     return False
 
 
-def build_transpiler(args):
+def build_transpiler(args: argparse.Namespace) -> None:
     nice = get_cmd_or_die("nice")
     cargo = get_cmd_or_die("cargo")
 
@@ -254,7 +256,7 @@ def build_transpiler(args):
             invoke(nice, *build_flags)
 
 
-def _parse_args():
+def _parse_args() -> argparse.Namespace:
     """
     define and parse command line arguments here.
     """
@@ -296,7 +298,7 @@ def _parse_args():
     return args
 
 
-def binary_in_path(binary_name) -> bool:
+def binary_in_path(binary_name: str) -> bool:
     try:
         # raises CommandNotFound exception if not available.
         _ = pb.local[binary_name]  # noqa: F841
@@ -305,7 +307,7 @@ def binary_in_path(binary_name) -> bool:
         return False
 
 
-def c2rust_bin_path(args):
+def c2rust_bin_path(args: argparse.Namespace) -> str:
     c2rust_bin_path = 'target/debug/c2rust' if args.debug \
                       else 'target/release/c2rust'
     c2rust_bin_path = os.path.join(c.ROOT_DIR, c2rust_bin_path)
@@ -314,14 +316,14 @@ def c2rust_bin_path(args):
     return os.path.relpath(c2rust_bin_path, abs_curdir)
 
 
-def print_success_msg(args):
+def print_success_msg(args: argparse.Namespace) -> None:
     """
     print a helpful message on how to run the c2rust binary.
     """
     print("success! you may now run", c2rust_bin_path(args))
 
 
-def _main():
+def _main() -> None:
     setup_logging()
     logging.debug("args: %s", " ".join(sys.argv))
 
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -198,6 +198,15 @@ class Config:
 config = Config()
 
 
+def get_host_triplet() -> str:
+    if on_linux():
+        return "x86_64-unknown-linux-gnu"
+    elif on_mac():
+        return "x86_64-apple-darwin"
+    else:
+        assert False, "not implemented"
+
+
 def update_or_init_submodule(submodule_path: str) -> None:
     git = get_cmd_or_die("git")
     invoke_quietly(git, "submodule", "update", "--init", submodule_path)
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -333,7 +333,7 @@ def ensure_dir(path: str) -> None:
         logging.debug("creating dir %s", path)
         os.makedirs(path, mode=0o744)
     if not os.path.isdir(path):
-        die("%s is not a directory", path)
+        die("%s is not a directory" % path)
 
 
 def is_elf_exe(path: str) -> bool:
@@ -556,9 +556,9 @@ def check_sig(afile: str, asigfile: str) -> None:
                 logging.warning("could not remove %s: not found.", f)
 
     if not os.path.isfile(afile):
-        die("archive file not found: %s", afile)
+        die("archive file not found: %s" % afile)
     if not os.path.isfile(asigfile):
-        die("signature file not found: %s", asigfile)
+        die("signature file not found: %s" % asigfile)
 
     # check that archive matches signature
     try:
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -12,7 +12,7 @@ import platform
 import multiprocessing
 
 from pathlib import Path
-from typing import List
+from typing import Any, Dict, List, NoReturn, Optional, Tuple, Union
 
 import plumbum as pb
 
@@ -30,10 +30,8 @@ class Colors:
 
 
 class Config:
-    BUILD_SUFFIX = ""
     # use custom build directory suffix if requested via env. variable
-    if os.getenv('C2RUST_BUILD_SUFFIX'):
-        BUILD_SUFFIX = os.getenv('C2RUST_BUILD_SUFFIX')
+    BUILD_SUFFIX = os.getenv('C2RUST_BUILD_SUFFIX') or ""
     BUILD_TYPE = "release"
 
     NCPUS = str(multiprocessing.cpu_count())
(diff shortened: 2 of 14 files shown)


fw-immunant:
Of course Ubuntu 18 has a Python with no `re.Pattern` type.

kkysen:
> It would be nice to run mypy (`mypy --ignore-missing-imports --disallow-untyped-defs scripts/*.py`) in CI, but that requires dealing with python packaging for distros, which is a nightmare and imo not worth holding up real fixes for.

Even if we can't add it to CI (Python installations really are frustrating), could we at least add that `mypy --ignore-missing-imports --disallow-untyped-defs scripts/*.py` to a script?  Perhaps `run_ci_checks.sh`, which contains the other CI checks, as this is something we'd at least like to run in CI.

scripts/common.py
```diff
@@ -12,7 +12,7 @@
 import multiprocessing
 
 from pathlib import Path
-from typing import List
+from typing import Any, Dict, List, NoReturn, Optional, Tuple, Union
```
kkysen, line 15: What Python version do we use?  Newer ones can use `|` for `Union`.

scripts/common.py
```diff
@@ -32,8 +32,9 @@ class Colors:
 class Config:
     BUILD_SUFFIX = ""
     # use custom build directory suffix if requested via env. variable
-    if os.getenv('C2RUST_BUILD_SUFFIX'):
-        BUILD_SUFFIX = os.getenv('C2RUST_BUILD_SUFFIX')
+    env_suffix = os.getenv('C2RUST_BUILD_SUFFIX')
+    if env_suffix is not None:
+        BUILD_SUFFIX = env_suffix
```
kkysen, line 37: Is it fine that this is not quite semantically equivalent?  It'll differ on `""`.

kkysen: Does doing just `if env_suffix` work?

scripts/common.py
```diff
@@ -170,6 +170,7 @@ def update_args(self, args=None):
 
         env_target_dir = os.getenv('CARGO_TARGET_DIR')
         self.TARGET_DIR = "{}/".format(build_type)
+        assert self.TARGET_DIR is not None  # for mypy
         if env_target_dir:
```
kkysen, line 174: See the above comment aboue `if env_suffix is not None`.  It seems to work this way here.

scripts/common.py
```diff
@@ -254,7 +253,7 @@ def on_linux() -> bool:
     return platform.system() == "Linux"
 
 
-def regex(raw: str):
+def regex(raw: str) -> 're.Pattern':
```
kkysen, line 265: Why does it need to be quoted?  Does that have to do with that Ubuntu 18 issue?

scripts/common.py
```diff
@@ -287,15 +286,15 @@ def est_parallel_link_jobs():
...
     return _invoke(True, cmd, *arguments)
 
 
-def invoke_quietly(cmd, *arguments):
+def invoke_quietly(cmd: Command, *arguments: Union[str, List[str]]) -> Tuple[int, str, str]:
     return _invoke(False, cmd, *arguments)
 
 
-def _invoke(console_output, cmd, *arguments):
+def _invoke(console_output: bool, cmd: Command, *arguments: Union[str, List[str]]) -> Tuple[int, str, str]:
```
kkysen, line 306: Can we extract something like this (not sure if these are the best names)?

```python
Args = Union[str, List[str]]
Invocation = Tuple[int, str, str]
```

scripts/common.py
```diff
@@ -476,6 +474,7 @@ def transpile(cc_db_path: str,
     """
     run the transpiler on all C files in a compile commands database.
     """
+    assert config.C2RUST_BIN is not None
```
kkysen, line 486: Can you add similar `# for mypy` comments where you use these assertions specifically for `mypy` like in one of the earlier ones?

scripts/convert_build_commands.py
```diff
@@ -43,6 +45,7 @@ def convert_entries(entries, out_dir=None):
             elif arg == "-o":
                 ei.output = next(arg_iter)
                 ei.new_args.append(arg)
+                assert ei.output is not None
```
kkysen, line 48: `# for mpypy` again (see above comment).

scripts/integration_test_translator.py
```diff
@@ -76,6 +76,7 @@
 
 
 def _test_minimal(code_snippet: str) -> bool:
+    assert c.TRANSPILER is not None
```
kkysen, line 79: `# for mpypy` again (see above comment).

scripts/integration_test_translator.py
```diff
@@ -215,6 +216,7 @@ def main() -> None:
     logging.debug("args: %s", " ".join(sys.argv))
 
     # check that the binaries have been built first
+    assert c.TRANSPILER is not None
```
kkysen, line 219: `# for mpypy` again (see above comment).

scripts/integration_test_translator.py
```diff
@@ -240,7 +242,8 @@ def main() -> None:
     success = True
     for t in tests:
         logging.debug("running test: %s", t.__name__)
-        success = success and t(args)
+        if not t(args):
+            success = False
```
kkysen, line 246: Why doesn't the old version type-check?

scripts/package.py
```diff
@@ -7,9 +7,13 @@
 
 import toml
 
+from typing import Any, Callable, Dict, Optional, Tuple
```
kkysen, line 10: Do we put stdlib imports (`typing`) above library ones (`toml`)?

scripts/rust_file.py
```diff
@@ -77,12 +77,15 @@ def __str__(self) -> str:
     def __hash__(self) -> int:
         return hash((self.visibility, self.name))
 
-    def __eq__(self, other: "RustMod") -> bool:
-        return self.name == other.name and self.visibility == other.visibility
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, RustMod):
+            return self.name == other.name and self.visibility == other.visibility
+        else:
+            return False
```
kkysen, line 84:
```suggestion
        return isinstance(other, RustMod) and self.name == other.name and self.visibility == other.visibility
```

Does this still type-check?  If so, I think it's preferred.

scripts/rust_file.py
```diff
@@ -92,14 +95,17 @@ def __str__(self) -> str:
     def __hash__(self) -> int:
         return hash((self.use, self.visibility))
 
-    def __eq__(self, other: "RustUse") -> bool:
-        return self.use == other.use and self.visibility == other.visibility
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, RustUse):
+            return self.use == other.use and self.visibility == other.visibility
+        else:
+            return False
```
kkysen, line 102:
```suggestion
        isinstance(other, RustUse) and return self.use == other.use and self.visibility == other.visibility
```

Same here.

scripts/test_translator.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, path: str, link_name: str,
 
 
 class CFile:
-    def __init__(self, log_level: str, path: str, flags: Set[str] = None) -> None:
+    def __init__(self, log_level: str, path: str, flags: Set[str] = set()) -> None:
```
kkysen, line 66: `set()` is mutable, so this will re-use the same set on separate `__init__` calls.  I think you have to set it to `None` and then do an

```python
if flags is not None:
    flags = set()
```

Very annoying, but oh well.

scripts/test_translator.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, path: str, link_name: str,
...
         if not flags:
             flags = set()
```
kkysen, line 68: Yeah, the initial code does it correctly except for the types.  It's annoying that `mypy` can't refine an existing variable's type, though, and argument names are public (for keyword args).

Instead, I think it's better to restrict the `flags` type to what's needed, `Container`, with its `__contains__` method and `in` operator.  Thus, we can set the default value to `()`, which is immutable.

```suggestion
    def __init__(self, log_level: str, path: str, flags: Container[str] = ()) -> None:
```

scripts/test_examples.py
```diff
@@ -273,86 +275,86 @@ def __init__(self, args):
...
+        self.build_flags: List[str] = []
         self.init_submodule()
         self.rust_src = os.path.join(self.repo_dir, 'rust')
 
-    def __del__(self):
+    def __del__(self) -> None:
         self.deinit_submodule()
 
-    def autotools(self):
+    def autotools(self, configure_args: List[str] = []) -> None:
```
kkysen, line 311: Default argument can't be mutable, like `[]`.

scripts/test_translator.py
```diff
@@ -75,10 +75,11 @@ def __init__(self, log_level: str, path: str, flags: Set[str] = None) -> None:
         self.reorganize_definitions = "reorganize_definitions" in flags
         self.emit_build_files = "emit_build_files" in flags
 
-    def translate(self, cc_db, ld_lib_path, extra_args: List[str] = []) -> RustFile:
+    def translate(self, cc_db: str, ld_lib_path: str, extra_args: List[str] = []) -> RustFile:
```
kkysen, line 78: Default argument can't be mutable, like `[]`.

scripts/test_translator.py
```diff
@@ -651,6 +654,7 @@ def main() -> None:
     logging.debug("args: %s", " ".join(sys.argv))
 
     # check that the binaries have been built first
+    assert c.TRANSPILER is not None
```
kkysen, line 657: `# for mpypy` again (see above comment).

scripts/test_translator.py
```diff
@@ -372,6 +373,8 @@ def run(self) -> List[TestOutcome]:
 
             return outcomes
 
+        assert static_library is not None
```
kkysen, line 376: `# for mypy` again (see above comment).

scripts/build_translator.py
```diff
@@ -25,13 +25,15 @@
...
 
-def download_llvm_sources():
+def download_llvm_sources() -> None:
     tar = get_cmd_or_die("tar")
 
     if not c.LLVM_SKIP_SIGNATURE_CHECKS:
         # make sure we have the gpg public key installed first
         install_sig(c.LLVM_PUBKEY)
 
+    assert c.LLVM_ARCHIVE_URLS is not None
```
kkysen, line 35: `# for mypy` again (see below comment).

scripts/common.py
```diff
@@ -202,6 +202,15 @@ def add_args(parser: argparse.ArgumentParser) -> None:
 config = Config()
 
 
+def get_host_triplet() -> str:
+    if on_linux():
+        return "x86_64-unknown-linux-gnu"
+    elif on_mac():
+        return "x86_64-apple-darwin"
+    else:
+        assert False, "not implemented"
```
kkysen, line 211:
```suggestion
        raise NotImplemented
```

thedataking:
> > It would be nice to run mypy (`mypy --ignore-missing-imports --disallow-untyped-defs scripts/*.py`) in CI, but that requires dealing with python packaging for distros, which is a nightmare and imo not worth holding up real fixes for.
> 
> Even if we can't add it to CI (Python installations really are frustrating), could we at least add that `mypy --ignore-missing-imports --disallow-untyped-defs scripts/*.py` to a script? Perhaps `run_ci_checks.sh`, which contains the other CI checks, as this is something we'd at least like to run in CI.

Having super nice Python scripts with annotations checked in CI is not a priority; the fixes in this PR are nice but should be wrapped up without creating additional work. These are approximately our priorities:

The priorities are: 
1. Improving the lifting (including testing). 
2. Fixing transpiler bugs reported by users 
3. Implementing support for C constructs that prevent translating C code bases (e.g. Atomics)
4. [Reserved]
5. CI stuff/scripting.

@kkysen please do not spend additional time on CI or scripting unless specifically asked to do so. Let's put our energy towards cutting a new release.

scripts/common.py
fw-immunant: This is just code motion; semantic changes are out of scope.

scripts/common.py
fw-immunant: Ancient on old distros in CI, so we can't rely on anything remotely new.

scripts/common.py
fw-immunant: Fixed.

scripts/common.py
fw-immunant: Obviated by simpler cleanups.

scripts/common.py
fw-immunant: Yes, Ubuntu 18's python is too old to know the `re.Pattern` type, so this is how it must be annotated.

scripts/common.py
fw-immunant: Sure, but it's not worth it for now.

scripts/convert_build_commands.py
fw-immunant: This is a correctness condition.

scripts/integration_test_translator.py
fw-immunant: Because `test_minimal` and `test_hello_world` returned None. They should really return bools propagated from their inner calls, though--I've changed them to do so, which obviates this change.

scripts/test_translator.py
fw-immunant: We don't mutate the set, so it's fine.

scripts/test_translator.py
fw-immunant: Resolved as we don't mutate this set either.

fw-immunant:
I cleaned this up and will merge so it doesn't rot if there are no critical objections today.

scripts/test_examples.py
fw-immunant: Used immutably, this is fine.

scripts/test_translator.py
fw-immunant: Again, used immutably.

PR #781, opened 2023-01-14
Remove the `apt-get` retry loop in `provision_dep.sh`

kkysen:
Remove the `apt-get` retry loop in `provision_dep.sh`.

It used to be there as the script ran in CI, but it no longer is, so we don't need to retry anymore.
See https://github.com/immunant/c2rust/pull/770#discussion_r1065225015.


thedataking (changes requested):
We still need to run `apt-get update -qq` as as the apt metadata may be missing or stale. My suggestion was that we remove the loop so we only attempt the update operation once.

kkysen:
> We still need to run `apt-get update -qq` as as the apt metadata may be missing or stale. My suggestion was that we remove the loop so we only attempt the update operation once.

Oops, I misunderstood.  It's fixed now.

thedataking:
I don't see the button to approve this PR but it looks great now, thanks! Going to merge it in.

PR #784, opened 2023-01-18
Skip `tests/simd` on non-`x86_64`

### LINKED ISSUE #783: Running `test_translator.py` on Apple M1 incorrectly runs `tests/simd/src/x86.c`
Running `test_translator.py` on Apple M1 runs the `tests/simd/src/x86.c/tests/simd/src/x86.c` test despite M1 being ARM, causing it to predictably fail immediately.

```sh
ivy:~/projects/c2rust(779_fix_test_translator_macOs$!?)via 🐍 v3.11.1 via 🦀 v1.65.0-nightly via 🐏 11GiB/16GiB | 1GiB/3GiB took 11s598ms  ./scripts/test_translator.py tests

...

 [ OK ]     test test_enums.rs - test_buffer4
 [ OK ]     test test_enums.rs - test_buffer5
simd:
 [ FAILED ] create libtest.a
In file included from /Users/ivy/projects/c2rust/tests/simd/src/x86.c:1:
/Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/xmmintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
#error "This header is only meant to be used on x86 and x64 architecture"
 ^
In file included from /Users/ivy/projects/c2rust/tests/simd/src/x86.c:1:
In file included from /Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/xmmintrin.h:17:
/Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/mmintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
#error "This header is only meant to be used on x86 and x64 architecture"
 ^
/Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/mmintrin.h:54:12: error: invalid conversion between vector type '__m64' (vector of 1 'long long' value) and integer type 'int' of different size
    return (__m64)__builtin_ia32_vec_init_v2si(__i, 0);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...

```

kkysen:
- Fixes #783.

`tests/simd`, which contains `x86.c`, is full of x86-specific SIMD code, so we need to skip it on non-x86 architectures.  Thus, this PR
- adds an `x86_64-unknown-linux-gnu` `target-tuple` file, which is for the Rust code
- renames `tests/{simd => simd.x86_64}`, which is for the initial target recognition

The non-native target skipping code is also after the C compilation, which doesn't work for skipping the platform-specific C compilation, too, so this moves it up to the beginning.


ivyraine (approved):
lgtm, the tests pass on my machine (Apple M1) In this PR.

fw-immunant (approved):
LGTM.

PR #786, opened 2023-01-18
add more lighttpd code to static analysis

aneksteind:


analysis/tests/lighttpd-minimal/src/main.rs
```diff
@@ -4,88 +4,1149 @@
...
 pub struct server {
+    pub plugin_slots: *mut libc::c_void,
+    pub config_context: *mut array,
+    pub config_captures: libc::c_int,
     pub ev: *mut fdevents,
-    pub cur_fds: libc::c_int,
+    pub network_backend_write: Option<
+        unsafe extern "C" fn(libc::c_int, *mut chunkqueue, off_t, *mut log_error_st) -> libc::c_int,
+    >,
+    pub request_env: Option<unsafe extern "C" fn(*mut request_st) -> handler_t>,
```
kkysen, line 231: Is leaving these in types fine?

c2rust-analyze/src/main.rs
```diff
@@ -742,7 +742,7 @@ fn fn_body_owners_postorder(tcx: TyCtxt) -> Vec<LocalDefId> {
 
         match tcx.def_kind(root_ldid) {
             DefKind::Fn | DefKind::AssocFn => {}
-            DefKind::AnonConst => continue,
+            DefKind::AnonConst | DefKind::Const => continue,
```
kkysen, line 745: What does this fix?

analysis/tests/lighttpd-minimal/src/main.rs
```diff
@@ -183,8 +1292,14 @@ unsafe extern "C" fn fdnode_free(fdn: *mut libc::c_void) {
 }
 
 pub unsafe extern "C" fn lighttpd_test() {
-    let fdarr = malloc(::std::mem::size_of::<*mut fdnode>() as libc::c_ulong); // TODO: handle cast as *mut *mut fdnode;
-    let fdes = malloc(::std::mem::size_of::<fdevents>() as libc::c_ulong); // TODO: handle cast as *mut fdevents;
+    let fdarr = calloc(
+        0 as libc::c_int as libc::c_ulong,
```
kkysen, line 1296: Can you just confirm these are supposed to be 0?  I remember we had a nullptr deref from something like this before.

kkysen (commented):
I think this should all be fine as long as it passes the tests.  That said, it's a bit hard to see what exactly changed here since the automatic diff isn't very clean, so it'd be helpful to separate this into a few commits (e.x. all the new function declarations and types, and then the function body changes) (I can take a look at this).

analysis/tests/lighttpd-minimal/src/main.rs
```diff
@@ -4,88 +4,1149 @@
...
         (*srv).ev,
         (*con).fd,
         // Some(
         //     connection_handle_fdevent
         //         as unsafe extern "C" fn(*mut libc::c_void, libc::c_int) -> handler_t,
         // ),
         // con as *mut libc::c_void,
-        fdn
+        // con as *mut libc::c_void,
+        fdn,
```
kkysen, line 1062: I think this duplicates the `con` arg.

kkysen:
> I think this should all be fine as long as it passes the tests. That said, it's a bit hard to see what exactly changed here since the automatic diff isn't very clean, so it'd be helpful to separate this into a few commits (e.x. all the new function declarations and types, and then the function body changes) (I can take a look at this).

I opened #792 for this for how I think it'd be better to split up the commits.

analysis/tests/lighttpd-minimal/src/main.rs
aneksteind: yes, although I'm not sure what you're concerned about

c2rust-analyze/src/main.rs
aneksteind: being able to handle things like `pub const HANDLER_GO_ON: handler_t = 0;` without associating a body with them

aneksteind:
closing in favor of #792

analysis/tests/lighttpd-minimal/src/main.rs
aneksteind: these are contrived for the time being, the actual code is more involved

kkysen: That's fine as long as we make sure to update them before they are used and not just immediately freed.

analysis/tests/lighttpd-minimal/src/main.rs
kkysen: I fixed this in #792.

analysis/tests/lighttpd-minimal/src/main.rs
kkysen: I just know we can't fully handle function pointers, so I wasn't sure if it's okay to have them in types and just not use them.

aneksteind: that's correct, only a subset of types are iterated over (ADTs) and it's their use in MIR bodies that would pose an issue

PR #788, opened 2023-01-20
Support casts to and from `*c_void` with `free`/`malloc`/`calloc`/`realloc`

aneksteind:


c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -271,17 +272,20 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-                );
-                */
-                Label::default()
-            }),
+            Rvalue::Cast(ck, _, ty) => {
+                // TODO: handle Unsize casts
 
+                // special case in casting to void*
+                if expect_ty.label.perm.contains(PermissionSet::FREE) {
+                    return expect_ty;
```
aneksteind, line 280: @spernsteiner these two `if` blocks feel incorrect, but I'm not sure what the proper label would be for types casted from `void*` to `my_struct*` that are eventually passed to `free`. There are downstream checks that the origin params are equivalent in size between LHS and RHS types, and so it seems like we either have to 
- create `origin_params` out of thin air
- copy the structure from `expect_ty` or
- avoid asserting `LTy` `origin_params` length parity when casting from `void*` to `my_struct*`.

What do you think?

spernsteiner: I think the general approach (for both borrowck and dataflow) should be to recognize this pattern:
```
_1 = malloc(...);
_2 = _1 as *mut T;
```
and instead treat it as if the MIR code read
```
_2 = malloc(...);
```
One option for implementing this would be to run a scan over the whole function up front and record some annotations when it sees the pattern: "use this `Place` instead of the real LHS" for the calls, and "ignore this statement" for the casts.  Then check for the relevant annotations in the `Call` and `Cast` cases of both borrowck and dataflow.

Another option (which might actually be simpler overall) would be to clone the entire MIR `Body` for each function and rewrite the pattern to the replacement in-place.

aneksteind: @spernsteiner I decided to go with the first option

c2rust-analyze/src/main.rs
```diff
@@ -424,7 +424,7 @@ fn run(tcx: TyCtxt) {
     for &ldid in &all_fn_ldids {
         let ldid_const = WithOptConstParam::unknown(ldid);
         let mir = tcx.mir_built(ldid_const);
-        let mir = mir.borrow();
+        let mir: std::cell::Ref<Body> = mir.borrow();
```
kkysen, line 427: What's this needed for?  Failed `rust-analyzer` intellisense?

Also,
```suggestion
        let mir: cell::Ref<Body> = mir.borrow();
```

c2rust-analyze/src/context.rs
```diff
@@ -84,11 +85,25 @@ pub struct AnalysisCtxt<'a, 'tcx> {
...
+    /// _1 = malloc(...);
+    /// _2 = _1 as *mut T;
+    /// ```
+    ///
+    /// In this case, `_1` would be mapped to `_2`, which is indicative
+    /// of the amended statement:
+    /// ```
+    /// _2 = malloc(...);
+    /// ```
+    pub special_casts: HashMap<Place<'tcx>, Place<'tcx>>,
```
kkysen, line 106: Can you newtype `SpecialCasts`?  I think a `struct SpecialCasts(HashMap<Place<'tcx>, Place<'tcx>>)` is better, as that way the intent is a bit clearer, you don't have to repeat the same docs twice, and you can add a few methods that better signify intent and reduce repetition?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -227,12 +227,18 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         match stmt.kind {
             StatementKind::Assign(ref x) => {
                 let (pl, ref rv) = **x;
+                if matches!(rv, Rvalue::Cast(_, Operand::Copy(p) | Operand::Move(p), _) if self.acx.special_casts.contains_key(&p) || self.acx.special_casts.contains_key(&pl))
```
kkysen, line 230: I think this part, specifically the `matches!(rv, Rvalue::Cast(_, Operand::Copy(p) | Operand::Move(p), _) if self.acx.special_casts.contains_key(&p)` can be extracted to a `SpecialCasts` method as its repeated above as well.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -295,18 +301,39 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        self.visit_place(*destination, Mutability::Mut);
                     }
 
                     Some(Callee::Calloc) => {
-                        self.visit_place(destination, Mutability::Mut);
+                        let destination = self
+                            .acx
+                            .special_casts
+                            .get(&destination)
+                            .unwrap_or(&destination);
```
kkysen, line 317: This logic of `special_casts.get(&x).unwrap_or(&x)` that's repeated a bunch of times could be put in `SpecialCasts`, too.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
...
+
+                    match util::ty_callee(tcx, func_ty) {
+                        Some(Malloc | Calloc) => {
+                            assert_libc_cvoid(&destination);
+                            acx.c_void_ptrs.insert(destination);
+                        }
+                        Some(Realloc) => {
+                            assert_libc_cvoid(&destination);
+                            acx.c_void_ptrs.insert(destination);
+                            acx.c_void_ptrs.insert(args[0].place().unwrap());
```
kkysen, line 491: Do you need an `assert_libc_cvoid(&args[0].place().unwrap())`, too, like for the other ones?

kkysen: Can you extract an `add_c_void_ptr` function that does both `assert_libc_cvoid` and `.insert`?

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
...
+                        let deref_ty = p
+                            .ty(acx.local_decls, acx.tcx())
+                            .ty
+                            .builtin_deref(true)
+                            .unwrap()
+                            .ty
+                            .kind();
+                        assert_matches!(deref_ty, TyKind::Adt(adt, _) => {
+                            assert_eq!(tcx.def_path(adt.did()).data[0].to_string(), "ffi");
+                            assert_eq!(tcx.item_name(adt.did()).as_str(), "c_void");
```
kkysen, line 479: What about `libc::c_void`?

Also, this works for both `std::` and `core::` `ffi::c_void`?

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
...
+                    ref args,
+                    destination,
+                    target: _,
+                    ..
+                } = term.kind
+                {
+                    let func_ty = func.ty(&mir.local_decls, tcx);
+                    use crate::util::Callee::*;
+
+                    let assert_libc_cvoid = |p: &Place<'tcx>| {
```
kkysen, line 469: Maybe just name this `assert_is_c_void` as it doesn't have to be `libc`.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
             assert_eq!(local, l);
         }
 
+        // Find all calls to `malloc`, `calloc`, `realloc`, and `free`
+        // and track their destination locals as being libc::c_void
+        // pointers to special-case downstream casts
+        for bb_data in mir.basic_blocks().iter() {
+            if let Some(term) = &bb_data.terminator {
```
kkysen, line 456:
```suggestion
        for term in mir.basic_blocks().iter().filter_map(|bb| bb.terminator) {
```

Something like this could help just to reduce the indentation level a bit.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
             assert_eq!(local, l);
         }
 
+        // Find all calls to `malloc`, `calloc`, `realloc`, and `free`
+        // and track their destination locals as being libc::c_void
```
kkysen, line 453: I think we should just call it `c_void` in general, as it could be `core::ffi::c_void` or `libc::c_void` depending on the Rust version, and `c_void` seems clear enough on its own.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
...
+                if let TerminatorKind::Call {
+                    ref func,
+                    ref args,
+                    destination,
+                    target: _,
+                    ..
+                } = term.kind
+                {
+                    let func_ty = func.ty(&mir.local_decls, tcx);
+                    use crate::util::Callee::*;
```
kkysen, line 467: Can you move this to right before use-site, right before the `match`?

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,65 @@ fn run(tcx: TyCtxt) {
...
+        // pointers to special-case downstream casts
+        for bb_data in mir.basic_blocks().iter() {
+            if let Some(term) = &bb_data.terminator {
+                let term: &Terminator = term;
+                if let TerminatorKind::Call {
+                    ref func,
+                    ref args,
+                    destination,
+                    target: _,
+                    ..
```
kkysen, line 463: Isn't the `target: _` redundant here with the `..`?

c2rust-analyze/src/context.rs
```diff
@@ -84,11 +85,25 @@ pub struct AnalysisCtxt<'a, 'tcx> {
 
     pub local_decls: &'a LocalDecls<'tcx>,
     pub local_tys: IndexVec<Local, LTy<'tcx>>,
+    pub c_void_ptrs: IndexSet<Place<'tcx>>,
```
kkysen, line 88: Can this just be a `HashSet`?  Order doesn't seem important as only `.insert` and `.contains` are used.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -25,6 +25,19 @@ struct TypeChecker<'tcx, 'a> {
...
+    /// _1 = malloc(...);
+    /// _2 = _1 as *mut T;
+    /// ```
+    ///
+    /// In this case, `_1` would be mapped to `_2`, which is indicative
+    /// of the amended statement:
+    /// ```
+    /// _2 = malloc(...);
+    /// ```
+    special_casts: &'a HashMap<Place<'tcx>, Place<'tcx>>,
```
kkysen, line 40: Is `special_casts` the best name?  What about `c_void_casts`, in line with `c_void_ptrs`?

c2rust-analyze/src/context.rs
kkysen: Why not put `c_void_ptrs` right next to `special_casts`?  There might be a reason I'm missing, but otherwise, it'd make sense to put them together and make it a bit easier to understand they're closely related.

c2rust-analyze/src/main.rs
```diff
@@ -465,7 +517,23 @@ fn run(tcx: TyCtxt) {
...
+                            // This is a special case for types being casted from *libc::c_void to a pointer
+                            // to some other type, e.g. `let foo = malloc(..) as *mut Foo;`
+                            // carry over the pointer id of *libc::c_void, but match the pointer ids
+                            // of the casted-to-type for the rest
+                            acx.special_casts.insert(p, lhs);
+                        }
+
+                        if acx.c_void_ptrs.contains(&lhs) {
+                            // This is a special case for types being casted to *libc::c_void
+                            // acx.special_casts.insert(p, lhs);
```
kkysen, line 531: What does this mean?

c2rust-analyze/src/context.rs
```diff
@@ -84,11 +85,25 @@ pub struct AnalysisCtxt<'a, 'tcx> {
...
     /// Types for certain [`Rvalue`]s.  Some `Rvalue`s introduce fresh [`PointerId`]s; to keep
     /// those `PointerId`s consistent, the `Rvalue`'s type must be stored rather than recomputed on
     /// the fly.
     pub rvalue_tys: HashMap<Location, LTy<'tcx>>,
+    /// A mapping for substituting [`Place`]s adhering to the
+    /// following pattern
+    /// ```
+    /// _1 = malloc(...);
+    /// _2 = _1 as *mut T;
+    /// ```
```
kkysen, line 99:
```suggestion
    /// ```mir
    /// _1 = malloc(...);
    /// _2 = _1 as *mut T;
    /// ```
```

Otherwise I think it may be interpreted as a Rust doctest, and also it's more clear.  This also applies to the other ` ``` `s.

c2rust-analyze/src/main.rs
aneksteind: addressed in 0fb24480

c2rust-analyze/src/context.rs
aneksteind: added in 528ec9e4

c2rust-analyze/src/main.rs
```diff
@@ -356,7 +356,7 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
     }
 }
 
-fn run(tcx: TyCtxt) {
+fn run<'tcx>(tcx: TyCtxt<'tcx>) {
```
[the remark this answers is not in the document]
aneksteind: because the closure `assert_libc_cvoid` takes in a `Place` that has a lifetime parameter that's ambiguous otherwise (the compiler assumes its data escapes the closure)

c2rust-analyze/src/main.rs
aneksteind: added in 9f72789c

c2rust-analyze/src/main.rs
aneksteind: AFAIK this works for `libc::c_void`. That's what `alloc.rs` and `lighttpd-minimal/src/main.rs` are tested with. What changes are you suggesting be made?

c2rust-analyze/src/main.rs
aneksteind: addressed in 9f72789c

c2rust-analyze/src/context.rs
aneksteind: addressed in 0d6dc10f

c2rust-analyze/src/main.rs
aneksteind: that line was left in by mistake. removed

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: addressed in e195b42c

c2rust-analyze/src/main.rs
aneksteind: will opt out of this change. feel free to suggest an entire working change if you feel strongly about it

c2rust-analyze/src/main.rs
aneksteind: addressed in cc90e3f7

c2rust-analyze/src/main.rs
aneksteind: addressed in 7a331276

c2rust-analyze/src/main.rs
aneksteind: 2e8801e7

c2rust-analyze/src/context.rs
aneksteind: 875aad88

c2rust-analyze/src/main.rs
kkysen: If it's helpful for intellisense, it's fine to keep it; just wondering.

aneksteind: I think you had it right, it was a rust-analyzer thing and I did not add it intentionally

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -25,6 +25,19 @@ struct TypeChecker<'tcx, 'a> {
...
+    /// ```mir
+    /// _1 = malloc(...);
+    /// _2 = _1 as *mut T;
+    /// ```
+    ///
+    /// In this case, `_1` would be mapped to `_2`, which is indicative
+    /// of the amended statement:
+    /// ```
+    /// _2 = malloc(...);
+    /// ```
```
kkysen, line 39: Can you move these docs to `struct CVoidCasts`?  Then it doesn't need to be duplicated.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,62 @@ fn run(tcx: TyCtxt) {
...
+                            .kind();
+                        assert_matches!(deref_ty, TyKind::Adt(adt, _) => {
+                            assert_eq!(tcx.def_path(adt.did()).data[0].to_string(), "ffi");
+                            assert_eq!(tcx.item_name(adt.did()).as_str(), "c_void");
+                        });
+
+                        acx.c_void_ptrs.insert(*p);
+                    };
+
+                    use crate::util::Callee::*;
```
kkysen, line 482: I think it can just be `use Callee::*` with a top-level `use crate::util::Callee`.

c2rust-analyze/src/main.rs
kkysen: Hmm, I'm guessing `libc::c_void` now just re-exports `core::ffi::c_void`, so from inside `rustc` we just see that it's `core::ffi::c_void`.  If we want to support older versions, then we'd need to allow `libc::c_void`, but I don't think that's important, so it should be good as is.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -348,8 +348,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
             StatementKind::Assign(ref x) => {
                 let (pl, ref rv) = **x;
 
-                if matches!(rv, Rvalue::Cast(_, Operand::Copy(p) | Operand::Move(p), _) if self.special_casts.contains_key(&p))
-                {
+                if self.special_casts.is_special(&pl, &rv) {
```
kkysen: `is_special` checks `self.0.contains_key(lhs)`, so `self.special_casts.contains_key(pl)`, which wasn't checked previously here.  This seems more correct, but I just want to check to make sure.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -348,7 +348,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
             StatementKind::Assign(ref x) => {
                 let (pl, ref rv) = **x;
 
-                if self.special_casts.is_special(&pl, &rv) {
+                if self.c_void_casts.is_special(&pl, &rv) {
```
kkysen, line 351: Should it still be `is_special` when it's `c_void_casts` now?

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -335,6 +347,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         match stmt.kind {
             StatementKind::Assign(ref x) => {
                 let (pl, ref rv) = **x;
+
+                if self.c_void_casts.is_special(&pl, &rv) {
+                    // skip this cast, because the local that is getting casted
+                    // originates from a call to an allocation that is handled
+                    // in a way that effectively elides the cast
```
kkysen, line 354: Can you add part of these docs to `is_special` (or whatever we call it), too?

c2rust-analyze/src/util.rs
```diff
@@ -2,11 +2,25 @@ use crate::labeled_ty::LabeledTy;
...
 use rustc_middle::ty::{AdtDef, DefIdTree, SubstsRef, Ty, TyCtxt, TyKind, UintTy};
+use std::collections::HashMap;
 use std::fmt::Debug;
 
+#[derive(Default)]
+pub struct CVoidCasts<'tcx>(pub HashMap<Place<'tcx>, Place<'tcx>>);
+
+impl<'tcx> CVoidCasts<'tcx> {
+    pub fn is_special(&self, lhs: &Place<'tcx>, rv: &Rvalue<'tcx>) -> bool {
+        matches!(rv, Rvalue::Cast(_, Operand::Copy(p) | Operand::Move(p), _) if self.0.contains_key(p) || self.0.contains_key(lhs))
```
kkysen, line 16: nit:

```suggestion
        matches!(rv, Rvalue::Cast(_, Operand::Copy(p) | Operand::Move(p), _) if self.0.contains_key(p)) || self.0.contains_key(lhs)
```

c2rust-analyze/src/main.rs
kkysen:
```suggestion
        for term in mir
            .basic_blocks()
            .iter()
            .filter_map(|bb| bb.terminator.as_ref())
        {
```

This should work (also needs to remove the corresponding `}` and re-format).  Not really that important, but it's helpful for reading to avoid very indented code.

c2rust-analyze/src/main.rs
```diff
@@ -465,7 +514,22 @@ fn run(tcx: TyCtxt) {
...
+                            acx.c_void_casts.0.insert(p, lhs);
+                        }
+
+                        if acx.c_void_ptrs.contains(&lhs) {
+                            // This is a special case for types being casted to *c_void
+                            acx.c_void_casts.0.insert(lhs, op.place().unwrap());
+                        }
+
+                        continue;
+                    }
```
kkysen, line 532: The `Operand` of an `Rvalue::Cast` is never `Operand::Constant`, right?  Then we can just

```rust
let p = op.place().expect("Rvalue::Cast is not Operant::Constant so it has a Place");
```

which simplifies things and makes the code's intent more clear (right now there's one `if let Some` and one `.unwrap()`).

c2rust-analyze/src/util.rs
```diff
@@ -2,11 +2,25 @@ use crate::labeled_ty::LabeledTy;
...
+    }
+
+    pub fn get_or_default_to(&self, p: &Place<'tcx>) -> Place<'tcx> {
+        self.0.get(p).unwrap_or(p).clone()
```
kkysen, line 20:
```suggestion
        *self.0.get(p).unwrap_or(p)
```

`Place` is `Copy`.

c2rust-analyze/src/main.rs
kkysen: Can you extract out the

```rust
if self.ptrs.contains(&p1) {
    self.casts.0.insert(p1, p2);
}
```

part of this into some `struct CVoids`?  Then it's clear the logic is the same for both, just flipped.

c2rust-analyze/src/borrowck/type_check.rs
kkysen:
```suggestion
                if self.c_void_casts.is_special(&pl, rv) {
```

c2rust-analyze/src/context.rs
```diff
@@ -84,12 +84,25 @@ pub struct AnalysisCtxt<'a, 'tcx> {
 
     pub local_decls: &'a LocalDecls<'tcx>,
     pub local_tys: IndexVec<Local, LTy<'tcx>>,
+    pub c_void_ptrs: HashSet<Place<'tcx>>,
+    /// A mapping for substituting [`Place`]s adhering to the
+    /// following pattern
+    /// ```mir
+    /// _1 = malloc(...);
+    /// _2 = _1 as *mut T;
+    /// ```
```
kkysen, line 93:
```suggestion
    /// ```
    /// where `_1` is `*c_void`.
```

spernsteiner:
It looks like the strategy here is:

1. Collect the destination and/or argument places of all `malloc`/`free`/`realloc` calls into `c_void_ptrs`
2. Record in `c_void_casts` the source and destination places of all casts involving one of the `c_void_ptrs`
3. When processing a `malloc`/`free`/`realloc` call, if its destination and/or argument place is in `c_void_casts`, use the place from the other side of the cast instead

My worry with this approach is that there could be other operations on the relevant places in between the cast and the `malloc`/`free`.  For example:
```Rust
_2 = _1 as *mut c_void;
_2 = _3;
free(_2);
```
I believe the current implementation would treat this as freeing `_1`, when in fact it frees `_3`.

Another potential issue would be the same place appearing in multiple casts.  For example:
```Rust
_3 = _1 as *mut c_void;
free(_3);
_3 = _2 as *mut c_void;
free(_3);
```
I'm not 100% sure but I think the current implementation would either treat both `free`s as freeing `_1` or else treat both as freeing `_2` (both of which are incorrect).

Note there may also be problems if the RHS of the cast (not just the LHS) is written between the cast and the `free`:
```Rust
_2 = _1 as *mut c_void;
_1 = _3;
free(_2);
```
Here it's wrong to say that `free(_2)` is equivalent to `free(_1)`, since `_1` and `_2` could actually have different values at the time of the `free`.

I think we should be much stricter about what code patterns we recognize when looking for the `malloc`/`free`+cast combo.  For `malloc`, we should require specifically a `malloc` call whose successor block begins with a cast of the destination, like this:
```Rust
    bb0: {
        _1 = malloc(const 8_usize) -> bb1;
    }

    bb1: {
        _0 = move _1 as *mut u8 (PtrToPtr);
        ...
    }
```
And for `free`, we want the immediate preceding statement to cast and set the argument:
```Rust
        _3 = move _4 as *mut libc::c_void (PtrToPtr);
        _2 = free(move _3) -> bb1;
```
`realloc` needs to check for casts both before and after, but this may be a bit more complicated since there could be arbitrary code involved in the evaluation of the size argument.  Here's an example where it calls an additional function, meaning the cast for the argument is actually in a different block from the `realloc` itself:
```Rust
bb0: {
    ...
    _3 = move _4 as *mut libc::c_void (PtrToPtr);
    _5 = get_size() -> bb1;
}

bb1: {
    _2 = libc::realloc(move _3, move _5) -> bb2;
}

bb2: {
    _0 = move _2 as *mut u8 (PtrToPtr);
    ...
}
```
For this, we'd need to ensure that the code between the cast and the `realloc` call updates neither the LHS nor the RHS of the cast operation, which could get complex, particularly if one of the two places involves a deref projection.  We should only handle as many cases here as we actually need to for lighttpd, which might be none at all (i.e. the cast must come immediately before the `realloc`, with no intervening code) .

Once the `malloc`/`free`+cast pattern has been identified, that information should be associated with the `Location` of one of the relevant statements (probably the call itself - I think that would be easiest) rather than a `Place`.  A `Place` can be reused in multiple parts of the same function, but a `Location` is unique.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: f7909a7b

c2rust-analyze/src/main.rs
aneksteind: 715944f6

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: e2b16ed3

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: this is the behavior I want, albeit unintended

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: 912688a9

c2rust-analyze/src/main.rs
aneksteind: `expect` added in 1d95fed5. will delay extracting the other code into a method -- that pattern only shows up twice AFAICT.

aneksteind:
Thanks for the feedback @spernsteiner, I'll work towards what you've suggested.

aneksteind:
@spernsteiner it looks like your example above is at least already in the test cases in `alloc.rs`:
```
bb0: {
    ...
    _3 = move _4 as *mut libc::c_void (PtrToPtr);
    _5 = get_size() -> bb1;
}

bb1: {
    _2 = libc::realloc(move _3, move _5) -> bb2;
}

bb2: {
    _0 = move _2 as *mut u8 (PtrToPtr);
    ...
}
```

```
bbx: {
 _3 = move (_11.0: i32), 
 StorageLive(_12), 
 StorageLive(_13), 
 StorageLive(_14), 
 _14 = _1, 
 _13 = move _14 as *mut libc::c_void (Misc), 
 StorageDead(_14), 
 StorageLive(_15), 
 StorageLive(_16),
 std::mem::size_of::<i32>() -> [return bb6, ...]
}

bb6: BasicBlockData { 
 _15 = move _16 as u64 (Misc),
 StorageDead(_16)], 
 realloc(move _13, move _15) ...
},
```

But I will change the tests such that the cast is in the same basic block as the `realloc` call.

I'll try to find a `lighttpd` example with `realloc` to test. AFAICT there is no call to one in the subset of functionality we've been testing.

spernsteiner:
The examples in that post are cases I think would be difficult for the new, stricter pattern matching I'm proposing.  It's possible they work fine on the current implementation, but I think they would break under the new one if it was overly strict.

aneksteind:
> I think they would break under the new one if it was overly strict.

@spernsteiner I agree and that's why I mention it. Locally I've implemented a strict version that fails in the above test case. But after changing the test such that there is no intermediate `std::mem::size_of` call, the issue goes away. I can't see `lighttpd` code that makes those intermediate calls so my hope is that the strict approach still works for lighttpd, which is what I will find out.

kkysen:
What would the future approach be to more general uses of `malloc`, `realloc`, `free`?  I think it's fine to limit it to the strict use cases we're coming across in `lighttpd` for now, but I think we should have an idea for how this would be implemented in a more general case to make sure it's feasible enough.

For example, something like this:

```c
#include <stdlib.h>

int main() {
    void* a = malloc(sizeof(int));
    if (a) {
        int* b = a;
        *b = 1;
        free(a);
    }
}
```

aneksteind:
@kkysen i'm not sure what you mean. I've added a test case similar to the example you provided and it passes tests: c8341290

spernsteiner:
>I've added a test case similar to the example you provided and it passes tests

The test you added is slightly different - it casts from `*mut c_void` to `*mut i32` immediately, rather than waiting until inside the conditional.  @kkysen's version ought to fail with the stricter checking, since there is nontrivial code between the `malloc` and the cast.

aneksteind:
@spernsteiner @kkysen yes, casting within the conditional does indeed fail the test

aneksteind:
@spernsteiner is there anything else you'd like to see addressed?

kkysen:
> @kkysen i'm not sure what you mean. I've added a test case similar to the example you provided and it passes tests: [387d50b](https://github.com/immunant/c2rust/commit/387d50b9119e61d6313deb7f823890dfd5b332f7)

Even though this is a bit different from the failing one I suggested, I don't think it's bad to have some extra tests.  I think you can drop the revert commit.

c2rust-analyze/src/util.rs
```diff
@@ -2,11 +2,40 @@ use crate::labeled_ty::LabeledTy;
...
+/// ```mir
+/// _1 = malloc(...);
+/// _2 = _1 as *mut T;
+/// ```
+///
+/// In this case, `_1` would be mapped to `_2`, which is indicative
+/// of the amended statement:
+/// ```
+/// _2 = malloc(...);
+/// ```
```
kkysen, line 22:
```suggestion
/// A mapping for substituting [`Place`]s adhering to the
/// following pattern
///
/// ```mir
/// _1 = malloc(...);
/// _2 = _1 as *mut T;
/// ```
///
/// where `_1` is [`*c_void`](core::ffi::c_void).
///
/// In this case, `_1` would be mapped to `_2`, which is indicative
/// of the amended statement:
///
/// ```mir
/// _2 = malloc(...);
/// ```
```

c2rust-analyze/src/util.rs
```diff
@@ -2,11 +2,40 @@ use crate::labeled_ty::LabeledTy;
...
+#[derive(Default)]
+pub struct CVoidCasts<'tcx>(pub HashMap<Place<'tcx>, Place<'tcx>>);
+
+impl<'tcx> CVoidCasts<'tcx> {
+    /// Checks if the casted-to or casted-from value
+    /// is a c_void pointer.
```
kkysen, line 28:
```suggestion
    /// Checks if the casted-to or casted-from value is a [`*c_void`](core::ffi::c_void).
```

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,121 @@ fn run(tcx: TyCtxt) {
...
+                                        x.0,
+                                        op.place().expect(
+                                            "Rvalue::Cast is not Operant::Constant so it has a Place",
+                                        ),
+                                    )),
+                                    _ => None,
+                                },
+                                _ => None,
+                            })
+                    };
```
kkysen, line 533: The two `find_map` closures are the same and can be refactored out.  I don't say this just to reduce repetition, but there are often little differences that sneak into one of the versions when they're separate.  Like there was for `CVoidCasts::contains` until it was refactored into a single version.

c2rust-analyze/src/util.rs
aneksteind: I recall you mentioning we should avoid specific `c_void` instances. Is that a concern here?

c2rust-analyze/src/main.rs
aneksteind: They are actually slightly different; they return tuples in reverse order of one another. But I will look for a way to factor it out anyway

c2rust-analyze/src/util.rs
aneksteind: Same question as https://github.com/immunant/c2rust/pull/788#discussion_r1102108079

c2rust-analyze/src/main.rs
aneksteind: 60fe64d8

kkysen: Thanks, that's a lot clearer now.  Much easier to see they're flipped in order now.

c2rust-analyze/src/util.rs
kkysen: I don't think so. This one is stable now and is just for docs anyways.

aneksteind:
@kkysen I reverted the revert in 28986313

c2rust-analyze/src/util.rs
```diff
@@ -2,11 +2,43 @@ use crate::labeled_ty::LabeledTy;
...
+/// where `_1` is [`*c_void`](core::ffi::c_void).
+///
+/// In this case, `_1` would be mapped to `_2`, which is indicative
+/// of the amended statement:
+///
+/// ```mir
+/// _2 = malloc(...);
+/// ```
+#[derive(Default)]
+pub struct CVoidCasts<'tcx>(pub HashMap<Place<'tcx>, Place<'tcx>>);
```
kkysen, line 28: Should the two mappings, the alloc one and the free one, be conflated?  Or should there be two maps?

aneksteind: For now it makes no difference. We treat casts to/from `*c_void` the same. When we start to treat them differently then it may make sense to separate them.

kkysen: We don't really treat them totally the same.

In

```rust
    pub fn contains(&self, lhs: &Place<'tcx>, rv: &Rvalue<'tcx>) -> bool {
        matches!(rv, Rvalue::Cast(_, Operand::Copy(p) | Operand::Move(p), _) if self.0.contains_key(p))
            || self.0.contains_key(lhs)
    }
```

one `.contains_key` is for to casts and one is for from casts.  Couldn't this potentially lead to a false positive?

aneksteind:

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,117 @@ fn run(tcx: TyCtxt) {
...
+                            .map(|(casted_to, casted_from)| (casted_from, casted_to))
+                    };
+
+                    let find_first_cast_succ_block = || {
+                        let successor_block_id = target.unwrap();
+                        mir.basic_blocks()[successor_block_id]
+                            .statements
+                            .iter()
+                            .find_map(|s| get_cast_op(s))
+                    };
```
spernsteiner, line 529: In the `malloc(...) as *mut T` pattern, can the `Rvalue::Cast` appear anywhere other than the first statement of the successor block?  My guess is that it can't (even in cases where the function arguments involve temporaries that need to be dropped, it looks like the drop happens after the cast).  If that's the case, then I think this should use something like `statements.first().filter_map(...)` rather than `statements.iter().find_map(...)` to ensure that it only considers the first statement in the block.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,117 @@ fn run(tcx: TyCtxt) {
...
+
+                    let find_last_cast_curr_block = || {
+                        bb_data
+                            .statements
+                            .iter()
+                            .rev()
+                            .find_map(|s| get_cast_op(s))
+                            .map(|(casted_to, casted_from)| (casted_from, casted_to))
+                    };
```
spernsteiner, line 521: We need a check somewhere to ensure that the LHS of the cast isn't modified between the cast and the `free` call.  With the current implementation, the correctness of our `free` handling depends on the fact that rustc currently produces suboptimal MIR for code like this:
```Rust
let mut q = p as *mut c_void;
q = x;
free(q);
```
The corresponding MIR:
```Rust
_4 = p;
q = move _4 as *mut libc::c_void (PtrToPtr);
_5 = x;
q = move _5;
_7 = q;
_6 = free(move _7) -> bb1;
```
If rustc changed to produce `_6 = free(q) -> bb1;` (eliminating the unnecessary temporary `_7`), then I believe our current implementation would wrongly recognize this as `free(_4)`.

Checking whether a statement could modify a particular `Place` is nontrivial (particularly if the `Place` involves `Deref` projections), but currently rustc always uses a temporary for the LHS of the cast in the desired `free(ptr as *mut c_void)` pattern, so I think we can conservatively treat the `Place` as unmodified only if (1) it's just a `Local`, with no projections, and (2) that local isn't mentioned in the LHS or RHS of any statements between the cast and the call.

c2rust-analyze/src/main.rs
```diff
@@ -449,13 +449,117 @@ fn run(tcx: TyCtxt) {
...
+                        _ => None,
+                    };
+
+                    let find_last_cast_curr_block = || {
+                        bb_data
+                            .statements
+                            .iter()
+                            .rev()
+                            .find_map(|s| get_cast_op(s))
+                            .map(|(casted_to, casted_from)| (casted_from, casted_to))
```
spernsteiner, line 520: This seems like a really confusing way to implement this.  For one thing, it means that the `casted` and `castee` variables above could either be the source and destination of the cast, or they could be the destination and the source.

kkysen: I had a follow-up PR to this, #826, which does I think clarify this part of the code with `struct CVoidCast`, if you think that's a clearer way to organize this logic.

c2rust-analyze/src/util.rs
spernsteiner: I agree with @kkysen that conflating casts from `void*` and casts to `void*` seems risky.  Here's an example:
```Rust
x = malloc(...);
p = x as *mut u8;
p = y;
free(x);
```
Here, `x` is cast to `p` after the `malloc`, so the map will contain an entry `(x, p)`.  There is no cast whose LHS is `x`, so the `free` won't add any entries to the map.  But when processing the `free(x)`, we'll see the entry `(x, p)` and treat it as `free(p)`, which is incorrect.

To reiterate from my comment above, I think the maps should be keyed on the MIR `Location` of the `malloc`/`free` so there can't be any ambiguity arising from the same `Place` appearing in more than one call.

c2rust-analyze/src/main.rs
aneksteind: > In the malloc(...) as *mut T pattern, can the Rvalue::Cast appear anywhere other than the first statement of the successor block?

Yes it can. We can have `StorageDead` statements in between. Should those be ignored? If they are ignored then I think you may be correct in that the cast is the next type of statement.

spernsteiner: >We can have StorageDead statements in between. Should those be ignored?

Yes, I think we should just ignore `StorageDead` at the start of the successor block.

`StorageDead` corresponds to `@llvm.lifetime.end`, which overwrites all bytes of the value with poison.  If we were doing an actual code transformation on the MIR, we'd need to be careful about where exactly we place those relative to the call, and we'd probably need to check for overlap between `StorageDead` locals and locals that appear in the inputs/outputs of the call and the cast.  But for the flow-insensitive analysis that we're actually using, that ordering of statements doesn't matter.

c2rust-analyze/src/main.rs
aneksteind: I merged #826

c2rust-analyze/src/main.rs
aneksteind: @spernsteiner I have something in progress in #848, would you mind taking a look? I'm almost positive that this won't be the case in the reverse with `realloc`, however. There can be statements and entire blocks between casts to `void*` and the `realloc` call. Because there won't be a fixed distance between the two in this case, and no longer a fixed distance between allocations and casts from void, I added something in #848 to key on the location of both the call and the cast. The former to know what updated `Place` to use, and the latter to know whether or not to skip a cast in `visit_statement`

c2rust-analyze/src/util.rs
aneksteind: Would you say #826 covered this adequately?

c2rust-analyze/src/main.rs
aneksteind: I've addressed this in #848

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,368 @@
...
+                |get_cast| bb_data.statements.iter().rev().find_map(get_cast);
+
+            for direction in CVoidCastDirection::from_callee(ty_callee(tcx, func_ty))
+                .iter()
+                .copied()
+            {
+                use CVoidCastDirection::*;
+                let c_void_ptr = match direction {
+                    From => destination,
+                    To => args[0].place().unwrap(),
```
spernsteiner, line 347: This `unwrap` will fail on the following code:
```Rust
use libc::{free, c_void};
const NULL: *mut c_void = 0 as *mut _;
unsafe fn f() {
    free(NULL);
}
```
The MIR for this has an `Operand::Constant` for the argument to `free`.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// Try to get the appropriate [`CVoidCast`]
+    /// from this [`Statement`] in the given [`CVoidCastDirection`].
+    ///
+    /// For the [`From`] direction, the [`*c_void`] is on the rhs.\
+    /// For the [`To`] direction, the [`*c_void`] is on the lhs.
+    ///
+    /// [`*c_void`]: core::ffi::c_void
+    /// [`From`]: CVoidCastDirection::From
+    /// [`To`]: CVoidCastDirection::To
+    pub fn get_cast_from_stmt(
```
spernsteiner, line 109: While reviewing #848, it took me a bit to figure out where exactly we check that the `Place` being cast from/to is actually `c_void_ptr_place`.  Can you update the doc comment here to clarify that `get_cast_from_stmt` only returns `Some` if the `void*` side of the cast matches `self`?

c2rust-analyze/src/c_void_casts.rs
aneksteind: If you're okay with it I'd like to make it an `expect` instead and address the case you outlined in the same PR as the one where we support null pointers

spernsteiner: Seems reasonable

aneksteind: 9c17f600

c2rust-analyze/src/c_void_casts.rs
aneksteind: 66cd1153

c2rust-analyze/src/util.rs
```diff
@@ -201,28 +202,58 @@ fn builtin_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, did: DefId) -> Option<C
...
+                return None;
+            }
+            if path.data.len() != 2 {
+                return None;
+            }
+            if path.data[0].to_string() != "mem" {
+                return None;
+            }
+            Some(Callee::Trivial)
+        }
```
kkysen, line 233: Btw, this has been removed in `master` by now.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -415,7 +441,13 @@ pub fn visit<'tcx>(
                 },
             );
         }
-        tc.visit_terminator(bb_data.terminator());
+        tc.visit_terminator(
+            bb_data.terminator(),
+            Location {
+                statement_index: bb_data.statements.len(),
+                block: bb,
+            },
```
kkysen, line 449:
```suggestion
            terminator_location(bb, bb_data)
```

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// nontrivial (particularly if the Place involves Deref projections),
+    /// but currently rustc always uses a temporary for the LHS of the
+    /// cast in the desired free(ptr as *mut c_void) pattern, so I think we
+    /// can conservatively treat the Place as unmodified only if (1) it's
+    /// just a Local, with no projections, and (2) that local isn't
+    /// mentioned in the LHS or RHS of any statements between the cast and
+    /// the call.
+    fn is_place_modified_by_statement(p: &Place, stmt: &Statement) -> bool {
+        if !p.projection.is_empty() {
+            return false;
```
kkysen, line 322:
```suggestion
            return true;
```

A `Place` is unmodified if (1) and (2), where (1) is a `Local` with no projections.  So if a `Place` has projections (`!p.projection.is_empty()`), the `Place` is considered modified, so this should `return true` here.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+    ///
+    /// [`From`]: CVoidCastDirection::From
+    /// [`To`]: CVoidCastDirection::To
+    other_ptr: Place<'tcx>,
+}
+
+/// A mapping from [`*c_void`](core::ffi::c_void)s ([`CVoidPtr`]s)
+/// to their properly typed pointers,
+/// like [`CVoidCasts`], but in a single direction, meaning it represents either
+/// [`CVoidCastDirection::From`] or [`CVoidCastDirection::To`].
```
kkysen, line 167: Needs to be updated.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+    calls: HashMap<Location, CVoidCast<'tcx>>,
+    casts: HashMap<Location, CVoidCast<'tcx>>,
+}
+
+impl<'tcx> CVoidCastsUniDirectional<'tcx> {
+    /// Get the adjusted [`Place`], skipping over [`*c_void`](core::ffi::c_void) intermediaries.
+    ///
+    /// That is, if `place` is a [`CVoidPtr`] in this map of [`CVoidCast`]s,
+    /// then the [`Place`] of its other, property-typed pointer is returned.
+    /// Otherwise, the same `place` is returned, as no adjustments are necessary.
```
kkysen, line 179: Needs to be updated.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// * the lhs [`Place`] of the [`Assign`] is a [`CVoidPtr`] in the [`To`] direction,
+    ///     as in the [`To`] direction, the [`CVoidPtr`] is on the lhs.
+    ///
+    /// * the rhs [`Place`] of the [`Cast`] of the [`Assign`] is a [`CVoidPtr`] in the [`From`] direction,
+    ///     as in the [`From`] direction, the [`CVoidPtr`] is on the rhs.
+    ///
+    /// [`Assign`]: rustc_middle::mir::StatementKind::Assign
+    /// [`Cast`]: rustc_middle::mir::Rvalue::Cast
+    /// [`From`]: CVoidCastDirection::From
+    /// [`To`]: CVoidCastDirection::To
```
kkysen, line 307: Needs to be updated.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    /// Search for the last cast to a void pointer in a sequence of
+    /// [Statement]s.
+    fn find_last_cast(
+        statements: &[Statement<'tcx>],
+        c_void_ptr: CVoidPtr<'tcx>,
+    ) -> Option<(usize, CVoidCast<'tcx>)> {
+        // Get the last cast in the current block, and ensure that the destination
+        // of that cast remains unmodified between the cast and the call in which
+        // the destination place is used.
```
kkysen, line 390: Can you move the comment up to part of the doc comment?

spernsteiner (commented):
The only remaining major issue for me is `is_place_modified_by_statement` being insufficiently conservative.  Otherwise it looks good

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+                ShallowInitBox(op, _) => op.place(),
+                CopyForDeref(..) => None,
+            };
+
+            if let Some(rv_local) = rv_place.map(|p| p.local) {
+                return local == rv_local;
+            }
+        }
+
+        false
```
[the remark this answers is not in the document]
aneksteind: 436d7007

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+                Repeat(op, _) => op.place(),
+                Ref(_, _, p) => Some(*p),
+                ThreadLocalRef(..) => None,
+                AddressOf(_, p) => Some(*p),
+                Len(p) => Some(*p),
+                Cast(_, op, _) => op.place(),
+                BinaryOp(..) => None,
+                CheckedBinaryOp(..) => None,
+                NullaryOp(..) => None,
+                UnaryOp(_, op) => op.place(),
```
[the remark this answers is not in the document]
aneksteind: I addressed what I thought were the obvious cases in a2d06b28

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+/// A mapping from [`*c_void`](core::ffi::c_void)s ([`CVoidPtr`]s)
+/// to their properly typed pointers,
+/// like [`CVoidCasts`], but in a single direction, meaning it represents either
+/// [`CVoidCastDirection::From`] or [`CVoidCastDirection::To`].
+#[derive(Default, Clone, Debug)]
+pub struct CVoidCastsUniDirectional<'tcx> {
+    calls: HashMap<Location, CVoidCast<'tcx>>,
+    casts: HashMap<Location, CVoidCast<'tcx>>,
```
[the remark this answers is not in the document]
aneksteind: It is for `calls` but not for `casts`. Changed in 850a0132

c2rust-analyze/src/c_void_casts.rs
aneksteind: ace07810

c2rust-analyze/src/c_void_casts.rs
aneksteind: ace07810

c2rust-analyze/src/c_void_casts.rs
aneksteind: ace07810

aneksteind: c9eb09c3

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,460 @@
...
+            StatementKind::Assign(assign) => {
+                let (lhs, rv) = &**assign;
+                if lhs.local == local {
+                    return true;
+                }
+
+                use Rvalue::*;
+                let rv_place = match rv {
+                    Use(op) => op.place(),
+                    Ref(_, _, p) => Some(*p),
```
spernsteiner, line 317: `AddressOf` should be handled the same as `Ref`, and treated as a potential write.  The reason is that once the address is taken, we can no longer tell which statements might modify the local.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,460 @@
...
+                    Cast(_, op, _) => op.place(),
+                    Discriminant(p) => Some(*p),
+                    ShallowInitBox(op, _) => op.place(),
+                    _ => return false,
```
spernsteiner, line 321: I think it would be safer here to explicitly spell out all the cases, even though it's a bit verbose.  Someday we will update to a newer version of `rustc`, and if it adds a new `Rvalue` kind, we want it to either fail to compile (so we know to check this code and decide what to do with the new variant) or to default to `true` (the conservative answer).

spernsteiner (approved):
Good to merge once the `AddressOf`/`Ref` thing is fixed

PR #796, opened 2023-01-25
Replace negation and unsigned casts with `.unsigned_abs()` to avoid overflows

### LINKED ISSUE #795: Debug build panicking when transpiling systemd: 'attempt to negate with overflow'
(PR fix incoming)

Context:
 - Fedora 37 x64
 - local debug build of c2rust in `../c2rust`
 - rustc 1.66.1 (90743e729 2023-01-10) 
 
Steps to reproduce:

    git clone git@github.com:systemd/systemd.git
    cd systemd
    env CC=$(which clang) meson build
    bear -- make all
    env RUST_BACKTRACE=1 ../c2rust/target/debug/c2rust transpile compile_commands.json

Actual output:
```
[...]
Transpiling libudev-device.c
thread 'main' panicked at 'attempt to negate with overflow', c2rust-transpile/src/translator/mod.rs:407:40
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
             at /rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/library/core/src/panicking.rs:48:5
   3: c2rust_transpile::translator::signed_int_expr
             at /var/home/fralalonde/Code/github/c2rust/c2rust-transpile/src/translator/mod.rs:407:40
   4: c2rust_transpile::translator::Translation::convert_decl
             at /var/home/fralalonde/Code/github/c2rust/c2rust-transpile/src/translator/mod.rs:1867:47
[...]
```
    
Expected output: no stack trace, same behavior as release build of c2rust (but slower, obviously)

fralalonde:
- Fixes #795.

This fixes instances of `\(?-[a-zA-Z_]+\)? as ` in which `(-value) as U` is done, where `value` is signed and `U` is unsigned.  This overflows (and panics in debug mode) on the minimum value.  What we want instead is `value.unsigned_abs()`, which does the negation and unsigned cast in one safe operation.

Note: I've replaced the `as u128` with `u128::from(..)` as it is considered more robust form but would not mind sticking with original cast style for code uniformity if required.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3986,7 +3986,7 @@ impl<'c> Translation<'c> {
 
             ConstIntExpr::I(n) => mk().unary_expr(
                 UnOp::Neg(Default::default()),
-                mk().lit_expr(mk().int_unsuffixed_lit((-n) as u128)),
+                mk().lit_expr(mk().int_unsuffixed_lit(n.unsigned_abs() as u128)),
             ),
         };
         Ok(expr)
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -404,7 +404,7 @@ pub fn signed_int_expr(value: i64) -> Box<Expr> {
     if value < 0 {
         mk().unary_expr(
             UnOp::Neg(Default::default()),
-            mk().lit_expr(mk().int_lit((-value) as u128, "")),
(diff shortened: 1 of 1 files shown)


kkysen (approved):

kkysen:
Good catch, @fralalonde!  Thanks for the fix.

`as` casts are wonky in Rust, so `::from` is fine.

kkysen:
There's actually another occurrence of this bug.  I'll push a fix for that to this branch, too.

PR #802, opened 2023-01-28
Release 0.17.0

### LINKED ISSUE #690: Update crates.io release
Please consider tagging and uploading a release for crates.io again, possibly after merging #677 [edit: and #693 ;-) ]. There have been several useful additions to c2rust around atomics, docstring handling and error handling; it would be great to have them at a simple `cargo install c2rust`.

Thanks for all the enhancements!

fw-immunant:
- Fixes #690.

This has been a long time coming, but we've now merged all the dep updates and gotten our release process in shape.

### DIFF
--- a/c2rust-analyze/Cargo.toml
+++ b/c2rust-analyze/Cargo.toml
@@ -19,11 +19,11 @@ assert_matches = "1.5.0"
 indexmap = "1.9.2"
 
 [build-dependencies]
-c2rust-build-paths = { path = "../c2rust-build-paths" }
+c2rust-build-paths = { path = "../c2rust-build-paths" , version = "0.17.0" }
 print_bytes = "1.1"
 
 [dev-dependencies]
-c2rust-build-paths = { path = "../c2rust-build-paths" }
+c2rust-build-paths = { path = "../c2rust-build-paths" , version = "0.17.0" }
 
 [package.metadata.rust-analyzer]
 rustc_private = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,7 +124,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "anyhow",
  "c2rust-build-paths",
@@ -140,7 +140,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analysis-rt"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "bincode",
  "enum_dispatch",
@@ -151,7 +151,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analyze"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "assert_matches",
  "bitflags",
@@ -164,11 +164,11 @@ dependencies = [
 
 [[package]]
 name = "c2rust-asm-casts"
-version = "0.16.0"
+version = "0.17.0"
 
 [[package]]
 name = "c2rust-ast-builder"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "proc-macro2",
  "syn",
@@ -176,7 +176,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-ast-exporter"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "bindgen",
  "c2rust-build-paths",
@@ -191,7 +191,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-ast-printer"
-version = "0.16.0"
+version = "0.17.0"
 dependencies = [
  "log",
  "prettyplease",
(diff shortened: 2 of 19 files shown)


fw-immunant:
Hoping to make the release late Monday if no issues crop up.

scripts/update-versions.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
+if [ "$#" -ne 1 ]; then
+	echo "usage: $(basename $0) <NEW_VERSION>" > /dev/stderr
+	exit 1
+fi
+version=$1
+
+toml_cli=${TOML_CLI:-${CARGO_HOME:-~/.cargo}/bin/toml}
```
kkysen, line 9: Would something like [`cargo-release`](https://github.com/crate-ci/cargo-release) work instead?

scripts/update-versions.sh
```diff
@@ -0,0 +1,58 @@
...
+
+update_key() {
+	toml=$1
+	key=$2
+	value=$3
+
+	old_value=$($toml_cli get $toml $key)
+	# only set value if already present
+	if [ ! -z "$old_value" ]; then
+		$toml_cli set $toml $key $value | sponge $toml
```
kkysen, line 34: What's `sponge`?

fw-immunant: https://manpages.debian.org/testing/moreutils/sponge.1.en.html

It avoids the pipeline truncating the .toml file to use as an output before it can be read from to use as an input.

scripts/update-versions.sh
fw-immunant: Possibly--I don't know well what packaging tools exist outside of Cargo itself, and the Rust/Cargo docs didn't point to cargo-release. I just wanted to automate what I knew I needed done both to make sure I didn't miss anything and as documentation for the next person to do a release.

kkysen (changes requested):
There are a few things missing from this.  Published packages require `description`, `license`/`license-file`, and `documentation`/`homepage`/`repository` fields, which most of the new packages don't have.  To fix part of this, I opened #804, which also unifies package versions, which makes this PR simpler.  #805 then adds the package descriptions.

Also, I don't think we should bump version numbers for the unmaintained packages, namely `c2rust-macros` and `c2rust-refactor`.  I don't think we should publish new versions of them if they are unchanged and unmaintained.

I'm also not sure the `update-versions.sh` script is the best idea vs. something much more polished and widely used like `cargo-release`.  `cargo-release` should get all the little details right that this script probably won't.  For example, it immediately detected that we're missing a bunch of the required package fields for publishing.  If there's a suitable alternative that's fairly widely used, I think that's better than our own script that we'll have to maintain, which also adds dependencies on things like `sponge`.

Regardless, I think we should merge #804 first, which then simplifies this PR, and also open a PR that includes `description`s for all of the new packages.

fw-immunant:
Sharing versions via workspace inheritance breaks the script anyway, so I've removed it. I also couldn't get `cargo-release` to do the right thing, trying `cargo release minor` and `cargo release --workspace minor` and also the `--dependent-version=upgrade` flag.

kkysen:
> I think we should bump version numbers for all crates in the repo to avoid divergence. If we want to delete the c2rust-refactor and c2rust-macros crates entirely, we can--they'll still be in the history.

Not sure it's the best idea, but I don't think it really matters either way.  They were bumped to `0.16.0` even after being retired anyways.

> We don't need to release old, deprecated crates on crates.io

👍 

> nor do I think it makes sense to release the new analysis/runtime/pdg crates; they aren't in any state where it would be a good idea to get them from crates.io rather than GitHub.

👍 

> We should release just these crates:
> 
> * c2rust
> * c2rust-asm-casts
> * c2rust-ast-builder
> * c2rust-ast-exporter
> * c2rust-ast-printer
> * c2rust-transpiler
> * c2rust-bitfields
> * c2rust-bitfields-derive
> * c2rust-build-paths
> * c2rust-transpile

👍 , except I think there's a typo in `c2rust-transpiler` (duplicate/nonexistent)

c2rust-analyze/Cargo.toml
```diff
@@ -19,11 +19,11 @@ assert_matches = "1.5.0"
 indexmap = "1.9.2"
 
 [build-dependencies]
-c2rust-build-paths = { path = "../c2rust-build-paths" }
+c2rust-build-paths = { path = "../c2rust-build-paths" , version = "0.17.0" }
```
kkysen, line 22: Usually `version` goes before `path` and there's no space before the `,`.

kkysen (approved):
One minor nit, but otherwise I think it's all good for a release now!

kkysen:
> Sharing versions via workspace inheritance breaks the script anyway, so I've removed it.

👍

> I also couldn't get `cargo-release` to do the right thing, trying `cargo release minor` and `cargo release --workspace minor` and also the `--dependent-version=upgrade` flag.

`cargo release version minor` at least worked for upgrading the versions and dependencies for me, though it doesn't upgrade them in non-workspace packages like the `analysis/tests/` or `tests/` crates.

PR #804, opened 2023-01-30
Use `workspace.package` inheritance for most package fields

kkysen:
This uses `workspace.package` inheritance for most package fields to standardize and sync fields.

Most packages repeat a lot of the package fields with minor variations, or omit them entirely (including ones like `license` that are mandatory for publishing).  This uses `cargo`'s [workspace inheritance](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) to set defaults in the workspace root's `Cargo.toml`, which are then either explicitly inherited in each `Cargo.toml` or overrided.

The defaults used here are:

```toml
version = "0.16.0"
authors = ["The C2Rust Project Developers <c2rust@immunant.com>"]
edition = "2021"
license = "BSD-3-Clause"
homepage = "https://c2rust.com/"
repository = "https://github.com/immunant/c2rust/"
readme = "README.md"
categories = ["development-tools", "development-tools::ffi", "command-line-utilities"]
keywords = ["transpiler", "migration", "translation", "c"]
```

which are taken mostly from `c2rust/Cargo.toml`.  `description` (a required field for publishing) is omitted as that should be unique for each package.

The major benefit of this is that we standardize the fields and have a single source of truth, which is especially useful for something like `version`, which we want to update in sync.

The changes caused by this "standardization" are:
* All versions are changed to `0.16.0`, the current version.
* The `authors` are standardized on `["The C2Rust Project Developers <c2rust@immunant.com>"]`.  Previously, they would sometimes include specific developers, but it was somewhat haphazard and not updated when new people worked on them.  I think it's much simpler to standardize on this general author.
  * The exception is `c2rust-ast-printer/Cargo.toml`, which is forked from `libsyntax` and thus also credits "The Rust Project Developers".
* `homepage`, `repository`, `readme`, `categories`, `keywords` are all set the same for each package, with the exception of `readme`, which is set to `readme = "README.md"` (the local one) if it exists.

Note that `c2rust-macros` and `c2rust-refactor` are not updated at all, as they are no longer part of the workspace.

This greatly simplifies things for the upcoming `0.17` release, as it
* unifies versions for all packages intended to be published
* sets the `license` and `homepage`/`repository` fields for all new packages, which are required to publish them (`description` is also needed, but I don't think that it good to inherit; we should fix that in a separate PR)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,7 +140,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analysis-rt"
-version = "0.1.0"
+version = "0.16.0"
 dependencies = [
  "bincode",
  "enum_dispatch",
@@ -151,7 +151,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analyze"
-version = "0.1.0"
+version = "0.16.0"
 dependencies = [
  "assert_matches",
  "bitflags",
@@ -164,7 +164,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-asm-casts"
-version = "0.2.0"
+version = "0.16.0"
 
 [[package]]
 name = "c2rust-ast-builder"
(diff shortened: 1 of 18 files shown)


fw-immunant (approved):
Individual changes look OK; this should simplify future version bumps, which is good.

Issue #794, opened 2023-01-24
Handle (skip) library functions that don't take or return pointers


kkysen:
As per https://github.com/immunant/c2rust/issues/785#issuecomment-1402746651, we should handle (by skipping) any library function that doesn't take or return any pointers, as there's nothing to analyze in it.  Thus, fixing this will also fix #785, as `u32::wrapping_sub` takes and returns no pointers.

PR #807, opened 2023-01-30
(`c2rust-analyze`) Rename `Callee::{MiscBuiltin => Trivial}` and expand it to any trivial library function

kkysen:
This renames `Callee::{MiscBuiltin => Trivial}` and expands it to any trivial library function.

The meaning of `MiscBuiltin`/`Trivial` is in that it has no effect on pointer permissions in its caller and thus requires no special handling, i.e. it is trivial. Thus, `Trivial` is a much more appropriate name than `Misc`.

This also expands the definition of `Trivial` to non-builtins, including any "library"/external function that takes or returns no pointers, as they are all trivial.

This is a small change before my fix for #794.

### DIFF
--- a/c2rust-analyze/src/util.rs
+++ b/c2rust-analyze/src/util.rs
@@ -89,14 +89,17 @@ pub enum Callee<'tcx> {
         mutbl: Mutability,
     },
 
-    /// A [`Trivial`] library function that requires no special handling.
+    /// A [`Trivial`] library function is one that has no effect on pointer permissions in its caller.
+    ///
+    /// Thus, a [`Trivial`] function call requires no special handling.
     ///
     /// A function is [`Trivial`] if it has no argument or return types that are or contain a pointer.
     /// Note that "contains a pointer" is calculated recursively.
     /// There must not be any raw pointer accessible from that type.
     ///
-    /// Int-to-ptr casts a la [`std::ptr::from_exposed_addr`] are not considered for this,
-    /// as doing so is very difficult and out of scope for now.
+    /// We ignore the possibility that a function may perform
+    /// int-to-ptr casts (a la [`std::ptr::from_exposed_addr`]) internally,
+    /// as handling such casts is very difficult and out of scope for now.
     ///
     /// References are allowed, because the existence of that reference in the first place
     /// carries much stronger semantics, so in the case that the reference is casted to a raw pointer,
--- a/c2rust-analyze/src/util.rs
+++ b/c2rust-analyze/src/util.rs
@@ -89,8 +89,21 @@ pub enum Callee<'tcx> {
         mutbl: Mutability,
     },
 
-    /// A built-in or standard library function that requires no special handling.
-    MiscBuiltin,
+    /// A [`Trivial`] library function that requires no special handling.
+    ///
+    /// A function is [`Trivial`] if it has no argument or return types that are or contain a pointer.
+    /// Note that "contains a pointer" is calculated recursively.
+    /// There must not be any raw pointer accessible from that type.
+    ///
+    /// Int-to-ptr casts a la [`std::ptr::from_exposed_addr`] are not considered for this,
+    /// as doing so is very difficult and out of scope for now.
+    ///
+    /// References are allowed, because the existence of that reference in the first place
+    /// carries much stronger semantics, so in the case that the reference is casted to a raw pointer,
+    /// we can simply use the pointer permissions guaranteed by that reference.
+    ///
+    /// [`Trivial`]: Self::Trivial
+    Trivial,
 
     /// libc::malloc
     Malloc,
@@ -194,7 +207,7 @@ fn builtin_callee<'tcx>(
             if path.data[0].to_string() != "process" {
                 return None;
             }
-            Some(Callee::MiscBuiltin)
+            Some(Callee::Trivial)
         }
 
         "size_of" => {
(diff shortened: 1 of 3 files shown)


kkysen:
@spernsteiner, I want to make sure I'm getting the semantics right.  Is a function trivial (i.e. requires no special handling) if all of its argument and return types:
* are not pointers
* contain no pointers
* are not a reference
* contain no references
* has no lifetime

Which of these apply?

I'm unclear on how our analysis interacts with references (e.x. partially lifted code).

I assume at least "are not pointers" and "contain no pointers" are correct, though the latter would need to determine that recursively.  If so, should that be cached?

spernsteiner:
I think ultimately we want "trivial" to mean that a call to this function has no effect on any pointer permissions.  A reasonable approximation for now would be to take the argument and return types (after `subst_and_normalize`-ing the generic arguments into the `fn_sig`) and walk over them, including traversing the fields of any ADTs that are mentioned, and make sure they contain no raw pointers.  This ensures that there are no raw pointers accessible to both the caller and the callee (more or less - the callee might do some pointer/integer casts and pass only `usize` back and forth, but code like that is pretty rare and I think we're okay ignoring that possibility for now).

I think it's safe to ignore references (i.e. safe to allow a function that takes/returns a reference to be marked `Trivial`) because I believe Rust's definition of undefined behavior requires us to "assume the worst" about any reference' that's present in the program.  Merely creating a value `x: &mut [T]` implicitly asserts that `x[0]` through `x[x.len() - 1]` are valid, writable memory.  So any conversions of `x` to or from a raw pointer can treat `x` as having exactly the permissions implied by the `&mut [T]` type - no more, no less.  And a call `f(x)` (assuming no UB) can't possibly require more permissions of `x` beyond the permissions assigned to it upon its creation.

kkysen:
@spernsteiner, does my documentation for `Callee::Trivial` look good now based on your comment above?

Also, for [`[subst_and_normalize_erasing_regions](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.subst_and_normalize_erasing_regions)`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.subst_and_normalize_erasing_regions), where should I get the `ParamEnv` needed for that?  `tcx.param_env(did)`?

kkysen:
Only failing CI due to another `tinycbor` release hash change (see #809), totally unrelated to this PR.

c2rust-analyze/src/util.rs
```diff
@@ -89,8 +89,21 @@ pub enum Callee<'tcx> {
...
-    /// A built-in or standard library function that requires no special handling.
-    MiscBuiltin,
+    /// A [`Trivial`] library function that requires no special handling.
+    ///
+    /// A function is [`Trivial`] if it has no argument or return types that are or contain a pointer.
+    /// Note that "contains a pointer" is calculated recursively.
+    /// There must not be any raw pointer accessible from that type.
+    ///
+    /// Int-to-ptr casts a la [`std::ptr::from_exposed_addr`] are not considered for this,
+    /// as doing so is very difficult and out of scope for now.
```
spernsteiner, line 99:
```suggestion
    /// We ignore the possibility that a function taking `usize` might perform int-to-ptr casts internally,
    /// as handling such casts is very difficult and out of scope for now.
```

c2rust-analyze/src/util.rs
```diff
@@ -89,8 +89,21 @@ pub enum Callee<'tcx> {
         mutbl: Mutability,
     },
 
-    /// A built-in or standard library function that requires no special handling.
-    MiscBuiltin,
+    /// A [`Trivial`] library function that requires no special handling.
```
spernsteiner, line 92:
```suggestion
    /// A [`Trivial`] library function is one that has no effect on pointer permissions in its caller.
```

Since the current implementation (looking for pointers in argument/return types) only approximates the actual `Trivial` property, I think the precise definition should be included here in the comment.

spernsteiner:
>Also, for [subst_and_normalize_erasing_regions](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.subst_and_normalize_erasing_regions), where should I get the `ParamEnv` needed for that? [tcx.param_env(did)](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.param_env)?

The major alternative is `ParamEnv::reveal_all()`, but looking at the docs of the two methods, I think `tcx.param_env(did)` is more appropriate here.

spernsteiner (approved):

c2rust-analyze/src/util.rs
kkysen: That makes sense, I agree.

kkysen: In fd48ed87242a603bced61135a612d57c182b9531, I made this change but kept the "requires no special handling" as

```rust
/// Thus, a [`Trivial`] function call requires no special handling.
```

c2rust-analyze/src/util.rs
kkysen: In fd48ed87242a603bced61135a612d57c182b9531, I made this change except for the "taking `usize`" part, as it doesn't have to be a `usize`, it could be any integer, or it could just be fabricated from nothing.

PR #808, opened 2023-01-30
c2rust-ast-exporter: update tinycbor

fw-immunant:
Somehow the md5sum of the tinycbor v0.5.3 `.tar.gz` archive from github has changed (but the decompressed .tar is identical to before, md5sum `206885d49c184edfed363b4f7d3227f3`). We need to fix the CI breakage from the hash changing, but while we're at it, we might as well just update the library to v0.6.0. We updated our Rust deps in prep for release, so this is a good time to go ahead with tinycbor too.

### DIFF
--- a/c2rust-ast-exporter/src/CMakeLists.txt
+++ b/c2rust-ast-exporter/src/CMakeLists.txt
@@ -4,9 +4,9 @@ project(ASTExporter)
 #################################################
 # TinyCBOR                                      #
 #################################################
-set(TINYCBOR_URL "https://github.com/intel/tinycbor/archive/v0.5.3.tar.gz"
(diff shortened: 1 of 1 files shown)


kkysen (approved):

kkysen:
It's changed again, to `3663e683dbf03f49cb7057ed316a7563` now.

fw-immunant:
Yeah, it turns out GitHub reverted their change as everyone in the world expects release archives to be stable: https://news.ycombinator.com/item?id=34586917

A better fix does seem worthwhile.

kkysen:
Found the culprit: https://github.com/git/git/commit/4f4be00d302bc52d0d9d5a3d4738bb525066c710.

PR #809, opened 2023-01-31
Switch `tinycbor` download to use `git` with a stable tag and commit hash vs. the unstable compressed release hash

kkysen:
CI keeps breaking due GitHub changing the compression on release tarballs, which results in different checksum hashes, which we use for integrity checks for the `tinycbor` dependency.  This has happened twice in the last day now (first fixed with #808, which just updated the hash).  Since it keeps happening (and who knows how many more times), I think it's better to switch to this approach (which I had implemented a while ago but never merged), which uses `cmake`'s `git` support to clone the `tinycbor` `git` repository at a certain release tag and also provide a commit hash for integrity checks.  This has the benefits of:
- using a source-based (commit) hash rather than a binary-based (`.tar.gz`) hash
    - this is really the main benefit, as it should be completely stable
- using SHA-1 instead of MD5
- being a bit easier to upgrade versions, since we just change the tag and its commit hash

This also requires adding `git` as a dependency in the Docker containers.  It's already a dependency in the `apt`-based ones, so this also adds it for the others.  This is also incidentally useful for local testing, where having `git` is useful.  This needs a Docker push, so I'll do that once this is approved.

### DIFF
--- a/c2rust-ast-exporter/src/CMakeLists.txt
+++ b/c2rust-ast-exporter/src/CMakeLists.txt
@@ -4,17 +4,22 @@ project(ASTExporter)
 #################################################
 # TinyCBOR                                      #
 #################################################
-set(TINYCBOR_URL "https://github.com/intel/tinycbor/archive/v0.6.0.tar.gz"
-    CACHE STRING "tinycbor download URL")
-set(TINYCBOR_MD5 "3663e683dbf03f49cb7057ed316a7563" CACHE STRING "tinycbor archive md5 sum")
+
+set(TINYCBOR_REPO "https://github.com/intel/tinycbor.git" CACHE STRING "tinycbor git repo URL")
+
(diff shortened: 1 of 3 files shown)


fw-immunant (approved):
Changes look good, but Fedora failed CI. Maybe images haven't deployed? This should merge when CI goes green; in the meantime we can just bump our tinycbor hash from GH again--they seem to have [learned their lesson](https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/) and aren't likely to modify archives again soon.

kkysen:
> Changes look good, but Fedora failed CI. Maybe images haven't deployed? This should merge when CI goes green; in the meantime we can just bump our tinycbor hash from GH again--they seem to have [learned their lesson](https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/) and aren't likely to modify archives again soon.

I'll push new Docker containers now (which should make CI pass on Fedora) and then merge this.

Issue #803, opened 2023-01-28
Problematic macro translation with --translate-const-macros


Remmirad:
I believe there exists a Problem with the `--translate-const-macros` option. A macro `#define TEST foo()` gets translated to `pub const TEST: t = foo();`. 

This does not compile because `foo()` is a C function and therefore an unsafe binding but no `unsafe` Block is generated. In addition `foo` would need to be compiletime const. Which as far as I know is not possible for bindings? And even if that would be possible c2rust would need to be able to check if it really is compiletime const, which it cant if only provided the functions definition.

And last the achieved behavior in Rust differs from the behavior in C. 
I think it would be ok to just completely forbid macro translation into a `const` if the macro expression is a function call? 

The only scenario where this would be a bit problematic, is when there exists a function `foo` and c2rust has access to the implementation of the function and is able to correctly translate it into Rust and check that it can be evaluated at compile time. Than the code could compile with the unsafe block added. But it would still not represent the C behavior as far as I can see. So maybe it is not really a problem.

## Test :
test.h
```C
int foo();

#define GET_FOO (foo())

int use_get_foo() {
    return GET_FOO;
}
```
transpile flags: 
```cli
--translate-const-macros
```

fw-immunant:
Thanks for the bug report--I can confirm the problematic behavior. The `--translate-const-macros` flag is pretty experimental and it's not surprising that it breaks here.

This issue is subtler than it may appear because the real constraint here is on the Rust side--we cannot generate a `const` whose body calls non-`const` functions. It isn't enough to look for function calls in the C source of the macro definition, though, because c2rust converts some C function calls (e.g. builtins) into Rust expressions that consist only of builtin operators and `const` function calls, and which we can translate. For example, `#define FOO __builtin_clz(34)` becomes `pub const FOO: std::ffi::c_int = (34 as std::ffi::c_int as std::ffi::c_uint).leading_zeros() as i32;` which compiles just fine.

We could, then, say "is the Rust code we generate for this macro definition `const`-safe?" as the predicate to use when deciding if we can translate a macro. This seems like the right way to capture the constraint that's actually relevant here. Unfortunately, c2rust rather naïvely builds an AST of the code to output and doesn't have a notion of which functions are const/not in that AST. It would be significant engineering effort to implement this notion, and as `--translate-const-macros` is already an experimental, somewhat hacky, approach, I think we probably should leave it as-is rather than complicating how it's broken without knowing it's a path to a fully reliable feature.

[...]

PR #810, opened 2023-01-31
Prohibit function call macro to const translation

Remmirad:
closes https://github.com/immunant/c2rust/issues/803

Tackles the issue, that C macros consisting of a function call are translated to rust `const` expressions which do not compile and are not semantically equivalent.  

As described in the code comment there might be a more thorough way to do this but that would, as far as i can tell, at the least require some redesigning of how C2rust handles/represents sideeffects and only would increase performance a bit in a special case, as described in the issue. So for now this here might suffice. 

The fix I wrote here stops C2Rust from producing code that does not compile in this case and achieves the behavior that one would expect: All occurrences of the macro in code are replaced by the function call. 

I also added corresponding test for this situation.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2231,6 +2231,19 @@ impl<'c> Translation<'c> {
                     .unwrap_or((*id, ty));
                 let expr = self.convert_expr(ctx, expr_id)?;
 
+                // Issue: https://github.com/immunant/c2rust/issues/803
+                // This actually should be catched by `to_unsafe_pure_expr` see below
+                // but a function call expression is translated as having no sideeffects by `convert_expr` (above)
+                // which is not necessarily correct, we just might not know them (e.g only decl in header exists).
+                // So we catch this here until there is a way to express unknown sideeffects of an expression. 
+                let expr = match expr.result_map(|v| {
+                    if matches!(*v, Expr::Call(_)) {Err(())}
+                    else {Ok(v)}
+                }) {
+                    Ok(v) => v,
+                    Err(_) => return Err(format_err!("Can not expand macro to function call"))
+                };
+
                 // Join ty and cur_ty to the smaller of the two types. If the
                 // types are not cast-compatible, abort the fold.
                 let ty_kind = self.ast_context.resolve_type(ty).kind.clone();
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -2231,6 +2231,22 @@ impl<'c> Translation<'c> {
...
+                let expr = match expr.result_map(|v| {
+                    if matches!(*v, Expr::Call(_)) {
+                        Err(())
+                    } else {
+                        Ok(v)
+                    }
+                }) {
+                    Ok(v) => v,
+                    Err(_) => return Err(format_err!("Can not expand macro to function call")),
+                };
```
fw-immunant, line 2248: This can be written more simply:

```suggestion
                let expr = expr.result_map(|v| {
                    if matches!(*v, Expr::Call(_)) {
                        Err(format_err!("Macro definition contains function call"))
                    } else {
                        Ok(v)
                    }
                })?;
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2231,6 +2231,22 @@ impl<'c> Translation<'c> {
                     .unwrap_or((*id, ty));
                 let expr = self.convert_expr(ctx, expr_id)?;
 
+                // Issue: https://github.com/immunant/c2rust/issues/803
+                // This actually should be catched by `to_unsafe_pure_expr` see below
+                // but a function call expression is translated as having no sideeffects by `convert_expr` (above)
+                // which is not necessarily correct, we just might not know them (e.g only decl in header exists).
+                // So we catch this here until there is a way to express unknown sideeffects of an expression.
```
fw-immunant, line 2238: The purpose of `to_unsafe_pure_expr` is not actually to enforce the kind of purity that Rust `const`s require. It checks a notion of purity based on "does the expression computing this value have prerequisite statements that we need to ensure are evaluated prior to the expression itself".

We sadly don't have a good notion of "is this Rust expression const-safe", and it's difficult to build one because `c2rust` is oriented toward reasoning about C code, not Rust code. See [this comment](https://github.com/immunant/c2rust/issues/803#issuecomment-1544754766).

fw-immunant (changes requested):
Thanks for the PR.

Unfortunately, I don't think this approach is what we want here. Forbidding call expressions on the Rust side is too restrictive (due to `const fn`s), and unfortunately so is forbidding call expressions on the C side (due to builtins which we may translate to Rust builtins or `const fn`s).

As this feature is experimental, I think it's best to simply leave it uncomplicated for now. `--translate-const-macros` is only useful for the simplest cases, and we don't have a clear path toward making it useful everywhere.

PR #811, opened 2023-02-01
Update `rustc` version incrementally to `nightly-2022-12-06`

kkysen:
This incrementally updates our `rustc` version to `nightly-2022-12-06` from `nightly-2022-08-08`.

I stopped there as the next nightly causes a panic in `c2rust-analyze` whose fix will probably be a bit more involved.

This helps us get closer back to the latest nightly so we don't stay too far behind, which makes updating later much harder.  

And as a result, the published [`rustc` nightly docs](https://doc.rust-lang.org/nightly/nightly-rustc/) become increasingly accurate as we near the latest nightly.  Docs are only published for the latest nightly, not the old versions we're using, and `rustup doc` doesn't seem to contain them anywhere (it contains the `rustc` book docs, but not the `rustdoc` docs of `rustc`'s code itself).  Docs could be built directly from [rust-lang/rust](https://github.com/rust-lang/rust), but that's a major pain as the build system is not trivial.

The changes should be a lot easier to see/review by commit rather than all at once.

The commits of meaningful changes are:
* 74f703fcc7a865cc27f83539eaa2a5afaff84a6a
* de457a95064d93b9c6e5448e777fc938121103ee
* 4a740ab2f3db148e11e5a5ae92fecf2c5f6669e7
* 76be3078892d5d3fd7c076568d447339070c265d
* 674d0e20458ca26b3d32bdb9e3e4bdc37bc29ae8
* 7fd06e33496b8e3817eb37a61f80e68042ab8b30

Of these, really only these are significant changes/fixes:
* 4a740ab2f3db148e11e5a5ae92fecf2c5f6669e7

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-08-29"
+channel = "nightly-2022-08-30"
 components = ["rustfmt-preview", "rustc-dev", "rust-src", "miri"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-09-17"
+channel = "nightly-2022-09-19"
 components = ["rustfmt-preview", "rustc-dev", "rust-src", "miri"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-10-08"
+channel = "nightly-2022-10-09"
 components = ["rustfmt-preview", "rustc-dev", "rust-src", "miri"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-10-09"
+channel = "nightly-2022-12-06"
 components = ["rustfmt-preview", "rustc-dev", "rust-src", "miri"]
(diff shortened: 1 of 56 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
```
ahomescu, line 1: Also commented on Slack: could we split this file into one per package? That way we can update some of them independently. E.g. I think `c2rust-analyze` and `c2rust-pdg` could be on stable Rust.

kkysen: `c2rust-analyze` and `c2rust-pdg` use rustc internals, so they need nightly.

kkysen: For `c2rust-transpile` and it's dependencies, those can build on stable, but `c2rust`, if it depends on things like `c2rust-analyze` (it doesn't currently, but that was the future intention), I'm not sure how to have `rust-toolchsin.toml` work there, as we want it to build on both nightly and stable. 

This is also an issue for `rav1d`, so figuring out how to do this would be nice.

kkysen:
This still isn't a priority right now (#1115 fixed the issue, though ideally that removal is temporary), but we do eventually want to do this, and I'd like to avoid situations like #1114 where someone ends up redoing work that was already done (and just needs a rebase), so I'll reopen this and just let it sit there.

GPHemsley:
The latest version not requiring any changes is now `nightly-2022-08-15`, after #942 and rust-lang/rust#100441.

Might be worth bumping just to that to avoid any further backsliding.

kkysen:
> The latest version not requiring any changes is now `nightly-2022-08-15`, after #942 and [rust-lang/rust#100441](https://github.com/rust-lang/rust/pull/100441).
> 
> Might be worth bumping just to that to avoid any further backsliding.

Thanks!  Good to know!

PR #815, opened 2023-02-07
analyze: implement source code rewriting

spernsteiner:
This extends `c2rust-analyze` with the ability to generate actual rewritten source code.  For example, when run on `insertion_sort.rs`, it produces output like this:
```Rust
#[no_mangle]
pub unsafe extern "C" fn insertion_sort(n: libc::c_int, p: &mut [(libc::c_int)]) {
    let mut i: libc::c_int = 1 as libc::c_int;
    while i < n {
        let tmp: libc::c_int = *&(&(&*(p))[((i as isize) as usize) ..])[0];
        let mut j: libc::c_int = i;
        while j > 0 as libc::c_int && *&(&(&*(p))[(((j - 1 as libc::c_int) as isize) as usize) ..])[0] > tmp {
            *&mut (&mut (p)[((j as isize) as usize) ..])[0] = *&(&(&*(p))[(((j - 1 as libc::c_int) as isize) as usize) ..])[0];
            j -= 1
        }
        *&mut (&mut (p)[((j as isize) as usize) ..])[0] = tmp;
        i += 1
    }
}
```
Which is ugly, but completely safe (after removing `#[no_mangle]` and `unsafe` from the function declaration).

The overall implementation strategy is:

1. Using the pointer permissions and flags inferred by the analysis, annotate MIR statements with the desired rewrites.  These MIR-level rewrites are abstract changes to MIR statements, such as adding a cast to a particular assignment statement.  This is defined in the `rewrite::expr::mir_op` module.
2. For each HIR expression, look at the MIR statements generated from this HIR expression and lift any MIR rewrites into HIR rewrites.  HIR rewrites are expressed as concrete operations on source code, such as replacing an expression with one of its subexpressions (both identified by their `Span`s) or wrapping an expression in a ref or deref operation.  The HIR-level rewrite type is `rewrite::Rewrite`; the `rewrite::expr::hir_op` module implements the lifting.
3. Apply the rewrites to the source code of the input program.  This reads the source of each file and emits a new string consisting of the file source with certain `Span`s rewritten as specified by the HIR rewrites.  The code for this is in `rewrite::apply`.

This covers rewriting of expressions; rewriting of types is similar but mostly skips step 1, since an abstract description of the changes to be made can be obtained by inspecting the pointer permissions and flags directly.  This code is in `rewrite::ty`.  All type and expr rewrites are collected and applied in one pass in step 3 (as rewriting in two passes would require us to update the `Span`s mentioned in the later rewrites to account for the changes in the source code produced by the earlier ones).

There are two additional bits of complexity:

* In expr rewriting, sometimes the code we're modifying has had autoderef and/or autoref `Adjustment`s applied to it.  To avoid unexpectedly changing which adjustments get applied, we "materialize" the `Adjustment`s, making them explicit in the source code.  For example, `vec.len()`, which implicitly applies deref and ref adjustments to `vec`, would be converted to `(&*vec).len()`, where the deref and ref operations are explicit, and might be further rewritten from there.  However, we don't want to materialize all adjustments, as this would make even non-rewritten code extremely verbose, so we try to materialize adjustments only on code that's subject to some rewrite.
* In type rewriting, sometimes we need to rewrite a pointer type that's behind a type alias, such as `type MyPtr = *mut u8;`.  To do this, we unfold the alias, replacing `x: MyPtr` with `x: *mut u8`, and applying further rewrites from there.  Again, we try to apply this transformation selectively, since we don't want to unfold all type aliases in the program.

There are a few TODOs remaining in the code, and in general we'll need to extend this to support some analysis features we've added more recently (like struct fields), but as this rewriting implementation works for `insertion_sort`, I think it's in good enough shape for review.


kkysen:
Could you add most of the above explanation, which is great, as module-level documentation in `rewrites/mod.rs`?

c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// `arr[idx..]`
+    SliceTail(Box<Rewrite>, Box<Rewrite>),
+    /// `e as usize`
+    CastUsize(Box<Rewrite>),
+    /// The integer literal `0`.
+    LitZero,
+
+    // Type builders
+    /// Emit a complete pretty-printed type, discarding the original annotation.
+    PrintTy(String),
```
aneksteind, line 40: why are expression rewrites and type rewrites part of the same enum?

c2rust-analyze/src/rewrite/span_index.rs
```diff
@@ -0,0 +1,54 @@
...
+        let mut v = it.into_iter().collect::<Vec<_>>();
+        v.sort_by_key(|(span, _)| span.lo());
+        SpanIndex { v }
+    }
+
+    /// Iterate over items whose spans are entirely contained within `span`.  The order of the
+    /// returned items is unspecified.
+    pub fn lookup<'a>(&'a self, span: Span) -> RangeIter<'a, T> {
+        let data = span.data();
+        let start = self.v.partition_point(|(span, _)| span.lo() < data.lo);
```
aneksteind, line 23: shouldn't the input param to `lookup`, ,`span`, have `span.lo()` available? you'd need to change the `span` in the `partition_point` closure to another name. I'm unsure why `data = span.data()` is necessary

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+use rustc_middle::ty::subst::GenericArg;
+use rustc_middle::ty::{self, ReErased, Ty, TyCtxt, TyKind};
+use rustc_span::Span;
+
+/// A label for use with `LabeledTy` to indicate what rewrites to apply at each position in a type.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
+struct RewriteLabel {
+    /// Rewrite a raw pointer, whose ownership and quantity have been inferred as indicated.
+    ty_desc: Option<(Ownership, Quantity)>,
+    /// If set, a child or other descendant of this type requires rewriting.
```
aneksteind, line 23: what is a child or descendant of a type? including that in the docs somehow (maybe a rustdoc link) would be helpful

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        } else {
+            let perms = asn.perms()[lty.label];
+            let flags = asn.flags()[lty.label];
+            // TODO: if the `Ownership` and `Quantity` exactly match `lty.ty`, then `ty_desc` can
+            // be `None` (no rewriting required).  This might let us avoid inlining a type alias
+            // for some pointers where no actual improvement was possible.
+            Some(type_desc::perms_to_desc(perms, flags))
+        };
+        let descendant_has_rewrite = args.iter().any(|child| {
+            let has_rewrite = child.label.ty_desc.is_some();
```
aneksteind, line 48: nit: why not `child.label.ty_desc.is_some() || child.label.descendant_has_rewrite`?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+    })
+}
+
+/// Extract arguments from `hir_ty` if it corresponds to the tcx type `ty`.  If the two types
+/// differ in shape (for example, if `hir_ty` is `_` or a type alias), then the result is `None`.
+/// If they have the same shape, the result is a list of `hir::Ty`s of arguments, where "argument"
+/// is defined the same way as for the `args` field of `LabeledTy`.
+fn deconstruct_hir_ty<'a, 'tcx>(
+    ty: Ty<'tcx>,
```
aneksteind, line 63: is this an MIR ty?

aneksteind: if so, I think it could help to use the qualified name or to make it clearer somehow. this might be something those who work with the compiler types often take for granted...?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        .cloned()
+        .find(|&krate| tcx.crate_name(krate).as_str() == "core")
+        .expect("failed to find crate `core`");
+
+    let cell_mod_child = tcx
+        .module_children(core_crate.as_def_id())
+        .iter()
+        .find(|child| child.ident.as_str() == "cell")
+        .expect("failed to find module `core::cell`");
+    let cell_mod = match cell_mod_child.res {
```
aneksteind, line 173: nit: rename to `cell_mod_did`

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        Res::Def(DefKind::Mod, did) => did,
+        ref r => panic!("unexpected resolution {:?} for `core::cell`", r),
+    };
+
+    let cell_struct_child = tcx
+        .module_children(cell_mod)
+        .iter()
+        .find(|child| child.ident.as_str() == "Cell")
+        .expect("failed to find struct `core::cell::Cell`");
+    let cell_struct = match cell_struct_child.res {
```
aneksteind, line 183: nit: rename to `cell_struct_did`

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+                let ty = mk_rewritten_ty(self.rw_lcx, rw_lty);
+                let printer = FmtPrinter::new(*self.rw_lcx, Namespace::TypeNS);
+                let s = ty.print(printer).unwrap().into_buffer();
+                self.hir_rewrites.push((hir_ty.span, Rewrite::PrintTy(s)));
+                return;
+            }
+        };
+
+        if let Some((own, qty)) = rw_lty.label.ty_desc {
+            assert_eq!(hir_args.len(), 1);
```
aneksteind, line 261: will this assertion break when we want to introduce ADT handling?

aneksteind: also, what about tuples?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+            if own == Ownership::Cell {
+                rw = Rewrite::TyCtor("core::cell::Cell".into(), vec![rw]);
+            }
+
+            rw = match qty {
+                Quantity::Single => rw,
+                Quantity::Slice => Rewrite::TySlice(Box::new(rw)),
+                // TODO: This should generate `OffsetPtr<T>` rather than `&[T]`, but `OffsetPtr` is
+                // NYI
+                Quantity::OffsetPtr => Rewrite::TySlice(Box::new(rw)),
```
aneksteind, line 273: what affect does this have on the insertion sort rewriting?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+            };
+
+            rw = match own {
+                Ownership::Raw => Rewrite::TyPtr(Box::new(rw), Mutability::Not),
+                Ownership::RawMut => Rewrite::TyPtr(Box::new(rw), Mutability::Mut),
+                Ownership::Imm => Rewrite::TyRef(Box::new(rw), Mutability::Not),
+                Ownership::Cell => Rewrite::TyRef(Box::new(rw), Mutability::Not),
```
aneksteind, line 280: am I interpreting this as "if something with cell ownership is immutable, rewrite it as an immutable reference" correct?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+
+impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
+    type NestedFilter = nested_filter::OnlyBodies;
+
+    fn nested_visit_map(&mut self) -> Self::Map {
+        self.acx.tcx().hir()
+    }
+}
+
+pub fn gen_ty_rewrites<'tcx>(
```
aneksteind, line 305: is this only for function signatures and not function bodies?

c2rust-analyze/tests/filecheck/field_temp.rs
```diff
@@ -0,0 +1,20 @@
...
+    // The temporary `(*l).data` requires a MIR `MutToImm` rewrite.
+    // CHECK: ([[#@LINE+2]]: (*l).data): &[i32]
+    // CHECK: rewrite Ref(Deref(Identity), Not) at {{.*}}:[[#@LINE+1]]:6: [[#@LINE+1]]:15
+    *(*l).data.offset(i as isize)
+}
+
+// CHECK-LABEL: final labeling for "list_set"
+pub unsafe fn list_set(l: *const MyList, i: usize, x: i32) {
+    // CHECK: ([[#@LINE+1]]: (*l).data): &mut [i32]
+    *(*l).data.offset(i as isize) = x;
```
aneksteind, line 19: add `// CHECK: rewrite Ref(Deref(Identity), Mut) at {{.*}}:[[#@LINE+1]]:6: [[#@LINE+1]]:15`?

c2rust-analyze/tests/filecheck/fields.rs
```diff
@@ -52,8 +52,10 @@ struct VecTup<'a> {
 // CHECK-DAG: assign Label { origin: Some(Origin([[P_REF_A_ORIGIN]]))
 
 // CHECK-LABEL: final labeling for "_field_access"
-// CHECK-DAG: ([[@LINE+3]]: ppd): addr_of = UNIQUE, type = READ | WRITE | UNIQUE
-// CHECK-DAG: ([[@LINE+2]]: ra): &mut A
+// CHECK-DAG: ([[@LINE+5]]: ppd): addr_of = UNIQUE, type = READ | WRITE | UNIQUE
```
aneksteind, line 55: just to be clear, it's rewrites of the fields that are currently not supported by this PR, but the ADT type itself should be covered?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+
+impl SpanLike for (u32, u32) {
+    fn lo(self) -> BytePos {
+        BytePos(self.0)
+    }
+    fn hi(self) -> BytePos {
+        BytePos(self.1)
+    }
+    fn contains(self, other: Self) -> bool {
+        self.0 <= other.0 && other.1 <= self.1
```
aneksteind, line 60: i think this and `overlaps` would be clearer if put in terms of `lo` and `hi`

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+impl<S: SpanLike> RewriteTree<S> {
+    pub fn build(
+        mut rws: Vec<(S, Rewrite<S>)>,
+    ) -> (Vec<RewriteTree<S>>, Vec<(S, Rewrite<S>, RewriteError<S>)>) {
+        // Sort by start position and then by decreasing length, so that each parent span comes
+        // before all its children.
+        rws.sort_by_key(|&(ref s, _)| (s.lo(), Reverse(s.hi() - s.lo())));
+
+        // The stack contains a `RewriteTree` node for the most recently processed item from `rws`
+        // along with all of its ancestors (nodes whose spans strictly contain its span).
```
aneksteind, line 76: i thought the parent's span completely contained the child's span? to me, this comment makes it seem like it's the other way around. which is correct?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+            stack: &mut Vec<RewriteTree<S>>,
+            out: &mut Vec<RewriteTree<S>>,
+            rt: RewriteTree<S>,
+        ) {
+            if let Some(parent) = stack.last_mut() {
+                debug_assert!(parent.span.contains(rt.span));
+                debug_assert!(parent
+                    .children
+                    .last()
+                    .map_or(true, |ch| ch.span.hi() <= rt.span.lo()),);
```
aneksteind, line 95: why is this not `ch.span.hi() <= rt.span.hi()`? could you explain the semantic property you're looking to assert here?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+    ) -> (Vec<RewriteTree<S>>, Vec<(S, Rewrite<S>, RewriteError<S>)>) {
+        // Sort by start position and then by decreasing length, so that each parent span comes
+        // before all its children.
+        rws.sort_by_key(|&(ref s, _)| (s.lo(), Reverse(s.hi() - s.lo())));
+
+        // The stack contains a `RewriteTree` node for the most recently processed item from `rws`
+        // along with all of its ancestors (nodes whose spans strictly contain its span).
+        // `stack[0]` is the outermost ancestor, and `stack.last()` is the most recently pushed
+        // node.  Nodes are "committed" once they are known to have no more children; when
+        // committed, the node is moved into the `children` list of its parent or (if it has no
```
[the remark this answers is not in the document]
aneksteind: nevermind, i understand the phrasing now

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+                    errs.push((span, rw, RewriteError::Conflict));
+                }
+                continue;
+            }
+
+            // Commit all rewrites that come strictly before `span`.  This leaves only nodes that
+            // at least partially overlap the current item.  These should normally be ancestors of
+            // the current item on the stack, but it's also possible that the current item
+            // erroneously partially overlaps a previous item.
+            while stack.last().map_or(false, |rt| rt.span.hi() <= lo) {
```
aneksteind, line 125: i think `ch.hi() <= parent.lo()` can be factored out into another method a la `contains` or `overlaps`, and would help make things clearer

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+
+            // Push a new node onto the stack.
+            stack.push(RewriteTree {
+                span,
+                rw,
+                children: Vec::new(),
+            });
+        }
+
+        while let Some(rt) = stack.pop() {
```
aneksteind, line 168: i think this deserves a comment similar to above where you explain that you're committing all rewrites that come before a particular span

aneksteind:
i think this PR should have been multiple, smaller, PRs. i won't formally request changes but would strongly encourage this to be broken up. there is a lot here for someone new to these concepts or portion of the code base to grok all at once, and staging the changes would have made them more digestible. granted, the commits within this PR are nicely organized.

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+    rts: &'a [RewriteTree],
+    span: Span,
+) -> (&'a [RewriteTree], &'a [RewriteTree], &'a [RewriteTree]) {
+    let lo = span.lo();
+    let hi = span.hi();
+
+    // Collect nodes from the front of `rts` until we find one that overlaps or comes after `span`.
+    let i = rts
+        .iter()
+        .position(|rt| rt.span.hi() > lo)
```
aneksteind, line 189: intuitively i'd expect the first overlapping span to have the property `rt.span.lo() <= hi`. what am i missing?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+        f(self);
+        if cond {
+            self.emit_str(")");
+        }
+    }
+
+    fn emit_rewrite(
+        &mut self,
+        rw: &Rewrite,
+        prec: usize,
```
aneksteind, line 226: could you document `prec`?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+                slf.emit_rewrite(idx, 0, emit_expr, emit_subexpr);
+                slf.emit_str("]");
+            }),
+            Rewrite::SliceTail(ref arr, ref idx) => self.emit_parenthesized(prec > 3, |slf| {
+                slf.emit_rewrite(arr, 3, emit_expr, emit_subexpr);
+                slf.emit_str("[");
+                // Rather than figure out the right precedence for `..`, just force
+                // parenthesization in this position.
+                slf.emit_rewrite(idx, 999, emit_expr, emit_subexpr);
+                slf.emit_str(" ..]");
```
aneksteind, line 270: is it valid (or even possible) to emit parenthesized expressions within things like this?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+        let (file_rts, rest) = rts.split_at(idx);
+        rts = rest;
+
+        let mut buf = String::new();
+        let mut emit = |s: &str| {
+            buf.push_str(s);
+        };
+        let mut emitter = Emitter {
+            file: &file,
+            emit: &mut emit,
```
aneksteind, line 384: why does `Emitter` have an API that takes an `emit` function as opposed to pushing to an internal string buffer? Do we expect to pass different kinds of `emit` functions?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+    new_src
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    type FakeSpan = (u32, u32);
+
+    fn mk(start: u32, end: u32, i: usize) -> (FakeSpan, Rewrite<FakeSpan>) {
```
aneksteind, line 401: are `start` and `end` equivalent to `lo` and `hi`? if so, i think it'd be clearer to having naming consistency

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+        Rewrite::Sub(i, (0, 0))
+    }
+
+    fn mk_rt(
+        start: u32,
+        end: u32,
+        i: usize,
+        children: Vec<RewriteTree<FakeSpan>>,
+    ) -> RewriteTree<FakeSpan> {
+        let span = (start, end);
```
aneksteind, line 417: use `mk(start, end, i)`?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+        let rw = mk_rewrite(i);
+        RewriteTree { span, rw, children }
+    }
+
+    #[test]
+    fn rewrite_tree_nesting() {
```
aneksteind, line 423: is this testing the sorting invariant? might be worth adding a comment

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+                5,
+                2,
+                vec![mk_rt(1, 2, 0, vec![]), mk_rt(3, 4, 1, vec![]),]
+            ),]
+        );
+    }
+
+    #[test]
+    fn rewrite_tree_identical() {
+        let (rts, errs) = RewriteTree::build(vec![
```
aneksteind, line 439: sorting + nub? I think a comment would help

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+                2,
+                vec![mk_rt(1, 2, 0, vec![]), mk_rt(3, 4, 1, vec![]),]
+            ),]
+        );
+    }
+
+    #[test]
+    fn rewrite_tree_nonidentical() {
+        let (rts, errs) = RewriteTree::build(vec![
+            mk(1, 2, 0),
```
aneksteind, line 462: since you're not explicitly testing sorting and nubbing here, i think it'd be helpful to group these by conflict, i.e. `0,5`, `1,2`, and `3,4` and consider annotating that each of these groups conflict.

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,506 @@
...
+        );
+    }
+
+    #[test]
+    fn rewrite_tree_overlap() {
+        let (rts, errs) =
+            RewriteTree::build(vec![mk(1, 3, 0), mk(3, 4, 1), mk(0, 5, 2), mk(2, 4, 3)]);
+        assert_eq!(
+            errs,
+            vec![((2, 4), mk_rewrite(3), RewriteError::PartialOverlap((1, 3))),]
```
aneksteind, line 494: can you call `mk(2,4,3)` rather than `((2, 4), mk_rewrite(3)`?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -0,0 +1,392 @@
...
+                    mir_op::RewriteKind::SliceFirst { mutbl } => {
+                        // `p` -> `&p[0]`
+                        let arr = hir_rw;
+                        let idx = Rewrite::LitZero;
+                        let elem = Rewrite::Index(Box::new(arr), Box::new(idx));
+                        hir_rw = Rewrite::Ref(Box::new(elem), mutbl_from_bool(mutbl));
+                    }
+
+                    mir_op::RewriteKind::MutToImm => {
+                        // `p` -> `&*p`
```
aneksteind, line 253: this is helpful -- can you place similar comments in other match arms?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -0,0 +1,392 @@
...
+        if self.materialize_adjustments {
+            let adjusts = self.typeck_results.expr_adjustments(ex);
+            assert!(
+                adjusts.is_empty() || !applied_mir_rewrite,
+                "combining adjustments ({:?}) with rewrite ({:?}) is NYI",
+                adjusts,
+                hir_rw,
+            );
+            for adj in adjusts {
+                match adj.kind {
```
aneksteind, line 283: i think this block should be refactored into its own function

c2rust-analyze/src/rewrite/apply.rs
kkysen: In a `..` range?  I'm pretty sure.

c2rust-analyze/src/rewrite/apply.rs
kkysen: Why do we need precedence?  Can't we just parenthesize everything and let another tool remove the unnecessary ones?  We already emit a bunch of unnecessary operations.

c2rust-analyze/src/rewrite/apply.rs
kkysen: I think the flexibility is fine here, but I would use `std::io::Write`.  That's much more standardized, flexible (as it has many other implementors), and you can use the `write!` macro.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        .expect("failed to find struct `core::cell::Cell`");
+    let cell_struct = match cell_struct_child.res {
+        Res::Def(DefKind::Struct, did) => did,
+        ref r => panic!("unexpected resolution {:?} for `core::cell::Cell`", r),
+    };
+
+    let cell_adt = tcx.adt_def(cell_struct);
+    let substs = tcx.mk_substs([GenericArg::from(ty)].into_iter());
+    tcx.mk_adt(cell_adt, substs)
+}
```
kkysen, line 191: I think the proper (and much easier and faster) way to do this is using diagnostic items.  We can call `tcx.get_diagnostic_item(Symbol::intern(...))` and get the `DefId` directly.  Only `Cell` doesn't seem to have one, which I'm surprised, since I would've thought it'd be used in some diagnostic.  `RefCell` does, though: https://github.com/rust-lang/rust/blob/068161ea483b1a80a959476cb3e31e6619a72737/library/core/src/cell.rs#L628-L630.  So you can at least skip most of this function to start at `RefCell`.

kkysen: Oh, that was added on 2022-09-27, and we're on 2022-08-08.  We can similarly use `tcx.lang_items().unsafe_cell_type()`.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        let mut ty = args[0];
+
+        if own == Ownership::Cell {
+            ty = mk_cell(tcx, ty);
+        }
+
+        ty = match qty {
+            Quantity::Single => ty,
+            Quantity::Slice => tcx.mk_slice(ty),
+            // TODO: This should generate `OffsetPtr<T>` rather than `&[T]`, but `OffsetPtr` is NYI
```
kkysen, line 212: What's NYI?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+            Quantity::OffsetPtr => tcx.mk_slice(ty),
+        };
+
+        ty = match own {
+            Ownership::Raw => tcx.mk_imm_ptr(ty),
+            Ownership::RawMut => tcx.mk_mut_ptr(ty),
+            Ownership::Imm => tcx.mk_imm_ref(tcx.mk_region(ReErased), ty),
+            Ownership::Cell => tcx.mk_imm_ref(tcx.mk_region(ReErased), ty),
+            Ownership::Mut => tcx.mk_mut_ref(tcx.mk_region(ReErased), ty),
+            Ownership::Rc => todo!(),
```
kkysen, line 222: There's a `tcx.mk_box`.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+            Ownership::Box => todo!(),
```
kkysen, line 223: There's a `tcx.mk_box`.

c2rust-analyze/src/rewrite/ty.rs
kkysen: What's `OffsetPtr`?  What is this supposed to generate?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+        };
+        let prim = match path.res {
+            Res::PrimTy(prim) => prim,
+            _ => return None,
+        };
+        if prim != expect_prim {
+            return None;
+        }
+        Some(Vec::new())
+    }
```
kkysen, line 84: Why not return `bool` or `Option<()>`?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -0,0 +1,392 @@
...
+                match rw.kind {
+                    mir_op::RewriteKind::OffsetSlice { mutbl } => {
+                        assert!(matches!(hir_rw, Rewrite::Identity));
+                        //assert_eq!(num_args, 2);
+                        // `p.offset(i)` -> `&p[i as usize ..]`
+                        let arr = self.get_subexpr(ex, 0);
+                        let idx = Rewrite::CastUsize(Box::new(self.get_subexpr(ex, 1)));
+                        let elem = Rewrite::SliceTail(Box::new(arr), Box::new(idx));
+                        hir_rw = Rewrite::Ref(Box::new(elem), mutbl_from_bool(mutbl));
+                    }
```
kkysen, line 242: For this `match`, can you hoist the `hir_rw = ` out of the `match` so it's clear it's unconditionally re-assigned?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -0,0 +1,392 @@
...
+                    Adjust::Borrow(AutoBorrow::Ref(_, mutbl)) => {
+                        hir_rw = Rewrite::Ref(Box::new(hir_rw), mutbl.into());
+                    }
+                    Adjust::Borrow(AutoBorrow::RawPtr(mutbl)) => {
+                        hir_rw = Rewrite::AddrOf(Box::new(hir_rw), mutbl.into());
+                    }
+                    Adjust::Pointer(PointerCast::Unsize) => {
+                        // TODO: figure out what kind of unsize this is and insert an explicit cast
+                        // (e.g. `&x` -> `&x as &[_]`).  In many cases this should work without a cast.
```
kkysen, line 298: Should this have a `todo!()`?

c2rust-analyze/src/rewrite/expr/hir_op.rs
kkysen: Actually, can you extract the whole `match` into its own method?

c2rust-analyze/src/rewrite/expr/hir_op.rs
kkysen: Then I'd do a `.fold` over `adjusts` with `hir_rw` as the accumulator.  For the prior `RewriteKind` `match`, too.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,370 @@
...
+fn convert_int_ty(x: ty::IntTy) -> ast::IntTy {
+    match x {
+        ty::IntTy::Isize => ast::IntTy::Isize,
+        ty::IntTy::I8 => ast::IntTy::I8,
+        ty::IntTy::I16 => ast::IntTy::I16,
+        ty::IntTy::I32 => ast::IntTy::I32,
+        ty::IntTy::I64 => ast::IntTy::I64,
+        ty::IntTy::I128 => ast::IntTy::I128,
+    }
+}
```
kkysen, line 133: I'm surprised things like this aren't already in a `From` `impl`.  We could coalesce them into a locally-defined (for the orphan rule) `trait Convert<T>` or something, which helps with future discoverability.

c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+        // Expr precedence:
+        // - Index, SliceTail: 3
+        // - Ref, Deref: 2
+        // - CastUsize: 1
+        //
+        // Currently, we don't have any type builders that require parenthesization.
+
+        match *self {
+            Rewrite::Identity => write!(f, "$e"),
+            Rewrite::Sub(i, _) => write!(f, "${}", i),
```
kkysen, line 77: What do the `$e` and `${}` mean?

c2rust-analyze/src/rewrite/ty.rs
aneksteind: `OffsetPtr` is essentially pointer arithmetic, e.g. `px + size_of<type_of<x>>`. Rather than `std::ptr::offset` it appears that the address is inherently (and temporarily) calculated by getting the address of an element at a particular index instead of calculating the offset

c2rust-analyze/src/rewrite/ty.rs
aneksteind: not yet implemented

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: i'm hesitant to add new places that can panic given our stated intention of making static analysis more robust (i.e. not crashing) for unsupported features

kkysen: Yeah, that's better actually.  Maybe just an `eprintln!` warning then?

c2rust-analyze/src/rewrite/mod.rs
spernsteiner: Mainly so we can apply all rewrites in a single pass.  Doing e.g. all the expressions in one pass and all the types in a second pass would be significantly more complicated.  To do it in a single pass, it's easier to use a single type for all rewrites because in the tree of rewrites built in `rewrite::apply`, an expr-rewrite node might have a ty-rewrite as a child, and (in theory) the reverse is possible as well.

c2rust-analyze/src/rewrite/span_index.rs
spernsteiner: This is a tiny optimization.  Calling `span.lo()` is equivalent to `span.data().lo`, and if `span` is interned, each `span.data()` is somewhat expensive since it accesses the span interner.  With this approach, we only access the interner once, no matter how many times the closure runs.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: By "child", I just mean a `Ty` that appears inside of this `Ty`.  For example, the children of `(u32, (u8, u8))` are `u32` and `(u8, u8)`.  "Descendant" is the transitive closure of "child".  I'll update this comment.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: Just clarifying that what we're checking here, abstractly, is "the child has a rewrite or one of its descendants has a rewrite".  The expression `child.label.ty_desc.is_some()` is not obviously connected to "child has a rewrite" IMO.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: That doesn't seem like it'll get too verbose, so I'll go ahead and switch to `mir::Ty` / `mir::TyKind` to avoid confusion.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: `rw_lty.label.ty_desc` is only `Some` for pointer and reference types.  The `type_desc` module in general is only about pointers - its purpose is to convert pointer permissions into a matching safe reference type.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: This causes insertion sort to use `&mut [c_int]` instead of the hypothetical `OffsetPtrMut<c_int>`.

`OffsetPtr` is a hypothetical library type that supports both forward and backward pointer arithmetic within a certain range, unlike slices which only support forward pointer arithmetic.  Whereas `&[T]` keeps a current position (`&slice[0]`) and max forward offset (`slice.len()`), `OffsetPtr<T>` would keep a current position, max forward offset, and max backward offset.  The max backward offset must be known in order to properly bounds-check pointer arithmetic with a negative offset.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: If a pointer `*mut T`/`*const T` has `Ownership::Cell`, we rewrite it into `&Cell<T>`.

>Where would the Cell and mut case go, then?

At this point, we've already decided which pointer type we're rewriting into (as indicated by `own` and `qty`), and we just need to do the rewrite.  The handling of various combinations of `WRITE`/`UNIQUE`/`CELL` permissions and flags is done in `type_desc::perms_to_desc`, which is what computes the `Ownership` and `Quantity` for each pointer type.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: Yes, function signatures only for now.  There are TODOs below for "update let bindings" and "update cast RHS types", which would cover types that appear inside function bodies.  There's currently no TODO comment about struct fields, since those weren't supported when I wrote this code, but I'll add one.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: You have it right.  Ancestors are "nodes whose spans strictly contain [the item's] span".

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Yes, it's legal to do things like `(1 + 2) .. (3 + 4)`

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: `do_prim` is used to factor out a bunch of case bodies in the match below, and the result type of that match is `Option<Vec<...>>` (because the non-primitive cases return non-empty lists when the result is `Some`).

c2rust-analyze/src/rewrite/ty.rs
kkysen: I agree; it's clearer this way.

c2rust-analyze/src/rewrite/mod.rs
spernsteiner: The syntax here is arbitrary since it's only used for debug prints.  I picked `$e` (like a macro placeholder) to represent the whole original expression, and `$0`, `$1`, etc for the subexpressions of the original expression.

kkysen: Oh okay, I didn't realize it was only for debug prints.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: Seems weird to depend on which types happen to be mentioned in some diagnostic suggestion.  I think we should just define a helper function for this lookup (once we're doing this for more than one type), and maybe add some caching if it turns out to be a performance problem.

Anyway, since `Cell` isn't a diagnostic item at the moment, I don't think there's anything to change here for now.

c2rust-analyze/src/rewrite/expr/hir_op.rs
kkysen: What about `mir_to_hir_rewrite`?

c2rust-analyze/tests/filecheck/field_temp.rs
spernsteiner: There's no such rewrite in `list_set`.  `list_get` uses a rewrite like this to upcast `(*l).data` from `&mut [i32]` to `&[i32]`, but `list_set` needs `&mut [i32]` and therefore doesn't cast.

c2rust-analyze/tests/filecheck/fields.rs
spernsteiner: Currently neither field expressions (e.g. `my_struct.x`) nor field type declarations (e.g. `x: *mut i32`) get rewritten.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Added a comment.  `a.hi <= b.lo` checks if `b` comes after `a` without overlapping.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Added a comment.  This is slightly different from the previous loop since it commits everything left over on the stack instead of only the ones that come before a certain span.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: I think that's more trouble than it's worth.  There are only four places with similar comparisons (2x `a.hi() <= b.lo()`, 1x `a.hi() <= lo`, 1x `hi <= b.lo()` - hence the trouble), all four already have comments explaining the high-level logic, and it's not too hard to figure out the exact meaning of the comparison regardless (low vs high comparisons like this are fairly common when working with ranges, I think).

Note the existing `SpanLike` methods exactly match methods on the actual `Span` type, and `Span` doesn't have a method for this sort of comparison.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: >Can't we just parenthesize everything and let another tool remove the unnecessary ones?

I tried implementing this, but it makes the output much harder to read.  Here's an example:
```diff
--- a/c2rust-analyze/tests/filecheck/insertion_sort_rewrites.rs
+++ b/c2rust-analyze/tests/filecheck/insertion_sort_rewrites.rs
@@ -9,20 +9,20 @@ extern crate libc;
 pub unsafe extern "C" fn insertion_sort(n: libc::c_int, p: *mut libc::c_int) {
     let mut i: libc::c_int = 1 as libc::c_int;
     while i < n {
-        // CHECK-DAG: [[#@LINE+2]]: p: &*$e
-        // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): &(&$0[($1 as usize) ..])[0]
+        // CHECK-DAG: [[#@LINE+2]]: p: (&(*($e)))
+        // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): (&((&(($0)[(($1) as usize) ..]))[(0)]))
         let tmp: libc::c_int = *p.offset(i as isize);
         let mut j: libc::c_int = i;
-        // CHECK-DAG: [[#@LINE+2]]: p: &*$e
-        // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): &(&$0[($1 as usize) ..])[0]
+        // CHECK-DAG: [[#@LINE+2]]: p: (&(*($e)))
+        // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): (&((&(($0)[(($1) as usize) ..]))[(0)]))
         while j > 0 as libc::c_int && *p.offset((j - 1 as libc::c_int) as isize) > tmp {
-            // CHECK-DAG: [[#@LINE+3]]: p: &*$e
-            // CHECK-DAG: [[#@LINE+2]]: p.offset({{.*}}): &(&$0[($1 as usize) ..])[0]
-            // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): &mut (&mut $0[($1 as usize) ..])[0]
+            // CHECK-DAG: [[#@LINE+3]]: p: (&(*($e)))
+            // CHECK-DAG: [[#@LINE+2]]: p.offset({{.*}}): (&((&(($0)[(($1) as usize) ..]))[(0)]))
+            // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): (&mut ((&mut (($0)[(($1) as usize) ..]))[(0)]))
             *p.offset(j as isize) = *p.offset((j - 1 as libc::c_int) as isize);
             j -= 1
         }
-        // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): &mut (&mut $0[($1 as usize) ..])[0]
+        // CHECK-DAG: [[#@LINE+1]]: p.offset({{.*}}): (&mut ((&mut (($0)[(($1) as usize) ..]))[(0)]))
         *p.offset(j as isize) = tmp;
         i += 1
     }
```
I find it a lot easier to tell what's going on in the old versions and to check if I'm getting the outputs I was expecting.

I added some comments to explain the purpose of `prec` instead.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Added a comment explaining the test.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Added comments explaining the test.

c2rust-analyze/src/rewrite/mod.rs
spernsteiner: Added a comment.

c2rust-analyze/src/rewrite/apply.rs
kkysen: That works I think.  I forgot we have to still interact with the rewrites themselves during testing, not just the final rewritten code after we can run `clippy` or `cargo check --fix` on it.

c2rust-analyze/src/rewrite/apply.rs
aneksteind: Alright, I don't feel too strongly about this but maintain it would make things clearer.

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,522 @@
...
+    pub fn build(
+        mut rws: Vec<(S, Rewrite<S>)>,
+    ) -> (Vec<RewriteTree<S>>, Vec<(S, Rewrite<S>, RewriteError<S>)>) {
+        // Sort by start position and then by decreasing length, so that each parent span comes
+        // before all its children.
+        rws.sort_by_key(|&(ref s, _)| (s.lo(), Reverse(s.hi() - s.lo())));
+
+        // The stack contains a `RewriteTree` node for the most recently processed item from `rws`
+        // along with all of its ancestors (nodes whose spans strictly contain the item's span).
+        // `stack[0]` is the outermost ancestor, and `stack.last()` is the most recently pushed
```
aneksteind, line 65: I think the relationship between the order of the trees in the stack and the order of `rws` can be made clearer. From a first read it was not obvious that the stack order implies that index `n` is a parent of index `n-1` if there is complete overlap, and the reverse for `rws`. That's made clearer in the code below with things like `if let Some(parent) = stack.last_mut()`

c2rust-analyze/src/rewrite/apply.rs
aneksteind: Right, I meant similar in the sense of describing why it commits a bunch of things. The comment LGTM

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,522 @@
...
+            commit(&mut stack, &mut out, rt);
+        }
+
+        (out, errs)
+    }
+}
+
+/// Split `rts` into the portion before `span`, the portion overlapping `span`, and the portion
+/// after `span`.  Nodes within `rts` should not overlap each other, and the list must be sorted by
+/// span; otherwise, the result are unspecified.
```
aneksteind, line 170: i think itd be inexpensive and worthwhile to assert that it is in facto sorted.  maybe a debug assert

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,522 @@
...
+    }
+
+    fn emit_span_with_rewrites(&mut self, span: Span, rts: &[RewriteTree]) {
+        let (_, overlap, _) = partition_nodes(rts, span);
+
+        let mut pos = span.lo();
+        for rt in overlap {
+            // Every child node is contained by the span of some `Rewrite::Identity` or
+            // `Rewrite::Sub` in its parent node.
+            debug_assert!(span.contains(rt.span));
```
aneksteind, line 328: would this assert be more appropriately placed in the partition function?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -0,0 +1,522 @@
...
+            "{:?}: warning: failed to apply rewrite {:?}: {:?}",
+            span, rw, err
+        );
+    }
+
+    let mut new_src = HashMap::new();
+    let mut rts = &rts as &[RewriteTree<Span>];
+    while rts.len() > 0 {
+        let file = source_map.lookup_source_file(rts[0].span.lo());
+        let idx = rts
```
aneksteind, line 362: i think this deserves a minor comment

c2rust-analyze/src/rewrite/span_index.rs
aneksteind, line 23: how do we know that the spans in the partitioned iterator all have a hi <= span.hi (otherwise would they not be fully contained)?

aneksteind: nevermind, i see this is handled in the next implementation

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: This is slightly stricter than the guarantee provided by `partition_nodes`, which implies only that `span.overlaps(rt.span)`.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: I edited the comments in `RewriteTree::build` - let me know if the new text seems clearer.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,394 @@
...
+    rw_lty: RwLTy<'tcx>,
+) -> ty::Ty<'tcx> {
+    let tcx = *lcx;
+    lcx.rewrite_unlabeled(rw_lty, &mut |ty, args, label| {
+        let (own, qty) = match label.ty_desc {
+            Some(x) => x,
+            None => return ty,
+        };
+
+        assert_eq!(args.len(), 1);
```
aneksteind, line 224: one argument because it's a pointer? if so can you add a comment explaining so?

c2rust-analyze/src/rewrite/expr/hir_op.rs
spernsteiner: I factored out the block into a function, but didn't switch to `fold` since I don't think it makes the code any clearer.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: Added a comment.

spernsteiner:
Updated based on feedback and rebased on current master

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -0,0 +1,394 @@
...
+            let perms = asn.perms()[lty.label];
+            let flags = asn.flags()[lty.label];
+            // TODO: if the `Ownership` and `Quantity` exactly match `lty.ty`, then `ty_desc` can
+            // be `None` (no rewriting required).  This might let us avoid inlining a type alias
+            // for some pointers where no actual improvement was possible.
+            Some(type_desc::perms_to_desc(perms, flags))
+        };
+        let descendant_has_rewrite = args.iter().any(|child| {
+            let has_rewrite = child.label.ty_desc.is_some();
+            has_rewrite || child.label.descendant_has_rewrite
```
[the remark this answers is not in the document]
aneksteind: i think it appears that it's relying on an invariant of some sorts. it wasn't clear to me that there was no other real way to implement `relabel_with_args`. i'm not exactly sure what would make it clearer but am open to what you think is reasonable

c2rust-analyze/tests/filecheck/field_temp.rs
```diff
@@ -0,0 +1,20 @@
...
+// that case when lifting MIR rewrites to HIR.
+pub struct MyList {
+    data: *mut i32,
+    len: usize,
+}
+
+// CHECK-LABEL: final labeling for "list_get"
+pub unsafe fn list_get(l: *const MyList, i: usize) -> i32 {
+    // The temporary `(*l).data` requires a MIR `MutToImm` rewrite.
+    // CHECK: ([[#@LINE+2]]: (*l).data): &[i32]
```
aneksteind, line 11: this is causing tests to fail, i believe you can omit the `#` to fix the issue

aneksteind (approved):

c2rust-analyze/src/rewrite/ty.rs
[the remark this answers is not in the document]
spernsteiner: Added a small comment

Issue #736, opened 2022-11-22
static analysis doesn't support `Aggregate` initialization


aneksteind:
The following:

```rust
let x = fdevents {
        fdarray: 0 as *mut *mut fdnode,
    };
```

which has the corresponding example MIR:
```
_22 = fdevents { fdarray: move _23 }
```

will trigger the following panic in `type_of_rvalue`:

```
thread 'rustc' panicked at 'not yet implemented: type_of Aggregate', c2rust-analyze/src/context.rs:358:56
```

kkysen:
Fixed by #856.

PR #817, opened 2023-02-07
(`c2rust-analyze`) Skip `Trivial` function calls.

### LINKED ISSUE #785: static analysis does not support `wrapping_sub` or `wrapping_add`
in line with #734 and #746, we'd like to be able to support something like the following:
```rust
unsafe extern "C" fn connections_get_new_connection(
    mut srv: *mut server,
    ...
) -> *mut connection {
    (*srv).lim_conns = ((*srv).lim_conns).wrapping_sub(1); // <--- currently causes a panic
```

### LINKED ISSUE #794: Handle (skip) library functions that don't take or return pointers

kkysen:
- Fixes #794.
- Fixes #785.

As per #807, this skips analysis of `Trivial` function calls, i.e. function calls that have no effect on pointer permissions in their caller, which means their argument and return types have no pointers within them.

This is done by defining a `trait IsTrivial` implemented for `FnSig`, `Ty`, etc.  `Ty` is the important one, and for `Ty`, we determine triviality this way:
* `RawPtr` is non-trivial
* primitive types are trivial
* `Never` is trivial
* `Foreign`, `Dynamic` are pessimistically non-trivial, as it's impossible to determine
    * of these, `Foreign` is likely non-trivial, and `Dynamic` is likely trivial
* `Ref` delegates to its inner type
* `Slice`, `Array`, `Tuple`, `Adt` all delegate to all of their inner types
* other types are pessimistically considered non-trivial as we're not sure how to handle them yet
    * these print a warning that they're not known yet, and for some, a guess is made:
        * `Opaque` tries to un-`Opaque` its type
        * `FnPtr`, `FnDef` are trivial, as they have contain no data
        * `Closure` delegates to its closure type (upvars)
        * `Generator` delegates to its inner types (upvars, resume, return, and yield types, and the witness)

As it's not currently needed yet, caching/supporting self-referential (through references) types (#834) and safe `std` types being `Trivial` (#820) are put off until later.

### DIFF
--- a/c2rust-analyze/src/util.rs
+++ b/c2rust-analyze/src/util.rs
@@ -345,7 +345,8 @@ impl<'tcx> IsTrivial<'tcx> for Ty<'tcx> {
     /// [`is_trivial`]: IsTrivial::is_trivial
     fn is_trivial(&self, tcx: TyCtxt<'tcx>) -> bool {
         let not_sure_yet = |is_trivial: bool| {
-            eprintln!("assuming non-trivial for now as a safe backup (guessed {is_trivial:?}): ty = {self:?}");
+            let kind = self.kind();
+            eprintln!("assuming non-trivial for now as a safe backup (guessed {is_trivial:?}): ty.kind() = {kind:?}, ty = {self:?}");
             false
         };
 
--- a/c2rust-analyze/src/util.rs
+++ b/c2rust-analyze/src/util.rs
@@ -4,12 +4,12 @@ use rustc_hir::def_id::DefId;
 use rustc_middle::mir::{
     Field, Local, Mutability, Operand, PlaceElem, PlaceRef, ProjectionElem, Rvalue,
 };
+use rustc_middle::ty::GenSig;
 use rustc_middle::ty::{
-    AdtDef, Binder, DefIdTree, FnSig, SubstsRef, Ty, TyCtxt, TyKind, TypeSuperVisitable,
-    TypeVisitable, TypeVisitor, UintTy,
+    self, subst::Subst, AdtDef, Binder, DefIdTree, EarlyBinder, FnSig, SubstsRef, Ty, TyCtxt,
+    TyKind, UintTy,
 };
 use std::fmt::Debug;
-use std::ops::ControlFlow;
 
 #[derive(Debug)]
 pub enum RvalueDesc<'tcx> {
@@ -140,7 +140,7 @@ pub fn ty_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> Option<Callee<'tcx>> 
         _ => return None,
     };
 
-    if let Some(callee) = builtin_callee(tcx, did, substs) {
+    if let Some(callee) = builtin_callee(tcx, ty, did) {
         return Some(callee);
     }
     Some(Callee::Other {
@@ -149,89 +149,8 @@ pub fn ty_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> Option<Callee<'tcx>> 
     })
 }
 
-trait IsTrivial {
-    /// Something [`is_trivial`] if it has no effect on pointer permissions,
-    /// and thus requires no special handling.
-    /// That is, it must contain no raw pointers.
-    ///
-    /// See [`Trivial`] for more info.
-    ///
-    /// [`is_trivial`]: Self::is_trivial
-    /// [`Trivial`]: Callee::Trivial
-    fn is_trivial(&self) -> bool;
-}
-
-impl IsTrivial for Ty<'_> {
-    /// A [`Ty`] [`is_trivial`] if
-    /// it is not a (raw) pointer and contains no (raw) pointers.
-    ///
-    /// [`is_trivial`]: IsTrivial::is_trivial
-    fn is_trivial(&self) -> bool {
-        // Similar logic to [`Ty::contains`].
-        struct Visitor;
-
-        impl<'tcx> TypeVisitor<'tcx> for Visitor {
-            type BreakTy = ();
-
-            fn visit_ty(&mut self, ty: Ty<'tcx>) -> ControlFlow<Self::BreakTy> {
-                if ty.is_unsafe_ptr() {
-                    ControlFlow::BREAK
-                } else {
-                    ty.super_visit_with(self)
-                }
-            }
-        }
-
-        self.visit_with(&mut Visitor) != ControlFlow::BREAK
-    }
-}
-
-impl IsTrivial for FnSig<'_> {
-    /// A [`FnSig`] [`is_trivial`] if
-    /// all of its argument and return types are [`Trivial`].
-    ///
-    /// [`is_trivial`]: IsTrivial::is_trivial
-    /// [`Trivial`]: Callee::Trivial
-    fn is_trivial(&self) -> bool {
-        self.inputs_and_output.iter().all(|ty| ty.is_trivial())
-    }
-}
-
-impl IsTrivial for Binder<'_, FnSig<'_>> {
-    /// A [`Binder`]`<`[`FnSig`]`>` [`is_trivial`] if
-    /// the [`FnSig`] without the [`Binder`] is trivial.
-    ///
-    /// Lifetimes do not affect [`Trivial`]ity,
-    /// so we can ignore higher-kinded lifetimes from the [`Binder`].
-    ///
-    /// [`is_trivial`]: IsTrivial::is_trivial
-    /// [`Trivial`]: Callee::Trivial
-    fn is_trivial(&self) -> bool {
-        // We don't care about higher-kinded lifetimes here, as we don't care about lifetimes at all.
-        self.skip_binder().is_trivial()
-    }
-}
-
-/// A callee [`is_trivial`] if
-/// its [`FnSig`] [`is_trivial`] after [`subst_and_normalize_erasing_regions`].
-///
-/// [`is_trivial`]: IsTrivial::is_trivial
-/// [`subst_and_normalize_erasing_regions`]: TyCtxt::subst_and_normalize_erasing_regions
-fn is_callee_trivial<'tcx>(tcx: TyCtxt<'tcx>, did: DefId, substs: SubstsRef<'tcx>) -> bool {
-    let param_env = tcx.param_env(did);
-    let did = tcx.subst_and_normalize_erasing_regions(substs, param_env, did);
-
-    // Typed because rust-analyzer can't resolve it.
-    let fn_sig: Binder<FnSig> = tcx.fn_sig(did);
-    fn_sig.is_trivial()
-}
-
-fn builtin_callee<'tcx>(
-    tcx: TyCtxt<'tcx>,
-    did: DefId,
-    substs: SubstsRef<'tcx>,
-) -> Option<Callee<'tcx>> {
-    if is_callee_trivial(tcx, did, substs) {
+fn builtin_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, did: DefId) -> Option<Callee<'tcx>> {
+    if ty.fn_sig(tcx).is_trivial(tcx) {
         return Some(Callee::Trivial);
     }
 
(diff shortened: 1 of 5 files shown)


kkysen:
@aneksteind, are there other instances of trivial function calls that were removed/I should add back in?  I did the two `u32::wrapping_{sub,add}` ones.

c2rust-analyze/src/util.rs
```diff
@@ -145,11 +149,92 @@ pub fn ty_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> Option<Callee<'tcx>>
...
+    ///
+    /// [`is_trivial`]: IsTrivial::is_trivial
+    fn is_trivial(&self) -> bool {
+        // Similar logic to [`Ty::contains`].
+        struct Visitor;
+
+        impl<'tcx> TypeVisitor<'tcx> for Visitor {
+            type BreakTy = ();
+
+            fn visit_ty(&mut self, ty: Ty<'tcx>) -> ControlFlow<Self::BreakTy> {
```
aneksteind, line 176: using `ControlFlow` feels like overkill. It's pretty obvious what returning `true` or `false` would mean here in this visitor, and no value in `Break(..)` or `Continue(..)` is used anyway.

aneksteind:
This LGTM after the comment is addressed

c2rust-analyze/src/util.rs
spernsteiner: This is implementing a trait method from rustc, and [the trait declaration](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/visit/trait.TypeVisitor.html) uses `ControlFlow`.

spernsteiner:
This implementation doesn't recurse into ADT fields, so it will wrongly consider e.g. `NonNull<T>` to be a trivial type.

Recursion into ADT fields will require some sort of caching, not only for performance but also because a struct might contain a reference to itself, and references are considered trivial and therefore don't break out of the traversal.  (The other way for a struct to refer to itself is through a raw pointer, which would break out due to the pointer.)  Here's an example:
```Rust
struct Foo {
    a: &'static Foo,
    b: *mut u8,
}
fn f(x: Foo) {}
```
Function `f` should be considered non-trivial, as it has access to a raw pointer through `x.b` (and also `x.a.b`, `x.a.a.b`, etc).

kkysen:
> This implementation doesn't recurse into ADT fields, so it will wrongly consider e.g. `NonNull<T>` to be a trivial type.

I did not realize that.  Thanks for the catch.  I based it off of `Ty::contains`, which says it recursively checks if a type contains a type, but I guess it only recurses into generic type parameters?  Not the best documentation.

> Recursion into ADT fields will require some sort of caching, not only for performance but also because a struct might contain a reference to itself, and references are considered trivial and therefore don't break out of the traversal. (The other way for a struct to refer to itself is through a raw pointer, which would break out due to the pointer.) Here's an example:
> 
> ```rust
> struct Foo {
>     a: &'static Foo,
>     b: *mut u8,
> }
> fn f(x: Foo) {}
> ```
> 
> Function `f` should be considered non-trivial, as it has access to a raw pointer through `x.b` (and also `x.a.b`, `x.a.a.b`, etc).

That makes sense.  I'll add catching for the self-referential types.

I did have a question, though.  References we consider safe/trivial and don't recurse into them.  Raw pointers are not, and so we exit early on them.  What about safe pointer wrappers like `Box` and `Vec`?  They ultimately contain raw pointers through `Unique<T>` => `NonNull<T>` => `*const T`, and so by the current `Trivial` definition, they're non-trivial.

spernsteiner:
>I based it off of Ty::contains, which says it recursively checks if a type contains a type, but I guess it only recurses into generic type parameters?

Looks like `TypeVisitor` ultimately traverses the `Ty` using [this function](https://github.com/rust-lang/rust/blob/bd39bbb4bb92df439bf6d85470e296cc6a47ffbd/compiler/rustc_middle/src/ty/structural_impls.rs#L685), which visits only the `Substs` of each `TyKind::Adt`, not the actual `AdtDef` (where the field declarations are listed).  I think in general, `TyKind::Adt` is treated as simply referring to the ADT, not containing it directly (it contains an `AdtDef` instead of a `DefId` only for convenience).

> What about safe pointer wrappers like Box and Vec? They ultimately contain raw pointers through Unique<T> => NonNull<T> => *const T, and so by the current Trivial definition, they're non-trivial.

Good point.  I think we can go ahead with treating those as nontrivial for now.  In the future we could extend the triviality check to treat certain safe library types as trivial if we encounter some functions that need it.

kkysen:
> Good point. I think we can go ahead with treating those as nontrivial for now. In the future we could extend the triviality check to treat certain safe library types as trivial if we encounter some functions that need it.

I'll make an issue for that as something we'll get to later.

kkysen:
@spernsteiner, can you check if the `Ty::is_trivial` function makes sense now?  I'm manually recursing through all the inner types now, and I think I fixed some errors with substituting generics correctly, but I haven't added the caching part yet.

For the recursing into inner types, I've basically done this:
* `RawPtr` is non-trivial
* primitive types and `Never`/`!` are trivial
* `Foreign` is assumed to be non-trivial
* `Ref`, `Slice`, `Array` delegate to their inner type
* `Tuple`, `Adt` delegate to all their fields
* `Dynamic` panics, as we don't know what type that could be (or I could assume it to be non-trivial)
* `FnPtr`, `FnDef` delegate to arguments and return type
* `Closure`, `Generator` (`async`) delegate to their compiler-generated types and function signatures
* `Opaque` tries to un-opaque the type and delegate to it
* the rest I'm not sure how to implement and may not come up, so they have a `todo!`

For the function-like types, `FnPtr`, `FnDef`, `Closure`, and `Generator`, I'm not sure if my implementation is correct.  I basically just check all inner types for triviality.  For a function call, I know that's correct for `FnDef`, but for a function pointer or something, should I be using that same logic?

For the caching, do you think I should have a global cache, or create a local one in each `builtin_callee` call?  The latter I assume is simpler, unless there's some way to store/attach information on `Ty`s or store a cache in `TyCtxt` that I don't know about.

I may also move the caching part to a separate PR.

spernsteiner:
This check is only approximate, and it's always safe to conservatively return non-trivial in any case we're not sure about.  Personally, I would just make all the non-obvious cases (everything on your list from `Dynamic` on down) return non-trivial for now, maybe with some TODO comments so we know we can improve it later if it becomes necessary.

>I may also move the caching part to a separate PR.

Caching is necessary to avoid a stack overflow on recursive ADTs like the `struct Foo` I posted above.

>For the caching, do you think I should have a global cache, or create a local one in each builtin_callee call?

A local cache is probably fine.

Normally there would be three possible states a particular `Ty` could be in: "haven't seen" this `Ty` before, "still processing" this `Ty` in some enclosing stack frame, or "yes"/"no" this `Ty` definitely is/isn't trivial.  And when looking up a `Ty`, we can optimistically treat an answer of "still processing" as "yes", since if the answer ultimately turns out to be "no" instead, that will be caught in frame is doing the processing.  But if the caching is purely local, and we return early (deleting the current cache) if anything ever comes back as "no", we can collapse this down to two states: "haven't seen" and "still processing or yes".

>unless there's some way to store/attach information on Tys or store a cache in TyCtxt that I don't know about.

We'd probably want to put the cache in the `GlobalAnalysisCtxt` or something like that.  There's no way to extend `Ty`/`TyCtxt` or create new queries from a separate crate.  The other option would be to put the cache in thread-local storage, which is probably overkill.

aneksteind:
> @aneksteind, are there other instances of trivial function calls that were removed/I should add back in? I did the two u32::wrapping_{sub,add} ones.

@kkysen I would checkout out the current `main.rs` in `lighttpd-minimal`, cross-reference that with the functions in `lighttpd-rs`, and check to see what may or may not be missing. I don't think there will be too much, but don't recall.

c2rust-analyze/src/util.rs
```diff
@@ -280,6 +284,134 @@ fn builtin_callee<'tcx>(
...
+
+            ty::Foreign(..) => false, // no introspection into a foreign, extern type, but as it's extern, it likely contains raw pointers
+
+            // delegate to the inner type
+            ty::Ref(_, ty, _) | ty::Slice(ty) | ty::Array(ty, _) => ty.is_trivial(tcx),
+
+            // delegate to all inner types
+            ty::Tuple(tys) => are_all_trivial(tcx, tys),
+            ty::Adt(adt_def, substs) => {
+                are_all_trivial(tcx, adt_def.all_fields().map(|field| field.ty(tcx, substs)))
```
aneksteind, line 360: a crude way to avoid stack overflow is to pass the mutable cache into all of these calls, but here is where itll be most important. i'm not sure where besides an ADT we need to worry about recursive definitions, but what @spernsteiner was talking about in another comment can be accomplished with fixed point iteration as is done [here](https://github.com/immunant/c2rust/blob/0a2b64b4f83b42f08fe13c3d4fbd8b5b167f07a8/c2rust-analyze/src/main.rs#L159). iteration ends when the cache remains the same from iteration `i-1` to `i`. In that linked block nothing is explicitly marked as "done", but all are presumed "done" when iteration ends

kkysen:
> This check is only approximate, and it's always safe to conservatively return non-trivial in any case we're not sure about. Personally, I would just make all the non-obvious cases (everything on your list from `Dynamic` on down) return non-trivial for now, maybe with some TODO comments so we know we can improve it later if it becomes necessary.

I think we need at least `FnDef` for the initial signature check.

> > I may also move the caching part to a separate PR.
> 
> Caching is necessary to avoid a stack overflow on recursive ADTs like the `struct Foo` I posted above.

I understand, I just want to reduce each PR's complexity if possible.  We don't have any recursive ADTs yet, so nothing's gonna crash yet, so I'll add the caching in a follow-up PR.

c2rust-analyze/src/util.rs
spernsteiner: >fixed point iteration

I think that's overkill here.  We can just treat "in progress" as "done" whenever it appears in a recursive call.

One way to think about what this code is doing is to say that we want to visit every distinct type that appears in the complete unrolling of the type.  The unrolling may be infinite if there are recursive types involved.  However, the set of distinct types in the unrolling is still finite.  If you look at the `struct Foo` example previously, unrolling `Foo` once produces `Foo` and `*mut u8`, but unrolling `Foo` a second time just produces `Foo` and `*mut u8` again, both of which were already visited.  So visiting the same type a second time can simply be skipped, as everything that it adds was already seen elsewhere in the traversal.  (Note what I just described is essentially a breadth-first traversal of the type, while what we've actually implemented is depth-first.  But the logic works out the same, just with the slight nuance that in the second unfolding of `Foo`, the type `*mut u8` hasn't actually been processed yet, but is guaranteed to be processed at some point due to the first unfolding of `Foo`.)

spernsteiner:
>I think we need at least FnDef for the initial signature check.

That could be special-cased at the top level.  Also, I suspect (though I haven't thought this out fully) that `FnDef` as an argument/result of a function should be considered trivial, since it's zero-sized and carries no actual data.

kkysen:
> That could be special-cased at the top level.

Yeah, I can do that.

> Also, I suspect (though I haven't thought this out fully) that `FnDef` as an argument/result of a function should be considered trivial, since it's zero-sized and carries no actual data.

I think you're right.  And then closures are just their enclosed type's triviality, but not the function part of it.

c2rust-analyze/src/util.rs
aneksteind: Yeah, you're right. I anchored my thinking with the previous approach :)

kkysen:
@spernsteiner, @aneksteind, can you check the code now for non-self-referential/non-cached correctness?  I'm going to add the caching in a follow-up PR so it doesn't get too complex here.

kkysen:
I've also added `FileCheck` tests for this, but they're blocked on #823 as they currently crash on `type_of Aggregate` (#736), but output the triviality checks first, so `FileCheck` can still work.

aneksteind (dismissed):
please include test cases for triviality when you're enabled to do so

kkysen:
> please include test cases for triviality when you're enabled to do so

Added now.

c2rust-analyze/src/util.rs
kkysen: Left for #834, which has a link back to these comments.

c2rust-analyze/src/util.rs
```diff
@@ -145,11 +146,13 @@ pub fn ty_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> Option<Callee<'tcx>>
     })
 }
 
-fn builtin_callee<'tcx>(
-    tcx: TyCtxt<'tcx>,
-    did: DefId,
-    _substs: SubstsRef<'tcx>,
-) -> Option<Callee<'tcx>> {
+fn builtin_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, did: DefId) -> Option<Callee<'tcx>> {
+    let is_trivial = ty.fn_sig(tcx).is_trivial(tcx);
```
spernsteiner, line 150: `ty.fn_sig(tcx)` will return sigs for `FnPtr`s and `Closure`s.  Replacing it with `tcx.fn_sig(did)` I think makes it clearer that we only expect to be working with `FnDef`s here.  It would also let you revert the change to the signature of `builtin_callee`.

spernsteiner (approved):

c2rust-analyze/src/util.rs
kkysen: `ty.fn_sig`, 


```rust
    pub fn fn_sig(self, tcx: TyCtxt<'tcx>) -> PolyFnSig<'tcx> {
        match self.kind() {
            FnDef(def_id, substs) => tcx.bound_fn_sig(*def_id).subst(tcx, substs),
            FnPtr(f) => *f,
            Error(_) => {
                // ignore errors (#54954)
                ty::Binder::dummy(FnSig::fake())
            }
            Closure(..) => bug!(
                "to get the signature of a closure, use `substs.as_closure().sig()` not `fn_sig()`",
            ),
            _ => bug!("Ty::fn_sig() called on non-fn type: {:?}", self),
        }
    }
```

doesn't return anything for `Closure`s, only for `FnDef` and `FnPtr`.  For `FnPtr`, I don't see why we'd treat function pointer calls different from function calls, as there's no hidden closure, and either way we just calculate the triviality based on the argument and return types.  `builtin_callee` is only called after we previously checked for `FnDef`, but we can relax that check when the need arises.  And if I use `tcx.fn_sig` instead, then I have to needlessly repeat the whole `tcx.bound_fn_sig(*def_id).subst(tcx, substs)`.

Issue #816, opened 2023-02-07
analyze: null pointer support


spernsteiner:
We should add support to `c2rust-analyze` for tracking (possibly) null pointers.

First, we should add a new pointer permission: `PermissionSet::NON_NULL`.  A pointer with this permission is known to never be null at run time.  `NON_NULL` is set (or not) when the pointer is created, and it flows forward along dataflow edges.  The results of `Rvalue::Ref` and `Rvalue::AddressOf` should have the `NON_NULL` permission set; `ptr::null()`, `ptr::null_mut()`, and casting the constant 0 to a pointer type do not set the permission; and casting other integers to pointer types is (still) unsupported and therefore panics.

The permission is `NON_NULL` rather than `NULL` because we allow dropping permissions in any assignment.  This means removing a permission from a pointer's `PermissionSet` must allow the pointer to take on more values, not restrict it to fewer values.

The reason to continue rejecting nonzero constant integer-to-pointer casts like `3 as *const u8` is to avoid losing information and potentially changing program behavior.  The result of `3 as *const u8` can't be a value of type `&u8` because it's not a valid pointer, and it can't be `None` because that would lose the fact that the value is 3 and not 0.  We don't want `0 as *const u8 == 3 as *const u8` to be rewritten into `None == None` and wrongly evaluate to `true`.

Second, we'll need to update the `type_desc` and `rewrite::ty` modules to recognize the lack of `NON_NULL` and use `Option<_>` when needed.

Third, `rewrite::expr` will need to insert `Some(_)`, `None`, and `_.unwrap()` as needed.  Casts from `NON_NULL` to nullable pointers will need to wrap the pointer expression in `Some(_)`.  Operations that produce null pointers should be replaced with `None`.  And dereferences of nullable pointers should call `_.unwrap()` first.

Some operations may need special handling.  I think `offset` should require a non-null pointer (meaning `rewrite::expr` should `unwrap()` the input if it's nullable, like it does for deref), since it's UB to call it on invalid pointers.  This means `offset` can unconditionally return `NON_NULL` regardless of its input.  And `is_null` can be converted to `is_some()` if the pointer is nullable or replaced with literal `true` if it's `NON_NULL`.  (Converting `if p.is_null() { ... }` to `if let Some(p) = p { ... }` would be nice to have, but isn't essential and would probably be a bit tricky.

[...]

PR #818, opened 2023-02-07
(`c2rust-analyze`) Add `PermissionSet::NON_NULL` with precise docs

kkysen:
- Fixes the first step of #816.

This adds `PermissionSet::NON_NULL` and precise docs explaining its meaning and intended usage, but doesn't contain any actual implementation or use of `NON_NULL`.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -41,6 +41,70 @@ bitflags! {
         const OFFSET_SUB = 0x0020;
         /// This pointer can be freed.
         const FREE = 0x0040;
+
+        /// This pointer is non-null.
+        ///
+        /// [`NON_NULL`] is set (or not) when the pointer is created,
+        /// and it flows forward along dataflow edges.
+        ///
+        /// The permission is [`NON_NULL`] rather than `NULL` because we allow dropping permissions in any assignment.
+        /// This means removing a permission from a pointer's [`PermissionSet`]
+        /// must allow the pointer to take on more values, not restrict it to fewer values.
+        ///
+        /// The following should be set to [`NON_NULL`]:
+        /// * the results of [`Rvalue::Ref`] and [`Rvalue::AddressOf`]
+        /// * a dereferent (i.e. [`READ`]` | `[`WRITE`]), as otherwise this would be UB
+        /// * the argument or result of a known function like [`_.offset`] where it is UB to pass it an invalid (null) pointer
+        ///
+        /// The following should not be set to [`NON_NULL`]:
+        /// * [`core::ptr::null`]
+        /// * [`core::ptr::null_mut`]
+        /// * [`core::ptr::from_exposed_addr`] with the constant `0`
+        /// * `0 as * {const,mut} _`, a cast from the constant `0` to a pointer type
+        /// * anything else
+        ///
+        /// Non-zero but invalid pointers, such as those produced by:
+        /// * [`core::ptr::invalid`]
+        /// * [`core::ptr::invalid_mut`]
+        /// * [`NonNull::dangling`]
+        /// * [`core::ptr::from_exposed_addr`]
+        /// * int-to-ptr casts though `as` or [`core::mem::transmute`]
+        ///
+        /// will not be [`NON_NULL`] but also will not be either [`READ`]` | `[`WRITE`],
+        /// which allows us to distinguish them.
+        ///
+        /// If a pointer is neither [`READ`] nor [`WRITE`], then it is invalid,
+        /// and thus will remain a pointer after rewriting.
+        ///
+        /// The rest below is for valid pointers ([`READ`]` | `[`WRITE`]):
+        ///
+        /// [`NON_NULL`] pointers will become references, e.x. `&T`.\
+        /// Non-[`NON_NULL`] pointers will become [`Option<&T>`].
+        ///
+        /// Casts/transitions from [`NON_NULL`] to non-[`NON_NULL`] will become [`Some(_)`].\
+        /// Casts/transitions from non-[`NON_NULL`] to [`NON_NULL`] will become [`_.unwrap()`].
+        ///
+        /// [`_.is_null()`] on a [`NON_NULL`] pointer will become [`false`].\
+        /// [`_.is_null()`] on a non-[`NON_NULL`] pointer will become [`_.is_some()`].
+        ///
+        /// Constant null pointers, like those produced by:
+        /// * [`core::ptr::null`]
+        /// * [`core::ptr::null_mut`]
+        /// * [`core::ptr::from_exposed_addr`] with the constant `0`
+        /// * `0 as * {const,mut} _`, a cast from the constant `0` to a pointer type
+        ///
+        /// will become [`None`].
+        ///
+        /// [`NON_NULL`]: Self::NON_NULL
+        /// [`READ`]: Self::READ
+        /// [`WRITE`]: Self::WRITE
+        /// [`_.offset`]: core::ptr::offset
+        /// [`NonNull::dangling`]: core::ptr::NonNull::dangling
+        /// [`Some(_)`]: Some
+        /// [`_.unwrap()`]: Option::unwrap
+        /// [`_.is_null()`]: core::ptr::is_null
+        /// [`_.is_some()`]: Option::is_some
+        const NON_NULL = 0x0080;
     }
 }
 


kkysen:
@spernsteiner, can you let me know if the definition/docs I added make sense/sound good?  They're a bit adjusted from what you wrote in #816 based on my own thoughts, some of which I commented on that issue, but the docs here are my final version so far.

c2rust-analyze/src/context.rs
```diff
@@ -41,6 +41,70 @@ bitflags! {
...
         const FREE = 0x0040;
+
+        /// This pointer is non-null.
+        ///
+        /// [`NON_NULL`] is set (or not) when the pointer is created,
+        /// and it flows forward along dataflow edges.
+        ///
+        /// The permission is [`NON_NULL`] rather than `NULL` because we allow dropping permissions in any assignment.
+        /// This means removing a permission from a pointer's [`PermissionSet`]
+        /// must allow the pointer to take on more values, not restrict it to fewer values.
```
spernsteiner, line 52: This bit from "we allow dropping permissions" onward applies to all permissions - could you move it to `PermissionSet`'s doc comment instead of `NON_NULL`'s?

c2rust-analyze/src/context.rs
```diff
@@ -41,6 +41,70 @@ bitflags! {
         const OFFSET_SUB = 0x0020;
         /// This pointer can be freed.
         const FREE = 0x0040;
+
+        /// This pointer is non-null.
+        ///
+        /// [`NON_NULL`] is set (or not) when the pointer is created,
+        /// and it flows forward along dataflow edges.
```
aneksteind, line 48: should it also flow backward? I'm thinking of a call to `offset`, e.g.

```
_1 = malloc(...);
_2 = std::ptr::offset(_1, ...);
something_else_with(_1); // pointer permissions for NON_NULL may be useful here
```

`_2` would get the `NON_NULL` permission and so would `_1`. It seems like this should happen regardless of the checks the initial C code made

c2rust-analyze/src/context.rs
```diff
@@ -41,6 +41,70 @@ bitflags! {
...
+        ///
+        /// The permission is [`NON_NULL`] rather than `NULL` because we allow dropping permissions in any assignment.
+        /// This means removing a permission from a pointer's [`PermissionSet`]
+        /// must allow the pointer to take on more values, not restrict it to fewer values.
+        ///
+        /// The following should be set to [`NON_NULL`]:
+        /// * the results of [`Rvalue::Ref`] and [`Rvalue::AddressOf`]
+        /// * a dereferent (i.e. [`READ`]` | `[`WRITE`]), as otherwise this would be UB
+        /// * the argument or result of a known function like [`_.offset`] where it is UB to pass it an invalid (null) pointer
```
spernsteiner, line 57: As this permission flows forward, it doesn't make sense to set it on the input to a dereference or to a function call, as there's nowhere for the permission to flow from there.  A more extreme approach, like taking the presence of `*p` to mean that the local `p` must be `NON_NULL`, would not fit well with our current flow-insensitive analysis.  Here's an example:
```
let p: *const u8 = either_null_or_valid();
if !p.is_null() {
    f(*p);
}
```
Since the analysis is flow-insensitive, there is no way for it to represent the idea that "`p` is nullable, but inside the `if` statement, it can be treated as `NON_NULL`".  `p` must either be nullable (for all time) or `NON_NULL` (for all time).  And making `p` `NON_NULL` and turning it into `&u8` means that there must be a panic somewhere in the case where `either_null_or_valid()` returns null, which (while allowed by our definition of correctness) is not what we want.

I think the "dereferent" point should be removed, and the "argument or result" point should be changed to something like "the result of a known function that only returns non-null pointers".

c2rust-analyze/src/context.rs
```diff
@@ -41,6 +41,70 @@ bitflags! {
...
+        ///
+        /// Non-zero but invalid pointers, such as those produced by:
+        /// * [`core::ptr::invalid`]
+        /// * [`core::ptr::invalid_mut`]
+        /// * [`NonNull::dangling`]
+        /// * [`core::ptr::from_exposed_addr`]
+        /// * int-to-ptr casts though `as` or [`core::mem::transmute`]
+        ///
+        /// will not be [`NON_NULL`] but also will not be either [`READ`]` | `[`WRITE`],
+        /// which allows us to distinguish them.
```
spernsteiner, line 74: I don't think it's true that invalid pointers will never be `READ` or `WRITE`.  Here's an example:
```Rust
/// Returns either a valid pointer, or a small integer (< 1000) cast to a pointer.
fn valid_or_small_int() -> *mut u8 { ... }

fn f() {
    let p: *mut u8 = valid_or_small_int();
    if p as usize < 1000 {
        // It's an integer converted to a pointer
        g1(p as usize);
    } else {
        // It's a real, valid pointer
        g2(*p);
    }
}
```
Here `p` is potentially invalid (e.g. it might be `3 as *mut u8`), but it is also `READ`, because the `else` case dereferences it.

In the future we might need a separate permission, `VALID`, meaning it's either a valid pointer or null.  Then `VALID | NON_NULL` means it's a real pointer (`&T`), `VALID` without `NON_NULL` means `Option<&T>`, and non-`VALID` means `Either<&T, usize>`.  But for now we reject any code that could produce a non-`VALID` pointer, and implicitly assume that `VALID` holds on all pointers we encounter (so only the first two cases can occur).

c2rust-analyze/src/context.rs
spernsteiner: I guess we had a race condition :)

In your example, it's fine to propagate backward, but if you were to wrap everything after the `malloc` in an `if !_1.is_null()`, then it wouldn't make sense to propagate `NON_NULL` outside that `if`.  And we don't currently have a notion of "this pointer X has permission Y, but only inside this block" - the pointer either has the permission or it doesn't.

aneksteind: based on other considerations with relaxing permissions, it sounds like it should not propagate backwards unless it's very clear there is no control flow, which we don't analyze anyway

c2rust-analyze/src/context.rs
kkysen: Sure, just this paragraph, right?

c2rust-analyze/src/context.rs
kkysen: Yeah, you're right about this.  I was mistakenly thinking about `READ` as must be readable, not may be read (may as in there's a chance, not may as it always capable).  I was thinking about a future `VALID` permission as well, though I was hoping we could get around it, but nevermind that.  But we'll leave that to when we have to handle any invalid pointers.  It should be `Either<&T, *T>`, though.  I'll change the docs back to forbidding invalid pointers for now.

c2rust-analyze/src/context.rs
spernsteiner: Right, just this paragraph.

c2rust-analyze/src/context.rs
kkysen: I Clarified some of these misconceptions I had in #868.

kkysen: Fixed in 287eaaf7b795545da904eafe62576d4a64c6e911.

c2rust-analyze/src/context.rs
kkysen: Also fixed in 1dde612baa11ac02692f9eafd8f5fe9f2b6cd7a2.

c2rust-analyze/src/context.rs
kkysen: @spernsteiner, in general, why is analysis flow-insensitive, or is that just how it currently is?  It seems like it could be significantly more powerful if it were flow-sensitive.  Then it could rewrite the above to:

```rust
let p: Option<*const u8> = either_null_or_valid();
if let Some(p) = p {
    f(*p);
}
```

and make `f` take a `NON_NULL`.  With nullability especially, it seems like we'll lose a ton of lifting possibilities without flow-sensitive analysis, as this is **the** common pattern of null checking in C.

That said, I can make the adjustments here (remove the deferent point and remove the arguments part) for now.  I was just wondering about this and our approach here.

kkysen: Fixed in 72fc9d031f7508d21a494727f9b4b124bbf277a8.

c2rust-analyze/src/context.rs
```diff
@@ -61,6 +61,60 @@ bitflags! {
...
+        ///
+        /// Non-zero but invalid pointers, such as those produced by:
+        /// * [`core::ptr::invalid`]
+        /// * [`core::ptr::invalid_mut`]
+        /// * [`NonNull::dangling`]
+        /// * [`core::ptr::from_exposed_addr`]
+        /// * int-to-ptr casts though `as` or [`core::mem::transmute`]
+        ///
+        /// will not be [`NON_NULL`], but for now,
+        /// we do not consider and do not allow such non-null invalid pointers at all.
```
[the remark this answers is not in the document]
kkysen: That makes sense.  Should it `may or not be [`NON_NULL`]`, though, as an invalid pointer could still be null (e.x. `ptr::invalid(0)`)?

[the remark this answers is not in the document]
spernsteiner: How about this?
```suggestion
        /// Non-zero but invalid pointers, such as those produced by
        /// the following operations (with nonzero arguments where applicable):
        /// * [`core::ptr::invalid`]
        /// * [`core::ptr::invalid_mut`]
        /// * [`NonNull::dangling`]
        /// * [`core::ptr::from_exposed_addr`]
        /// * int-to-ptr casts though `as` or [`core::mem::transmute`]
        ///
        /// will be [`NON_NULL`], but for now,
        /// we do not consider and do not allow such non-null invalid pointers at all.
```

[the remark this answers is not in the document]
kkysen: Wait, actually I think your original suggestion is good.  I wrote `Non-zero but invalid pointers`, so I already specified that basically.

spernsteiner (approved):

PR #826, opened 2023-02-13
(`c2rust-analyze`) Separate the two `*c_void` cast directions

kkysen:
This is a rework of #788 that properly separates the two `*c_void` cast directions.

In #788, these are conflated.  I'm not exactly sure if this can lead to a bug (the `Place`s in the map are not necessary unique), but the point is, the conflation makes it harder to understand and reason about, so it is harder to tell if there is one or not.  It may be correct due to details of our particular analysis and how MIR works, but then you have to understand all the details to know if it's correct or not, and it's a lot more brittle to future changes.

By separating the `From` and `To` directions here, it IMO makes it much easier to understand.  It allows the logic to be more easily separated, such as in
* `CVoidCastDirection::from_callee`
* `CVoidPtr::checked`
* `CVoidPtr::get_cast_from_stmt`
* `CVoidCasts::should_skip_stmt`

Separating the logic also allows each part to be more documented, which IMO makes it a lot easier to readon about and understand.

It also separates the `*c_void` casts code into its own module.

### DIFF
--- a/c2rust-analyze/src/c_void_casts.rs
+++ b/c2rust-analyze/src/c_void_casts.rs
@@ -300,56 +300,57 @@ impl<'tcx> CVoidCasts<'tcx> {
     /// [`*c_void`]: core::ffi::c_void
     fn insert_all_from_body(&mut self, body: &Body<'tcx>, tcx: TyCtxt<'tcx>) {
         for bb_data in body.basic_blocks().iter() {
-            if let Some(term) = &bb_data.terminator {
-                let term: &Terminator = term;
-                if let TerminatorKind::Call {
+            let term: &Terminator = match &bb_data.terminator {
+                Some(term) => term,
+                None => continue,
+            };
+            let (func, args, destination, target) = match term.kind {
+                TerminatorKind::Call {
                     ref func,
                     ref args,
                     destination,
                     target,
                     ..
-                } = term.kind
-                {
-                    let func_ty = func.ty(&body.local_decls, tcx);
+                } => (func, args, destination, target),
+                _ => continue,
+            };
+            let func_ty = func.ty(&body.local_decls, tcx);
 
-                    // For [`CVoidCastDirection::From`], we only count
-                    // a cast from `*c_void` to an arbitrary type in the subsequent block,
-                    // searching forward.
-                    let find_first_cast_succ_block = |get_cast| {
-                        let successor_block_id = target.unwrap();
-                        body.basic_blocks()[successor_block_id]
-                            .statements
-                            .iter()
-                            .find_map(get_cast)
-                    };
+            // For [`CVoidCastDirection::From`], we only count
+            // a cast from `*c_void` to an arbitrary type in the subsequent block,
+            // searching forward.
+            let find_first_cast_succ_block = |get_cast| {
+                let successor_block_id = target.unwrap();
+                body.basic_blocks()[successor_block_id]
+                    .statements
+                    .iter()
+                    .find_map(get_cast)
+            };
 
-                    // For [`CVoidCastDirection::From`], we only count
-                    // a cast to `*c_void` from an arbitrary type in the same block,
-                    // searching backwards.
-                    let find_last_cast_curr_block =
-                        |get_cast| bb_data.statements.iter().rev().find_map(get_cast);
+            // For [`CVoidCastDirection::From`], we only count
+            // a cast to `*c_void` from an arbitrary type in the same block,
+            // searching backwards.
+            let find_last_cast_curr_block =
+                |get_cast| bb_data.statements.iter().rev().find_map(get_cast);
 
-                    for direction in CVoidCastDirection::from_callee(ty_callee(tcx, func_ty))
-                        .iter()
-                        .copied()
-                    {
-                        use CVoidCastDirection::*;
-                        let c_void_ptr = match direction {
-                            From => destination,
-                            To => args[0].place().unwrap(),
-                        };
-                        let c_void_ptr = CVoidPtr::checked(c_void_ptr, &body.local_decls, tcx);
-                        let get_cast = move |stmt: &Statement<'tcx>| {
-                            c_void_ptr.get_cast_from_stmt(direction, stmt)
-                        };
-                        let cast = match direction {
-                            From => find_first_cast_succ_block(get_cast),
-                            To => find_last_cast_curr_block(get_cast),
-                        };
-                        if let Some(cast) = cast {
-                            self.insert(direction, cast);
-                        }
-                    }
+            for direction in CVoidCastDirection::from_callee(ty_callee(tcx, func_ty))
+                .iter()
+                .copied()
+            {
+                use CVoidCastDirection::*;
+                let c_void_ptr = match direction {
+                    From => destination,
+                    To => args[0].place().unwrap(),
+                };
+                let c_void_ptr = CVoidPtr::checked(c_void_ptr, &body.local_decls, tcx);
+                let get_cast =
+                    move |stmt: &Statement<'tcx>| c_void_ptr.get_cast_from_stmt(direction, stmt);
+                let cast = match direction {
+                    From => find_first_cast_succ_block(get_cast),
+                    To => find_last_cast_curr_block(get_cast),
+                };
+                if let Some(cast) = cast {
+                    self.insert(direction, cast);
                 }
             }
         }
--- a/c2rust-analyze/src/c_void_casts.rs
+++ b/c2rust-analyze/src/c_void_casts.rs
@@ -369,4 +369,10 @@ impl<'tcx> CVoidCasts<'tcx> {
             }
         }
     }
+
+    pub fn new(mir: &Body<'tcx>, tcx: TyCtxt<'tcx>) -> Self {
+        let mut this = Self::default();
+        this.insert_all_from_mir(mir, tcx);
+        this
+    }
 }
--- a/c2rust-analyze/src/c_void_casts.rs
+++ b/c2rust-analyze/src/c_void_casts.rs
@@ -1,0 +1,372 @@
+use std::borrow::Borrow;
+use std::collections::HashMap;
+
+use rustc_middle::{
+    mir::{Body, LocalDecls, Place, Rvalue, Statement, StatementKind, Terminator, TerminatorKind},
+    ty::{TyCtxt, TyKind},
+};
+
+use assert_matches::assert_matches;
+
+use crate::util::{ty_callee, Callee};
+
+/// The direction of a [`*c_void`](core::ffi::c_void) cast.
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub enum CVoidCastDirection {
+    /// From [`*c_void`](core::ffi::c_void) to another pointer type.
+    ///
+    /// This is used immediately after allocating,
+    /// after both [`Callee::Malloc`], [`Callee::Calloc`], and [`Callee::Realloc`].
+    From,
+
+    /// To [`*c_void`](core::ffi::c_void) from another pointer type.
+    ///
+    /// This is used immediately before freeing,
+    /// before both [`Callee::Free`] and [`Callee::Realloc`].
+    To,
+}
+
+impl CVoidCastDirection {
+    /// For applicable [`Callee`]s that take or return [`*c_void`](core::ffi::c_void),
+    /// return the [`CVoidCastDirection`]s that they use for casts.
+    ///
+    /// That is, these [`Callee`]s are [`CVoidCastDirection::From`]:
+    /// * [`Callee::Malloc`]
+    /// * [`Callee::Calloc`]
+    /// * [`Callee::Realloc`]
+    ///
+    /// And these [`Callee`]s are [`CVoidCastDirection::To`]:
+    /// * [`Callee::Free`]
+    /// * [`Callee::Realloc`]
+    pub fn from_callee(callee: Option<Callee>) -> &'static [Self] {
+        let callee = match callee {
+            None => return &[],
+            Some(it) => it,
+        };
+        use CVoidCastDirection::*;
+        use Callee::*;
+        match callee {
+            Malloc | Calloc => &[From][..],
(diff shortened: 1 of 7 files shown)


c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+                        let c_void_ptr = match direction {
+                            From => destination,
+                            To => args[0].place().unwrap(),
+                        };
+                        let c_void_ptr = CVoidPtr::checked(c_void_ptr, &mir.local_decls, tcx);
+                        let get_cast = move |stmt: &Statement<'tcx>| {
+                            c_void_ptr.get_cast_from_stmt(direction, stmt)
+                        };
+                        let cast = match direction {
+                            From => find_first_cast_succ_block(get_cast),
```
aneksteind, line 361: I'm wondering if it's worth keeping the casted and castee in each of `From` and `To` so that things like `args[0].place().unwrap()` only need to be specified in their construction. WDYT?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub fn checked_by_eq(&self, place: Place<'tcx>) -> Result<Self, Place<'tcx>> {
+        if self.place == place {
+            Ok(Self { place })
+        } else {
+            Err(place)
+        }
+    }
+}
+
+fn get_cast_place<'tcx>(rv: &Rvalue<'tcx>) -> Option<Place<'tcx>> {
```
aneksteind, line 101: i think functions like these belong in their own non-cvoid-related place or module

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+    match rv {
+        Rvalue::Cast(_, op, _) => op.place(),
+        _ => None,
+    }
+}
+
+fn get_assign_sides<'tcx, 'a>(
```
aneksteind, line 108: ditto own module

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+    stmt: &'a Statement<'tcx>,
+) -> Option<(Place<'tcx>, &'a Rvalue<'tcx>)> {
+    let (pl, ref rv) = match &stmt.kind {
+        StatementKind::Assign(it) => Some(&**it),
+        _ => None,
+    }?;
+    Some((*pl, rv))
+}
+
+impl<'tcx> CVoidPtr<'tcx> {
```
aneksteind, line 118: why have a separate `impl` block for these methods?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+                    let find_last_cast_curr_block =
+                        |get_cast| bb_data.statements.iter().rev().find_map(get_cast);
+
+                    for direction in CVoidCastDirection::from_callee(ty_callee(tcx, func_ty))
+                        .iter()
+                        .copied()
+                    {
+                        use CVoidCastDirection::*;
+                        let c_void_ptr = match direction {
+                            From => destination,
```
aneksteind, line 353: if this means it's casted from a void ptr, then intuitively i'd expect the destination to be the non-void place. but here the destination is the pointer. should `From` and `To` be swapped?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    pub fn new(mir: &Body<'tcx>, tcx: TyCtxt<'tcx>) -> Self {
+        let mut this = Self::default();
+        this.insert_all_from_mir(mir, tcx);
```
aneksteind, line 375: why keep `insert_all_from_mir`? that function also wouldn't need to stay public afaict

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -0,0 +1,378 @@
...
+impl<'tcx> CVoidCastsUniDirectional<'tcx> {
+    pub fn contains(&self, place: Place<'tcx>) -> bool {
+        self.0.contains_key(&place)
+    }
+
+    /// Get the adjusted [`Place`], skipping over [`*c_void`](core::ffi::c_void) intermediaries.
+    ///
+    /// That is, if `place` is a [`CVoidPtr`] in this map of [`CVoidCast`]s,
+    /// then the [`Place`] of its other, property-typed pointer is returned.
+    /// Otherwise, the same `place` is returned, as no adjustmenets are necessary.
```
aneksteind, line 192: `adjustments`

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -299,24 +299,28 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                     Some(Callee::Realloc) => {
-                        let destination = self.acx.c_void_casts.get_or_default_to(&destination);
-                        let input_ptr_pl = args[0].place().unwrap();
-                        let first_arg = self.acx.c_void_casts.get_or_default_to(&input_ptr_pl);
-                        self.visit_place(destination, Mutability::Mut);
-                        let pl_lty = self.acx.type_of(destination);
+                        let out_ptr = self
+                            .acx
+                            .c_void_casts
+                            .get_adjusted_place(CVoidCastDirection::From, destination);
```
aneksteind, line 313: this new naming `get_adjusted_place` doesn't make it clear that it will default to the input `place` if there's been more adjustment. consider renaming to make this clear so that one doesn't have to look up the definition

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -299,24 +299,28 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        let in_ptr = self
+                            .acx
+                            .c_void_casts
+                            .get_adjusted_place(CVoidCastDirection::To, in_ptr);
+                        self.visit_place(out_ptr, Mutability::Mut);
+                        let pl_lty = self.acx.type_of(out_ptr);
                         assert!(args.len() == 2);
-                        self.visit_place(input_ptr_pl, Mutability::Not);
-                        let rv_lty = self.acx.type_of(first_arg);
+                        self.visit_place(in_ptr, Mutability::Not);
```
aneksteind, line 322: i think the semantics of this have changed. in the previous version, the original argument to realloc was visited, and here it seems to be the adjustment that's visited. is this intentional? if not, check to see if a similar change exists for the other callees. something about this feels more correct -- we're avoiding visiting a place that will cease to be relevant given that we're skipping the cast. but i wonder if we risk traversing the adjustment, the new place, superfluously and if that has any unintended side effects

c2rust-analyze/src/c_void_casts.rs
kkysen: I'll put them in `util`.

c2rust-analyze/src/c_void_casts.rs
kkysen: It was just so the order of those helper methods were close to their use site.  But I'll move them to `util` so it won't be a problem anymore.

c2rust-analyze/src/c_void_casts.rs
kkysen: I matched the previous code,

```rust
Some(Malloc | Calloc) => {
    handle_c_void_cast(&destination, &find_first_cast_succ_block);
}
```

`malloc` and `calloc` are from `*c_void`s, so I turned that code into the current code.

It seems like the destination should be non-`*c_void` here, though.  Yet when I try to flip them, it crashes in `CVoidPtr::checked` where it checks if it actually is a `*c_void`.

I think this `destination` is the destination of the `malloc`, etc. call, so what it returns, `*c_void`, not the destination of the cast.  In the flipped direction, the first arg of `free`, etc. is the `*c_void`.

c2rust-analyze/src/c_void_casts.rs
kkysen: Fixed in 278df7d9361f7d9815104deca07351c5304e3a03 (`rebase` `fixup`).

c2rust-analyze/src/c_void_casts.rs
kkysen: Fixed in 707f6c020d381f1e017a548725138571a6f4afae.

c2rust-analyze/src/c_void_casts.rs
kkysen: Fixed in 707f6c020d381f1e017a548725138571a6f4afae.

c2rust-analyze/src/c_void_casts.rs
kkysen: Fixed in 707f6c020d381f1e017a548725138571a6f4afae.

c2rust-analyze/src/c_void_casts.rs
kkysen: Fixed in 4326435687e15f1e266f52545b78a4daed6166fb.

I had thought `insert_all_from_mir` could be callable outside of `new`, but since `Place`s are local to a func `Body`, that shouldn't happen, so now `CVoidCasts::new` is the only way to create and modify a `CVoidCasts`.

I also renamed `mir` to `body`, as I'm not sure why we just call a `Body` `mir`.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: What I meant by `adjusted` is that the post-`*c_void` fixes is the adjusted `Place`, whether or not it has to be adjusted.  It's in the documentation for that function.  If you have a name that better indicates that, I can change it, but I'm not sure what to name it to otherwise.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -326,12 +330,15 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                     }
                     Some(Callee::Free) => {
-                        let input_ptr_pl = args[0].place().unwrap();
-                        let first_arg = self.acx.c_void_casts.get_or_default_to(&input_ptr_pl);
+                        let in_ptr = args[0].place().unwrap();
+                        let in_ptr = self
+                            .acx
+                            .c_void_casts
+                            .get_adjusted_place(CVoidCastDirection::To, in_ptr);
                         self.visit_place(destination, Mutability::Mut);
```
kkysen, line 338: Following up on https://github.com/immunant/c2rust/pull/826/files#r1105146453:

Here, for `free`, `self.visit_place` is called on `destination`, while for `realloc`, it's called on `input_ptr_pl` (the non-adjusted one), while I changed it to `in_ptr` (the adjusted one).  Is this correct?

c2rust-analyze/src/dataflow/type_check.rs
kkysen: You're right, I accidentally changed it.  But I was under the impression they should be the same (i.e. I meant to do this, but the change was accidental; I didn't realize it was different before).

See https://github.com/immunant/c2rust/pull/826/files#r1106252674 also.

c2rust-analyze/src/c_void_casts.rs
aneksteind: > I think this destination is the destination of the malloc, etc. call, so what it returns, *c_void, not the destination of the cast. In the flipped direction, the first arg of free, etc. is the *c_void.

Right, I conveniently forgot this

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: i think appending `_or_default_to` or something similar would make it much clearer

kkysen: I can switch that if you think it's a lot clearer, though I thought it was clear enough already.  The adjustment can be zero/nothing.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: let's revive the old semantics until we determine that a change to them is desired/necessary

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: the old code for `Realloc` also visited the `destination`: https://github.com/immunant/c2rust/pull/826/files#diff-59cb8a15b14c434fe5d4773bb9ef18f70ab3fa8a6022fe424df3b9e0a6d680cbL315. `Realloc` should be visiting both it's argument(s) and its destination. Both `destination` and `input_ptr_pl` are pre-adjustment and so it's those semantics I'd like to preserve. Now in this PR it's post-adjustment places that are visited in `Realloc` but pre-adjustment `Places` that are visited in `Free`. I think if anything we should be visiting the argument to `free`, as I don't see that.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: i do think it's a lot clearer

c2rust-analyze/src/c_void_casts.rs
aneksteind: on second thought, it's too minor a point to dwell over

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Shouldn't `free` call it on the argument, not the destination?  `free` doesn't return anything.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Fixed in eb0840379814d745288f4ac165788c67b4c1ef64.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: visiting the destination isn't harmful afaik, and it does have one. we should also be visiting the argument

aneksteind (approved):

Issue #793, opened 2023-01-24
static analysis erroneously wants to visit MIR bodies of extern function calls


aneksteind:
Running static analysis on the following results in an exception

```rust
extern "C" {
    // ...
    fn foreign(r: *mut bar);
    // ...
}

unsafe extern "C" fn func(mut b: *mut bar) {
    // ...
    foreign(b);
    // ...
}
```

```
thread 'rustc' panicked at 'not yet implemented: call to unknown function DefId(0:66 ~ main[5047]::{extern#0}::foreign)', c2rust-analyze/src/dataflow/type_check.rs:353:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
``` 

In [`visit_call_other`](https://github.com/immunant/c2rust/blob/212dfa970fbd760db963e0eabc8b8ee5d1eb296b/c2rust-analyze/src/dataflow/type_check.rs#L345) we might at least like to iterate over the function signature to relate pointer permissions.

kkysen:
> we might at least like to iterate over the function signature to relate pointer permissions.

What do you mean by relating pointer permissions using only the signature?

aneksteind:
> What do you mean by relating pointer permissions using only the signature?

E.g. when there is a function call there is a terminator with `_dest = call(_arg1, ..., _argN)` and static analysis calls `do_assign` for those [arguments](https://github.com/immunant/c2rust/blob/212dfa970fbd760db963e0eabc8b8ee5d1eb296b/c2rust-analyze/src/dataflow/type_check.rs#L364) and for the [output](https://github.com/immunant/c2rust/blob/212dfa970fbd760db963e0eabc8b8ee5d1eb296b/c2rust-analyze/src/dataflow/type_check.rs#L371) of the call. We miss out on those relationships if the call doesn't have an MIR body to traverse, as things currently stand.

kkysen:
Split into:
* #842
* #843

PR #830, opened 2023-02-13
(`c2rust-analyze`) Support `Repeat` array initializers with no pointers

aneksteind:
`Repeat` initialization shows up a bunch in `lighttpd` code, e.g.

```rust
...
let mut x: [uint32_t; 16] = [0; 16];
...
```

As far as I can tell, though, no `Repeat` initialization in `lighttpd` includes arrays of pointer types. So, this PR minimally adds support for `Repeat` initialization for non-pointer types

### DIFF
--- a/analysis/tests/lighttpd-minimal/src/main.rs
+++ b/analysis/tests/lighttpd-minimal/src/main.rs
@@ -1369,6 +1369,22 @@ pub unsafe extern "C" fn lighttpd_test() {
     free(fdes /* TODO: handle cast as *mut libc::c_void */);
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn buffer_append_int(mut b: *mut buffer, mut val: intmax_t) {
+    let mut buf: [libc::c_char; 22] = [0; 22];
+    // let str: *const libc::c_char = itostr(buf.as_mut_ptr(), val);
+    // buffer_append_string_len(
+    //     b,
+    //     str,
+    //     buf
+    //         .as_mut_ptr()
+    //         .offset(
+    //             ::std::mem::size_of::<[libc::c_char; 22]>() as libc::c_ulong as isize,
+    //         )
+    //         .offset_from(str) as libc::c_long as size_t,
+    // );
+}
+
 fn main() {
     unsafe {
         lighttpd_test();
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -299,6 +299,15 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
             Rvalue::UnaryOp(_, ref op) => self.visit_operand(op),
 
+            Rvalue::Repeat(ref op, _) => {
+                if op.ty(self.local_decls, self.tcx).is_any_ptr() {
+                    todo!("Repeat types over pointers not yet implemented");
+                }
+                let ty = rv.ty(self.local_decls, *self.ltcx);
+                // TODO: create fresh origins for all pointers in `ty`, and generate subset
+                // relations between the regions of the array and the regions of its elements
+                self.ltcx.label(ty, &mut |_ty| Label::default())
+            }
             ref rv => panic!("unsupported rvalue {:?}", rv),
         }
     }
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -115,6 +115,18 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 assert_matches!(lty.args, [elem_lty] => {
                     // Pseudo-assign from the operand to the element type of the array.
                     let op_lty = self.acx.type_of(op);
+                    /*
+                        TODO: This is the right thing to do here, though currently
+                        it's a no-op because type_of_rvalue (whose result is passed
+                        into this function as lty) constructs its result using
+                        type_of(op), so elem_lty and op_lty will always be identical.
+                        Eventually we'll want to change this so the Rvalue::Repeat gets
+                        its own LTy with its own PointerIds, similar to the handling of
+                        array aggregates. This would let us detect the need for a
+                        cast on the operand of the repeat (we can't cast [&[u32]; 3]
+                        to [&u32; 3], but we could cast the operand from &[u32] to
+                        &u32 before doing the repeat, e.g. let x = [&my_slice[0]; 3];).
+                    */
                     self.do_assign(elem_lty, op_lty);
                 });
             }
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -1,6 +1,7 @@
 use super::DataflowConstraints;
 use crate::context::{AnalysisCtxt, LTy, PermissionSet, PointerId};
 use crate::util::{self, describe_rvalue, Callee, RvalueDesc};
+use assert_matches::assert_matches;
 use rustc_hir::def_id::DefId;
 use rustc_middle::mir::{
     AggregateKind, BinOp, Body, Location, Mutability, Operand, Place, PlaceRef, ProjectionElem,
@@ -111,11 +112,11 @@ impl<'tcx> TypeChecker<'tcx, '_> {
             Rvalue::Use(ref op) => self.visit_operand(op),
             Rvalue::Repeat(ref op, _) => {
                 assert!(matches!(lty.kind(), TyKind::Array(..)));
-                assert_eq!(lty.args.len(), 1);
-                let elem_lty = lty.args[0];
-                // Pseudo-assign from the operand to the element type of the array.
-                let op_lty = self.acx.type_of(op);
-                self.do_assign(elem_lty, op_lty);
+                assert_matches!(lty.args, [elem_lty] => {
+                    // Pseudo-assign from the operand to the element type of the array.
+                    let op_lty = self.acx.type_of(op);
+                    self.do_assign(elem_lty, op_lty);
+                });
             }
             Rvalue::Ref(..) => {
                 unreachable!("Rvalue::Ref should be handled by describe_rvalue instead")
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -109,7 +109,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
         match *rv {
             Rvalue::Use(ref op) => self.visit_operand(op),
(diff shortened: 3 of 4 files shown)


c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -109,7 +109,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
         match *rv {
             Rvalue::Use(ref op) => self.visit_operand(op),
-            Rvalue::Repeat(..) => todo!("visit_rvalue Repeat"),
+            Rvalue::Repeat(ref op, _) => {
+                assert!(matches!(lty.kind(), TyKind::Array(..)));
```
kkysen, line 113: There's a `Ty::is_array` method.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -299,6 +299,15 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
             Rvalue::UnaryOp(_, ref op) => self.visit_operand(op),
 
+            Rvalue::Repeat(ref op, _) => {
+                if op.ty(self.local_decls, self.tcx).is_any_ptr() {
```
kkysen, line 303: I think we have to be careful about what we consider pointers throughout `c2rust-analyze`.

- `Ty::is_any_ptr` checks `Ref`, `RawPtr`, and `FnPtr`
- `assign_pointer_ids` checks `Ref` and `RawPtr` manually
- and `is_trivial` checks just `RawPtr`

aneksteind: I think that's fair, but here we simply just don't want any of `Ref`, `RawPtr`, or `FnPtr`.

c2rust-analyze/tests/filecheck/aggregate1.rs
```diff
@@ -14,3 +14,8 @@ pub unsafe fn aggregate1_array1(p: *mut i32) {
     let arr = [p];
     *arr[0] = 1;
 }
+
+pub unsafe fn repeat() {
+    let x = 22;
+    let _buf: [u32; 22] = [x; 22];
+}
```
kkysen, line 21: Can we put this in a new `aggregate_initializer.rs` file?

aneksteind: For what purpose?

analysis/tests/lighttpd-minimal/src/main.rs
```diff
@@ -1369,6 +1369,22 @@ pub unsafe extern "C" fn lighttpd_test() {
     free(fdes /* TODO: handle cast as *mut libc::c_void */);
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn buffer_append_int(mut b: *mut buffer, mut val: intmax_t) {
+    let mut buf: [libc::c_char; 22] = [0; 22];
+    // let str: *const libc::c_char = itostr(buf.as_mut_ptr(), val);
```
kkysen, line 1375: What's `itostr`?  As in, where is it defined?  It doesn't seem to be a standard C function.  Do we need to add it's declaration in the `extern` block, like there is for `buffer_append_string_len` below`.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -109,7 +109,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                assert_eq!(lty.args.len(), 1);
+                let elem_lty = lty.args[0];
```
kkysen, line 115: nit: I've noticed this in general.  I think we can do something like an `assert_matches!` here and destructure the array instead.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: wonderful, will use that

analysis/tests/lighttpd-minimal/src/main.rs
aneksteind: it's defined in `buffer.rs` in `lighttpd`, and can be ignored for now. Once https://github.com/immunant/c2rust/issues/793 is addressed we would not need to include the definition here.

kkysen (changes requested):
See above comments.

Also, I'm wondering if we can extend this to all aggregate initializers with the same no-pointer restriction.  It might not make much of a difference in the `lighttpd` codebase as most `struct` initializations have pointers, but just a thought.

analysis/tests/lighttpd-minimal/src/main.rs
kkysen: We're always going to need to include the declaration.

c2rust-analyze/tests/filecheck/aggregate1.rs
kkysen: Smaller tests are generally better, right?  It's testing something different here, and one could fail while the other doesn't.

aneksteind:
I don't think Aggregates are worth considering in this PR. I think we should file an issue for that. But this unblocks a sizeable set of `lighttpd` code. I believe we should focus on correctness of the implementation first and towards enabling more `lighttpd` code analysis.

kkysen:
> I don't think Aggregates are worth considering in this PR. I think we should file an issue for that. But this unblocks a sizeable set of `lighttpd` code. I believe we should focus on correctness of the implementation first and towards enabling more `lighttpd` code analysis.

I can split those off into issues.

There are a lot of `None`s in `lighttpd`, which are aggregates.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: 831aad17

analysis/tests/lighttpd-minimal/src/main.rs
aneksteind: What's the thing you'd like to see changed?

c2rust-analyze/tests/filecheck/aggregate1.rs
aneksteind: Maybe. The nature of test failure won't change much having it here vs. in another file. And aggregate / repeat initialization is not substantially different from the other test cases in this file; I think it definitely fits in here.

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: 681e55f6

c2rust-analyze/src/borrowck/type_check.rs
kkysen: This should be fine for this PR, but I think it's something we should discuss at some point to clarify each of the definitions.

analysis/tests/lighttpd-minimal/src/main.rs
kkysen: Looking at the definition, https://github.com/immunant/lighttpd-rust/blob/4aec35e15309d6d66d9104f29d5d5e2b6b6f5c46/src/buffer.c#L303, it seems there's only a definition, no separate declaration, which I assumed there would be (like there are from headers).  I think it's fine to just leave things as is then.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -109,7 +110,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
         match *rv {
             Rvalue::Use(ref op) => self.visit_operand(op),
-            Rvalue::Repeat(..) => todo!("visit_rvalue Repeat"),
+            Rvalue::Repeat(ref op, _) => {
+                assert!(lty.ty.is_array());
+                assert_matches!(lty.args, [elem_lty] => {
+                    // Pseudo-assign from the operand to the element type of the array.
+                    let op_lty = self.acx.type_of(op);
+                    self.do_assign(elem_lty, op_lty);
```
spernsteiner, line 118: This is the right thing to do here, though currently it's a no-op because `type_of_rvalue` (whose result is passed into this function as `lty`) constructs its result using `type_of(op)`, so `elem_lty` and `op_lty` will always be identical.  Eventually we'll want to change this so the `Rvalue::Repeat` gets its own `LTy` with its own `PointerId`s, similar to the handling of array aggregates.  This would let us detect the need for a cast on the operand of the repeat (we can't cast `[&[u32]; 3]` to `[&u32; 3]`, but we could cast the operand from `&[u32]` to `&u32` before doing the repeat, e.g. `let x = [&my_slice[0]; 3];`).

spernsteiner (approved):

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -109,7 +110,26 @@ impl<'tcx> TypeChecker<'tcx, '_> {
 
         match *rv {
             Rvalue::Use(ref op) => self.visit_operand(op),
-            Rvalue::Repeat(..) => todo!("visit_rvalue Repeat"),
+            Rvalue::Repeat(ref op, _) => {
+                assert!(lty.ty.is_array());
```
kkysen, line 114: This is always the case, right?  So this is just a sanity check?

analysis/tests/lighttpd-minimal/src/main.rs
```diff
@@ -1369,6 +1369,22 @@ pub unsafe extern "C" fn lighttpd_test() {
...
+    // buffer_append_string_len(
+    //     b,
+    //     str,
+    //     buf
+    //         .as_mut_ptr()
+    //         .offset(
+    //             ::std::mem::size_of::<[libc::c_char; 22]>() as libc::c_ulong as isize,
+    //         )
+    //         .offset_from(str) as libc::c_long as size_t,
```
kkysen, line 1384: Can this part of the call be uncommented?

c2rust-analyze/tests/filecheck/aggregate1.rs
kkysen: I'd still put this into an `analyze`-only test file.  It's a lot easier to test small things individually.  I guess it doesn't matter so much, but I'd prefer it like this and it's a trivial change.

Also, it doesn't need to be `unsafe`, right?

kkysen (approved):

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: Yes it's just making the same check that's made in the `Aggregate`/`Array` case

analysis/tests/lighttpd-minimal/src/main.rs
aneksteind: I'm leaving it commented until https://github.com/immunant/c2rust/issues/842 is fixed to address `itostr`.

PR #835, opened 2023-02-14
(`c2rust-analyze/tests`) Define `FileCheck` tests statically, rather than searching for them dynamically

kkysen:
Define `FileCheck` tests statically, rather than searching for them dynamically.

This has a few major advantages:

- You can run a single test if you want. This is really useful when you're developing and testing one test, and you have no need to run the other tests yet, which are just slower and add to extra output.

- The tests are automatically parallelized, which makes them take ~0.5s vs. ~2s, 4x faster.

- You can add some custom logic for one particular test, for example, if it crashes, or you want to skip it for now as it's a WIP.

The macros `define_test!` and `define_tests!` are added here as a convenience for quickly and succinctly defining all the tests, but they don't have to be used for, e.x. a custom test.

I've also now added `check_for_missing_tests` tests than ensure we don't have test files that aren't being tested by a `#[test] fn`.

### DIFF
--- a/c2rust-analyze/tests/filecheck.rs
+++ b/c2rust-analyze/tests/filecheck.rs
@@ -1,30 +1,49 @@
 pub mod common;
 
-use std::fs;
+use std::path::PathBuf;
 
 use common::{Analyze, FileCheck};
 
-#[test]
-fn filecheck() {
+fn test(file_name: &str) {
     let analyze = Analyze::resolve();
     let file_check = FileCheck::resolve();
+    let path = ["tests", "filecheck", file_name]
+        .iter()
+        .collect::<PathBuf>();
+    let output_path = analyze.run(&path);
+    file_check.run(&path, &output_path);
+}
 
-    for entry in fs::read_dir("tests/filecheck").unwrap() {
-        let entry = entry.unwrap();
-
-        if !entry.file_type().unwrap().is_file() {
-            continue;
+macro_rules! define_test {
+    ($name:ident) => {
+        #[test]
+        fn $name() {
+            test(concat!(stringify!($name), ".rs"));
         }
+    };
+}
 
-        let name = entry.file_name();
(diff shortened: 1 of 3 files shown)


c2rust-analyze/tests/filecheck.rs
```diff
@@ -1,30 +1,47 @@
...
+use std::path::PathBuf;
 
 use common::{Analyze, FileCheck};
 
-#[test]
-fn filecheck() {
+fn test(file_name: &str) {
     let analyze = Analyze::new();
     let file_check = FileCheck::resolve();
+    let path = ["tests", "filecheck", file_name]
```
aneksteind, line 10: nit: I think `PathBuf::from` would be cleaner

aneksteind:
What is kind of unsatisfactory about this is that now adding a test involves two steps (adding the file and specifying it here) rather than just the one step of adding the test to the directory. It's also silent in that, if someone forgets to add their test to this macro invocation, nothing complains. I'm happy about the speedup and making the tests individually runnable, though. Is there a way to make this complain if a user forgets the step of adding the test to the macro invocation?

c2rust-analyze/tests/filecheck.rs
kkysen: It adds an extra allocation, which is why I didn't use it.

aneksteind:
I've had to add more filechecks tests recently and would have definitely forgotten to add the mention of those files. i think there needs to be a sanity check to make sure all files in the filecheck directory are listed and panic otherwise

kkysen:
@aneksteind, I figured out how to add a test that checks for any missing tests, so now it'll be impossible to forget to add new tests, and we'll still get all the other benefits.

aneksteind (approved):

spernsteiner (approved):

Issue #833, opened 2023-02-14
(`c2rust-analyze`) Handle transpiled string literals (`b"...\0" as *const u8 as *const libc::c_char`)


kkysen:
Currently, `c2rust-analyze` can't handle `c2rust transpile`d string literals.  String literals like `"literal"` in C become `b"literal\0" as *const u8 as *const libc::c_char` in Rust, but the ptr-to-ptr cast of different pointer types is not handled, and this happens:

```rust
b"\0" as *const u8 as *const libc::c_char;
```

```shell
visit_statement(_19 = const b"\x00")
thread 'rustc' panicked at 'unexpected pointer type in &[u8; 1]', c2rust-analyze/src/context.rs:503:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/library/core/src/panicking.rs:142:14
   2: c2rust_analyze::context::label_no_pointers::{{closure}}
             at ./src/context.rs:503:9
   3: c2rust_analyze::labeled_ty::LabeledTyCtxt<L>::label
             at ./src/labeled_ty.rs:157:21
   4: c2rust_analyze::context::label_no_pointers
             at ./src/context.rs:502:5
   5: <rustc_middle::mir::syntax::Operand as c2rust_analyze::context::TypeOf>::type_of
             at ./src/context.rs:494:41
   6: <&T as c2rust_analyze::context::TypeOf>::type_of
             at ./src/context.rs:464:9
   7: c2rust_analyze::context::AnalysisCtxt::type_of
             at ./src/context.rs:241:9
   8: c2rust_analyze::context::AnalysisCtxt::type_of_rvalue
             at ./src/context.rs:304:36
   9: c2rust_analyze::dataflow::type_check::TypeChecker::visit_statement
             at ./src/dataflow/type_check.rs:233:30
  10: c2rust_analyze::dataflow::type_check::visit
             at ./src/dataflow/type_check.rs:388:13
  11: c2rust_analyze::dataflow::generate_constraints
             at ./src/dataflow/mod.rs:326:5
  12: c2rust_analyze::run
             at ./src/main.rs:480:45
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

[...]

PR #838, opened 2023-02-16
(`c2rust-analyze/tests`) Add tests for string literals and casts

kkysen:
Add tests for string literals and casts.

They are currently unsupported:
- #833
- #837

so for now the tests are skipped over with `#[cfg(any())]`.  Once those issues are fixed, these tests will be turned back on, but it's easier to start with the tests existing.

These tests just check if `c2rust-analyze` doesn't crash on them, similar to the existing `lighttpd-minimal` test.  For the `FileCheck` tests, `FileCheck` requires at least one `CHECK:` command, which we don't want.  Thus, I've renamed `lighttpd.rs` to `analyze.rs` and it'll be for `c2rust-analyze`-only tests, as opposed to `filecheck.rs` for `c2rust-analyze` + `FileCheck` tests.

Furthermore, running `cargo` concurrently (due to multiple tests) for `c2rust-analyze` crashes in macOS CI, so now the `c2rust-analyze` binary is found to run directly, rather than going through `cargo` again.

Skipping going through `cargo` makes the tests run much faster, but more importantly, prevents them from competing for the `target/` lock, which caused issues in CI on macOS sometimes.  It's also a bit simpler now, not needing to go through `cargo`.

### DIFF
--- a/c2rust-analyze/tests/analyze.rs
+++ b/c2rust-analyze/tests/analyze.rs
@@ -4,15 +4,15 @@ pub mod common;
 
 #[test]
 fn string_literals() {
-    Analyze::new().run("tests/analyze/string_literals.rs");
+    Analyze::resolve().run("tests/analyze/string_literals.rs");
 }
 
 #[test]
 fn string_casts() {
-    Analyze::new().run("tests/analyze/string_casts.rs");
+    Analyze::resolve().run("tests/analyze/string_casts.rs");
 }
 
 #[test]
 fn lighttpd_minimal() {
-    Analyze::new().run("../analysis/tests/lighttpd-minimal/src/main.rs");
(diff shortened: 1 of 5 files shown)


c2rust-analyze/tests/analyze.rs
```diff
@@ -0,0 +1,18 @@
+use common::Analyze;
+
+pub mod common;
+
+#[test]
+fn string_literals() {
+    Analyze::new().run("tests/analyze/string_literals.rs");
```
aneksteind, line 7: could you try making these filecheck tests instead, given the CI failure?

kkysen: Per our discussion, we decided we'd avoid running through `cargo` to fix the CI failure, a simplification.  Then when we need `FileCheck` tests for these, we'll move them to `filecheck/` and add the directives.

aneksteind (approved):

PR #839, opened 2023-02-16
(`c2rust-analyze`) Support ptr-to-ptr casts between safely transmutable types, for now limited to same-sized integers

### LINKED ISSUE #833: (`c2rust-analyze`) Handle transpiled string literals (`b"...\0" as *const u8 as *const libc::c_char`)

### LINKED ISSUE #840: (`c2rust-analyze`) Handle compatible ptr-to-ptr casts
Currently, `c2rust-analyze` can't handle ptr-to-ptr casts when the pointee types are different, even when they're compatible/safely transmutable like `u8` and `i8`/`c_char`.  Thus, fixing this is necessary for `c2rust transpile`d string literal support (#833).

```rust
fn cast_only(s: *const u8) {
    s as *const core::ffi::c_char;
}
```

```shell
callgraph traversal order:
  DefId(0:3 ~ string_casts[c5c4]::cast_only)
visit_statement(StorageLive(_2))
visit_statement(StorageLive(_3))
visit_statement(_3 = _1)
rvalue = _1, desc = Project { base: PlaceRef { local: _1, projection: [] }, proj: [] }, base_lty = g0#*const u8[NONE#u8[]]
visit_rvalue(_1), desc = Some(Project { base: PlaceRef { local: _1, projection: [] }, proj: [] })
equate NONE#u8[] = NONE#u8[]
visit_statement(_2 = move _3 as *const i8 (Misc))
thread 'rustc' panicked at 'unexpected pointer type in *const i8', c2rust-analyze/src/context.rs:503:9
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: c2rust_analyze::context::label_no_pointers::{{closure}}
             at ./src/context.rs:503:9
   3: c2rust_analyze::labeled_ty::LabeledTyCtxt<L>::label
             at ./src/labeled_ty.rs:157:21
   4: c2rust_analyze::context::label_no_pointers
             at ./src/context.rs:502:5
   5: c2rust_analyze::context::AnalysisCtxt::type_of_rvalue
             at ./src/context.rs:368:17
   6: c2rust_analyze::dataflow::type_check::TypeChecker::visit_statement
             at ./src/dataflow/type_check.rs:233:30
   7: c2rust_analyze::dataflow::type_check::visit
             at ./src/dataflow/type_check.rs:388:13
   8: c2rust_analyze::dataflow::generate_constraints
             at ./src/dataflow/mod.rs:326:5
   9: c2rust_analyze::run
             at ./src/main.rs:480:45
```

kkysen:
- Fixes #840

This introduces the concept of equivalent/compatible/safely transmutable types: https://github.com/immunant/c2rust/blob/2915b8d0c71add21dee1f9d540958ea863478212/c2rust-analyze/src/util.rs#L356-L380

Thus, we can now allow ptr-to-ptr casts between safely transmutable pointee types, whereas previously they were only allowed for equal types.  In particular, this enables support for string casts, which are produced by `c2rust transpile` as `b"" as *const u8 as *const core::ffi::c_char`, where `c_char = i8`.  Thus, this fixes #840.

New tests are added in `string_casts.rs` to cover various ptr casts, though some of them crash in the rewriter due to having implicitly inserted MIR statements like implicit `&raw`s, which are inserted with `addr_of!`s.  Thus, for some of these (where it works), there are versions with explicit `addr_of!`s that succeed end-to-end.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -3,7 +3,7 @@ use crate::pointer_id::{
     GlobalPointerTable, LocalPointerTable, NextGlobalPointerId, NextLocalPointerId, PointerTable,
     PointerTableMut,
 };
-use crate::util::{self, are_transmutable_ptrs, describe_rvalue, RvalueDesc};
+use crate::util::{self, describe_rvalue, is_transmutable_ptr_cast, RvalueDesc};
 use crate::AssignPointerIds;
 use bitflags::bitflags;
 use rustc_hir::def_id::DefId;
@@ -358,7 +358,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
                 // In particular, this allows casts from `*u8` to `*core::ffi::c_char`.
                 let from_ty = op_lty.ty;
                 let to_ty = ty;
-                match are_transmutable_ptrs(from_ty, to_ty) {
+                match is_transmutable_ptr_cast(from_ty, to_ty) {
                     // Label the to type with the same [`PointerId`]s as the from type in all positions.
                     // This works because the two types have the same structure.
                     Some(true) => self.lcx().mk(ty, op_lty.args, op_lty.label),
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -3,7 +3,7 @@ use crate::pointer_id::{
     GlobalPointerTable, LocalPointerTable, NextGlobalPointerId, NextLocalPointerId, PointerTable,
     PointerTableMut,
 };
-use crate::util::{self, describe_rvalue, RvalueDesc};
+use crate::util::{self, are_transmutable_ptrs, describe_rvalue, RvalueDesc};
 use crate::AssignPointerIds;
 use bitflags::bitflags;
 use rustc_hir::def_id::DefId;
@@ -349,23 +349,22 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             Rvalue::Cast(_, ref op, ty) => {
                 let op_lty = self.type_of(op);
 
-                // We support this category of pointer casts as a special case.
-                let op_is_ptr = matches!(op_lty.ty.kind(), TyKind::Ref(..) | TyKind::RawPtr(..));
-                let op_pointee = op_is_ptr.then(|| op_lty.args[0]);
-                let ty_pointee = match *ty.kind() {
-                    TyKind::Ref(_, ty, _) => Some(ty),
-                    TyKind::RawPtr(tm) => Some(tm.ty),
-                    _ => None,
-                };
-                if op_pointee.is_some() && op_pointee.map(|lty| lty.ty) == ty_pointee {
-                    // The source and target types are both pointers, and they have identical
-                    // pointee types.  We label the target type with the same `PointerId`s as the
-                    // source type in all positions.  This works because the two types have the
-                    // same structure.
-                    return self.lcx().mk(ty, op_lty.args, op_lty.label);
+                // We only support pointer casts when:
+                // * both types are pointers
+                // * they have compatible (safely transmutable) pointee types
+                // Safe transmutability is difficult to check abstractly,
+                // so we limit it to integer types of the same size
+                // (but potentially different signedness).
+                // In particular, this allows casts from `*u8` to `*core::ffi::c_char`.
+                let from_ty = op_lty.ty;
+                let to_ty = ty;
+                match are_transmutable_ptrs(from_ty, to_ty) {
+                    // Label the to type with the same [`PointerId`]s as the from type in all positions.
+                    // This works because the two types have the same structure.
+                    Some(true) => self.lcx().mk(ty, op_lty.args, op_lty.label),
+                    Some(false) => todo!("unsupported ptr-to-ptr cast between pointee types not yet supported as safely transmutable: `{from_ty:?} as {to_ty:?}`"),
+                    None => label_no_pointers(self, ty),
                 }
-
-                label_no_pointers(self, ty)
             }
             Rvalue::Len(..)
             | Rvalue::BinaryOp(..)
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -1,21 +1,32 @@
+use crate::borrowck::{AdtMetadata, FieldMetadata, OriginArg, OriginParam};
 use crate::c_void_casts::CVoidCasts;
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
+use crate::panic_detail::PanicDetail;
 use crate::pointer_id::{
     GlobalPointerTable, LocalPointerTable, NextGlobalPointerId, NextLocalPointerId, PointerTable,
     PointerTableMut,
 };
 use crate::util::{self, describe_rvalue, PhantomLifetime, RvalueDesc};
 use crate::AssignPointerIds;
+use assert_matches::assert_matches;
 use bitflags::bitflags;
-use rustc_hir::def_id::DefId;
+use indexmap::IndexSet;
+use log::*;
+use rustc_ast::Mutability;
+use rustc_hir::def::DefKind;
+use rustc_hir::def_id::{DefId, LocalDefId};
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::interpret::{self, AllocId, ConstValue, GlobalAlloc};
 use rustc_middle::mir::{
     Body, Constant, ConstantKind, Field, HasLocalDecls, Local, LocalDecls, Location, Operand,
     Place, PlaceElem, PlaceRef, Rvalue,
 };
-use rustc_middle::ty::{AdtDef, FieldDef, Ty, TyCtxt, TyKind};
+use rustc_middle::ty::{
+    tls, AdtDef, FieldDef, GenericArgKind, GenericParamDefKind, Ty, TyCtxt, TyKind,
+};
+use rustc_type_ir::RegionKind::{ReEarlyBound, ReStatic};
 use std::collections::HashMap;
+use std::fmt::Debug;
 use std::ops::Index;
 
 bitflags! {
@@ -135,6 +146,12 @@ bitflags! {
         /// way, and it can't be freely discarded (or its inverse freely added) as is the case for
         /// everything in `PermissionSet`.
         const CELL = 0x0001;
+
+        /// This pointer's type is fixed; rewrites must not change it.  This is used for all safe
+        /// references (which we assume already have the correct types), for raw pointers that
+        /// cross an FFI boundary, and for arguments and return values of functions we can't
+        /// rewrite.
+        const FIXED = 0x0002;
     }
 }
 
@@ -149,26 +166,138 @@ pub struct LFnSig<'tcx> {
     pub output: LTy<'tcx>,
 }
 
+bitflags! {
+    /// Basic information about a pointer, computed with minimal analysis before running `dataflow`
+    /// or `borrowck`.
+    ///
+    /// When `PointerId`s are merged into a single `PointerId` per equivalence class, the
+    /// `PointerInfo` of each resulting `PointerId` is the union of the `PointerInfo`s of all the
+    /// members of the class.  Thus, flags describing properties of the declaration that produced a
+    /// given `PointerId` should be formulated as "at least one declaration has property X", as a
+    /// single `PointerId` may correspond to several declarations.
+    #[derive(Default)]
+    pub struct PointerInfo: u16 {
+        /// This `PointerId` was generated for a `TyKind::Ref`.
+        const REF = 0x0001;
+
+        /// At least one declaration that produced this `PointerId` used an explicit type
+        /// annotation.
+        const ANNOTATED = 0x0002;
+
+        /// This `PointerId` has at least one local declaration that is not a temporary reference
+        /// arising from an `&x` or `&mut x` expression in the source.
+        const NOT_TEMPORARY_REF = 0x0004;
+    }
+}
+
+pub struct AdtMetadataTable<'tcx> {
+    pub table: HashMap<DefId, AdtMetadata<'tcx>>,
+    pub struct_dids: Vec<DefId>,
+}
+
+impl<'tcx> Debug for AdtMetadataTable<'tcx> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        fn fmt_string(lty: LabeledTy<'_, &[OriginArg]>) -> String {
+            let args: Vec<String> = lty.args.iter().map(|t| fmt_string(t)).collect();
+            use rustc_type_ir::TyKind::*;
+            match lty.kind() {
+                Ref(..) | RawPtr(..) => {
+                    format!("&{:?} {:}", lty.label[0], args[0])
+                }
+                Adt(adt, _) => {
+                    let mut s = format!("{adt:?}");
+                    let params = lty
+                        .label
+                        .iter()
+                        .map(|p| format!("{:?}", p))
+                        .into_iter()
+                        .chain(args.into_iter())
+                        .collect::<Vec<_>>()
+                        .join(",");
+
+                    if !params.is_empty() {
+                        s.push('<');
+                        s.push_str(&params);
+                        s.push('>');
+                    }
+                    s
+                }
+                Tuple(_) => {
+                    format!("({:})", args.join(","))
+                }
+                _ => format!("{:?}", lty.ty),
+            }
+        }
+
+        tls::with_opt(|tcx| {
+            let tcx = tcx.unwrap();
+            for k in &self.struct_dids {
+                let adt = &self.table[k];
+                let other_param_names = tcx.generics_of(k).params.iter().filter_map(|p| {
+                    if !matches!(p.kind, GenericParamDefKind::Lifetime) {
+                        Some(p.name.to_ident_string())
+                    } else {
+                        None
+                    }
+                });
+                write!(f, "struct {:}", tcx.item_name(*k))?;
+                write!(f, "<")?;
+                let lifetime_params_str = adt
+                    .lifetime_params
+                    .iter()
+                    .map(|p| format!("{:?}", p))
+                    .chain(other_param_names)
+                    .collect::<Vec<_>>()
+                    .join(",");
+                write!(f, "{lifetime_params_str:}")?;
+                writeln!(f, "> {{")?;
+                for (fdid, fmeta) in &adt.field_info {
+                    write!(f, "\t{:}: ", tcx.item_name(*fdid))?;
+                    let field_string_lty = fmt_string(fmeta.origin_args);
+
+                    write!(f, "{field_string_lty:}")?;
+
+                    writeln!(f)?;
+                }
+
+                writeln!(f, "}}\n")?;
+            }
+            writeln!(f)
+        })
+    }
+}
+
 pub struct GlobalAnalysisCtxt<'tcx> {
     pub tcx: TyCtxt<'tcx>,
     pub lcx: LTyCtxt<'tcx>,
 
-    pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
+    ptr_info: GlobalPointerTable<PointerInfo>,
 
-    pub field_tys: HashMap<DefId, LTy<'tcx>>,
+    /// Map from a function to all of its callers.
+    pub fn_callers: HashMap<DefId, Vec<DefId>>,
+
+    pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
+    /// `DefId`s of functions where analysis failed, and a [`PanicDetail`] explaining the reason
+    /// for each failure.
+    pub fns_failed: HashMap<DefId, PanicDetail>,
+
+    pub field_ltys: HashMap<DefId, LTy<'tcx>>,
 
     pub static_tys: HashMap<DefId, LTy<'tcx>>,
+    pub addr_of_static: HashMap<DefId, PointerId>,
 
-    next_ptr_id: NextGlobalPointerId,
+    pub adt_metadata: AdtMetadataTable<'tcx>,
 }
 
 pub struct AnalysisCtxt<'a, 'tcx> {
     pub gacx: &'a mut GlobalAnalysisCtxt<'tcx>,
 
+    ptr_info: LocalPointerTable<PointerInfo>,
+
     pub local_decls: &'a LocalDecls<'tcx>,
     pub local_tys: IndexVec<Local, LTy<'tcx>>,
-    pub c_void_casts: CVoidCasts<'tcx>,
     pub addr_of_local: IndexVec<Local, PointerId>,
+    pub c_void_casts: CVoidCasts<'tcx>,
     /// Types for certain [`Rvalue`]s.  Some `Rvalue`s introduce fresh [`PointerId`]s; to keep
     /// those `PointerId`s consistent, the `Rvalue`'s type must be stored rather than recomputed on
     /// the fly.
@@ -176,8 +305,6 @@ pub struct AnalysisCtxt<'a, 'tcx> {
 
     /// [`Location`]s of (byte-)string literal [`rvalue_tys`](Self::rvalue_tys).
     pub string_literal_locs: Vec<Location>,
-
-    next_ptr_id: NextLocalPointerId,
 }
 
 impl<'a, 'tcx> AnalysisCtxt<'_, 'tcx> {
@@ -209,11 +336,191 @@ impl<'a, 'tcx> AnalysisCtxt<'_, 'tcx> {
 }
 
 pub struct AnalysisCtxtData<'tcx> {
+    ptr_info: LocalPointerTable<PointerInfo>,
     local_tys: IndexVec<Local, LTy<'tcx>>,
     addr_of_local: IndexVec<Local, PointerId>,
+    c_void_casts: CVoidCasts<'tcx>,
     rvalue_tys: HashMap<Location, LTy<'tcx>>,
     string_literal_locs: Vec<Location>,
-    next_ptr_id: NextLocalPointerId,
+}
+
+fn construct_adt_metadata<'tcx>(tcx: TyCtxt<'tcx>) -> AdtMetadataTable {
+    let struct_dids: Vec<_> = tcx
+        .hir_crate_items(())
+        .definitions()
+        .filter_map(|ldid: LocalDefId| {
+            use DefKind::*;
+            let did = ldid.to_def_id();
+            if matches!(tcx.def_kind(did), Struct | Enum | Union) {
+                return Some(did);
+            }
+
+            None
+        })
+        .collect();
+
+    let mut adt_metadata_table = AdtMetadataTable {
+        table: HashMap::new(),
+        struct_dids,
+    };
+
+    // Gather known lifetime parameters for each struct
+    for struct_did in &adt_metadata_table.struct_dids {
+        let struct_ty = tcx.type_of(struct_did);
+        if let TyKind::Adt(adt_def, substs) = struct_ty.kind() {
+            adt_metadata_table
+                .table
+                .insert(adt_def.did(), AdtMetadata::default());
+            eprintln!("gathering known lifetimes for {adt_def:?}");
+            for sub in substs.iter() {
+                if let GenericArgKind::Lifetime(r) = sub.unpack() {
+                    eprintln!("\tfound lifetime {r:?} in {adt_def:?}");
+                    assert_matches!(r.kind(), ReEarlyBound(eb) => {
+                        let _ = adt_metadata_table
+                        .table
+                        .entry(adt_def.did())
+                        .and_modify(|metadata| {
+                            metadata.lifetime_params.insert(OriginParam::Actual(eb));
+                        });
+                    });
+                }
+            }
+        } else {
+            panic!("{struct_ty:?} is not a struct");
+        }
+    }
+
+    let ltcx = LabeledTyCtxt::<'tcx, &[OriginArg<'tcx>]>::new(tcx);
+    let mut loop_count = 0;
+    loop {
+        /*
+            This loop iterates over all structs and gathers metadata for each.
+            If there were no recursive or mutually-recursive data structures,
+            this loop would only need one iteration to complete. To support
+            recursive and mutually-recursive structs, the loop iterates until
+            the metadata gathered for each struct reaches a fixed point.
+        */
+        loop_count += 1;
+        assert!(loop_count < 1000);
+
+        eprintln!("---- running fixed point struct field analysis iteration #{loop_count:?} ----");
+        let old_adt_metadata = adt_metadata_table.table.clone();
+        let mut next_hypo_origin_id = 0;
+
+        // for each struct, gather lifetime information (actual and hypothetical)
+        for struct_did in &adt_metadata_table.struct_dids {
+            let adt_def = tcx.adt_def(struct_did);
+            eprintln!("gathering lifetimes and lifetime parameters for {adt_def:?}");
+            for field in adt_def.all_fields() {
+                let field_ty: Ty = tcx.type_of(field.did);
+                eprintln!("\t{adt_def:?}.{:}", field.name);
+                let field_origin_args = ltcx.label(field_ty, &mut |ty| {
+                    let mut field_origin_args = IndexSet::new();
+                    match ty.kind() {
+                        TyKind::RawPtr(ty) => {
+                            eprintln!(
+                                "\t\tfound pointer that requires hypothetical lifetime: *{:}",
+                                if let Mutability::Mut = ty.mutbl {
+                                    "mut"
+                                } else {
+                                    "const"
+                                }
+                            );
+                            adt_metadata_table
+                                .table
+                                .entry(*struct_did)
+                                .and_modify(|adt| {
+                                    let origin_arg = OriginArg::Hypothetical(next_hypo_origin_id);
+                                    let origin_param =
+                                        OriginParam::Hypothetical(next_hypo_origin_id);
+                                    eprintln!(
+                                        "\t\t\tinserting origin {origin_param:?} into {adt_def:?}"
+                                    );
+
+                                    adt.lifetime_params.insert(origin_param);
+                                    next_hypo_origin_id += 1;
+                                    field_origin_args.insert(origin_arg);
+                                });
+                        }
+                        TyKind::Ref(reg, _ty, _mutability) => {
+                            eprintln!("\t\tfound reference field lifetime: {reg:}");
+                            assert_matches!(reg.kind(), ReEarlyBound(..) | ReStatic);
+                            let origin_arg = OriginArg::Actual(*reg);
+                            adt_metadata_table
+                                .table
+                                .entry(*struct_did)
+                                .and_modify(|adt| {
+                                    if let ReEarlyBound(eb) = reg.kind() {
+                                        eprintln!("\t\t\tinserting origin {eb:?} into {adt_def:?}");
+                                        adt.lifetime_params.insert(OriginParam::Actual(eb));
+                                    }
+
+                                    field_origin_args.insert(origin_arg);
+                                });
+                        }
+                        TyKind::Adt(adt_field, substs) => {
+                            eprintln!("\t\tfound ADT field base type: {adt_field:?}");
+                            for sub in substs.iter() {
+                                if let GenericArgKind::Lifetime(r) = sub.unpack() {
+                                    eprintln!("\tfound field lifetime {r:?} in {adt_def:?}.{adt_field:?}");
+                                    eprintln!("\t\t\tinserting {adt_field:?} lifetime param {r:?} into {adt_def:?}.{:} lifetime parameters", field.name);
+                                    assert_matches!(r.kind(), ReEarlyBound(..) | ReStatic);
+                                    field_origin_args.insert(OriginArg::Actual(r));
+                                }
+                            }
+                            if let Some(adt_field_metadata) =
+                                adt_metadata_table.table.get(&adt_field.did()).cloned()
+                            {
+                                // add a metadata entry for the struct field matching the metadata entry
+                                // for the struct definition of said field
+                                adt_metadata_table
+                                    .table
+                                    .insert(field.did, adt_field_metadata.clone());
+
+                                for adt_field_lifetime_param in adt_field_metadata.lifetime_params.iter() {
+                                    adt_metadata_table.table.entry(*struct_did).and_modify(|adt| {
+                                        if let OriginParam::Hypothetical(h) = adt_field_lifetime_param {
+                                            eprintln!("\t\t\tbubbling {adt_field:?} origin {adt_field_lifetime_param:?} up into {adt_def:?} origins");
+                                            field_origin_args.insert(OriginArg::Hypothetical(*h));
+                                            adt.lifetime_params.insert(*adt_field_lifetime_param);
+                                        }
+                                    });
+                                }
+                            }
+                        }
+                        _ => (),
+                    }
+
+                    if field_origin_args.is_empty() {
+                        return &[];
+                    }
+                    let field_origin_args: Vec<_> = field_origin_args.into_iter().collect();
+                    ltcx.arena().alloc_slice(&field_origin_args[..])
+                });
+
+                adt_metadata_table
+                    .table
+                    .entry(*struct_did)
+                    .and_modify(|adt| {
+                        adt.field_info.insert(
+                            field.did,
+                            FieldMetadata {
+                                origin_args: field_origin_args,
+                            },
+                        );
+                    });
+            }
+
+            eprintln!();
+        }
+
+        if adt_metadata_table.table == old_adt_metadata {
+            eprintln!("reached a fixed point in struct lifetime reconciliation\n");
+            break;
+        }
+    }
+
+    adt_metadata_table
 }
 
 impl<'tcx> GlobalAnalysisCtxt<'tcx> {
@@ -221,10 +528,14 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         GlobalAnalysisCtxt {
             tcx,
             lcx: LabeledTyCtxt::new(tcx),
+            ptr_info: GlobalPointerTable::empty(),
+            fn_callers: HashMap::new(),
             fn_sigs: HashMap::new(),
-            field_tys: HashMap::new(),
+            fns_failed: HashMap::new(),
+            field_ltys: HashMap::new(),
             static_tys: HashMap::new(),
-            next_ptr_id: NextGlobalPointerId::new(),
+            addr_of_static: HashMap::new(),
+            adt_metadata: construct_adt_metadata(tcx),
         }
     }
 
@@ -240,12 +551,16 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         AnalysisCtxt::from_data(self, mir, data)
     }
 
-    pub fn new_pointer(&mut self) -> PointerId {
-        self.next_ptr_id.next()
+    pub fn new_pointer(&mut self, info: PointerInfo) -> PointerId {
+        self.ptr_info.push(info)
     }
 
     pub fn num_pointers(&self) -> usize {
-        self.next_ptr_id.num_pointers()
+        self.ptr_info.len()
+    }
+
+    pub fn ptr_info(&self) -> &GlobalPointerTable<PointerInfo> {
+        &self.ptr_info
     }
 
     /// Update all [`PointerId`]s in `self`, replacing each `p` with `map[p]`.  Also sets the "next
@@ -259,11 +574,17 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         let GlobalAnalysisCtxt {
             tcx: _,
             lcx,
+            ref mut ptr_info,
+            fn_callers: _,
             ref mut fn_sigs,
-            ref mut field_tys,
+            fns_failed: _,
+            ref mut field_ltys,
             ref mut static_tys,
-            ref mut next_ptr_id,
+            ref mut addr_of_static,
+            adt_metadata: _,
         } = *self;
+
+        *ptr_info = remap_global_ptr_info(ptr_info, map, counter.num_pointers());
 
         for sig in fn_sigs.values_mut() {
             sig.inputs = lcx.mk_slice(
@@ -275,7 +596,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             sig.output = remap_lty_pointers(lcx, map, sig.output);
         }
 
-        for labeled_field in field_tys.values_mut() {
+        for labeled_field in field_ltys.values_mut() {
             *labeled_field = remap_lty_pointers(lcx, map, labeled_field);
         }
 
@@ -283,22 +604,52 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             *labeled_static = remap_lty_pointers(lcx, map, labeled_static);
         }
 
-        *next_ptr_id = counter;
+        for ptr in addr_of_static.values_mut() {
+            if !ptr.is_none() {
+                *ptr = map[*ptr];
+            }
+        }
     }
 
     pub fn assign_pointer_to_static(&mut self, did: DefId) {
-        let lty = self.assign_pointer_ids(self.tcx.static_ptr_ty(did));
+        trace!("assign_pointer_to_static({:?})", did);
+        // Statics always have full type annotations.
+        let lty = self.assign_pointer_ids_with_info(self.tcx.type_of(did), PointerInfo::ANNOTATED);
+        let ptr = self.new_pointer(PointerInfo::empty());
         self.static_tys.insert(did, lty);
+        self.addr_of_static.insert(did, ptr);
     }
 
     pub fn assign_pointer_to_field(&mut self, field: &FieldDef) {
-        let lty = self.assign_pointer_ids(self.tcx.type_of(field.did));
-        self.field_tys.insert(field.did, lty);
+        let lty =
+            self.assign_pointer_ids_with_info(self.tcx.type_of(field.did), PointerInfo::ANNOTATED);
+        self.field_ltys.insert(field.did, lty);
     }
 
     pub fn assign_pointer_to_fields(&mut self, did: DefId) {
         for field in self.tcx.adt_def(did).all_fields() {
             self.assign_pointer_to_field(field);
+        }
+    }
+
+    pub fn fn_failed(&mut self, did: DefId) -> bool {
+        self.fns_failed.contains_key(&did)
+    }
+
+    pub fn mark_fn_failed(&mut self, did: DefId, detail: PanicDetail) {
+        if self.fns_failed.contains_key(&did) {
+            return;
+        }
+
+        self.fns_failed.insert(did, detail);
+
+        // This is the first time marking `did` as failed, so also mark all of its callers.
+        let callers = self.fn_callers.get(&did).cloned().unwrap_or(Vec::new());
+        for caller in callers {
+            self.mark_fn_failed(
+                caller,
+                PanicDetail::new(format!("analysis failed on callee {:?}", did)),
+            );
         }
     }
 }
@@ -311,13 +662,13 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
         let tcx = gacx.tcx;
         AnalysisCtxt {
             gacx,
+            ptr_info: LocalPointerTable::empty(),
             local_decls: &mir.local_decls,
             local_tys: IndexVec::new(),
-            c_void_casts: CVoidCasts::new(mir, tcx),
             addr_of_local: IndexVec::new(),
+            c_void_casts: CVoidCasts::new(mir, tcx),
             rvalue_tys: HashMap::new(),
             string_literal_locs: Default::default(),
-            next_ptr_id: NextLocalPointerId::new(),
         }
     }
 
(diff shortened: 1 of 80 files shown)


c2rust-analyze/src/util.rs
```diff
@@ -305,3 +306,44 @@ pub fn lty_project<'tcx, L: Debug>(
...
+pub fn are_transmutable<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> bool {
+    let transmutable_ints = {
+        use IntTy::*;
+        use UintTy::*;
+        match (a.kind(), b.kind()) {
+            (ty::Uint(u), ty::Int(i)) | (ty::Int(i), ty::Uint(u)) => {
+                matches!((u, i), |(Usize, Isize)| (U8, I8)
+                    | (U16, I16)
+                    | (U32, I32)
+                    | (U64, I64))
```
fw-immunant, line 327: I hate this formats as a closure.

c2rust-analyze/src/context.rs
```diff
@@ -349,23 +349,22 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
...
-                    // source type in all positions.  This works because the two types have the
-                    // same structure.
-                    return self.lcx().mk(ty, op_lty.args, op_lty.label);
+                // We only support pointer casts when:
+                // * both types are pointers
+                // * they have compatible (safely transmutable) pointee types
+                // Safe transmutability is difficult to check abstractly,
+                // so we limit it to integer types of the same size
+                // (but potentially different signedness).
+                // In particular, this allows casts from `*u8` to `core::ffi::c_char`.
```
fw-immunant, line 358: should be "...to `*core::ffi::c_char`"

fw-immunant (approved):
LGTM.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -206,10 +206,12 @@ impl<'tcx> TypeChecker<'tcx, '_> {
     /// that position.  For example, given `lty1 = *mut /*l1*/ *const /*l2*/ u8` and `lty2 = *mut
     /// /*l3*/ *const /*l4*/ u8`, this function will unify `l1` with `l3` and `l2` with `l4`.
```
fw-immunant, line 207: You might update "identical" to "transmutable" or "compatible" in the earlier bit of this comment block.

c2rust-analyze/src/util.rs
kkysen: Yep, I hate it, too, but I couldn't figure out a way around `rustfmt`.

kkysen: It makes it look like a closure.

c2rust-analyze/src/context.rs
kkysen: Fixed in a [rebase](https://github.com/immunant/c2rust/compare/a7f06c4e32e33f7dbed6c580b77125e0c8287fb8..19d0dac835e5db42b4446725f51328d115ac06ab).

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Fixed in 255a8bec72950888deed3f3ffcefd7a17162858c.

kkysen: Does GitHub not let you comment on the earlier lines since it's too far?  That sucks.

c2rust-analyze/src/util.rs
```diff
@@ -305,3 +306,44 @@ pub fn lty_project<'tcx, L: Debug>(
...
+/// Determine if two types (e.x. in a cast) are pointers,
+/// and if they are, if the pointee types are compatible,
+/// i.e. they are safely transmutable to each other.
+///
+/// This returns [`Some`]`(is_transmutable)` if they're both pointers,
+/// and [`None`] if its some other types.
+///
+/// See [`are_transmutable`] for the definition of safe transmutability.
+pub fn are_transmutable_ptrs<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> Option<bool> {
+    let a = a.builtin_deref(true)?.ty;
```
aneksteind, line 346: calls to `builtin_deref` are still valid on non-pointer types even when `explicit == true`, it's just that `explicit` will be ignored (e.g. comparing `Box<i32>` and `Box<u32>` will still work here. so i think it'd be worthwhile to add an assertion here that the inputs are raw pointers if that's all you'd like to limit it to. if boxes are permissible then please ignore

aneksteind (approved):

c2rust-analyze/tests/analyze/string_casts.rs
```diff
@@ -1,4 +1,3 @@
-#[cfg(any())]
```
aneksteind, line 1: are these sorts of casts representative of most of `lighttpd` code instances of casting between transmutable types. if not, we might want to expand what counts as transmutable in another PR

c2rust-analyze/src/util.rs
kkysen: `builtin_deref` works on `&`, `*`, and `Box`, but there can't be an `as Box<T>` cast, right?  That won't pass typeck.

c2rust-analyze/tests/analyze/string_casts.rs
kkysen: The `u8` to `i8` (`c_char`) seems to be the most common cast, used for transpiled string literals.  There are also the decaying casts (also a part of transpiled string literals), which I expanded the transmutability definition to in #841.

c2rust-analyze/src/util.rs
aneksteind: I think you're right

spernsteiner:
It seems odd to declare that `usize` and `*mut u8` are not safely transmutable, when you can safely convert between them using `as`, and that conversion behaves exactly the same as `mem::transmute`.  But I agree this conversion shouldn't be supported in this PR - `usize` has nowhere to attach a `PointerId`, which is basically the same problem we have with general use of `void*`.  Maybe we need a different name for this property?

Also, it would be good to clearly distinguish between the precise logical property, which may be difficult or impossible to check, and the conservative approximation of that property that we use in the implementation.  For example, in the trivial functions PR, the precise property is "calls to this function have no effect on pointer permissions in the caller", and the approximation is "there are no raw pointers reachable through the types in this function's signature".

The casts allowed by this PR will require some additional library support during rewriting - there's no way to safely convert `&str` to `&[i8]` using only the standard library.

kkysen:
> It seems odd to declare that `usize` and `*mut u8` are not safely transmutable, when you can safely convert between them using `as`, and that conversion behaves exactly the same as `mem::transmute`. But I agree this conversion shouldn't be supported in this PR - `usize` has nowhere to attach a `PointerId`, which is basically the same problem we have with general use of `void*`. Maybe we need a different name for this property?

Ptr-to-int and int-to-ptr transmutes are very weird, much more than casts, and may be UB or provenance dropping casts.  They're not quite safe to always transmute.

https://github.com/rust-lang/unsafe-code-guidelines/issues/286

> Also, it would be good to clearly distinguish between the precise logical property, which may be difficult or impossible to check, and the conservative approximation of that property that we use in the implementation. For example, in the trivial functions PR, the precise property is "calls to this function have no effect on pointer permissions in the caller", and the approximation is "there are no raw pointers reachable through the types in this function's signature".

I thought I did?  Let me check.  I at least did in the follow-up PR to this one.

> The casts allowed by this PR will require some additional library support during rewriting - there's no way to safely convert `&str` to `&[i8]` using only the standard library.

Wdym by this?  This PR doesn't allow that cast, nor anything to do with `&str`.

spernsteiner:
> It's in the documentation here.

Ah, got it - the official definition is the bit about whether `*(a as *const B)` and `*(b as *const A)` are safe.

>Ptr-to-int transmutes implicitly drop provenance (still being decided) and int-to-ptr transmutes are just UB. They're not safe

I haven't read the full thread you linked, but I saw the bit at the end about ptr-to-int transmute dropping provenance, which I believe is the same behavior you get from the safe cast `my_ptr as usize`.  Do you happen to know whether they intend to make `mem::transmute::<usize, *mut u8>` UB, contrary to the safe cast `my_usize as *mut u8`?

I guess this is a bit of a moot point, since "safely transmutable" is effectively defined in terms of "whatever the Rust devs decide is well-defined".

>Wdym by this? This PR doesn't allow that cast, nor anything to do with &str.

Sorry, I forgot the example was using `b""` rather than `""`.  But I think the same point applies.  The example `b"" as *const u8 as *const core::ffi::c_char` will get rewritten into something involving a cast from `&[u8]` to `&[i8]`.  This operation is well-defined, but `std` doesn't provide a safe function for performing it (as far as I know).  So we'll eventually need to write or import a library that does provide such a function so we can do the necessary rewriting.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -202,15 +202,18 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-        assert_eq!(
-            self.acx.tcx().erase_regions(lty1.ty),
-            self.acx.tcx().erase_regions(lty2.ty)
-        );
+        let ty1 = lty1.ty;
+        let ty2 = lty2.ty;
+        assert!(are_transmutable(
+            self.acx.tcx().erase_regions(ty1),
+            self.acx.tcx().erase_regions(ty2),
+        ), "types not transmutable (compatible), so PointerId unification cannot be done: {ty1:?} !~ {ty2:?}");
```
spernsteiner, line 216: This relaxed check should go somewhere specific to `Rvalue::Cast`, not in `do_unify`, where it would affect all `Assign` and pseudo-assignment operations.  Unfortunately, the limited cast handling I implemented previously involves some hacks that will need to be backed out before this is possible.

Regarding `do_unify`, in some other comment I wrote out the subtyping rule that this whole analysis is based on: `*mut /*1*/ T <: *mut /*2*/ U` if `perms(2) ⊆ perms(1)` and `T = U`.  `do_unify` is part of the implementation of this rule.  Specifically, `do_assign(*mut /*2*/ U, *mut /*1*/ T)` asserts the constraint `*mut /*1*/ T <: *mut /*2*/ U`, `do_assign_pointer_ids(2, 1)` asserts `perms(2) ⊆ perms(1)`, and `do_unify(U, T)` asserts `T = U`.  All three functions should be left as-is so that `Assign` and pseudo-assignments still follow the proper rules.

Now, regarding casts, there are three types involved in a cast like `(q: V) = (p: T) as U`.  The relationships required between these types are `U <: V` and `T is-castable-to U`.  However, the hacky logic in the `Cast` case of `AnalysisCtxt::type_of_rvalue` does not return the actual type `U`, but instead combines parts of `T` and parts of `U` to produce a new type `X`.  As a side effect, this has to assert that `T` and `U` are sufficiently similar (prior to this branch, this meant they had to have the same pointee type).  Then `dataflow::type_check` simply checks `X <: V`, and implements no `is-castable-to` check at all (in fact it can't, since it doesn't have access to `U`).

This PR relaxes some of the assertions in `type_of_rvalue`, but leaves `type_of_rvalue` returning the mixed type `X`.  But under these relaxed checks, in some cases `X` is not similar enough to `U` to actually pass the subtyping check `X <: V`, so this branch is forced to relax some assertions in `do_unify` as well.

A cleaner way to implement this would be to have `type_of_rvalue` actually return `U`, and have `dataflow::type_check` assert exactly the required constraints `U <: V` and `T is-castable-to U`.  Specifically:
1. In `main.rs`, when populating `acx.rvalue_tys`, create `LTy`s for the target types of all `Rvalue::Cast`s.  (The TODO comment there says "certain cases", but I think you can just do all casts instead.)  This `LTy` will be the `U` type for each cast.
2. In `AnalysisCtxt::type_of_rvalue`, remove all the `Cast` cases.  Casts should instead be handled by the `self.rvalue_tys` lookup at the top (similar to array aggregates).
3. In `dataflow::type_check`, update the `Rvalue::Cast` case in `visit_rvalue` to assert the constraint that `T is-castable-to U` (looking at the `Aggregate` case, I believe `T` is `acx.type_of(op)` and `U` is the `lty` argument of `visit_rvalue`).  This should accept the same casts that are currently allowed by `type_of_rvalue`, plus the transmutability-related relaxations from this PR.
This shouldn't require any changes to `do_unify` or the rest of `Assign` handling (`U <: V` is just the normal constraint produced by the `Assign` case).

kkysen:
> Ah, got it - the official definition is the bit about whether `*(a as *const B)` and `*(b as *const A)` are safe.

Yeah.

> I haven't read the full thread you linked, but I saw the bit at the end about ptr-to-int transmute dropping provenance, which I believe is the same behavior you get from the safe cast `my_ptr as usize`. Do you happen to know whether they intend to make `mem::transmute::<usize, *mut u8>` UB, contrary to the safe cast `my_usize as *mut u8`?
> 
> I guess this is a bit of a moot point, since "safely transmutable" is effectively defined in terms of "whatever the Rust devs decide is well-defined".

The ptr-to-int transmute drops provenance, corresponding to `ptr.addr()`.  That's different from `as usize`, which is `ptr.expose_addr()` and preserves provenance.  That's why it can't be round tripped back into a pointer, since it has no provence.  If it has no provenance, turning it into a pointer is UB since the pointer needs provenance.

> Sorry, I forgot the example was using `b""` rather than `""`. But I think the same point applies. The example `b"" as *const u8 as *const core::ffi::c_char` will get rewritten into something involving a cast from `&[u8]` to `&[i8]`. This operation is well-defined, but `std` doesn't provide a safe function for performing it (as far as I know). So we'll eventually need to write or import a library that does provide such a function so we can do the necessary rewriting.

It is in `core`.  You can convert safely through `CStr`:
* from `&[u8]`: [`CStr::from_bytes_with_nul`](https://doc.rust-lang.org/stable/core/ffi/struct.CStr.html#method.from_bytes_with_nul)
* to `*const c_char`: [`CStr::as_ptr`](https://doc.rust-lang.org/stable/core/ffi/struct.CStr.html#method.as_ptr)

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Now that I think about it, why do we need any check here in the first place?  Why not run static analysis after typeck (`Callbacks::after_analysis`) and simply allow everything that is valid Rust?

kkysen: We also should put some of these explanations in documentation.  They're not clear at all from the code, but a lot clearer now.

spernsteiner:
> If it has no provenance, turning it into a pointer is UB since the pointer needs provenance.

It surely can't be UB to create such a pointer, since `core::ptr::invalid` is safe and does exactly that.  (Dereferencing it would of course still be UB.)  If ptr-to-int transmute is equivalent to `ptr.addr()` (a safe operation), and int-to-ptr transmute is equivalent to `ptr::invalid` (a safe operation), then I think that means `usize` and `*mut u8` are safely transmutable under our definition.

However, I think we definitely don't want to allow casts between `*mut usize` and `*mut *mut u8`.  We might want to rewrite that `*mut u8` to `&[u8]`, for example, and `usize` and `&[u8]` definitely are not safely transmutable.  So I think we want some further restriction on casts beyond "pointee types are safely transmutable", something like (1) the pointee types must have `PointerId`s in roughly the same positions, and (2) those `PointerId`s must have identical permissions (same equivalence class).  This still allows for `i8` <-> `u8` and such, since those have no `PointerId`s at all.

> You can convert safely through CStr

`CStr` is extremely limited - it includes a null terminator check, and returns a raw pointer rather than a safe reference.  This PR also allows a wide variety of other casts, such as `*mut i32 -> *mut u32`, which will get rewritten into something like `&mut i32 -> &mut u32` or `&[i32] -> &[u32]`, and (once we implement rewriting for such casts) these will require some kind of library support to do in safe code.

kkysen:
> > If it has no provenance, turning it into a pointer is UB since the pointer needs provenance.
> 
> It surely can't be UB to create such a pointer, since `core::ptr::invalid` is safe and does exactly that. (Dereferencing it would of course still be UB.) If ptr-to-int transmute is equivalent to `ptr.addr()` (a safe operation), and int-to-ptr transmute is equivalent to `ptr::invalid` (a safe operation), then I think that means `usize` and `*mut u8` are safely transmutable under our definition.

Sorry, I didn't mean the transmute/cast itself is UB, but any dereference after.  My safe transmutability definition includes safely dereferencing the pointer, so this doesn't count.

> However, I think we definitely don't want to allow casts between `*mut usize` and `*mut *mut u8`. We might want to rewrite that `*mut u8` to `&[u8]`, for example, and `usize` and `&[u8]` definitely are not safely transmutable. So I think we want some further restriction on casts beyond "pointee types are safely transmutable", something like (1) the pointee types must have `PointerId`s in roughly the same positions, and (2) those `PointerId`s must have identical permissions (same equivalence class). This still allows for `i8` <-> `u8` and such, since those have no `PointerId`s at all.

Is this still a problem when we don't count `usize` at all?  `usize` to ptr casts are not included under my definition or implementation.

> `CStr` is extremely limited - it includes a null terminator check, and returns a raw pointer rather than a safe reference.

I'm confused.  `&CStr` is the reference, and you can convert it to either raw pointer at unsafe boundaries.  `&CStr` is the idiomatic way to do this in handwritten Rust, right?  Before `c""` strings are implemented.

> This PR also allows a wide variety of other casts, such as `*mut i32 -> *mut u32`, which will get rewritten into something like `&mut i32 -> &mut u32` or `&[i32] -> &[u32]`, and (once we implement rewriting for such casts) these will require some kind of library support to do in safe code.

Allowing the cast, i.e. not crashing on it, doesn't mean we have to make it safe.  We're trying to expand the set of code that we don't crash on, and ultimately, this is a valid cast.  This is normally done in Rust with unsafe.  Why can't we just keep it to a well-scoped unsafe block with one of the `from_raw_parts` functions?

spernsteiner:
> My safe transmutability definition includes safely dereferencing the pointer, so this doesn't count.

The definition dereferences a pointer to the `*mut u8`, not the `*mut u8` itself:

>Formally, safe transmutability defines an equivalence relation on types, named `~` here. `A ~ B` iff `*(a as *const B)` and `*(b as *const A)` are safe, where `a: *const A` and `b: *const B`.

Let `A = *mut u8` and `B = usize`.  `*((a: *const *mut u8) as *const usize)` is well defined: from the discussion above, it's equivalent to `(*a).addr()`.  `*((b: *const usize) as *const *mut u8)` is also well defined: it's equivalent to `ptr::invalid(*b)`.  Therefore `*mut u8 ~ usize`.

The current documentation suggests that we think the cast handling strategy implemented here is valid for any cast from `*const A` to `*const B` where `A ~ B`.  But that's not the case.  Rewriting cannot handle casting `*const usize` to `*const *mut u8`, not even using unsafe code, at least for some combinations of reference types (e.g. `&[usize]` to `&[&[u8]]`).  And for the dataflow analysis, the `*const usize` to `*const *mut u8` cast is awfully similar to int-to-ptr and/or a (general) `void*` cast, which we currently don't know a good strategy for handling.

This is purely a documentation issue - I'm convinced the cast handling in this PR will work fine for `*const u32` to `*const i32` and such.  But it's wrong to imply that the only thing needed to handle `*const usize` to `*const *mut u8` is to make the `A ~ B` check more precise.

>&CStr is the idiomatic way to do this in handwritten Rust

I'm thinking of non-C-string use cases.  For example, suppose I have `let x: [u8; 4] = [1, 0, 2, 0]` and I want to convert `&x` to `&[i8]`.  I can't use `CStr` for this, first because it doesn't actually return `&[i8]`, and second because it will either truncate or error at the first 0.

> Why can't we just keep it to a well-scoped unsafe block with one of the from_raw_parts functions?

I don't think we should leave extra `unsafe` in the user's code just to avoid adding a dependency.  `c2rust-transpile` already uses a few support libraries like this in the translation of certain features.

My comment about needing a support library wasn't meant as a criticism of this PR - I was just pointing out that supporting these casts will require one, whereas so far we've gotten by with just the standard library.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Now that #883 has been merged, 6d20f40f39d28085e2361eacf84bd90e8635dbd3 reverts this part.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -113,31 +110,13 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-            CastKind::Pointer(..) => {
-                // The source and target types are both pointers, and they have identical pointee types.
-                // TODO: remove or move check to `is_castable_to`
-                assert!(from_lty.args[0].ty == to_lty.args[0].ty);
-                assert!(is_castable_to(from_lty, to_lty));
-            }
             _ => {
-                // A cast such as `T as U`
-                assert!(is_castable_to(from_lty, to_lty));
+                // TODO add dataflow constraints
```
kkysen, line 114: As the dataflow constraints are already not done correctly/at all prior to this PR, I'm not going to fix that here yet, but will fix that in a follow-up PR so that this one doesn't get too large (it's already very old).

kkysen:
I'm getting two errors from this now that I'm not sure about:

## `as_ptr.rs`

```sh
running 1 test
tests/filecheck/as_ptr.rs:33:15: error: CHECK-DAG: expected string not found in input
// CHECK-DAG: ([[@LINE+1]]: x): &{{\[\[i32; 10]]}}
              ^
<stdin>:1367:47: note: scanning from here
type assignment for "array_as_ptr_offset_load":
                                              ^
<stdin>:1367:47: note: with "@LINE+1" equal to "34"
type assignment for "array_as_ptr_offset_load":
                                              ^
<stdin>:1369:5: note: possible intended match here
_1 (34: x): &[i32; 10]
    ^

Input file: <stdin>
Check file: tests/filecheck/as_ptr.rs

-dump-input=help explains the following input dump.

Input was:
<<<<<<
          .
          .
          .
       1362: _5 (38: q): addr_of = l9, type = l8#*const i32[NONE#i32[]]
       1363: _6 (38: p): addr_of = UNIQUE, type = READ | UNIQUE | OFFSET_ADD | OFFSET_SUB#*const i32[(empty)#i32[]]
       1364: _6 (38: p): addr_of flags = (empty), type flags = (empty)#*const i32[(empty)#i32[]]
       1365: _6 (38: p): addr_of = l11, type = l10#*const i32[NONE#i32[]]
       1366:
       1367: type assignment for "array_as_ptr_offset_load":
dag:33'0                                                   X~ error: no match found
dag:33'1                                                      with "@LINE+1" equal to "34"
       1368: _0 (34: i32): i32
dag:33'0     ~~~~~~~~~~~~~~~~~~
       1369: _1 (34: x): &[i32; 10]
dag:33'0     ~~~~~~~~~~~~~~~~~~~~~~~
dag:33'2         ?                   possible intended match
       1370: _2 (36: p): &[i32]
dag:33'0     ~~~~~~~~~~~~~~~~~~~
       1371: _3 (36: x.as_ptr()): &[[i32]]
dag:33'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       1372: _4 (36: x.as_ptr()): &[i32; 10]
dag:33'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       1373: _5 (38: q): &i32
dag:33'0     ~~~~~~~~~~~~~~~~~
       1374: _6 (38: p): &[i32]
dag:33'0     ~~~~~~~~~~~~~~~~~~~
          .
          .
          .
>>>>>>
test filecheck ... FAILED
```

and

## `string_casts.rs`

These are caused by these three functions in particular, but not the other cast ones:

```rust
pub fn cast_array_to_slice_ptr(s: &[u8; 0]) {
    s as *const [u8];
}

pub fn cast_array_to_ptr(s: &[u8; 0]) {
    s as *const u8;
}

pub fn cast_from_literal() {
    b"" as *const u8 as *const core::ffi::c_char;
}
```

```sh
thread 'rustc' panicked at 'expected to find only one Assign statement, but got multiple:
expr = Expr { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 4 }, kind: Path(Resolved(None, Path { span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:6 (#0), res: Local(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 2 }), segments: [PathSegment { ident: s#0, hir_id: Some(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 3 }), res: Some(Local(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 2 })), args: None, infer_args: true }] })), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:6 (#0) }
matches = [Left(_3 = &raw const (*_1)), Left(_2 = move _3 as *const [u8] (Pointer(Unsize)))]', c2rust-analyze/src/rewrite/expr/hir_op.rs:126:21
```

which follows a bunch of warnings like this:

```sh
warning: find_primary_location: unsupported expr Expr { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 12 }, kind: Block(Block { stmts: [Stmt { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 10 }, kind: Semi(Expr { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 9 }, kind: Cast(Expr { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 4 }, kind: Path(Resolved(None, Path { span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:6 (#0), res: Local(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 2 }), segments: [PathSegment { ident: s#0, hir_id: Some(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 3 }), res: Some(Local(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 2 })), args: None, infer_args: true }] })), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:6 (#0) }, Ty { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 8 }, kind: Ptr(MutTy { ty: Ty { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 7 }, kind: Slice(Ty { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 5 }, kind: Path(Resolved(None, Path { span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:18: 14:20 (#0), res: PrimTy(Uint(U8)), segments: [PathSegment { ident: u8#0, hir_id: Some(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 6 }), res: Some(PrimTy(Uint(U8))), args: None, infer_args: false }] })), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:18: 14:20 (#0) }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:17: 14:21 (#0) }, mutbl: Not }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:10: 14:21 (#0) }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:21 (#0) }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:22 (#0) }], expr: None, hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 11 }, rules: DefaultBlock, span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:13:45: 15:2 (#0), targeted_by_break: false }, None), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:13:45: 15:2 (#0) }
warning: find_primary_location: unsupported expr Expr { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 9 }, kind: Cast(Expr { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 4 }, kind: Path(Resolved(None, Path { span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:6 (#0), res: Local(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 2 }), segments: [PathSegment { ident: s#0, hir_id: Some(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 3 }), res: Some(Local(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 2 })), args: None, infer_args: true }] })), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:6 (#0) }, Ty { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 8 }, kind: Ptr(MutTy { ty: Ty { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 7 }, kind: Slice(Ty { hir_id: HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 5 }, kind: Path(Resolved(None, Path { span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:18: 14:20 (#0), res: PrimTy(Uint(U8)), segments: [PathSegment { ident: u8#0, hir_id: Some(HirId { owner: DefId(0:6 ~ string_casts[c5c4]::cast_array_to_slice_ptr), local_id: 6 }), res: Some(PrimTy(Uint(U8))), args: None, infer_args: false }] })), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:18: 14:20 (#0) }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:17: 14:21 (#0) }, mutbl: Not }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:10: 14:21 (#0) }), span: /home/kkysen/work/rust/c2rust2/c2rust-analyze/tests/analyze/string_casts.rs:14:5: 14:21 (#0) }
```

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -113,31 +110,13 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-
-                assert_matches!(from_lty.kind(), TyKind::Ref(..) | TyKind::RawPtr(..));
-                let pointee_from_lty = assert_matches!(from_lty.args, [lty] => lty);
-
-                let elem_to_ty = assert_matches!(pointee_to_ty.kind(), &TyKind::Slice(ty) => ty);
-                assert!(matches!(pointee_from_lty.kind(), TyKind::Array(..)));
-                let elem_from_lty = assert_matches!(pointee_from_lty.args, [lty] => lty);
-                assert_eq!(elem_from_lty.ty, elem_to_ty);
-                assert_eq!(pointee_from_lty.label, PointerId::NONE);
-                self.do_assign_pointer_ids(to_lty.label, from_lty.label);
```
aneksteind, line 130: i think this deletion could be why `as_ptr.rs` is now failing

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -104,6 +100,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
     fn visit_cast(&mut self, cast_kind: CastKind, op: &Operand<'tcx>, to_lty: LTy<'tcx>) {
         let to_ty = to_lty.ty;
         let from_lty = self.acx.type_of(op);
+        let from_ty = from_lty.ty;
```
aneksteind, line 103: nit: just pass `from_lty.ty` to `is_transmutable_ptr_cast`

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Yeah, it is.  I added it back now and now `as_ptr.rs` succeeds.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: It lets me use `{from_ty:?}`, as only variables work in that position, not expressions.

c2rust-analyze/tests/analyze/string_casts.rs
```diff
@@ -10,6 +10,15 @@ pub fn cast_slice_ptr_to_ptr(s: *const [u8]) {
     s as *const u8;
 }
 
+/// For the below disabled (`#[cfg(any())]`ed) tests, they currently crash in the rewriter
+/// due to it not being able to handle implicitly inserted `std::ptr::addr_of!`s yet.
```
aneksteind, line 14: I would replace this wording with implicitly-inserted `&raw` statements

kkysen: Fixed in 11bf3514929133cd5ee44fe4eb5956bb0e4c07e8.

kkysen:
@spernsteiner, could you take a look at this again?  I got two more of the tests working with explicit `addr_of!`s, though the third still didn't work, and is just disabled.  Does the rest of the cast and transmutability logic look good, barring fixing the missing dataflow constraints and rewriter issues that will be fixed separately?

c2rust-analyze/src/util.rs
```diff
@@ -351,3 +352,70 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+/// so this function only checks a subset of it,
+/// with these formal rules for all types `A`, `B`:
+///
+/// * `A = B => A ~ B`
+/// * `A ~ B => *A ~ *B`
+/// * `A ~ B => [A] ~ B`
+/// * `A ~ B => [A; N] ~ B`, where `const N: usize`
+/// * `A ~ B => [A] ~ [B; N]`, where `const N: usize`
+/// * `A ~ B => [A; N] ~ [B]`, where `const N: usize`
+/// * `uN ~ iN`, `iN ~ uN`, where `N` is an integer width
```
spernsteiner, line 376: These rules let you derive the following: `u8 = u8 => u8 ~ u8 => [u8; 1] ~ [u8] => *[u8; 1] ~ *[u8]`.  This implies that it's safe to do `*(a as *const *mut [u8])` where `a: *const *mut [u8; 1]`.  But even when `*a` is safe, `*(a as *const *mut [u8])` will often be unsafe, since it reads a two-word `*mut [u8]` from a location that might only contain a one-word `*mut [u8; 1]`.

c2rust-analyze/src/util.rs
spernsteiner: Does removing the leading `|` make rustfmt do something more reasonable here?  Failing that, can you make rustfmt ignore this bit of code?  The closure-style formatting is actively misleading.

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: Please open an issue for this and make it a high priority.  Debugging this sort of constraint-solving analysis is hard enough already - we shouldn't knowingly leave it in a broken state on top of that.

c2rust-analyze/src/util.rs
kkysen: Oh actually yes!  Without the leading `|` it's like this:

```rust
                matches!(
                    (u, i),
                    (Usize, Isize) | (U8, I8) | (U16, I16) | (U32, I32) | (U64, I64)
                )
```

Much better.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: I will.  I was planning on working on this next.

c2rust-analyze/src/util.rs
```diff
@@ -351,3 +352,70 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+
+/// Determine if `from` can be safely transmuted to `to`,
+/// which is defined as `*(from as *const To)` being a safe operation,
+/// where `from: *const From` and assuming `*from` already was safe.
+///
+/// Note that this is one-way, and is slightly different from [`core::mem::transmute`],
+/// and more similar to [`core::mem::transmute_copy`].
+///
+/// This forms a non-symmetric (one-way) equivalence relation, named `~` below.
+/// Formally, `A ~ B` iff `*a` and `*(a as *const B)` are safe, where `a: *const A`.
```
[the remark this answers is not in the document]
kkysen: Fixed in 44ac9f497b0d073104e625c228a8cedf20f11ac9.  I replaced "equivalence" with "reflexive, transitive", since it's still those things.

c2rust-analyze/src/util.rs
kkysen: Fixed in 65e51406868ad4829d8c93a12ba4743c16c4821b.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Tracked in #913.

c2rust-analyze/src/util.rs
kkysen: I was thinking carefully about that case and thought it wasn't sound, but it did pass `miri`, so I thought it meant it was, but I think I may have miswritten/mistested it slightly.

kkysen: I'm going to remove the `[A; N] ~ [B]` rule, but keep the `[A] ~ [B; N]` one, as this fails:

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aa387753ff96bbdaf9696d4251bb5132)

```rust
unsafe fn f(a: *const *mut [u8]) {
    let _ = *(a as *const *mut [u8; 1]);
}

fn main() {
    let s = &mut [b'a'] as *mut [u8];
    unsafe {
        f(&s);
    }
}
```

but this works:

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7d25cc382885804299a48c414cdf0d8f)

```rust
unsafe fn f(a: *const *mut [u8; 1]) {
    let _ = *(a as *const *mut [u8]);
}

fn main() {
    let s = &mut [b'a'] as *mut [u8; 1];
    unsafe {
        f(&s);
    }
}
```

fw-immunant: I believe you have the two code blocks (but not playground links or descriptions) switched.

kkysen: You're right; fixed them.

kkysen: Reverted in 720c2fbb841a4cf83460c59db64b18256c280ebc.

kkysen:
@spernsteiner, could you take a look at the new separated `CastKind` handling (4bbb3065d6f8aa4961701fdab5cb6ffae21a823f) if that and the reversion in 720c2fbb841a4cf83460c59db64b18256c280ebc addresses your comment here: https://github.com/immunant/c2rust/pull/839#discussion_r1185587118?

Also, regarding https://github.com/immunant/c2rust/issues/909#issuecomment-1559963756, I'm not sure why #919 is causing  that crash related to `FIXED`.  Even without #839 at all, this function

```rust
pub fn cast_array_to_ptr_explicit(s: &[u8; 0]) {
    std::ptr::addr_of!(*s) as *const u8;
}
```

causes `master` to crash with that same error:

```shell
thread 'rustc' panicked at 'building TypeDesc for FIXED pointer requires a related pointee type', c2rust-analyze/src/type_desc.rs:75:5
```

spernsteiner:
I'll investigate that panic next week and also take a look at this PR

kkysen:
@spernsteiner, I've merged `master` into this PR after you just merged #945, so that crash is fixed now.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +354,64 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+///
+/// However, safe transmutability is difficult to check completely,
+/// so this function only checks a subset of it,
+/// with these formal rules for all types `A`, `B`:
+///
+/// * `A = B => A ~ B`
+/// * `A ~ B => *A ~ *B`
+/// * `uN ~ iN`, `iN ~ uN`, where `N` is an integer width
+/// * `[A] ~ A`
+/// * `[A; N] ~ A`, where `const N: usize`
```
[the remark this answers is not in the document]
spernsteiner: Here's an issue with zero-length arrays:
```Rust
let arr = [];
let p = &arr as *const [i32; 0];
let _ = *p;  // Well-defined: loads 0 bytes
let q = p as *const i32;
let _ = *q;  // Undefined behavior: loads 4 bytes from an object of size 0
```
There is a similar issue with zero-length slices, which is harder to address since the length is not part of the type.

Which of these rules do we actually need for lighttpd?  If we only need `*const u8 <-> *const i8` then maybe it would be best to remove the array and slice rules for now.  Otherwise, I'm not sure what we'd want to change regarding the rules and/or the definition of `A ~ B`.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +354,64 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+///
+/// Thus, [`true`] means it is definitely transmutable,
+/// while [`false`] means it may not be transmutable.
+pub fn is_transmutable_to<'tcx>(from: Ty<'tcx>, to: Ty<'tcx>) -> bool {
+    let transmutable_ints = || {
```
spernsteiner, line 381: I assume `transmutable_ints` and `one_way_transmutable` are closures so the check at the end can short-circuit to avoid them?  I suspect that's not necessary - LLVM should be smart enough to delay the memory loads as late as possible.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +354,64 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+                    (u, i),
+                    (Usize, Isize) | (U8, I8) | (U16, I16) | (U32, I32) | (U64, I64)
+                )
+            }
+            _ => false,
+        }
+    };
+
+    let one_way_transmutable = || match from.kind() {
+        &ty::Array(from, _) | &ty::Slice(from) => is_transmutable_to(from, to),
```
spernsteiner, line 396: This appears to implement `A ~ B => [A] ~ B` rather than `[A] ~ A`.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -114,31 +112,40 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                // it ensures if [`PointerCast`] is changed in a future `rustc` version,
+                // this won't compile until we've checked that this reasoning is still accurate.
+                match ptr_cast {
+                    PointerCast::ReifyFnPointer => {}
+                    PointerCast::UnsafeFnPointer => {}
+                    PointerCast::ClosureFnPointer(_) => {}
+                    PointerCast::MutToConstPointer => {}
+                    PointerCast::ArrayToPointer => {}
+                    PointerCast::Unsize => {}
+                }
```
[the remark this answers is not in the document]
kkysen: I'm planning to do this in the next PR right after this (which I've already started).  I'll keep this in mind.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -114,31 +112,40 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                self.do_assign_pointer_ids(to_lty.label, from_lty.label)
+                // TODO add other dataflow constraints
             }
-            _ => {
-                // A cast such as `T as U`
-                assert!(is_castable_to(from_lty, to_lty));
+            CastKind::Misc => {
+                match is_transmutable_ptr_cast(from_ty, to_ty) {
+                    Some(true) => {
+                        // TODO add other dataflow constraints
```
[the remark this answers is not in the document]
kkysen: I know.  I've already added this in the followup PR.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -295,23 +302,31 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         }
     }
 
-    /// Unify corresponding `PointerId`s in `lty1` and `lty2`.
+    /// Unify corresponding [`PointerId`]s in `pl_lty` and `rv_lty`.
```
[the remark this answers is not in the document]
kkysen: Okay, I'll revert that.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +354,64 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+
+/// Determine if `from` can be safely transmuted to `to`,
+/// which is defined as `*(from as *const To)` being a safe operation,
+/// where `from: *const From` and assuming `*from` already was safe.
+///
+/// Note that this is one-way, and is slightly different from [`core::mem::transmute`],
+/// and more similar to [`core::mem::transmute_copy`].
+///
+/// This forms a reflexive, transitive, and non-symmetric (one-way) relation, named `~` below.
+/// Formally, `A ~ B` iff `*a` and `*(a as *const B)` are safe, where `a: *const A`.
```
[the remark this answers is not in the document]
kkysen: By well-defined, do you mean not undefined as in UB?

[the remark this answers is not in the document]
spernsteiner: Yes, I just mean "not UB"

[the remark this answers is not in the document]
kkysen: Okay, I'll change this.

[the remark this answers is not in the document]
kkysen: Fixed in a73e64d53985c42fda66b55030ce7fecdd3a37e4.

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
kkysen: Fixed in 435e399677662f5bc8a9ae4846b55f8626affeb4.

c2rust-analyze/src/util.rs
kkysen: > I assume `transmutable_ints` and `one_way_transmutable` are closures so the check at the end can short-circuit to avoid them?

Yes.

> I suspect that's not necessary - LLVM should be smart enough to delay the memory loads as late as possible.

`one_way_transmutable` is recursive, so I don't think this is simple to elide.  Besides, I'd rather not depend on LLVM being super clever as the closure is very simple (just adding `||` to definitions and `()`s to calls).

c2rust-analyze/src/util.rs
kkysen: I meant to have the rule be `A ~ B => [A] ~ B` (and the same for arrays).  Would that change your analysis above in https://github.com/immunant/c2rust/pull/839#discussion_r1218720249 at all?

c2rust-analyze/src/util.rs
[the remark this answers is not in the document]
kkysen: > This writeup might still be useful for reasoning about the rules, once they're fixed.
> 
> The main property we want is that if `A ~ B`, then `sizeof(A) >= sizeof(B)` and `alignof(A) >= alignof(B)`. For rules 1 and 3, this obviously holds. For rule 2, this holds as long as `A ~ B` implies that if `*B` is a fat pointer, then `*A` is also a fat pointer.
> 
> This latter property obviously holds for rules 1 and 3. For rule 4, it holds because the `*[A]` is always a fat pointer. For rule 5, the property holds because when `*A` is a fat pointer, `A` is unsized, and therefore `[A; N]` is ill-formed. For rule 2, it holds because `**A` and `**B` are always thin pointers.

I can include this in the docs if you think that's good.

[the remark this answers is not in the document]
kkysen: > Which of these rules do we actually need for lighttpd? If we only need `*const u8 <-> *const i8` then maybe it would be best to remove the array and slice rules for now. Otherwise, I'm not sure what we'd want to change regarding the rules and/or the definition of `A ~ B`.

It's been a while so I'm trying to remember, but I'm pretty sure we need the array rule, as that's the intermediary cast for string literals.  For arrays, I can restrict the rule to `A ~ B, N > 0 => [A; N] ~ B`, right, as `N` will be a constant I check at compile-time.

> There is a similar issue with zero-length slices, which is harder to address since the length is not part of the type.

For slices, that's a lot trickier.  Let me check if we need them first.

c2rust-analyze/src/util.rs
kkysen: Update the rules in 1317e034b6c8fe847ff693e8f8f44ce74a4bc83b.

c2rust-analyze/src/util.rs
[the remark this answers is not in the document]
kkysen: I added `N > 0` as a pre-requisite to the array rule in 6b3bdd88b70103e40435cf575f4f520d8955e4ae.

For slices, I don't think we need them, so I'll remove them for now until we either need them or figure out a way to make them sound.

c2rust-analyze/src/util.rs
spernsteiner: I thought we previously discussed the rule `A ~ B => [A] -> B`, concluded that it was unsound, and replaced it with the more limited `[A] -> A` instead.

kkysen: Oh, maybe.  Let me check.

kkysen: I'm pretty sure what we discussed before and that I removed were the slice to array and array to slice casts/transmutes.  I don't remember any issues from `A ~ B, N > 0 => [A; N] ~ B` (the slice version I've removed now).

spernsteiner: I thought about this more and realized you can derive `[A; N] ~ B` from `[A; N] ~ A` and `A ~ B`, so there's no new issue with the `[A; N] ~ B` version of the rule.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +354,76 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+    };
+
+    let one_way_transmutable = || match *from.kind() {
+        ty::Array(from, n) => {
+            is_transmutable_to(from, to) && {
+                let is_zero = n
+                    .kind()
+                    .try_to_scalar_int()
+                    .map(|i| i.is_null())
+                    .unwrap_or(false);
```
spernsteiner, line 405: Can this be shortened using [`n.try_eval_target_usize(...)`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/struct.Const.html#method.try_eval_target_usize)?

spernsteiner (approved):
Looks good once the conflict + CI is fixed

c2rust-analyze/src/util.rs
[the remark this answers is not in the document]
kkysen: I added the fat ptr proof in 7f87dc1ece162561e675d185a25e9ede4b3bf459.

kkysen:
@spernsteiner, given #934 is all approved now, if you plan to merge it tomorrow, I'll wait until that merges to rebase/merge this before merging it into `master`, as I also want to check if #934 fixes the implicit `&raw` issue (#909).

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
kkysen: Fixed in 3c038606a32329d3e4ce21e308e8a49635a6d44f .

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
kkysen: Fixed in b20cea742881dee4c5b4dfe3003345c7db161250.

c2rust-analyze/src/util.rs
spernsteiner: >the array length has to be a scalar.

I don't think that's true - it could be a const generic argument instead.  Though it's fine to not handle that case for now.

PR #841, opened 2023-02-16
(`c2rust-analyze`) Relaxed the transmutable checks from two-way to one-way, now allowing for arrays and slices to decay

kkysen:
Relaxed the transmutable checks from two-way to one-way, now allowing for arrays and slices to decay.

This expands the definition of safe transmutability to be one-way. That is, it checks if `*T as *U` is safe, rather than also `*U as *T`.

Thus, we can now allow for casts decaying pointers to arrays and slices to pointers to their element type.

`do_unify` is modified to also be one-way, which it was already in all call sites.

New tests are also added to `string_casts.rs` for all the types of ptr-to-ptr casts.

Out of the full string cast, `b"" as *const u8 as *const core::ffi::c_char`, this adds support for the `as *const u8` (from `&[u8; _]`), so only support for the string literal itself remains.


c2rust-analyze/src/util.rs
```diff
@@ -307,19 +307,23 @@ pub fn lty_project<'tcx, L: Debug>(
...
+/// * decaying arrays and slices to their element type
+///     * e.x. `*[u8; 0] as *u8`
+///     * e.x. `*[u8] as *u8`
 ///
 /// Thus, [`true`] means it is definitely transmutable,
 /// while [`false`] means it may not be transmutable.
-pub fn are_transmutable<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> bool {
-    let transmutable_ints = {
+pub fn is_transmutable_to<'tcx>(from_ty: Ty<'tcx>, to_ty: Ty<'tcx>) -> bool {
+    let transmutable_ints = || {
```
fw-immunant, line 323: This documentation isn't completely correct now that the function allows `from_ty` to decay: for example, `is_transmutable_to([u8; 4], u8)` returns `true`, but you can't transmute from `[u8; 4]` to `u8` in Rust. I think the docs and name should be updated to clarify that this only applies behind a pointer, and perhaps it makes the most sense to simply fold this into `is_transmutable_ptr_cast`, as that's the only caller as well as the necessary context to understand the predicate being applied.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -205,14 +205,14 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-        ), "types not transmutable (compatible), so PointerId unification cannot be done: {ty1:?} !~ {ty2:?}");
-        for (sub_lty1, sub_lty2) in lty1.iter().zip(lty2.iter()) {
+    fn do_unify(&mut self, pl_lty: LTy<'tcx>, rv_lty: LTy<'tcx>) {
+        let pl_ty = pl_lty.ty;
+        let rv_ty = rv_lty.ty;
+        assert!(is_transmutable_to(
+            self.acx.tcx().erase_regions(rv_ty),
+            self.acx.tcx().erase_regions(pl_ty),
+        ), "types not transmutable (compatible), so PointerId unification cannot be done: *{rv_ty:?} as *{pl_ty:?}");
+        for (sub_lty1, sub_lty2) in pl_lty.iter().zip(rv_lty.iter()) {
```
fw-immunant, line 215: I'm confused here: are `pl_lty` née `lty1` and `rv_lty` née `lty2` the (possibly) pointer types, or types that have already been dereferenced? If the former, I don't see how their transmute compatibility can be handled by `is_transmutable_to` née `are_transmutable`, because that function doesn't do any dereferencing. If they've already been dereferenced, then the comment giving the example of `lty1 = *mut /*l1*/ *const /*l2*/ u8` seems misleading.

In any event, update the `lty1`/`lty2` names in the comment.

c2rust-analyze/src/util.rs
kkysen: `do_unify` also calls `is_transmutable_to`, that's why they're separate.  I'll fix the docs, though, and clarify that it's for behind a pointer, or more similar to `transmute_copy` than `transmute` now.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: > I'm confused here: are `pl_lty` née `lty1` and `rv_lty` née `lty2` the (possibly) pointer types, or types that have already been dereferenced? If the former, I don't see how their transmute compatibility can be handled by `is_transmutable_to` née `are_transmutable`, because that function doesn't do any dereferencing. If they've already been dereferenced, then the comment giving the example of `lty1 = *mut /*l1*/ *const /*l2*/ u8` seems misleading.

`pl_ty` and `rv_ty` are the possibly pointer types, but `tcx.erase_regions` strips the pointer (if it's there) from what I can tell.

> In any event, update the `lty1`/`lty2` names in the comment.

Thanks for catching that, I'll do that.

fw-immunant: I could be wrong, but I don't think `erase_regions` strips an outer pointer: looking at [this source](https://doc.rust-lang.org/beta/nightly-rustc/src/rustc_middle/ty/erase_regions.rs.html), I would expect `fold_ty` to do dispatch on the outer kind, but it only either dispatches to itself via the generic visitor (`super_fold_with`), or to a new instance of itself (`self.tcx.erase_regions_ty(ty)`).

kkysen: > In any event, update the `lty1`/`lty2` names in the comment.

Fixed in f05df654f7d3a12d0d104419b03fefd24639fbc9.

c2rust-analyze/src/util.rs
```diff
@@ -330,20 +334,24 @@ pub fn are_transmutable<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> bool {
...
     };
 
-    // only check for transmutable ints so far
-    a == b || transmutable_ints
+    let one_way_transmutable = || match from_ty.kind() {
+        &ty::Array(from_ty, _) | &ty::Slice(from_ty) => is_transmutable_to(from_ty, to_ty),
+        _ => false,
+    };
+
+    from_ty == to_ty || transmutable_ints() || one_way_transmutable()
```
aneksteind, line 342: are these function calls here for laziness?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -199,21 +199,29 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+    ///
+    /// ```
+    /// # fn(
+    /// pl_lty: *mut /*l1*/ *const /*l2*/ u8,
+    /// rv_lty: *mut /*l3*/ *const /*l4*/ u8,
+    /// # ) {}
+    /// ```
+    ///
+    /// this function will unify `l1` with `l3` and `l2` with `l4`.
+    fn do_unify(&mut self, pl_lty: LTy<'tcx>, rv_lty: LTy<'tcx>) {
```
aneksteind, line 217: i think we want to abstract this away from implying that unification is only from the perspective of a MIR assignment, so I don't think the `pl_lty` and `rv_lty` names will always be appropriate

aneksteind: on second glance, it does look like it's only done in the context of some sort of assignment. so maybe it's fine. i'm not sure where we may want to do equivalence relations outside of assignment, maybe @spernsteiner can say

c2rust-analyze/src/util.rs
kkysen: Yeah.  I want to put all the main operations in one line for clarity, but run them in the specific order I want.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Yeah, I looked at the call sites and they're always `pl_lty` and `rv_lty`, too, which is why I named it that.  I can also rename it to `from_lty` and `to_lty`, similar to the `is_transmutable_to` arg names.

c2rust-analyze/src/util.rs
kkysen: The docs have been fixed/updated and the definition formalized in b84b4129cfa2e85fd910932bed795434e01c66f3.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: > I could be wrong, but I don't think `erase_regions` strips an outer pointer

Yeah, you're right.  I misunderstood.

It seems that the `pl_ty` and `rv_ty` types passed to `do_unify` have already been dereferenced, which happens in `do_equivalence_nested` and other call sites.  I'm not sure what the example means.  @spernsteiner?

spernsteiner: `*const *mut *mut u8` is a legal type.  `do_equivalence_nested` will strip off the outermost `*const` and pass `*mut *mut u8` to `do_unify`.

aneksteind (approved):

kkysen:
I'm going to merge this into #839, then merge master into #839 (vs. a rebase since it's quite old and diverged and I don't want to lose the old history), fixup #839, and then have it reviewed before merging into master.

PR #848, opened 2023-02-23
Feat.cast.loc

aneksteind:
Separating some improvements for #788, namely
- keying casts by `Location`
- expecting casts from void to be the first statement in the destination block of an allocation call
- ensuring that the argument to `free` is not modified between its cast and its use

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -442,7 +442,7 @@ pub fn visit_body<'tcx>(
             };
             tc.current_location = loc;
 
-            if !c_void_casts.should_skip_stmt(&loc) {
+            if !c_void_casts.should_skip_stmt(loc) {
                 tc.visit_statement(stmt);
             }
         }
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -331,10 +331,6 @@ impl<'tcx> TypeChecker<'tcx, '_> {
     }
 
     pub fn visit_statement(&mut self, stmt: &Statement<'tcx>) {
-        if self.c_void_casts.should_skip_stmt(stmt) {
-            return;
-        }
-
         // TODO(spernsteiner): other `StatementKind`s will be handled in the future
         #[allow(clippy::single_match)]
         match stmt.kind {
@@ -440,17 +436,21 @@ pub fn visit_body<'tcx>(
         c_void_casts,
     };
 
-    for (bb, bb_data) in mir.basic_blocks().iter_enumerated() {
+    for (block, bb_data) in mir.basic_blocks().iter_enumerated() {
         for (idx, stmt) in bb_data.statements.iter().enumerate() {
-            tc.current_location = Location {
-                block: bb,
+            let loc = Location {
+                block,
                 statement_index: idx,
             };
-            tc.visit_statement(stmt);
+            tc.current_location = loc;
+
+            if !c_void_casts.should_skip_stmt(&loc) {
+                tc.visit_statement(stmt);
+            }
         }
 
         tc.current_location = Location {
-            block: bb,
+            block,
             statement_index: bb_data.statements.len(),
         };
         tc.visit_terminator(bb_data.terminator());
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -26,7 +26,6 @@ struct TypeChecker<'tcx, 'a> {
     local_decls: &'a IndexVec<Local, LocalDecl<'tcx>>,
     current_location: Location,
     adt_metadata: &'a AdtMetadataTable<'tcx>,
-    c_void_casts: &'a CVoidCasts<'tcx>,
 }
 
 impl<'tcx> TypeChecker<'tcx, '_> {
@@ -433,7 +432,6 @@ pub fn visit_body<'tcx>(
         local_decls: &mir.local_decls,
         current_location: Location::START,
         adt_metadata,
-        c_void_casts,
     };
 
     for (block, bb_data) in mir.basic_blocks().iter_enumerated() {
--- a/c2rust-analyze/src/c_void_casts.rs
+++ b/c2rust-analyze/src/c_void_casts.rs
@@ -2,7 +2,7 @@ use std::borrow::Borrow;
 use std::collections::HashMap;
 
 use rustc_middle::{
-    mir::{Body, LocalDecls, Place, Statement, Terminator, TerminatorKind},
+    mir::{BasicBlock, Body, LocalDecls, Location, Place, Statement, Terminator, TerminatorKind, StatementKind},
     ty::{TyCtxt, TyKind},
 };
 
@@ -59,7 +59,7 @@ impl CVoidCastDirection {
 /// It is checked to be a [`*c_void`] upon construction.
 ///
 /// [`*c_void`]: core::ffi::c_void
-#[derive(PartialEq, Eq, Hash, Clone, Copy)]
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
 pub struct CVoidPtr<'tcx> {
     place: Place<'tcx>,
 }
@@ -135,6 +135,7 @@ impl<'tcx> Borrow<Place<'tcx>> for CVoidPtr<'tcx> {
     }
 }
 
+#[derive(Clone)]
 /// A cast to/from a [`*c_void`](core::ffi::c_void) to/from a properly typed pointer.
 pub struct CVoidCast<'tcx> {
     /// The [`*c_void`](core::ffi::c_void) side of the cast.
@@ -158,12 +159,12 @@ pub struct CVoidCast<'tcx> {
 /// to their properly typed pointers,
 /// like [`CVoidCasts`], but in a single direction, meaning it represents either
 /// [`CVoidCastDirection::From`] or [`CVoidCastDirection::To`].
-#[derive(Default)]
-pub struct CVoidCastsUniDirectional<'tcx>(HashMap<CVoidPtr<'tcx>, Place<'tcx>>);
+#[derive(Default, Clone, Debug)]
+pub struct CVoidCastsUniDirectional<'tcx>(HashMap<Location, (CVoidPtr<'tcx>, Place<'tcx>)>);
 
 impl<'tcx> CVoidCastsUniDirectional<'tcx> {
-    pub fn contains(&self, place: Place<'tcx>) -> bool {
-        self.0.contains_key(&place)
+    pub fn contains(&self, loc: &Location) -> bool {
+        self.0.contains_key(loc)
     }
 
     /// Get the adjusted [`Place`], skipping over [`*c_void`](core::ffi::c_void) intermediaries.
@@ -171,12 +172,16 @@ impl<'tcx> CVoidCastsUniDirectional<'tcx> {
     /// That is, if `place` is a [`CVoidPtr`] in this map of [`CVoidCast`]s,
     /// then the [`Place`] of its other, property-typed pointer is returned.
     /// Otherwise, the same `place` is returned, as no adjustments are necessary.
-    pub fn get_adjusted_place_or_default_to(&self, place: Place<'tcx>) -> Place<'tcx> {
-        *self.0.get(&place).unwrap_or(&place)
+    pub fn get_adjusted_place_or_default_to(
+        &self,
+        loc: &Location,
+        place: Place<'tcx>,
+    ) -> Place<'tcx> {
+        *self.0.get(loc).map(|(void, subst)| subst).unwrap_or(&place)
     }
 
-    pub fn insert(&mut self, cast: CVoidCast<'tcx>) {
-        self.0.insert(cast.c_void_ptr, cast.other_ptr);
+    pub fn insert(&mut self, loc: Location, cast: CVoidCast<'tcx>) {
+        self.0.insert(loc, (cast.c_void_ptr, cast.other_ptr));
     }
 }
 
@@ -216,7 +221,7 @@ impl<'tcx> CVoidCastsUniDirectional<'tcx> {
 /// _1 = ...;
 /// free(_1);
 /// ```
-#[derive(Default)]
+#[derive(Default, Clone)]
 pub struct CVoidCasts<'tcx> {
     from: CVoidCastsUniDirectional<'tcx>,
     to: CVoidCastsUniDirectional<'tcx>,
@@ -245,16 +250,17 @@ impl<'tcx> CVoidCasts<'tcx> {
     /// See [`CVoidCastsUniDirectional::get_adjusted_place_or_default_to`].
     pub fn get_adjusted_place_or_default_to(
         &self,
+        loc: &Location,
         direction: CVoidCastDirection,
         place: Place<'tcx>,
     ) -> Place<'tcx> {
         self.direction(direction)
-            .get_adjusted_place_or_default_to(place)
+            .get_adjusted_place_or_default_to(loc, place)
     }
 
     /// See [`CVoidCastsUniDirectional::insert`].
-    fn insert(&mut self, direction: CVoidCastDirection, cast: CVoidCast<'tcx>) {
-        self.direction_mut(direction).insert(cast)
+    fn insert(&mut self, loc: Location, direction: CVoidCastDirection, cast: CVoidCast<'tcx>) {
+        self.direction_mut(direction).insert(loc, cast)
     }
 
     /// Determine if this [`Statement`] should be skipped
@@ -274,16 +280,11 @@ impl<'tcx> CVoidCasts<'tcx> {
     /// [`Cast`]: rustc_middle::mir::Rvalue::Cast
     /// [`From`]: CVoidCastDirection::From
     /// [`To`]: CVoidCastDirection::To
-    pub fn should_skip_stmt(&self, stmt: &Statement<'tcx>) -> bool {
-        let (lhs, rv) = match get_assign_sides(stmt) {
-            None => return false,
-            Some(it) => it,
-        };
-        self.to.contains(lhs)
-            || self.from.contains(match get_cast_place(rv) {
-                None => return false,
-                Some(it) => it,
-            })
+    pub fn should_skip_stmt(&self, loc: &Location) -> bool {
+        eprintln!("{loc:?}");
+        eprintln!("to: {:?}", self.to);
+        eprintln!("from: {:?}", self.from);
+        self.to.contains(loc) || self.from.contains(loc)
     }
 
     /// Insert all applicable [`*c_void`] casts
@@ -303,7 +304,7 @@ impl<'tcx> CVoidCasts<'tcx> {
     ///
     /// [`*c_void`]: core::ffi::c_void
     fn insert_all_from_body(&mut self, body: &Body<'tcx>, tcx: TyCtxt<'tcx>) {
-        for bb_data in body.basic_blocks().iter() {
+        for (block, bb_data) in body.basic_blocks().iter().enumerate() {
             let term: &Terminator = match &bb_data.terminator {
                 Some(term) => term,
                 None => continue,
@@ -328,14 +329,25 @@ impl<'tcx> CVoidCasts<'tcx> {
                 body.basic_blocks()[successor_block_id]
                     .statements
                     .iter()
-                    .find_map(get_cast)
+                    .enumerate()
+                    .find(|(_, s)| !matches!(s.kind, StatementKind::StorageDead(_)))
+                    .map(get_cast)
+                    .and_then(|(idx, cast)| Some((idx, cast?)))
             };
 
             // For [`CVoidCastDirection::From`], we only count
             // a cast to `*c_void` from an arbitrary type in the same block,
             // searching backwards.
-            let find_last_cast_curr_block =
-                |get_cast| bb_data.statements.iter().rev().find_map(get_cast);
+            let find_last_cast_curr_block = |get_cast| {
+                bb_data
+                    .statements
+                    .iter()
+                    .enumerate()
+                    .map(get_cast)
+                    .rev()
+                    .filter_map(|(idx, cast)| Some((idx, cast?)))
+                    .next()
+            };
 
             for direction in CVoidCastDirection::from_callee(ty_callee(tcx, func_ty))
                 .iter()
@@ -348,13 +360,26 @@ impl<'tcx> CVoidCasts<'tcx> {
                 };
                 let c_void_ptr = CVoidPtr::checked(c_void_ptr, &body.local_decls, tcx);
                 let get_cast =
-                    move |stmt: &Statement<'tcx>| c_void_ptr.get_cast_from_stmt(direction, stmt);
+                    move |(idx, stmt): (usize, &Statement<'tcx>) | (idx, c_void_ptr.get_cast_from_stmt(direction, stmt));
                 let cast = match direction {
                     From => find_first_cast_succ_block(get_cast),
                     To => find_last_cast_curr_block(get_cast),
                 };
-                if let Some(cast) = cast {
-                    self.insert(direction, cast);
+                if let Some((statement_index, cast)) = cast {
+                    let loc = Location {
+                        statement_index,
+                        block: match direction {
+                            To => BasicBlock::from_usize(block),
+                            From => target.unwrap(),
+                        },
+                    };
+                    self.insert(loc, direction, cast.clone());
+
+                    let loc = Location {
+                        statement_index: bb_data.statements.len(),
+                        block: BasicBlock::from_usize(block),
+                    };
+                    self.insert(loc, direction, cast);
                 }
             }
         }
--- a/c2rust-analyze/src/c_void_casts.rs
+++ b/c2rust-analyze/src/c_void_casts.rs
@@ -138,7 +138,7 @@ impl<'tcx> Borrow<Place<'tcx>> for CVoidPtr<'tcx> {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 /// A cast to/from a [`*c_void`](core::ffi::c_void) to/from a properly typed pointer.
 pub struct CVoidCast<'tcx> {
     /// The [`*c_void`](core::ffi::c_void) side of the cast.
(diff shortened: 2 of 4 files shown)


c2rust-analyze/src/c_void_casts.rs
```diff
@@ -347,14 +362,28 @@ impl<'tcx> CVoidCasts<'tcx> {
                     To => args[0].place().unwrap(),
                 };
                 let c_void_ptr = CVoidPtr::checked(c_void_ptr, &body.local_decls, tcx);
-                let get_cast =
-                    move |stmt: &Statement<'tcx>| c_void_ptr.get_cast_from_stmt(direction, stmt);
+                let get_cast = move |(idx, stmt): (usize, &Statement<'tcx>)| {
```
aneksteind, line 365: @kkysen I don't really like this, and it was a function of needing type annotations for this closure in `find_last_cast_curr_block` for example. I'm open to cleaner ways of going about this.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -440,17 +436,21 @@ pub fn visit_body<'tcx>(
...
-            tc.current_location = Location {
-                block: bb,
+            let loc = Location {
+                block,
                 statement_index: idx,
             };
-            tc.visit_statement(stmt);
+            tc.current_location = loc;
+
+            if !c_void_casts.should_skip_stmt(&loc) {
```
aneksteind, line 447: I did this to avoid passing `loc` into `visit_statement`, but if it's inappropriate to check `c_void_casts` here I can move it into `visit_statement`

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -158,25 +162,33 @@ pub struct CVoidCast<'tcx> {
...
+            .0
+            .get(loc)
+            .map(|(_void, subst)| subst)
+            .unwrap_or(&place)
     }
 
-    pub fn insert(&mut self, cast: CVoidCast<'tcx>) {
-        self.0.insert(cast.c_void_ptr, cast.other_ptr);
+    pub fn insert(&mut self, loc: Location, cast: CVoidCast<'tcx>) {
+        self.0.insert(loc, (cast.c_void_ptr, cast.other_ptr));
```
spernsteiner, line 191: This is maybe excessively cautious, but could you assert here than `loc` isn't already in the map?  This should always be the case since we currently have at most one "from" cast and at most one "to" cast per `Location`.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -158,25 +162,33 @@ pub struct CVoidCast<'tcx> {
...
-        *self.0.get(&place).unwrap_or(&place)
+    pub fn get_adjusted_place_or_default_to(
+        &self,
+        loc: &Location,
+        place: Place<'tcx>,
+    ) -> Place<'tcx> {
+        *self
+            .0
+            .get(loc)
+            .map(|(_void, subst)| subst)
```
spernsteiner, line 186: Similarly, you could assert here than `void == place`

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
-            None => return false,
-            Some(it) => it,
-        };
-        self.to.contains(lhs)
-            || self.from.contains(match get_cast_place(rv) {
-                None => return false,
-                Some(it) => it,
-            })
+    pub fn should_skip_stmt(&self, loc: &Location) -> bool {
+        self.to.contains(loc) || self.from.contains(loc)
```
spernsteiner, line 291: Can you please add a separate field to `CVoidCastsUniDirectional` for tracking "casts to skip", separate from the map for tracking "calls with a cast on the argument/result"?  Right now, an entry in the `CVoidCastsUniDirectional` map means "either this is a call, and has the following cast on its argument/result, OR this is a cast, and it should be skipped".  This will work, but IMO it would be clearer to have separate maps/sets for the two cases.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+    }
+
+    fn is_place_local_modified(p: &Place, stmt: &Statement) -> bool {
+        if !p.projection.is_empty() {
+            return false;
+        }
```
spernsteiner, line 297: What is this function checking, exactly?  I'm not sure whether this first case should return `false` because `p` isn't a local, or `true` because we can't be sure that `stmt` leaves `p` unmodified.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -320,41 +405,44 @@ impl<'tcx> CVoidCasts<'tcx> {
...
-                let c_void_ptr = match direction {
+                let c_void_ptr_place = match direction {
                     From => destination,
                     To => args[0].place().unwrap(),
                 };
-                let c_void_ptr = CVoidPtr::checked(c_void_ptr, &body.local_decls, tcx);
+                let c_void_ptr = CVoidPtr::checked(c_void_ptr_place, &body.local_decls, tcx);
                 let get_cast =
-                    move |stmt: &Statement<'tcx>| c_void_ptr.get_cast_from_stmt(direction, stmt);
+                    |stmt: &Statement<'tcx>| c_void_ptr.get_cast_from_stmt(direction, stmt);
```
spernsteiner, line 419: Since this is the only callback ever passed to `find_*_cast_*_block`, would it make sense to pass `c_void_ptr` and `stmt` to those functions instead, and inline this `get_cast_from_stmt` call?

spernsteiner (approved):

c2rust-analyze/src/c_void_casts.rs
aneksteind: In particular it was meant to check what your [comment](https://github.com/immunant/c2rust/pull/788#discussion_r1113481984) outlined:

> I think we can conservatively treat the Place as unmodified only if (1) it's just a Local, with no projections, and (2) that local isn't mentioned in the LHS or RHS of any statements between the cast and the call.

c2rust-analyze/src/c_void_casts.rs
aneksteind: I don't mind either way, I'll play around with it to see if one appears to be better than the other.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -158,25 +162,33 @@ pub struct CVoidCast<'tcx> {
 /// to their properly typed pointers,
 /// like [`CVoidCasts`], but in a single direction, meaning it represents either
 /// [`CVoidCastDirection::From`] or [`CVoidCastDirection::To`].
-#[derive(Default)]
-pub struct CVoidCastsUniDirectional<'tcx>(HashMap<CVoidPtr<'tcx>, Place<'tcx>>);
+#[derive(Default, Clone, Debug)]
+pub struct CVoidCastsUniDirectional<'tcx>(HashMap<Location, (CVoidPtr<'tcx>, Place<'tcx>)>);
```
kkysen, line 166: `(CVoidPtr<'tcx>, Place<'tcx>)` can just be `CVoidCast<'tcx>`.  They were only split before due to being the key and value of the `HashMap`.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -158,25 +162,33 @@ pub struct CVoidCast<'tcx> {
...
 
 impl<'tcx> CVoidCastsUniDirectional<'tcx> {
-    pub fn contains(&self, place: Place<'tcx>) -> bool {
-        self.0.contains_key(&place)
+    pub fn contains(&self, loc: &Location) -> bool {
+        self.0.contains_key(loc)
     }
```
kkysen, line 171: For here and other places, you can just take `Location` by value as it's `Copy`.

c2rust-analyze/src/c_void_casts.rs
kkysen: Can you add that as docs?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+        false
+    }
+
+    fn find_first_cast_succ_block<F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>>(
+        mut get_cast: F,
+        statements: &[Statement<'tcx>],
+    ) -> Option<(usize, CVoidCast<'tcx>)> {
+        // For [`CVoidCastDirection::From`], we only count
+        // a cast from `*c_void` to an arbitrary type in the subsequent block,
+        // searching forward.
```
kkysen, line 340: Make this rustdoc.  It was only a local comment before since it was in a closure.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+    }
+
+    fn find_last_cast_curr_block<F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>>(
+        mut get_cast: F,
+        statements: &[Statement<'tcx>],
+        place: &Place<'tcx>,
+    ) -> Option<(usize, CVoidCast<'tcx>)> {
+        // For [`CVoidCastDirection::From`], we only count
+        // a cast to `*c_void` from an arbitrary type in the same block,
+        // searching backwards.
```
kkysen, line 362: Same for these docs.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+        }
+
+        None
+    }
+
+    fn find_last_cast_curr_block<F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>>(
+        mut get_cast: F,
+        statements: &[Statement<'tcx>],
+        place: &Place<'tcx>,
+    ) -> Option<(usize, CVoidCast<'tcx>)> {
```
kkysen, line 359:
```suggestion
    fn find_last_cast_curr_block<F>(
        mut get_cast: F,
        statements: &[Statement<'tcx>],
        place: &Place<'tcx>,
    ) -> Option<(usize, CVoidCast<'tcx>)>
    where
        F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>,
    {
```

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+                Cast(_, op, _) => op.place(),
+                // BinaryOp
+                // CheckedBinaryOp
+                // NullaryOp
+                UnaryOp(_, op) => op.place(),
+                Discriminant(p) => Some(p),
+                // Aggregate
+                ShallowInitBox(op, _) => op.place(),
+                _ => None,
+            };
```
kkysen, line 324: If you're naming all the variants by comments anyways, just include them as actual arms and no `_ =>`.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+
+    fn is_place_local_modified(p: &Place, stmt: &Statement) -> bool {
+        if !p.projection.is_empty() {
+            return false;
+        }
+
+        let local = p.local;
+
+        if let StatementKind::Assign(asgn) = stmt.kind.clone() {
+            let (lhs, rv) = *asgn;
```
kkysen, line 302: nit:
```suggestion
        if let StatementKind::Assign(assign) = stmt.kind.clone() {
            let (lhs, rv) = *assign;
```

c2rust-analyze/src/c_void_casts.rs
kkysen, line 324:
```suggestion
        if let StatementKind::Assign(assign) = &stmt.kind {
            let (lhs, rv) = &**assign;
            if lhs.local == local {
                return true;
            }

            use Rvalue::*;
            let rv_place = match rv {
                Use(op) => op.place(),
                Repeat(op, _) => op.place(),
                Ref(_, _, p) => Some(*p),
                // ThreadLocalRef
                AddressOf(_, p) => Some(*p),
                Len(p) => Some(*p),
                Cast(_, op, _) => op.place(),
                // BinaryOp
                // CheckedBinaryOp
                // NullaryOp
                UnaryOp(_, op) => op.place(),
                Discriminant(p) => Some(*p),
                // Aggregate
                ShallowInitBox(op, _) => op.place(),
                _ => None,
            };
```

Don't need to clone `StatementKind` and its `Box`.

c2rust-analyze/src/c_void_casts.rs
kkysen: The function is different now, though, so can you update the docs accordingly, too?

c2rust-analyze/src/c_void_casts.rs
kkysen: I only added the closure initially to avoid passing `direction` and making the code a bit shorter.  You can switch to whatever's simpler now.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -303,7 +388,7 @@ impl<'tcx> CVoidCasts<'tcx> {
     ///
     /// [`*c_void`]: core::ffi::c_void
     fn insert_all_from_body(&mut self, body: &Body<'tcx>, tcx: TyCtxt<'tcx>) {
-        for bb_data in body.basic_blocks().iter() {
+        for (block, bb_data) in body.basic_blocks().iter().enumerate() {
```
kkysen, line 391:
```suggestion
        for (block, bb_data) in body.basic_blocks().iter_enumerated() {
```

Then you don't need the `BasicBlock::from_usize(block)`s.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -245,16 +257,17 @@ impl<'tcx> CVoidCasts<'tcx> {
...
     ) -> Place<'tcx> {
         self.direction(direction)
-            .get_adjusted_place_or_default_to(place)
+            .get_adjusted_place_or_default_to(loc, place)
     }
 
     /// See [`CVoidCastsUniDirectional::insert`].
-    fn insert(&mut self, direction: CVoidCastDirection, cast: CVoidCast<'tcx>) {
-        self.direction_mut(direction).insert(cast)
+    fn insert(&mut self, loc: Location, direction: CVoidCastDirection, cast: CVoidCast<'tcx>) {
```
kkysen, line 269: I'd switch `loc` and `direction`.  That's the first lookup before `location`.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -245,16 +257,17 @@ impl<'tcx> CVoidCasts<'tcx> {
     /// See [`CVoidCastsUniDirectional::get_adjusted_place_or_default_to`].
     pub fn get_adjusted_place_or_default_to(
         &self,
+        loc: &Location,
```
kkysen, line 260: Same here about switching `loc` and `direction`.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+    }
+
+    fn find_first_cast_succ_block<F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>>(
+        mut get_cast: F,
+        statements: &[Statement<'tcx>],
+    ) -> Option<(usize, CVoidCast<'tcx>)> {
+        // For [`CVoidCastDirection::From`], we only count
+        // a cast from `*c_void` to an arbitrary type in the subsequent block,
+        // searching forward.
+        for (sidx, stmt) in statements.iter().enumerate() {
```
kkysen, line 341: You can use `.iter_enumerated()` to keep the index typed.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+
+    fn find_last_cast_curr_block<F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>>(
+        mut get_cast: F,
+        statements: &[Statement<'tcx>],
+        place: &Place<'tcx>,
+    ) -> Option<(usize, CVoidCast<'tcx>)> {
+        // For [`CVoidCastDirection::From`], we only count
+        // a cast to `*c_void` from an arbitrary type in the same block,
+        // searching backwards.
+        for (sidx, stmt) in statements.iter().enumerate().rev() {
```
kkysen, line 363: `.iter_enumerated()` again.

c2rust-analyze/src/c_void_casts.rs
kkysen: Nevermind, I didn't realize these were just normal slices.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+                continue;
+            }
+            if let Some(cast) = get_cast(stmt) {
+                return Some((sidx, cast));
+            } else {
+                break;
+            }
+        }
+
+        None
```
kkysen, line 352:
```suggestion
        statements
            .iter()
            .enumerate()
            .skip_while(|(_, stmt)| matches!(stmt.kind, StatementKind::StorageDead(_)))
            .find_map(|(i, stmt)| Some((i, get_cast(stmt)?)))
```

I think the intent is a lot clearer in a functional style.  `.skip_while` and `.find_map` work very well here.

kkysen: Written this way, docs are much less needed, though I still think they'd be useful.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+    }
+
+    fn find_last_cast_curr_block<F: FnMut(&Statement<'tcx>) -> Option<CVoidCast<'tcx>>>(
+        mut get_cast: F,
+        statements: &[Statement<'tcx>],
+        place: &Place<'tcx>,
```
kkysen, line 358: `Place` is copy, so you can take it by value.

kkysen (commented):
Can you make sure to update the docs when things are changed, and include new ones for non-trivial functions?  Also, can you put in the PR description what's being fixed?  Probably just Stuart's comment from #788.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -320,41 +405,44 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+                            From => target.unwrap(),
+                        },
+                    };
+                    self.insert(loc, direction, cast.clone());
+
+                    let loc = Location {
+                        statement_index: bb_data.statements.len(),
+                        block: BasicBlock::from_usize(block),
+                    };
+                    self.insert(loc, direction, cast);
```
kkysen, line 445: Can you inline `loc` both times?  It's a bit confusing to reuse the variable name when it's totally different.

Also, why do we need to insert the cast at two `Location`s?  Can you document why that's done?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -320,41 +405,44 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+                        statement_index,
+                        block: match direction {
+                            To => BasicBlock::from_usize(block),
+                            From => target.unwrap(),
+                        },
+                    };
+                    self.insert(loc, direction, cast.clone());
+
+                    let loc = Location {
+                        statement_index: bb_data.statements.len(),
```
kkysen, line 442: What does it mean to have an out-of-bounds `statement_index` for a `Location`?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -320,41 +405,44 @@ impl<'tcx> CVoidCasts<'tcx> {
...
                 .copied()
             {
                 use CVoidCastDirection::*;
-                let c_void_ptr = match direction {
+                let c_void_ptr_place = match direction {
                     From => destination,
                     To => args[0].place().unwrap(),
                 };
-                let c_void_ptr = CVoidPtr::checked(c_void_ptr, &body.local_decls, tcx);
+                let c_void_ptr = CVoidPtr::checked(c_void_ptr_place, &body.local_decls, tcx);
```
kkysen, line 417: This was supposed to shadow the first `c_void_ptr`, now `c_void_ptr_place`.  Why the change?

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
-        self.to.contains(lhs)
-            || self.from.contains(match get_cast_place(rv) {
-                None => return false,
-                Some(it) => it,
-            })
+    pub fn should_skip_stmt(&self, loc: &Location) -> bool {
+        self.to.contains(loc) || self.from.contains(loc)
+    }
+
+    fn is_place_local_modified(p: &Place, stmt: &Statement) -> bool {
```
kkysen, line 294:
```suggestion
    fn is_place_unmodified_by_stmt(p: &Place, stmt: &Statement) -> bool {
```

I think this was meant to be "unmodified", right?  Since Stuart said it's
> unmodified only if (1) it's just a `Local`, with no projections, and
and here it's returning `false` when there is a projection.

Also, I think `is_place_unmodified_by_stmt` reads more clearly.

c2rust-analyze/src/c_void_casts.rs
kkysen: Is this why there are two insertions into `c_void_casts`?  I agree it would be much clearer to separate the two.

c2rust-analyze/src/c_void_casts.rs
```diff
@@ -274,16 +287,88 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+        // searching backwards.
+        for (sidx, stmt) in statements.iter().enumerate().rev() {
+            let cast = get_cast(stmt);
+            if let Some(cast) = cast {
+                return Some((sidx, cast));
+            } else if Self::is_place_local_modified(place, stmt) {
+                return None;
+            }
+        }
+        None
```
kkysen, line 371: Can you explain (in docs, too) what this is doing?

c2rust-analyze/src/c_void_casts.rs
kkysen: I think a function like this in `util.rs` would make things a lot clearer.  I'd've thought it'd be in `rustc` already.

```rust
pub fn terminator_location(block: BasicBlock, block_data: &BasicBlockData) -> Location {
    Location {
        block,
        statement_index: block_data.statements.len(),
    }
}
```

c2rust-analyze/src/c_void_casts.rs
aneksteind: 91bad4f2

c2rust-analyze/src/c_void_casts.rs
aneksteind: c27a916f

c2rust-analyze/src/c_void_casts.rs
aneksteind: b6e492d4

c2rust-analyze/src/c_void_casts.rs
kkysen: nit: you can use `assert_eq!`.

c2rust-analyze/src/c_void_casts.rs
aneksteind: d2cd89bf

c2rust-analyze/src/c_void_casts.rs
aneksteind: 8fcc2a8a

c2rust-analyze/src/c_void_casts.rs
aneksteind: 7fb2d448

c2rust-analyze/src/c_void_casts.rs
aneksteind: 8fcc2a8a

c2rust-analyze/src/c_void_casts.rs
aneksteind: bc23fd71

c2rust-analyze/src/c_void_casts.rs
aneksteind: a8c64a5b

c2rust-analyze/src/c_void_casts.rs
kkysen: These two functions are a bit different now, like this one with skipping the `StorageDead`s at the beginning, can you explain that in the docs?  The other one is a bit more complex, so that one is more in need of updated docs.

c2rust-analyze/src/c_void_casts.rs
aneksteind: 4bb545f4

c2rust-analyze/src/c_void_casts.rs
aneksteind: f62cb76a

c2rust-analyze/src/c_void_casts.rs
aneksteind: inlined in 54830958

c2rust-analyze/src/c_void_casts.rs
aneksteind: f6741d57

c2rust-analyze/src/c_void_casts.rs
aneksteind: 93a30394

c2rust-analyze/src/c_void_casts.rs
aneksteind: 93a30394, it's checking for modification, not unmodification

c2rust-analyze/src/c_void_casts.rs
aneksteind: ed3c93a9

c2rust-analyze/src/c_void_casts.rs
aneksteind: a7ddc515

c2rust-analyze/src/c_void_casts.rs
aneksteind: 65d1bce3

c2rust-analyze/src/c_void_casts.rs
aneksteind: 89fc7fd9

c2rust-analyze/src/c_void_casts.rs
aneksteind: 59dffcc3

c2rust-analyze/src/c_void_casts.rs
aneksteind: 827606fc

PR #850, opened 2023-02-24
(`c2rust-analyze`) Split `Callee::Other` into `Callee::{Local, Unknown}Def` and include `fn` ptrs in `Callee::UnknownDef`

kkysen:
Split `Callee::Other` into `Callee::{Local, Unknown}Def` and include `fn` ptrs in `Callee::UnknownDef`.

Currently, `Callee::Other` is as it sounds, just any other function that is not first recognized to be `Trivial` or a builtin.  The docs for `Callee::Other` say it is statically known, but `extern` functions with statically unknown definitions are still included, and these currently panic when we look up `LFnSig` in `visit_call_other`.

Also, calls to `fn` ptrs are completely unsupported, but `lighttpd` does have a lot of them.

This tightens `Callee::Other` into what it says it is, `Callee::LocalDef`, which is a statically-known function with a known definition in the current (local) crate (functions in other crates don't have an accessible definition).

On the other hand, `Callee::UnknownDef` is added for cases where the function's definition is not known, and is meant to be treated as a potential fallback for any function call. This includes `extern` functions, whose definition may be completely unknown and dynamically linked.

And it also includes `fn` ptrs who may have multiple definitions.  It may be possible to statically determine the set of function definitions it could be, but not always, and doing so would be an optimization over this fallback.

Plus, it also includes normal Rust functions from other crates, as their definitions aren't available.  The vast majority of these should be safe (like from `std`), and so with the expanded `Callee::Trivial` definition, most of them should be handled already.

Also, since we can now handle `FnPtr`s, we also consider them for triviality checking.

This PR also folds `None::<Callee>` into `Callee::UnknownDef`, as it should error.  Previously nothing was done for `None`, which was incorrect.

### DIFF
--- a/c2rust-analyze/src/util.rs
+++ b/c2rust-analyze/src/util.rs
@@ -72,24 +72,6 @@ pub fn describe_rvalue<'tcx>(rv: &Rvalue<'tcx>) -> Option<RvalueDesc<'tcx>> {
 
 #[derive(Debug)]
 pub enum Callee<'tcx> {
-    /// `<*mut T>::offset` or `<*const T>::offset`.
-    PtrOffset {
-        pointee_ty: Ty<'tcx>,
-        mutbl: Mutability,
-    },
-
-    /// `<[T]>::as_ptr` and `<[T]>::as_mut_ptr` methods.  Also covers the array and str versions.
-    SliceAsPtr {
-        /// The pointee type.  This is either `TyKind::Slice`, `TyKind::Array`, or `TyKind::Str`.
-        pointee_ty: Ty<'tcx>,
-
-        /// The slice element type.  For `str`, this is `u8`.
-        elem_ty: Ty<'tcx>,
-
-        /// Mutability of the output pointer.
-        mutbl: Mutability,
-    },
-
     /// A [`Trivial`] library function is one that has no effect on pointer permissions in its caller.
     ///
     /// Thus, a [`Trivial`] function call requires no special handling.
@@ -108,6 +90,24 @@ pub enum Callee<'tcx> {
     ///
     /// [`Trivial`]: Self::Trivial
     Trivial,
+
+    /// `<*mut T>::offset` or `<*const T>::offset`.
+    PtrOffset {
+        pointee_ty: Ty<'tcx>,
+        mutbl: Mutability,
+    },
+
+    /// `<[T]>::as_ptr` and `<[T]>::as_mut_ptr` methods.  Also covers the array and str versions.
+    SliceAsPtr {
+        /// The pointee type.  This is either `TyKind::Slice`, `TyKind::Array`, or `TyKind::Str`.
+        pointee_ty: Ty<'tcx>,
+
+        /// The slice element type.  For `str`, this is `u8`.
+        elem_ty: Ty<'tcx>,
+
+        /// Mutability of the output pointer.
+        mutbl: Mutability,
+    },
 
     /// libc::malloc
     Malloc,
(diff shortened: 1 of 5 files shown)


spernsteiner:
It seems to me that we probably want to treat non-crate-local Rust functions similarly to `extern` functions.  In both cases, we have only a signature, not a body, so we can't analyze or rewrite any part of the function definition.  So I think non-crate-local `FnDef`s should be classified as `UnknownDef`, not `Normal`.

Also, I'm not a huge fan of the name `Normal`.  What do you think about having `Callee::LocalDef` for statically-known functions defined in the local crate and `Callee::UnknownDef` for all other functions (including `FnPtr`s)?

c2rust-analyze/src/util.rs
```diff
@@ -123,36 +148,43 @@ pub enum Callee<'tcx> {
...
+            }
+        }
+        ty::FnPtr(..) => {
+            if is_trivial() {
+                Callee::Trivial
+            } else {
+                Callee::UnknownDef { ty }
+            }
+        }
+        _ => Callee::Trivial,
```
spernsteiner, line 183: >This PR also folds None::<Callee> into Callee::Trivial, as Trivial needs no special handling, which is exactly how None is treated.

I think it was incorrect for us to treat `None` that way - that case should have been a panic instead.  If we don't know what's being called, then I think the right thing to return is `UnknownDef`.

kkysen:
> Also, I'm not a huge fan of the name `Normal`. What do you think about having `Callee::LocalDef` for statically-known functions defined in the local crate and `Callee::UnknownDef` for all other functions (including `FnPtr`s)?

Yeah, I wasn't sure about that name.  `LocalDef` sounds good, or maybe `KnownDef`, though `LocalDef` seems good in the same manner of `LocalDefId`.

c2rust-analyze/src/util.rs
kkysen: Okay, so the fix here would just be 
```suggestion
        _ => Callee::UnknownDef { ty },
```
?

kkysen:
> It seems to me that we probably want to treat non-crate-local Rust functions similarly to `extern` functions. In both cases, we have only a signature, not a body, so we can't analyze or rewrite any part of the function definition. So I think non-crate-local `FnDef`s should be classified as `UnknownDef`, not `Normal`.

That makes sense.  There's no way to get the MIR of those other non-crate-local functions, right?  We just hope the vast majority will be `Trivial` under the new, expanded definition?  This can be determined by `def_id.is_local()`, right?

c2rust-analyze/src/util.rs
spernsteiner: Yes, that looks right to me

kkysen: Fixed in 1a54c3cf4a344ebb9743f3a62ced9a66571fc820.

kkysen:
> It seems to me that we probably want to treat non-crate-local Rust functions similarly to `extern` functions. In both cases, we have only a signature, not a body, so we can't analyze or rewrite any part of the function definition. So I think non-crate-local `FnDef`s should be classified as `UnknownDef`, not `Normal`.

Fixed in d74b48839426ccc19e9776c6db20f84aa255743d.

> Also, I'm not a huge fan of the name `Normal`. What do you think about having `Callee::LocalDef` for statically-known functions defined in the local crate and `Callee::UnknownDef` for all other functions (including `FnPtr`s)?

Fixed in 257708a8a6823ceee653c9382a2eac351bd8ea9e.

aneksteind (approved):

PR #851, opened 2023-02-24
(`c2rust-analyze`) `log::error!` rather than `todo!` on `Callee::UnknownDef` to test skipping it

### LINKED ISSUE #842: (`c2rust-analyze`) Skip non-builtin `extern` function calls
Currently, `c2rust-analyze` tries to visit the bodies of `extern` functions when they're called.  Ultimately, we would like one of two cases to happen:
* properly link `extern`s defined in other modules with `use` imports
* if the `extern` function is not defined somewhere else (e.x. in a dynamic library), default to pessimistic pointer permissions based only on the signature
    * for common ones like from libc, these could be made builtins

But for now, we simply want to stop panicking on these `extern` function calls, so we'll skip them entirely.  Then we'll switch to pessimistic signature-based pointer permissions, and then properly link the functions as an optimization.

kkysen:
Fixes
* #842

`log::error!` rather than `todo!` on `Callee::UnknownDef`.

This allows us to test what happens when these `UnknownDef` calls are not crashed on by setting `RUST_LOG_PANIC=off` (see #854) to disable the normal panic on `log::error!`.

This is incorrect, but will allow us to avoid crashing on a lot more code and see what other issues there may be.

### DIFF
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -280,7 +280,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 match callee {
                     Callee::Trivial => {}
                     Callee::UnknownDef { .. } => {
-                        todo!("visit Callee::{callee:?}");
+                        log::error!("TODO: visit Callee::{callee:?}");
                     }
 
                     Callee::LocalDef { def_id, substs } => {


spernsteiner:
This is the wrong approach to reducing crashes and would almost certainly bite us at some point in the future.

However, for testing purposes, I think it would make sense to have a way to downgrade as many panics as possible into warnings, so the analysis can make progress past unsupported features.  For example, we could define a new `xpanic!(...)` macro that works like `panic!(...)` by default, but when built with `--features debug_dont_panic`, it instead prints a warning and continues on.  Similarly, we could have `xassert!`, `xtodo!`, etc, and replace normal `panic!`/`assert!`/`todo!` with these in places where it makes sense.  (Though the changeover from `panic!` to `xpanic!` should probably be done lazily - if you're doing some testing and you hit a panic, then change that one panic to `xpanic!` if possible, and try again - rather than trying to update every `panic!` in the codebase up front.)

This would still be useful even once we start catching panics during function analysis, since at least for the near future, a panic during analysis of a particular function will cause the tool to skip all further processing of that function and its (transitive) callers.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -280,7 +280,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 match callee {
                     Callee::Trivial => {}
                     Callee::UnknownDef { .. } => {
-                        todo!("visit Callee::{callee:?}");
+                        eprintln!("visit Callee::{callee:?}");
```
aneksteind, line 283: given that `eprintln!` is used everywhere else, could we make it even more obvious that this is unknown and unvisited?

c2rust-analyze/tests/common/mod.rs
```diff
@@ -48,6 +48,18 @@ impl Analyze {
...
+                let len_truncated = output.len() - max_len;
+                let (truncated, rest) = output.split_at(len_truncated);
+                let num_lines_truncated = truncated.split('\n').count();
+                let truncated_msg = format!(
+                    "truncated the first {len_truncated} bytes ({num_lines_truncated} lines)"
+                );
+                format!("[{truncated_msg}] ...\n{rest}\n[note: {truncated_msg}]")
+            } else {
+                output
+            };
```
aneksteind, line 62: why?

kkysen: When I was testing which things to uncomment, if there were errors from `lighttpd-minimal`, it would print 20 million bytes to the terminal, so I added this code to truncate it.  I could remove this code, but it seemed useful and I had to write it anyways to not overload my terminal while testing things.

kkysen:
> This is the wrong approach to reducing crashes and would almost certainly bite us at some point in the future.

That's what I was thinking long term, but David thought it'd be useful to expose other things first/in the meantime.

> However, for testing purposes, I think it would make sense to have a way to downgrade as many panics as possible into warnings, so the analysis can make progress past unsupported features. For example, we could define a new `xpanic!(...)` macro that works like `panic!(...)` by default, but when built with `--features debug_dont_panic`, it instead prints a warning and continues on.

I think having it toggleable at runtime would be better than re-compiling.  An environment variable would work.

`panic!` also has a `!` return type and this allows a lot of things to compile.  One that has `()` as a return type might not always be substitutable.

I'm also wondering if a logging framework would be helpful here.  Do you know if they're able to, at runtime, turn warnings into panicking errors, for example?  Something like that could work and it'd be off-the-shelf.

kkysen:
I think we can use something simple but pre-built like `env_logger` for this.  We can set custom formatting with [`Builder::format`](https://docs.rs/env_logger/0.10.0/env_logger/struct.Builder.html#method.format) and check the [`log::Record::level`] to see if we should panic, based on an environment variable or something.

aneksteind:
> David thought it'd be useful to expose other things first/in the meantime.

To clarify my take on this, I think what @spernsteiner is suggesting is in line with what I'd like to see, which is implementing some way to allow us to move past these forms of static analysis failure and come back to handling them correctly in subsequent stages. Having said that, why do you think it's better to toggle this at runtime and not re-compile?

c2rust-analyze/tests/common/mod.rs
aneksteind: I think it's best if we remove it

kkysen:
> why do you think it's better to toggle this at runtime and not re-compile?

I don't see what the value is in re-compiling.  If it's a compile-time switch, we risk other code paths not being tested (if we don't test it regularly, all of a sudden the feature might not even compile anymore).  And a runtime switch is faster during testing, as we don't need to wait to re-compile.

aneksteind:
> we risk other code paths not being tested (if we don't test it regularly, all of a sudden the feature might not even compile anymore)

I don't see why this wouldn't be the case for any feature gate, dynamic or static

spernsteiner:
If we define `xpanic!(...)` as either `if true { panic!(...) }` or `eprintln!(...)`, then it will return `()` in both configurations, so we shouldn't run into trouble with the less common mode unexpectedly failing to build.

kkysen:
I still think runtime configuration is better.  It's like `RUSTFLAGS="-D warnings"`.  Do you see a downside?

spernsteiner:
I think it's fine either way, but having a few extra recompiles when using a relatively obscure debug feature doesn't seem like a big deal to me, and the compile-time flag seems like it would probably be simpler to implement.

c2rust-analyze/tests/common/mod.rs
kkysen: Removed for now, stored in [kkysen/analyze-tests-limit-output](https://github.com/immunant/c2rust/tree/kkysen/analyze-tests-limit-output) in case we ever need it.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: It should be clearer with the `log::error!` and `[ERROR @ ...]` printed now, I think.

spernsteiner (approved):
Looks good to merge after #854

PR #856, opened 2023-03-01
Support Aggregate literals in static analysis

aneksteind:


aneksteind:
@spernsteiner this is currently failing:

```
thread 'rustc' panicked at 'called `Result::unwrap()` on an `Err` value: SourceNotAvailable { filename: Real(Remapped { local_path: None, virtual_name: "/rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/library/core/src/ptr/mod.rs" }) }', c2rust-analyze/src/main.rs:716:57
```

```rust
fn describe_span(tcx: TyCtxt, span: Span) -> String {
    let s = tcx.sess.source_map().span_to_snippet(span).unwrap();
    ...
```

Changing `uwrap()` to `unwrap_or_default` allows the tests to pass but I'm sure there must be some better solution. What do you think?

spernsteiner:
>Changing uwrap() to unwrap_or_default() allows the tests to pass but I'm sure there must be some better solution. What do you think?

Looks like this is coming from a macro expansion - I looked up `ptr/mod.rs:1997` and it's the body of `ptr::addr_of!`.  So one option for handling this would be to walk up the `Span`'s macro expansion stack (maybe using `Span::parent_callsite` or `Span::source_callsite` - I'm not sure exactly what's the right thing to call here) until we find one that's in the local crate or at least has source available.

c2rust-analyze/src/labeled_ty.rs
```diff
@@ -220,6 +220,47 @@ impl<'tcx, L: Copy> LabeledTyCtxt<'tcx, L> {
...
+    /// (usually this means the caller is providing the wrong list of type arguments as `substs`).
+    ///
+    /// This produces a [`LabeledTy`] with the right structure and also substitutes the
+    /// underlying [`Ty`]s
+    pub fn subst_full(
+        &self,
+        lty1: LabeledTy<'tcx, L>,
+        lty2: LabeledTy<'tcx, L>,
+        substs: &[LabeledTy<'tcx, L>],
+    ) -> LabeledTy<'tcx, L> {
```
spernsteiner, line 234: This approach seems quite tricky - we should fix `subst` instead so that there's no need to mess with two different `LTy`s when doing a substitution.  I think the strategy for that is:

1. Convert `substs` to a `SubstsRef` (discarding labels/args and keeping only each `x.ty`)
2. Do the actual `Ty` substitution with `EarlyBinder(lty.ty).subst(...)` (based on [`subst_and_normalize_erasing_regions`](https://github.com/rust-lang/rust/blob/7c306f6dcd600b1fcc74439c780e8ebef338d84c/compiler/rustc_middle/src/ty/normalize_erasing_regions.rs#L133-L156))
3. Traverse and update `args` as is currently done in `LabeledTyCtxt::subst` (preferably with some `debug_assert`s in there to ensure the substituted `ty` has the expected shape to match the `args`)

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -282,13 +282,102 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        by traversing the struct definition. Additionally, the label for the
+                        `Operand` corresponding to `s` has an `Origin('1)` (because it is
+                        a reference). Because the `OriginArgs` are missing for the field, the
+                        labeled types for the field operand and for the field declaration do not
+                        have the same shape, and so `do_assign` cannot create subset relations.
+
+                        Additionally, the regions referenced within the operand types are
+                        erased, and so it also would not be possible to know which ADT
+                        OriginParams those correspond to because there would need to be a
+                        non-erased region to compare with.
```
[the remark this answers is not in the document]
aneksteind: @spernsteiner how would you feel about me creating an issue for supporting this and generics? AFAICT this isn't going to be a consideration until we want to incrementally lift (I don't see any transpiled code that uses generics in lighttpd)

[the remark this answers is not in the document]
spernsteiner: Putting off generics for now is fine, but I think having a separate function for computing the field type would still make the logic below much clearer.

c2rust-analyze/src/labeled_ty.rs
aneksteind: `subst` doesn't seem to be available, maybe I'm missing something:

```
no method named `subst` found for struct `EarlyBinder` in the current scope
   --> c2rust-analyze/src/labeled_ty.rs:238:37
    |
238 |                 *t = EarlyBinder(t).subst(self.tcx, substs);
    |                                     ^^^^^ method not found in `EarlyBinder<&mut rustc_middle::ty::Ty<'tcx>>`
```

I found its definition in `subst.rs`, though.

c2rust-analyze/src/borrowck/type_check.rs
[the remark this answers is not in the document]
aneksteind: I'm a bit confused about the refactoring feedback. I can imagine a way to break out the substitution into its own function but don't see the vision of refactoring specifically to return a relabeled field type; this block doesn't relabel field types and performs substitution on the RHS of the assignment, not the LHS / `foo` which should already have the appropriate `Origins` from the labeling that occurs in `visit_place` in `adt_func`. What am I missing?

c2rust-analyze/src/labeled_ty.rs
kkysen: There's a `Subst` trait you have to import.

c2rust-analyze/src/borrowck/type_check.rs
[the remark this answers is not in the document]
spernsteiner: I didn't look at the implementation here in detail, but I was expecting it to follow the same general structure as in `dataflow`, where we effectively unroll the aggregate assignment into a series of field assignments.  For example, we would treat `s = S { x: a, y: b }` as if it were `s.x = a; s.y = b;`.  So the logic for each field would be:
1. From the struct type (which in this case I think is `expect_ty`), obtain the type of the field
2. Get the type of the value being stored in the field (here it's `op_label` - this should maybe be called `op_lty`, since it's the whole `LTy`, not just the label)
3. Pseudo-assign the value type into the field type

It sounds like what you're describing is the other way around:

>performs substitution on the RHS of the assignment, not the LHS / foo

It sounds like the idea here is to handle `s = S { x: a, y: b }` by building up an `LTy` for `S { x: a, y: b }` based on the `LTy`s for `a` and `b` and then pseudo-assigning that `LTy` into the type of `s`.  But I don't think this is possible in general.  Suppose we have the following setup:

```Rust
struct S<'a> {
    x: &'a i32,
    y: &'a u8,
}
let a: &'o1 i32 = ...;
let b: &'o2 u8 = ...;
let s: S<'o3> = S { x: a, y: b };
```
There's no single origin to pick for `'a` in the type of `S { x: a, y: b }` - picking `S<'o1>` means ignoring the interaction with `'o2` and vice versa.

c2rust-analyze/src/labeled_ty.rs
aneksteind: @spernsteiner let me know if 5db7e8d8 doesn't fit the requirement

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -173,6 +173,21 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                        for (field, op) in fields.iter().zip(ops.iter()) {
+                            let op_lty = self.acx.type_of(op);
+                            let unresolved_field_lty = self.acx.gacx.field_tys[&field.did];
+                            // resolve the generic type arguments in `field_lty` by referencing the `Ty` of `op`
+                            let resolved_field_lty =
+                                self.acx
+                                    .lcx()
+                                    .subst_full(unresolved_field_lty, op_lty, lty.args);
+                            // Pseudo-assign from each operand to the element type of the field.
+                            self.do_assign(resolved_field_lty, op_lty);
```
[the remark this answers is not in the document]
aneksteind: Any action items here? A comment about the no-op perhaps?

c2rust-analyze/src/borrowck/type_check.rs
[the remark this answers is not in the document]
aneksteind: > There's no single origin to pick for 'a in the type of S { x: a, y: b } - picking S<'o1> means ignoring the interaction with 'o2 and vice versa.

You're right, this is a bug. But `visit_statement` calls `do_assign` on the LHS and RHS, which also assigns the `Origins` of the lifetime parameters. It seems like the concrete origins for the `Aggregate` literal on the RHS need to be present, so how can I reconcile that with the sort of example you posted above where the fields have two different concrete origins but there is only one lifetime parameter for the struct? Should there be a different origin `'o3` that outlives `'o1` and `'o2`? Should that sort of assignment not happen at all?

I realized there was never really any assignment between fields of the LHS and RHS, so I've factored out the function and called the assignment in 249d6a55.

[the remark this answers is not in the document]
spernsteiner: One option is to take `expect_lty` to be the type of the struct literal.  Then you would use that as the base for the field projections, pseudo-assign `a` into `proj(expect_lty, x)` and `b` into `proj(expect_lty, y)`, and finally return `expect_lty` as the overall type of the rvalue.

>Should there be a different origin `'oX` that outlives `'o1` and `'o2`?

This is the other option, which matches what we do in `dataflow` with `rvalue_tys`.  We would label each `Rvalue` up front, including generating fresh `Origin`s for all its real and hypothetical region arguments.  Then we would pseudo-assign `a` into `proj(rvalue_lty, x)` etc, and return `rvalue_lty` to be pseudo-assigned into the LHS type (which is equal to `expect_lty`).

However, I'm not sure if this extra indirection is necessary here.  We use it in `dataflow` because it makes it easier to attach casts to the individual field values later on, but for `borrowck` we don't have such a requirement.  So it's probably okay to go with the first option unless you actually run into trouble with that approach.

c2rust-analyze/src/labeled_ty.rs
```diff
@@ -224,30 +224,28 @@ impl<'tcx, L: Copy> LabeledTyCtxt<'tcx, L> {
...
+    /// This produces a [`LabeledTy`] with the right structure and also substitutes the
+    /// underlying [`Ty`]s
+    pub fn subst(&self, lty: LabeledTy<'tcx, L>, substs: SubstsRef<'tcx>) -> LabeledTy<'tcx, L> {
+        let substituted_ty = EarlyBinder(lty.ty).subst(self.tcx, substs);
+
         if let TyKind::Param(ref ty) = lty.ty.kind() {
-            if let Some(p) = substs.get(ty.index as usize) {
-                return p;
+            if substs.get(ty.index as usize).is_some() {
+                return self.mk(substituted_ty, &[], lty.label);
```
spernsteiner, line 234: This case is incorrect - it produces an `LTy` with empty `args` even if `substituted_ty` ought to have some args.  I think this should return `p` as before, but also assert that `p.ty` matches `substituted_ty`.

c2rust-analyze/src/labeled_ty.rs
```diff
@@ -224,30 +224,28 @@ impl<'tcx, L: Copy> LabeledTyCtxt<'tcx, L> {
...
             }
         }
 
-        self.mk(lty.ty, self.subst_slice(lty.args, substs), lty.label)
+        let substituted_args = self.subst_slice(lty.args, substs);
+        debug_assert_eq!(substituted_args.len(), lty.args.len());
```
spernsteiner, line 239: This assert only checks that `self.subst_slice` returns a slice of the same length as its input, which is not a mistake I think we're likely to make.

The sort of bug I want to defend against is this: suppose we call rustc's `subst` and get `(i32, u8)`, and then we do substitution on our args and get the slice `[u8, i32]`.  This might happen due to some unforeseen in the way rustc's `subst` works, or due to a mismatch between the way our `subst` and rustc's `subst` handles `SubstRef`s.  Now we have an `LTy` where `lty.ty` is `(i32, u8)` but `lty.args` is `[u8, i32]` - this is not a valid `LTy`.

Maybe the right way to go about this is to write a function that checks that an `LTy` is well-formed - that is, check that `lty.args[i].ty` is equal to the i'th "argument" (as defined by `LabeledTyCtxt::label`) of the underlying `lty.ty` - and `debug_assert!` before returning from this function that the `LTy` we're returning really is well-formed.

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
spernsteiner: No action needed - just pointing out that this will be fine once `subst_full`/`subst` is fixed.

c2rust-analyze/src/labeled_ty.rs
aneksteind: p is no longer and LTy, though, because the type of substs has changed.

spernsteiner: Ah, I missed that change.  The argument here should still be a list of `LTy`s, and it should build the `SubstsRef` internally.  The point of this function is to do things like substitute `*mut /*ptr1*/ u32` into `Vec<T>` to get `Vec<*mut /*ptr1*/ u32>`, so the inputs still need to be labeled.

c2rust-analyze/src/borrowck/type_check.rs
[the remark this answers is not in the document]
aneksteind: Ok, I've pushed option 1 in 9b9dc7a

c2rust-analyze/src/labeled_ty.rs
aneksteind: addressed in a7152bc

c2rust-analyze/src/borrowck/type_check.rs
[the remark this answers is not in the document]
aneksteind: And made a small change in 1e8d88d

c2rust-analyze/src/labeled_ty.rs
aneksteind: I think this should be covered [here](https://github.com/immunant/c2rust/commit/a7152bc6957456c8e6da4f4857e1c67ab7e62316#diff-85be9e434a05c5ae5841fcd74ecf82b89c062c10a4dbd858135f4c7781007d40R248) but let me know if I've gotten that wrong

spernsteiner: The latest `subst` implementation in this PR still produces wrong results.  However, when I tried to fix it myself, I discovered that the `subst` API is also broken - the `lty` and `substs` arguments it currently receives aren't sufficient to perform the right substitution.  The issue is that the index in `TyKind::Param` is an index into the full generic argument list, which includes not only types but also regions and consts, while `substs` is a list covering only the types, and we don't know how indices into the two lists match up.  For example, if `substs = [A, B]` and we see a `Param` with index 1, it might be that we're substituting into `struct Foo<T, U>` where `Param(1)` is `U` and we should return the type `B`, or it might be that we're substituting into `struct<'a, T, U>` where `Param(1)` is `T` and we should return the type `A`.

I think the thing to do here might be to revert to the old `subst` implementation (which is a bit simpler) but add an assert that `substs.len() == 0` so that the current bugs can't be triggered.  Then we can fix it in a separate PR after figuring out exactly what information it needs and how to get it.

aneksteind: OBE in 2cc47277

c2rust-analyze/src/labeled_ty.rs
aneksteind: No longer used in d8876a7a

c2rust-analyze/src/labeled_ty.rs
aneksteind: No longer used in d8876a7a

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -37,18 +37,20 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-        let mut adt_func = |base_lty: LTy<'tcx>, base_adt_def: AdtDef, field: Field| {
-            let base_origin_param_map: IndexMap<OriginParam, Origin> =
-                IndexMap::from_iter(base_lty.label.origin_params.to_vec());
-            let field_def: &FieldDef = &base_adt_def.non_enum_variant().fields[field.index()];
-            let perm = self.field_permissions[&field_def.did];
-            let base_metadata = &self.adt_metadata.table[&base_adt_def.did()];
-            let field_metadata = &base_metadata.field_info[&field_def.did];
-
-            self.ltcx.relabel(
+    pub fn projection_lty(
```
spernsteiner, line 40: Should probably be named `field_lty` or something like that, since it only works for fields, not projections in general.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -173,6 +173,16 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                         }
                     }
+                    AggregateKind::Adt(adt_did, ..) => {
+                        let base_adt_def = self.acx.tcx().adt_def(adt_did);
+                        let fields = &base_adt_def.non_enum_variant().fields;
+                        for (field, op) in fields.iter().zip(ops.iter()) {
+                            let op_lty = self.acx.type_of(op);
+                            let field_lty = self.acx.gacx.field_tys[&field.did];
+                            // Pseudo-assign from each operand to the element type of the field.
+                            self.do_assign(field_lty, op_lty);
```
spernsteiner, line 183: We know the unsubstituted `field_lty` is the wrong thing to use here in general.  Please at least add a TODO comment (or better yet, a relevant assert), or else put the `subst` call back in (it's fine if it panics when `args.len() > 0` for now, until we fix `subst` properly).

c2rust-analyze/src/main.rs
```diff
@@ -467,6 +467,13 @@ fn run(tcx: TyCtxt) {
                             let args = acx.lcx().mk_slice(&[elem_lty]);
                             acx.lcx().mk(array_ty, args, PointerId::NONE)
                         }
+                        AggregateKind::Adt(_, _, substs, ..) => {
+                            let args = acx.lcx().mk_slice(acx.lcx().arena().alloc_from_iter(
+                                substs.types().map(|sub_ty| acx.assign_pointer_ids(sub_ty)),
+                            ));
+                            let adt_ty = rv.ty(&acx, acx.tcx());
+                            acx.lcx().mk(adt_ty, args, PointerId::NONE)
```
spernsteiner, line 475: Is this any different from just doing `acx.assign_pointer_ids(adt_ty)`?  I see the `AggregateKind::Array` case works the same way as this one, but that case I think could also be simplified (`acx.assign_pointer_ids(array_ty)`).

spernsteiner (commented):
A few minor comments - otherwise this looks good at this point

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: renamed in 7de4be76

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: renamed in 7de4be76

c2rust-analyze/src/main.rs
aneksteind: addressed in 09e255f3

aneksteind:
@spernsteiner would you mind taking another look at this when you have time?

spernsteiner (approved):

PR #858, opened 2023-03-06
Don't attempt to evaluate constant values that don't exist

dberlin:
ASTExporter attempts to convert constant int expressions to actual int values through evaluation.

However, even when evaluation fails, it still attempts to convert the result into an APint, leading to an assertion in APInt (due to trying to convert a none value to APInt).

Fixed by respecting the result of evaluation and exiting if it says no evaluated value was produced.


fw-immunant:
Hi, thanks for the PR! Could you provide a test case that triggers the behavior you're describing?

If it's too much trouble to navigate the project's testing infrastructure to add a test case, feel free to just post some code and I'll integrate it into our tests.

PR #859, opened 2023-03-06
Handle two argument libc builtin functions properly

### LINKED ISSUE #857: builtin handling for string functions expects wrong arguments
Attempts to convert strchr/strrchr (and some others in that list) will fail with a panic in convert_libc_fns, say "`convert_libc_fns` must have exactly 3 arguments: [dst, c, len]""

Length does not exist, strchr/strrchr is a 2 argument function :)
It actually should just work fine if you translate the 2 argument ones straight through.

I'll make a patch.

dberlin:
This fixes #857 by passing through 2 argument libc builtin functions.
Tested on builtin_strrchr.


kkysen (approved):
LGTM

PR #864, opened 2023-03-14
Add dataflow and borrowck support for null pointers derived from cast from 0

aneksteind:


kkysen:
I don't think we need to handle it yet, since the transpiler doesn't currently output it, but we'll probably want to also handle `core::ptr::{null,null_mut}` along with this, as that's the correct way of doing the `0 as *{const,mut} T` casts that the transpiler probably should emit instead, so just leaving this as a note.

aneksteind:
@kkysen is there any reason why that wouldn't be handled properly in at least the `dataflow` and `borrowck` portions?

kkysen:
> @kkysen is there any reason why that wouldn't be handled properly in at least the `dataflow` and `borrowck` portions?

It's an `Callee::UnknownDef` now, so we need to make it a builtin `Callee` and then handle it properly.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -271,7 +271,50 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                 })
             }
-
+            Rvalue::Cast(CastKind::PointerFromExposedAddress, ref op, _ty) => {
+                // We support only one case here, which is the case of null pointers
+                // constructed via casts such as `0 as *const T`
+                if let Some(0) = op
+                    .constant()
+                    .and_then(|c| c.literal.try_to_scalar_int())
+                    .and_then(|s| s.try_to_machine_usize(self.tcx).ok())
```
kkysen, line 280: See #866.  I'm not sure this is totally correct and there seems to be a simpler implementation possible.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -328,6 +328,35 @@ fn run_polonius<'tcx>(
...
+                .map(|o| {
+                    let pairing = (*o, amaps.origin());
+                    eprintln!("pairing lifetime parameter with origin: {pairing:?}");
+                    pairing
+                })
+                .collect()
+        })
+        .unwrap_or_default();
+
+    ltcx.arena().alloc_from_iter(origins)
```
kkysen, line 357:
```suggestion
    let default = Default::default();
    let origins = adt_metadata
        .table
        .get(&adt_def.did())
        .map_or(&default, |adt| &adt.lifetime_params)
        .iter()
        .map(|origin| (*origin, amaps.origin()))
        .inspect(|pairing| eprintln!("pairing lifetime parameter with origin: {pairing:?}"));
    ltcx.arena().alloc_from_iter(origins)
```

Don't need the `.collect::<Vec<_>>()`.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: Utilizing your changes now

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -271,7 +271,46 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                     Label::default()
                 })
             }
-
+            Rvalue::Cast(CastKind::PointerFromExposedAddress, ref op, _ty) => {
+                // We support only one case here, which is the case of null pointers
+                // constructed via casts such as `0 as *const T`
+                if let Some(true) = op.constant().cloned().map(util::is_null_const) {
+                    let adt_metadata = &construct_adt_metadata(self.tcx);
+                    self.ltcx.relabel(expect_ty, &mut |lty| {
```
spernsteiner, line 279: I had to think about why relabeling `expect_ty` is a reasonable thing to do here.  I think it works but is a bit of a hack.  Labeling `ty` seems more correct, but the problem is we don't know what `perms` to set for it (we particularly care about `UNIQUE` here).  The correct solution is probably to add an `rvalue_tys` mechanism for borrowck similar to what we have for dataflow, but the approach here should work for now.

Can you add a comment noting that we're only extracting `perms` from `expect_ty`, and otherwise this is conceptually similar to labeling the cast target `ty`?

spernsteiner (approved):
This looks generally plausible to me.  I guess the dataflow changes are so simple because we aren't tracking `NON_NULL` permissions yet.

For borrowck, I guess the strategy for `0 as *mut T` is to generate a fresh origin with no loans associated with it.  Have you checked how this compares to what rustc actually does for cases like `None::<&mut T>` (the rewritten equivalent)?

aneksteind:
> Have you checked how this compares to what rustc actually does for cases like None::<&mut T> (the rewritten equivalent)?

@spernsteiner AFAICT, polonius generates no new loans in this case, but many subset relations. There are 5 regions introduced in `use_of_var_derefs_origin.facts` for a contrived example `let x: Option<&S<&u32, &u16, &u32, &i32>> = None;`, which aligns with my expectations, and subset relations between those and other regions (presumably from the RHS type). I believe that is in line what this PR introduces, but please correct me if I am wrong.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: done in ef0a754e

PR #868, opened 2023-03-17
(`c2rust-analyze`) Clarify some points in `PermissionSet` docs

kkysen:
This clarifies and elaborates on some points in the `PermissionSet` docs.

These were things that I found confusing working through the semantics of other PRs, so I just wanted to confirm that I'm understanding them correctly now and in preparation for other PRs like #818.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -26,6 +26,10 @@ bitflags! {
     /// as opposed to something like `ALIASED` (a pointer capability),
     /// as removing [`UNIQUE`] (`&mut`) allows more values to be taken on (`&`).
     ///
+    /// Currently, we assume that all pointers are valid
+    /// (see [the `std::ptr` safety docs](https://doc.rust-lang.org/std/ptr/index.html#safety)).
+    /// We do not yet (here) consider null, unaligned, or cast-from-integer pointers.
+    ///
     /// [`UNIQUE`]: Self::UNIQUE
     #[derive(Default)]
     pub struct PermissionSet: u16 {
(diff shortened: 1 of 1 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -26,27 +26,45 @@ bitflags! {
     /// That's why, for example [`UNIQUE`] is named as such,
     /// as opposed to something like `ALIASED` (a pointer capability),
     /// as removing [`UNIQUE`] (`&mut`) allows more values to be taken on (`&`).
+    ///
+    /// Furthermore, when "can" is used for a permission,
+    /// that does not mean that the pointer is always capable of, being [`READ`] for example.
+    /// That is, "can be read" means might be capable of being read,
+    /// or in other words, there are cases in which this pointer will be read.
```
spernsteiner, line 33: I don't think this is accurate.  If the pointer is `READ` then we will rewrite it into `&T` (or similar), which asserts to rustc that it is, in fact, valid and readable, regardless of whether it ever actually gets read/dereferenced.

This will change slightly when `NON_NULL` is added, since `READ` without `NON_NULL` will mean "either readable or null" (those being the valid states of a type like `Option<&T>`).

kkysen: I was basing this on the example you gave in https://github.com/immunant/c2rust/pull/818#discussion_r1099038252.

kkysen: This part has been removed now.  I think I was thinking ahead to when `NON_NULL` and `VALID` would be permissions, too, in which case this would apply (e.x. `READ` doesn't mean it's readable, as it could be null, only `READ | NON_NULL` does).

spernsteiner (approved):

PR #869, opened 2023-03-17
(`c2rust-analyze/tests`) Switch from simple directives to `clap`-parsed args for analyze tests

kkysen:
Switch from simple directives to `clap`-parsed args for analyze tests.

This also adds the `--env` arg, which is needed to set `RUST_LOG_PANIC` in #855 to allow its tests to work.  This is more complex than just implementing logic like setting `RUST_LOG_PANIC` inline in a `#[test] fn` in `c2rust-analyze/tests/filecheck.rs`, but has the benefit of keeping all the testing parameters within the test file itself, as @spernsteiner pointed out.

This switches the previous simple directives (just boolean flags, `,`-separated) to full `clap`- and `shlex`-parsed args because we need arg values and so that we don't waste any time re-implementing parsing.  It's also very easy to just add a `//! --help` to see all the options for args now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,7 +129,7 @@ dependencies = [
  "anyhow",
  "c2rust-build-paths",
  "c2rust-transpile",
- "clap",
+ "clap 3.2.23",
  "env_logger",
  "git-testament",
  "is_executable",
@@ -156,12 +156,14 @@ dependencies = [
  "assert_matches",
  "bitflags",
  "c2rust-build-paths",
+ "clap 4.1.9",
  "env_logger",
  "indexmap",
  "log",
  "polonius-engine",
  "print_bytes",
  "rustc-hash",
+ "shlex",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


aneksteind:
why does this have a base associated with #835?

kkysen:
> why does this have a base associated with #835?

It's meant to build on top of it.

aneksteind:
> It's meant to build on top of it.

@kkysen does it need to? If not, let's set the base to `master` and adjust accordingly

kkysen:
Hmm, I guess you're right.  It doesn't necessarily need #835.  That said, it makes rebasing more complex if the PR stack has to keep changing, as now #855 is based on this one, but also has parts based on #835.

spernsteiner (approved):
Massive overkill, but it'll work.

aneksteind (changes requested):
please base on `master`. Any merges into master would affect rebasing to some extent, I don't feel like that difficulty makes a valid reason to merge this into a PR that is already under scrutiny and may not get merged.

kkysen:
Okay, I'll rebase on `master`, as soon as I figure out how to do that (a simple `git rebase master` doesn't work).

aneksteind (approved):

Issue #874, opened 2023-03-27
(`c2rust-analyze`) Remove unsound `Trivial`ity


kkysen:
After a lot of discussion in #855, we've determined that the previous `Trivial`ity is unsound, and other attempts to create a correct and more general solution to classifying and handling `Callee::UnknownDef` calls were extremely difficult in determining what the correct semantics would be.  As `lighttpd` has now been compiled as an amalgamation (https://github.com/immunant/lighttpd-rust/pull/2, #871), this removes the vast majority of `Callee::UnknownDef` calls that we were previously trying to handle.  Thus, the only calls that are left are `std`/`core` and `libc` calls.  For these, to get optimal semantics, we want to implement something like #865, so we might as well abandon a conservative but general solution and focus on #865.  Since the previous `Trivial`ity was unsound, that will now be removed.

PR #876, opened 2023-03-29
c2rust-analyze: initial implementation of error recovery

spernsteiner:
This branch adds a basic implementation of error recovery: if the static analysis panics on some function, it now catches the panic, marks that function as invalid, and continues with analysis of the remaining functions.  At the end, after completing as much analysis and rewriting as it could, the tool prints a report of all the panics it encountered, including a stack trace for each one.

Currently, when marking a function as failed/invalid, we immediately mark all of its (transitive) callers invalid as well.  In the future we can probably relax this, falling back on something like `Trivial` or on dynamic analysis results to let the callers reason about the unanalyzed function.

The current approach produces a lot of invalid Rust code.  Suppose `f` calls `g`, `f` fails analysis, but `g` succeeds and gets rewritten.  Currently we don't rewrite `f` at all due to the previous failure, and if `g`'s signature gets rewritten, `f`'s call to `g` will no longer be valid.  In the future we'll need a better scheme for calling back and forth between analyzed and unanalyzed functions.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -277,8 +277,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
     pub fn_callers: HashMap<DefId, Vec<DefId>>,
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
-    /// `DefId`s of functions where analysis failed, and a `String` explaining the reason for each
-    /// failure.
+    /// `DefId`s of functions where analysis failed, and a [`PanicDetail`] explaining the reason
+    /// for each failure.
     pub fns_failed: HashMap<DefId, PanicDetail>,
 
     pub field_ltys: HashMap<DefId, LTy<'tcx>>,
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -26,7 +26,7 @@ use rustc_middle::ty::{
 };
 use rustc_type_ir::RegionKind::{ReEarlyBound, ReStatic};
 use std::collections::HashMap;
-use std::fmt::{Debug, Display};
+use std::fmt::Debug;
 use std::ops::Index;
 
 bitflags! {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -25,7 +25,7 @@ use rustc_middle::ty::{
 };
 use rustc_type_ir::RegionKind::{ReEarlyBound, ReStatic};
 use std::collections::HashMap;
-use std::fmt::Debug;
+use std::fmt::{Debug, Display};
 use std::ops::Index;
 
 bitflags! {
@@ -276,6 +276,9 @@ pub struct GlobalAnalysisCtxt<'tcx> {
     pub fn_callers: HashMap<DefId, Vec<DefId>>,
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
+    /// `DefId`s of functions where analysis failed, and a `String` explaining the reason for each
+    /// failure.
+    pub fns_failed: HashMap<DefId, String>,
 
     pub field_ltys: HashMap<DefId, LTy<'tcx>>,
 
@@ -527,6 +530,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             ptr_info: GlobalPointerTable::empty(),
             fn_callers: HashMap::new(),
             fn_sigs: HashMap::new(),
+            fns_failed: HashMap::new(),
             field_ltys: HashMap::new(),
             static_tys: HashMap::new(),
             addr_of_static: HashMap::new(),
@@ -572,6 +576,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             ref mut ptr_info,
             fn_callers: _,
             ref mut fn_sigs,
+            fns_failed: _,
             ref mut field_ltys,
             ref mut static_tys,
             ref mut addr_of_static,
@@ -623,6 +628,24 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
     pub fn assign_pointer_to_fields(&mut self, did: DefId) {
         for field in self.tcx.adt_def(did).all_fields() {
             self.assign_pointer_to_field(field);
+        }
+    }
+
+    pub fn fn_failed(&mut self, did: DefId) -> bool {
+        self.fns_failed.contains_key(&did)
+    }
+
+    pub fn mark_fn_failed(&mut self, did: DefId, reason: impl Display) {
+        if self.fns_failed.contains_key(&did) {
+            return;
+        }
+
+        self.fns_failed.insert(did, reason.to_string());
+
+        // This is the first time marking `did` as failed, so also mark all of its callers.
+        let callers = self.fn_callers.get(&did).cloned().unwrap_or(Vec::new());
+        for caller in callers {
+            self.mark_fn_failed(caller, format_args!("analysis failed on callee {:?}", did));
         }
     }
 }
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -1,6 +1,7 @@
 use crate::borrowck::{AdtMetadata, FieldMetadata, OriginArg, OriginParam};
 use crate::c_void_casts::CVoidCasts;
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
+use crate::panic_detail::PanicDetail;
 use crate::pointer_id::{
     GlobalPointerTable, LocalPointerTable, NextGlobalPointerId, NextLocalPointerId, PointerTable,
     PointerTableMut,
@@ -278,7 +279,7 @@ pub struct GlobalAnalysisCtxt<'tcx> {
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
     /// `DefId`s of functions where analysis failed, and a `String` explaining the reason for each
     /// failure.
-    pub fns_failed: HashMap<DefId, String>,
+    pub fns_failed: HashMap<DefId, PanicDetail>,
 
     pub field_ltys: HashMap<DefId, LTy<'tcx>>,
 
@@ -635,17 +636,20 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         self.fns_failed.contains_key(&did)
     }
 
-    pub fn mark_fn_failed(&mut self, did: DefId, reason: impl Display) {
+    pub fn mark_fn_failed(&mut self, did: DefId, detail: PanicDetail) {
         if self.fns_failed.contains_key(&did) {
             return;
         }
 
-        self.fns_failed.insert(did, reason.to_string());
+        self.fns_failed.insert(did, detail);
 
         // This is the first time marking `did` as failed, so also mark all of its callers.
         let callers = self.fn_callers.get(&did).cloned().unwrap_or(Vec::new());
         for caller in callers {
-            self.mark_fn_failed(caller, format_args!("analysis failed on callee {:?}", did));
+            self.mark_fn_failed(
+                caller,
+                PanicDetail::new(format!("analysis failed on callee {:?}", did)),
+            );
         }
     }
 }
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -272,6 +272,9 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     ptr_info: GlobalPointerTable<PointerInfo>,
 
+    /// Map from a function to all of its callers.
+    pub fn_callers: HashMap<DefId, Vec<DefId>>,
+
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
     pub field_ltys: HashMap<DefId, LTy<'tcx>>,
@@ -522,6 +525,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx,
             lcx: LabeledTyCtxt::new(tcx),
             ptr_info: GlobalPointerTable::empty(),
+            fn_callers: HashMap::new(),
             fn_sigs: HashMap::new(),
             field_ltys: HashMap::new(),
             static_tys: HashMap::new(),
@@ -566,6 +570,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx: _,
             lcx,
             ref mut ptr_info,
+            fn_callers: _,
             ref mut fn_sigs,
             ref mut field_ltys,
             ref mut static_tys,
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -49,6 +49,7 @@ mod dataflow;
 mod equiv;
 mod labeled_ty;
 mod log;
+mod panic_detail;
 mod pointer_id;
 mod rewrite;
 mod trivial;
@@ -411,7 +412,7 @@ fn run(tcx: TyCtxt) {
         let (dataflow, equiv_constraints) = match r {
             Ok(x) => x,
             Err(e) => {
-                gacx.mark_fn_failed(ldid.to_def_id(), panic_to_string(e));
+                gacx.mark_fn_failed(ldid.to_def_id(), panic_detail::catch(&e));
                 continue;
             }
         };
@@ -534,7 +535,7 @@ fn run(tcx: TyCtxt) {
             match r {
                 Ok(()) => {}
                 Err(e) => {
-                    gacx.mark_fn_failed(ldid.to_def_id(), panic_to_string(e));
+                    gacx.mark_fn_failed(ldid.to_def_id(), panic_detail::catch(&e));
                     continue;
                 }
             }
@@ -615,7 +616,7 @@ fn run(tcx: TyCtxt) {
         match r {
             Ok(()) => {}
             Err(e) => {
-                gacx.mark_fn_failed(ldid.to_def_id(), panic_to_string(e));
+                gacx.mark_fn_failed(ldid.to_def_id(), panic_detail::catch(&e));
                 continue;
             }
         }
@@ -666,11 +667,32 @@ fn run(tcx: TyCtxt) {
     rewrite::apply_rewrites(tcx, all_rewrites);
 
     // Report errors that were caught previously
+    eprintln!("\nerror details:");
     for ldid in tcx.hir().body_owners() {
-        if let Some(reason) = gacx.fns_failed.get(&ldid.to_def_id()) {
-            eprintln!("analysis of {:?} failed: {:?}", ldid, reason);
+        if let Some(detail) = gacx.fns_failed.get(&ldid.to_def_id()) {
+            if !detail.has_backtrace() {
+                continue;
+            }
+            eprintln!("\nerror in {:?}:\n{}", ldid, detail.to_string_full());
         }
     }
+
+    eprintln!("\nerror summary:");
+    for ldid in tcx.hir().body_owners() {
+        if let Some(detail) = gacx.fns_failed.get(&ldid.to_def_id()) {
+            eprintln!(
+                "analysis of {:?} failed: {}",
+                ldid,
+                detail.to_string_short()
+            );
+        }
+    }
+
+    eprintln!(
+        "\nsaw errors in {} / {} functions",
+        gacx.fns_failed.len(),
+        all_fn_ldids.len()
+    );
 }
 
 trait AssignPointerIds<'tcx> {
@@ -897,20 +919,6 @@ fn for_each_callee(tcx: TyCtxt, ldid: LocalDefId, f: impl FnMut(LocalDefId)) {
     CalleeVisitor { tcx, mir, f }.visit_body(mir);
 }
 
-fn panic_to_string(e: Box<dyn Any + Send + 'static>) -> String {
-    let e = match e.downcast::<&'static str>() {
-        Ok(s) => return s.to_string(),
-        Err(e) => e,
-    };
-
-    let e = match e.downcast::<String>() {
-        Ok(s) => return *s,
-        Err(e) => e,
-    };
-
-    format!("unknown error: {:?}", e.type_id())
-}
-
 struct AnalysisCallbacks;
 
 impl rustc_driver::Callbacks for AnalysisCallbacks {
@@ -928,6 +936,7 @@ impl rustc_driver::Callbacks for AnalysisCallbacks {
 
 fn main() -> rustc_interface::interface::Result<()> {
     init_logger();
+    panic::set_hook(Box::new(panic_detail::panic_hook));
     let args = env::args().collect::<Vec<_>>();
     rustc_driver::RunCompiler::new(&args, &mut AnalysisCallbacks).run()
 }
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -382,7 +382,7 @@ fn run(tcx: TyCtxt) {
 
         let mut acx = gacx.function_context(&mir);
 
-        let r = panic::catch_unwind(AssertUnwindSafe(|| {
+        let r = panic_detail::catch_unwind(AssertUnwindSafe(|| {
             // Assign PointerIds to local types
             assert!(acx.local_tys.is_empty());
             acx.local_tys = IndexVec::with_capacity(mir.local_decls.len());
@@ -412,8 +412,8 @@ fn run(tcx: TyCtxt) {
 
         let (dataflow, equiv_constraints) = match r {
             Ok(x) => x,
-            Err(e) => {
-                gacx.mark_fn_failed(ldid.to_def_id(), panic_detail::catch(&e));
+            Err(pd) => {
+                gacx.mark_fn_failed(ldid.to_def_id(), pd);
                 continue;
             }
         };
@@ -519,7 +519,7 @@ fn run(tcx: TyCtxt) {
             let acx = gacx.function_context_with_data(&mir, info.acx_data.take());
             let mut asn = gasn.and(&mut info.lasn);
 
-            let r = panic::catch_unwind(AssertUnwindSafe(|| {
+            let r = panic_detail::catch_unwind(AssertUnwindSafe(|| {
                 // `dataflow.propagate` and `borrowck_mir` both run until the assignment converges
                 // on a fixpoint, so there's no need to do multiple iterations here.
                 info.dataflow.propagate(&mut asn.perms_mut());
@@ -535,8 +535,8 @@ fn run(tcx: TyCtxt) {
             }));
             match r {
                 Ok(()) => {}
-                Err(e) => {
-                    gacx.mark_fn_failed(ldid.to_def_id(), panic_detail::catch(&e));
+                Err(pd) => {
+                    gacx.mark_fn_failed(ldid.to_def_id(), pd);
                     continue;
                 }
             }
@@ -597,7 +597,7 @@ fn run(tcx: TyCtxt) {
 
         eprintln!();
 
-        let r = panic::catch_unwind(AssertUnwindSafe(|| {
+        let r = panic_detail::catch_unwind(AssertUnwindSafe(|| {
             let hir_body_id = tcx.hir().body_owned_by(ldid);
             let expr_rewrites = rewrite::gen_expr_rewrites(&acx, &asn, &mir, hir_body_id);
             let ty_rewrites = rewrite::gen_ty_rewrites(&acx, &asn, &mir, ldid);
@@ -616,8 +616,8 @@ fn run(tcx: TyCtxt) {
         }));
         match r {
             Ok(()) => {}
-            Err(e) => {
-                gacx.mark_fn_failed(ldid.to_def_id(), panic_detail::catch(&e));
+            Err(pd) => {
+                gacx.mark_fn_failed(ldid.to_def_id(), pd);
                 continue;
             }
         }
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -300,11 +300,24 @@ fn run(tcx: TyCtxt) {
 
     // Follow a postorder traversal, so that callers are visited after their callees.  This means
     // callee signatures will usually be up to date when we visit the call site.
-    let all_fn_ldids = fn_body_owners_postorder(tcx);
+    let (all_fn_ldids, fn_callers) = fn_body_owners_postorder(tcx);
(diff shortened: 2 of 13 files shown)


kkysen:
> In the future we can probably relax this, falling back on something like `Trivial`

What do you mean by this?  I was planning on entirely removing `Trivial`ity (#874) as the current version is unsound and we were having trouble determining a sound version, and I was just going to hardcode the `libc` functions.

spernsteiner:
>We also need to change the existing feature-specific tests to check that there were no caught panics

Which tests are these?  I'd prefer to fix it in this PR so `master` isn't put into an inconsistent state.

>What do you mean by this? I was planning on entirely removing `Trivial`

I haven't fully thought this through, but I think we may want a way to approximate the behavior of a function `g` on its caller `f` even if `g` can't be fully analyzed.  Otherwise, a failure in `g` would mean we can't analyze `f` at all.  I'm not sure what soundness guarantees, if any, we can provide in the case where analysis has panicked on some functions; if we can't provide guarantees regardless, then this approximation wouldn't need to be sound.

spernsteiner:
Panic reports now include a source span within the code being processed, if one is known.  The output format looks like this:
```
panic at c2rust-analyze/src/context.rs:546:9: unexpected pointer type in *mut *mut _IO_FILE
related location: <rustc_middle::mir::syntax::Operand as c2rust_analyze::context::TypeOf>::type_of::h618b9ae9e6b3fd09 @ /.../c2rust-analyze/src/context.rs:537:41
source location: /.../lighttpd_rust_amalgamated//src/main.rs:91380:13: 91380:19 (#0)
   0: c2rust_analyze::panic_detail::panic_hook
             at src/panic_detail.rs:72:14
   1: core::ops::function::Fn::call
             at /rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/library/core/src/ops/function.rs:77:5
   2: std::panicking::rust_panic_with_hook
             at /rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/library/std/src/panicking.rs:702:17
   3: ...
```
This makes it easier to understand the context in which the panic occurs.

c2rust-analyze/src/main.rs
```diff
@@ -440,76 +442,92 @@ fn run(tcx: TyCtxt) {
...
-                LocalKind::Var | LocalKind::Temp => acx.assign_pointer_ids(decl.ty),
-                LocalKind::Arg => {
-                    debug_assert!(local.as_usize() >= 1 && local.as_usize() <= mir.arg_count);
-                    lsig.inputs[local.as_usize() - 1]
-                }
-                LocalKind::ReturnPointer => lsig.output,
-            };
-            let l = acx.local_tys.push(lty);
-            assert_eq!(local, l);
+        let r = panic::catch_unwind(AssertUnwindSafe(|| {
```
aneksteind, line 457: please use a more descriptive name

c2rust-analyze/src/main.rs
```diff
@@ -440,76 +442,92 @@ fn run(tcx: TyCtxt) {
     // computed during this the process is kept around for use in later passes.
     let mut global_equiv = GlobalEquivSet::new(gacx.num_pointers());
     for &ldid in &all_fn_ldids {
+        // The function might already be marked as failed if one of its callees previously failed.
+        if gacx.fn_failed(ldid.to_def_id()) {
```
aneksteind, line 446: would it be much extra work to run this in a fixed point loop instead? that way we can avoid the problem if `g` calling `f` where `f` is rewritten but `g` isn't and we get a compiler error. I can take that on if you feel like it's reasonable approach and can be in a separate PR regardless of whether or not that's the approach to take

c2rust-analyze/tests/filecheck/catch_panic.rs
```diff
@@ -0,0 +1,29 @@
...
+// Analysis of `bad` should fail because it calls an unsupported library function involving raw
+// pointers, namely `NonNull::as_ptr`.
+// CHECK-NOT: final labeling for "bad"
+unsafe fn bad(p: NonNull<u8>) {
+    *p.as_ptr() = 1;
+}
+
+// Analysis of `bad2` should also fail because it has a callee on which analysis failed.
+// CHECK-NOT: final labeling for "bad2"
+unsafe fn bad2(p: NonNull<u8>) {
```
aneksteind, line 20: Thoughts on adding a test asserting the functionality we know to be broken, e.g. that `good2` defined before `bad3` where `bad3` calls `good2`, will not produce any of the error messages below? That way we can develop against this test case later on

c2rust-analyze/src/panic_detail.rs
```diff
@@ -136,3 +147,24 @@ fn panic_to_string(e: &(dyn Any + Send + 'static)) -> String {
...
+    static CURRENT_SPAN: Cell<Span> = Cell::new(DUMMY_SP);
+}
+
+pub struct CurrentSpanGuard {
+    old: Span,
+}
+
+impl Drop for CurrentSpanGuard {
+    fn drop(&mut self) {
+        CURRENT_SPAN.with(|cell| cell.set(self.old));
```
aneksteind, line 161: am I interpreting this correctly as a funky RAII that replaces the old span when parent scope is entered?

c2rust-analyze/src/context.rs
```diff
@@ -182,6 +193,27 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             self.assign_pointer_to_field(field);
         }
     }
+
+    pub fn fn_failed(&mut self, did: DefId) -> bool {
+        self.fns_failed.contains_key(&did)
+    }
+
+    pub fn mark_fn_failed(&mut self, did: DefId, detail: PanicDetail) {
```
aneksteind, line 201: I feel like I must be missing something; where is this called on any failing code? I only see the recursive call below

c2rust-analyze/src/main.rs
spernsteiner: What's the change you're proposing here?  Currently failure is propagated "up" the callgraph (from callees to callers) in `AnalysisCtxt::mark_fn_failed`.  This part doesn't require a fixpoint, just a depth-first traversal of the callgraph.

The "easy" way to avoid having calls between rewritten and nonrewritten functions is to propagate failure in both directions, but (assuming there's no dead code) that means a failure anywhere will propagate to all parts of the program (first up the callgraph into `main`, then down the callgraph into everything else).

c2rust-analyze/src/panic_detail.rs
spernsteiner: Yes, `set_current_span(x)` sets `CURRENT_SPAN` to `x`, and then when the guard goes out of scope, it resets `CURRENT_SPAN` to whatever value it held before (assuming you follow the normal pattern `let _g = set_current_span(x);` and don't move/swap/forget the guards).

c2rust-analyze/src/main.rs
aneksteind: I think you're right, and I hadn't considered the "easy" way either. Nevermind

kkysen:
> > We also need to change the existing feature-specific tests to check that there were no caught panics
> 
> Which tests are these? I'd prefer to fix it in this PR so `master` isn't put into an inconsistent state.

Basically all of them.  They all currently check that `c2rust-analyze` doesn't panic, but now the meaning of panicking changes, so those tests should be changed to check that `c2rust-analyze` has 0 caught panics.

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+    None
+}
+
+fn panic_to_string(e: &(dyn Any + Send + 'static)) -> String {
+    match e.downcast_ref::<&'static str>() {
+        Some(s) => return s.to_string(),
+        None => {}
+    }
```
kkysen, line 141:
```suggestion
    if let Some(s) = e.downcast_ref::<&'static str>() {
        return s.to_string();
    }
```

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    match e.downcast_ref::<String>() {
+        Some(s) => return (*s).clone(),
+        None => {}
+    }
```
kkysen, line 146:
```suggestion
    if let Some(s) = e.downcast_ref::<String>() {
        return (*s).clone()
    }
```

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+use std::panic::PanicInfo;
+
+/// Detailed information about a panic.
+#[derive(Clone, Debug)]
+pub struct PanicDetail {
+    msg: String,
+    loc: Option<String>,
+    relevant_loc: Option<String>,
+    backtrace: Option<Backtrace>,
+    sp: Span,
```
kkysen, line 16: Can this just be `span`?

c2rust-analyze/src/context.rs
```diff
@@ -182,6 +193,27 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
+        if self.fns_failed.contains_key(&did) {
+            return;
+        }
+
+        self.fns_failed.insert(did, detail);
+
+        // This is the first time marking `did` as failed, so also mark all of its callers.
+        let callers = self.fn_callers.get(&did).cloned().unwrap_or(Vec::new());
```
kkysen, line 209:
```suggestion
        let callers = self.fn_callers.get(&did).cloned().unwrap_or_default();
```

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl Drop for CurrentSpanGuard {
+    fn drop(&mut self) {
+        CURRENT_SPAN.with(|cell| cell.set(self.old));
+    }
+}
+
+/// Set the current span.  Returns a guard that will reset the current span to its previous value
+/// on drop.
+pub fn set_current_span(sp: Span) -> CurrentSpanGuard {
```
kkysen, line 167: This is very neat.

Would `enter_current_span` or just `enter_span` be a bit clearer (similar to Python's `__enter__` and `__exit__`)?

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+    match e.downcast_ref::<String>() {
+        Some(s) => return (*s).clone(),
+        None => {}
+    }
+
+    format!("unknown error: {:?}", e.type_id())
+}
+
+thread_local! {
+    static CURRENT_SPAN: Cell<Span> = Cell::new(DUMMY_SP);
```
kkysen, line 152:
```suggestion
    static CURRENT_SPAN: Cell<Span> = Default::default;
```

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+impl PanicDetail {
+    /// Create a new `PanicDetail` containing only a message, with no location or backtrace
+    /// information.
+    pub fn new(msg: String) -> PanicDetail {
+        PanicDetail {
+            msg,
+            loc: None,
+            relevant_loc: None,
+            backtrace: None,
+            sp: DUMMY_SP,
```
kkysen, line 28:
```suggestion
            sp: Default::default(),
```

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
+use backtrace::Backtrace;
+use log::warn;
+use rustc_span::{Span, DUMMY_SP};
+use std::any::Any;
+use std::cell::Cell;
+use std::fmt::Write as _;
```
kkysen, line 6: I didn't realize you could do this for a trait; neat!

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Get the [`PanicDetail`] of the most recent panic.  This clears the internal storage, so if this
+/// is called twice in a row without an intervening panic, the second call always returns `None`.
+pub fn take_current() -> Option<PanicDetail> {
+    CURRENT_PANIC_DETAIL.with(|cell| cell.take())
+}
+
+/// Get the [`PanicDetail`] of the most recent panic; if it's not available, build a placeholder
+/// `PanicDetail` from the panic payload `e`.  This is useful in conjunction with
+/// [`std::panic::catch_unwind`].
+pub fn catch(e: &(dyn Any + Send + 'static)) -> PanicDetail {
```
kkysen, line 95: Could we just define a `panic_detail::catch_unwind` that wraps `std::panic::catch_unwind` with this?

c2rust-analyze/src/main.rs
```diff
@@ -849,6 +946,7 @@ impl rustc_driver::Callbacks for AnalysisCallbacks {
 
 fn main() -> rustc_interface::interface::Result<()> {
     init_logger();
+    panic::set_hook(Box::new(panic_detail::panic_hook));
```
kkysen, line 949: What happens if there is a panic outside out of the `panic::catch_unwind`s?  Then we don't get the normal panic that prints the error message and panics, right, but we want that?

I think we should either have a solution similar to the span guard, where we set and then remove the panic hook only during/immediately surrounding a `panic::catch_unwind`.  Or have a top-level `panic::catch_unwind` that carries out the default panic behavior.  Otherwise, we'll start getting silent panics, and that seems very bad.

kkysen (commented):
Could 059628af03c7a7de937947284ec315d64eeaaeb0 and 86dfd302ec6797fd1565a7d45f4181b7f5fca469 be split out into a separate PR and merged immediately?  Those 2 commits look good to me and are independent of the panicking stuff in the rest of this PR.

c2rust-analyze/src/context.rs
```diff
@@ -182,6 +193,27 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
+    pub fn fn_failed(&mut self, did: DefId) -> bool {
+        self.fns_failed.contains_key(&did)
+    }
+
+    pub fn mark_fn_failed(&mut self, did: DefId, detail: PanicDetail) {
+        if self.fns_failed.contains_key(&did) {
+            return;
+        }
+
+        self.fns_failed.insert(did, detail);
```
kkysen, line 206: Do we always want to skip a 2nd error?  Could the 2nd error be a panicking error, and the 1st one just be a propagated one?  In that case, the panicking error would be more useful to keep.

kkysen: We could also just store all of the `PanicDetails`.

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// Detailed information about a panic.
+#[derive(Clone, Debug)]
+pub struct PanicDetail {
+    msg: String,
+    loc: Option<String>,
+    relevant_loc: Option<String>,
+    backtrace: Option<Backtrace>,
+    sp: Span,
+}
```
kkysen, line 17: Could we separate `PanicDetail` (an actual panic occurred) and `FnFailure` (either a panic occurred or a called function failed)?

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,170 @@
...
+    static CURRENT_PANIC_DETAIL: Cell<Option<PanicDetail>> = Cell::new(None);
+}
+
+/// Panic hook for use with [`std::panic::set_hook`].  This builds a `PanicDetail` for each panic
+/// and stores it for later retrieval by [`take_current`].
+pub fn panic_hook(info: &PanicInfo) {
+    let bt = Backtrace::new();
+    let detail = PanicDetail {
+        msg: panic_to_string(info.payload()),
+        loc: info.location().map(|l| l.to_string()),
```
kkysen, line 75: I think it's fine to do `info.location().unwrap()`.  Currently, it always returns `Some`, and `std` `.unwrap()`s, too.  That would simplify a bunch of the logic.

c2rust-analyze/src/context.rs
spernsteiner: This only removes a type hint that might be useful for the reader.

c2rust-analyze/src/panic_detail.rs
spernsteiner: This is again less clear for the reader.  Rustc also mainly uses `DUMMY_SP` instead of `default()` for spans.

c2rust-analyze/src/panic_detail.rs
spernsteiner: I don't see much value in doing it this way.

c2rust-analyze/src/context.rs
kkysen: This is a `clippy::or_fun_call` warning.  I don't see a good reason not to follow it.  The fact that it's `Vec` isn't important here.

c2rust-analyze/src/panic_detail.rs
spernsteiner: Functions named `enter_foo` or `with_foo` usually take a closure and set up some kind of context for the duration of that closure.  Here we're using a guard because it's less obtrusive at call sites and we're not likely to screw up the nesting.

c2rust-analyze/src/panic_detail.rs
spernsteiner: Good idea - done

c2rust-analyze/src/main.rs
spernsteiner: Good point.  I've fixed it to call the default panic hook if a panic occurs outside our `panic_detail::catch_unwind` wrapper.

c2rust-analyze/tests/filecheck/catch_panic.rs
spernsteiner: I added a new `bad_call_good` function, which fails analysis and also calls `good`, and a check that this does not result in the failure propagating from `bad_call_good` to its callee `good`.  But I'm not sure if that's what you were suggesting.

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,205 @@
...
+        let old = cell.replace(PanicState::OutsideCatchUnwind);
+        match old {
+            PanicState::OutsideCatchUnwind => {
+                // No special behavior is needed.  Call the default panic hook instead.
+                default_hook(info);
+                return;
+            }
+            PanicState::InsideCatchUnwind => {}
+            PanicState::Unwinding(pd) => {
+                warn!("discarding old panic detail: {:?}", pd);
```
kkysen, line 95: Would this happen if there's a panic during unwinding, for example?  In that case, it might be better to use the original `PanicDetail`, which would be the real error we care about, and just warn about a panic happening during unwinding.  Or maybe it's just best to abort here.

c2rust-analyze/src/panic_detail.rs
```diff
@@ -0,0 +1,205 @@
...
+            .as_ref()
+            .or(self.loc.as_ref())
+            .map_or("[unknown]", |s| &*s);
+        format!("{}: {}", loc_str, self.msg.trim())
+    }
+
+    /// Return a full description of this panic, including a complete backtrace if available.
+    pub fn to_string_full(&self) -> String {
+        let mut s = String::new();
+        let loc_str = self.loc.as_ref().map_or("[unknown]", |s| &*s);
```
kkysen, line 50:
```suggestion
            .map_or("[unknown]", |s| s);
        format!("{}: {}", loc_str, self.msg.trim())
    }

    /// Return a full description of this panic, including a complete backtrace if available.
    pub fn to_string_full(&self) -> String {
        let mut s = String::new();
        let loc_str = self.loc.as_ref().map_or("[unknown]", |s| s);
```

A `clippy::borrow_deref_ref` warning.  The `&*` isn't necessary for a reborrow.

c2rust-analyze/src/main.rs
```diff
@@ -956,6 +1054,10 @@ impl rustc_driver::Callbacks for AnalysisCallbacks {
 
 fn main() -> rustc_interface::interface::Result<()> {
     init_logger();
+    let default_panic_hook = panic::take_hook();
+    panic::set_hook(Box::new(move |info| {
+        panic_detail::panic_hook(&default_panic_hook, info)
+    }));
```
kkysen, line 1060: Could this just be moved to a `panic_detail::set_hook` function?  There shouldn't be other ways of calling it.

c2rust-analyze/src/main.rs
```diff
@@ -729,21 +784,31 @@ fn run(tcx: TyCtxt) {
...
+                "\ngenerated {} expr rewrites + {} ty rewrites for {:?}:",
+                expr_rewrites.len(),
+                ty_rewrites.len(),
+                name
+            );
+            for &(span, ref rw) in expr_rewrites.iter().chain(ty_rewrites.iter()) {
+                eprintln!("  {}: {}", describe_span(tcx, span), rw);
+            }
+            all_rewrites.extend(expr_rewrites);
+            all_rewrites.extend(ty_rewrites);
```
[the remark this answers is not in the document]
spernsteiner: I don't think that would change anything.  This already gives the behavior we want, which is to add entries to `all_rewrites` only if all rewrite generation for this function was successful.

c2rust-analyze/src/panic_detail.rs
spernsteiner: Panicking during unwinding already triggers an immediate abort, so I don't think we need any special handling here.  This case should be impossible, so I'll change the `warn!` to a `panic!`.

kkysen: Oh, perfect!  That should work then.

c2rust-analyze/src/main.rs
spernsteiner: I don't think a descriptive name is necessary here, since the value is used only once, immediately after it's defined.

spernsteiner:
> > > We also need to change the existing feature-specific tests to check that there were no caught panics
> > 
> > 
> > Which tests are these?
> 
> Basically all of them.

I've updated the `tests/analyze/` test cases to disable the panic-catching machinery.  The filecheck tests leave it enabled, since a panic there is likely to cause the analysis to produce wrong output, which FileCheck will detect.  @kkysen, let me know if this isn't what you meant.

c2rust-analyze/src/panic_detail.rs
spernsteiner: > Could we separate `PanicDetail` (an actual panic occurred) and `FnFailure` (either a panic occurred or a called function failed)?

I think it's okay to leave this for now, since the "called function failed" will (hopefully) be going away some time soon.

>we could separate the data itself with the formatting/displaying

I'm not entirely clear on what you're suggesting, but this also seems fine to me as-is.  Such implementation details of the `panic_detail` module don't affect much outside the module itself, which is small and seems unlikely to need major changes in the future.

c2rust-analyze/src/context.rs
spernsteiner: This is a false positive in Clippy.  "This is only bad if it allocates or does some non-trivial amount of work," but "The lint also cannot figure out whether the function you call is actually expensive to call or not."  In this case the function being called is `const`, so it doesn't need to do any run-time work at all.  The lint in question has also been [downgraded to allow-by-default](https://github.com/rust-lang/rust-clippy/pull/9829) due to the high rate of false positives.

c2rust-analyze/src/context.rs
spernsteiner: Since we visit functions in callgraph order (callee first, then caller), in nearly all cases where there's both a propagated error and an error in the caller itself, we will set the propagated error first and then never analyze the caller (so we won't see the non-propagated error at all).  The only exception is for mutually recursive functions.

We'll soon be getting rid of this style of error propagation through the callgraph, and I think it's fine to leave this as-is for now.

c2rust-analyze/src/panic_detail.rs
spernsteiner: This change only simplifies this single line (`loc` still needs to be `Option` since `PanicDetail::new` doesn't provide a location), and it would make us reliant on an implementation detail of the standard library, so I don't think it's worth it.

spernsteiner:
Rebased and addressed all the comments so far

c2rust-analyze/src/util.rs
```diff
@@ -276,7 +278,7 @@ fn builtin_callee(tcx: TyCtxt, did: DefId) -> Option<Callee> {
             if tcx.impl_trait_ref(parent_did).is_some() {
                 return None;
             }
-            let parent_impl_ty = tcx.type_of(parent_did);
+            let parent_impl_ty = EarlyBinder(tcx.type_of(parent_did)).subst(tcx, substs);
```
aneksteind, line 281: could you comment on why this is necessary

spernsteiner: This was from #918, which I mistakenly merged into this branch instead of into `master`

aneksteind: ok, I'd still like to know why it was done. I should have commented on #918

kkysen: In #918, I assumed it was to get the actual type instead of the generic type.

c2rust-analyze/src/panic_detail.rs
kkysen: Most of these fields are optional, all except for `msg`.  They are all set when an actual panic occurs, but when it's a function failing due to one its callees failing, only `msg` is used.  I was thinking it could like this:

```rust
struct PanicDetail {
    msg: String,
    loc: String,
    backtrace: Backtrace,
    span: Span,
}

impl PanicDetail {
    pub fn from_info(info: &PanicInfo) -> Self {
        Self {
            msg: panic_to_string(info.payload()),
            loc: info.location().unwrap().to_string(),
            backtrace: Backtrace::new_unresolved(),
            span: CURRENT_SPAN.with(|cell| cell.get()),
        }
    }
}

enum FnFailure {
    Panic(PanicDetail),
    CalleeFailed(String),
}
```

Also, as I commented in https://github.com/immunant/c2rust/pull/876#discussion_r1155265976, doing `info.location().unwrap()` simplifies things here more than just that one line, as then uses of `PanicDetail::loc` don't need to handle the `Option`.

I'm not sure what you mean by the "called function failed" will be going away soon, though.

c2rust-analyze/src/context.rs
kkysen: This one is a slightly different warning.  When it suggests using a closure that calls a `const fn`, that is more of a false warning as you were saying, but in this case, `.unwrap_or_default()` is functionally different (I'm surprised it falls under the same `clippy` lint, as it's quite different).  Using `Default::default()` for this is more semantically clear (we don't care that it's actually `Vec::new()`, but that it's the default, empty container).  We could conceivably change it to `ThinVec`, `Box<[T]>`, `IndexSet`, etc., in which case the `.unwrap_or_default()` wouldn't need to be changed since it's the same semantic logic.

That said, I don't really want to spend time debating the merits of `clippy` lints.  I think we should just do what `clippy` says here.  There's no major reason not to, and using `Default` here is more semantically specific.

c2rust-analyze/tests/common/mod.rs
```diff
@@ -13,6 +13,7 @@ use clap::Parser;
 #[derive(Default)]
 pub struct Analyze {
     path: PathBuf,
+    dont_catch_panic: bool,
```
kkysen, line 16: Instead of adding this here, I'd add it as a directive arg like this (before `AnalyzeArgs::env`):

```rust
    /// Don't catch any panics during analysis of functions.
    ///
    /// The default behavior is to catch panics,
    /// try to recover as much as possible,
    /// and analyze as many functions as possible.
    /// Setting this will make panics fatal instead.
    #[arg(long)]
    dont_catch_panic: bool,
```

and then add `//! --dont-catch-panic` to all of the current tests, `analyze` and `filecheck` ones, as that was their previous behavior and we want to keep testing the same behavior.  When we add a new test or want to change an existing test to test behavior post-panicking, then we can remove the `//! --dont-catch-panic` from that file.

c2rust-analyze/src/main.rs
```diff
@@ -1078,7 +1078,13 @@ impl rustc_driver::Callbacks for AnalysisCallbacks {
 
 fn main() -> rustc_interface::interface::Result<()> {
     init_logger();
-    panic_detail::set_hook();
+
+    let dont_catch = env::var_os("C2RUST_TEST_ANALYZE_DONT_CATCH_PANIC").is_some();
```
kkysen, line 1082: nit: This is a flag for `c2rust-analyze`, so I'd leave out the `TEST` part (or at least, order it `C2RUST_ANALYZE_TEST`).

```suggestion
    let dont_catch = env::var_os("C2RUST_ANALYZE_DONT_CATCH_PANIC").is_some();
```

c2rust-analyze/tests/common/mod.rs
spernsteiner: I'm going to turn on "don't catch panic" mode by default in all tests and add a `//! --catch-panic` option to opt out instead.  Catching panics by default covered up a significant regression in #919, which I originally developed on top of this branch.

kkysen: That makes sense.  I was wondering, too, if the default should be panicking, too.  nit: I'd call it `--catch-panics`, though.

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
kkysen: It should work fine as is, but I thought it'd be a bit clearer if we minimize the part inside the `catch_unwind` to exactly the parts we want to catch panics from.  If something in this highlighted part panics, we probably want that to be fatal.

c2rust-analyze/src/panic_detail.rs
spernsteiner: >I'm not sure what you mean by the "called function failed" will be going away soon, though.

I'm currently working on better error recovery, so that we'll no longer need to propagate analysis failures up to callers.

c2rust-analyze/src/context.rs
spernsteiner: >I think we should just do what clippy says here. There's no major reason not to

There's no good reason to make this change.  The fact that Clippy suggested it isn't a good reason on its own, because Clippy often gives bad advice.  (Any clippy lints that do regularly give good advice tend to get uplifted into rustc.)

> The fact that it's Vec isn't important here.

I disagree.  Knowing the type of `xs` is useful for knowing what you'll get if you do `for x in xs`.

As far as I know, the official policy for this repo does not require the code to be clippy-clean.  Unless that changes (which I would oppose), I don't see any value in making the code (very slightly) worse just to appease the tool.

spernsteiner:
Rebased and addressed some final comments about panic-catching defaults in tests

kkysen (approved):
There were a few things I thought should be done a bit differently, but they're ultimately minor.  The main parts of this PR—catching and correctly handling panics, and making sure our tests still test the same thing—all look good to me now.

PR #879, opened 2023-03-30
User type annotation rewrite

aneksteind:
Rewrite user-supplied type annotations where appropriate.

### DIFF
--- a/c2rust-analyze/src/rewrite/ty.rs
+++ b/c2rust-analyze/src/rewrite/ty.rs
@@ -4,6 +4,8 @@
 //! with the materialization of adjustments in expr rewriting, we try to apply this transformation
 //! selectively, since we don't want to unfold all type aliases in the program.
 
+use std::collections::HashMap;
+
 use crate::context::{AnalysisCtxt, Assignment, LTy};
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
 use crate::rewrite::Rewrite;
@@ -258,6 +260,7 @@ struct HirTyVisitor<'a, 'tcx> {
     rw_lcx: LabeledTyCtxt<'tcx, RewriteLabel>,
     mir: &'a Body<'tcx>,
     hir_rewrites: Vec<(Span, Rewrite)>,
+    hir_span_to_mir_local: HashMap<Span, rustc_middle::mir::Local>,
 }
 
 impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
@@ -330,13 +333,13 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
         match s.kind {
             // A local with a user type annotation
             hir::StmtKind::Local(hir_local) if hir_local.ty.is_some() => {
-                for (mir_local, mir_local_decl) in self.mir.local_decls.iter_enumerated() {
-                    if mir_local_decl.source_info.span == hir_local.pat.span {
-                        let lty = self.acx.local_tys[mir_local];
-                        let rw_lty = relabel_rewrites(self.asn, self.rw_lcx, lty);
-                        let hir_ty = hir_local.ty.unwrap();
-                        self.handle_ty(rw_lty, hir_ty);
-                    }
+                if let Some(mir_local) = self.hir_span_to_mir_local.get(&hir_local.pat.span) {
+                    let mir_local_decl = &self.mir.local_decls[*mir_local];
+                    assert_eq!(mir_local_decl.source_info.span, hir_local.pat.span);
+                    let lty = self.acx.local_tys[*mir_local];
+                    let rw_lty = relabel_rewrites(self.asn, self.rw_lcx, lty);
+                    let hir_ty = hir_local.ty.unwrap();
+                    self.handle_ty(rw_lty, hir_ty);
                 }
             }
             _ => (),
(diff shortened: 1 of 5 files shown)


c2rust-analyze/tests/filecheck/type_annotation_rewrite.rs
```diff
@@ -0,0 +1,37 @@
...
+    let x: *mut i32;
+    x = std::ptr::addr_of_mut!(z);
+}
+
+// CHECK-DAG: pub unsafe fn let_decl_cell()
+pub unsafe fn let_decl_cell() {
+    let mut z = 2;
+    // CHECK-DAG: let x: &core::cell::Cell<(i32)>;
+    let x: *mut i32;
+    x = std::ptr::addr_of_mut!(z);
```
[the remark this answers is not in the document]
spernsteiner: I think there are three issues here:
1. `x = std::ptr::addr_of_mut!(z)` should be rewritten to `x = &z` or `x = &mut z`, depending on the type of `x`
2. `z`'s initializer should be rewritten to `let z = Cell::new(2);` so that `let x: &Cell<i32> = &z;` will work (`z` should have `CELL` in its `addr_of` flags, which is how we know to do this)
3. `z` should be `UNIQUE` and so shouldn't require `Cell` at all (the `*x = *x` operation below should not trigger a borrowck error)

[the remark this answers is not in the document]
aneksteind: Are these things that you'd support having a separate PR for or do you want them addressed here?

kkysen:
What do you mean by user-supplied type annotations?

aneksteind:
@kkysen rustc refers to type annotations as user_ty so I was going off of that and its description

c2rust-analyze/tests/filecheck/type_annotation_rewrite.rs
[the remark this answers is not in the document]
aneksteind: As per our discussion offline, only rewriting borrows in cases of `Use(Copy)`: f1989754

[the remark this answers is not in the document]
spernsteiner: Go ahead.  The `mir_op` part of the implementation should be straightforward.  The `hir_op` side might be a bit tricky, though, since there's a macro involved.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -325,19 +326,38 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
     fn nested_visit_map(&mut self) -> Self::Map {
         self.acx.tcx().hir()
     }
+
+    fn visit_stmt(&mut self, s: &'tcx hir::Stmt<'tcx>) {
+        match s.kind {
+            // A local with a user type annotation
+            StmtKind::Local(hir_local) if hir_local.ty.is_some() => {
```
spernsteiner, line 333: For consistency, this should use the qualified path `hir::StmtKind::Local` (and remove the `use hir::StmtKind` above).  Since this module is juggling a mix of HIR and `TyCtxt` representations, which often have overlapping names, I always use the fully qualified path `hir::Foo` or `ty::Foo` for clarity.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -325,19 +326,38 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
...
+                let rw_lcx = LabeledTyCtxt::new(self.acx.tcx());
```
spernsteiner, line 334: This code should use `self.rw_lcx` instead of creating a new one.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -325,19 +326,38 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
...
+                for (mir_local, mir_local_decl) in self.mir.local_decls.iter_enumerated() {
```
spernsteiner, line 335: Rather than make this an O(N^2) operation, we should build a map up front from `Span` to the MIR `Local` with that span.  This would also give us a chance to check for ambiguous cases where there are two locals with the same span (maybe due to macro tricks).

c2rust-analyze/src/rewrite/ty.rs
aneksteind: 3bc6f339

c2rust-analyze/src/rewrite/ty.rs
aneksteind: dc598e4a

c2rust-analyze/src/rewrite/ty.rs
aneksteind: does bd8f4a36 accomplish what you're asking for? I was sort of unsure what subset of Spans from HIR we wanted to build the map with and if there was some each way to iterate over all of those. in that commit I just limit it to the spans of `hir::Local`s

spernsteiner: That looks like it works, but seems a bit fragile since it relies on MIR locals being declared in a particular order.  What I was thinking was actually to just loop over the `mir.local_decls` and insert with `mir_local_decl.source_info.span` as the map key (not looking at HIR at all).  Then you can still do lookups exactly like you've got in that commit, since we expect the MIR local and the HIR local to have the same `Span`.

aneksteind: I've made the simplification in a6596623

Issue #878, opened 2023-03-29
c2rust-analyze: use rvalue_tys for all casts


spernsteiner:
On lighttpd, our largest source of panics is the `Rvalue::Cast` case in `AnalysisCtxt::type_of_rvalue`.  Only a few kinds of casts are handled here; otherwise, it asserts that there are no pointer types in the cast's target type.  Separately, casts of the constant `0` to a pointer type are supported because `main.rs` adds an entry to `acx.rvalue_tys` for each such cast.  Having an `rvalue_tys` entry causes the entire `type_of_rvalue` `Cast` case to be bypassed in favor of the `LTy` stored in that entry.

We should change the analysis to add `rvalue_tys` entries for all `Cast`s, and we should move the restrictions on cast kinds and input/output types, which are currently implemented partly in `main.rs` and partly in `type_of_value`, into `dataflow::type_check`.  This will make cast handling more uniform, and should make it easier to add support for additional cast kinds later on.  This should also let us provide more specific panic messages for unsupported casts, which will make it easier to classify the various kinds of unsupported casts in lighttpd.

kkysen:
Is this at all similar to your feedback here: https://github.com/immunant/c2rust/pull/839#discussion_r1115160479?

fw-immunant:
Should this be closed now that #883 has merged? cc @aneksteind

kkysen:
I believe so.

PR #883, opened 2023-04-07
Refactor.analysis.cast

aneksteind:
Lays the groundwork to check whether or not certain types of ptr-to-ptr casts are supported.

Implements the reorganization specified in https://github.com/immunant/c2rust/issues/878, and lays the groundwork for https://github.com/immunant/c2rust/pull/839 as outlined in https://github.com/immunant/c2rust/pull/839#discussion_r1115160479


kkysen (commented):
It's a bit hard to tell what changed vs. just moved in this PR.  For f2b89e354cd648930649434cac4337fb8eb3fc8a, there are 2 moves, but also a few changes.  Could those be split out?

c2rust-analyze/src/main.rs
```diff
@@ -361,6 +362,109 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
...
+                StatementKind::Assign(x) => *x.clone(),
+                _ => continue,
+            };
+
+            if acx.c_void_casts.should_skip_stmt(Location {
+                statement_index: i,
+                block: bb,
+            }) {
+                continue;
+            }
```
kkysen, line 377: Why is this part added?  Was this something missed earlier?

c2rust-analyze/src/main.rs
```diff
@@ -361,6 +362,109 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
     }
 }
 
+fn label_rvalue_tys<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
+    for (bb, bb_data) in mir.basic_blocks().iter_enumerated() {
+        for (i, stmt) in bb_data.statements.iter().enumerate() {
+            let (_, rv) = match &stmt.kind {
+                StatementKind::Assign(x) => *x.clone(),
```
kkysen, line 368: `&**x` avoids the `Box` `.clone()` I believe.

c2rust-analyze/src/main.rs
kkysen, line 1: Could this commit be split into 2, one for the pure move of the loop in `run` to `label_rvalue_tys`, and then another for the move of the code from `type_of_rvalue` to `label_rvalue_tys`?  If those 2 commits are their own PR, I think it becomes a lot easier to review the rest of the PR with its more substantive changes with smaller diffs.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -41,6 +41,11 @@ struct TypeChecker<'tcx, 'a> {
     equiv_constraints: Vec<(PointerId, PointerId)>,
 }
 
+fn is_castable_to(_t: LTy, _u: LTy) -> bool {
+    // TODO: implement
+    true
+}
```
kkysen: Is this meant to be something similar to what I implemented in #839 and #841?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -142,7 +147,13 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                 self.visit_place(pl, Mutability::Not);
             }
-            Rvalue::Cast(_, ref op, _) => self.visit_operand(op),
+            Rvalue::Cast(_cast_kind, ref op, _) => {
+                // A cast such as T as U
+                let t_lty = self.acx.type_of(op);
+                let u_lty = rvalue_lty;
+                assert!(is_castable_to(t_lty, u_lty));
+                self.visit_operand(op)
+            }
```
kkysen: I think something like `from` and `to`, or `pl` and `rv` are clearer than `t` and `u`.

kkysen (commented):
I'm confused by the change in 57ff503e1d63163d69d5f4b209f72c92f6f85b6a alone, as that code seems to be moved (not sure if changed at all) in 47a053f66ba86ff9c28b1943f03844735c4694c1.  Could those commits be combined so it's clear that it's a move?  Otherwise, I'm not sure how 57ff503e1d63163d69d5f4b209f72c92f6f85b6a does what the commit message says.

c2rust-analyze/src/main.rs
aneksteind: the thing this was checking before happened downstream of a check of this kind. now it's moved to upstream of that check so another check needed to be added

c2rust-analyze/src/main.rs
aneksteind: i'm not in favor of making those changes their own PR but will split up the commit into two at a minimum

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: yes, the idea was that this is where the bulk of your checks in those PRs would take place

c2rust-analyze/src/main.rs
aneksteind: On second look, I think it's only combining the two into one location (assuming you're talking about f2b89e35). I think we should keep them as one commit unless I'm missing some extra changes on top of that in the changelog

aneksteind:
@kkysen https://github.com/immunant/c2rust/commit/47a053f66ba86ff9c28b1943f03844735c4694c1 wasn't meant to signify a movement. It's just some additional checks that were made elsewhere previously. Yes, checks that were in one place are now present in another, but semantically those two commits represent 1. cutting out unnecessary checks given what's now returned (the labeled casted-to type) and 2. adding some cast-specific type checks to `type_check.rs`.

As far as https://github.com/immunant/c2rust/commit/57ff503e1d63163d69d5f4b209f72c92f6f85b6a itself is concerned, before we were taking portions of the casted-from type (T) and the casted-to type (U) and combining them into another type (X). This instead just labels and returns LTy(U).

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: https://github.com/immunant/c2rust/pull/883/commits/d9a6de02ce88087a4cc60a239122696907863d87

spernsteiner:
Any idea why this changes the behavior on the `as_ptr.rs` test case, which doesn't contain any casts?

Otherwise all tests are producing identical output, with the exception of `insertion_sort_driver.rs`, which I suspect is another instance of whatever change we're seeing in `as_ptr.rs`.

spernsteiner:
I think there is actually a bug in this `as_ptr.rs` example.  Specifically, it seems like we're missing a dataflow edge somewhere between `_4` and `_3`.

Both this branch and `master` give `_4` offset permissions:
```
_3 (37: x.as_ptr()): addr_of = UNIQUE, type = READ | UNIQUE | OFFSET_ADD | OFFSET_SUB#&[i32][(empty)#[i32][(empty)#i32[]]]
```

However, `master` gives the same permissions to `_3`, while this branch gives no permissions:

```diff
-_4 (37: x.as_ptr()): addr_of = UNIQUE, type = READ | UNIQUE | OFFSET_ADD | OFFSET_SUB#&[i32; 10][(empty)#[i32; 10][(empty)#i32[]]]
+_4 (36: x.as_ptr()): addr_of = UNIQUE, type = UNIQUE#&[i32; 10][(empty)#[i32; 10][(empty)#i32[]]]
```

This branch similarly gives no permissions to `_1`, but since `_1` only flows into `_4`, the lack of permissions on `_4` seems like the real problem.

aneksteind:
Ah yes, I see that the way `X` was being constructed before included using the `LTy` of `op` and its args to construct `LTy` of `X`. So now that connection's been severed. I'll look for a way to maintain those permissions, but I think we risk running into the same parity-checking in terms of structure shape as before.

spernsteiner:
I think it would be cleaner and more in line with the purpose of this branch to revert that last commit (so the `rvalue_tys` entry for a cast is always just the labeled target type) and instead implement the equivalent of the old rules by setting up appropriate dataflow edges inside `dataflow::type_check`.  For the `Unsize` in this example, with an operand type of `& /*p0*/ [T0; n]` and an rvalue type of `& /*p1*/ [T1]`, it should record an edge from `p0` to `p1` (that is, make `p1`'s permissions a subset of `p0`'s) and assert that `T0` is equivalent to `T1`.  This is similar to what we normally do for assignments (like an assignment from `& /*p2*/ T2` to `& /*p3*/ T3`), but here one side's pointee type is wrapped in `TyKind::Array` while the other's uses `TyKind::Slice` (and also we're doing this "assignment" between the operand type and the expected/`rvalue_tys` type of a single rvalue, rather than between the LHS and RHS of a `StatementKind::Assign`).

spernsteiner:
Forgot to mention - since this branch is meant to be purely a refactor, it's probably better to leave fixes like the `&[[..]]` thing to a separate branch.

aneksteind:
@spernsteiner reverted the last commit and setup dataflow rules in 28073265. I'm fine saving fixes for another PR but am still curious if you agree whether or not I have the right idea regarding the cause.

c2rust-analyze/src/main.rs
```diff
@@ -361,6 +361,51 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
...
+                        let tuple_ty = rv.ty(acx, acx.tcx());
+                        acx.assign_pointer_ids(tuple_ty)
+                    }
+                    _ => continue,
+                },
+                Rvalue::Cast(_, _, ty) => acx.assign_pointer_ids(*ty),
+                _ => continue,
+            };
+            let loc = Location {
+                block: bb,
```
aneksteind, line 401: dedup with above

aneksteind: 92c1aaa8

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -142,7 +148,59 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                assert!(is_castable_to(op_lty, rvalue_lty));
+                self.visit_operand(op)
+            }
+            Rvalue::Cast(_cast_kind, ref op, _) => {
+                // A cast such as `T as U`
+                let casted_from = self.acx.type_of(op);
+                let casted_to = rvalue_lty;
+                assert!(is_castable_to(casted_from, casted_to));
+                self.visit_operand(op)
+            }
```
kkysen, line 203: Ptr and non-ptr casts are quite different; I don't think there's really any `is_castable_to` function we could conceivably put here.  Non-ptr casts are going to be things like scalar casts without any pointers, so we don't need to really care about those, right, as there are no pointers?  Even if it's some sort of transmuting cast not going through a pointer, `rustc` typeck isn't going to allow that in the first place.  Ptr-to-ptr casts are the main thing we care about here, and that's how I designed the #839 and #841 PRs.

c2rust-analyze/src/main.rs
```diff
@@ -361,6 +361,50 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
     }
 }
 
+fn label_rvalue_tys<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
```
kkysen, line 364: This will conflict with #886.  Should I change #886 to run in a separate, second loop, or should we rename this function (given it'll label constant tys, too)?  The downside of two loops is then the match is separate into two places.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -142,7 +148,59 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                    _ => unreachable!("unsize cast has non-pointer output {:?}?", ty),
+                };
+
+                let op_lty = self.acx.type_of(op);
+                assert!(matches!(
+                    op_lty.kind(),
+                    TyKind::Ref(..) | TyKind::RawPtr(..)
+                ));
+                assert_eq!(op_lty.args.len(), 1);
+                let op_pointee_lty = op_lty.args[0];
```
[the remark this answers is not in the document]
aneksteind: 9573df7e

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: I don't think there is harm in keeping the check here and making it trivially true for non-ptr types. Do you?

c2rust-analyze/src/main.rs
aneksteind: I'm fine with it being renamed and updated to include code to label consts too. I don't think two loops is desirable, at least to me. For these reasons and what I mentioned [here](https://github.com/immunant/c2rust/pull/886#issuecomment-1509189413) I think this is why having one `rvalue_tys` dict in #886 is best. It's not a hill I'm willing to die on, and leave it to your judgement. We'll have to adjust to this one way or another and my primary concern is that it's correct and that it's clear to the reader what's going on.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -142,7 +148,59 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+
+                if let TyKind::Slice(elem_ty) = *pointee_ty.kind() {
```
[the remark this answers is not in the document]
aneksteind: no subset relation is made between casted-from and casted-to value. I'm not sure where it's possible to have an unsize cast and not have the RHS be a slice, but that would be when it'd happen

aneksteind:

c2rust-analyze/src/main.rs
kkysen: @spernsteiner just had a good suggestion for keeping the single `rvalue_tys` map and adding a `Vec` of locations that are constants as a way to iterate through them easily later, so I could try that.

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
kkysen: If we're not sure when it wouldn't be a slice, should we just make this is an `assert_matches!`?  If the other cases come up later, we'll handle them, but I don't want them silently passing through if we don't know how to handle them yet.

c2rust-analyze/src/main.rs
kkysen: Sure, I'll be working on it today.

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
aneksteind: i think there is an argument for that, but even though its a minor change i'd rather make it in a follow on PR. i want this one to be limited to the semantics described in the original issue and comments mentioned in the PR description

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: `CastKind::PointerExposeAddress` will fall into this case at the moment, so it's at least possible for something nontrivial to happen here.

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
spernsteiner: Please make it panic, and include the panic as part of this PR.  The usual rule applies here: anywhere we're not sure what the right answer is, the analysis should panic, not try to guess.

spernsteiner:
>I think the rewriting bug I thought this original solution had fixed (but didn't) is because of the propagation of the offset permission not taking into account that the original type is already an array type that can be offset (so it rewrites it with an extra []). Maybe the propagation upward should only go as far as the as_ptr call. WDYT?

Sounds plausible to me.  And I think your proposed fix sounds good - the `*mut T` output of `as_ptr` might be `OFFSET` even when the `*mut [T; N]` input is not.  So `as_ptr` should not propagate `OFFSET` from its output to its input.

c2rust-analyze/src/dataflow/type_check.rs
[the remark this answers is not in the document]
aneksteind: added in 8473be01

aneksteind:
> Sounds plausible to me. And I think your proposed fix sounds good - the *mut T output of as_ptr might be OFFSET even when the *mut [T; N] input is not. So as_ptr should not propagate OFFSET from its output to its input.

Thanks @spernsteiner, I will look into this and add the fix in another PR

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: Marking resolved. Feel free to mark unresolved with specific changes to request

c2rust-analyze/src/main.rs
kkysen: @aneksteind, it's done in #886's 8665bc2d8cb5c78fc840da1b3bdeabf96f68af9f now.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: I'll figure it out when I get to #839 and #841.

kkysen (approved):
LGTM now.  As I commented earlier, there are some parts where I'm not sure it'll be right once #839 and #841 land, but I'll just fix them in those PRs.

Also, could you just update the PR title to reflect the description?  Thanks!

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -142,7 +148,60 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+
+                assert_matches!(pointee_ty.kind(), TyKind::Slice(..));
+                if let TyKind::Slice(elem_ty) = *pointee_ty.kind() {
+                    assert!(matches!(op_pointee_lty.kind(), TyKind::Array(..)));
+                    let elem_lty = assert_matches!(op_pointee_lty.args, [elem_lty] => {
+                        elem_lty
+                    });
+                    assert_eq!(elem_lty.ty, elem_ty);
+                    assert_eq!(op_pointee_lty.label, PointerId::NONE);
+                    self.do_assign_pointer_ids(rvalue_lty.label, op_lty.label);
```
kkysen, line 182: Why is `do_assign_pointer_ids` only needed in the `Unsize` case?  (I'm going to ask some questions here to help with integrating the string cast PRs.)

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -142,7 +148,60 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                });
+
+                assert_matches!(pointee_ty.kind(), TyKind::Slice(..));
+                if let TyKind::Slice(elem_ty) = *pointee_ty.kind() {
+                    assert!(matches!(op_pointee_lty.kind(), TyKind::Array(..)));
+                    let elem_lty = assert_matches!(op_pointee_lty.args, [elem_lty] => {
+                        elem_lty
+                    });
+                    assert_eq!(elem_lty.ty, elem_ty);
+                    assert_eq!(op_pointee_lty.label, PointerId::NONE);
```
kkysen, line 181: What is this trying to assert exactly?  If `pointee_ty` is `Slice`, then how could `op_pointee_lty.label` ever be a non-`NONE` `PointerId`?  Or is this a kind of sanity check for that?

aneksteind: AFAIK only `Ref` and `RawPtr` `TyKind`s get pointer id's, and `op_pointee_lty` is already determined to be a `TyKind::Array` a few lines above

c2rust-analyze/src/dataflow/type_check.rs
aneksteind: It's needed at as because of the subset permission relationship described [here](https://github.com/immunant/c2rust/pull/883#issuecomment-1505578516):

> with an operand type of & /*p0*/ [T0; n] and an rvalue type of & /*p1*/ [T1], it should record an edge from p0 to p1

Why others don't have it is I think something that may be appropriate but I'm not sure. @spernsteiner ?

c2rust-analyze/src/dataflow/type_check.rs
kkysen: So this is just double-checking that `PointerId`s were labeled correctly?  And it shouldn't ever crash here?

aneksteind: It seems that way

kkysen: Okay, good to know for generalizing to other cases.

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: The `visit_statement` case for handling `pl = rv` works as follows:
```Rust
let pl_lty = acx.type_of(pl);
let rv_lty = acx.type_of_rvalue(rv);
do_assign(pl_lty, rv_lty);`
```
where the `do_assign` call establishes dataflow constraints to ensure that it's valid to assign a value of type `rv_lty` into a location of type `pl_lty`.  As usual, this means `rv_lty` must be a subtype of `pl_lty`.  Our subtyping rules are as follows:

* First, the unannotated types (`pl_lty.ty` and `rv_lty.ty`) must be identical.
* If the outermost type is a pointer or reference, then there are outermost `PointerId`s `pl_lty.label` and `rv_lty.label`.  We require that `perms(pl_lty.label) ⊆ perms(rv_lty.label)`.  That is, you can discard permissions on the outermost pointer type during an assignment, but can't add them.
* For all other `PointerId`s in `pl_lty`, there must be a corresponding `PointerId` at the same location within `rv_lty` (and vice versa), and the permissions on those corresponding `PointerId`s must be identical.

For simple cases, like when `rv` is a local variable, this logic for `StatementKind::Assign` works fine: `rv_lty` is exactly the type of the local, and we add some constraints between that and `pl_lty` (which is usually another local).  But in cases where there's an entry in `rvalue_tys`, there is no direct relationship between `rv_lty` and `rv`.  It's the job of `visit_rvalue` to establish such a connection in these cases.

For example, suppose we have an `AggregateKind::Array` assignment like `(ys: [V; n]) = [(x1: T1), (x2: T2)]: [U; n]`.  Here `[U; n]` is `rv_lty`, taken from `rvalue_tys`.  `visit_statement` will establish the subtyping constraint `[U; n] <: [V; n]`.  And `visit_rvalue` will make "pseudo-assignments", roughly as if the programmer had written `tmp[0] = x1; tmp[1] = x2; ys = tmp;` rather than `ys = [x1, x2];`, which establish the constraints `T1 <: U` and `T2 <: U`.

In the `CastKind::Unsize` case for `(y: &[V]) = (x: &[T; n]) as &[U]` we are doing a sort of "modified pseudo-assignment", as in `let tmp: &[U] = x: &[T; n]; y: &[V] = tmp: &[U];`.  The pseudo-assignment `tmp = x` would normally fail, because our subtyping rules don't allow for `&[T; n] <: &[U]` (the first condition is violated because the unlabeled types are not equal), but as a special case, here we allow the pointee type to change from `[_; n]` to `[_]` because that's the purpose of an `Unsize` cast.

We should have similar pseudo-assignments in all the `visit_rvalue` cases where we expect an `rvalue_tys` entry to be present.  I see the `CastKind::Pointer` case doesn't have one, which I think is a bug - on `y = (x: *mut u8) as *const u8`, we would fail to propagate permissions from `y` to `x` due to a missing edge between `rv_lty = *const u8` and `op_lty = *mut u8`.  I think the `Unsize` case should also be establishing constraints between any `PointerId`s within the pointee types; of the three subtyping rules given above, this case is enforcing the first (in modified form) and the second, but not the third.

kkysen: Thanks for the clear, thorough explanation!  I understand much better now.  As I integrate the string cast PRs, before or after (maybe after, as things should be more consolidated then), I'll make sure to properly establish the constraints in all rvalue cases.

kkysen: Tracked in #913.

PR #885, opened 2023-04-10
support tuples in static analysis

aneksteind:


spernsteiner (approved):
Looks like this is basically in line with the `AggregateKind::Adt` case, which seems like the right thing to do.

PR #886, opened 2023-04-10
(`c2rust-analyze`) Handle inline `const` refs, including string literals

### LINKED ISSUE #837: (`c2rust-analyze`) Handle string literals
Currently, `c2rust-analyze` can't handle any string literals, such as `""` or `b""`:

```shell
callgraph traversal order:
  DefId(0:3 ~ string_casts[c5c4]::f)
visit_statement(StorageLive(_1))
visit_statement(_1 = const "")
thread 'rustc' panicked at 'unexpected pointer type in &str', c2rust-analyze/src/context.rs:506:9
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: c2rust_analyze::context::label_no_pointers::{{closure}}
             at ./src/context.rs:503:9
   3: c2rust_analyze::labeled_ty::LabeledTyCtxt<L>::label
             at ./src/labeled_ty.rs:157:21
   4: c2rust_analyze::context::label_no_pointers
             at ./src/context.rs:502:5
   5: <rustc_middle::mir::syntax::Operand as c2rust_analyze::context::TypeOf>::type_of
             at ./src/context.rs:494:41
   6: <&T as c2rust_analyze::context::TypeOf>::type_of
             at ./src/context.rs:464:9
   7: c2rust_analyze::context::AnalysisCtxt::type_of
             at ./src/context.rs:241:9
   8: c2rust_analyze::context::AnalysisCtxt::type_of_rvalue
             at ./src/context.rs:304:36
   9: c2rust_analyze::dataflow::type_check::TypeChecker::visit_statement
             at ./src/dataflow/type_check.rs:233:30
  10: c2rust_analyze::dataflow::type_check::visit
             at ./src/dataflow/type_check.rs:388:13
  11: c2rust_analyze::dataflow::generate_constraints
             at ./src/dataflow/mod.rs:326:5
  12: c2rust_analyze::run
             at ./src/main.rs:480:45
```

Fixing this is a prerequisite to #833.

kkysen:
Fixes #837.

This only handles `const` refs that are inline and thus local to a function, as determined by being `ConstantKind::Val`.  Thus, this handles string literals like `""` (`inline_str`) and `b""` (`inline_bstr`), which is the main thing we're trying to support.

This PR also adds tests for other similar string usages, though they're not handled and turned off for now.  The outline cases, where the `const` ref/string literal is used through a named constant, aren't handled, as they are globally accessible and that makes them more complex.  Since supporting them isn't necessary for string literals, which is the main usage in `lighttpd-rust`.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -53,16 +53,16 @@ bitflags! {
 }
 
 impl PermissionSet {
-    pub fn for_const(constant: ConstantKind) -> PermissionSet {
+    pub fn for_const(constant: ConstantKind) -> Self {
         let ref_ty = constant.ty();
         let ty = match ref_ty.kind() {
             ty::Ref(_, ty, _) => ty,
             _ => panic!("expected only `Ref`s for constants: {ref_ty:?}"),
         };
         if ty.is_array() || ty.is_str() {
-            PermissionSet::READ | PermissionSet::OFFSET_ADD
+            Self::READ | Self::OFFSET_ADD
         } else if ty.is_primitive_ty() {
-            PermissionSet::READ
+            Self::READ
         } else {
             panic!("expected an array, str, or primitive type: {ty:?}");
         }
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -98,7 +98,7 @@ pub struct AnalysisCtxt<'a, 'tcx> {
     /// those `PointerId`s consistent, the `Rvalue`'s type must be stored rather than recomputed on
     /// the fly.
     pub rvalue_tys: HashMap<Location, LTy<'tcx>>,
-    pub const_tys: HashMap<ConstantKind<'tcx>, LTy<'tcx>>,
+    pub const_ref_tys: HashMap<ConstantKind<'tcx>, LTy<'tcx>>,
     next_ptr_id: NextLocalPointerId,
 }
 
@@ -106,7 +106,7 @@ pub struct AnalysisCtxtData<'tcx> {
     local_tys: IndexVec<Local, LTy<'tcx>>,
     addr_of_local: IndexVec<Local, PointerId>,
     rvalue_tys: HashMap<Location, LTy<'tcx>>,
-    const_tys: HashMap<ConstantKind<'tcx>, LTy<'tcx>>,
+    const_ref_tys: HashMap<ConstantKind<'tcx>, LTy<'tcx>>,
     next_ptr_id: NextLocalPointerId,
 }
 
@@ -199,7 +199,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             c_void_casts: CVoidCasts::new(mir, tcx),
             addr_of_local: IndexVec::new(),
             rvalue_tys: HashMap::new(),
-            const_tys: HashMap::new(),
+            const_ref_tys: HashMap::new(),
             next_ptr_id: NextLocalPointerId::new(),
         }
     }
@@ -213,7 +213,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             local_tys,
             addr_of_local,
             rvalue_tys,
-            const_tys,
+            const_ref_tys,
             next_ptr_id,
         } = data;
         AnalysisCtxt {
@@ -223,7 +223,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             c_void_casts: CVoidCasts::default(),
             addr_of_local,
             rvalue_tys,
-            const_tys,
+            const_ref_tys,
             next_ptr_id,
         }
     }
@@ -233,7 +233,7 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             local_tys: self.local_tys,
             addr_of_local: self.addr_of_local,
             rvalue_tys: self.rvalue_tys,
-            const_tys: self.const_tys,
+            const_ref_tys: self.const_ref_tys,
             next_ptr_id: self.next_ptr_id,
         }
     }
@@ -426,7 +426,7 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
             ref mut local_tys,
             ref mut addr_of_local,
             ref mut rvalue_tys,
-            ref mut const_tys,
+            ref mut const_ref_tys,
             ref mut next_ptr_id,
         } = *self;
 
@@ -444,7 +444,7 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
             *lty = remap_lty_pointers(lcx, &map, lty);
         }
 
-        for lty in const_tys.values_mut() {
+        for lty in const_ref_tys.values_mut() {
             *lty = remap_lty_pointers(lcx, &map, lty);
         }
 
@@ -515,7 +515,7 @@ impl<'tcx> TypeOf<'tcx> for Operand<'tcx> {
             Operand::Move(pl) | Operand::Copy(pl) => acx.type_of(pl),
             Operand::Constant(ref c) => {
                 let c = &**c;
-                match acx.const_tys.get(&c.literal) {
+                match acx.const_ref_tys.get(&c.literal) {
                     Some(lty) => lty,
                     None => label_no_pointers(acx, c.ty()),
                 }
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -53,7 +53,7 @@ bitflags! {
 }
 
 impl PermissionSet {
-    pub fn for_const(constant: ConstantKind) -> Self {
+    pub fn for_const_ref(constant: ConstantKind) -> Self {
         let ref_ty = constant.ty();
         let ty = match ref_ty.kind() {
             ty::Ref(_, ty, _) => ty,
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -4,7 +4,7 @@ use crate::pointer_id::{
     GlobalPointerTable, LocalPointerTable, NextGlobalPointerId, NextLocalPointerId, PointerTable,
     PointerTableMut,
 };
-use crate::util::{self, describe_rvalue, RvalueDesc};
+use crate::util::{self, describe_rvalue, PhantomLifetime, RvalueDesc};
 use crate::AssignPointerIds;
 use bitflags::bitflags;
 use rustc_hir::def_id::DefId;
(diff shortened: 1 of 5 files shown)


kkysen:
I'm not sure if/how this PR should add exact `READ | OFFSET_ADD` permissions for string literals (and `READ` permissions for scalar literal refs).

c2rust-analyze/src/context.rs
```diff
@@ -98,13 +98,15 @@ pub struct AnalysisCtxt<'a, 'tcx> {
...
     pub rvalue_tys: HashMap<Location, LTy<'tcx>>,
+    pub const_tys: HashMap<ConstantKind<'tcx>, LTy<'tcx>>,
     next_ptr_id: NextLocalPointerId,
 }
 
 pub struct AnalysisCtxtData<'tcx> {
     local_tys: IndexVec<Local, LTy<'tcx>>,
     addr_of_local: IndexVec<Local, PointerId>,
     rvalue_tys: HashMap<Location, LTy<'tcx>>,
+    const_tys: HashMap<ConstantKind<'tcx>, LTy<'tcx>>,
```
spernsteiner, line 109: It looks to me like we could use `rvalue_tys` for this instead of adding a new map.

spernsteiner:
>I'm not sure if/how this PR should add exact `READ | OFFSET_ADD` permissions for string literals

I think the best approach for this is to assign `PointerId`s to these literals (as you've implemented here), set up the permissions/flags for those `PointerId`s in the initial `Assignment` prior to running `dataflow` and `borrowck`, and assert at the end that the permissions/flags weren't modified.  This will catch cases where the program tries to mutate a string literal (or other constant) and will stop us from trying to generate a bogus rewrite, like trying to turn `"foo"` into `&mut str`.

kkysen:
> assert at the end that the permissions/flags weren't modified

If I do it this way, would it make more sense to keep `const_tys` separate from `rvalue_tys`?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -484,5 +504,11 @@ pub fn visit<'tcx>(
         );
     }
 
+    for (&constant, const_lty) in &acx.const_tys {
+        let _ptr_id = const_lty.label;
+        let _expected_perms = const_perms(constant);
+        // TODO: check that perms match the expected ones
+    }
```
kkysen, line 511: How can I lookup the `PermissionSet` of a `PointerId` after dataflow typeck has run?

spernsteiner: That mapping is in the `Assignment`

c2rust-analyze/src/main.rs
```diff
@@ -454,6 +454,7 @@ fn run(tcx: TyCtxt) {
             assert_eq!(local, l);
         }
 
+        let mut const_ref_tys = HashMap::new();
```
aneksteind, line 457: why not insert into the `acx` map directly?

aneksteind:
> If I do it this way, would it make more sense to keep const_tys separate from rvalue_tys?

@kkysen why not go about it this way, i.e. why is `const_tys` used instead?

kkysen:
I was going to change it back to using `rvalue_tys`, but if I need to iterate through the constant ones at the end to check the permissions weren't modified, then it's a lot simpler if there's a separate map.

aneksteind:
I think it would be cleaner to keep the `rvalue_tys` dict alone and instead have logic that checks permissions and matches against the `RValue` ty and checks there. If we want to check similar things in the future we would end up adding another map to the context to maintain the pattern and I don't think that'd be the best design

kkysen:
There's not really a way to tell an `LTy` came from a constant, though, right?  I'm not sure how to retrieve that information back.
.

c2rust-analyze/src/main.rs
kkysen: There'd be multiple mutable references to `acx` then, due to `acx.rvalue_tys`..  It'd probably be better to construct the `AnalysisCtxt` with the maps already constructed, but that'd be a bit more of a change, so I just did it this way for now.

kkysen:
@spernsteiner, the post-typeck check is erroring, saying the permissions have expanded from `READ | OFFSET_ADD` to `READ | UNIQUE | OFFSET_ADD`.  Is this correct?

kkysen:
@spernsteiner, also, do you think any `FileCheck` tests are needed for this, or is the post-typeck check that the permissions are as expected more than enough?

c2rust-analyze/src/main.rs
kkysen: Oh, well it might not have been that specifically then, but there definitely was a borrow checker error.  I tried this way first but it didn't work.

c2rust-analyze/src/context.rs
kkysen: Discussion moved here: https://github.com/immunant/c2rust/pull/886#issuecomment-1517568518.

c2rust-analyze/src/util.rs
```diff
@@ -348,3 +348,6 @@ pub fn is_null_const(constant: Constant) -> bool {
         _ => false,
     }
 }
+
+pub trait PhantomLifetime<'a> {}
```
aneksteind, line 352: why is this necessary?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -465,6 +465,10 @@ pub fn visit<'tcx>(
         equiv_constraints: Vec::new(),
     };
 
+    for (ptr, perms) in acx.const_ref_perms() {
+        tc.constraints.add_all_perms(ptr, perms);
```
aneksteind, line 469: why not add the constraints as you see consts in `visit_rvalue`?

c2rust-analyze/src/main.rs
```diff
@@ -489,6 +489,18 @@ fn run(tcx: TyCtxt) {
...
+                    Rvalue::Use(Operand::Constant(c)) => {
+                        // Constants can include, for example, `""` and `b""` string literals.
+                        if let ConstantKind::Val(_, ty) = c.literal && ty.is_ref() {
+                            // The [`Constant`] is an inline value and thus local to this function,
+                            // as opposed to a global, named `const`s, for example.
+                            // This might miss local, named `const`s,
+                            acx.assign_pointer_ids(ty)
+                        } else {
+                            // TODO: Handle global, named `const`s.
+                            continue;
```
aneksteind, line 501: should we emit a warning here?

aneksteind (commented):
I'm not sure why `PhantomLifetime` is necessary. Without knowing for sure it seems like an unnecessary complication. I also think the vector of locations for const refs is similarly unnecessary in that the permissions can be checked in another visit of the body after the permissions have been propagated (maybe in the `borrowck` phase or somewhere in between) in combination with adding constraints during the `dataflow` visitation.

aneksteind (approved):
Despite my comments, however, I think this looks good enough. Insofar as the points made above are valid, consider reducing the complexity slightly.

c2rust-analyze/src/main.rs
kkysen: Yeah maybe.  I think this `continue` eventually turns into the same `label_no_pointers` error, which isn't very descriptive.  We should only warn on references, though, as non-references will be fine in the `label_no_pointers` later on.

kkysen:
> I'm not sure why `PhantomLifetime` is necessary. Without knowing for sure it seems like an unnecessary complication.

It was needed due to how lifetimes and `impl Trait` work where I can't do multiple `impl Trait + 'a + 'b`s and have it work, but I can do `impl Trait + PhantomLifetime<'a> + 'b` and that works.  It's similar to `PhantomData` and type parameters, but with lifetimes.

kkysen:
> I also think the vector of locations for const refs is similarly unnecessary in that the permissions can be checked in another visit of the body after the permissions have been propagated (maybe in the `borrowck` phase or somewhere in between) in combination with adding constraints during the `dataflow` visitation.

Do you mean re-iterating through `BasicBlock`s, `Statement`s, `Rvalue`s, etc. to find the `Constant`s again?  Why is that better than doing it once and storing the result?

c2rust-analyze/src/context.rs
```diff
@@ -98,13 +114,44 @@ pub struct AnalysisCtxt<'a, 'tcx> {
...
 }
 
+impl<'a, 'tcx> AnalysisCtxt<'_, 'tcx> {
+    pub fn const_ref_tys(&'a self) -> impl Iterator<Item = LTy<'tcx>> + 'a {
+        self.const_ref_locs.iter().map(|loc| self.rvalue_tys[loc])
+    }
+
+    pub fn const_ref_perms(
+        &'a self,
+    ) -> impl Iterator<Item = (PointerId, PermissionSet)> + PhantomLifetime<'tcx> + 'a {
```
spernsteiner, line 131: Is the `PhantomLifetime<'tcx>` actually necessary here?  I don't think the returned iterator borrows anything with lifetime `'tcx`.  `self.const_ref_tys()` returns an iterator that doesn't borrow `'tcx`, and the closure provided to `map` doesn't capture anything.

c2rust-analyze/src/util.rs
spernsteiner, line 352: Should this be `trait PhantomLifetime<'a>: 'a {}` or something like that?  As-is, it seem like it doesn't constrain anything.

c2rust-analyze/src/context.rs
kkysen: It is.  `self.const_ref_tys()` returns `LTy<'tcx>`s, and so I need that `'tcx` in the return type somewhere.  Since there's no more `LTy`, I need to stick it somewhere.

c2rust-analyze/src/util.rs
kkysen: I don't think so?  Someone suggested this online when I was searching up the lifetime error.  Adding a `: 'a` makes it not compile.  I think the point of the `PhantomLifetime` is just to mark that `'a` is used.  Just like there's

```rust
pub struct PhantomData<T: ?Sized>;
```

kkysen:
> an invariant of location being present in the index operations

What do you mean by this?  That `const_ref_locs`'s `Location`s are present in `rvalue_tys`?  That seems like a pretty simple invariant.  Or I might be misunderstanding.

c2rust-analyze/src/util.rs
kkysen: See https://github.com/immunant/c2rust/pull/886#discussion_r1174153441.  It'd due to Rust not letting you add multiple lifetimes (at least correctly?) to an `impl Trait`.  I tried many other ways, but nothing worked, and I saw someone suggest this and it worked perfectly.

c2rust-analyze/src/main.rs
kkysen: I made it into an error here: 6a0608331dfc43a9a429f7f6f73847621fb89ee6.  It'll turn into an error eventually in `label_no_pointers` anyways, and erroring here gives a much clearer error.

PR #889, opened 2023-04-14
Rewrite calls to `std::ptr::addr_of` and `std::ptr::addr_of_mut`

aneksteind:
- [x] *mut
- [x] *const

Pointers with `CELL` permissions will be supported in another PR

### DIFF
--- a/c2rust-analyze/src/rewrite/expr/hir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/hir_op.rs
@@ -160,9 +160,16 @@ impl<'a, 'tcx> HirRewriteVisitor<'a, 'tcx> {
                 opt_assign_loc
             }
             hir::ExprKind::AddrOf(..) => {
+                if ex.span == ex.span.source_callsite() {
+                    // FIXME: this is a hacky proxy for checking if &raw comes from
+                    // std::ptr::addr_of(_mut)! or is compiler-generated. If the
+                    // spans are equal, it was most likely compiler-generated and so we
+                    // skip this particular type of expression for rewriting
+                    return None
+                }
                 let assign_loc = self
                     .find_sole_location_matching(
-                        ex.span.source_callsite(),
+                        ex.span,
                         |stmt| matches!(stmt.kind, mir::StatementKind::Assign(..)),
                         |_term| false,
                     )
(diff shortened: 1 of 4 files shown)


c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -159,6 +159,23 @@ impl<'a, 'tcx> HirRewriteVisitor<'a, 'tcx> {
                     .unwrap_or_else(|err| panic_sole_location_error(err, "Assign statement"));
                 opt_assign_loc
             }
+            hir::ExprKind::AddrOf(..) => {
+                if ex.span == ex.span.source_callsite() {
+                    // FIXME: this is a hacky proxy for checking if &raw comes from
```
aneksteind, line 164: this is addressed in https://github.com/immunant/c2rust/pull/893/files#diff-f27494e760407f9ad0a4266c21cf7c2e553c775034319d545a0fba0e58989965L163

fw-immunant (approved):
LGTM, makes sense.

I don't like how the rewrites become `&(x)` when we do try to keep track of precedence, but I believe that's caused by `Rewrite::Sub` always being parenthesized (as is also the case for `Rewrite::Identity`) rather than by anything in this PR.

aneksteind:
@spernsteiner would you mind taking a look at this?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -221,6 +238,12 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirRewriteVisitor<'a, 'tcx> {
     fn visit_expr(&mut self, ex: &'tcx hir::Expr<'tcx>) {
         let mut hir_rw = Rewrite::Identity;
 
+        // This span will be used to apply the actual rewrite.
+        // To prevent attempts to rewrite macros in their definition
+        // location (e.g. `std::ptr::addr_of`) we instead specify that
+        // the rewrite should occur at the callsite
+        let callsite_span = ex.span.source_callsite();
```
spernsteiner, line 245: Always looking through one level of macro expansion is probably not the right answer in all situations, but should work for the code we get out of `c2rust-transpile` at least.

spernsteiner (approved):

PR #893, opened 2023-04-18
Rewrite Cell

aneksteind:
This adds partial support for pointer with `CELL` permissions. Namely it rewrites:

- [x] assignments where the address of the local on the LHS of the assignment has `CELL` permissions with `Cell::new`
- [x] reads through pointers with `CELL` permissions with `Cell::get`
- [x] assignments where the local on the LHS is a pointer with `CELL` permissions whose address gets written to with `Cell::set`

Still to be supported in another PR:

- rewriting `struct` definitions, namely fields, with updated `Cell` types


c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -62,6 +62,12 @@ pub enum Rewrite<S = Span> {
     Cast(Box<Rewrite>, String),
     /// The integer literal `0`.
     LitZero,
+    /// Cell::new
+    CellNew,
+    /// Cell::get
+    CellGet(Box<Rewrite>),
+    /// cell.set(x)
+    CellSet(Box<Rewrite>, Box<Rewrite>),
```
spernsteiner, line 70: It seems likely that we'll add more function/method-call rewrites in the future, so it might be worth generalizing these with something like `FunctionCall(String, Vec<Rewrite>)` for calling arbitrary functions (and similarly for methods).  This would be similar to `TyCtor` which can build arbitrary generic types.

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -127,60 +155,51 @@ impl<'a, 'tcx> HirRewriteVisitor<'a, 'tcx> {
...
+                );
             }
-            hir::ExprKind::AddrOf(..) => {
-                if ex.span == ex.span.source_callsite() {
-                    // FIXME: this is a hacky proxy for checking if &raw comes from
-                    // std::ptr::addr_of(_mut)! or is compiler-generated. If the
-                    // spans are equal, it was most likely compiler-generated and so we
-                    // skip this particular type of expression for rewriting
-                    return None;
-                }
```
fw-immunant, line 169: How does `find_all_locations_matching` with a `mir::StatementKind::Assign` pattern predicate make the distinction between compiler-generated `&raw` and ones corresponding to a source-level `addr_of[_mut]!` call? This seems subtle enough to merit a comment.

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -127,60 +155,51 @@ impl<'a, 'tcx> HirRewriteVisitor<'a, 'tcx> {
...
-                    .unwrap_or_else(|err| panic_sole_location_error(err, "Assign statement"));
-                Some(assign_loc)
-            }
-            _ => {
-                eprintln!("warning: find_primary_location: unsupported expr {:?}", ex);
-                None
+                    .unwrap_or_else(|err| panic_location_error(err, "Assignment statement"));
+                locations.push(assign_loc)
             }
+            _ => {}
```
fw-immunant, line 199: Should we still warn in this case?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -326,29 +361,47 @@ fn mutbl_from_bool(m: bool) -> hir::Mutability {
...
 
+fn materialize_adjustments<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    adjustments: &[Adjustment<'tcx>],
+    hir_rw: Rewrite,
+) -> Rewrite {
+    let adj_kinds: Vec<&_> = adjustments.iter().map(|a| &a.kind).collect();
+    match (hir_rw, &adj_kinds[..]) {
+        (Rewrite::Identity, []) => Rewrite::Identity,
+        (hir_rw @ Rewrite::Identity, _) => {
```
fw-immunant, line 392: This rebinding is really confusing--you already have a name for `hir_rw`, so the @-pattern isn't necessary here. Right?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -128,8 +135,52 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
-                let pl_ty = self.acx.type_of(pl);
-                self.enter_assign_rvalue(|v| v.visit_rvalue(rv, pl_ty));
+                let pl_lty = self.acx.type_of(pl);
+
+                if pl.is_indirect() && self.acx.local_tys[pl.local].ty.is_any_ptr() {
+                    let local_addr = self.acx.local_tys[pl.local].label;
+                    let perms = self.perms[local_addr];
+                    let flags = self.flags[local_addr];
+                    if let (Ownership::Cell, _) = type_desc::perms_to_desc(perms, flags) {
+                        // this is an assignment like `let x = 2` but `x` has CELL permissions
```
fw-immunant, line 145: `let` seems inappropriate here if this assignment corresponds to `CellSet` and not `CellNew`.

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -244,47 +263,71 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirRewriteVisitor<'a, 'tcx> {
...
+        let locs = self.find_locations(ex);
+        for loc in &locs {
+            self.locations_visited.insert(*loc);
+        }
+
+        let all_rws: Vec<&_> = locs
+            .iter()
+            .filter_map(|loc| self.rewrites.get(loc))
+            .flatten()
+            .collect();
```
spernsteiner, line 275: I'm not sure if I like this strategy.  My original thinking when writing `find_sole_location` was that for each HIR expr, we should find all the MIR statements it produced and label each one with its purpose/origin within the expr, since `mir_op` rewrites on different parts may need to be applied in different ways.  For `addr_of!` this will work, but I worry that lumping together all the rewrites will cause problems for more complex situations where a single expr might produce 3-4 different statements corresponding to different parts of the expression.

spernsteiner:
It seems like there's a mix of `Cell` and `addr_of!` handling in this PR - is that intentional?

fw-immunant (changes requested):
Comments inline. I'd also prefer the `all_locations` infrastructure to be split out into a separate commit from the domain logic changes around `Cell` ownership, if possible.

c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -62,6 +62,12 @@ pub enum Rewrite<S = Span> {
     Cast(Box<Rewrite>, String),
     /// The integer literal `0`.
     LitZero,
+    /// Cell::new
+    CellNew,
```
spernsteiner, line 66: `CellNew` should take a `Box<Rewrite>` argument for consistency, even if it's always just `Identity` at the moment.  Also, there are some checks we plan to implement (TODO comments in `apply.rs`) that involve looking at which parts of the original expr (either `Identity` or `Sub`-exprs) appear in its `Rewrite`.

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: It currently does not make a distinction. Before a distinction was necessary because `find_sole_location` would be trying to get only one location where an HIR statement could generate multiple (this line skipped the second and assumed it was a compiler-generated thing based on the span, but it wasn't very robust and wasn't in an ideal place in terms of logic).

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: I've simplified it in 1608de9d. The shadowing is still necessary without the bind because `hir_rw` is moved in the match.

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: Agreed, fixed in c7a8216b

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: is the `flatten` the crux of the issue then, with the unflattened version being the crudest possibility of separating groups of MIR statements resulting from HIR expressions?

aneksteind:
> It seems like there's a mix of Cell and addr_of! handling in this PR - is that intentional?

@spernsteiner this branch started from the test cases in `addr_of.rs` emitting `Cell` types but not rewriting them, so that's why it's included here. I've moved the cell-specific ones to `cell.rs` in a391ef66

c2rust-analyze/src/rewrite/mod.rs
aneksteind: addressed in d0d400d

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: Would it be reasonable to group rewrites by the span of the HIR expression? Something analogous to a rewrite tree per HIR expression? I'm not sure how else to support multiple MIR statements per HIR Assign expression, for example. What do you recommend?

spernsteiner: There's no general strategy for this.  Essentially what we're doing here is reverse-engineering parts of rustc's HIR-to-MIR lowering so that we can lift rewrites from the MIR to the HIR.  I think the safest way to have separate cases for all the specific patterns we know how to handle.  For `addr_of!` specifically, the pattern is something like: a HIR `ExprKind::AddrOf` whose span came from a macro expansion, and exactly two adjacent MIR statements with the same span where the first is an `Rvalue::Ref` assignment and the second is an `Rvalue::AddressOf` assignment.  And in this case, a `mir::RewriteKind::RawToRef` on the second MIR statement is lifted into a corresponding HIR-level `Rewrite::Ref`, while other kinds of rewrites on the second statement and any rewrites on the first statement are ignored / cause a panic / get some default handling (I'm not sure of the exact details here).

It's possible to make this stricter (e.g. we expect that the LHS of the first assignment is a temporary, which appears only in the RHS of the second assignment), which reduces the possibility of bad rewrites but also means more work for us in writing and adjusting ultra-specific patterns.  It's also possible to make it more lenient (e.g. the current approach where all rewrites are pooled together), but it seems likely to me that this will lead to bugs in the future around things like method calls, where the call itself and the deref adjustment on the receiver have the same span but may have two different sets of MIR rewrites that need to be handled in different ways.

c2rust-analyze/src/rewrite/mod.rs
aneksteind: Does 957bbfee cover the extent of what you were thinking?

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -244,47 +263,112 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirRewriteVisitor<'a, 'tcx> {
...
+
+        let all_rws_unflattened: Vec<_> = locs
+            .iter()
+            .filter_map(|loc| self.rewrites.get(loc))
+            .map(|rws| rws.iter().map(|rw| &rw.kind).collect::<Vec<_>>())
+            .collect();
+        let all_rws_unflattened = all_rws_unflattened
+            .iter()
+            .map(|v| v.as_slice())
+            .collect::<Vec<_>>();
```
spernsteiner, line 279: This might be clearer without the filter, so that a list of N locations turns into N lists of rewrites (one from each location), where some of those lists might be empty.

c2rust-analyze/src/rewrite/expr/hir_op.rs
```diff
@@ -244,47 +263,112 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirRewriteVisitor<'a, 'tcx> {
...
+        use mir_op::RewriteKind::*;
+        if !all_rws_unflattened.is_empty() {
+            hir_rw = match &all_rws_unflattened[..] {
+                [[rw @ RawToRef { .. }]] if callsite_span != ex.span && is_addr_of_expansion() => {
+                    rewrite_from_mir_rws(rw, hir_rw)
+                }
+                [rws @ [OffsetSlice { .. }, SliceFirst { .. }]] => rws
+                    .iter()
+                    .fold(hir_rw, |acc, rw| rewrite_from_mir_rws(rw, acc)),
+                [[rw]] => rewrite_from_mir_rws(rw, hir_rw),
```
spernsteiner, line 366: With the filter removal from the previous comment, I think these two cases could be combined into a single case that applies when there is only one location.

c2rust-analyze/src/rewrite/expr/hir_op.rs
spernsteiner: Yes, that looks pretty good.  We can rearrange the individual cases later if needed.

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: Without the filter there are cases such as `[[], [RawToRef { mutbl: false }]]` to match against. I'm not sure we want to match against specific cases here, and we can instead do

```rust
             rws => rws
                    .iter()
                    .flat_map(|rws| rws.iter())
                    .fold(hir_rw, |acc, rw| rewrite_from_mir_rws(rw, acc))
```
but this matches against all cases and presumably we want to panic in some cases, no?

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: I've combined them in 849d4d16 without the filter

aneksteind: I'm not sure I like the https://github.com/immunant/c2rust/commit/849d4d163316880197fadf100160613985a5e5ca approach, though. I think the original constrains the space well-enough that we can catch any obviously-weird cases in the panic

aneksteind: I've added back the constraints and kept it one match arm in fbe4e2d0

aneksteind:
@fw-immunant would you mind reviewing this again, sans rebase? I'll give the rebase a good-faith effort before merging regardless

c2rust-analyze/src/rewrite/expr/hir_op.rs
spernsteiner: >Without the filter there are cases such as `[[], [RawToRef { mutbl: false }]]` to match against.

This is just the sort of thing I was thinking of in https://github.com/immunant/c2rust/pull/893#discussion_r1177062926 above: we expect to see two `Location`s, one with a `Ref` and the other with an `AddressOf` (checked by `is_addr_of_expansion`), where the first has no rewrites and the second has only `RawToRef` (checked by this pattern).

c2rust-analyze/src/rewrite/expr/hir_op.rs
spernsteiner: It works either way, but the `[rws] => ...` version is what I was thinking.  I think it's more important to filter based on the number of locations than on the number of rewrites.  If there's only one location but it has many rewrites, the most obvious thing to do is to apply them all in order (and `mir_op` is careful to generate rewrites in a sensible order to allow this).  It's only when there are multiple locations that it becomes unclear what exactly you should do with each set of rewrites.

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: Good point, I missed that. I've made the change in 8114679a, but there is one other type of instance where the compiler will generate these two ref/raw MIR statements:

```rust
let r: *mut request_st = &mut (*con).request;
```

which doesn't occur in a macro call. I've accounted for this case in the match arm

c2rust-analyze/src/rewrite/expr/hir_op.rs
aneksteind: Agreed, addressed in 8114679a

aneksteind:
@fw-immunant just rebased splitting things up slightly further with https://github.com/immunant/c2rust/pull/893/commits/c2c37acd95dc2054d8e6cb7e5102defd06b420e5 and https://github.com/immunant/c2rust/pull/893/commits/5834096e104a111724adabe9630ee6e606c49604

spernsteiner (approved):

fw-immunant (approved):
LGTM, sorry to take so long to review--kind of swamped at the moment. Thanks for the history changes.

Issue #709, opened 2022-10-18
Make `c2rust-analyze` be a `cargo` and `rustc` wrapper like `c2rust-instrument` is now


kkysen:
It appears this is now necessary to analyze the full, amalgamated (#871) (or not, just full) `lighttpd`.  Full `lighttpd` has bitfields, and thus depends on `c2rust-bitfields`.  Having a crate dependency that's not the bundled `libc` (which we can get away with an `extern crate libc` and no `cargo`) complicates things a lot, as the `rustc` invocation we pass to `c2rust-analyze` gets way more complex, and depends on previous `rustc` invocations for dependencies.  Thus, I think the best way forward to is to wrap `cargo` and let it does what it's meant to do instead of trying to re-implement any of that.  We can base this on how `c2rust-instrument` is currently wrapping `rustc`, which itself was based on `miri`, `clippy`, and other `rustc`-wrappers.

PR #894, opened 2023-04-20
[c2rust-analyze] Run analysis on pdg input

fw-immunant:
This is a rebased version of @spernsteiner's [analyze-pdg-input](https://github.com/immunant/c2rust/tree/analyze-pdg-input) branch.


fw-immunant:
Crashes in CI because the `pdg.bc` file isn't found on `cargo check`.

c2rust-analyze/src/main.rs
```diff
@@ -544,6 +546,110 @@ fn run(tcx: TyCtxt) {
...
 
+    // Process PDG
+    let mut func_def_path_hash_to_ldid = HashMap::new();
+    for &ldid in &all_fn_ldids {
+        let def_path_hash: (u64, u64) = tcx.def_path_hash(ldid.to_def_id()).0.as_value();
+        eprintln!("def_path_hash {:?} = {:?}", def_path_hash, ldid);
+        func_def_path_hash_to_ldid.insert(def_path_hash, ldid);
+    }
+
+    if let Ok(f) = std::fs::File::open("pdg.bc") {
```
kkysen, line 557: This probably shouldn't be hardcoded at all, right?  Ideally a CLI arg, though we don't have anything set up for that at the moment and it's a bit complicated with wrapping `rustc` right now (#709 would fix that), but we probably want at least an environment variable for this.

c2rust-analyze/src/main.rs
```diff
@@ -544,6 +546,110 @@ fn run(tcx: TyCtxt) {
...
+                }
+                if node_info.flows_to.store.is_some() {
+                    perms.insert(PermissionSet::WRITE);
+                }
+                if node_info.flows_to.pos_offset.is_some() {
+                    perms.insert(PermissionSet::OFFSET_ADD | PermissionSet::OFFSET_SUB);
+                }
+                if node_info.flows_to.neg_offset.is_some() {
+                    perms.insert(PermissionSet::OFFSET_ADD | PermissionSet::OFFSET_SUB);
+                }
```
kkysen, line 631: Shouldn't `pos_offset` correspond to `OFFSET_ADD` and `neg_offset` to `OFFSET_SUB`?

pdg/src/main.rs
```diff
@@ -191,6 +182,10 @@ fn main() -> eyre::Result<()> {
     pdg.graphs.assert_all_tests();
     let repr = pdg.repr(&args.print);
     println!("{repr}");
+
+    let f = std::fs::File::create("pdg.bc")?;
```
kkysen, line 186: This can be added as a CLI arg in `Args` pretty easily, something like `--output` maybe.

pdg/src/util/serde.rs
kkysen, line 1: I don't think we really need this.  The only types we access are `Node`'s `.function: Func`, `.dest: Option<MirPlace>`, and `.info: Option<NodeInfo>` fields, which are all either in `c2rust_pdg` or `c2rust_analysis_rt`, not `rustc` private crates, meaning we can freely `#[derive(Serialize, Deserialize)]` on them.

aneksteind (approved):
I think the LGTM after confirming it's tested. And I think @kkysen has a point about `pdg.bc` being a CLI arg. We have a `pdg` script that generates these files that it would be good to align with.

c2rust-analyze/src/main.rs
fw-immunant: CC @spernsteiner?

pdg/src/main.rs
fw-immunant: Done in a78e49693fc02f3e70e9972f1933001683f2bce0.

c2rust-analyze/src/main.rs
fw-immunant: Filename is now (cb8307e) obtained from the `PDG_FILE` env var, whose presence decides whether we expect a serialized pdg.

pdg/src/util/serde.rs
fw-immunant: This seems like a good clean-up for when we have time, but not worth blocking this as it's part of integrating the pipeline from instrumentation to rewriting. I'll file an issue.

kkysen:
Actually, now that I think about it, why is serialization necessary at all?  If `c2rust-analyze` is using `c2rust-pdg` as a library in the first place, it can just generate the PDG in the same way `pdg/src/main.rs` does.  That seems way simpler.

aneksteind:
Would it not be useful to be able to save a PDG offline for later processing?

kkysen:
> Would it not be useful to be able to save a PDG offline for later processing?

Is it?  We can just run `c2rust-pdg` if we need that.  Having to serialize the PDG complicates things a lot for a use case we don't need yet at all.  If later we want to do the PDG generation as a separate step as some sort of optimization, we can add it back, but I don't see a reason why need that extra functionality at the moment.

c2rust-analyze/src/main.rs
spernsteiner: That sounds right - not sure why I did it this way.

fw-immunant:
I'm generally very in favor of avoiding plumbing data through serialization and the filesystem if we can. AIUI, this branch was revived so that we start using the results of dynamic analysis as the starting point for static analysis; that doesn't require c2rust-pdg to be a separate program, so I'd rather use it as a library and save on complexity.

That said, that approach means `c2rust-analyze` now needs to take the event log and metadata as inputs. Are we back to specifying file paths in environment variables for those, pending #709?

spernsteiner:
>that doesn't require c2rust-pdg to be a separate program, so I'd rather use it as a library and save on complexity

Does it really save that much complexity?  Serializing and deserializing with serde should be pretty straightforward.  Also, if processing the instrumentation trace to build the PDG takes a nontrivial amount of time, it would be nice during development to be able to do that step only once and test various analysis changes with the same PDG.

kkysen:
> Also, if processing the instrumentation trace to build the PDG takes a nontrivial amount of time, it would be nice during development to be able to do that step only once and test various analysis changes with the same PDG.

`c2rust-pdg` generally runs very fast, significantly faster than `c2rust-analyze` currently does, so I don't think it will be a bottleneck at all.  Also, wouldn't we have to run `c2rust-pdg` to generate the graphs pretty much always in order to ensure things to get out of sync (in `cargo test`)?  #835 is a much simpler and greater speedup for iteration speed on tests compared to this, which I think will be trivial until we start getting to really large/long event traces.

If we want to serialize everything, I think it's more worth it to only serialize the information necessary for `c2rust-analyze`, which is much less than all of the `Graphs` (see https://github.com/immunant/c2rust/pull/894#discussion_r1173607469), and doesn't require custom `serde` serialization `impl`s.  So I think it's better to just use `c2rust-pdg` as a library for now, recompute the PDG every time, which takes very little time, and then later move to a more refined serialization if we end up needing it for performance or something.

spernsteiner (approved):
I don't see much upside to throwing out the serialization code now that we've got it, but I think it will be fine either way

kkysen:
> That said, that approach means `c2rust-analyze` now needs to take the event log and metadata as inputs. Are we back to specifying file paths in environment variables for those, pending #709?

I think so?  We could try to add CLI args in somewhere, but I think it's more trouble than its worth before we do #709, and environment variables are pretty simple and seem good enough for now for this.  I think we'd also want to move `struct Pdg` and `fn Pdg::new` to the library, but not `fn Pdg::repr` and `struct PdgRepr`.

c2rust-analyze/src/main.rs
fw-immunant: Changed in db5ef66.

fw-immunant:
From this week's meeting, it sounds like we should just go ahead and merge this; doing so doesn't create any more work for us if we decide to minimize what gets communicated between c2rust-pdg and c2rust-analyze or move things in-process.

I'll add a test that exercises this and then hit the green button.

Issue #877, opened 2023-03-29
c2rust-analyze: handle uses of statics


spernsteiner:
On lighttpd, our second-largest source of panics the `Operand::Constant` case of `type_of`, which currently doesn't allow any pointers or references to appear in the type of the constant.  Constants of reference type are used for accessing statics - here is an example:
```Rust
static MY_STATIC: i32 = 12345;
let x = MY_STATIC;
// Produces the following MIR:
_2 = const {alloc1: &i32};
_1 = (*_2);
```
Here, `alloc1` is the static allocation used to store the value of `MY_STATIC`.  The value assigned to `_2` is an `Operand::Constant` of type `&i32`, which currently causes a panic in our static analysis.

There are several different cases observed in lighttpd that we'd eventually like to handle:

* [ ] Extern statics, like `stderr` in `fprintf(stderr, "some message\n")`
  * These can't be rewritten in general, but in some cases we might be able to apply a higher-level rewrite, such as `fprintf(stderr, ...)` to `eprintln!(...)`.
* [ ] `static mut` of primitive type, like lighttpd's `clockid_mono_coarse` and `log_monotonic_secs`
  * I think these can be rewritten into atomics, such as converting `i32` to `AtomicI32`.  We don't have to be too specific beyond that because even `Ordering::Relaxed` would provide more guarantees than the program was previously getting.
* [ ] `static mut` containing pointers, like `graceful_sockets` and `log_con_jqueue`
  * I don't have a good strategy for this, though we could maybe make some initial progress by assuming that the program is single-threaded.  This would let us use `thread_local!` `Cell<Box<T>>` or similar, instead of needing more complicated `Mutex` + `Arc` machinery.
* [ ] Lookup tables that don't actually need to be `mut`, like `name2id`, `static_table`, and `yy_shift_ofst`
  * `c2rust-transpile` already generates safe indexing operations for some of these, so all that's needed is to identify `static mut`s that are never mutated and convert them to non-`mut`.
* [ ] String and bytestring literals, like the `expect` call in `connection_write_cq`
  * These are not exactly statics, but look fairly similar in that they use an `Operand::Constant` of reference type.

Lookup tables are probably the easiest, but the handling of that case is entirely separate from the main analysis.  Aside from that, strings and bytestrings also seem relatively simple, and adding support for these to the dataflow analysis could provide some ideas on how to handle the more complex cases.

aneksteind:
[...]

PR #895, opened 2023-04-21
Rewrite never-mutated `static mut`s sans `mut`

fw-immunant:
This addresses point 4 of #877. ~It needs automated validation of the output of the testcase.~ Done.


kkysen:
By the way, I have noticed in rav1d that `c2rust transpile` translates all `const` `static`s into `static mut`s in Rust.

fw-immunant:
CI failed trying to clone tinycbor git. I love it.

c2rust-analyze/src/main.rs
```diff
@@ -456,6 +456,26 @@ fn run(tcx: TyCtxt) {
...
+        eprintln!("  {:?}", ldid);
+    }
+
+    // Assign global `PointerId`s for types of `static` items.
+    assert!(gacx.static_tys.is_empty());
+    gacx.static_tys = HashMap::with_capacity(all_static_ldids.len());
+    for &ldid in &all_static_ldids {
+        let did = ldid.to_def_id();
+        use DefKind::*;
+        if !matches!(tcx.def_kind(did), Static(..)) {
```
spernsteiner, line 472:
```suggestion
        if !matches!(tcx.def_kind(did), DefKind::Static(..)) {
```

c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -74,6 +76,10 @@ pub enum Rewrite<S = Span> {
     TySlice(Box<Rewrite>),
     /// `Foo<T1, T2>`
     TyCtor(String, Vec<Rewrite>),
+
+    // `static` builders
+    /// `static` mutability (`static` <-> `static mut`)
+    StaticMut(Mutability, S),
```
spernsteiner, line 82: It's unfortunate that this requires such a weird and specialized rewrite.  I was going to suggest adding a `Rewrite::Delete` variant that simply deletes the text in the span it's applied to, but looking at `hir::Item`, getting the exact span of the `mut` token looks like it would be a real pain.

c2rust-analyze/src/rewrite/statics.rs
```diff
@@ -0,0 +1,55 @@
...
+    let mut hir_rewrites = Vec::new();
+    for (did, lty) in gacx.static_tys.iter() {
+        // The map of statics and their ty + permissions tracks statics by did; map this to an Item
+        // node to look at the static's spans and declared mutability.
+        let item = if let Some(Node::Item(item)) = gacx.tcx.hir().get_if_local(*did) {
+            item
+        } else {
+            panic!("def id {:?} not found", did);
+        };
+        if let ItemKind::Static(_ty, mut_, _body_id) = item.kind {
```
spernsteiner, line 22: Variables of type `Mutability` are more commonly named `mutbl` (both here and in rustc)

spernsteiner (approved):
This all looks about right to me

c2rust-analyze/src/main.rs
```diff
@@ -456,6 +456,25 @@ fn run(tcx: TyCtxt) {
...
+    for &ldid in &all_static_ldids {
+        eprintln!("  {:?}", ldid);
+    }
+
+    // Assign global `PointerId`s for types of `static` items.
+    assert!(gacx.static_tys.is_empty());
+    gacx.static_tys = HashMap::with_capacity(all_static_ldids.len());
+    for &ldid in &all_static_ldids {
+        let did = ldid.to_def_id();
+        if !matches!(tcx.def_kind(did), DefKind::Static(..)) {
```
aneksteind, line 471: given the definition of `all_static_items`, when would this not be true? could you cut out the middleman and return a list of `DefId`s instead of `LocalDefId`s that have `DefKind::Static`?

c2rust-analyze/src/main.rs
```diff
@@ -679,6 +698,35 @@ fn run(tcx: TyCtxt) {
...
+            if lty.label == PointerId::NONE {
+                PermissionSet::empty()
+            } else {
+                gasn.perms[lty.label]
+            }
+        });
+        eprintln!(
+            "{:?}: addr_of = {:?}, type = {:?}",
+            name,
+            //describe_local(tcx, decl),
```
aneksteind, line 717: nit: remove before merging

c2rust-analyze/src/context.rs
```diff
@@ -445,11 +459,51 @@ impl<'tcx> TypeOf<'tcx> for PlaceRef<'tcx> {
...
+                            match acx.gacx.static_tys.get(&did) {
+                                Some(lty) => lty,
+                                None => panic!("did {:?} not found", did),
+                            }
+                        } else {
+                            panic!("no static found for alloc id {:?}", alloc_id)
+                        }
+                    } else {
+                        panic!("constant was of ptr type but not a Scalar pointing into a static allocation: {:?}", c)
+                    }
```
kkysen, line 502: nit: I think these `if let Some(...) = ... { ... } else { panic!(...); }`s can be `.unwrap_or_else(|| panic!(...))`s and that would get rid of the nesting.

c2rust-analyze/src/context.rs
```diff
@@ -445,11 +459,51 @@ impl<'tcx> TypeOf<'tcx> for PlaceRef<'tcx> {
...
 }
 
+fn const_alloc_id(c: &Constant) -> Option<AllocId> {
+    if let ConstantKind::Val(ConstValue::Scalar(interpret::Scalar::Ptr(ptr, _)), _ty) = c.literal {
+        return Some(ptr.provenance);
+    }
+    None
+}
+
+fn find_static_for_alloc<'tcx>(tcx: &TyCtxt<'tcx>, id: AllocId) -> Option<DefId> {
```
kkysen, line 469:
```suggestion
fn find_static_for_alloc(tcx: &TyCtxt, id: AllocId) -> Option<DefId> {
```

c2rust-analyze/src/main.rs
```diff
@@ -679,6 +698,35 @@ fn run(tcx: TyCtxt) {
...
+            addr_of1,
+            ty1,
+        );
```
kkysen, line 720: I'd also use inline formatting args (like `{name:?}`) here, as it's easy to mix up the order and filecheck needs them in a specific order (it happens with `perms` and `flags` in [fw/rewrite-never-mutated-static-muts-rebase](https://github.com/immunant/c2rust/tree/fw/rewrite-never-mutated-static-muts-rebase)).

c2rust-analyze/src/main.rs
fw-immunant: Yep, will do.

PR #897, opened 2023-04-24
`c2rust-ast-exporter`: fix segfault from iterator invalidation

maribu:
Using an iterator to iterate over an vector while modifying the underlying vector is invalid. This caused an segfault for a specific input, which is fixed by using operator[] for accessing the vector members instead.


kkysen (approved):
Thanks for hunting this down!  Looks like it was a difficult bug in https://github.com/RIOT-OS/RIOT/pull/19495, though at least the fix is quite simple.

PR #898, opened 2023-04-24
Fix mistranslation of side-effects inside unused unary operations

### LINKED ISSUE #896: c2rust-transpile: bug about side effect of expression
c2rust v0.17.0
Ubuntu 20.04.2

Consider the following C programs:
```c
// output:
// global_v = 0
// global_v = 1

// C programs
#include <stdio.h>
int global_v = 0;
int func();
int main(){ 
    printf("global_v = %d\n", global_v); 
    // the result of func() is not used, but necessary
    -func();
    printf("global_v = %d\n", global_v); 
}

int func() { 
    // change the value of global_v
    global_v = 1; 
    return global_v;
}
```
the generated Rust program ignore to translate the expression: "-func();", which make the different output between C and Rust.
```rust
// output:
// global_v = 0
// global_v = 0

// Rust program
use ::correctness_1::*;
extern "C" {
}
#[no_mangle]
pub static mut global_v: libc::c_int = 0 as libc::c_int;
    printf(b"global_v = %d\n\0" as *const u8 as *const libc::c_char, global_v);
    // ignore the expression: "-func();"
    printf(b"global_v = %d\n\0" as *const u8 as *const libc::c_char, global_v);
    return 0;
}
#[no_mangle]
pub unsafe extern "C" fn func() -> libc::c_int {
    global_v = 1 as libc::c_int;
    return global_v;
}
pub fn main() {
}
```
however, when the C code just call func() as the following:
```c
int main(){ 
    printf("global_v = %d\n", global_v); 
    // the result of func() is not used, but necessary
    func();
    printf("global_v = %d\n", global_v); 
}
```
the translated Rust code is correct:
```rust
    printf(b"global_v = %d\n\0" as *const u8 as *const libc::c_char, global_v);
    func();
    printf(b"global_v = %d\n\0" as *const u8 as *const libc::c_char, global_v);
    return 0;
}
```

So, the error occurs in c2rust/c2rust-transpile/src/translator/operators.rs
the following code about `Unop` doesn't consider about side effect of expression:
```rust
            // c2rust/c2rust-transpile/src/translator/operators.rs 1004-1029
            c_ast::UnOp::Plus => self.convert_expr(ctx.used(), arg), // promotion is explicit in the clang AST

            c_ast::UnOp::Negate => {
                let val = self.convert_expr(ctx.used(), arg)?;

                if resolved_ctype.kind.is_unsigned_integral_type() {
                    Ok(val.map(wrapping_neg_expr))
                } else {
                    Ok(val.map(neg_expr))
                }
            }
            c_ast::UnOp::Complement => Ok(self
                .convert_expr(ctx.used(), arg)?
                .map(|a| mk().unary_expr(UnOp::Not(Default::default()), a))),

            c_ast::UnOp::Not => {
                let val = self.convert_condition(ctx, false, arg)?;
                Ok(val.map(|x| mk().cast_expr(x, mk().path_ty(vec!["libc", "c_int"]))))
            }
            c_ast::UnOp::Extension => {
                let arg = self.convert_expr(ctx, arg)?;
                Ok(arg)
            }
```
could chage like the following, thus it can correctly consider about side effcet: 
```rust
           c_ast::UnOp::Plus => {
                let mut res = self.convert_expr(ctx.used(), arg)?;
                if ctx.is_unused() {
                    let v = res.clone().into_value();
                    res.stmts_mut().push(Stmt::Semi(*v, Default::default()));
                }
                Ok(res)
            } // promotion is explicit in the clang AST

            c_ast::UnOp::Negate => {
                let val = self.convert_expr(ctx.used(), arg)?;

                let mut res = if resolved_ctype.kind.is_unsigned_integral_type() {
                    val.map(wrapping_neg_expr)
                } else {
                    val.map(neg_expr)
                };
                if ctx.is_unused() {
                    let v = res.clone().into_value();
                    res.stmts_mut().push(Stmt::Semi(*v, Default::default()));
                }
                Ok(res)
            }
            c_ast::UnOp::Complement => {
                let mut res = self
                    .convert_expr(ctx.used(), arg)?
                    .map(|a| mk().unary_expr(UnOp::Not(Default::default()), a));
                if ctx.is_unused() {
                    let v = res.clone().into_value();
                    res.stmts_mut().push(Stmt::Semi(*v, Default::default()));
                }
                Ok(res)
            }

            c_ast::UnOp::Not => {
                let val = self.convert_condition(ctx, false, arg)?;
                let mut res = val.map(|x| mk().cast_expr(x, mk().path_ty(vec!["libc", "c_int"])));
                if ctx.is_unused() {
                    let v = res.clone().into_value();
                    res.stmts_mut().push(Stmt::Semi(*v, Default::default()));
                }
                Ok(res)
            }
            c_ast::UnOp::Extension => {
                let arg = self.convert_expr(ctx, arg)?;
                Ok(arg)
            }
```

fw-immunant:
Fixes #896.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -816,7 +816,7 @@ impl<'c> Translation<'c> {
         let ty = self.convert_type(ctype)?;
         let resolved_ctype = self.ast_context.resolve_type(ctype);
 
-        match name {
+        let mut unary = match name {
             c_ast::UnOp::AddressOf => {
                 let arg_kind = &self.ast_context[arg].kind;
 
@@ -944,6 +944,14 @@ impl<'c> Translation<'c> {
             c_ast::UnOp::Real | c_ast::UnOp::Imag | c_ast::UnOp::Coawait => {
                 panic!("Unsupported extension operator")
             }
+        }?;
+
+        // Unused unary operators (`-foo()`) may have side effects, so we need
+        // to add them to stmts.
+        if ctx.is_unused() {
+            let v = unary.clone().into_value();
+            unary.stmts_mut().push(Stmt::Semi(*v, Default::default()));
         }
+        Ok(unary)
     }
 }
--- a/tests/misc/src/exprs.c
+++ b/tests/misc/src/exprs.c
@@ -1,0 +1,15 @@
+int side_effect(int* x) {
+        *x = 1;
+        return 0;
+}
+
+void exprs(const unsigned n, int * const buffer) {
+
+        unsigned long i = 0;
+        char arr[1] = {0};
+
+        side_effect(&buffer[0]);
+        -side_effect(&buffer[1]);
+        &""[side_effect(&buffer[2])];
+        ++arr[side_effect(&buffer[3])];
+}
--- a/tests/misc/src/test_exprs.rs
+++ b/tests/misc/src/test_exprs.rs
@@ -1,0 +1,24 @@
+use crate::exprs::rust_exprs;
+use libc::c_int;
+use libc::c_uint;
+
+#[link(name = "test")]
+extern "C" {
+    fn exprs(_: c_uint, _: *mut c_int);
+}
+
+const BUFFER_SIZE: usize = 60;
+
+pub fn test_exprs() {
+    let mut buffer = [0; BUFFER_SIZE];
+    let mut rust_buffer = [0; BUFFER_SIZE];
+
+    unsafe {
+        exprs(BUFFER_SIZE as c_uint, buffer.as_mut_ptr());
+        rust_exprs(BUFFER_SIZE as c_uint, rust_buffer.as_mut_ptr());
+    }
+
+    for x in 0..BUFFER_SIZE {
+        assert_eq!(buffer[x], rust_buffer[x], "index {}", x);
+    }
+}
--- a/tests/misc/src/test_sizeofs.rs
+++ b/tests/misc/src/test_sizeofs.rs
@@ -1,4 +1,4 @@
-//! feature_core_intrinsics,
+//! feature_core_intrinsics, feature_label_break_value
 
 use crate::sizeofs::rust_sizeofs;
 use libc::c_int;


tests/misc/src/test_exprs.rs
```diff
@@ -0,0 +1,26 @@
+//! feature_core_intrinsics,
```
kkysen, line 1: Why do we need `#![feature(core_intrinsics)]` for this?

tests/misc/src/exprs.c
```diff
@@ -0,0 +1,17 @@
...
+void exprs(const unsigned n, int * const buffer) {
+
+        unsigned long i = 0;
+		char arr[1] = {0};
+
+        side_effect(&buffer[0]);
+        -side_effect(&buffer[1]);
+        &""[side_effect(&buffer[2])];
+		++arr[side_effect(&buffer[3])];
+}
```
kkysen, line 17: nit: Indentation.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -944,6 +944,14 @@ impl<'c> Translation<'c> {
             c_ast::UnOp::Real | c_ast::UnOp::Imag | c_ast::UnOp::Coawait => {
                 panic!("Unsupported extension operator")
             }
+        }?;
+
+        // Unused unary operators (`-foo()`) may have side effects, so we need
```
kkysen, line 949:
```suggestion
        // Unused unary operators (e.x. `-foo()`) may have side effects, so we need
```

kkysen (approved):
LGTM besides the little things I commented on.  Thanks, @zone-hai, for finding the bug and the clear fix!

tests/misc/src/test_exprs.rs
kkysen: I think it does need `//! feature_label_break_value`, though, as `#![feature(label_break_value)]` was only recently stabilized and we're still stuck on an older `rustc`.

fw-immunant: We don't--it was a holdover from copy-pasting another test case. Removed.

fw-immunant: Added `feature_label_break_value` for the `sizeof` test. I'm not sure why that one wasn't failing in CI prior to now.

tests/misc/src/exprs.c
kkysen: I meant the 8-space indentation.

tests/misc/src/test_exprs.rs
kkysen: Yeah, that's weird, but I guess if it's fixed now it's all good.

tests/misc/src/exprs.c
fw-immunant: I was following the pattern of `sizeofs.c`. We should normalize all these, but I didn't do a survey to see what was most common.

kkysen: Ahh, okay.

PR #899, opened 2023-04-25
Fix c2rust-ast-exporter's bugs when compiling on FreeBSD

oooutlk:
1. FreeBSD 10 and later versions remove libstd++ and switch to libc++. This is the same as macOS, so build.rs should detect `cfg!(target_os = "freebsd")`

2. FreeBSD make does not support gmake's -j option without an argument. Use gmake instead of make on FreeBSD when compiling tinycbor.


fw-immunant (approved):
Changes look sane. Thanks for working on improved platform support!

PR #900, opened 2023-04-25
(`c2rust-analyze`) Cleanup/simplify `visit_cast` (extracted from #883)

kkysen:
This cleans up and simplifies some of the code from #883 in preparation for the string cast PRs, #739 and #741.

It extracts things into a separate `visit_cast` method, simplifies a few things, and clarifies exactly which `Ty`s and `LTy`s are from and to in a cast.  This makes it easier to understand and read, and makes it simpler to rebase #739 onto it.

### DIFF
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -127,14 +127,12 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 assert_matches!(op_lty.kind(), TyKind::Ref(..) | TyKind::RawPtr(..));
                 let op_pointee_lty = assert_matches!(op_lty.args, [lty] => lty);
 
-                assert_matches!(pointee_ty.kind(), TyKind::Slice(..));
-                if let TyKind::Slice(elem_ty) = *pointee_ty.kind() {
-                    assert!(matches!(op_pointee_lty.kind(), TyKind::Array(..)));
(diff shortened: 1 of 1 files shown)


c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -148,59 +196,8 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-                self.visit_operand(op)
-            }
-            Rvalue::Cast(_cast_kind, ref op, _) => {
-                // A cast such as `T as U`
-                let casted_from = self.acx.type_of(op);
-                let casted_to = rvalue_lty;
-                assert!(is_castable_to(casted_from, casted_to));
-                self.visit_operand(op)
+            Rvalue::Cast(cast_kind, ref op, ty) => {
+                self.visit_cast(cast_kind, op, ty, rvalue_lty);
```
aneksteind, line 200: `rvalue_lty.ty` and `ty` should be the same here, so I believe you don't need two arguments and can pass `rvalue_lty` alone

aneksteind: you could make an assertion that they're the same if you'd like to be safe

aneksteind (approved):
LGTM after the arguments comment is addressed

c2rust-analyze/src/dataflow/type_check.rs
kkysen: Fixed in 3906c892b16a349053cf7e9c65d66f192d34e23a.

PR #901, opened 2023-04-27
(`c2rust-analyze`) Further clarify the behavior and semantics of const refs

kkysen:


kkysen:
I was wrong about some of the semantics here, so closing this in favor of #902.

PR #902, opened 2023-04-27
(`c2rust-analyze`) Restrict const ref handling to just string literals

kkysen:
This restricts the current const ref handling to just string literals, as const refs are too complex as a whole.  I previously though `ConstantKind::Val` vs. `::Ty` was a good way to disambiguate them, but it works differently than I thought (statics, for example).

This currently uses a very hacky way to detect string literals from `Constant`s by using its `Display` `impl`.
~I'll replace that soon.~  On second thought, a more correct implementation derived from `Constant`'s `Display` `impl` is exceedingly complex and quite large, so I think this hacky way might actually be better.

### DIFF
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -361,6 +361,33 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
     }
 }
 
+/// Label const refs.
+///
+/// By const ref, we mean the ref itself is constant.
+///
+/// This can be a reference to an inline value ([`ConstantKind::Val`]), such as:
+/// * `1`
+/// * `[]`
+/// * `()`
+/// * `""`
+/// * `b""`
+///
+/// As these are only inline values, they do not need
+/// [`DataflowConstraints`] to their pointee (unlike their named counterparts),
+/// because the values cannot be accessed elsewhere,
+/// and their permissions are predetermined (see [`PermissionSet::for_const_ref_ty`]).
+///
+/// Or it can be a reference to a named item ([`ConstantKind::Ty`])
+/// that is accessible at compile time, such as:
+/// * `const`s
+/// * `static`s
+/// * `fn`s
+///
+/// Note that these named items are often global,
+/// but could also be local to a function or smaller scope.
+///
+/// When these are handled (they are not yet currently), they will need to have
+/// the correct [`DataflowConstraints`] to their pointee.
 fn label_const_refs<'tcx>(
     acx: &mut AnalysisCtxt<'_, 'tcx>,
     c: &Constant<'tcx>,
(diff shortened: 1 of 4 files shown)


kkysen:
@fw-immunant, I think with this, your statics PR (#895) should work again.  The `::log::error!` here on non-string literal const refs can be removed/replaced with whatever static handling is needed.

fw-immunant:
> @fw-immunant, I think with this PR, your statics PR (#895) should work again. The `::log::error!` here on non-string literal const refs can be removed/replaced with whatever static handling is needed.

Atop this PR, #895 passes tests if I comment out the `::log::error!` that you mention. Probably the right thing to do is to add a no-op case for constants that reference static items, and still `log::error` otherwise. We don't need to do anything other than a no-op for pointer IDs of `static`s in `label_string_literals` because statics are handled separately, and we want to assign them pointer IDs even if they are never used as `Rvalue`s.

Speaking of which, the naming is a bit confusing: if `label_string_literals` is supposed to do what it says on the tin, it shouldn't be responsible for labelling "non-string literal const refs" (the case where it calls `log::error`, should it?

kkysen:
I could make `label_string_literals` be a complete no-op for non-string literals (no `::log::error!`), or I could rename it `label_const_refs` and leave the `::log::error!`, which will you'll replace with a no-op in #895 for just statics.  Which do you think I should do?

fw-immunant:
I would say the former--when we know what to do with other const refs we can decide whether it makes sense to do it in the same place or not (e.g. for const refs of statics it doesn't).

kkysen:
> I would say the former--when we know what to do with other const refs we can decide whether it makes sense to do it in the same place or not (e.g. for const refs of statics it doesn't).

The `log::error!` is meant to just alert you of an error earlier than when it's caught later on in a `label_no_pointers`, which is less clear.  That's why it's a `log::error!` not a `panic!`, too, so it can be bypassed with `RUST_LOG_PANIC`.  But I think I'll still just take it out here.

c2rust-analyze/src/main.rs
```diff
@@ -361,6 +361,53 @@ impl<'tcx> Debug for AdtMetadataTable<'tcx> {
...
+/// [`Display`]: std::fmt::Display
+fn is_string_literal(c: &Constant) -> bool {
+    let s = c.to_string();
+    s.ends_with('"') && {
+        let s = match s.strip_prefix("const ") {
+            Some(s) => s,
+            None => return false,
+        };
+        s.starts_with('"') || s.starts_with("b\"")
+    }
```
fw-immunant, line 379: This really is horrible, but the predicate is well-defined so a different implementation should be easy to slot in.

kkysen: Yeah, it's unfortunate.  I think this actually also has a lower chance of breaking in future `rustc` versions than an implementation based on what the `Display` `impl` does.

aneksteind (approved):

spernsteiner (approved):

Issue #873, opened 2023-03-23
Currently neither field expressions (e.g. `my_struct.x`) nor field type declarations (e.g. `x: *mut i32`) get rewritten.


aneksteind:
Currently neither field expressions (e.g. `my_struct.x`) nor field type declarations (e.g. `x: *mut i32`) get rewritten.

_Originally posted by @spernsteiner in https://github.com/immunant/c2rust/pull/815#discussion_r1119219040_

spernsteiner:
Lack of rewriting of field types is a major source of compilation errors on lighttpd.  For example:

```
error[E0608]: cannot index into a value of type `*mut i8`
    --> ../src/main.new.rs:5992:12
     |
5992 |         &(&(((*hdr).buf))[(((*hdr).name_offset as libc::c_int as isize) as usize) ..])[0]
     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The problem here is that the `buf` field hasn't been rewritten into a slice.

aneksteind:
I'm working on this

PR #910, opened 2023-05-03
Rewrite field declarations

aneksteind:
Addresses https://github.com/immunant/c2rust/issues/873 in part.

This only rewrites struct field declarations. It doesn't rewrite the struct lifetime parameters and their use within the struct's field declarations.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -468,10 +468,9 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
             let field_def = &adt_def.non_enum_variant().fields[field.index()];
             let field_def_name = field_def.name;
             eprintln!("projecting into {adt_def:?}.{field_def_name:}");
-            let field_lty: &crate::labeled_ty::LabeledTyS<PointerId> =
-                self.gacx.field_ltys.get(&field_def.did).unwrap_or_else(|| {
-                    panic!("Could not find {adt_def:?}.{field_def_name:?} in field type map")
-                });
+            let field_lty: LTy = self.gacx.field_ltys.get(&field_def.did).unwrap_or_else(|| {
+                panic!("Could not find {adt_def:?}.{field_def_name:?} in field type map")
+            });
             field_lty
         };
         util::lty_project(lty, proj, projection_lty)
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -678,7 +678,7 @@ impl LocalAssignment {
 }
 
 pub struct Assignment<'a> {
-    global: &'a mut GlobalAssignment,
+    pub global: &'a mut GlobalAssignment,
     local: &'a mut LocalAssignment,
 }
 
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -155,7 +155,7 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
-    pub field_tys: HashMap<DefId, LTy<'tcx>>,
+    pub field_ltys: HashMap<DefId, LTy<'tcx>>,
 
     pub static_tys: HashMap<DefId, LTy<'tcx>>,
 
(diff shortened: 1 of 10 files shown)


c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -42,6 +42,12 @@ pub use self::statics::gen_static_rewrites;
 pub use self::ty::dump_rewritten_local_tys;
 pub use self::ty::gen_ty_rewrites;
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum LifetimeType {
```
fw-immunant, line 46: Bikeshed: `LifetimeName` rather than `LifetimeType` to avoid confusion with actual types.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -380,25 +423,26 @@ pub fn gen_ty_rewrites<'tcx>(
...
-        v.handle_ty(rw_lty, hir_ty);
+        let rw_lty = relabel_rewrites(&asn.perms(), &asn.flags(), rw_lcx, lty_sig.output);
+        v.handle_ty(rw_lty, hir_ty, LifetimeType::Elided);
     }
 
     let hir_body_id = acx.tcx().hir().body_owned_by(ldid);
     let body = acx.tcx().hir().body(hir_body_id);
     intravisit::Visitor::visit_body(&mut v, body);
 
-    // TODO: wrap locals in `Cell` if `addr_of_local` indicates that it's needed
```
fw-immunant, line 396: Is this TODO outdated (and the Cell PR should have removed this comment), or am I missing wheres this PR is doing something to locals?

aneksteind: the Cell PR should have removed this comment

fw-immunant (approved):
Seems sane in general, though I am curious how it will look to "get actual lifetime from ADT/Field metadata". A couple minor comments inline.

c2rust-analyze/src/rewrite/mod.rs
aneksteind: eec91a8e

c2rust-analyze/src/context.rs
```diff
@@ -463,15 +463,16 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
...
+    pub fn projection_lty(&self, lty: LTy<'tcx>, proj: &PlaceElem<'tcx>) -> LTy<'tcx> {
         let projection_lty = |_lty: LTy, adt_def: AdtDef, field: Field| {
             let field_def = &adt_def.non_enum_variant().fields[field.index()];
             let field_def_name = field_def.name;
             eprintln!("projecting into {adt_def:?}.{field_def_name:}");
-            let res = *self.gacx.field_tys.get(&field_def.did).unwrap_or_else(|| {
-                panic!("Could not find {adt_def:?}.{field_def_name:?} in field type map")
-            });
-            res
+            let field_lty: &crate::labeled_ty::LabeledTyS<PointerId> =
```
spernsteiner, line 471: Is this annotation autogenerated?  It would be clearer to use the `LTy` here, or `LabeledTy<_>`, rather than the more verbose and less common `&LabeledTyS<_>`.  Or else remove the annotation and add the `*` back to the RHS.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -264,7 +272,12 @@ struct HirTyVisitor<'a, 'tcx> {
 }
 
 impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
-    fn handle_ty(&mut self, rw_lty: RwLTy<'tcx>, hir_ty: &hir::Ty<'tcx>) {
+    fn handle_ty(
+        &mut self,
+        rw_lty: RwLTy<'tcx>,
+        hir_ty: &hir::Ty<'tcx>,
+        lifetime_type: LifetimeName,
+    ) {
```
spernsteiner, line 280: Would it make sense to include the `LifetimeName` in the `RwLTy` label instead of passing it separately?  That might make it easier to work with when there are nested pointers with different lifetimes at each level.

spernsteiner (approved):

c2rust-analyze/src/context.rs
aneksteind: Yes, it was auto-generated. Updated in 4b2537b4

c2rust-analyze/src/rewrite/ty.rs
aneksteind: I did something like this in #915 [here](https://github.com/immunant/c2rust/pull/915/files#diff-1f3bd06b6665a1221155482f5a88af065b3e807fc3fa48e543e39050416eedb3R385) (need to expand `ty.rs` to see)

PR #911, opened 2023-05-03
analyze: bring static handling in line with locals

spernsteiner:
Gives statics separate `addr_of_static` (`PointerId`) and `static_tys` (`LTy`) information, analogous to `addr_of_local` and `local_tys`.  Previously they had only a single `LTy` based on `tcx.static_ptr_ty(did)`, which gives somewhat weird results (it's either a `RawPtr` or a `Ref` depending on whether or not the static is declared as mut).


aneksteind (approved):
I think the changes for `print_labeling_for_var` should be broken out into its own commit

PR #915, opened 2023-05-05
Feat.rewrite.lifetime

aneksteind:
Rewrites structs to reflect hypothetical lifetimes accepted by polonius.

e.g.
```rust
pub struct Data<'d> {
    pub pi: *mut i32,
    pub pa: *mut A<'d>,
    pub a: A<'d>,
}
pub struct A<'a> {
    pub rd: &'a Data<'a>,
    pub pra: *mut &'a mut A<'a>,
}

struct VecTup<'a> {
    bar: *mut Vec<(VecTup<'a>, *mut A<'a>)>,
}
```

gets rewritten to

```rust
pub struct Data<'d,'h0,'h1,'h2> {
    pub pi: &'h0 mut (i32),
    pub pa: &'h1 mut (A<'d,'h0,'h1,'h2>),
    pub a: A<'d,'h0,'h1,'h2>,
}

pub struct A<'a,'h0,'h1,'h2> {
    pub rd: &'a (Data<'a,'h0,'h1,'h2>),
    pub pra: &'h2 core::cell::Cell<(&'a (A<'a,'h0,'h1,'h2>))>,
}

struct VecTup<'a,'h3,'h4,'h0,'h1,'h2> {
    bar: &'h3 (Vec<(VecTup<'a,'h3,'h4,'h0,'h1,'h2>, &'h4 (A<'a,'h0,'h1,'h2>))>),
}
```

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -1,3 +1,4 @@
+use crate::borrowck::{AdtMetadata, FieldMetadata, OriginArg, OriginParam};
 use crate::c_void_casts::CVoidCasts;
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
 use crate::pointer_id::{
@@ -6,17 +7,25 @@ use crate::pointer_id::{
 };
 use crate::util::{self, describe_rvalue, PhantomLifetime, RvalueDesc};
 use crate::AssignPointerIds;
+use assert_matches::assert_matches;
 use bitflags::bitflags;
+use indexmap::IndexSet;
 use log::*;
-use rustc_hir::def_id::DefId;
+use rustc_ast::Mutability;
+use rustc_hir::def::DefKind;
+use rustc_hir::def_id::{DefId, LocalDefId};
 use rustc_index::vec::IndexVec;
 use rustc_middle::mir::interpret::{self, AllocId, ConstValue, GlobalAlloc};
 use rustc_middle::mir::{
     Body, Constant, ConstantKind, Field, HasLocalDecls, Local, LocalDecls, Location, Operand,
     Place, PlaceElem, PlaceRef, Rvalue,
 };
-use rustc_middle::ty::{AdtDef, FieldDef, Ty, TyCtxt, TyKind};
+use rustc_middle::ty::{
+    tls, AdtDef, FieldDef, GenericArgKind, GenericParamDefKind, Ty, TyCtxt, TyKind,
+};
+use rustc_type_ir::RegionKind::{ReEarlyBound, ReStatic};
 use std::collections::HashMap;
+use std::fmt::Debug;
 use std::ops::Index;
 
 bitflags! {
@@ -180,6 +189,83 @@ bitflags! {
     }
 }
 
+pub struct AdtMetadataTable<'tcx> {
+    pub table: HashMap<DefId, AdtMetadata<'tcx>>,
+    pub struct_dids: Vec<DefId>,
+}
+
+impl<'tcx> Debug for AdtMetadataTable<'tcx> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        fn fmt_string(lty: LabeledTy<'_, &[OriginArg]>) -> String {
+            let args: Vec<String> = lty.args.iter().map(|t| fmt_string(t)).collect();
+            use rustc_type_ir::TyKind::*;
+            match lty.kind() {
+                Ref(..) | RawPtr(..) => {
+                    format!("&{:?} {:}", lty.label[0], args[0])
+                }
+                Adt(adt, _) => {
+                    let mut s = format!("{adt:?}");
+                    let params = lty
+                        .label
+                        .iter()
+                        .map(|p| format!("{:?}", p))
+                        .into_iter()
+                        .chain(args.into_iter())
+                        .collect::<Vec<_>>()
+                        .join(",");
+
+                    if !params.is_empty() {
+                        s.push('<');
+                        s.push_str(&params);
+                        s.push('>');
+                    }
+                    s
+                }
+                Tuple(_) => {
+                    format!("({:})", args.join(","))
+                }
+                _ => format!("{:?}", lty.ty),
+            }
+        }
+
+        tls::with_opt(|tcx| {
+            let tcx = tcx.unwrap();
+            for k in &self.struct_dids {
+                let adt = &self.table[k];
+                let other_param_names = tcx.generics_of(k).params.iter().filter_map(|p| {
+                    if !matches!(p.kind, GenericParamDefKind::Lifetime) {
+                        Some(p.name.to_ident_string())
+                    } else {
+                        None
+                    }
+                });
+                write!(f, "struct {:}", tcx.item_name(*k))?;
+                write!(f, "<")?;
+                let lifetime_params_str = adt
+                    .lifetime_params
+                    .iter()
+                    .map(|p| format!("{:?}", p))
+                    .chain(other_param_names)
+                    .collect::<Vec<_>>()
+                    .join(",");
+                write!(f, "{lifetime_params_str:}")?;
+                writeln!(f, "> {{")?;
+                for (fdid, fmeta) in &adt.field_info {
+                    write!(f, "\t{:}: ", tcx.item_name(*fdid))?;
+                    let field_string_lty = fmt_string(fmeta.origin_args);
+
+                    write!(f, "{field_string_lty:}")?;
+
+                    writeln!(f)?;
+                }
+
+                writeln!(f, "}}\n")?;
+            }
+            writeln!(f)
+        })
+    }
+}
+
 pub struct GlobalAnalysisCtxt<'tcx> {
     pub tcx: TyCtxt<'tcx>,
     pub lcx: LTyCtxt<'tcx>,
@@ -192,6 +278,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub static_tys: HashMap<DefId, LTy<'tcx>>,
     pub addr_of_static: HashMap<DefId, PointerId>,
+
+    pub adt_metadata: AdtMetadataTable<'tcx>,
 }
 
 pub struct AnalysisCtxt<'a, 'tcx> {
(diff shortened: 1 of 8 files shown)


c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -330,10 +440,19 @@ impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
             self.hir_rewrites.push((hir_ty.span, rw));
         }
 
+        if let TyKind::Adt(adt_def, substs) = rw_lty.ty.kind() {
+            if let Some(adt_metadata) = &self.acx.gacx.adt_metadata.table.get(&adt_def.did()) {
+                self.hir_rewrites.push((
+                    hir_ty.span,
+                    adt_ty_rw(adt_def, &adt_metadata.lifetime_params, substs),
```
spernsteiner, line 447: `adt_metadata.lifetime_params` tells you what lifetimes are declared on the type, not what lifetimes should be passed at this particular use of it.  Due to the way we infer hypothetical lifetimes in ADTs, it happens to be the case that the names match up within struct fields (given `struct Foo<'h0> { ... }`, all uses will look like `struct Bar<'h0, 'h1> { foo: &'h1 Foo<'h0> }`, with the same name `'h0` in both places), but this isn't true in general.

```Rust
struct Foo<'h0> { ... }
fn foo1<'a>(x: Foo<'a>, y: Foo<'a>) { ... }
fn foo2<'a, 'b>(x: Foo<'a>, y: Foo<'b>) { ... }
```

Given `Foo`, both the `foo1` and `foo2` signatures are valid, and they mean different things, so we need a way to express both of them.

`RewriteLabel` probably needs a field like `&'tcx [OriginArg<'tcx>]` representing the lifetime arguments of a particular `Ty`, with similar meaning to the labels of `FieldMetadata`'s `origin_args` field.  That will allow for distinctions like the above `foo1`/`foo2` example.

For struct field types, the lifetime arguments can be set from the existing `FieldMetadata::origin_args` labels.  For types in function signatures, we'll eventually need a similar setup to generate hypothetical parameters and so on, but for now that doesn't exist, so any region arguments should either be elided or left as `'static`.  Inside function bodies, the regions can always be elided.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -25,57 +25,124 @@ use rustc_hir::Mutability;
...
 
-type RwLTy<'tcx> = LabeledTy<'tcx, RewriteLabel>;
+type RwLTy<'tcx> = LabeledTy<'tcx, RewriteLabel<'tcx>>;
+
+fn has_lifetime_rws(rw_lty: &RwLTy, adt_metadata: &AdtMetadataTable) -> bool {
+    let has_pointer_lifetime = rw_lty
+        .label
+        .lifetime
+        .iter()
+        .any(|lt| matches!(lt, OriginArg::Hypothetical(..)));
```
kkysen, line 52:
```suggestion
    let has_pointer_lifetime = matches!(rw_lty.label.lifetime, Some(&OriginArg::Hypothetical(..)));
```

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -25,57 +25,124 @@ use rustc_hir::Mutability;
...
 
 /// A label for use with `LabeledTy` to indicate what rewrites to apply at each position in a type.
 #[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
-struct RewriteLabel {
+struct RewriteLabel<'a> {
     /// Rewrite a raw pointer, whose ownership and quantity have been inferred as indicated.
     ty_desc: Option<(Ownership, Quantity)>,
     /// If set, a child or other descendant of this type requires rewriting.
     descendant_has_rewrite: bool,
+    // A lifetime rewrite for a pointer or reference
```
kkysen, line 41:
```suggestion
    /// A lifetime rewrite for a pointer or reference.
```

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -138,7 +205,28 @@ fn deconstruct_hir_ty<'a, 'tcx>(
...
+                        for mir_arg in substs.types().into_iter().skip(type_args.len()) {
+                            eprintln!("\t{:?}", mir_arg)
+                        }
+                    } else if type_args.len() != substs.types().count() {
+                        panic!("mismatched number of type arguments for {adt_def:?} and {hir_ty:?}")
+                    }
+                    Some(type_args)
+                }
+                _ => None,
+            }
```
kkysen, line 224:
```suggestion
            hir_generic_ty_args(hir_ty).map(|type_args| {
                if type_args.len() < substs.types().count() {
                    // this situation occurs when there are hidden type arguments
                    // such as the allocator `std::alloc::Global` type argument in `Vec`
                    eprintln!("warning: extra MIR type argument for {adt_def:?}:");
                    for mir_arg in substs.types().into_iter().skip(type_args.len()) {
                        eprintln!("\t{:?}", mir_arg)
                    }
                } else if type_args.len() != substs.types().count() {
                    panic!("mismatched number of type arguments for {adt_def:?} and {hir_ty:?}")
                }
                type_args
            })
```

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -268,20 +356,37 @@ fn mk_rewritten_ty<'tcx>(
...
 impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
-    fn handle_ty(
-        &mut self,
-        rw_lty: RwLTy<'tcx>,
-        hir_ty: &hir::Ty<'tcx>,
-        lifetime_type: LifetimeName,
-    ) {
-        if rw_lty.label.ty_desc.is_none() && !rw_lty.label.descendant_has_rewrite {
+    fn handle_ty(&mut self, rw_lty: RwLTy<'tcx>, hir_ty: &hir::Ty<'tcx>) {
+        if !matches!(rw_lty.ty.kind(), TyKind::Adt(..))
```
kkysen, line 386:
```suggestion
        if !rw_lty.ty.is_adt()
```

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -347,10 +466,63 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
...
+
+                        for ul in new_lifetime_params_iter {
+                            new_lifetimes.push(Rewrite::PrintTy(format!("{:?}", ul)))
+                        }
+
+                        updated_lifetimes
+                            .into_iter()
+                            .chain(new_lifetimes.into_iter())
+                            .chain(other_params.into_iter())
+                            .collect()
```
kkysen, line 517:
```suggestion
                        [updated_lifetimes, new_lifetimes, other_params].into_iter().flatten().collect()
```

c2rust-analyze/src/rewrite/ty.rs
aneksteind: Thanks for the catch, the tests were wrong and didn't complain about this. I've addressed it in 37e14bc1 and took out the test for the function lifetime rewriting for the time being; I'll create an issue for it

c2rust-analyze/src/rewrite/ty.rs
aneksteind: addressed in 37e14bc1

c2rust-analyze/src/rewrite/ty.rs
aneksteind: Issue created #921

aneksteind:

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -266,20 +348,37 @@ fn mk_rewritten_ty<'tcx>(
...
+    rw_lcx: LabeledTyCtxt<'tcx, RewriteLabel<'tcx>>,
     mir: &'a Body<'tcx>,
     hir_rewrites: Vec<(Span, Rewrite)>,
     hir_span_to_mir_local: HashMap<Span, rustc_middle::mir::Local>,
 }
 
+fn adt_ty_rw<S>(
+    adt_def: &AdtDef,
+    lifetime_params: &[OriginArg],
+    substs: &&List<GenericArg>,
```
spernsteiner, line 360:
```suggestion
    substs: &List<GenericArg>,
```
Haven't tested this, but I wouldn't expect it to break anything, since there's an implicit coercion from `&&T` to `&T`

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -328,10 +433,19 @@ impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
             self.hir_rewrites.push((hir_ty.span, rw));
         }
 
+        if let TyKind::Adt(adt_def, substs) = rw_lty.ty.kind() {
+            if !rw_lty.label.lifetime.is_empty() {
+                self.hir_rewrites.push((
```
spernsteiner, line 438: I was expecting this to generate a rewrite on every ADT with at least one lifetime, which would lead to unnecessary unfolding of type aliases.  But I didn't see that effect when I ran it on this code:

```Rust
struct Test<'a>(&'a u8);
type Alias<'a> = Test<'a>;
fn f<'a>(x: Alias<'a>) {}
```

I was expecting a rewrite from `x: Alias<'a>` to `x: Test<'a>`, but no such rewrite occurred.  Under what conditions is `label.lifetime` non-empty?  I was expecting it to be set to the original lifetime args `['a]` in this case.  Is it empty for structs that have no hypothetical lifetimes?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -345,10 +459,62 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
...
+                                    let updated_lifetime_param = new_lifetime_params_iter
+                                        .next()
+                                        .expect("Not enough updated_lifetime_params");
+                                    updated_lifetimes.push(Rewrite::PrintTy(format!(
+                                        "{:?}",
+                                        updated_lifetime_param
+                                    )))
+                                }
+                                _ => {
+                                    other_params.push(Rewrite::PrintTy(gp.name.ident().to_string()))
```
spernsteiner, line 497: Will this swallow constraints, rewriting `struct Foo<T: Clone>(...)` into `struct Foo<T>(...)`?

spernsteiner (commented):
This generally looks okay, but I have questions about a couple specific rewrite cases

spernsteiner:
I think ideally we would be able to either insert additional lifetime arguments before/after some existing argument (`<T>` -> `<'h0, T>` or `<'a>` -> `<'a, 'h0>`, with the rewrite being placed on a zero-width span `<{}T>` or `<'a{}>`) or generate a whole generics list from scratch if none is present (`Foo` -> `Foo<'h0>`, again on the zero-width span `Foo{}`).  This would minimize disruption of the existing source code.  But it might be better to just add a `TODO` or open an issue to avoid this complexity for now.

c2rust-analyze/src/context.rs
```diff
@@ -219,6 +307,185 @@ pub struct AnalysisCtxtData<'tcx> {
     next_ptr_id: NextLocalPointerId,
 }
 
+fn construct_adt_metadata<'tcx>(tcx: TyCtxt<'tcx>) -> AdtMetadataTable {
```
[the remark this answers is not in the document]
aneksteind: correct

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -218,13 +300,13 @@ fn mk_cell<'tcx>(tcx: TyCtxt<'tcx>, ty: ty::Ty<'tcx>) -> ty::Ty<'tcx> {
     };
 
     let cell_adt = tcx.adt_def(cell_struct_did);
-    let substs = tcx.mk_substs([GenericArg::from(ty)].into_iter());
+    let substs = tcx.mk_substs([subst::GenericArg::from(ty)].into_iter());
```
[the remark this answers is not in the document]
aneksteind: yes, there is a difference between ones at the HIR and MIR level

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -22,55 +24,117 @@ use rustc_hir::Mutability;
...
+    pointer_lty: LTy<'tcx>,
+    args: &[RwLTy<'tcx>],
+    perms: &impl Index<PointerId, Output = PermissionSet>,
+    flags: &impl Index<PointerId, Output = FlagSet>,
+    lifetime: &'tcx [OriginArg<'tcx>],
+    adt_metadata: &AdtMetadataTable,
+) -> RewriteLabel<'tcx> {
+    let ty_desc = Some(pointer_lty.label)
+        .filter(|ptr| !ptr.is_none())
+        .map(|ptr| {
```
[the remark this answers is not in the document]
aneksteind: I agree, changed in 86da1947

c2rust-analyze/src/rewrite/ty.rs
aneksteind: `Alias` doesn't get rewritten because it doesn't appear in the ADT metadata table, and that's because it's `DefKind` is a `TyAlias` and not one of `Struct | Enum | Union`

aneksteind: Is support for `TyAlias` for `structs` something that can be done in a follow-up PR?

c2rust-analyze/src/rewrite/ty.rs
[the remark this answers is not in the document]
spernsteiner: Both the types I'm looking at are MIR/Ty level, though.   The `GenericArg` in scope is `rustc_middle::ty::GenericArg`, while `subst::GenericArg` is `rustc_middle::ty::subst::GenericArg`.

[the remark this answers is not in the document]
aneksteind: fixed in bca93849

c2rust-analyze/src/rewrite/ty.rs
aneksteind: Issues created in https://github.com/immunant/c2rust/issues/926

c2rust-analyze/src/rewrite/ty.rs
aneksteind: Issue created in https://github.com/immunant/c2rust/issues/927. Would you mind elaborating on that issue as to how fresh hypothetical lifetimes would be spliced in?

spernsteiner (approved):

PR #918, opened 2023-05-09
analyze: minor fixes

spernsteiner:
A couple small fixes split out from the `FIXED` work

* 85bd2ac11982dcd4892331f59bd7c1dc20dfa8e4: Makes `pointer_table[PointerId::NONE]` panic with a clearer message (it already panics, but the message is rather opaque).
* e02e0b8ade0a98519cd1fcd78306860b8cb40c4f: The `pointee_ty` field of `Callee::PtrOffset` currently is always set to the `T` from `impl<T> *mut T { ... }`; this commit fixes it to be the actual pointee type at the use site.  For example, on `(x: *mut i32).offset(...)`, `pointee_ty` will now be `i32`.

Based on #876 for now; once this is approved, I'll rebase onto `master`.

### DIFF
--- a/c2rust-analyze/src/util.rs
+++ b/c2rust-analyze/src/util.rs
@@ -7,7 +7,9 @@ use rustc_middle::mir::{
     BasicBlock, BasicBlockData, Constant, Field, Local, Location, Mutability, Operand, Place,
     PlaceElem, PlaceRef, ProjectionElem, Rvalue, Statement, StatementKind,
 };
-use rustc_middle::ty::{self, AdtDef, DefIdTree, SubstsRef, Ty, TyCtxt, TyKind, UintTy};
+use rustc_middle::ty::{
+    self, AdtDef, DefIdTree, EarlyBinder, Subst, SubstsRef, Ty, TyCtxt, TyKind, UintTy,
+};
 use std::fmt::Debug;
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


aneksteind (approved):

kkysen (approved):

c2rust-analyze/src/util.rs
```diff
@@ -203,7 +205,7 @@ fn builtin_callee(tcx: TyCtxt, did: DefId) -> Option<Callee> {
             if tcx.impl_trait_ref(parent_did).is_some() {
                 return None;
             }
-            let parent_impl_ty = tcx.type_of(parent_did);
+            let parent_impl_ty = EarlyBinder(tcx.type_of(parent_did)).subst(tcx, substs);
```
spernsteiner, line 208: > could you comment on why this is necessary
>
> _Originally posted by @aneksteind in https://github.com/immunant/c2rust/pull/876#discussion_r1190176434_

Previously, for a call like `(x: *mut i32).offset(...)`, we would look up the `offset` method, find its containing `impl<T> *mut T { ... }`, take the impl type `*mut T`, extract the pointee type `T`, and record that as the `pointee_ty` in `Callee::PtrOffset`.  But having `T` here is completely useless (all `PtrOffset`s had the same `pointee_ty` due to this bug).  What we actually want in this example is for `pointee_ty` to be `i32`, which is the pointee type for this particular `offset` call.  This substitution takes the impl type `*mut T` and substitutes in the `substs` used for this particular call, which in our example produces `*mut i32`.  From there, the remaining steps extract the desired `i32` pointee type, and we end up with `PtrOffset { pointee_ty: i32 }`.

PR #919, opened 2023-05-09
analyze: refactor MIR rewrite generation

spernsteiner:
This branch refactors `mir_op` to make the generation of conversions more uniform.  Ideally, we should call `emit_cast_*` in roughly every place where `dataflow::type_check` adds a dataflow edge.  This branch doesn't fully implement that, but gets us a little closer.

This also adds a new `TypeDesc` type, which bundles up `Ownership`, `Quantity`, and a pointee `Ty` into a complete description of a rewritten pointer type.  This is used for some of the new `emit_cast_*` machinery.

Based on #918 for now; once this is approved, I'll rebase onto `master`

### DIFF
--- a/c2rust-analyze/src/rewrite/expr/mir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/mir_op.rs
@@ -135,6 +135,17 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
         match stmt.kind {
             StatementKind::Assign(ref x) => {
                 let (pl, ref rv) = **x;
+
+                if matches!(rv, Rvalue::Cast(..)) && self.acx.c_void_casts.should_skip_stmt(loc) {
+                    // This is a cast to or from `void*` associated with a `malloc`, `free`, or
+                    // other libc call.
+                    //
+                    // TODO: we should probably emit a rewrite here to remove the cast; then when
+                    // we implement rewriting of the actual call, it won't need to deal with
+                    // additional rewrites at a separate location from the call itself.
+                    return;
+                }
+
                 let pl_lty = self.acx.type_of(pl);
 
                 if pl.is_indirect() && self.acx.local_tys[pl.local].ty.is_any_ptr() {
--- a/c2rust-analyze/src/rewrite/expr/mir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/mir_op.rs
@@ -28,6 +28,9 @@ pub enum SubLoc {
     CallArg(usize),
     /// The Nth operand of an rvalue.  `Rvalue -> Operand`
     RvalueOperand(usize),
+    /// The Nth place of an rvalue.  Used for cases like `Rvalue::Ref` that directly refer to a
+    /// `Place`.  `Rvalue -> Place`
+    RvaluePlace(usize),
     /// The place referenced by an operand.  `Operand::Move/Operand::Copy -> Place`
     OperandPlace,
     /// The pointer used in the Nth innermost deref within a place.  `Place -> Place`
@@ -101,8 +104,7 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
         r
     }
 
-    #[allow(dead_code)]
-    fn _enter_dest<F: FnOnce(&mut Self) -> R, R>(&mut self, f: F) -> R {
+    fn enter_dest<F: FnOnce(&mut Self) -> R, R>(&mut self, f: F) -> R {
         self.enter(SubLoc::Dest, f)
     }
 
@@ -116,6 +118,10 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
 
     fn enter_rvalue_operand<F: FnOnce(&mut Self) -> R, R>(&mut self, i: usize, f: F) -> R {
         self.enter(SubLoc::RvalueOperand(i), f)
+    }
+
+    fn enter_rvalue_place<F: FnOnce(&mut Self) -> R, R>(&mut self, i: usize, f: F) -> R {
+        self.enter(SubLoc::RvaluePlace(i), f)
     }
 
     #[allow(dead_code)]
@@ -195,7 +201,7 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
                     _ => {}
                 };
 
-                self.enter_assign_rvalue(|v| v.visit_rvalue(rv, pl_lty));
+                self.enter_assign_rvalue(|v| v.visit_rvalue(rv, Some(pl_lty)));
                 // TODO: visit place
             }
             StatementKind::FakeRead(..) => {}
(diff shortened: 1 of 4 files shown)


c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -166,11 +170,11 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
                             if rv_place.is_indirect()
                                 && self.acx.local_tys[rv_place.local].ty.is_any_ptr()
                             {
-                                let local_addr = self.acx.local_tys[rv_place.local].label;
-                                let perms = self.perms[local_addr];
-                                let flags = self.flags[local_addr];
-                                if let (Ownership::Cell, _) = type_desc::perms_to_desc(perms, flags)
-                                {
+                                let local_lty = self.acx.local_tys[rv_place.local];
```
aneksteind, line 180: i see why you wrapped this up into a variable to pass `local_lty.ty` to `perms_to_desc` but I think this is a step down in clarity for understanding that the `label` is the `local_addr`

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -360,20 +361,22 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
-
-        let arg_expect_own = result_own;
-        // TODO: infer `arg_expect_qty` based on the type of offset this is (positive / unknown)
-        let arg_expect_qty = match result_qty {
-            Quantity::Single => Quantity::Slice,
-            Quantity::Slice => Quantity::Slice,
-            Quantity::OffsetPtr => Quantity::OffsetPtr,
+        let result_desc =
+            type_desc::perms_to_desc(result_ty.ty, self.perms[result_ptr], self.flags[result_ptr]);
+        let result_own = result_desc.own;
```
aneksteind, line 389: i think it's worth destructuring `result_desc` rather than grabbing field by field

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -391,9 +394,16 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
-            type_desc::perms_to_desc(self.perms[result_ptr], self.flags[result_ptr]);
+        let op_desc = type_desc::perms_to_desc(op_lty.ty, self.perms[op_ptr], self.flags[op_ptr]);
+        let op_own = op_desc.own;
+        let op_qty = op_desc.qty;
+        let result_desc = type_desc::perms_to_desc(
+            result_lty.ty,
+            self.perms[result_ptr],
+            self.flags[result_ptr],
+        );
+        let result_own = result_desc.own;
```
aneksteind, line 428: ditto destructuring

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -58,7 +58,8 @@ where
             // TODO: if the `Ownership` and `Quantity` exactly match `lty.ty`, then `ty_desc` can
             // be `None` (no rewriting required).  This might let us avoid inlining a type alias
             // for some pointers where no actual improvement was possible.
-            Some(type_desc::perms_to_desc(perms, flags))
+            let desc = type_desc::perms_to_desc(lty.ty, perms, flags);
+            Some((desc.own, desc.qty))
```
aneksteind, line 62: why not give RewriteLabel a `desc` that is now `TypeDesc`?

c2rust-analyze/src/type_desc.rs
```diff
@@ -52,3 +58,39 @@ pub fn perms_to_desc(perms: PermissionSet, flags: FlagSet) -> (Ownership, Quanti
...
+    }
+}
+
+pub fn local_perms_to_desc<'tcx>(
+    local_ty: Ty<'tcx>,
+    perms: PermissionSet,
+    flags: FlagSet,
+) -> TypeDesc<'tcx> {
+    let (own, qty) = perms_to_own_and_qty(perms, flags);
+    let pointee_ty = local_ty;
```
aneksteind, line 90: this seems unnecessary and in my opinion would be clearer as the argument name changed to `pointee_ty` and function name changed to something like `pointer_perms_to_desc`

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -140,22 +140,26 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
                 let pl_lty = self.acx.type_of(pl);
 
                 if pl.is_indirect() && self.acx.local_tys[pl.local].ty.is_any_ptr() {
-                    let local_addr = self.acx.local_tys[pl.local].label;
-                    let perms = self.perms[local_addr];
-                    let flags = self.flags[local_addr];
-                    if let (Ownership::Cell, _) = type_desc::perms_to_desc(perms, flags) {
-                        // this is an assignment like `x = 2` but `x` has CELL permissions
+                    // FIXME: should use pl_lty instead
```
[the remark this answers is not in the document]
spernsteiner: Updated: c737e42a

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -146,8 +146,11 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
                 let (pl, ref rv) = **x;
                 let pl_lty = self.acx.type_of(pl);
 
+                // FIXME: Needs changes to handle CELL pointers in struct fields.  Suppose `pl` is
+                // something like `*(_1.0)`, where the `.0` field is CELL.  This should be
+                // converted to a `Cell::get` call, but we would fail to enter this case because
+                // `_1` fails the `is_any_ptr()` check.
```
aneksteind, line 152: should the check instead look for what the projection_ty is? would that allow the block to be entered at least, even if the rewrite is incorrect?

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: The previous version sometimes used `local_addr` for the `addr_of_local` `PointerId` and other times used it for the outermost `PointerId` of the local's type (`local_tys[local].label`).  I changed it so that `local_addr` is used exclusively for the former.

In this module, we're pretty consistent about using the name `lty` only for things of type `crate::LTy`, so `some_lty.label` is always a `PointerId`.  I could name it as `local_ptr` or something, though, if you think that would help.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: I changed this and the other case below to just use `result_desc.own` instead of defining a separate `result_own` variable: 7384e347

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: It would need to look at the type of the outermost `Deref`, since that determines what location is actually being written to.  For example, in `*((*_1).0) = 3;`, we should look at the outermost `PointerId` on `(*_1).0`, and if it's `CELL`, rewrite to `(*_1).0.set(3);`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: i think it would help people like me, primarily because i'm not keeping in mind what module i'm in when looking at `LTy` and so `label` is something I'd have to jump to to know what it is. There is at least `RwLty` which makes it more easily guessable and so `.label` variable names don't necessarily need the same massaging to communicate meaning

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: could we create issues from these FIXME's?

aneksteind (approved):
LGTM, some nice-to-have but nothing blocking

c2rust-analyze/src/type_desc.rs
spernsteiner: This function isn't meant to be a general-purpose `TypeDesc` constructor, even if it happens to be implemented like one.  Added a doc comment to clarify: 7a2b68cd

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: OK, added `local_ptr` vars in a3c72909

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Opened #924

spernsteiner:
I rebased this onto `master`, but a test is now failing

spernsteiner:
Fixed the test failure by skipping rewrite generation for `void*` casts.  We can do something more intelligent with those later when we add proper rewriting of `malloc`/`free`/`memcpy`/etc.

PR #920, opened 2023-05-09
analyze: introduce FlagSet::FIXED

spernsteiner:
Adds a new flag, `FIXED`, which indicates that the pointer type must not be changed during rewriting.  This branch only adds the flag and sets it on some `PointerId`s where it's appropriate; nothing consumes this new flag (yet).

For now, we set the `FIXED` flag on all `TyKind::Ref`s in the input program.  This will eventually allow us to avoid rewriting code that's already safe.  However, we don't set the `FIXED` flag for MIR temporaries introduced to hold the results of `&x` or `&mut x` expressions.  This is because c2rust-transpile still uses `&mut x as *mut T` instead of `addr_of_mut!(x)` for the C address-of operator, and we sometimes need to rewrite these expressions in a way that changes their types.  This also applies to cases like `array.as_mut_ptr()`, which implicitly works like `(&mut array).as_mut_ptr()`.

Based on #919 for now; once this is approved, I'll rebase onto `master`.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -136,6 +136,12 @@ bitflags! {
         /// way, and it can't be freely discarded (or its inverse freely added) as is the case for
         /// everything in `PermissionSet`.
         const CELL = 0x0001;
+
+        /// This pointer's type is fixed; rewrites must not change it.  This is used for all safe
+        /// references (which we assume already have the correct types), for raw pointers that
+        /// cross an FFI boundary, and for arguments and return values of functions we can't
+        /// rewrite.
+        const FIXED = 0x0002;
     }
 }
 
@@ -407,6 +413,10 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
         self.gacx.ptr_info.and(&self.ptr_info)
     }
 
+    pub fn local_ptr_info(&self) -> &LocalPointerTable<PointerInfo> {
+        &self.ptr_info
+    }
+
     pub fn type_of<T: TypeOf<'tcx>>(&self, x: T) -> LTy<'tcx> {
         x.type_of(self)
     }
@@ -550,6 +560,10 @@ impl<'tcx> AnalysisCtxtData<'tcx> {
         for lty in rvalue_tys.values_mut() {
             *lty = remap_lty_pointers(lcx, &map, lty);
         }
+    }
+
+    pub fn local_ptr_info(&self) -> &LocalPointerTable<PointerInfo> {
+        &self.ptr_info
     }
 
     pub fn num_pointers(&self) -> usize {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -157,12 +157,17 @@ pub struct LFnSig<'tcx> {
 }
 
 bitflags! {
-    /// Additional flags describing a given pointer type.  These are mainly derived from
-    /// `PermissionSet`, but don't follow the normal subtyping rules and propagation algorithm.
+    /// Basic information about a pointer, computed with minimal analysis before running `dataflow`
+    /// or `borrowck`.
     #[derive(Default)]
     pub struct PointerInfo: u16 {
         /// This `PointerId` was generated for a `TyKind::Ref`.
         const REF = 0x0001;
+
+        /// This `PointerId` was recognized as belonging to a temporary reference.  Given a MIR
+        /// statement like `_1 = &mut _2`, where `_1` is a temporary and is (statically) used only
+        /// once, we apply this flag to the outermost `PointerId` of `_1`.
+        const RECOGNIZED_TEMPORARY_REF = 0x0002;
     }
 }
 
@@ -411,6 +416,10 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
 
     pub fn ptr_info(&self) -> PointerTable<PointerInfo> {
         self.gacx.ptr_info.and(&self.ptr_info)
+    }
+
+    pub fn ptr_info_mut(&mut self) -> PointerTableMut<PointerInfo> {
+        self.gacx.ptr_info.and_mut(&mut self.ptr_info)
     }
 
     pub fn local_ptr_info(&self) -> &LocalPointerTable<PointerInfo> {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -150,9 +150,21 @@ pub struct LFnSig<'tcx> {
     pub output: LTy<'tcx>,
 }
 
+bitflags! {
+    /// Additional flags describing a given pointer type.  These are mainly derived from
+    /// `PermissionSet`, but don't follow the normal subtyping rules and propagation algorithm.
+    #[derive(Default)]
+    pub struct PointerInfo: u16 {
+        /// This `PointerId` was generated for a `TyKind::Ref`.
+        const REF = 0x0001;
+    }
+}
+
 pub struct GlobalAnalysisCtxt<'tcx> {
     pub tcx: TyCtxt<'tcx>,
     pub lcx: LTyCtxt<'tcx>,
+
+    ptr_info: GlobalPointerTable<PointerInfo>,
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
@@ -160,12 +172,12 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub static_tys: HashMap<DefId, LTy<'tcx>>,
     pub addr_of_static: HashMap<DefId, PointerId>,
-
-    next_ptr_id: NextGlobalPointerId,
 }
 
 pub struct AnalysisCtxt<'a, 'tcx> {
     pub gacx: &'a mut GlobalAnalysisCtxt<'tcx>,
+
+    ptr_info: LocalPointerTable<PointerInfo>,
 
     pub local_decls: &'a LocalDecls<'tcx>,
     pub local_tys: IndexVec<Local, LTy<'tcx>>,
@@ -178,8 +190,6 @@ pub struct AnalysisCtxt<'a, 'tcx> {
 
     /// [`Location`]s of (byte-)string literal [`rvalue_tys`](Self::rvalue_tys).
     pub string_literal_locs: Vec<Location>,
-
-    next_ptr_id: NextLocalPointerId,
 }
 
 impl<'a, 'tcx> AnalysisCtxt<'_, 'tcx> {
@@ -211,12 +221,12 @@ impl<'a, 'tcx> AnalysisCtxt<'_, 'tcx> {
 }
 
 pub struct AnalysisCtxtData<'tcx> {
+    ptr_info: LocalPointerTable<PointerInfo>,
     local_tys: IndexVec<Local, LTy<'tcx>>,
     addr_of_local: IndexVec<Local, PointerId>,
     c_void_casts: CVoidCasts<'tcx>,
     rvalue_tys: HashMap<Location, LTy<'tcx>>,
     string_literal_locs: Vec<Location>,
-    next_ptr_id: NextLocalPointerId,
 }
 
 impl<'tcx> GlobalAnalysisCtxt<'tcx> {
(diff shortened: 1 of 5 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -303,14 +321,12 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
         }
-
-        *next_ptr_id = counter;
     }
 
     pub fn assign_pointer_to_static(&mut self, did: DefId) {
         trace!("assign_pointer_to_static({:?})", did);
         let lty = self.assign_pointer_ids(self.tcx.type_of(did));
-        let ptr = self.new_pointer();
+        let ptr = self.new_pointer(PointerInfo::empty());
```
aneksteind, line 340: semantically, what's happening here with `empty`?

c2rust-analyze/src/context.rs
```diff
@@ -577,6 +600,43 @@ where
...
+) -> GlobalPointerTable<PointerInfo> {
+    let mut new_info = GlobalPointerTable::<PointerInfo>::new(num_pointers);
+    for (old, &new) in map.iter() {
+        new_info[new] |= ptr_info[old];
+    }
+    new_info
+}
+
+/// Renumber the keys of `old_local_ptr_info`, producing a new local table.  `new_global_ptr_info`
+/// must already be remapped with `remap_global_ptr_info`.
```
aneksteind, line 642: nit: is it worth wrapping `new_global_ptr_info` in a type so that the invariant becomes type safe? seems easy to mess this up otherwise by calling before `remap_global_ptr_info` is called

c2rust-analyze/src/context.rs
aneksteind: nevermind, it seems that it's a default without a reference annotation

c2rust-analyze/src/main.rs
```diff
@@ -460,6 +460,77 @@ fn label_rvalue_tys<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
...
+                "update_pointer_info: visit assignment: {:?}[{}]: {:?}",
+                bb, i, stmt
+            );
+
+            // Note we ignore `c_void_casts` here.  It shouldn't affect any of the patterns we're
+            // looking for.
+
+            if !pl.is_indirect() {
+                // This is a write directly to `pl.local`.
+                *write_count.entry(pl.local).or_insert(0) += 1;
```
aneksteind, line 484: do we intentionally capture writes to fields here too?

c2rust-analyze/src/main.rs
```diff
@@ -460,6 +460,77 @@ fn label_rvalue_tys<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
     }
 }
 
+fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
```
aneksteind, line 463: docstring?

c2rust-analyze/src/main.rs
```diff
@@ -460,6 +460,77 @@ fn label_rvalue_tys<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
...
+                eprintln!("  record write to LHS {:?}", pl);
+            }
+
+            let ref_pl = match *rv {
+                Rvalue::Ref(_rg, _kind, pl) => Some(pl),
+                Rvalue::AddressOf(_mutbl, pl) => Some(pl),
+                _ => None,
+            };
+            if let Some(ref_pl) = ref_pl {
+                // For simplicity, we consider taking the address of a local to be a write.  We
```
aneksteind, line 494: why is this considered a write? what does that imply downstream?

c2rust-analyze/src/context.rs
```diff
@@ -252,12 +273,18 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         AnalysisCtxt::from_data(self, mir, data)
     }
 
-    pub fn new_pointer(&mut self) -> PointerId {
-        self.next_ptr_id.next()
+    pub fn new_pointer(&mut self, info: PointerInfo) -> PointerId {
+        let i = self.ptr_info.len();
+        self.ptr_info.push(info);
+        PointerId::global(u32::try_from(i).unwrap())
```
kkysen, line 279:
```suggestion
        self.ptr_info.push(info)
```

c2rust-analyze/src/context.rs
kkysen: Do you mean the `` `new_global_ptr_info` must already be remapped with `remap_global_ptr_info` `` invariant?

c2rust-analyze/src/main.rs
```diff
@@ -641,9 +713,24 @@ fn run(tcx: TyCtxt) {
...
+    }
+
     for info in func_info.values_mut() {
         let num_pointers = info.acx_data.num_pointers();
-        let lasn = LocalAssignment::new(num_pointers, PermissionSet::UNIQUE, FlagSet::empty());
+        let mut lasn = LocalAssignment::new(num_pointers, PermissionSet::UNIQUE, FlagSet::empty());
+
+        for (ptr, info) in info.acx_data.local_ptr_info().iter() {
+            if info.contains(PointerInfo::REF)
+                && !info.contains(PointerInfo::RECOGNIZED_TEMPORARY_REF)
```
kkysen, line 728: Could you put this in a `fn PointerInfo::is_fixed` method?  I think it would be clearer right next to the `PointerInfo` definition.

spernsteiner: I think that would be confusing, since just above this is the global case that sets `FIXED` for any `REF`, regardless of whether it was `RECOGNIZED` or not.

kkysen: Oh, true, if the conditions are different for global ptrs.  But why is `FIXED` being set for all `REF`s globally?  Why don't we skip temporaries there, too?

c2rust-analyze/src/context.rs
aneksteind: yes

c2rust-analyze/src/main.rs
spernsteiner: Yes, we count any kind of modification of the `Local`, just to be safe.  "Directly" is in contrast with things like `*_1 = ...;`, which writes indirectly through `_1` but doesn't modify `_1` itself.

The difference between writes to a local and writes to fields of a local doesn't actually matter for the pattern we're interested in, since references have no fields.

c2rust-analyze/src/main.rs
spernsteiner: Again, just being conservative.  What we're looking for here is the code that's generated by `&x`/`&mut x` expressions in the source, which consists of a temporary `_1` that's initialized with a statement `_1 = &mut ...;` and which otherwise isn't written in any way.

We might need to do something slightly smarter here in the future, if we eventually need to handle `&mut &mut x`.

c2rust-analyze/src/context.rs
```diff
@@ -252,12 +262,18 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         AnalysisCtxt::from_data(self, mir, data)
     }
 
-    pub fn new_pointer(&mut self) -> PointerId {
-        self.next_ptr_id.next()
+    pub fn new_pointer(&mut self, info: PointerInfo) -> PointerId {
+        let i = self.ptr_info.len();
```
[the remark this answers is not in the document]
kkysen: Superseded by https://github.com/immunant/c2rust/pull/920#discussion_r1190235780.

c2rust-analyze/src/main.rs
spernsteiner: Actually, after looking at this more, I think there are some adjustments I want to make to the rules for applying `FIXED`.  This will involve unifying the local and global cases, at which point it would make sense to apply a change like you suggested.

c2rust-analyze/src/context.rs
spernsteiner: I think it's not worth it because this function is called in only one place in the entire codebase.  But I'll think about it a little and see if there's an easy refactor to make it more robust.

c2rust-analyze/src/main.rs
spernsteiner: Added: eaba10925978bb9f4b931e08c2a33fb6fc5f0707

kkysen:
> > Would the temporary refs analysis still be necessary if the transpiler only used `addr_of!` and `addr_of_mut!` in these cases?
> 
> I think the analysis wouldn't be needed in that case, but I didn't want to block these changes on that transpiler bug being fixed.

Yeah, that makes sense.  It's good to know, though, that fixing that bug would simplify these kinds of analyses as well as eliminate UB.

spernsteiner:
I've reformulated the rules for applying FIXED: a63fca4a0fa285e443806c02d96e6182e98fdd9c.  This should produce the same results as the previous version, but it should be easier to reason about and more consistent in its handling of complex cases.

Tomorrow I'll add a filecheck test, and then I think it will be ready to merge.

spernsteiner:
Added a test case (and fixed a small bug it uncovered in the new calculation of `NOT_TEMPORARY_REF`)

c2rust-analyze/src/main.rs
aneksteind: I think some or all of your response is worth adding as a comment

c2rust-analyze/src/main.rs
```diff
@@ -456,6 +458,65 @@ fn label_rvalue_tys<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>) {
...
+                }
+
+                rhs_is_ref.insert(pl.local);
+            }
+        }
+    }
+
+    for local in mir.local_decls.indices() {
+        let is_temp_ref = mir.local_kind(local) == LocalKind::Temp
+            && write_count.get(&local).copied().unwrap_or(0) == 1
```
aneksteind, line 509: am i reading this right that you're only looking for cases where an address of a local is taken but not a reborrow?

c2rust-analyze/src/main.rs
```diff
@@ -547,6 +609,7 @@ fn run(tcx: TyCtxt) {
         assert!(acx.local_tys.is_empty());
         acx.local_tys = IndexVec::with_capacity(mir.local_decls.len());
         for (local, decl) in mir.local_decls.iter_enumerated() {
+            // TODO: set PointerInfo::ANNOTATED for the parts of the type with user annotations
```
aneksteind, line 612: can this even be gleaned from MIR? I thought type annotations were only visible at the HIR level

c2rust-analyze/src/main.rs
```diff
@@ -718,8 +809,11 @@ fn run(tcx: TyCtxt) {
     eprintln!("\nfinal labeling for static items:");
     let lcx1 = crate::labeled_ty::LabeledTyCtxt::new(tcx);
     let lcx2 = crate::labeled_ty::LabeledTyCtxt::new(tcx);
-    for (did, lty) in gacx.static_tys.iter() {
-        let name = tcx.item_name(*did);
+    let mut static_dids = gacx.static_tys.keys().cloned().collect::<Vec<_>>();
+    static_dids.sort();
```
aneksteind, line 813: why sorted? add a comment?

c2rust-analyze/src/main.rs
```diff
@@ -732,8 +826,11 @@ fn run(tcx: TyCtxt) {
     }
 
     eprintln!("\nfinal labeling for fields:");
-    for (did, field_lty) in gacx.field_ltys.iter() {
-        let name = tcx.item_name(*did);
+    let mut field_dids = gacx.field_ltys.keys().cloned().collect::<Vec<_>>();
+    field_dids.sort();
```
aneksteind, line 830: ditto unclear why dids are sorted

aneksteind (commented):
wasn't this going to fix rewriting bugs removing `mut` from `&mut A` in `fields.rs` in the function it has? Why is that no longer the case?

c2rust-analyze/src/main.rs
spernsteiner: 1429293d

c2rust-analyze/src/main.rs
spernsteiner: No, we don't care what `Place` the temporary reference points to.  `_1 = &_2`, `_1 = &_2.0`, and `_1 = &*_2` are all treated the same for our purposes.  We only care whether the LHS variable looks like it's a temporary storing the result of a source-level `&...` expression.

c2rust-analyze/src/main.rs
spernsteiner: Determinism.  Hash tables always need to be sorted before iterating if iteration order might affect the output, since iteration order is either random or unreliable in nearly all languages and libraries.

spernsteiner:
> wasn't this going to fix rewriting bugs removing `mut` from `&mut A` in `fields.rs` in the function it has?

This PR only adds code for computing the `FIXED` flag.  The next PR in this series, #923, updates the rewriter to respect `FIXED`, and that one does fix the `&mut A` -> `&A` issue in `fields.rs`.

PR #923, opened 2023-05-10
analyze: preserve refs

spernsteiner:
This branch updates `mir_op` to respect the new `FIXED` flag, thereby stopping most rewriting of already-safe code.

The main effect is that when `emit_cast_*` sees a `FIXED` type on one side of the cast, it tries to obtain a `TypeDesc` by deconstructing the existing `Ty`, rather than by looking at the `perms` and `flags` the analysis came up with.  This ensure we keep the original `Ty` unchanged, even if the `perms` and `flags` would result in a different type.  For example, given a `FIXED` pointer of type `&[T]`, we would produce `(Ownership::Ref, Quantity::Slice, T)`, which accurately reflects the type `&[T]`, regardless of what the `perms` and `flags` say.

Currently this is only used to preserve existing refs, and in that case, the `perms` and `flags` should generally agree with the `FIXED` `Ty`.  But in the future we'll extend this to raw pointers (such as FFI and unanalyzed functions), where `perms` and `flags` will normally suggest some kind of safe rewrite (e.g. `*mut T` -> `&[T]`), but we don't actually want to perform that rewrite.

Depends on #920; once #920 is merged, I'll rebase this onto master

### DIFF
--- a/c2rust-analyze/src/type_desc.rs
+++ b/c2rust-analyze/src/type_desc.rs
@@ -196,49 +196,51 @@ pub fn unpack_pointer_type<'tcx>(
         }
     }
 
+    // Parse the sequence `steps` to extract the ownership and quantity.
     let mut i = 0;
-    let next = |i: &mut usize| -> Option<&Step> {
-        let s = steps.get(*i);
-        *i += 1;
-        s
-    };
-
-    let own = match next(&mut i) {
-        Some(&Step::Ref(Mutability::Not)) => {
-            if let Some(&Step::Cell) = steps.get(i) {
-                i += 1;
-                Ownership::Cell
-            } else {
-                Ownership::Imm
-            }
-        }
-        Some(&Step::Ref(Mutability::Mut)) => Ownership::Mut,
-        Some(&Step::RawPtr(Mutability::Not)) => Ownership::Raw,
-        Some(&Step::RawPtr(Mutability::Mut)) => Ownership::RawMut,
-        Some(&Step::Box) => Ownership::Box,
-        Some(&Step::Rc) => Ownership::Rc,
-        _ => {
-            // Un-consume this step.
-            i -= 1;
-            panic!(
-                "failed to deconstruct {:?} as a pointer to {:?}: \
-                    steps {:?} don't start with a pointer",
-                ty,
-                pointee_ty,
-                &steps[i - 1..]
-            );
+    let mut eat = |s: Step| -> bool {
+        if steps.get(i) == Some(&s) {
+            i += 1;
+            true
+        } else {
+            false
         }
     };
 
-    let qty = match next(&mut i) {
-        Some(&Step::Slice) => Quantity::Slice,
-        Some(&Step::OffsetPtr) => Quantity::OffsetPtr,
-        Some(&Step::Array) => Quantity::Array,
-        _ => {
-            // Un-consume this step.
-            i -= 1;
-            Quantity::Single
+    let own = if eat(Step::Ref(Mutability::Not)) {
+        if eat(Step::Cell) {
+            Ownership::Cell
+        } else {
+            Ownership::Imm
         }
+    } else if eat(Step::Ref(Mutability::Mut)) {
+        Ownership::Mut
+    } else if eat(Step::RawPtr(Mutability::Not)) {
+        Ownership::Raw
+    } else if eat(Step::RawPtr(Mutability::Mut)) {
+        Ownership::RawMut
+    } else if eat(Step::Box) {
+        Ownership::Box
+    } else if eat(Step::Rc) {
+        Ownership::Rc
+    } else {
+        panic!(
+            "failed to deconstruct {:?} as a pointer to {:?}: \
+                steps {:?} don't start with a pointer",
+            ty,
+            pointee_ty,
+            &steps[i..]
+        );
+    };
+
+    let qty = if eat(Step::Slice) {
+        Quantity::Slice
+    } else if eat(Step::OffsetPtr) {
+        Quantity::OffsetPtr
+    } else if eat(Step::Array) {
+        Quantity::Array
+    } else {
+        Quantity::Single
     };
 
     assert!(
--- a/c2rust-analyze/src/type_desc.rs
+++ b/c2rust-analyze/src/type_desc.rs
@@ -1,6 +1,6 @@
 use crate::context::{FlagSet, PermissionSet};
 use rustc_middle::mir::Mutability;
-use rustc_middle::ty::{AdtDef, Ty, TyCtxt, TyKind, TypeAndMut};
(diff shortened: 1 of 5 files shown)


c2rust-analyze/src/type_desc.rs
```diff
@@ -98,3 +99,169 @@ pub fn local_perms_to_desc<'tcx>(
...
+        perms_to_own_and_qty(perms, flags)
+    };
+    TypeDesc {
+        own,
+        qty,
+        pointee_ty,
+    }
+}
+
+/// Unpack an existing `Ty` into its ownership, quantity, and pointee type.
```
aneksteind, line 128: the docs don't seem to match up with the function which just unpacks into ownership and quantity. am i missing something?

spernsteiner: That's outdated - I'll fix it

c2rust-analyze/src/type_desc.rs
```diff
@@ -98,3 +99,169 @@ pub fn local_perms_to_desc<'tcx>(
...
+    match *ty.kind() {
+        TyKind::Adt(adt_def, substs) if is_cell(tcx, adt_def) => Some(substs.type_at(0)),
+        _ => None,
+    }
+}
+
+/// Returns `true` if `adt_def` is the type `std::cell::Cell`.
+fn is_cell<'tcx>(tcx: TyCtxt<'tcx>, adt_def: AdtDef<'tcx>) -> bool {
+    // TODO
+    false
```
aneksteind, line 260: replace these with `todo!()`?

aneksteind (approved):

c2rust-analyze/src/type_desc.rs
```diff
@@ -94,3 +105,169 @@ pub fn local_perms_to_desc<'tcx>(
...
+        }
+    };
+
+    assert!(
+        i == steps.len(),
+        "failed to deconstruct {:?} as a pointer to {:?}: got extra steps {:?}",
+        ty,
+        pointee_ty,
+        &steps[i..],
+    );
```
kkysen, line 244:
```suggestion
    struct Cursor<'a, T> {
        a: &'a [T],
        i: usize,
    }

    impl<'a, T> Cursor<'a, T> {
        pub fn new(a: &'a [T]) -> Self {
            Self { a, i: 0 }
        }

        pub fn next(&mut self) -> Option<&'a T> {
            let e = self.a.get(self.i);
            self.i += 1;
            e
        }

        pub fn prev(&mut self) {
            self.i -= 1;
        }

        pub fn remaining(&self) -> &'a [T] {
            &self.a[self.i..]
        }
    }

    let mut steps = Cursor::new(&steps);

    let own = match steps.next().copied() {
        Some(Step::Ref(Mutability::Not)) => {
            if let Some(Step::Cell) = steps.next().copied() {
                Ownership::Cell
            } else {
                steps.prev();
                Ownership::Imm
            }
        }
        Some(Step::Ref(Mutability::Mut)) => Ownership::Mut,
        Some(Step::RawPtr(Mutability::Not)) => Ownership::Raw,
        Some(Step::RawPtr(Mutability::Mut)) => Ownership::RawMut,
        Some(Step::Box) => Ownership::Box,
        Some(Step::Rc) => Ownership::Rc,
        _ => {
            steps.prev();
            panic!(
                "failed to deconstruct {:?} as a pointer to {:?}: \
                    steps {:?} don't start with a pointer",
                ty,
                pointee_ty,
                steps.remaining(),
            );
        }
    };

    let qty = match steps.next().copied() {
        Some(Step::Slice) => Quantity::Slice,
        Some(Step::OffsetPtr) => Quantity::OffsetPtr,
        Some(Step::Array) => Quantity::Array,
        _ => {
            steps.prev();
            Quantity::Single
        }
    };

    assert!(
        steps.remaining().is_empty(),
        "failed to deconstruct {:?} as a pointer to {:?}: got extra steps {:?}",
        ty,
        pointee_ty,
        steps.remaining(),
    );
```

spernsteiner:
Rebased onto `master`

c2rust-analyze/src/type_desc.rs
aneksteind: i think these diffs are pretty huge and hard to know what changed without scrutiny that surpasses whats needed for a normal pr diff, id recommend trying to make it clearer somehow what the change is

aneksteind:
@spernsteiner i still approve, but would like to see responses to the comments

c2rust-analyze/src/type_desc.rs
kkysen: The diff isn't quite as large as GitHub shows, as it shows unchanged lines in the diff, too.  The main change was to add `Cursor` as an easier to understand abstraction than a free `i` variable with some closures, which was much less clear to me.  This way the usages of `steps.{next, prev, remaining}()` are a lot more self-explanatory and it's clearer how the steps work.

c2rust-analyze/src/type_desc.rs
```diff
@@ -94,3 +105,169 @@ pub fn local_perms_to_desc<'tcx>(
...
+            }
+
+            _ => panic!(
+                "failed to deconstruct {:?} as a pointer to {:?}: unexpected {:?}",
+                ty,
+                pointee_ty,
+                cur_ty.kind()
+            ),
+        }
+    }
```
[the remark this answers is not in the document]
kkysen: This just removes a bunch of the redundant lines in each of the match arms so that it looks simpler and is easier to view all at once.

c2rust-analyze/src/type_desc.rs
spernsteiner: Returning `false` effectively disables the `Cell` (or `Rc` / `OffsetPtr`) case for now.  We can implement the proper checks later when they eventually become necessary.  Having these cases disabled will already cause a panic - calling `unpack_pointer_type` with `ty = &Cell<i32>` and `pointee_ty = i32` will be unable to peel off the `Cell` (because the `Cell` case never fires), and will panic ("failed to deconstruct ...") because it can't reach the `i32` pointee type.  Using `todo!` here would move the panic somewhere else, with a less helpful message.

c2rust-analyze/src/type_desc.rs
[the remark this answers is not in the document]
spernsteiner: Did something similar to this in 18282a161983f4c9108fe2c0ba6ae8d29bbdb8a4

c2rust-analyze/src/type_desc.rs
spernsteiner: This part is essentially a parser, so I made it look more like one, with a single `eat` operation: eb55a007ad1080578f4958ef3d2c425421e8e7bc.  The backward stepping (`i -= 1`) was a hack to avoid repeating `i += 1` in almost every branch.

PR #929, opened 2023-05-15
analyze: add support for some unsupported casts

spernsteiner:
This adds various cases to `emit_cast_desc_desc` to eliminate all the "unsupported cast" warnings in the current tests, and also changes "unsupported cast" to a panic.  In general, "unsupported cast" normally means we're generating code that won't compile (due to type errors), so this is actually a major error for the rewriter.

Currently based on #923; I will rebase onto `master` once that's been merged.

### DIFF
--- a/c2rust-analyze/src/rewrite/expr/mir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/mir_op.rs
@@ -12,6 +12,7 @@ use crate::panic_detail;
 use crate::pointer_id::PointerTable;
 use crate::type_desc::{self, Ownership, Quantity, TypeDesc};
 use crate::util::{ty_callee, Callee};
+use log::*;
 use rustc_ast::Mutability;
 use rustc_middle::mir::{
     BasicBlock, Body, Location, Operand, Place, Rvalue, Statement, StatementKind, Terminator,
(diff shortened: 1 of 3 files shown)


c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -471,27 +472,57 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
             self.acx.tcx().erase_regions(to.pointee_ty),
         );
 
+        let orig_from = from;
+        let mut from = orig_from;
+
+        // Ignore differences in lifetimes.
+        if from.pointee_ty != to.pointee_ty {
```
aneksteind, line 479: check if `erase_regions(from.pointee_ty) == erase_regions(to.pointee_ty)` as well?

aneksteind (approved):

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: That's already asserted above - the only difference that's possible at this point is in the lifetimes.  I've updated the comment here to mention this.

aneksteind: I think I had the mistaken interpretation that `erase_regions` did erase the lifetimes from the type, if this check compares lifetimes but the assertion above does not. Where can I learn about the difference here? From polonius docs I've gotten used to origins, regions, and lifetimes used interchangeably

spernsteiner: Not sure what you mean.  `erase_regions(ty)` returns a copy of `ty` where all regions have been erased, e.g. `&'a i32` -> `&'_ i32` (here using `'_` to mean "erased region" rather than "infer this region").

If this function is called with `&'a i32` and `&'b i32`, the assertion above passes because `&'_ i32 == &'_ i32`.  But the check here returns `true` because `&'a i32 != &'b i32` - specifically, `'a != 'b`.  Our goal in this function is to end up with `from == to` (there's a check to this effect at the end), and here we handle the case of differences in the lifetimes, which are irrelevant.

aneksteind: I think what was confusing me was that this check was happening at all; if the difference in lifetimes doesn't matter, the assertion is made above that the types are equal, and the only way they can differ here is in the lifetimes, why not unconditionally set `from.pointee_ty = to.pointee_ty`?

spernsteiner:
I rebased this onto `master` and ran into a test failure with the newly-added test function from #945.  To fix it, I took several commits from #936 (all related to expanding `mir_op`'s cast generation, so it makes sense to include them here) and added them to this branch.  The commits in question are [here](https://github.com/immunant/c2rust/compare/7b7780a21347838af4ffa7b7ee43a08b456fa9fc...76481866b1c309e2b853f06a75c487c2095ee9ca).

I'm not sure if it's worth re-reviewing, since both this an #945 were already approved.  @aneksteind, I've re-requested review just in case - feel free to just hit approve if you don't think it's necessary.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Ah, good point.  How's this? dc785af4d161e0b3729a42719481850ce9a097f8

kkysen:
The new commits from #936 LGTM.

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: that lgtm

aneksteind:
@spernsteiner consider this re-approval 👍

PR #932, opened 2023-05-17
(`c2rust-analyze`) Fix bleeding of `RUST_LOG` into `RUST_LOG_PANIC`'s logger

### LINKED ISSUE #931: analyze: setting `RUST_LOG=c2rust_analyze::context=trace` causes panic
On current `master` (b34d4e1465f3c1c0cf73811955c1e383aa25c2fd), the following command produces a test failure:
```sh
RUST_LOG=c2rust_analyze::context=trace cargo test statics
```
The failure is due to a panic on a `trace!` invocation:
```
thread 'rustc' panicked at '
[TRACE @ c2rust-analyze/src/context.rs:604 @ c2rust_analyze::context]: assign_pointer_to_static(DefId(0:3 ~ statics[bcaa]::UNUSED))
', c2rust-analyze/src/log.rs:50:17
```
But only `error!` calls should trigger a panic.

Setting `RUST_LOG=trace` does not produce a panic (the test succeeds).

@kkysen, any ideas?

kkysen:
Fixes #931.

Use `env_logger::Builder::from_env` instead of `env_logger::builder().parse_env`, as the latter first reads from the default env (`RUST_LOG`) and thus reads `RUST_LOG` for `RUST_LOG_PANIC`.

### DIFF
--- a/c2rust-analyze/src/log.rs
+++ b/c2rust-analyze/src/log.rs
@@ -33,8 +33,8 @@ pub fn init_logger() {
     let log_env = Env::default().default_filter_or(LevelFilter::Debug.as_str());
     let panic_env = Env::default().filter_or("RUST_LOG_PANIC", LevelFilter::Error.as_str());
 
-    let log_logger = env_logger::builder().parse_env(log_env).build();
-    let panic_logger = env_logger::builder().parse_env(panic_env).build();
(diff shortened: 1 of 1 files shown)


spernsteiner (approved):
Thanks for the quick fix!

PR #933, opened 2023-05-17
(`c2rust-analyze/tests`) Use absolute paths for locating test files so it works from all CWDs

kkysen:
I came across this in debugging #931 and coming up with the fix in #932.  When trying to debug the tests, it would use a different working directory, and not be able to find the correct test file paths.  This just uses the same (correct) logic that I already used in `check_for_missing_tests_for`.


spernsteiner (approved):
Seems reasonable

PR #934, opened 2023-05-18
analyze: new implementation of MIR-to-HIR rewrite lifting

spernsteiner:
This replaces `rewrite::expr::hir_op` with three new modules:

**`rewrite::expr::unlower`:** Builds the "unlowering map", which maps a MIR "precise location" (a `Location` plus a `Vec<SubLoc>` path) to information about the HIR that was lowered to produce that piece of the MIR.  Specifically, it records the `HirId` of the HIR `Expr` that was lowered along with a description of how the MIR relates to the HIR.    For example:

For example:

```Rust
fn f(a: i32) -> i32 {
    a + 1
}

fn g(x: i32, y: i32) -> i32 {
    x + f(y)
}
```

For `f`, the unlowering map annotates the MIR as follows:

```
block bb0:
  bb0[0]: StorageLive(_2)
  bb0[1]: _2 = _1
    []: StoreIntoLocal, `a`
    [Rvalue]: Expr, `a`
  bb0[2]: _0 = Add(move _2, const 1_i32)
    []: StoreIntoLocal, `a + 1`
    [Rvalue]: Expr, `a + 1`
  bb0[3]: StorageDead(_2)
  bb0[4]: Terminator { source_info: ..., kind: return }
```

The statement `_2 = _1` is associated with the expression `a`; the statement as a whole is storing the result of evaluating `a` into a MIR local, and the statement's rvalue `_1` represents the expression `a` itself.  Similarly, `_0 = Add(move _2, const 1)` stores the result of `a + 1` into a local.  If needed, we could extend the `unlower` pass to also record that `move _2` (a.k.a. `bb0[2]` `[Rvalue, RvalueOperand(0)]`) is lowered from the `Expr` `a`.

On `g`, the unlowering map includes the following (among other entries):

```
bb0[5]: Terminator { source_info: ..., kind: _4 = f(move _5) -> [return: bb1, unwind: bb2] }
  []: StoreIntoLocal, `f(y)`
  [Rvalue]: Expr, `f(y)`
  [Rvalue, CallArg(0)]: Expr, `y`
bb1[1]: _0 = Add(move _3, move _4)
  []: StoreIntoLocal, `x + f(y)`
  [Rvalue]: Expr, `x + f(y)`
```

The call terminator `_4 = f(move _5)` computes `f(y)` and stores the result into a local; its rvalue is `f(y)` itself, and the first argument of the rvalue is `y`.

**`rewrite::expr::distribute`:** Given the MIR rewrites produced by `mir_op` and the unlowering map from `unlower`, this module distributes MIR rewrites to HIR nodes (identified by their `HirId`s).  This step also checks for ambiguous case, such as multiple rewrites from different MIR locations being mapped to the same HIR node, where it's unclear in which order to apply the rewrites, and also detects MIR rewrites that couldn't be mapped to any HIR node.

Using the example from above:

```
bb0[5]: Terminator { source_info: ..., kind: _4 = f(move _5) -> [return: bb1, unwind: bb2] }
  []: StoreIntoLocal, `f(y)`
  [Rvalue]: Expr, `f(y)`
  [Rvalue, CallArg(0)]: Expr, `y`
```

A MIR rewrite on `bb0[5]` `[Rvalue, CallArg(0)]` would be attached to the MIR
`Expr` `y`, and a rewrite on `bb0[5]` `[Rvalue`] would be attached to `f(y)`.
A MIR rewrite on `bb0[5]` `[]` (i.e. on the call terminator itself) would
result in an error, since there is no good place in the HIR to attach such a
rewrite.

**`rewrite::expr::convert`:** Converts the `MirRewrite`s for each HIR `Expr` into `Span`-based `rewrite::Rewrite`s.  These are returned as the final result of `expr::gen_expr_rewrites` and are later applied to the input source code using `rewrite::apply`.

---

Overall, this branch simplifies the handling of `SubLoc`s; `distribute` treats `Vec<SubLoc>` paths as mostly opaque (it expects each MIR rewrite's `sub_loc` field to exactly match an entry in the unlowering map), and `convert` doesn't deal with `SubLoc`s at all.  More precise `SubLoc` handling, which will be needed for some tricky pointer-to-pointer cases, goes in the `unlower` module, whose sole purpose is establishing the mapping between `SubLoc`s and HIR.  And  `unlower` makes the mapping between MIR and HIR explicit (unlike `hir_op`, where it was computed implicitly while building rewrites), so it can be inspected for easier debugging.

All test cases produce the same rewrites as before this branch, except for one place in `lighttpd-minimal`, where we now place a `Cell::from_mut` cast in a more appropriate location.


c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
+//! Builds the *unlowering map*, which maps each piece of the MIR to the HIR `Expr` that was
+//! lowered to produce it.
+//!
+//! For example:
+//!
+//! ```Rust
```
kkysen, line 6:
```suggestion
//! ```rust
```

`Rust` isn't recognized by `rustdoc`.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
+//! Builds the *unlowering map*, which maps each piece of the MIR to the HIR `Expr` that was
```
kkysen, line 1: I would move these module docs to `fn unlower` instead, as usually I look to function documentation first, and all of this documentation makes sense for `fn unlower`, too.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+
+    // Build `span_index`, which maps `Span`s to MIR `Locations`.
+    let span_index = build_span_index(mir);
+
+    // Run the visitor.
+    let hir = tcx.hir().body(hir_body_id);
+
+    let mut v = UnlowerVisitor {
```
kkysen, line 332: Could this be renamed `visitor`?  I think of `value` for `v` usually, and it is used for `value` later in this function, too.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+                if k.0 != loc {
+                    break;
+                }
+                let ex = tcx.hir().expect_expr(v.hir_id);
+                eprintln!("      {:?}: {:?}, {:?}", k.1, v.desc, ex.span);
+            }
+        }
+    }
+
+    v.mir_map
```
kkysen, line 378: This seems like a perfect case for a `Display` `impl`.

```suggestion
    let unlowering = UnloweringMap {
        tcx,
        mir,
        map: visitor.mir_map,
    };

    // Print results.
    eprintln!("{}", unlowering);

    unlowering.map
```

along with

```rust
struct UnloweringMap<'tcx, 'a> {
    tcx: TyCtxt<'tcx>,
    mir: &'a Body<'tcx>,
    map: BTreeMap<(Location, Vec<SubLoc>), MirOrigin>,
}

impl Display for UnloweringMap<'_, '_> {
        writeln!(f, "unlowering for {:?}:", self.mir.source)?;
        for (bb_id, bb) in self.mir.basic_blocks().iter_enumerated() {
            writeln!(f, "  block {:?}:", bb_id)?;
            for (i, stmt) in bb.statements.iter().enumerate() {
                let loc = Location {
                    block: bb_id,
                    statement_index: i,
                };

                writeln!(f, "    {:?}: {:?}", loc, stmt)?;
                for (k, v) in self.map.range(&(loc, vec![])..) {
                    if k.0 != loc {
                        break;
                    }
                    let ex = self.tcx.hir().expect_expr(v.hir_id);
                    writeln!(f, "      {:?}: {:?}, {:?}", k.1, v.desc, ex.span)?;
                }
            }

            {
                let term = bb.terminator();
                let loc = Location {
                    block: bb_id,
                    statement_index: bb.statements.len(),
                };

                writeln!(f, "    {:?}: {:?}", loc, term)?;
                for (k, v) in self.map.range(&(loc, vec![])..) {
                    if k.0 != loc {
                        break;
                    }
                    let ex = self.tcx.hir().expect_expr(v.hir_id);
                    writeln!(f, "      {:?}: {:?}, {:?}", k.1, v.desc, ex.span)?;
                }
            }
        }
        Ok(())
    }
}
```

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+        self.visit_expr_inner(ex);
+        intravisit::walk_expr(self, ex);
+    }
+}
+
+pub fn unlower<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    mir: &Body<'tcx>,
+    hir_body_id: hir::BodyId,
+) -> BTreeMap<(Location, Vec<SubLoc>), MirOrigin> {
```
kkysen, line 317: Is a `BTreeMap` map used here to sort by `(Location, Vec<SubLoc>)`?  If so, could you add that to the docs?

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen, line 332:
```suggestion
    let hir = tcx.hir().body(hir_body_id);

    // Run the visitor.
    let mut visitor = UnlowerVisitor {
```

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+    // If this MIR body came from a `#[derive]`, ignore it.
+    let def_id = mir.source.def_id();
+    if let Some(parent_def_id) = tcx.opt_parent(def_id) {
+        if tcx.has_attr(parent_def_id, sym::automatically_derived) {
+            return BTreeMap::new();
+        }
+    }
```
kkysen, line 324: nit: I find these long functions like `unlower` much easier to follow and understand when the non-core, high-level logic of it is moved into its own functions.

```suggestion
    if is_automatically_derived(mir, tcx) {
        return BTreeMap::new();
    }
```

along with

```rust
/// Check if a [`Body`] is from an `impl` marked `#[automatically_derived]`,
/// which indicates it came from a `#[derive(...)]`.
fn is_automatically_derived<'tcx>(body: &Body<'tcx>, tcx: TyCtxt<'tcx>) -> bool {
    tcx.opt_parent(body.source.def_id())
        .map(|parent_def_id| tcx.has_attr(parent_def_id, sym::automatically_derived))
        .unwrap_or(false)
}
```

which could also go in `util`.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+//!   bb0[0]: StorageLive(_2)
+//!   bb0[1]: _2 = _1
+//!     []: StoreIntoLocal, `a`
+//!     [Rvalue]: Expr, `a`
+//!   bb0[2]: _0 = Add(move _2, const 1_i32)
+//!     []: StoreIntoLocal, `a + 1`
+//!     [Rvalue]: Expr, `a + 1`
+//!   bb0[3]: StorageDead(_2)
+//!   bb0[4]: Terminator { source_info: ..., kind: return }
+//! ```
```
kkysen, line 29: The format used here isn't very clear at all.  If you could add a bit of an explainer, that'd be great and make it a lot easier to understand, or at least intra-doc link to the code that prints this format.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: I also think using `{var:?}`-style format strings would make it easier to match the code here with the docs using the output of this, which would then make it easier to understand which things are meant to be which in the docs.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+            span: ex.span,
+            desc,
+        };
+        match self.mir_map.entry((loc, sub_loc.to_owned())) {
+            Entry::Vacant(e) => {
+                e.insert(origin);
+            }
+            Entry::Occupied(e) => {
+                if *e.get() != origin {
+                    let old_origin = e.get().clone();
```
kkysen, line 121:
```suggestion
                    let old_origin = *e.get();
```

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    fn record(&mut self, loc: Location, sub_loc: &[SubLoc], ex: &hir::Expr) {
+        self.record_desc(loc, sub_loc, ex, MirOriginDesc::Expr);
+    }
+
+    fn record_desc(
+        &mut self,
+        loc: Location,
+        sub_loc: &[SubLoc],
+        ex: &hir::Expr,
```
kkysen, line 107: Is `ex` an abbreviation for `expr` commonly used in rustc?  Otherwise, it's pretty confusing, so I'd prefer `expr`, which we normally use elsewhere.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// This MIR stores the result of the HIR expression into a MIR local of some kind.
+    StoreIntoLocal,
+}
+
+struct UnlowerVisitor<'a, 'tcx> {
+    tcx: TyCtxt<'tcx>,
+    mir: &'a Body<'tcx>,
+    span_index: SpanIndex<Location>,
+    /// Maps MIR (sub)locations to the HIR node that produced each one, if known.
+    mir_map: BTreeMap<(Location, Vec<SubLoc>), MirOrigin>,
```
kkysen, line 89: `unlower_map`, `mir_to_hir_map`, or `origin_map` would make more sense than `mir_map` I think, which doesn't say much.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+            return None;
+        }
+        self.get_last_assign(locs)
+    }
+
+    fn get_last_assign(
+        &self,
+        locs: &[Location],
+    ) -> Option<(Location, mir::Place<'tcx>, &'a mir::Rvalue<'tcx>)> {
+        let &loc = locs.last()?;
```
kkysen, line 153:
```suggestion
        let loc = *locs.last()?;
```

The `&` destructure in a normal `let` is rarely used I think, so `*` would IMO be a lot clearer.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+    fn get_last_call(
+        &self,
+        locs: &[Location],
+    ) -> Option<(
+        Location,
+        mir::Place<'tcx>,
+        &'a mir::Operand<'tcx>,
+        &'a [mir::Operand<'tcx>],
+    )> {
+        let &loc = locs.last()?;
```
kkysen, line 170:
```suggestion
        let loc = *locs.last()?;
```

Same here.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen, line 121: nit:

```suggestion
                let old_origin = *e.get();
                if old_origin != origin {
```

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+    fn visit_expr_inner(&mut self, ex: &'tcx hir::Expr<'tcx>) {
+        let _g = panic_detail::set_current_span(ex.span);
+
+        let mut locs = Vec::with_capacity(1);
+        for &loc in self.span_index.lookup_exact(ex.span) {
+            if !self.should_ignore_statement(loc) {
+                locs.push(loc);
+            }
+        }
+        if locs.len() == 0 {
```
kkysen, line 209: nit:

```suggestion
        if locs.is_empty() {
```

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+            }
+
+            hir::ExprKind::Call(_, args) | hir::ExprKind::MethodCall(_, args, _) => {
+                let (loc, _mir_pl, _mir_func, mir_args) = match self.get_last_call(&locs) {
+                    Some((l, pl, f, a)) if self.is_var(pl) => (l, pl, f, a),
+                    _ => {
+                        warn(self, "expected final Call to store into var");
+                        return;
+                    }
+                };
```
kkysen, line 249:
```suggestion
                let (loc, _mir_pl, _mir_func, mir_args) = match self
                    .get_last_call(&locs)
                    .filter(|(_, pl, _, _)| self.is_var(*pl))
                {
                    Some(x) => x,
                    _ => {
                        warn(self, "expected final Call to store into var");
                        return;
                    }
                };
```

This way we don't need the confusing short variable names like `l`, `f`, `a`.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+                if locs.len() > 1 {
+                    warn!("NYI: extra locations {:?} in Call", &locs[..locs.len() - 1]);
+                }
+            }
+
+            _ => {
+                // For all other `ExprKind`s, we expect the last `loc` to be an assignment storing
+                // the final result into a temporary.
+                let (loc, _mir_pl, mir_rv) = match self.get_last_assign(&locs) {
+                    Some((l, pl, r)) if self.is_var(pl) => (l, pl, r),
```
kkysen, line 266:
```suggestion
                let (loc, _mir_pl, mir_rv) = match self
                    .get_last_assign(&locs)
                    .filter(|(_, pl, _)| self.is_var(*pl))
                {
                    Some(x) => x,
```

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+        for &loc in self.span_index.lookup_exact(ex.span) {
+            if !self.should_ignore_statement(loc) {
+                locs.push(loc);
+            }
+        }
+        if locs.len() == 0 {
+            return;
+        }
+
+        let warn = |slf: &Self, desc| {
```
kkysen, line 213: I don't think we need the `Self` arg here.  I tried removing it and everything still works, so it's not a borrowck thing.

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    fn is_var(&self, pl: mir::Place<'tcx>) -> bool {
+        pl.projection.len() == 0
+    }
+
+    fn visit_expr_inner(&mut self, ex: &'tcx hir::Expr<'tcx>) {
+        let _g = panic_detail::set_current_span(ex.span);
+
+        let mut locs = Vec::with_capacity(1);
```
kkysen, line 203: This is `1` because we expect, in the most common case, a 1-to-1 mapping from MIR to HIR nodes, right?

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+                | mir::StatementKind::StorageLive(..)
+                | mir::StatementKind::StorageDead(..)
+                | mir::StatementKind::Nop => return true,
+                _ => {}
+            }
+        }
+        false
+    }
+
+    fn is_var(&self, pl: mir::Place<'tcx>) -> bool {
```
kkysen, line 196: Does this need a `&self` arg?

c2rust-analyze/src/rewrite/expr/mod.rs
```diff
@@ -14,6 +16,8 @@ pub fn gen_expr_rewrites<'tcx>(
     hir_body_id: BodyId,
 ) -> Vec<(Span, Rewrite)> {
     let mir_rewrites = mir_op::gen_mir_rewrites(acx, asn, mir);
-    let hir_rewrites = hir_op::gen_hir_rewrites(acx.tcx(), mir, hir_body_id, &mir_rewrites);
+    let unlower_map = unlower::unlower(acx.tcx(), mir, hir_body_id);
+    let rewrites_by_expr = distribute::distribute(acx.tcx(), unlower_map, mir_rewrites);
+    let hir_rewrites = convert::convert_rewrites(acx.tcx(), hir_body_id, rewrites_by_expr);
```
kkysen, line 21: nit: Can the functions just be imported and called directly?

c2rust-analyze/src/rewrite/expr/mod.rs
```diff
@@ -4,8 +4,10 @@ use rustc_hir::BodyId;
 use rustc_middle::mir::Body;
 use rustc_span::Span;
 
-mod hir_op;
```
kkysen, line 7: A bunch of other documentation (like the `mod rewrite` top-level docs) reference `hir_op` a lot, so could those be updated to reference the new `convert`, `distribute`, `unlower` `mod`s?

kkysen (commented):
Could d875e3128c6e680e692246f87d7f79bd92fd31b9 be split into a separate PR?  It seems mostly separate and simpler than the rest of the PR.

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: Done (also did the same in the `distribute` and `convert` modules as well)

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: Done

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I've slightly expanded the paragraph below in 85959eeaf802e208747da14b34fe5007a98e6018 - let me know if that helps.

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: >it'd be more easily useable from any other potential contexts.

I don't think we're likely to want this exact debug output in any other contexts.

>using `{var:?}`-style format strings would

Done in ddf3da625fa2f883a35b0766f48767ef32d21c10

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: Done (though currently we only rely on this feature for the debug prints in this function)

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I applied the other suggestion instead, which supersedes this

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I just changed this to `Vec::new()` since it's not clear whether this is a win.  Using `with_capacity` forces an allocation even when there are no locations, which happens about 40% of the time according to some quick profiling on our test suite.

c2rust-analyze/src/rewrite/expr/mod.rs
spernsteiner: I generally prefer to import the module and call `module::function()` rather than import the function directly.  This makes it clear when the callee is in a different module.

c2rust-analyze/src/rewrite/expr/mod.rs
spernsteiner: I updated the `rewrite/mod.rs` doc comment, which is the only place I saw a mention of `hir_op`.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: That works well, too!  I forgot about the `@` trick.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: That's probably better.  In that case, https://github.com/immunant/c2rust/pull/934#discussion_r1210103196 could also use a `.collect::<Vec<_>>()` to simplify things even further.

c2rust-analyze/src/rewrite/expr/mod.rs
kkysen: I only actually saw it referenced there, so this should be perfect.

kkysen: Actually, could you also update 9d2510327274776803bacfb0248870cf67c574eb to use intra-doc links (e.x. `` [`item`] `` style)?  That way such dangling references would be caught automatically.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: > I don't think we're likely to want this exact debug output in any other contexts.

It could be helpful in debugging.  If you feel strongly that it shouldn't be a `Display` `impl`, though, I'd still prefer for it to at least be separated into its own function, both to divide the cognitive complexity of `fn unlower` and to intra-doc link to it from where it's output is shown in the docs.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: That's a lot clearer, thanks!  Especially the part about prefixes and looking up by `Location` only for any `SubLoc`s.

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I've factored it out, though IMO this makes little difference for readability (anyone uninterested in the printing of results could already skip over the section clearly labeled "print results")

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I've removed it.  (In a previous version, this code used an `is_temp_var` method, which does need access to `self.mir`.)

c2rust-analyze/src/rewrite/expr/mod.rs
spernsteiner: Rustdoc doesn't seem to like those links, presumably because the targets are private - it reports "no item named `convert` in module `expr`" etc.

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: `ex` is the name used in [the HIR `Visitor` trait](https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/intravisit/trait.Visitor.html#method.visit_expr), and generally seems pretty unambiguous

c2rust-analyze/src/rewrite/expr/mod.rs
kkysen: Hmm, that's annoying, oh well.

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: Renamed to `unlower_map`

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: Perhaps it's a MIR vs. HIR distinction?  Fairly odd abbreviation but if it's already used widely it's fine.

c2rust-analyze/src/rewrite/expr/unlower.rs
kkysen: Yeah, that's a lot clearer now!

By the way, I generally try to have line breaks at semantic breaks in the sentence vs. hard wrapping at a certain length.  That way diffs are much cleaner and don't spill over to other unchanged lines.  Also, it's IMO easier to read.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -368,6 +186,11 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirRewriteVisitor<'a, 'tcx> {
             }
         };
 
+        for mir_rw in mir_rws {
+            hir_rw = rewrite_from_mir_rws(&mir_rw, hir_rw);
+        }
+
+        /*
```
kkysen, line 193: Is this going to be used later?  Is there a reason it's just commented out vs. removed?

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: >I generally try to have line breaks at semantic breaks in the sentence vs. hard wrapping at a certain length. That way diffs are much cleaner and don't spill over to other unchanged lines.

I used to do the same, but it turns out `git diff --word-diff` does a pretty good job of unwrapping the diff:

```
/// {+Each statement is shown here with its MIR `Location`, such as `bb0[0]`; within each statement,+}
{+/// annotations are associated with particular `SubLoc` paths, such as `[]` or `[Rvalue]`.+}  The
{+///+} statement `_2 = _1` is associated with the expression `a`; the statement[-///-] as a whole {+(`bb0[1]`+}
{+/// `[]`)+} is storing the result of evaluating `a` into a MIR local, and the[-///-] statement's rvalue
{+/// (`bb0[1]` `[Rvalue]`)+} `_1` represents the expression `a` itself.  Similarly, `_0 =[-///-] Add(move _2,
{+///+} const 1)` stores the result of `a + 1` into a local.  If needed,[-///-] we could extend the `unlower`
{+///+} pass to also record that `move _2` (a.k.a. `bb0[2]`[-///-] `[Rvalue, RvalueOperand(0)]`) is lowered
{+///+} from the `Expr` `a`.
```

(Looks nicer in color - try `git show 85959ee --word-diff`)

kkysen: I use that, too, but I don't know how to get GitHub to use it (that'd be amazing!).

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+    let mut info_map = HashMap::<HirId, Vec<RewriteInfo>>::new();
+
+    for (loc, mir_rws) in mir_rewrites {
+        for mir_rw in mir_rws {
+            let mut key = (loc, mir_rw.sub_loc);
+
+            let mut origin = unlower_map.get(&key);
+            if origin.is_none() && matches!(key.1.last(), Some(&SubLoc::RvalueOperand(0))) {
+                // Hack: try without the `RvalueOperand(0)` at the end.
+                // TODO: remove SubLoc::RvalueOperand from mir_op
```
kkysen, line 48: Could you explain a bit more why this hack is needed?

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    fn visit_expr_inner(&mut self, ex: &'tcx hir::Expr<'tcx>) {
+        let _g = panic_detail::set_current_span(ex.span);
+
+        let mut locs = Vec::with_capacity(1);
+        for &loc in self.span_index.lookup_exact(ex.span) {
+            if !self.should_ignore_statement(loc) {
+                locs.push(loc);
+            }
+        }
```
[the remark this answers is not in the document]
spernsteiner: I don't think this is any clearer than the original.  This foreach/if/push pattern is extremely common - I think most developers have seen code like this hundreds of times in dozens of different languages and would recognize the pattern instantly.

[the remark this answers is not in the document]
kkysen: Since we're using `Vec::new()` now, this can be written like this instead:

```suggestion
        let locs = self
            .span_index
            .lookup_exact(ex.span)
            .copied()
            .filter(|&loc| !self.should_ignore_statement(loc))
            .collect::<Vec<_>>();
```

From what I've seen, that's a lot more common and idiomatic in Rust.  `locs` doesn't have to be `mut` anymore, and `.collect()` can take advantage of the `.size_hint()` that `.push()` can't.

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+///   [Rvalue, CallArg(0)]: Expr, `y`
+/// ```
+///
+/// A MIR rewrite on `bb0[5]` `[Rvalue, CallArg(0)]` would be attached to the MIR
+/// `Expr` `y`, and a rewrite on `bb0[5]` `[Rvalue]` would be attached to `f(y)`.
+/// A MIR rewrite on `bb0[5]` `[]` (i.e. on the call terminator itself) would
+/// result in an error, since there is no good place in the HIR to attach such a
+/// rewrite.
+pub fn distribute<'tcx>(
+    tcx: TyCtxt<'tcx>,
```
kkysen, line 35:
```suggestion
pub fn distribute(
    tcx: TyCtxt,
```

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+    unlower_map: BTreeMap<(Location, Vec<SubLoc>), MirOrigin>,
+    mir_rewrites: HashMap<Location, Vec<MirRewrite>>,
+) -> HashMap<HirId, Vec<mir_op::RewriteKind>> {
+    let mut info_map = HashMap::<HirId, Vec<RewriteInfo>>::new();
+
+    for (loc, mir_rws) in mir_rewrites {
+        for mir_rw in mir_rws {
+            let mut key = (loc, mir_rw.sub_loc);
```
kkysen, line 43: I think adding a

```rust
struct PreciseLoc {
    loc: Location,
    sub_loc: Vec<SubLoc>,
}
```

would make this a lot more readable (`key.1` isn't very) (and I think we may also be able to `impl Borrow<Q> for PreciseLoc` to avoid `sub_loc` clones, though I'm not sure that's worth the complexity).

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+            }
+            let mut origin = match origin {
+                Some(x) => x,
+                None => {
+                    error!("unlower_map has no origin for {:?}", key);
+                    continue;
+                }
+            };
+
+            if origin.desc == MirOriginDesc::StoreIntoLocal && key.1.len() == 0 {
```
kkysen, line 63:
```suggestion
            if origin.desc == MirOriginDesc::StoreIntoLocal && key.1.is_empty() {
```

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+                // Hack: try with an extra `Rvalue` in the key.
```
kkysen, line 64: Same (https://github.com/immunant/c2rust/pull/934/files#r1210910617) for this hack.

c2rust-analyze/src/rewrite/span_index.rs
```diff
@@ -18,14 +19,26 @@ impl<T> SpanIndex<T> {
...
+
+    /// Iterate over items whose spans are exactly equal to `span`.  The order of the returned
+    /// items is unspecified.
+    pub fn lookup_exact(&self, span: Span) -> ExactIter<T> {
+        let data = span.data();
+        let start = self.v.partition_point(|(span, _)| span.lo() < data.lo);
+        ExactIter {
+            inner: self.v[start..].iter(),
+            lo: data.lo,
+            span: span,
```
kkysen, line 39: nit:

```suggestion
            span,
```

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    // If a single `HirId` has rewrites from multiple different pieces of MIR, it's ambiguous how
+    // to order those rewrites.  (`mir_rewrites` only establishes an ordering between rewrites on
+    // the same `Location`.)  For now, we complain if we see this ambiguity; in the future, we may
+    // need to add rules to resolve it in a particular way, such as prioritizing one `SubLoc` or
+    // `MirOriginDesc` over another.
+    for (&hir_id, infos) in &info_map {
+        let all_same_loc = infos.iter().skip(1).all(|i| {
+            i.loc == infos[0].loc && i.sub_loc == infos[0].sub_loc && i.desc == infos[0].desc
+        });
```
kkysen, line 106: If we add `itertools` as a dependency, which we already have in most other `c2rust-*` crates, we can use their `.all_equal()` method.  And I think the `.map(|i| (i.loc, &i.sub_loc, i.desc))` more clearly shows which fields are being checked.

```suggestion
        let all_same_loc = infos
            .iter()
            .map(|i| (i.loc, &i.sub_loc, i.desc))
            .all_equal();
```

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            info_map
+                .entry(origin.hir_id)
+                .or_default()
+                .push(RewriteInfo {
+                    rw: mir_rw.kind,
+                    loc,
+                    sub_loc: key.1,
+                    desc: origin.desc,
+                });
```
kkysen, line 94: I think this might be clearer if written as grouping (i.e. `.into_group_map()` from `itertools`).

c2rust-analyze/src/rewrite/span_index.rs
```diff
@@ -18,14 +19,26 @@ impl<T> SpanIndex<T> {
...
+        }
+    }
```
kkysen, line 41: If we return an `impl Iterator` and use iterator methods, we don't need `ExactIter`, and I think the iterator chaining way is a lot easier to understand.

```suggestion
    pub fn lookup_exact(&self, span: Span) -> impl Iterator<Item = &T> {
        let data = span.data();
        let start = self.v.partition_point(|(span, _)| span.lo() < data.lo);
        self.v[start..]
            .iter()
            // The list is sorted by `lo`, so once we move past `self.lo`, we won't see it again.
            .take_while(move |(span, _)| span.data().lo == data.lo)
            // This is not the requested span.  Note we compare the entire span, not just `lo`
            // and `hi`, because spans with the same `lo` and `hi` can come from two different
            // macro expansions (thus having different `SyntaxContext`s).
            .filter(move |(span_, _)| *span_ == span)
            .map(|(_, t)| t)
    }
```

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -477,55 +301,34 @@ fn materialize_adjustments<'tcx>(
     }
 }
 
-pub fn gen_hir_rewrites<'tcx>(
+/// Convert the MIR rewrites attached to each HIR node into `Span`-based `rewrite::Rewrite`s.
+pub fn convert_rewrites<'tcx>(
     tcx: TyCtxt<'tcx>,
```
kkysen, line 306: nit:

```suggestion
pub fn convert_rewrites(
    tcx: TyCtxt,
```

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -477,55 +301,34 @@ fn materialize_adjustments<'tcx>(
...
     intravisit::Visitor::visit_body(&mut v, hir);
 
-    // Check that all locations with rewrites were visited at least once.
-    let mut locs = rewrites.keys().cloned().collect::<Vec<_>>();
-    locs.sort();
-    let mut found_unvisited_loc = false;
-    for loc in locs {
-        if v.locations_visited.contains(&loc) {
-            continue;
+    if v.mir_rewrites.len() > 0 {
```
kkysen, line 323: nit:

```suggestion
    if !v.mir_rewrites.is_empty() {
```

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -477,55 +301,34 @@ fn materialize_adjustments<'tcx>(
...
-        mir,
-        span_index,
-        rewrites,
-        locations_visited: HashSet::new(),
-        hir_rewrites: Vec::new(),
+        mir_rewrites,
+        rewrites: Vec::new(),
         materialize_adjustments: false,
     };
     intravisit::Visitor::visit_body(&mut v, hir);
```
kkysen, line 321: Why not just

```suggestion
    v.visit_body(hir);
```

?  Is there a reason?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -281,25 +114,10 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirRewriteVisitor<'a, 'tcx> {
...
-            })
-            .collect();
-
-        let all_rws_unflattened = all_rws_unflattened
-            .iter()
-            .map(|v| v.as_slice())
-            .collect::<Vec<_>>();
+        let mir_rws = self
+            .mir_rewrites
+            .remove(&ex.hir_id)
```
kkysen, line 119: Previously (on `self.rewrites`), this was `.get` instead of `.remove`.  If I'm understanding things correctly, that could've been `.remove`, too, right, since `visit_expr` is never called on the same `Expr` in the same visitor pass?  If that's the case, then this is all good.

kkysen (commented):
By the way, is there a reason the base branches for this and other PRs is different (in name) but equal (in commit) from the PR branch of the other PRs?  It makes it confusing which PR is stacked on top of which PR.

spernsteiner:
> By the way, is there a reason the base branches for this and other PRs is different (in name) but equal (in commit) from the PR branch of the other PRs?

Keeps github from getting confused about what to show in the diff when I do more work (add commits and/or rebase) on the various branches or merge PRs out of order.  I forget what exactly triggered this, but I definitely ran into some trouble along these lines with my previous batch of PRs.

c2rust-analyze/src/rewrite/expr/unlower.rs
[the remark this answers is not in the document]
spernsteiner: That looks absolutely horrible (`rustfmt` tends to make a mess of these kinds of iterator chains), but you're right that `collect` is the thing to use here.  I didn't check the context and assumed there were more `push`es elsewhere in the function.

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: Added a `struct PreciseLoc`.  `impl Borrow` I don't think is possible because it needs to return `&Q` where `Q` is something like `(Location, &[SubLoc])`, but `PreciseLoc` can't own such a thing.

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Removed.  I had that commented out initially because I thought I might need to port that logic over to the new design, and forgot to delete it afterward.

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: I did some `mir_op` cleanup that I initially put off and then removed the hacks from `distribute`.

c2rust-analyze/src/rewrite/expr/convert.rs
kkysen: Okay, the `.remove` here seems correct to me then.

c2rust-analyze/src/rewrite/expr/distribute.rs
kkysen: Great, that looks much cleaner!

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: This was copied from the previous version.  I changed the code (in both `unlower` and `convert`) to import `intravisit::Visitor` and call `v.visit_body(hir)` and there weren't any problems.

c2rust-analyze/src/rewrite/span_index.rs
aneksteind: in either of these implementations why is it necessary to partition and cut off iteration? it seems like filtering by equality with span would optimize for clarity at the (maybe minor?) expense of iterating past the span's lo byte

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -0,0 +1,107 @@
...
+/// ```text
+/// bb0[5]: Terminator { source_info: ..., kind: _4 = f(move _5) -> [return: bb1, unwind: bb2] }
+///   []: StoreIntoLocal, `f(y)`
+///   [Rvalue]: Expr, `f(y)`
+///   [Rvalue, CallArg(0)]: Expr, `y`
+/// ```
+///
+/// A MIR rewrite on `bb0[5]` `[Rvalue, CallArg(0)]` would be attached to the MIR
+/// `Expr` `y`, and a rewrite on `bb0[5]` `[Rvalue]` would be attached to `f(y)`.
+/// A MIR rewrite on `bb0[5]` `[]` (i.e. on the call terminator itself) would
```
aneksteind, line 30: Is this the case with all terminators, and only terminators? For a statement that's not a terminator like `*x = 2` where `x` has `CELL` permissions I would expect an `MIR` rewrite for the entire `Location` with the `SubLocs` equal to `[]`

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,400 @@
...
+
+            hir::ExprKind::Call(_, args) | hir::ExprKind::MethodCall(_, args, _) => {
+                let (loc, _mir_pl, _mir_func, mir_args) = match self.get_last_call(&locs) {
+                    Some(x @ (_, pl, _, _)) if is_var(pl) => x,
+                    _ => {
+                        warn("expected final Call to store into var");
+                        return;
+                    }
+                };
+                self.record_desc(loc, &[], ex, MirOriginDesc::StoreIntoLocal);
```
aneksteind, line 203: why is there a distinction between `StoreIntoLocal` and `Expr` with a `Dest`, i.e. why not `self.record(loc, &[SubLoc::Dest], mir_pl);`?

aneksteind (approved):
Left some questions for my own understanding and posterity. Regardless of the response, with those answered this LGTM

c2rust-analyze/src/rewrite/span_index.rs
spernsteiner: I think the expense is actually significant here.  Using `filter` instead of breaking early means visiting roughly every span in the function on every lookup, making the overall behavior O(N^2).  Since we know the list is sorted, we can do better by looking only at the range that could plausibly contain matching entries.

aneksteind: That sounds fine but, just for my own understanding, isn't the behavior O(N^2) anyway because of the edge case where all spans match the criteria?

spernsteiner: Yes, in the worst case it's still O(N^2), but in most cases we expect to see, it should be closer to O(N).

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: Ideally there would be only one `MirOriginDesc::Expr` entry for each HIR expr, indicating the piece of MIR that's primarily responsible for implementing the evaluation of that expr.  For `x + y` this should be an `Rvalue::BinaryOp`, for `f(x)` it's a `TerminatorKind::Call`, etc.  For a store into a local like `_z = _x + _y`, it would be confusing to label both the `Place` (`SubLoc::Dest`) and the `Rvalue` as being "the expr itself".

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: We generally treat statements and terminators the same in `unlower` and `distribute`.  For `CellSet` in particular, it looks like `mir_op` does emit the rewrite on the statement itself (empty `Vec<SubLoc>`).

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -0,0 +1,329 @@
...
+                    _ => {
+                        warn(self, "expected final Call to store into var");
+                        return;
+                    }
+                };
+                self.record_desc(loc, &[], ex, MirOriginDesc::StoreIntoLocal);
+                self.record(loc, &[SubLoc::Rvalue], ex);
+                for (i, (arg, mir_arg)) in args.iter().zip(mir_args).enumerate() {
+                    self.record(loc, &[SubLoc::Rvalue, SubLoc::CallArg(i)], arg);
+                    // TODO: distribute extra `locs` among the various args
```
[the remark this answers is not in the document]
spernsteiner: Expanded on this comment in e7063dfc - let me know if this helps

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: It looks like this would require restructuring the whole loop into a big iterator chain, which IMO is generally less readable than a standard for loop.

c2rust-analyze/src/rewrite/span_index.rs
spernsteiner: I switched to iterator chains here and in `fn lookup()`.

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: Done.

spernsteiner:
Rebased and addressed all comments

kkysen (approved):

c2rust-analyze/src/rewrite/expr/distribute.rs
aneksteind: That's better

c2rust-analyze/src/rewrite/expr/unlower.rs
aneksteind: I think I understand. Plus, I'm not sure there is an `hir::Expr` that represents the destination of the method call anyway, right?

spernsteiner: Right - the LHS in the MIR is a temporary, so there is no corresponding variable in the HIR code.

spernsteiner:
@kkysen https://github.com/immunant/c2rust/pull/839#issuecomment-1584110904

> Actually, there seem to be issues with https://github.com/immunant/c2rust/pull/934 running on the string_casts.rs tests

Did you test `string_casts.rs` on #934 alone, or on a merge of #934 and #839?  I've rebased this branch (#934) onto `master` (which now includes #839) and I haven't seen any test failures.

kkysen:
> Did you test `string_casts.rs` on #934 alone, or on a merge of #934 and #839? I've rebased this branch (#934) onto `master` (which now includes #839) and I haven't seen any test failures.

I merged #934 (this one) into #839 and tested it.  If rebase #934 on `master`, which now includes #839, works, then that's great and there's no issue at all.

PR #936, opened 2023-05-23
analyze: generate casts around call arguments and results

spernsteiner:
This branch updates the rewriter to add casts around function calls and their arguments.  We mainly need this so we can add casts to and from raw pointers around calls to non-rewritten functions.

More specifically, this branch does the following:

* Updates `rewrite::expr::mir_op` to emit casts around function calls.  Given the MIR statement `_2 = f(_1)`, we now generate a cast from the type (`LTy`) of `_1` to the argument type of `f` and a cast from the return type of `f` to the type of `_2`.
* Updates `rewrite::expr::distribute` to handle rewrites from multiple MIR locations being applied to the same HIR node.  For example, the HIR code `f(x)` compiles down to `_tmp = _x; f(_tmp)`; in some test cases, code like this resulted in a cast on the rvalue `_x` of the first assignment and another cast on the argument `_tmp` of the call.  But both MIR locations originate from the same HIR expression, `x`, so having rewrites on both locations was rejected because it's ambiguous which set of rewrites should be applied first to the HIR.  On this branch, `distribute` now properly prioritizes these rewrites and allows this case.
* Adds support for casts between `TyKind::Ref`s and `TyKind::RawPtr`s and between different `RawPtr` types.  This consists of some cleanup and expansion of `emit_cast_desc_desc` in `rewrite::expr::mir_op` and the addition of a few new `Rewrite` variants.
* Adds some new function attributes `#[c2rust_analyze_test::fixed_signature]`, `#[c2rust_analyze_test::skip_rewrite]`, and `#[c2rust_analyze_test::fail_analysis]` for use in test cases.

### DIFF
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -163,6 +163,10 @@ impl<'tcx> intravisit::Visitor<'tcx> for ConvertVisitor<'tcx> {
                     let ty = if *to_mutbl { "*mut _" } else { "*const _" };
                     Rewrite::Cast(Box::new(hir_rw), ty.to_owned())
                 }
+                mir_op::RewriteKind::UnsafeCastRawToRef { mutbl } => {
+                    let rw_pl = Rewrite::Deref(Box::new(hir_rw));
+                    Rewrite::Ref(Box::new(rw_pl), mutbl_from_bool(*mutbl))
+                }
 
                 mir_op::RewriteKind::CellNew => {
                     // `x` to `Cell::new(x)`
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -155,6 +155,15 @@ impl<'tcx> intravisit::Visitor<'tcx> for ConvertVisitor<'tcx> {
                     Rewrite::Ref(Box::new(self.get_subexpr(ex, 0)), mutbl_from_bool(*mutbl))
                 }
 
+                mir_op::RewriteKind::CastRefToRaw { mutbl } => {
+                    let rw_pl = Rewrite::Deref(Box::new(hir_rw));
+                    Rewrite::AddrOf(Box::new(rw_pl), mutbl_from_bool(*mutbl))
+                }
+                mir_op::RewriteKind::CastRawToRaw { to_mutbl } => {
+                    let ty = if *to_mutbl { "*mut _" } else { "*const _" };
+                    Rewrite::Cast(Box::new(hir_rw), ty.to_owned())
+                }
+
                 mir_op::RewriteKind::CellNew => {
                     // `x` to `Cell::new(x)`
                     Rewrite::Call("std::cell::Cell::new".to_string(), vec![Rewrite::Identity])
--- a/c2rust-analyze/src/rewrite/expr/distribute.rs
+++ b/c2rust-analyze/src/rewrite/expr/distribute.rs
@@ -30,6 +30,15 @@ struct RewriteInfo {
     loc: Location,
     sub_loc: Vec<SubLoc>,
     desc: MirOriginDesc,
+    priority: Priority,
+}
+
+/// This enum defines a sort order for [`RewriteInfo`], from innermost (applied earlier) to
+/// outermost (applied later).
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+enum Priority {
+    Eval,
+    LoadResult,
 }
 
 pub fn distribute<'tcx>(
@@ -75,14 +84,18 @@ pub fn distribute<'tcx>(
                 }
             }
 
-            if origin.desc != MirOriginDesc::Expr && origin.desc != MirOriginDesc::LoadFromTemp {
-                error!(
-                    "can't distribute rewrites onto {:?} origin {:?}\n\
-                        key = {:?}\n\
-                        mir_rw = {:?}",
-                    origin.desc, origin, key, mir_rw.kind
-                );
-            }
+            let priority = match origin.desc {
+                MirOriginDesc::Expr => Priority::Eval,
+                MirOriginDesc::LoadFromTemp => Priority::LoadResult,
+                _ => {
+                    panic!(
+                        "can't distribute rewrites onto {:?} origin {:?}\n\
+                            key = {:?}\n\
+                            mir_rw = {:?}",
+                        origin.desc, origin, key, mir_rw.kind
+                    );
+                }
+            };
 
             info_map
                 .entry(origin.hir_id)
@@ -92,19 +105,27 @@ pub fn distribute<'tcx>(
                     loc,
                     sub_loc: key.1,
                     desc: origin.desc,
+                    priority,
                 });
         }
     }
 
-    // If a single `HirId` has rewrites from multiple different pieces of MIR, it's ambiguous how
-    // to order those rewrites.  (`mir_rewrites` only establishes an ordering between rewrites on
-    // the same `Location`.)  For now, we complain if we see this ambiguity; in the future, we may
-    // need to add rules to resolve it in a particular way, such as prioritizing one `SubLoc` or
-    // `MirOriginDesc` over another.
-    for (&hir_id, infos) in &info_map {
-        let all_same_loc = infos.iter().skip(1).all(|i| {
-            i.loc == infos[0].loc && i.sub_loc == infos[0].sub_loc && i.desc == infos[0].desc
-        });
+    // If a single `HirId` has rewrites from multiple different pieces of MIR at the same
+    // `Priority`, it's ambiguous how to order those rewrites.  (`mir_rewrites` only establishes an
+    // ordering between rewrites on the same `Location`.)  For now, we complain if we see this
+    // ambiguity; in the future, we may need to add rules to resolve it in a particular way, such
+    // as prioritizing one `SubLoc` over another.
+    for (&hir_id, infos) in &mut info_map {
+        infos.sort_by_key(|i| i.priority);
+        let mut all_same_loc = true;
+        for (a, b) in infos.iter().zip(infos.iter().skip(1)) {
+            if a.priority != b.priority {
+                continue;
+            }
+            if a.loc != b.loc || a.sub_loc != b.sub_loc || a.desc != b.desc {
+                all_same_loc = false;
+            }
+        }
         if !all_same_loc {
             info!("rewrite info:");
             for i in infos {
--- a/c2rust-analyze/src/rewrite/expr/distribute.rs
+++ b/c2rust-analyze/src/rewrite/expr/distribute.rs
@@ -75,7 +75,7 @@ pub fn distribute<'tcx>(
                 }
             }
 
-            if origin.desc != MirOriginDesc::Expr {
+            if origin.desc != MirOriginDesc::Expr && origin.desc != MirOriginDesc::LoadFromTemp {
                 error!(
                     "can't distribute rewrites onto {:?} origin {:?}\n\
                         key = {:?}\n\
--- a/c2rust-analyze/src/rewrite/expr/mir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/mir_op.rs
@@ -495,6 +495,19 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
             return;
         }
 
+        if from.own == Ownership::Imm && matches!(to.own, Ownership::Raw | Ownership::RawMut) {
+            self.emit(RewriteKind::CastRefToRaw { mutbl: false });
+            from.own = Ownership::Raw;
+        }
+        if from.own == Ownership::Mut && to.own == Ownership::RawMut {
+            self.emit(RewriteKind::CastRefToRaw { mutbl: true });
+            from.own = Ownership::RawMut;
+        }
+        if from.own == Ownership::Raw && to.own == Ownership::RawMut {
+            self.emit(RewriteKind::CastRawToRaw { to_mutbl: true });
+            from.own = Ownership::RawMut;
+        }
+
         if (from.qty, to.qty) == (Quantity::OffsetPtr, Quantity::Slice)
             || (from.qty, to.qty) == (Quantity::Slice, Quantity::OffsetPtr)
         {
--- a/c2rust-analyze/src/rewrite/expr/mir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/mir_op.rs
@@ -51,6 +51,13 @@ pub enum RewriteKind {
     RemoveAsPtr,
     /// Replace &raw with & or &raw mut with &mut
     RawToRef { mutbl: bool },
+
+    /// Cast `&` to `*const` or `&mut` to `*mut`.
+    CastRefToRaw { mutbl: bool },
+    /// Cast `*const` to `*mut` or vice versa.  If `to_mutbl` is true, we are casting to `*mut`;
+    /// otherwise, we're casting to `*const`.
+    CastRawToRaw { to_mutbl: bool },
+
     /// Replace `y` in `let x = y` with `Cell::new(y)`, i.e. `let x = Cell::new(y)`
     /// TODO: ensure `y` implements `Copy`
     CellNew,
--- a/c2rust-analyze/src/rewrite/expr/mir_op.rs
+++ b/c2rust-analyze/src/rewrite/expr/mir_op.rs
@@ -12,6 +12,7 @@ use crate::panic_detail;
 use crate::pointer_id::PointerTable;
 use crate::type_desc::{self, Ownership, Quantity, TypeDesc};
 use crate::util::{ty_callee, Callee};
+use log::*;
 use rustc_ast::Mutability;
 use rustc_middle::mir::{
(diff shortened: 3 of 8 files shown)


kkysen (changes requested):
Could the `#[c2rust_analyze_test::*]` attribute commits (and a test/example for them) be split into their own PR based on `master`?
* e213f85164a41b4b8d88ad895b50e0e4a6115557
* c0cab1376e73f86da42e78562a73054f828ee982

I have some comments about them, but they also seem like not the main focus of this PR at all

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -92,19 +105,27 @@ pub fn distribute<'tcx>(
...
-        });
+    // If a single `HirId` has rewrites from multiple different pieces of MIR at the same
+    // `Priority`, it's ambiguous how to order those rewrites.  (`mir_rewrites` only establishes an
+    // ordering between rewrites on the same `Location`.)  For now, we complain if we see this
+    // ambiguity; in the future, we may need to add rules to resolve it in a particular way, such
+    // as prioritizing one `SubLoc` over another.
+    for (&hir_id, infos) in &mut info_map {
+        infos.sort_by_key(|i| i.priority);
+        let mut all_same_loc = true;
+        for (a, b) in infos.iter().zip(infos.iter().skip(1)) {
```
kkysen, line 121:
```suggestion
        for [a, b] in infos.windows(2).map(|w| [&w[0], &w[1]]) {
```

I think the intent with this would be clearer vs. the zipping way.

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -92,19 +105,27 @@ pub fn distribute<'tcx>(
...
-    for (&hir_id, infos) in &info_map {
-        let all_same_loc = infos.iter().skip(1).all(|i| {
-            i.loc == infos[0].loc && i.sub_loc == infos[0].sub_loc && i.desc == infos[0].desc
-        });
+    // If a single `HirId` has rewrites from multiple different pieces of MIR at the same
+    // `Priority`, it's ambiguous how to order those rewrites.  (`mir_rewrites` only establishes an
+    // ordering between rewrites on the same `Location`.)  For now, we complain if we see this
+    // ambiguity; in the future, we may need to add rules to resolve it in a particular way, such
+    // as prioritizing one `SubLoc` over another.
+    for (&hir_id, infos) in &mut info_map {
```
kkysen, line 118: How about this?

```rust
    for (&hir_id, infos) in &info_map {
        for (priority, infos) in infos.iter().into_group_map_by(|i| i.priority) {
            if !infos
                .iter()
                .map(|i| (i.loc, &i.sub_loc, i.desc))
                .all_equal()
            {
                info!("rewrite info of {priority:?}:");
                for i in infos {
                    info!("  {:?}, {:?}, {:?}: {:?}", i.loc, i.sub_loc, i.desc, i.rw);
                }
                let ex = tcx.hir().expect_expr(hir_id);
                error!(
                    "multiple distinct locations produced rewrites for {:?} {:?}",
                    ex.span, ex,
                );
            }
        }
    }
```

Using `.into_group_map_by` and `.all_equal` would make things a lot more explicit/declarative, and this way we also move the `error!` to the priority group that is ambiguous, instead of printing out all the other infos, too.

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -30,6 +30,15 @@ struct RewriteInfo {
...
     desc: MirOriginDesc,
+    priority: Priority,
+}
+
+/// This enum defines a sort order for [`RewriteInfo`], from innermost (applied earlier) to
+/// outermost (applied later).
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+enum Priority {
+    Eval,
+    LoadResult,
```
kkysen, line 41: Docs for these would be nice, as I'm not exactly what the names are supposed to mean.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -51,6 +52,15 @@ pub enum RewriteKind {
     RemoveAsPtr,
     /// Replace &raw with & or &raw mut with &mut
     RawToRef { mutbl: bool },
+
+    /// Cast `&` to `*const` or `&mut` to `*mut`.
+    CastRefToRaw { mutbl: bool },
+    /// Cast `*const` to `*mut` or vice versa.  If `to_mutbl` is true, we are casting to `*mut`;
+    /// otherwise, we're casting to `*const`.
+    CastRawToRaw { to_mutbl: bool },
```
kkysen, line 60: Is this meant to be casts from `*{const,mut} T` to `*{const,mut} T`, or also between different types?  Could you include that in the docs?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -155,6 +155,19 @@ impl<'tcx> intravisit::Visitor<'tcx> for ConvertVisitor<'tcx> {
                     Rewrite::Ref(Box::new(self.get_subexpr(ex, 0)), mutbl_from_bool(*mutbl))
                 }
 
+                mir_op::RewriteKind::CastRefToRaw { mutbl } => {
+                    let rw_pl = Rewrite::Deref(Box::new(hir_rw));
+                    Rewrite::AddrOf(Box::new(rw_pl), mutbl_from_bool(*mutbl))
+                }
+                mir_op::RewriteKind::CastRawToRaw { to_mutbl } => {
+                    let ty = if *to_mutbl { "*mut _" } else { "*const _" };
```
kkysen, line 163: If `CastRawToRaw` is meant only for `const` to `mut` and vice versa casts (as opposed to transmuting casts between different types), I'm worried the `_` could be inferred in some contexts to be another type (I'm not sure exactly how the inference rules work).  Given we don't have the full type here to use, what about using [`.cast_const`](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast_const) and [`.cast_mut`](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast_mut).  These methods were added for this reason: to avoid accidentally changing the type.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -155,6 +155,19 @@ impl<'tcx> intravisit::Visitor<'tcx> for ConvertVisitor<'tcx> {
                     Rewrite::Ref(Box::new(self.get_subexpr(ex, 0)), mutbl_from_bool(*mutbl))
                 }
 
+                mir_op::RewriteKind::CastRefToRaw { mutbl } => {
+                    let rw_pl = Rewrite::Deref(Box::new(hir_rw));
+                    Rewrite::AddrOf(Box::new(rw_pl), mutbl_from_bool(*mutbl))
+                }
```
kkysen, line 161: If it's already a ref, why do we need to deref and addr_of vs. just casting to the ptr type?

spernsteiner:
> Could the `#[c2rust_analyze_test::*]` attribute commits (and a test/example for them) be split into their own PR based on `master`?

Opened #942

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Good idea - I changed it to use these methods.

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: I guess that works - I'm just used to the Haskell version `zip xs (tail xs)`.  It's a shame `<[T]>::array_windows` isn't stable yet (we could avoid the `map`), but this version is fine.

kkysen: Yeah, I tried to use `.array_windows`, too, but almost all of those array versions are still unstable unfortunately.

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Added a comment.

c2rust-analyze/src/rewrite/expr/distribute.rs
spernsteiner: Added a comment explaining the ordering: a2ebd4f1

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -266,25 +276,32 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+                                        v.enter_call_arg(i, |v| v.visit_operand(op, Some(lty)));
+                                    } else {
+                                        // This is a call to a variadic function, and we've gone
+                                        // past the end of the declared arguments.
+                                        // TODO: insert a cast to turn `op` back into its original
+                                        // declared type (i.e. upcast the chosen reference type
+                                        // back to a raw pointer)
+                                        continue;
+                                    }
+                                }
```
kkysen, line 295:
```suggestion
                                let (args, _variadic_args) = args.split_at(lsig.inputs.len());
                                for (i, (op, &lty)) in iter::zip(args, lsig.inputs).enumerate() {
                                    v.enter_call_arg(i, |v| v.visit_operand(op, Some(lty)));
                                }
                                // TODO: insert a cast to turn `op` back into its original
                                // declared type (i.e. upcast the chosen reference type
                                // back to a raw pointer)
```

I think this would be clearer (and simpler) about the variadic args being the extra args at the end rather than matching on `slice::get`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -266,25 +276,32 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
-                for (i, _op) in args.iter().enumerate() {
-                    if i >= sig.inputs().len() {
-                        // This is a call to a variadic function, and we've gone past the end of
-                        // the declared arguments.
-                        // TODO: insert a cast to turn `op` back into its original declared type
-                        // (i.e. upcast the chosen reference type back to a raw pointer)
-                        continue;
+                    Callee::LocalDef { def_id, substs: _ } => {
+                        // TODO: handle substs (if nonempty)
+                        if let Some(lsig) = self.acx.gacx.fn_sigs.get(&def_id) {
```
kkysen, line 282: There shouldn't be any cases where this returns `None`, right?  Those should be `UnknownDef`s.  In other cases like this, we do `.fn_sigs.get(&def_id).unwrap()` or similar.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -485,51 +502,170 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+                }
+                (Quantity::OffsetPtr, Quantity::Slice) => {
+                    let _rw = match opt_mutbl {
+                        // Currently a no-op, since `Slice` and `OffsetPtr` are identical.
+                        Some(_mutbl) => (),
+                        None => break,
+                    };
+                    // self.emit(rw);
+                    from.qty = Quantity::Slice;
+                }
```
kkysen, line 553: Could these be combined since this is a bidirectional conversion?

```suggestion
                (Quantity::Slice, Quantity::OffsetPtr) | (Quantity::OffsetPtr, Quantity::Slice) => {
                    let _rw = match opt_mutbl {
                        // Currently a no-op, since `Slice` and `OffsetPtr` are identical.
                        Some(_mutbl) => (),
                        None => break,
                    };
                    // self.emit(rw);
                    from.qty = to.qty;
                }
```

Also, why is the `_rw` still needed?  Even if it's used, it's `()`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -485,51 +502,170 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+                // `Slice` and `OffsetPtr` convert to `Single` the same way.
+                // TODO: may be better to use `slice.as_ptr()` to avoid panic on 0-length inputs
```
kkysen, line 555:
```suggestion
                // TODO: We should use `slice.as_ptr()` to avoid panic on 0-length inputs
```

Can we just do this now?  What's the idea behind doing it this way (`&a[0]`) at all in the first place?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -485,51 +502,170 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
-            from.own = to.own;
+        // Late `Ownership` casts.
+        while from.own != to.own {
+            match self.cast_ownership_one_step(from, to, false) {
+                Some(new_own) => {
+                    from.own = new_own;
+                }
+                None => break,
+            }
         }
```
kkysen, line 584: Could we put this in some `cast_ownership_multiple_steps` function given we're doing it many times?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -485,51 +502,170 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+    fn cast_ownership_one_step(
+        &mut self,
+        from: TypeDesc,
+        to: TypeDesc,
+        early: bool,
+    ) -> Option<Ownership> {
+        match from.own {
+            Ownership::Box => match to.own {
+                Ownership::Raw | Ownership::Imm => {
+                    error!("NYI: cast Box to Imm");
```
kkysen, line 603: By the way, for these `NYI`s in `error!`s, can we use `TODO` instead so that we can grep for `TODO`.  Normally these would be `todo!`s, right?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -485,51 +502,170 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
         }
     }
 
+    fn cast_ownership_one_step(
+        &mut self,
+        from: TypeDesc,
+        to: TypeDesc,
+        early: bool,
+    ) -> Option<Ownership> {
+        match from.own {
```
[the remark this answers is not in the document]
kkysen: A `use Ownership::*;` would help a lot as well, I think.

c2rust-analyze/tests/filecheck/call_cast.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    // CHECK: let y: &mut (i32)
+    // CHECK-SAME: pass_single(core::ptr::addr_of_mut!(*(x)))
+    let y: *mut i32 = pass_single(x);
+    use_single(y);
+}
+
+// CHECK-LABEL: fn g(x: &mut [(i32)])
+unsafe fn g(x: *mut i32) {
+    // CHECK: use_single(core::ptr::addr_of_mut!(*&mut (x)[0]))
```
aneksteind, line 49: why isn't this the same as the test above:
```rust
// CHECK: use_single(core::ptr::addr_of_mut!(*(x)))
```

aneksteind (approved):

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: `&x[0]` goes from `&[T]` to `&T`; `x.as_ptr()` goes from `&[T]` to `*const T`.  The latter doesn't fit well into the current design because it changes ownership and quantity at the same time, though I think we could add it as a special case if it's needed at some point.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: I thought the idea with #854 is that we should now use `error!` instead of `panic!`/`todo!` when possible so that the error can be made non-fatal by adjusting the `RUST_LOG_PANIC` env var.

c2rust-analyze/src/rewrite/expr/mir_op.rs
kkysen: I'm curious, in what cases is this done where `&x[0]` is not used as a slice?

c2rust-analyze/src/rewrite/expr/mir_op.rs
kkysen: Yeah, that's right.  I just mean doing something like `error!("TODO: ...")` so we can easily search for TODOs.  Or we could add a wrapper `todo!` macro in `mod log`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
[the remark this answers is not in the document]
spernsteiner: I tried that, but IMO it makes it harder to see the grouping (first all the cases where `from.own == Box`, then all `Rc`, etc)

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Ah, that makes sense.  I'll do that.

c2rust-analyze/src/rewrite/expr/mir_op.rs
kkysen: Looks to be fixed in 6d099d07176a24079c9b3ba8d51d1d29b4979bb2.

kkysen (approved):

PR #939, opened 2023-05-26
analyze: generate shims for calls from non-rewritten to rewritten code

spernsteiner:
Implements generation of unsafe shims when calling rewritten code from non-rewritten code.  Given this input:

```Rust
fn bad(x: *const i32) {
    good(x);
    // Do some unsupported operation
}

fn good(x: *const i32) {
    // ...
}
```

The tool now produces output like this:

```Rust
fn bad(x: *const i32) {
    good_shim(x);
    // Do some unsupported operation
}

fn good(x: &i32) {
    // ...
}

unsafe fn good_shim(x: *const i32) {
    good(&*x);
}
```

Here, the tool has rewritten `good` to change its argument type from `*const i32` to `&i32`, but analysis failed on `bad`, so the tool can't apply corresponding rewrites in the body of `bad`.  Previously, this would cause a type error: `bad` would continue passing `*const i32` to `good`, but `good` now expects `&i32`.  Now, the tool generates a helper function `good_shim`, which wraps `good` but keeps the original signature from before rewriting, and rewrites `bad` to call `good_shim` instead of `good`, resulting in a well-typed program.  Renaming the callee in `bad` doesn't require any analysis results, so it works even though analysis failed on `bad`.

Implementation strategy: for each failed (non-rewritten) function, we walk over the HIR and look for `ExprKind::Path`s and `ExprKind::MethodCall`s that resolve to rewritten functions.  For each one, we generate a rewrite that appends `_shim` to the end of the function name and record the `DefId` of the callee.  Then, for each callee we encountered, we generate a rewrite that inserts the shim function definition after the definition of the callee.  The shim function has to cast arguments and return values between the original unsafe types and the safe types produced by rewriting; for this, we reuse the cast-generation machinery from `rewrite::expr::mir_op` and `rewrite::expr::convert`.

This is a counterpart to #936: that one handles calls from rewritten code into non-rewritten code, and this one handles the reverse.

This PR also marks failed `fn`s as `FIXED` and thus doesn't propagate analysis failures to callers.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -657,6 +657,10 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             );
         }
     }
+
+    pub fn iter_fns_failed<'a>(&'a self) -> impl Iterator<Item = DefId> + 'a {
+        self.fns_failed.keys().copied()
+    }
 }
 
 impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -647,15 +647,6 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         }
 
         self.fns_failed.insert(did, detail);
-
-        // This is the first time marking `did` as failed, so also mark all of its callers.
-        let callers = self.fn_callers.get(&did).cloned().unwrap_or(Vec::new());
-        for caller in callers {
-            self.mark_fn_failed(
-                caller,
-                PanicDetail::new(format!("analysis failed on callee {:?}", did)),
-            );
-        }
     }
 
     pub fn iter_fns_failed<'a>(&'a self) -> impl Iterator<Item = DefId> + 'a {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -273,9 +273,6 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     ptr_info: GlobalPointerTable<PointerInfo>,
 
-    /// Map from a function to all of its callers.
-    pub fn_callers: HashMap<DefId, Vec<DefId>>,
-
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
     /// `DefId`s of functions where analysis failed, and a [`PanicDetail`] explaining the reason
@@ -532,7 +529,6 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx,
             lcx: LabeledTyCtxt::new(tcx),
             ptr_info: GlobalPointerTable::empty(),
-            fn_callers: HashMap::new(),
             fn_sigs: HashMap::new(),
             fns_failed: HashMap::new(),
             field_ltys: HashMap::new(),
@@ -579,7 +575,6 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             tcx: _,
             lcx,
             ref mut ptr_info,
-            fn_callers: _,
             ref mut fn_sigs,
             fns_failed: _,
             ref mut field_ltys,
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -27,6 +27,7 @@ use rustc_middle::ty::{
 use rustc_type_ir::RegionKind::{ReEarlyBound, ReStatic};
 use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
+use std::iter;
 use std::ops::Index;
 
 bitflags! {
@@ -164,6 +165,12 @@ pub type LTyCtxt<'tcx> = LabeledTyCtxt<'tcx, PointerId>;
 pub struct LFnSig<'tcx> {
     pub inputs: &'tcx [LTy<'tcx>],
     pub output: LTy<'tcx>,
+}
+
+impl<'tcx> LFnSig<'tcx> {
+    pub fn inputs_and_output(&self) -> impl Iterator<Item = LTy<'tcx>> + 'tcx {
+        self.inputs.iter().copied().chain(iter::once(self.output))
+    }
 }
 
 bitflags! {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -644,7 +644,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         self.fns_failed.insert(did, detail);
     }
 
-    pub fn iter_fns_failed<'a>(&'a self) -> impl Iterator<Item = DefId> + 'a {
+    pub fn iter_fns_failed(&self) -> impl Iterator<Item = DefId> + '_ {
         self.fns_failed.keys().copied()
     }
 }
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -400,24 +400,11 @@ fn run(tcx: TyCtxt) {
 
     // Follow a postorder traversal, so that callers are visited after their callees.  This means
     // callee signatures will usually be up to date when we visit the call site.
-    let (all_fn_ldids, fn_callers) = fn_body_owners_postorder(tcx);
+    let all_fn_ldids = fn_body_owners_postorder(tcx);
     eprintln!("callgraph traversal order:");
     for &ldid in &all_fn_ldids {
         eprintln!("  {:?}", ldid);
     }
-
-    gacx.fn_callers = fn_callers
-        .into_iter()
-        .map(|(ldid, caller_ldids)| {
-            (
-                ldid.to_def_id(),
-                caller_ldids
-                    .into_iter()
-                    .map(|caller_ldid| caller_ldid.to_def_id())
-                    .collect(),
-            )
-        })
-        .collect();
 
     // Assign global `PointerId`s for all pointers that appear in function signatures.
     for &ldid in &all_fn_ldids {
@@ -1153,12 +1140,9 @@ fn all_static_items(tcx: TyCtxt) -> Vec<DefId> {
 /// Return all `LocalDefId`s for all `fn`s that are `body_owners`, ordered according to a postorder
 /// traversal of the graph of references between bodies.  Also returns the callgraph itself, in the
 /// form of a map from callee `LocalDefId` to a set of caller `LocalDefId`s.
-fn fn_body_owners_postorder(
-    tcx: TyCtxt,
-) -> (Vec<LocalDefId>, HashMap<LocalDefId, HashSet<LocalDefId>>) {
+fn fn_body_owners_postorder(tcx: TyCtxt) -> Vec<LocalDefId> {
     let mut seen = HashSet::new();
     let mut order = Vec::new();
-    let mut callers = HashMap::<_, HashSet<_>>::new();
     enum Visit {
         Pre(LocalDefId),
         Post(LocalDefId),
@@ -1187,7 +1171,6 @@ fn fn_body_owners_postorder(
                         stack.push(Visit::Post(ldid));
                         for_each_callee(tcx, ldid, |callee_ldid| {
                             stack.push(Visit::Pre(callee_ldid));
-                            callers.entry(callee_ldid).or_default().insert(ldid);
                         });
                     }
                 }
@@ -1198,7 +1181,7 @@ fn fn_body_owners_postorder(
         }
     }
 
-    (order, callers)
+    order
 }
 
 fn for_each_callee(tcx: TyCtxt, ldid: LocalDefId, f: impl FnMut(LocalDefId)) {
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -879,8 +879,20 @@ fn run(tcx: TyCtxt) {
 
     let (shim_rewrites, shim_fn_def_ids) = rewrite::gen_shim_call_rewrites(&gacx, &gasn);
     all_rewrites.extend(shim_rewrites);
-    let shim_def_rewrites = rewrite::gen_shim_definition_rewrites(&gacx, &gasn, shim_fn_def_ids);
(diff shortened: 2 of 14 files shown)


c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -391,6 +404,36 @@ impl<S: Sink> Emitter<'_, S> {
...
                 self.emit_sub(0, span)
             }
+
+            Rewrite::DefineFn {
+                ref name,
+                ref arg_tys,
+                ref return_ty,
+                ref body,
+            } => {
+                self.emit_fmt(format_args!("\nunsafe fn {name}("))?;
```
aneksteind, line 414: is it worth making `unsafe` an optional addition?

c2rust-analyze/tests/filecheck/unrewritten_calls.rs
```diff
@@ -0,0 +1,28 @@
...
+    // CHECK: let y = &*((g(core::ptr::addr_of_mut!(*(x)))) as *const _);
+    let y = g(x);
+    *y
+}
+
+// The middle function, `g`, will be marked as failed.  We wait until analysis has finished before
+// marking it so that the `WRITE` requirement from `h` will be propagated up to `f`.
+#[c2rust_analyze_test::fail_before_rewriting]
+// CHECK-LABEL: fn g(x: *mut i32) -> *mut i32
+unsafe fn g(x: *mut i32) -> *mut i32 {
```
aneksteind, line 16: can you relabel these as `good`, `bad`, etc for clarity

aneksteind (approved):

c2rust-analyze/tests/filecheck/unrewritten_calls.rs
```diff
@@ -3,21 +3,26 @@
...
-    h(x);
+    h(x)
 }
 
-// CHECK-LABEL: fn h(x: &mut (i32))
-unsafe fn h(x: *mut i32) {
+// CHECK-LABEL: fn h(x: &mut (i32)) -> &(i32)
+unsafe fn h(x: *mut i32) -> *mut i32 {
     *x = 1;
+    // CHECK: &*(x)
```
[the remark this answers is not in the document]
spernsteiner: Yes, this reborrow will get the same lifetime.

In general, the implicit coercion from `&mut T` to `&T` produces MIR like `_2 = &(*_1)`.  [Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=311ffa5474289e9cdf086ff3f6632a99).  Here, we're just doing the same thing explicitly.

c2rust-analyze/src/pointer_id.rs
```diff
@@ -336,6 +336,14 @@ impl<'a, T> PointerTable<'a, T> {
     }
 }
 
+impl<'a, T> Clone for PointerTable<'a, T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<'a, T> Copy for PointerTable<'a, T> {}
```
kkysen, line 345: Can this just be a `#[derive(Clone, Copy)]`?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -391,6 +404,36 @@ impl<S: Sink> Emitter<'_, S> {
...
+                for (i, arg_ty) in arg_tys.iter().enumerate() {
+                    if i > 0 {
+                        self.emit_str(", ")?;
+                    }
+                    self.emit_fmt(format_args!("arg{i}: "))?;
```
kkysen, line 419: Does preserving the argument names matter?

kkysen (commented):
What happens for shim functions for non-just a ptr to reference cases, such as slices and other such non-trivial conversions from raw pointers to safe types?

c2rust-analyze/src/pointer_id.rs
spernsteiner: `#[derive(Clone, Copy)]` would add `T: Clone` / `T: Copy` bounds, which aren't necessary.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: I don't think it matters, as long as the function declaration and body agree on the names.  That's why I have it using `arg0`, `arg1`, etc instead of the original names.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: We don't need that flexibility at the moment, and it should be easy to add later if that changes.

c2rust-analyze/src/pointer_id.rs
kkysen: Ah, it's that derive bug.  Could you just add a comment here saying that's why we're `impl`ing it manually?

c2rust-analyze/src/rewrite/apply.rs
kkysen: Okay, that should be good enough for now.  I think ideally we'd want to match the arg names of the original function, but it's not too important as long as the arg declarations and usages match as you said.

kkysen:
Can I take a little bit more of a look at this after it's rebased?

c2rust-analyze/src/main.rs
```diff
@@ -508,14 +495,15 @@ fn run(tcx: TyCtxt) {
...
+    // failed at this point.
     for &ldid in &all_fn_ldids {
-        if !util::has_test_attr(tcx, ldid, TestAttr::FailAnalysis) {
+        if !util::has_test_attr(tcx, ldid, TestAttr::FailBeforeAnalysis) {
             continue;
         }
         gacx.mark_fn_failed(
             ldid.to_def_id(),
-            PanicDetail::new("explicit fail_analysis for testing".to_owned()),
+            PanicDetail::new("explicit fail_before_analysis for testing".to_owned()),
```
kkysen, line 506: nit:

```suggestion
            PanicDetail::new(format!("explicit {} for testing", TestAttr::FailBeforeAnalysis::name())),
```

c2rust-analyze/src/main.rs
```diff
@@ -576,6 +564,35 @@ fn run(tcx: TyCtxt) {
...
 
+    // For testing, putting #[c2rust_analyze_test::fail_before_rewriting] on a function marks it as
+    // failed at this point.
+    for &ldid in &all_fn_ldids {
+        if !util::has_test_attr(tcx, ldid, TestAttr::FailBeforeRewriting) {
+            continue;
+        }
+        gacx.mark_fn_failed(
+            ldid.to_def_id(),
+            PanicDetail::new("explicit fail_before_rewriting for testing".to_owned()),
```
kkysen, line 575: nit:

```suggestion
            PanicDetail::new(format!("explicit {} for testing", TestAttr::FailBeforeRewriting.name())),
```

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -154,21 +140,6 @@ impl<'tcx> Visitor<'tcx> for ConvertVisitor<'tcx> {
...
-                    Rewrite::MethodCall(method.to_string(), Box::new(hir_rw), vec![])
-                }
-                mir_op::RewriteKind::UnsafeCastRawToRef { mutbl } => {
-                    let rw_pl = Rewrite::Deref(Box::new(hir_rw));
-                    Rewrite::Ref(Box::new(rw_pl), mutbl_from_bool(*mutbl))
-                }
-
                 mir_op::RewriteKind::CellNew => {
                     // `x` to `Cell::new(x)`
                     Rewrite::Call("std::cell::Cell::new".to_string(), vec![Rewrite::Identity])
```
kkysen, line 145: Why not include `CellNew` in `convert_cast_rewrite`?  It doesn't modify the original expression and is a kind of cast or conversion, similar to `CellFromMut`, which is included in `convert_cast_rewrite`?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -484,13 +485,70 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+    PT2: Index<PointerId, Output = FlagSet>,
+    F: FnMut(RewriteKind),
+{
+    pub fn new(
+        tcx: TyCtxt<'tcx>,
+        perms: &'a PT1,
+        flags: &'a PT2,
+        emit: F,
+    ) -> CastBuilder<'a, 'tcx, PT1, PT2, F> {
+        CastBuilder {
```
kkysen, line 536: nit:

```suggestion
    ) -> Self {
        Self {
```

c2rust-analyze/src/context.rs
```diff
@@ -642,15 +637,10 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
 
-        // This is the first time marking `did` as failed, so also mark all of its callers.
-        let callers = self.fn_callers.get(&did).cloned().unwrap_or(Vec::new());
-        for caller in callers {
-            self.mark_fn_failed(
-                caller,
-                PanicDetail::new(format!("analysis failed on callee {:?}", did)),
-            );
-        }
+    pub fn iter_fns_failed<'a>(&'a self) -> impl Iterator<Item = DefId> + 'a {
```
kkysen, line 642:
```suggestion
    pub fn iter_fns_failed(&self) -> impl Iterator<Item = DefId> + '_ {
```

c2rust-analyze/src/main.rs
```diff
@@ -576,6 +564,35 @@ fn run(tcx: TyCtxt) {
...
+        );
+    }
+
+    // Before generating rewrites, add the FIXED flag to the signatures of all functions that
+    // failed analysis.
+    for did in gacx.iter_fns_failed() {
+        let lsig = match gacx.fn_sigs.get(&did) {
+            Some(x) => x,
+            None => continue,
+        };
```
kkysen, line 585:
```suggestion
        let lsig = gacx.fn_sigs[&did];
```

This lookup shouldn't fail, right?

c2rust-analyze/src/main.rs
```diff
@@ -576,6 +564,35 @@ fn run(tcx: TyCtxt) {
...
+        for sig_lty in lsig.inputs.iter().copied().chain(iter::once(lsig.output)) {
```
kkysen, line 586: nit: A `.inputs_and_output()` method would be nice.  I think the `rustc` `FnSig` equivalent has something similar.

c2rust-analyze/src/main.rs
```diff
@@ -636,6 +653,23 @@ fn run(tcx: TyCtxt) {
         info.acx_data.set(acx.into_data());
     }
 
+    let (shim_call_rewrites, shim_fn_def_ids) = rewrite::gen_shim_call_rewrites(&gacx, &gasn);
```
kkysen, line 656: Can this never panic?

c2rust-analyze/src/rewrite/shim.rs
```diff
@@ -0,0 +1,240 @@
...
+        FnRetTy::DefaultReturn(..) => None,
+        FnRetTy::Return(ty) => Some(Box::new(Rewrite::Extract(ty.span))),
+    };
+
+    // `def_id` should always refer to a rewritten function, and all rewritten functions have
+    // valid `fn_sigs` entries.
+    let lsig = gacx
+        .fn_sigs
+        .get(&def_id)
+        .unwrap_or_else(|| panic!("missing lsig for {:?}", def_id));
```
kkysen, line 201: nit: We have a bunch of these, and I think we can just use `[...]` indexing here vs. custom panic messages.

```suggestion
    let lsig = gacx.fn_sigs[&def_id];
```

kkysen (changes requested):
See the above comments, especially the ones about `unsafe fn good2_shim` in the test and not catching panics in `gen_shim_call_rewrites`, as well as a bunch of smaller ones.

Could you also add to the PR description that this marks failed `fn`s as `FIXED` and thus doesn't propagate analysis failures to callers, as that's a significant change and improvement from/over before?  Thanks!

c2rust-analyze/src/rewrite/shim.rs
```diff
@@ -0,0 +1,240 @@
...
+            .inputs
+            .iter()
+            .cloned()
+            .chain(iter::once(lsig.output))
+            .flat_map(|lty| lty.iter())
+            .any(|lty| {
+                let ptr = lty.label;
+                !ptr.is_none() && !self.gasn.flags[ptr].contains(FlagSet::FIXED)
+            });
+        if !has_non_fixed_ptr {
```
kkysen, line 65: nit: I think `if has_all_fixed_ptrs` would be a bit clearer here, flipping the `.any` to `.all`.

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: That makes sense - I've moved it.

c2rust-analyze/src/main.rs
spernsteiner: Good idea.  In the future we might expand analysis to static initializers, and it would be good for this to panic if a non-function has failed analysis instead of silently ignoring it.

c2rust-analyze/src/main.rs
spernsteiner: Added a comment:
```Rust
    // This call never panics, which is important because this is the fallback if the more
    // sophisticated analysis and rewriting above did panic.
```

c2rust-analyze/src/rewrite/shim.rs
```diff
@@ -0,0 +1,237 @@
...
+        // has at least one non-`FIXED` pointer in its signature.
+        let lsig = match self.gacx.fn_sigs.get(&def_id) {
+            Some(x) => x,
+            None => return,
+        };
+        let has_non_fixed_ptr = lsig
+            .inputs
+            .iter()
+            .copied()
+            .chain(iter::once(lsig.output))
```
kkysen, line 59:
```suggestion
        let has_non_fixed_ptr = lsig.inputs_and_output()
```

kkysen (approved):
I still have a few simple comments/suggestions that would be nice to have, but the important changes have all been made, so this LGTM now.

PR #942, opened 2023-05-31
analyze: add function attrs for testing

spernsteiner:
Adds support for some function attributes that are useful for testing:

* `#[c2rust_analyze_test::fixed_signature]`: Marks all pointers in the function signature as `FIXED`
* `#[c2rust_analyze_test::fail_analysis]`: Forces an analysis failure for the function
* `#[c2rust_analyze_test::skip_rewrite]`: Skips generating rewrites for the function


c2rust-analyze/src/main.rs
```diff
@@ -493,6 +497,30 @@ fn run(tcx: TyCtxt) {
         info.lasn.set(lasn);
     }
 
+    // For testing, putting #[c2rust_analyze_test::fixed_signature] on a function makes all
+    // pointers in its signature FIXED.
+    for &ldid in &all_fn_ldids {
+        if !has_test_attr(tcx, ldid, "fixed_signature") {
```
kkysen, line 503: Could we make an `enum` for all the possible attributes?

kkysen: That way we avoid any little typos and can put the docs for each attribute there.

spernsteiner: I went ahead and did this, even though I don't think it's really necessary.  Typos here seem likely to be caught quickly (if you're adjusting the test attributes, it's probably because you're working on a test and will be checking its behavior closely), and `git grep foo` already got you pretty close to the comment describing the `foo` attribute.

kkysen (changes requested):
Can you add a simple example/test for this, too?

kkysen (commented):
`#![feature(register_tool)]` (https://github.com/rust-lang/rust/issues/66079) seems fairly unstable, but I don't see another way of adding this functionality simply, so I think it's fine here.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +355,44 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+            TestAttr::SkipRewrite => "skip_rewrite",
+        }
+    }
+}
+
+pub fn has_test_attr(tcx: TyCtxt, ldid: LocalDefId, attr: TestAttr) -> bool {
+    let tool_sym = Symbol::intern("c2rust_analyze_test");
+    let name_sym = Symbol::intern(attr.name());
+
+    for attr in tcx.get_attrs_unchecked(ldid.to_def_id()) {
```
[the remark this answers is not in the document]
kkysen: Actually, `TyCtxt::get_attrs` seems to require an attr with only 1 segment, and I don't see any other alternatives to `TyCtxt::get_attrs_unchecked`, so it's probably fine.

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +355,44 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+        let path = match attr.kind {
+            AttrKind::Normal(ref item, _) => &item.path,
+            AttrKind::DocComment(..) => continue,
+        };
+        let (a, b) = match &path.segments[..] {
+            &[ref a, ref b] => (a, b),
```
kkysen, line 390: nit:

```suggestion
            [a, b] => (a, b),
```

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +355,44 @@ pub fn is_null_const(constant: Constant) -> bool {
 
 pub trait PhantomLifetime<'a> {}
 impl<'a, T: ?Sized> PhantomLifetime<'a> for T {}
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
+pub enum TestAttr {
+    /// `#[c2rust_analyze_test::fixed_signature]`: Mark all pointers in the function signature as
+    /// `FIXED`.
```
kkysen, line 362:
```suggestion
    /// [`FIXED`](crate::context::FlagSet::FIXED).
```

c2rust-analyze/src/util.rs
```diff
@@ -353,3 +355,44 @@ pub fn is_null_const(constant: Constant) -> bool {
...
+        match self {
+            TestAttr::FixedSignature => "fixed_signature",
+            TestAttr::FailAnalysis => "fail_analysis",
+            TestAttr::SkipRewrite => "skip_rewrite",
+        }
+    }
+}
+
+pub fn has_test_attr(tcx: TyCtxt, ldid: LocalDefId, attr: TestAttr) -> bool {
+    let tool_sym = Symbol::intern("c2rust_analyze_test");
```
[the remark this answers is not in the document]
spernsteiner: I'm inclined to leave it for now.  We can do the rename just as easily in the future if we decide we actually want to use these for something in lighttpd.

spernsteiner:
>Can you add a simple example/test for this, too?

Added a basic test case in 2244cbe14828895ce7aa9f4da0725c9ee5ee5d59

c2rust-analyze/src/util.rs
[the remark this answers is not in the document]
kkysen: Okay, that's fine.

c2rust-analyze/src/util.rs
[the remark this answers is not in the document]
kkysen: I eventually realized the same.  The `FIXME` is still there in today's nightly.

c2rust-analyze/src/util.rs
spernsteiner: I mostly avoid using the "match ergonomics" feature - it makes it less clear whether the `match` is binding by reference or by value, which is generally useful to know IMO.

kkysen: Isn't that clear from the `&` in the `&path.segments[..]`?  The `&` and `ref` are basically inverses, right, so doing `&[ref` seems quite redundant.

kkysen: IMO `&[ref a, ref b]` is much more confusing than `[a, b]`.  Match ergonomics are used all over the place, in most idiomatic Rust code and ours.

kkysen (approved):

c2rust-analyze/src/main.rs
```diff
@@ -493,6 +495,30 @@ fn run(tcx: TyCtxt) {
...
+    // For testing, putting #[c2rust_analyze_test::fixed_signature] on a function makes all
+    // pointers in its signature FIXED.
+    for &ldid in &all_fn_ldids {
+        if !util::has_test_attr(tcx, ldid, TestAttr::FixedSignature) {
+            continue;
+        }
+        let lsig = match gacx.fn_sigs.get(&ldid.to_def_id()) {
+            Some(x) => x,
+            None => continue,
+        };
```
kkysen, line 507: @spernsteiner, I vaguely remember you answering this before, but I can't find the comment.  Why do we `continue` here instead of panicking?  When would the lookup fail?

spernsteiner: Not sure.  I tried changing the `continue` to a `panic!()` just now and nothing broke in the tests (though only a few of those use `fixed_signature`).

PR #947, opened 2023-06-09
(`c2rust-analyze`) Add more (still incomplete) dataflow constraints for ptr casts

kkysen:
3c038606a32329d3e4ce21e308e8a49635a6d44f adds the simple (still incomplete) dataflow constraints for `CastKind::Misc`.  The full constraints are trickier since the types can be recursive, so I wanted to split this simpler part out into its own smaller PR first.

b20cea742881dee4c5b4dfe3003345c7db161250 adds complete dataflow constraints for `CastKind::Pointer`, which is modeled on `do_assign`, but adjusted given its a ptr cast.

### DIFF
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -13,6 +13,7 @@ use rustc_middle::mir::{
 };
 use rustc_middle::ty::adjustment::PointerCast;
 use rustc_middle::ty::{SubstsRef, Ty, TyKind};
+use std::iter;
 
 /// Visitor that walks over the MIR, computing types of rvalues/operands/places and generating
 /// constraints as a side effect.
@@ -135,8 +136,30 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                     PointerCast::ArrayToPointer => {}
                     PointerCast::Unsize => {}
                 }
-                self.do_assign_pointer_ids(to_lty.label, from_lty.label)
-                // TODO add other dataflow constraints
+
+                // We would normally do `self.do_assign`,
+                // but we can't do a normal `self.do_equivalence_nested` (inside `self.do_assign`),
+                // so we do the `self.do_assign_pointer_ids` from `self.do_assign`,
+                // and then do the body of `self.do_equivalence_nested`,
+                // but after the extra outer layers are stripped.
+
+                self.do_assign_pointer_ids(to_lty.label, from_lty.label);
+
+                // Each of these ptr casts needs at least the outer layer stripped (e.x. `&`, `*mut`, `fn`),
+                // but then after that, these ptr casts below
+                // needs more layers stripped to reach the types that are equivalent.
+                let strip_from = |lty: LTy<'tcx>| match ptr_cast {
+                    PointerCast::ArrayToPointer => lty.args[0],
+                    PointerCast::Unsize => lty.args[0],
+                    _ => lty,
+                };
+                let strip_to = |lty: LTy<'tcx>| match ptr_cast {
+                    PointerCast::Unsize => lty.args[0],
+                    _ => lty,
+                };
+                for (&from_lty, &to_lty) in iter::zip(from_lty.args, to_lty.args) {
+                    self.do_unify(strip_from(from_lty), strip_to(to_lty));
+                }
             }
             CastKind::Misc => {
                 match is_transmutable_ptr_cast(from_ty, to_ty) {
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -141,6 +141,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
             CastKind::Misc => {
                 match is_transmutable_ptr_cast(from_ty, to_ty) {
                     Some(true) => {
+                        self.do_assign_pointer_ids(to_lty.label, from_lty.label);
                         // TODO add other dataflow constraints
                     },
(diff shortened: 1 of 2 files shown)


c2rust-analyze/tests/analyze/string_casts.rs
```diff
@@ -33,3 +33,57 @@ pub fn cast_from_literal() {
 pub fn cast_from_literal_explicit() {
     std::ptr::addr_of!(*b"\0") as *const u8 as *const core::ffi::c_char;
 }
+
+/// [`PointerCast::ReifyFnPointer`]
+/// Can't figure out how to create a [`PointerCast::ReifyFnPointer`].
+#[cfg(any())]
```
aneksteind, line 39: where do these tests crash downstream? create an issue outlining them?

aneksteind (approved):
would like @spernsteiner's okay but this LGTM as a start

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -135,12 +136,35 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                // needs more layers stripped to reach the types that are equivalent.
+                let strip_from = |lty: LTy<'tcx>| match ptr_cast {
+                    PointerCast::ArrayToPointer => lty.args[0],
+                    PointerCast::Unsize => lty.args[0],
+                    _ => lty,
+                };
+                let strip_to = |lty: LTy<'tcx>| match ptr_cast {
+                    PointerCast::Unsize => lty.args[0],
+                    _ => lty,
+                };
```
spernsteiner, line 159: I think it would be clearer to match on the "from" and "to" `TyKind`s, or else to have a single case per `PointerCast` variant and assert that the types have a particular shape (for both sanity-checking and documentation purposes).  As is, it's hard for me to tell whether or not these rules are correct.  Also, I think matching on `TyKind`s is necessary to distinguish `(x: &[u8; 3]) as &[u8]` from `(x: &[u8; 3]) as &dyn Debug` (both are `PointerCast::Unsize`, but they need different handling here).

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -135,12 +136,35 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                for (&from_lty, &to_lty) in iter::zip(from_lty.args, to_lty.args) {
+                    self.do_unify(strip_from(from_lty), strip_to(to_lty));
+                }
```
spernsteiner, line 162: Similar to my previous comment, I think it would be good to be clear about how many args we expect to see for each `PointerCast` case.  For example, `Unsize` should always have 1 arg in both `from_lty` and `to_lty`, while for `UnsafeFnPointer` we only expect that `from_lty` and `to_lty` will have the same amount.

I also think it's okay to panic on the tricky cases `ReifyFnPointer` and `ClosureFnPointer` for now.

c2rust-analyze/tests/analyze/string_casts.rs
```diff
@@ -33,3 +33,57 @@ pub fn cast_from_literal() {
 pub fn cast_from_literal_explicit() {
     std::ptr::addr_of!(*b"\0") as *const u8 as *const core::ffi::c_char;
 }
+
+/// [`PointerCast::ReifyFnPointer`]
+/// Can't figure out how to create a [`PointerCast::ReifyFnPointer`].
```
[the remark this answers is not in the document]
kkysen: I thought that would happen here but it didn't (it turned into some other cast).  I casted `impl Fn(u8) -> i8` to `fn(u8) -> i8`, and I thought `impl Fn(u8) -> i8` could be a `FnDef`.  Let me try again with a more explicit case like your example.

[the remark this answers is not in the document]
spernsteiner: > I thought `impl Fn(u8) -> i8` could be a `FnDef`

It could be, but isn't guaranteed to be, so the MIR can't use the `FnDef`-specific `PointerCast`.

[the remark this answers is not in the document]
kkysen: Ahh.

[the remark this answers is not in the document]
kkysen: Fixed in e02903c7f11fc719062ecc13bb0ca5a0d9e2655d.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: That's a good point about `dyn` fat ptrs; I hadn't thought of them.  Should I just `error!` on `as &dyn _`s since there shouldn't be any `dyn`s in `lighttpd` or any other `c2rust transpile`d codebase?

c2rust-analyze/src/dataflow/type_check.rs
kkysen: How tricky is `ReifyFnPointer`?  If it's not too complex, I'd rather not error on it as it is used commonly in `lighttpd` and other `c2rust transpile`d code for `fn` ptrs.

spernsteiner: With `ReifyFnPointer`, I haven't thought about what constraints we ought to generate for pointer arguments in the function's signature.  For example, given `fn f(p: *mut /*p0*/ i32)` and the expression `f as fn(*mut /*p1*/ i32)`, do we require `p1`'s permissions to be a subset of `p0`'s, or do we require the reverse, or do we require the two to be equal?

PR #956, opened 2023-06-12
Add hypothetical lifetime to structure rewrites with empty generic type parameter lists

aneksteind:
Deconstructing an HIR field type would fail in the case where there where the field type had no generic parameters, in that it would not be rewritten to include the new hypothetical lifetimes. Now the type is deconstructed correctly by providing an empty generic type parameter list, which allows the hir_args to have the correct value and allows `adt_ty_rw` to run and generate a rewrite for the ADT

### DIFF
--- a/c2rust-analyze/src/rewrite/ty.rs
+++ b/c2rust-analyze/src/rewrite/ty.rs
@@ -120,19 +120,19 @@ where
     })
 }
 
-// Gets the generic type arguments of an HIR type. If the type has no generic
-// arguments, an empty vector is returned
-fn hir_generic_ty_args<'tcx>(ty: &hir::Ty<'tcx>) -> Vec<&'tcx hir::Ty<'tcx>> {
+// Gets the generic type arguments of an HIR type.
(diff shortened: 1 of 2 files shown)


c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -140,6 +142,7 @@ fn hir_generic_ty_args<'tcx>(ty: &hir::Ty<'tcx>) -> Option<Vec<&'tcx hir::Ty<'tc
             })
             .collect()
     })
+    .unwrap_or_default()
```
fw-immunant, line 145: I think it makes more sense for this function to still return `Option<_>`, because that preserves the distinction between a `TyKind` that *can't* have generic parameters (where it would be an error for us to add lifetime parameters) and one that can but doesn't yet. To effect the semantic change we want, the `=> args` in this function would become ``=> Some(args.unwrap_or_default()) /* convert missing type arguments to an empty `Vec` */``, and `deconstruct_hir_ty` would be reverted to what it was before this PR.

fw-immunant (approved):
I believe this fixes the bug (so approving), but I think the handling should change; see comment.

Issue #952, opened 2023-06-09
analyze: bad rewrite of derive(Clone)


spernsteiner:
Input:
```Rust
#[derive(Clone)]
struct Foo {
    x: i32,
}
```
Output:
```Rust
#[derive(Foo)]
struct Foo {
    x: i32,
}
```
Notice `derive(Clone)` has been replaced with `derive(Foo)`, which is bogus.

We previously had some logic to suppress rewrites coming from `derive`-generated functions, but this seems not to work any more.

Issue #949, opened 2023-06-09
analyze: missing lifetime args for structs with no pre-existing lifetimes


spernsteiner:
```Rust
struct Foo {
    ptr: *mut u8,
}

struct Bar {
    foo: *mut Foo,
}
```

`struct Foo` gains a hypothetical lifetime argument to use on `ptr`.  This means the `foo` field should get rewritten to something like `foo: &'h1 Foo<'h2>`, but instead it gets rewritten to `foo: &'h1 Foo`, which is missing the lifetime argument.

PR #958, opened 2023-06-13
rewrite: add #[derive(Clone)] test case

aneksteind:
It turns out #952 was addressed indirectly via #956. This adds a test case to demonstrate the functionality is now working as expected.


fw-immunant (approved):
I don't understand quite how that change fixed this, but test coverage is good in any case.

aneksteind:
It fixed the issue because the prior code would:
- try to deconstruct the HIR type according the the MIR type
- that would fail because of #949 and so `Foo` would indiscriminately get pushed as a rewrite for `Clone`

PR #960, opened 2023-06-14
Do not rewrite ADTs mentioned in extern blocks

### LINKED ISSUE #948: analyze: avoid rewriting structs mentioned in extern declarations
Structs like `FILE`, which are used in external libc functions like `fwrite`, should have their fields marked `FIXED` so we don't try to rewrite them.  Even changing pointers to `&`/`&mut` should be forbidden, even though it doesn't change the layout, because the correctness of that change depends on details of the behavior of the libc functions that use the type.

aneksteind:
Fixes #948. Fixes an issue where we would rewrite structs and their fields even though they appear in extern blocks.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -288,7 +288,7 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub adt_metadata: AdtMetadataTable<'tcx>,
 
-    pub foreign_mentioned_tys: HashSet<Ty<'tcx>>,
+    pub foreign_mentioned_tys: HashSet<DefId>,
 }
 
 pub struct AnalysisCtxt<'a, 'tcx> {
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -25,7 +25,7 @@ use rustc_middle::ty::{
     tls, AdtDef, FieldDef, GenericArgKind, GenericParamDefKind, Ty, TyCtxt, TyKind,
 };
 use rustc_type_ir::RegionKind::{ReEarlyBound, ReStatic};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
 use std::ops::Index;
 
@@ -287,6 +287,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
     pub addr_of_static: HashMap<DefId, PointerId>,
 
     pub adt_metadata: AdtMetadataTable<'tcx>,
+
+    pub foreign_mentioned_tys: HashSet<Ty<'tcx>>,
 }
 
 pub struct AnalysisCtxt<'a, 'tcx> {
@@ -536,6 +538,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             static_tys: HashMap::new(),
             addr_of_static: HashMap::new(),
             adt_metadata: construct_adt_metadata(tcx),
+            foreign_mentioned_tys: HashSet::new(),
         }
     }
 
@@ -582,6 +585,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             ref mut static_tys,
             ref mut addr_of_static,
             adt_metadata: _,
+            foreign_mentioned_tys: _,
         } = *self;
 
         *ptr_info = remap_global_ptr_info(ptr_info, map, counter.num_pointers());
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -280,6 +280,9 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
 
 /// Walks the type `ty` and applies a function `f` to it if it's an ADT
 /// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
+/// If `f` returns false, recursion terminates.
+/// We only look for ADTs rather than other FFI-crossing types because ADTs
+/// are the only nominal ones, which are the ones that we may rewrite.
 fn walk_args_and_fields<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
 where
     F: FnMut(DefId) -> bool,
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -303,7 +303,9 @@ fn foreign_mentioned_tys(tcx: TyCtxt) -> HashSet<DefId> {
 
 /// Walks the type `ty` and applies a function `f` to it if it's an ADT
 /// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
-/// If `f` returns false, recursion terminates.
+/// If `f` returns false, recursion terminates cases where the ADT is recursive,
+/// otherwise the function will naturally terminate given that the generic types
+/// of a type are finite in length.
 /// We only look for ADTs rather than other FFI-crossing types because ADTs
 /// are the only nominal ones, which are the ones that we may rewrite.
 fn walk_ffi_safe_ty<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -295,18 +295,27 @@ where
         }
     }
 
-    if let TyKind::Adt(adt_def, _) = ty.kind() {
-        if !f(adt_def.did()) {
-            return;
-        }
-        for field in adt_def.all_fields() {
-            let field_ty = tcx.type_of(field.did);
-            for arg in field_ty.walk() {
-                if let GenericArgKind::Type(ty) = arg.unpack() {
-                    walk_args_and_fields(tcx, ty, f)
+    match ty.kind() {
+        TyKind::Adt(adt_def, _) => {
+            if !f(adt_def.did()) {
+                return;
+            }
+            for field in adt_def.all_fields() {
+                let field_ty = tcx.type_of(field.did);
+                for arg in field_ty.walk() {
+                    if let GenericArgKind::Type(ty) = arg.unpack() {
+                        walk_args_and_fields(tcx, ty, f)
+                    }
                 }
             }
         }
+        TyKind::FnPtr(sig) => {
+            let sig = tcx.erase_late_bound_regions(*sig);
+            for ty in sig.inputs_and_output {
+                walk_args_and_fields(tcx, ty, f)
+            }
+        }
+        _ => (),
     }
 }
 
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -537,16 +537,15 @@ fn run(tcx: TyCtxt) {
     // FIX the fields of structs mentioned in extern blocks
     for adt_did in &gacx.adt_metadata.struct_dids {
         if gacx.foreign_mentioned_tys.contains(adt_did) {
-            if let TyKind::Adt(adt_def, _) = tcx.type_of(adt_did).kind() {
-                let fields = adt_def.all_fields();
-                for field in fields {
-                    let field_lty = gacx.field_ltys[&field.did];
-                    eprintln!(
-                        "adding FIXED permission for {adt_did:?} field {:?}",
-                        field.did
-                    );
-                    make_ty_fixed(&mut gasn, field_lty);
-                }
+            let adt_def = tcx.adt_def(adt_did);
+            let fields = adt_def.all_fields();
+            for field in fields {
+                let field_lty = gacx.field_ltys[&field.did];
+                eprintln!(
+                    "adding FIXED permission for {adt_did:?} field {:?}",
+                    field.did
+                );
+                make_ty_fixed(&mut gasn, field_lty);
             }
         }
     }
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -278,8 +278,8 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
     }
 }
 
-// Walks the type `ty` and applies a function `f` to it if it's an ADT
-// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
+/// Walks the type `ty` and applies a function `f` to it if it's an ADT
+/// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
 fn walk_args_and_fields<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
 where
     F: FnMut(DefId) -> bool,
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -294,10 +294,9 @@ where
         if !f(adt_def.did()) {
             return;
         }
-        let fields = adt_def.all_fields();
-        for field in fields {
-            let field_lty = tcx.type_of(field.did);
-            for arg in field_lty.walk() {
+        for field in adt_def.all_fields() {
+            let field_ty = tcx.type_of(field.did);
+            for arg in field_ty.walk() {
                 if let GenericArgKind::Type(ty) = arg.unpack() {
                     walk_args_and_fields(tcx, ty, f)
                 }
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -34,7 +34,7 @@ use rustc_middle::mir::{
     AggregateKind, BindingForm, Body, Constant, LocalDecl, LocalInfo, LocalKind, Location, Operand,
     Rvalue, StatementKind,
 };
-use rustc_middle::ty::{Ty, TyCtxt, TyKind, WithOptConstParam};
+use rustc_middle::ty::{GenericArgKind, Ty, TyCtxt, TyKind, WithOptConstParam};
 use rustc_span::Span;
 use std::collections::{HashMap, HashSet};
 use std::env;
@@ -278,6 +278,32 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
     }
 }
 
+// Walks the type `ty` and applies a function `f` to it if it's an ADT
+// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
+fn walk_args_and_fields<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
+where
+    F: FnMut(DefId),
+{
+    for arg in ty.walk().skip(1) {
+        if let GenericArgKind::Type(ty) = arg.unpack() {
+            walk_args_and_fields(tcx, ty, f);
+        }
+    }
+
+    if let TyKind::Adt(adt_def, _) = ty.kind() {
+        f(adt_def.did());
+        let fields = &adt_def.non_enum_variant().fields;
+        for field in fields {
+            let field_lty = tcx.type_of(field.did);
+            for arg in field_lty.walk() {
+                if let GenericArgKind::Type(ty) = arg.unpack() {
+                    walk_args_and_fields(tcx, ty, f)
+                }
+            }
+        }
+    }
+}
+
 fn run(tcx: TyCtxt) {
     let mut gacx = GlobalAnalysisCtxt::new(tcx);
     let mut func_info = HashMap::new();
@@ -478,18 +504,23 @@ fn run(tcx: TyCtxt) {
     // don't want to rewrite those
     let mut foreign_mentioned_tys = HashSet::new();
     for item in tcx.hir_crate_items(()).foreign_items() {
-        let ldid = item.def_id.to_def_id();
-        match tcx.def_kind(ldid) {
+        let did = item.def_id.to_def_id();
+        match tcx.def_kind(did) {
             DefKind::Fn | DefKind::AssocFn => {
-                let sig = tcx.fn_sig(ldid);
+                let sig = tcx.fn_sig(did);
                 let sig = tcx.erase_late_bound_regions(sig);
                 for ty in sig.inputs_and_output {
-                    foreign_mentioned_tys.insert(ty);
+                    walk_args_and_fields(tcx, ty, &mut |did| {
+                        foreign_mentioned_tys.insert(did);
+                    });
                 }
             }
             DefKind::Static(_) => {
-                eprintln!("adding static def {ldid:?}");
-                foreign_mentioned_tys.insert(tcx.type_of(ldid));
+                eprintln!("adding static def {did:?}");
+                let ty = tcx.type_of(did);
+                walk_args_and_fields(tcx, ty, &mut |did| {
+                    foreign_mentioned_tys.insert(did);
+                });
             }
             _ => continue,
         }
@@ -506,8 +537,7 @@ fn run(tcx: TyCtxt) {
 
     // FIX the fields of structs mentioned in extern blocks
     for adt_did in &gacx.adt_metadata.struct_dids {
-        let ty = &tcx.type_of(adt_did);
-        if gacx.foreign_mentioned_tys.contains(ty) {
+        if gacx.foreign_mentioned_tys.contains(adt_did) {
             if let TyKind::Adt(adt_def, _) = tcx.type_of(adt_did).kind() {
                 let fields = &adt_def.non_enum_variant().fields;
                 for field in fields {
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -474,11 +474,47 @@ fn run(tcx: TyCtxt) {
                 || info.contains(PointerInfo::NOT_TEMPORARY_REF))
     }
 
+    // track all types mentioned in extern blocks, we
+    // don't want to rewrite those
+    let mut foreign_mentioned_tys = HashSet::new();
+    for item in tcx.hir_crate_items(()).foreign_items() {
+        let ldid = item.def_id.to_def_id();
+        match tcx.def_kind(ldid) {
+            DefKind::Fn | DefKind::AssocFn => {
+                let sig = tcx.fn_sig(ldid);
+                let sig = tcx.erase_late_bound_regions(sig);
+                for ty in sig.inputs_and_output {
+                    foreign_mentioned_tys.insert(ty);
+                }
+            }
+            _ => continue,
+        }
+    }
+    gacx.foreign_mentioned_tys = foreign_mentioned_tys;
+
     let mut gasn =
         GlobalAssignment::new(gacx.num_pointers(), PermissionSet::UNIQUE, FlagSet::empty());
     for (ptr, &info) in gacx.ptr_info().iter() {
         if should_make_fixed(info) {
             gasn.flags[ptr].insert(FlagSet::FIXED);
+        }
+    }
+
+    // FIX the fields of structs mentioned in extern blocks
+    for adt_did in &gacx.adt_metadata.struct_dids {
+        let ty = &tcx.type_of(adt_did);
+        if gacx.foreign_mentioned_tys.contains(ty) {
+            if let TyKind::Adt(adt_def, _) = tcx.type_of(adt_did).kind() {
+                let fields = &adt_def.non_enum_variant().fields;
+                for field in fields {
+                    let field_lty = gacx.field_ltys[&field.did];
+                    eprintln!(
+                        "adding FIX permission for {adt_did:?} field {:?}",
+                        field.did
+                    );
+                    make_ty_fixed(&mut gasn, field_lty);
+                }
+            }
         }
     }
 
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -290,7 +290,7 @@ fn foreign_mentioned_tys(tcx: TyCtxt) -> HashSet<DefId> {
             _ => None,
         })
     {
-        walk_args_and_fields(tcx, ty, &mut |did| foreign_mentioned_tys.insert(did));
+        walk_adts(tcx, ty, &mut |did| foreign_mentioned_tys.insert(did));
     }
     foreign_mentioned_tys
 }
@@ -302,7 +302,7 @@ fn foreign_mentioned_tys(tcx: TyCtxt) -> HashSet<DefId> {
 /// of a type are finite in length.
 /// We only look for ADTs rather than other FFI-crossing types because ADTs
 /// are the only nominal ones, which are the ones that we may rewrite.
-fn walk_args_and_fields<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
+fn walk_adts<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
 where
     F: FnMut(DefId) -> bool,
 {
@@ -310,7 +310,7 @@ where
     // relevant handling for that is below, so we can skip it here
     for arg in ty.walk().skip(1) {
         if let GenericArgKind::Type(ty) = arg.unpack() {
-            walk_args_and_fields(tcx, ty, f);
+            walk_adts(tcx, ty, f);
         }
     }
 
@@ -320,7 +320,7 @@ where
         }
         for field in adt_def.all_fields() {
             let field_ty = tcx.type_of(field.did);
-            walk_args_and_fields(tcx, field_ty, f);
+            walk_adts(tcx, field_ty, f);
         }
     }
 }
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -287,13 +287,13 @@ fn foreign_mentioned_tys(tcx: TyCtxt) -> HashSet<DefId> {
                 let sig = tcx.fn_sig(did);
                 let sig = tcx.erase_late_bound_regions(sig);
                 for ty in sig.inputs_and_output {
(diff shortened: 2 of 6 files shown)


kkysen:
Also, I changed "addresses" to "fixes" as I don't think "addresses" is [one of the keywords for closing an issue from a PR](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

c2rust-analyze/tests/filecheck/foreign.rs
```diff
@@ -0,0 +1,25 @@
+extern "Rust" {
```
kkysen, line 1: Don't we want to mainly test `extern "C"`?

c2rust-analyze/src/main.rs
```diff
@@ -482,6 +500,24 @@ fn run(tcx: TyCtxt) {
...
+    // FIX the fields of structs mentioned in extern blocks
+    for adt_did in &gacx.adt_metadata.struct_dids {
+        let ty = &tcx.type_of(adt_did);
+        if gacx.foreign_mentioned_tys.contains(ty) {
+            if let TyKind::Adt(adt_def, _) = tcx.type_of(adt_did).kind() {
+                let fields = &adt_def.non_enum_variant().fields;
+                for field in fields {
+                    let field_lty = gacx.field_ltys[&field.did];
+                    eprintln!(
+                        "adding FIX permission for {adt_did:?} field {:?}",
```
kkysen, line 512: Isn't it `FIXED`?

c2rust-analyze/src/main.rs
```diff
@@ -474,6 +474,24 @@ fn run(tcx: TyCtxt) {
                 || info.contains(PointerInfo::NOT_TEMPORARY_REF))
     }
 
+    // track all types mentioned in extern blocks, we
+    // don't want to rewrite those
+    let mut foreign_mentioned_tys = HashSet::new();
+    for item in tcx.hir_crate_items(()).foreign_items() {
+        let ldid = item.def_id.to_def_id();
+        match tcx.def_kind(ldid) {
+            DefKind::Fn | DefKind::AssocFn => {
```
kkysen, line 483: Why are we only looking at functions?  What about statics, for example, or whatever else there may be?

spernsteiner (commented):
Looks like a good approach overall

c2rust-analyze/src/context.rs
```diff
@@ -287,6 +287,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
     pub addr_of_static: HashMap<DefId, PointerId>,
 
     pub adt_metadata: AdtMetadataTable<'tcx>,
+
+    pub foreign_mentioned_tys: HashSet<Ty<'tcx>>,
```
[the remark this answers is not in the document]
aneksteind: I feel like the current name is fine

c2rust-analyze/tests/filecheck/foreign.rs
aneksteind: Can i have that and still put the Rust types as arguments? I didnt think one type of extern block or another would make a difference for this feature

c2rust-analyze/src/main.rs
aneksteind: I thought about statics but I wasnt sure, could you give an example of a declaration item in the extern block youre thinking of and I can add support for it?

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -473,6 +473,15 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for HirTyVisitor<'a, 'tcx> {
...
             ItemKind::Struct(VariantData::Struct(field_defs, _), generics) => {
+                if self
+                    .acx
+                    .gacx
+                    .foreign_mentioned_tys
+                    .contains(&self.acx.tcx().type_of(did))
+                {
+                    eprintln!("Avoiding rewrite for foreign-mentioned type: {did:?}");
+                    return;
+                }
```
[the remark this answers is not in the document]
aneksteind: It’s needed to avoid adding hypothetical lifetimes

c2rust-analyze/src/main.rs
spernsteiner: Here's an example involving an extern static with struct type:

```C
struct S {
    int x;
    int y;
};

extern S s;

int f() { return s.x; }
```

```Rust
extern "C" {
    static mut s: S;
}
#[derive(Copy, Clone)]
#[repr(C)]
pub struct S {
    pub x: libc::c_int,
    pub y: libc::c_int,
}
#[no_mangle]
pub unsafe extern "C" fn f() -> libc::c_int {
    return s.x;
}
```

c2rust-analyze/src/rewrite/ty.rs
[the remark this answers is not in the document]
spernsteiner: Ah, I see - this check is not running on the field types, but rather on mentions of the struct itself.

At some point we'll need to modify the rewriting of hypothetical lifetimes to handle cases where a pointer in the original struct becomes something other than a reference (such as `Box<T>`, or in this case being left as `*mut T`), but this should work well enough for now.

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
kkysen: As @spernsteiner said in https://github.com/immunant/c2rust/pull/960#discussion_r1229915402, this should probably be computed recursively.  What we want to be `FIXED` is any type that ever crosses an FFI boundary, not just types that are used in foreign items (shallowly).  `rustc` generally uses "foreign" to mean "in an `extern` block", but this should include types that are not in `extern` blocks as well, as they still cross an FFI boundary.

c2rust-analyze/src/main.rs
spernsteiner: Extern types don't have any rewritable parts, so we don't actually need to do any work in that case.

c2rust-analyze/src/main.rs
```diff
@@ -474,6 +474,24 @@ fn run(tcx: TyCtxt) {
...
+                let sig = tcx.fn_sig(ldid);
+                let sig = tcx.erase_late_bound_regions(sig);
+                for ty in sig.inputs_and_output {
+                    foreign_mentioned_tys.insert(ty);
+                }
```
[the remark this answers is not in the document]
spernsteiner: After reading @kkysen's comment https://github.com/immunant/c2rust/pull/960#discussion_r1229986055, I think this should actually recurse into fields as well (for `#[repr(C)]` structs at minimum).
```Rust
#[repr(C)]
struct Foo {
    x: *mut i32,
    y: *mut i32,
}
#[repr(C)]
struct Bar {
    foo: *mut Foo,
}
extern "C" {
    fn f(bar: *mut Bar);
}
```
Here, it's possible the C code for `f` might access `bar->foo->y`, in which case it relies on the layout of `Foo`, even though `Foo` itself isn't mentioned in any `extern` signatures.  So all the fields of `Foo` would need to be FIXED in this case.

c2rust-analyze/src/main.rs
kkysen: Yeah, that's true.  I think they'd also be mentioned in `extern` `fn`s` and `static`s anyways.

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
kkysen: Does `rustc` have a `fn` to do this already?  As `rustc` checks (as a warning) to make sure your FFI-crossing types are FFI-safe.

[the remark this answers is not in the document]
spernsteiner: I'm not sure.  It doesn't look like `TypeFolder` will walk into fields, at least.  It might be worth looking at the code for the FFI-safety lint to see if it's built on some kind of reusable framework.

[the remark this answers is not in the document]
kkysen: > Does `rustc` have a `fn` to do this already? As `rustc` checks (as a warning) to make sure your FFI-crossing types are FFI-safe.

It doesn't seem so, at least not in a way neatly separated from the lint.

c2rust-analyze/src/main.rs
```diff
@@ -482,6 +500,24 @@ fn run(tcx: TyCtxt) {
         }
     }
 
+    // FIX the fields of structs mentioned in extern blocks
+    for adt_did in &gacx.adt_metadata.struct_dids {
+        let ty = &tcx.type_of(adt_did);
+        if gacx.foreign_mentioned_tys.contains(ty) {
```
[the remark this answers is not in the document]
kkysen: The `Ty`s at this stage should still be generic, right?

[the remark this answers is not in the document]
spernsteiner: The `Ty` produced by `tcx.type_of(adt_did)` will be generic.  The `Ty`s in the `foreign_mentioned_tys` set won't be, since they come from actual uses of the struct.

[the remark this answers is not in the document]
kkysen: Ah, okay.  Then I agree that using `DefId`s seems smarter here.

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
aneksteind: besides repr C or extern blocks what would constitute crossing an FFI boundary?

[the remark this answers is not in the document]
kkysen: `#[repr(C)]` doesn't necessarily mean it should be `FIXED`, and all types from `c2rust transpile` will be `#[repr(C)]`.  `#[repr(C)]` still allows more complex, safe types, so we want to rewrite those, just not the ones that actually cross the FFI boundary.  `extern` blocks indicate they will cross the FFI boundary as an import, and as for exports, I think `extern "C"`, too.  The problem is that all `fn`s from `c2rust transpile` will be marked `extern "C"`.  Maybe a better way to detect exports is `#[no_mangle]`.  I'm not sure if mangle `fn`s will ever be called over an FFI boundary or if Rust's name mangling is stable and you are allowed to do that or not.  If something like [the `#[export]` RFC](https://github.com/rust-lang/rfcs/pull/3435) lands, that might be good, but it hasn't even been accepted as an RFC much less implemented and stabilized yet.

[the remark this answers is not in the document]
aneksteind: Given the ambiguity here and the way c2rust currently transpiles, I’m just going to focus on the extern imports and recursing properly and creating issues for the rest.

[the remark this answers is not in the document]
spernsteiner: > Maybe a better way to detect exports is `#[no_mangle]` (rather, `#[no_mangle]` and `extern` non-`Rust`).

I think we should ignore the library use case for now and assume that Rust functions are never called from C.  `c2rust-transpile` makes all non-`static` C functions into `#[no_mangle]` Rust functions, so there are a lot of functions in lighttpd that are `#[no_mangle]` just because they're called from a different compilation unit.

> I'm not sure if mangle `fn`s will ever be called over an FFI boundary or if Rust's name mangling is stable and you are allowed to do that or not.

I think we should assume this doesn't happen (regardless of what we decide on the previous point), even though it's technically possible and might even be well-defined.

>besides repr C or extern blocks what would constitute crossing an FFI boundary?

One way to formulate this is in terms of typed expressions/declarations:

* All arguments and return values of `extern`-block `fn`s are FFI-exposed.
* All `extern`-block `static`s are FFI-exposed.
* If `x: *const T` or `x: *mut T` is FFI-exposed, then `*x: T` is also FFI-exposed.
* If `s: S` is FFI-exposed, `S` is a struct type, and `struct S` is declared with `#[repr(C)]`, then each field of `S` is FFI-exposed.
* If `x: T` is FFI-exposed, then all pointers that appear in `T` must be marked `FIXED`.  ("A pointer appears in T" roughly means that if you print `T`, you will see the pointer.  Two pointers appear in `*mut *const i32`; no pointers appear in the type `MyStruct`, regardless of what fields it has.)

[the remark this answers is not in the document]
kkysen: > I think we should ignore the library use case for now and assume that Rust functions are never called from C. `c2rust-transpile` makes all non-`static` C functions into `#[no_mangle]` Rust functions, so there are a lot of functions in lighttpd that are `#[no_mangle]` just because they're called from a different compilation unit.

I forgot most are `#[no_mangle]`, too.  I agree, then, we should just ignore the export case.  It may not be unreasonable to require users to mark their exports with something like `#[c2rust_analyze_test::fixed_signature]` or `#[c2rust_analyze::export]`.

> * If `x: *const T` or `x: *mut T` is FFI-exposed, then `*x: T` is also FFI-exposed.

I'm not sure what `*x: T` means here.

> * If `s: S` is FFI-exposed, `S` is a struct type, and `struct S` is declared with `#[repr(C)]`, then each field of `S` is FFI-exposed.

Shouldn't this apply to all compound types?  We could do `struct`s only for now if it's simpler, but shouldn't we do it for all compound types, i.e. everything `rustc`'s FFI-safety lint recurses into?

Also, I don't think whether or not the type is `#[repr(C)]` matters.  The field types are still contained in the `struct` and cross the FFI boundary all the same.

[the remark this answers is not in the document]
spernsteiner: > Shouldn't this apply to all compound types? We could do `struct`s only for now if it's simpler, but shouldn't we do it for all compound types, i.e. everything `rustc`'s FFI-safety lint recurses into?

Oh, yes, I suppose we need similar rules for arrays and for unions.

Are there any other FFI-safe aggregates?  I know tuples are always implicitly `#[repr(Rust)]`.

> Also, I don't think whether or not the type is `#[repr(C)]` matters. The field types are still contained in the `struct` and cross the FFI boundary all the same.

If the struct isn't either `#[repr(C)]` or #[repr(transparent)]`, then there are no layout guarantees, and the C code can't (or at least really shouldn't) access any fields.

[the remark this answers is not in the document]
kkysen: > Are there any other FFI-safe aggregates? I know tuples are always implicitly `#[repr(Rust)]`.

From [`ImproperCTypesVisitor::check_type_for_ffi`](https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_lint/types.rs.html#893), it seems that we should recurse into:
* `struct`s
* `union`s
* `enum`s
* references
* raw pointers
* arrays
* `fn` ptrs

Do we already treat non-crate-local types as `FIXED`?

c2rust-analyze/tests/filecheck/foreign.rs
aneksteind: aa23b35d

kkysen: Can you make `Bar` and `Baz` `#[repr(C)]`?

c2rust-analyze/src/main.rs
```diff
@@ -482,6 +500,24 @@ fn run(tcx: TyCtxt) {
...
+            if let TyKind::Adt(adt_def, _) = tcx.type_of(adt_did).kind() {
```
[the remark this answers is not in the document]
aneksteind: Are any other types expected to be supported?

c2rust-analyze/src/main.rs
aneksteind: Added in 9f35bbfe

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
kkysen: Yeah, so this suggestion doesn't really apply/is useful anymore.  See https://github.com/immunant/c2rust/pull/960/files/b90b6465c179fb8b5913d3b09fa5c17b220c3c1d#r1230230044 for which types to recurse into.

c2rust-analyze/src/main.rs
aneksteind: e7890392

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: 9f35bbf

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: How would I get `DefId`s of function signature inputs and outputs, which are an iterator over types, to add those to the set?

c2rust-analyze/src/rewrite/ty.rs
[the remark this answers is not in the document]
aneksteind: Added #961

c2rust-analyze/tests/filecheck/foreign.rs
aneksteind: they are in 9f35bbf

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
spernsteiner: For each arg/return/field type, I think you'd want to walk over it (maybe with [`Ty::walk`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.walk) or `LTy::iter` or a custom recursive function) and do something with all the `TyKind::Adt` `DefId`s you see.  You could potentially also look for `PointerId`s and set their FIXED flags in the same traversal, which might be simpler depending on how exactly you implement the traversal of struct fields.

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
spernsteiner: > Do we already treat non-crate-local types as `FIXED`?

Roughly, yes.  IIRC we don't even assign `PointerId`s to fields of non-local types, since we can't rewrite them.

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: There are other comments outlining the need for recursive visitation, so closing this comment and will close the others once recursive visitation is supported.

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: Thanks, that looks very useful

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
kkysen: Okay, so for this recursion, we should make sure to stop eagerly at non-crate-local types.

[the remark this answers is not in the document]
aneksteind: Let me know if 795438ae doesn't accomplish what's desired

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: Walking in 795438ae

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: Walking in 795438ae

c2rust-analyze/src/main.rs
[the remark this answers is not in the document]
aneksteind: So AFAIK ADTs are the only types that we recurse into that we want to avoid rewriting explicitly. I believe the rest are at least recursed into in 795438ae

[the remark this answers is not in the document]
kkysen: Ah, okay.  That makes sense, as ADTs are all the nominal (so not aliases) aggregate types that we'd rewrite.  In that case, you can use `tcx.adt_def` here again if it applies.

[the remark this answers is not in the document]
aneksteind: addressed in f15d8041

c2rust-analyze/src/main.rs
```diff
@@ -278,6 +278,35 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
     }
 }
 
+// Walks the type `ty` and applies a function `f` to it if it's an ADT
+// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
```
kkysen, line 282:
```suggestion
/// Walks the type `ty` and applies a function `f` to it if it's an ADT
/// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
```

kkysen: Could you also document what the return value of `f` means?

c2rust-analyze/src/main.rs
```diff
@@ -278,6 +278,35 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
...
+where
+    F: FnMut(DefId) -> bool,
+{
+    // the first type encountered in the walk is `ty`, and the
+    // relevant handling for that is below, so we can skip it here
+    for arg in ty.walk().skip(1) {
+        if let GenericArgKind::Type(ty) = arg.unpack() {
+            walk_args_and_fields(tcx, ty, f);
+        }
+    }
```
kkysen, line 293: I don't think we want to walk the generics.  A type could be a generic type parameter but not actually be in the type layout (e.x. `PhantomData`).

c2rust-analyze/src/main.rs
```diff
@@ -278,6 +278,35 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
...
+        }
+        for field in adt_def.all_fields() {
+            let field_ty = tcx.type_of(field.did);
+            for arg in field_ty.walk() {
+                if let GenericArgKind::Type(ty) = arg.unpack() {
+                    walk_args_and_fields(tcx, ty, f)
+                }
+            }
+        }
+    }
```
kkysen, line 307: I'm not sure the generic walking here is correct.  From what I understand, this is what we want to do:

1. If `ty` is an ADT, call `f` on its `DefId`, which represents the generic ADT.
2. Substitute the ADTs generic args into the ADT.
3. Then recurse on each of its fields.

c2rust-analyze/src/main.rs
```diff
@@ -278,6 +278,35 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
     }
 }
 
+// Walks the type `ty` and applies a function `f` to it if it's an ADT
+// `f` gets applied recursively to `ty`'s generic types and fields (if applicable)
+fn walk_args_and_fields<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
```
kkysen, line 283: I think `walk_adts` would be a better name.  `f` is called on all ADTs found.

c2rust-analyze/src/main.rs
```diff
@@ -474,6 +503,29 @@ fn run(tcx: TyCtxt) {
...
+    let mut foreign_mentioned_tys = HashSet::new();
+    for item in tcx.hir_crate_items(()).foreign_items() {
+        let did = item.def_id.to_def_id();
+        match tcx.def_kind(did) {
+            DefKind::Fn | DefKind::AssocFn => {
+                let sig = tcx.fn_sig(did);
+                let sig = tcx.erase_late_bound_regions(sig);
+                for ty in sig.inputs_and_output {
+                    walk_args_and_fields(tcx, ty, &mut |did| foreign_mentioned_tys.insert(did));
+                }
```
kkysen, line 517: Once the above comments are addressed an `walk_ffi_safe_adts` handles `FnPtr`s, you could do

```rust
walk_ffi_safe_adts(tcx, TyKind::FnPtr(sig), &mut |did| foreign_mentioned_tys.insert(did));
```

c2rust-analyze/src/main.rs
```diff
@@ -474,6 +503,29 @@ fn run(tcx: TyCtxt) {
                 || info.contains(PointerInfo::NOT_TEMPORARY_REF))
     }
 
+    // track all types mentioned in extern blocks, we
+    // don't want to rewrite those
+    let mut foreign_mentioned_tys = HashSet::new();
```
kkysen, line 508: Could you move this block calculating `foreign_mentioned_tys` to its own `fn`?  `fn run` is big enough already.

c2rust-analyze/src/main.rs
spernsteiner: This loop also walks array elements, pointees of raw pointers, and such, which we don't want to ignore.

c2rust-analyze/src/main.rs
kkysen: Also, once the other changes are applied, I think you could simplify it a lot more with some iterator methods:

```rust
for ty in tcx.hir_crate_items(())
    .foreign_items()
    .map(|item| item.def_id.to_def_id())
    .filter_map(|did| match tcx.def_kind(did) {
        DefKind::Fn | DefKind::AssocFn => Some(TyKind::FnPtr(tcx.fn_sig(did))),
        DefKind::Static(_) => Some(tcx.type_of(did)),
        _ => None,
    }) {
    walk_ffi_safe_adts(tcx, ty, &mut |did| foreign_mentioned_tys.insert(did));
}
```

c2rust-analyze/src/main.rs
kkysen: Can you also add that we only look for ADTs rather than other FFI-crossing types because ADTs are the only nominal ones, which are the ones that we may rewrite?

c2rust-analyze/src/main.rs
spernsteiner: > a. Substitute the ADTs generic args into the ADT.
> b. Then recurse on each of its fields.

That's overkill IMO and makes things significantly more complex.  We would need to keep a separate set of visited `Ty`s in order to avoid infinite recursion without potentially skipping some structs, whereas currently we reuse the output set (a set of `DefId`s) for this instead.

I think we should walk over generics (at least in the ADT case, if we go ahead with explicit matching on the `TyKind` instead of using `walk`), which gives us an overapproximation of reachable types but should be accurate in the vast majority of cases.

kkysen: > That's overkill IMO and makes things significantly more complex. We would need to keep a separate set of visited `Ty`s in order to avoid infinite recursion without potentially skipping some structs, whereas currently we reuse the output set (a set of `DefId`s) for this instead.

Why doesn't the `f(adt_def.did())` check prevent infinite recursion?  The current way is already recursing into the ADT's fields.

aneksteind:
@spernsteiner I’m not clear on what further changes you’d like to see. Could you let me know if I’ve missed any concerns of yours?

c2rust-analyze/src/main.rs
aneksteind: a2e99a57

c2rust-analyze/src/main.rs
aneksteind: Given [this comment](https://github.com/immunant/c2rust/pull/960#discussion_r1231639983), marking resolved

c2rust-analyze/src/main.rs
aneksteind: I don't see how I can call `walk` on a `TyKind`, I think things are fine as they are here

c2rust-analyze/src/main.rs
aneksteind: moved to it's own function in d0a6aac9. `tcx.type_of` returns a `Ty` and that's not a `TyKind` so your suggestion will not work.

c2rust-analyze/src/main.rs
aneksteind: I renamed to `walk_ffi_safe_ty` i  88472f72 now that it processes `FnPtr` too

c2rust-analyze/src/main.rs
aneksteind: Now that this is also walking a `FnPtr` I've renamed to `walk_ffi_safe_ty` in 88472f72

c2rust-analyze/src/main.rs
kkysen: Oh, oops.  You could do `tcx.mk_fn_ptr(sig)` instead of the `TYKind`.

c2rust-analyze/src/main.rs
aneksteind: Btw `if !f(adt_def.did())` does prevent infinite recursion in the case that the ADT is recursive (mutually or not)`. I added documentation in 79502f74

aneksteind: `walk` does seem to walk `FnPtr`s anyway, so I'd like to leave the function as-is

```rust
            ty::FnPtr(sig) => {
                stack.push(sig.skip_binder().output().into());
                stack.extend(sig.skip_binder().inputs().iter().copied().rev().map(|ty| ty.into()));
            }
```

c2rust-analyze/src/main.rs
aneksteind: It does handle `FnPtrs`, see https://github.com/immunant/c2rust/pull/960#discussion_r1232403470

c2rust-analyze/src/main.rs
aneksteind: The way `walk` walks types is kind of misleading in the name. It implicitly casts types and consts like array lengths to `GenericArgs` for convenience. It still iterates over all reachable types regardless of whether or not the item is a generic arg. For `FnPtrs` it iterates over the inputs and outputs

c2rust-analyze/src/main.rs
aneksteind: see https://github.com/immunant/c2rust/pull/960#discussion_r1232411197 for explanation of why walk deals with GenericArgs

c2rust-analyze/src/main.rs
kkysen: I updated the above snippet so they're both `Ty`s now.

aneksteind: Added your suggestion in 8e341449

c2rust-analyze/src/main.rs
aneksteind: 29fa2691

kkysen (approved):
Just one small nit left: `walk_adts` instead of `walk_adt`, as the latter sounds like it's walking over only one.  Otherwise, it all LGTM now!

c2rust-analyze/src/main.rs
```diff
@@ -278,6 +278,53 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
...
+where
+    F: FnMut(DefId) -> bool,
+{
+    // the first type encountered in the walk is `ty`, and the
+    // relevant handling for that is below, so we can skip it here
+    for arg in ty.walk().skip(1) {
+        if let GenericArgKind::Type(ty) = arg.unpack() {
+            walk_adts(tcx, ty, f);
+        }
+    }
```
spernsteiner, line 315: I think this can be exponential in the size of the type (which is usually small, but still).  Suppose you call `walk_adts(*mut *mut T)`: this loop will call `walk_adts(*mut T)` and `walk_adts(T)`.  Then the call to `walk_adts(*mut T)` will call `walk_adts(T)` a second time.  If `T = *mut *mut U`, then we call `walk_adts(U)` at least four times, and so on.

One easy fix for this would be to do only the `if let TyKind::Adt ...` work instead of recursing here.  For the types we visit in `walk()`, we don't want to walk them again (since the outer walk will already cover their children), we only want to do the `TyKind::Adt` check.

c2rust-analyze/src/main.rs
```diff
@@ -482,6 +533,22 @@ fn run(tcx: TyCtxt) {
         }
     }
 
+    // FIX the fields of structs mentioned in extern blocks
```
spernsteiner, line 536: Are `extern`-block `fn` signatures already `FIXED` elsewhere, or does that still need to be added (in this PR or elsewhere)?

c2rust-analyze/src/main.rs
aneksteind: Is 78fce6ab what you had in mind?

c2rust-analyze/src/main.rs
aneksteind: I didn't think they needed to be marked fixed because I thought only body-owners were visited?

c2rust-analyze/src/main.rs
spernsteiner: Yes, except the `return` should be changed to `continue`

c2rust-analyze/src/main.rs
spernsteiner: Ah - I don't actually remember whether non-body-owners get visited by `rewrite::ty`, but you might be right.  Either way, I think the args/returns should be marked `FIXED` so that `rewrite::expr` on callers comes up with the right `TypeDesc`s for the callee's args and generates the right casts.

c2rust-analyze/src/main.rs
aneksteind: Done

c2rust-analyze/src/main.rs
aneksteind: I'd like to do that in another PR

aneksteind: Created an issue #965

c2rust-analyze/src/main.rs
```diff
@@ -278,6 +278,49 @@ fn update_pointer_info<'tcx>(acx: &mut AnalysisCtxt<'_, 'tcx>, mir: &Body<'tcx>)
...
+/// of a type are finite in length.
+/// We only look for ADTs rather than other FFI-crossing types because ADTs
+/// are the only nominal ones, which are the ones that we may rewrite.
+fn walk_adts<'tcx, F>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, f: &mut F)
+where
+    F: FnMut(DefId) -> bool,
+{
+    for arg in ty.walk() {
+        if let GenericArgKind::Type(ty) = arg.unpack() {
+            if let TyKind::Adt(adt_def, _) = ty.kind() {
```
kkysen, line 311: Why don't we need to substitute generics here?  Then generic fields will just be `TyKind::Param` and we won't recognize that it may be an ADT or other type we need to recurse into.

kkysen: Oh wait, that should be covered by `.walk` instead, right?

spernsteiner (approved):

PR #962, opened 2023-06-15
Fix the permissions of some scripts

mjgarton:
This makes `provision_dnf.sh` and `provision_yum.sh` executable.


kkysen (approved):
I'm not really sure how this was getting run here, https://github.com/immunant/c2rust/blob/a4ab0bc75acf6bb24b11567d928874b9da8b3e25/scripts/provision.sh#L48, if it wasn't executable, but it apparently does.  Either way, these files should be executable, so thanks for finding that.

PR #963, opened 2023-06-15
Improve Fedora detection

mjgarton:
Instead of relying on `$NAME`, which has changed in recent releases, use `$ID`, which appears more stable.


kkysen (commented):
Is `$ID` more stable than `$NAME` for just Fedora, or for other distros, too?

mjgarton:
It seems that only Fedora has changed the `NAME` value. The others have all remained the same for a long time.  I ran the following to check a few distros:

kkysen (approved):
That looks pretty thorough.  Using `$ID` for just Fedora makes sense then, and if any others ever change, we can use `$ID` for them, too.

PR #966, opened 2023-06-16
mark extern block function signatures as FIXED

### LINKED ISSUE #965: analyze: mark the inputs and output of foreign, non-body-owners as fixed
Comment [here](https://github.com/immunant/c2rust/pull/960#discussion_r1232592042)

With this change, `rewrite::expr` on callers comes up with the right TypeDescs for the callee's args and generates the right casts.

aneksteind:
Fixes #965. The inputs and outputs of body-owners are already marked as fixed, but the traversal of `all_fn_dids` does not include function declarations in `extern` blocks

### DIFF
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -321,6 +321,46 @@ where
     }
 }
 
+fn mark_foreign_fixed<'tcx>(
+    gacx: &mut GlobalAnalysisCtxt<'tcx>,
+    gasn: &mut GlobalAssignment,
+    tcx: TyCtxt<'tcx>,
+) {
+    // FIX the inputs and outputs of function declarations in extern blocks
+    for did in gacx.foreign_mentioned_tys.clone() {
+        if let DefKind::Fn = tcx.def_kind(did) {
+            let sig = tcx.erase_late_bound_regions(tcx.fn_sig(did));
+            let inputs = sig
+                .inputs()
+                .iter()
+                .map(|&ty| gacx.assign_pointer_ids_with_info(ty, PointerInfo::ANNOTATED))
+                .collect::<Vec<_>>();
+            for input in inputs {
+                make_ty_fixed(gasn, input);
+            }
+
+            let output = gacx.assign_pointer_ids_with_info(sig.output(), PointerInfo::ANNOTATED);
+            make_ty_fixed(gasn, output)
+        }
+    }
+
+    // FIX the fields of structs mentioned in extern blocks
+    for adt_did in &gacx.adt_metadata.struct_dids {
+        if gacx.foreign_mentioned_tys.contains(adt_did) {
+            let adt_def = tcx.adt_def(adt_did);
+            let fields = adt_def.all_fields();
+            for field in fields {
+                let field_lty = gacx.field_ltys[&field.did];
+                eprintln!(
+                    "adding FIXED permission for {adt_did:?} field {:?}",
+                    field.did
+                );
+                make_ty_fixed(gasn, field_lty);
+            }
+        }
+    }
+}
+
 fn run(tcx: TyCtxt) {
     let mut gacx = GlobalAnalysisCtxt::new(tcx);
     let mut func_info = HashMap::new();
@@ -529,21 +569,7 @@ fn run(tcx: TyCtxt) {
         }
     }
 
-    // FIX the fields of structs mentioned in extern blocks
-    for adt_did in &gacx.adt_metadata.struct_dids {
-        if gacx.foreign_mentioned_tys.contains(adt_did) {
-            let adt_def = tcx.adt_def(adt_did);
-            let fields = adt_def.all_fields();
-            for field in fields {
-                let field_lty = gacx.field_ltys[&field.did];
-                eprintln!(
-                    "adding FIXED permission for {adt_did:?} field {:?}",
-                    field.did
-                );
-                make_ty_fixed(&mut gasn, field_lty);
-            }
-        }
-    }
+    mark_foreign_fixed(&mut gacx, &mut gasn, tcx);
 
     for info in func_info.values_mut() {
         let num_pointers = info.acx_data.num_pointers();
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -321,26 +321,35 @@ where
     }
 }
 
+fn gather_foreign_sigs<'tcx>(gacx: &mut GlobalAnalysisCtxt<'tcx>, tcx: TyCtxt<'tcx>) {
+    for did in tcx
+        .hir_crate_items(())
+        .foreign_items()
+        .map(|item| item.def_id.to_def_id())
+        .filter(|did| matches!(tcx.def_kind(did), DefKind::Fn | DefKind::AssocFn))
+    {
+        let sig = tcx.erase_late_bound_regions(tcx.fn_sig(did));
+        let inputs = sig
+            .inputs()
+            .iter()
+            .map(|&ty| gacx.assign_pointer_ids_with_info(ty, PointerInfo::ANNOTATED))
+            .collect::<Vec<_>>();
+        let inputs = gacx.lcx.mk_slice(&inputs);
+        let output = gacx.assign_pointer_ids_with_info(sig.output(), PointerInfo::ANNOTATED);
+        let lsig = LFnSig { inputs, output };
+        gacx.fn_sigs.insert(did, lsig);
+    }
+}
+
 fn mark_foreign_fixed<'tcx>(
     gacx: &mut GlobalAnalysisCtxt<'tcx>,
     gasn: &mut GlobalAssignment,
     tcx: TyCtxt<'tcx>,
 ) {
     // FIX the inputs and outputs of function declarations in extern blocks
-    for did in gacx.foreign_mentioned_tys.clone() {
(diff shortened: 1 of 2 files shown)


c2rust-analyze/src/main.rs
```diff
@@ -321,6 +321,46 @@ where
...
+    tcx: TyCtxt<'tcx>,
+) {
+    // FIX the inputs and outputs of function declarations in extern blocks
+    for did in gacx.foreign_mentioned_tys.clone() {
+        if let DefKind::Fn = tcx.def_kind(did) {
+            let sig = tcx.erase_late_bound_regions(tcx.fn_sig(did));
+            let inputs = sig
+                .inputs()
+                .iter()
+                .map(|&ty| gacx.assign_pointer_ids_with_info(ty, PointerInfo::ANNOTATED))
```
kkysen, line 336: Is it right to reuse `ANNOTATED` for what when we mean `FIXED`?

kkysen: Or is this `ANNOTATED` merely because it's from a `fn` signature, and those must be annotated?

c2rust-analyze/src/main.rs
```diff
@@ -321,6 +321,46 @@ where
...
+                .collect::<Vec<_>>();
+            for input in inputs {
+                make_ty_fixed(gasn, input);
+            }
+
+            let output = gacx.assign_pointer_ids_with_info(sig.output(), PointerInfo::ANNOTATED);
+            make_ty_fixed(gasn, output)
+        }
+    }
```
kkysen, line 345:
```suggestion
    let foreign_mentioned_tys = mem::take(&mut gacx.foreign_mentioned_tys);
    for ty in foreign_mentioned_tys
        .iter()
        .filter(|&did| tcx.def_kind(did) == DefKind::Fn)
        .map(|did| tcx.fn_sig(did))
        .map(|sig| tcx.erase_late_bound_regions(sig))
        .flat_map(|sig| sig.inputs_and_output)
    {
        let lty = gacx.assign_pointer_ids_with_info(ty, PointerInfo::ANNOTATED);
        make_ty_fixed(gasn, lty);
    }
    let _ = mem::replace(&mut gacx.foreign_mentioned_tys, foreign_mentioned_tys);
```

The `mem::take`/`mem::replace` (needs a `use std::mem`) would avoid the need to `.clone()` the `HashSet` while still getting around the borrow checker without refactoring things.  And we shouldn't need to `.collect()` the `inputs` either, I think.

c2rust-analyze/src/main.rs
```diff
@@ -321,6 +321,46 @@ where
...
+
+    // FIX the fields of structs mentioned in extern blocks
+    for adt_did in &gacx.adt_metadata.struct_dids {
```
kkysen, line 348: By the way, it would be easier to review (though in this case, it's a simple change so it's not bad) if moves were separated from actual changes in separate commits.

kkysen (approved):
LGTM once the unneeded allocations are removed and the usage of `PointerInfo::ANNOTATED` is confirmed (I'm assuming my latter thought is right, in which case it's all good).

c2rust-analyze/src/main.rs
```diff
@@ -321,6 +321,46 @@ where
...
+                .inputs()
+                .iter()
+                .map(|&ty| gacx.assign_pointer_ids_with_info(ty, PointerInfo::ANNOTATED))
+                .collect::<Vec<_>>();
+            for input in inputs {
+                make_ty_fixed(gasn, input);
+            }
+
+            let output = gacx.assign_pointer_ids_with_info(sig.output(), PointerInfo::ANNOTATED);
+            make_ty_fixed(gasn, output)
```
spernsteiner, line 343: This loop creates a bunch of `LTy`s and marks them `FIXED` but doesn't store them anywhere, so they have no effect on anything.  `inputs` and `output` should probably be bundled into an `LFnSig` and then stored in either `gacx.fn_sigs` or some new map.

c2rust-analyze/src/main.rs
```diff
@@ -321,6 +321,46 @@ where
...
 }
 
+fn mark_foreign_fixed<'tcx>(
+    gacx: &mut GlobalAnalysisCtxt<'tcx>,
+    gasn: &mut GlobalAssignment,
+    tcx: TyCtxt<'tcx>,
+) {
+    // FIX the inputs and outputs of function declarations in extern blocks
+    for did in gacx.foreign_mentioned_tys.clone() {
+        if let DefKind::Fn = tcx.def_kind(did) {
```
spernsteiner, line 331: Does `foreign_mentioned_tys` really include `DefId`s of functions as well as types?  I thought it only stored the types (in which case we need some other way of visiting `extern`-block `fn`s here).  If `foreign_mentioned_tys` really does store both fns and types, then I think it should be renamed to make that clear.

c2rust-analyze/src/main.rs
spernsteiner: `ANNOTATED` is correct because there's an explicit type annotation in the source code corresponding to these `Ty`s.  Normal `fn`s are similarly marked `ANNOTATED`.

spernsteiner (changes requested):

c2rust-analyze/src/main.rs
kkysen: You're right; it only stores ADTs `DefId`s, not `fn` `DefId`s at all.  We could change this loop to iterating over all foreign items and filtering for just `DefKind::Fn` (and `DefKind::AssocFn`?) like we do in `fn foreign_mentioned_tys`.  Also, given this confusion, I think renaming the `tys` in `foreign_mentioned_tys` to `adts` might be helpful.

c2rust-analyze/src/main.rs
kkysen: `gacx.fn_sigs` just stores `LFnSig`s for body owners, so no body-less `extern fn`s (how would some like a non-default `trait` `fn` be handled, by the way?) are included.  Doesn't that mean we would never rewrite such an `extern fn` in the first place?

spernsteiner: We shouldn't rewrite `extern`-block `fn`s, but we need to retrieve their signatures anywhere they're called.  That's why I was thinking it might be reasonable to put them in `fn_sigs`, alongside the signatures of all the normal `fn`s.

kkysen: Don't we just currently crash on calls to `extern fn`s?  That's why I'm working on fixing with hardcoding permissions for `libc` `fn`s, but that's probably going to work a bit differently from `fn_sigs`.  Right now, and for any `extern fn` we don't hardcode, we'll crash on the call when we try to lookup the `LFnSig` from `fn_sigs`.  I just want to make sure we don't silently do the wrong thing instead of crashing if we include these `extern fn`s in `fn_sigs`.

c2rust-analyze/src/main.rs
aneksteind: No longer necessary with 7f8bddbe

c2rust-analyze/src/main.rs
aneksteind: I've added them to `gacx.fn_sigs` in 7f8bddbe

c2rust-analyze/src/main.rs
aneksteind: You're right, this have been fixed in 7f8bddbe

c2rust-analyze/src/main.rs
spernsteiner: > Don't we just currently crash on calls to `extern fn`s? That's why I'm working on fixing with hardcoding permissions for `libc` `fn`s

Yes - there are two parts to libc function handling:
1. For analysis, we need to know the `READ`/`WRITE`/`OFFSET` permissions on the function signature.  This is the part involving hard-coded permissions.
2. For rewriting, we need to generate casts around calls to libc functions.  Setting `FIXED` on the signature triggers insertion of these casts.



> Right now, and for any `extern fn` we don't hardcode, we'll crash on the call when we try to lookup the `LFnSig` from `fn_sigs`. I just want to make sure we don't silently do the wrong thing instead of crashing if we include these `extern fn`s in `fn_sigs`.

Good point.  Currently `util::ty_callee` checks for `extern`-block `fn`s directly (by looking at the parent `DefKind`), so that part will still work after this change.  But are there other places in the code that assume `extern`-block `fn`s will be absent from `fn_sigs`?

spernsteiner (approved):
Would be nice to have a test for this, but I guess that might be too much trouble when you can't actually call `extern`-block `fn`s yet.

aneksteind:
> Would be nice to have a test for this, but I guess that might be too much trouble when you can't actually call extern-block fns yet.

@spernsteiner how do you want to handle that in the short term, is it worth supporting minimally or in some more principled way? @kkysen didn't you have a WIP for this?

c2rust-analyze/src/main.rs
kkysen: @aneksteind @spernsteiner.  I could move the foreign `fn` sigs to a new map, right?  As long as I handle it during calls (which should be quite simple as I'm already changing that stuff), it should be fine, right?  I think it can simplify things with #980 and #999.

spernsteiner: I'm not sure how that helps, but yeah, you could move them as long as you update all the places where calls are handled (which includes at least `dataflow::type_check` and `rewrite::expr::mir_op` - I don't think `borrowck::type_check` handles calls yet).

PR #968, opened 2023-06-23
add support for emitting casts to and from *mut T and Cell<T>

aneksteind:


kkysen:
The `*mut T` -> `*const T` -> `*const Cell<T>` is safe because `Cell` is `#[repr(transparent)]`, right?

aneksteind:
> The *mut T -> *const T -> *const Cell<T> is safe because Cell is #[repr(transparent)], right?

That sounds correct to me

c2rust-analyze/tests/filecheck/cast.rs
```diff
@@ -16,3 +16,26 @@ pub unsafe fn null_ptr() {
...
+
+#[repr(C)]
+pub struct Foo {
+    y: *mut i32,
+}
+
+extern "C" {
+    // necessary to fix the type of Foo::y
+    fn bar(f: Foo);
+}
```
spernsteiner, line 28: Maybe it would be nice to have a test attribute like `#[fixed_signature]` that can be used on struct fields.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -615,7 +619,13 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
                 }
                 _ => None,
             },
-            Ownership::Cell => None,
+            Ownership::Cell => match to.own {
+                Ownership::RawMut if !early => {
```
spernsteiner, line 623: This case should probably trigger on `Ownership::Raw` as well, since the conversion to `RawMut` can be used as the first step in `&Cell<T> -> *mut T -> *const T`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -634,6 +644,13 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
                     self.emit(RewriteKind::UnsafeCastRawToRef { mutbl: true });
                     Some(Ownership::Mut)
                 }
+                Ownership::Cell if !early => {
+                    self.emit(RewriteKind::CastRawMutToCellPtr {
+                        ty: format!("{:?}", to.pointee_ty),
```
spernsteiner, line 649: Is `Debug` the right thing to use here?  I'm not sure whether we can rely on it producing well-formed Rust syntax.  I know there are some cases in `rewrite::ty` where we have to convert a `ty::Ty` to syntax - how is it handled there?

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -208,18 +208,31 @@ impl<'a, 'tcx> UnlowerVisitor<'a, 'tcx> {
...
+                        .map(|pl| {
+                            if is_var(pl) && self.mir.local_kind(pl.local) == mir::LocalKind::Temp {
+                                MirOriginDesc::LoadFromTemp
+                            } else {
+                                MirOriginDesc::Expr
+                            }
+                        })
+                        .unwrap_or(MirOriginDesc::Expr),
+                    _ => MirOriginDesc::Expr,
+                };
```
spernsteiner, line 232: This duplicates the `op_is_temp` logic from `record_operand` - would it make sense to factor that out?

c2rust-analyze/src/rewrite/expr/unlower.rs
```diff
@@ -208,18 +208,31 @@ impl<'a, 'tcx> UnlowerVisitor<'a, 'tcx> {
...
                 self.record(loc, &[], ex);
                 self.record(loc, &[SubLoc::Dest], pl);
-                self.record(loc, &[SubLoc::Rvalue], rv);
+                self.record_desc(loc, &[SubLoc::Rvalue], ex, desc);
```
spernsteiner, line 235: What's the reason for the change from `rv` to `ex` here?  This seems incorrect.  Using `ex` here says that MIR rewrites attached to the `Rvalue` of this MIR statement should be applied to the whole assignment.  So, for example, if there's a cast on the `Rvalue`, then I think it would get rewritten as `(x = y) as T;` instead of `x = y as T;`.

aneksteind: I think you're right that it's incorrect. Does b56abb41 address the issue?

c2rust-analyze/tests/filecheck/cast.rs
aneksteind: Agreed. I think that should go into a separate PR.

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: Is dd00d8d0 sufficient?

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: I've matched what's in `rewrite::ty` with ca6bf033

c2rust-analyze/src/rewrite/expr/unlower.rs
aneksteind: Added in 8172666f

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Yes, that's what I was thinking

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I was thinking more like `fn op_is_temp(&self, op: &Operand<'tcx>) -> bool`, or maybe `fn operand_desc(&self, op: &Operand<'tcx>) -> MirOriginDesc`.  I'm not a fan of the `get_desc` from 8172666 because I don't think its logic makes sense outside of the specific context of `op_is_temp`.  For example, when looking at a `StatementKind::Assign` that stores into a temporary, calling `get_desc` on the LHS `Place` would produce `LoadFromTemp`, which seems wrong.

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: Yes, that looks better.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -232,6 +228,11 @@ fn materialize_adjustments<'tcx>(
             hir_rw
         }
         (rw @ Rewrite::Ref(..), &[Adjust::Deref(..), Adjust::Borrow(..)]) => rw,
+        (ref rw @ Rewrite::MethodCall(ref s, ..), &[Adjust::Deref(..), Adjust::Borrow(..)])
+            if s == "as_ptr" =>
+        {
+            rw.clone()
+        }
```
spernsteiner, line 235: Tying this to the method name `as_ptr` seems weird.  I think we could allow any method name here, since method calls will automatically reborrow (`Adjust::Deref` + `Adjust::Borrow`) when needed.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -697,6 +712,13 @@ where
                     (self.emit)(RewriteKind::UnsafeCastRawToRef { mutbl: true });
                     Some(Ownership::Mut)
                 }
+                Ownership::Cell if !early => {
+                    let printer = FmtPrinter::new(self.tcx, Namespace::TypeNS);
+                    let ty = to.pointee_ty.print(printer).unwrap().into_buffer();
+                    (self.emit)(RewriteKind::CastRawMutToCellPtr { ty });
+                    (self.emit)(RewriteKind::MutToImm);
```
spernsteiner, line 719: This should be an `UnsafeCastRawToRef { mutbl: false }` instead.  `MutToImm` is for `&mut T -> &T`, which is not what we have here (though it happens to generate code that works for this case).

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -18,10 +19,14 @@ use rustc_middle::mir::{
     BasicBlock, Body, Location, Operand, Place, Rvalue, Statement, StatementKind, Terminator,
     TerminatorKind,
 };
+use rustc_middle::ty::print::FmtPrinter;
 use rustc_middle::ty::{Ty, TyCtxt, TyKind};
 use std::collections::HashMap;
 use std::ops::Index;
 
+extern crate rustc_resolve;
+use rustc_resolve::Namespace;
```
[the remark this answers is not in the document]
aneksteind: 0ce727a8

c2rust-analyze/src/rewrite/expr/convert.rs
aneksteind: 3ebc774d

c2rust-analyze/src/rewrite/expr/mir_op.rs
aneksteind: 74d5f314

c2rust-analyze/src/rewrite/expr/unlower.rs
spernsteiner: I was thinking more like `fn op_is_temp(&self, op: &Operand<'tcx>) -> bool`, or maybe `fn operand_desc(&self, op: &Operand<'tcx>) -> MirOriginDesc`.  The current `get_desc` doesn't entirely make sense outside the specific context of `op_is_temp`.  For example, if you have a `StatementKind::Assign` that stores into a temporary, calling `get_desc` on the LHS would return `LoadFromTemp`, which probably isn't what you'd want.

aneksteind: 5cc3fb7e

spernsteiner (approved):

PR #972, opened 2023-06-29
analyze: README: fix invocation

fw-immunant:
This makes things a little more concise and a little less brittle (e.g. not Linux-specific, if someone is using macOS or BSD).


c2rust-analyze/README.md
```diff
@@ -1,5 +1,5 @@
 ```sh
-cargo run --bin c2rust-analyze -- tests/filecheck/insertion_sort.rs -L "$(rustc --print sysroot)/lib/rustlib/x86_64-unknown-linux-gnu/lib" --crate-type rlib
+cargo run --bin c2rust-analyze -- tests/filecheck/insertion_sort.rs -L "$(rustc --print target-libdir)" --crate-type rlib
```
kkysen, line 2: I didn't realize `rustc` had that `--print` option.  That's helpful.

PR #973, opened 2023-06-29
instrument: README: fix invocation

fw-immunant:
The invocation here was out-of-date as of #554; the command now works as intended.

### DIFF
--- a/dynamic_instrumentation/README.md
+++ b/dynamic_instrumentation/README.md
@@ -1,9 +1,10 @@
 ```sh
-cargo run -- analysis_metadata.bc ../analysis/tests/misc/src/main.rs -L ~/.rustup/toolchains/nightly-2022-02-14-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/ -L ../target/debug/deps/ --extern c2rust_analysis_rt=../target/debug/deps/libc2rust_analysis_rt-4c5a3dcb09c8161c.rlib
-INSTRUMENT_BACKEND=debug METADATA_FILE=analysis_metadata.bc ./main
+cargo build
+${CARGO_TARGET_DIR:-../target/debug/c2rust-instrument} --metadata instrument.target/debug/metadata.bc -- build --manifest-path analysis/tests/misc/Cargo.toml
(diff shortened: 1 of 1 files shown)


kkysen (commented):
How'd you determine these new invocation commands?  The most up-to-date commands are in `pdg/src/main.rs` in `mod tests`, as those are the ones tested in `cargo test` and CI.

Could we just say to look there in the `README.md`?

fw-immunant:
This is basically what's in `pdg.sh`, but as you note also matches `src/pdg/main,rs`. I don't think suggesting to read the source (even with a pointer to where) is helpful for a user-facing README--we should have a shell command so people can get started quickly.

dynamic_instrumentation/README.md
```diff
@@ -1,9 +1,10 @@
 ```sh
-cargo run -- analysis_metadata.bc ../analysis/tests/misc/src/main.rs -L ~/.rustup/toolchains/nightly-2022-02-14-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/ -L ../target/debug/deps/ --extern c2rust_analysis_rt=../target/debug/deps/libc2rust_analysis_rt-4c5a3dcb09c8161c.rlib
-INSTRUMENT_BACKEND=debug METADATA_FILE=analysis_metadata.bc ./main
+cargo build
+${CARGO_TARGET_DIR:-../target/debug/c2rust-instrument} --metadata instrument.target/debug/metadata.bc -- build --manifest-path analysis/tests/misc/Cargo.toml
+(cd analysis/tests/misc/instrument.target/debug; INSTRUMENT_BACKEND=debug INSTRUMENT_RUNTIME=bg METADATA_FILE=metadata.bc ./c2rust-analysis-tests-misc)
```
kkysen, line 4: I think this still might be broken, which is why I hadn't approved it earlier.  The `${CARGO_TARGET_DIR:-../target/debug/c2rust-instrument}` assumes the cwd is `dynamic_instrumentation/` (I think, or at least some other directory), but `analysis/tests/misc/Cargo.toml` is relative to the repo root.  Also, I don't remember if `--metadata` is supposed to be relative to the cwd or to the manifest dir.

fw-immunant: I pushed a fix that works correctly with the repo root as cwd. Even when building the entire repo with `cargo build` in the repo root, `c2rust-instrument` ends up in the `debug` subdir of the cargo target directory; the other commands already expected to be there.

`--metadata` is relative to the cwd, which does mean that we need it to include the base path of the `Cargo.toml`.

PR #974, opened 2023-06-29
scripts/pdg.sh: also run analysis

fw-immunant:
This is fairly hacky and best-effort, but until c2rust-analyze is a rustc wrapper for cargo, it's hard to do better, and this avoids duplicating the work of figuring out these hacks.

To see it in action: `./scripts/pdg.sh analysis/tests/minimal`

### DIFF
--- a/scripts/pdg.sh
+++ b/scripts/pdg.sh
@@ -41,6 +41,7 @@ main() {
     local event_log="${test_dir}/log.bc"
     local runtime="analysis/runtime"
 
+    local crate_name="$(cargo pkgid --manifest-path "${test_dir}/Cargo.toml" | grep -P -o '(?<=#)[^@]+(?=@)' | tr '-' '_')"
     (
         unset RUSTFLAGS # transpiled code has tons of warnings; don't allow `-D warnings`
         export RUST_BACKTRACE=1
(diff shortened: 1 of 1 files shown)


scripts/pdg.sh
```diff
@@ -77,6 +78,32 @@ main() {
...
+        local deps_dir="${CARGO_TARGET_DIR:-target}/${profile_dir_name}/deps"
+        local libc_dep="$(ls -1 --sort=time -r ${deps_dir}/liblibc-*.rlib | head -n1)"
+        cargo run \
+            --bin c2rust-analyze \
+            "${profile_args[@]}" \
+            -- \
+            --crate-name "${crate_name}" \
+            --edition 2021 \
+            --crate-type bin \
+            --sysroot $(rustc --print sysroot) \
```
kkysen, line 94:
```suggestion
            --sysroot "$(rustc --print sysroot)" \
```

scripts/pdg.sh
```diff
@@ -77,6 +78,32 @@ main() {
...
+            "-C" \
+            "opt-level=3" \
+            "-C" \
+            "embed-bitcode=no" \
```
kkysen, line 98: I don't think we need these, as they don't affect `c2rust-analyze`.

scripts/pdg.sh
```diff
@@ -77,6 +78,32 @@ main() {
...
+            "--cfg" \
+            "feature=\"c2rust-analysis-rt\"" \
```
kkysen, line 100: This is specific to `c2rust instrument`.  We shouldn't need it for `c2rust-analyze` since we're not linking to the runtime.

```suggestion
```

scripts/pdg.sh
```diff
@@ -77,6 +78,32 @@ main() {
...
+            "-C" \
+            "metadata=4095517b1921578c" \
```
kkysen, line 102: Do we need this?  What's that hash for?  It looks fragile.

scripts/pdg.sh
```diff
@@ -77,6 +78,32 @@ main() {
...
+            "-L" "dependency=${deps_dir}" \
+            "--extern" "libc=${libc_dep}" \
```
kkysen, line 104: We also need `c2rust_bitfields`.

See https://github.com/immunant/lighttpd-rust/blob/385ca548101242cfb8f8bc61533f1073042b5036/lighttpd_rust_amalgamated/c2rust-analyze.sh#L36-L37.

kkysen (commented):
What is this meant to run end-to-end on?  `lighttpd-minimal`?  `lighttpd-rust-amalgamated`?  Or the `c2rust-analyze` tests (which aren't full crates)?  Anything else?

kkysen (commented):
I think it would be better to move these additions to Rust code in `c2rust_pdg::tests` rather than adding to this shell script, even if that's not their eventual location once `c2rust-analyze` gets `cargo`-wrapped.  Right now, that's the canonical invocation of `c2rust-instrument` that is run in CI tests.  `scripts/pdg.sh` isn't used regularly anymore and I'm not sure if it's fully up-to-date (I'd prefer to remove it entirely).

fw-immunant:
Superseded by #1069.

PR #975, opened 2023-07-05
provide statics with hypothetical origins

### LINKED ISSUE #776: Add support for static lifetimes in ADT fields
Once #768 is merged, code such as the following will require support:

```rust
struct Foo {
  bar: &'static i32
}

fn baz(mut foo: Foo, i: &'static i32) {
    foo.bar = i;
}
```

In #768, it currently results in a panic:

```
thread 'rustc' panicked at ''static lifetimes not yet supported: ()', c2rust-analyze/src/borrowck/type_check.rs:63:34
```

The solution is to create a unique `Origin` for the `'static` lifetime that's accessible from `TypeChecker` and add it to the polonius universal region set.

### LINKED ISSUE #943: analyze: borrowck assertion failure on statics with hypothetical region params
Test case:
```Rust
// Regression test for an assertion failure in `borrowck::type_check` `do_assign` related to
// mismatched numbers of origin params when accessing statics.

static mut oneshot_fdn: fdnode = fdnode { ctx: 0 as *mut u8 };

pub struct fdnode {
    pub ctx: *mut u8,
}

unsafe extern "C" fn server_free() -> bool {
    oneshot_fdn.ctx.is_null()
}
```

Panics on [this assertion](https://github.com/immunant/c2rust/blob/d1e05f3203efa29d1277eead56a33d93d549ac0b/c2rust-analyze/src/borrowck/type_check.rs#L423):
```
thread 'rustc' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', c2rust-analyze/src/borrowck/type_check.rs:423:13
```

This error currently appears in 44 places in lighttpd

aneksteind:
Fixes #943 and fixes #776. Assigns origins to static variables, which provides parity in the number of `OriginParams` when comparing equality of LHS and RHS of assignments to local variables (which already had the correct number of `OriginParams`).

Questions:
- polonius does not appear to issue loans for statics, but it accounts for regions associated with them. Does the `assign_origins` routine for statics adequately cover this?
  - in a statement such as `_2 = const {alloc1: *mut fdnode}` the RHS is not issued a loan; is that correct? Polonius appears not to issue any loans here
- what should be done for string literals?


fw-immunant (commented):
Mostly looks good, a couple questions inline.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -164,6 +165,39 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         match *op {
             Operand::Copy(pl) | Operand::Move(pl) => self.visit_place(pl),
             Operand::Constant(ref c) => {
+                if c.ty().is_any_ptr() {
+                    if let Some(alloc_id) = const_alloc_id(c) {
+                        if let Some(did) = find_static_for_alloc(&self.acx.tcx(), alloc_id) {
```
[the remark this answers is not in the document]
aneksteind: They're the same, and I didn't appreciate `tcx` was there. Now that `acx` is there and used for other pieces of the module, I'm only passing `acx` now in 70f83061

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -164,6 +165,39 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                                origin_params: &[],
+                                perm,
+                            };
+                            return self.ltcx.mk(c.ty(), args, label);
+                        } else {
+                            // string literals
+                        }
+                    } else {
+                        // string literals
+                    }
```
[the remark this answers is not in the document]
aneksteind: You're right, I was conflating things based on what I was seeing in testing `string_literals.rs`. The only difference I can see is that inline bytestring literals appear to have a `ConstValue::Scalar`, and string literals have a `ConstValue::Slice`. I'm open to recommendations regarding the comments if you think they're worthwhile, but for now I've removed them in a186e9cb

spernsteiner:
> polonius does not appear to issue loans for statics, but it accounts for regions associated with them

Sounds right.  Loans are created on `&`/`&mut` operations; for statics, those operations essentially happen before the start of main, and aren't part of any function (so we don't analyze them).



> Does the `assign_origins` routine for statics adequately cover this?

Have you checked what Polonius normally does for statics that contain `'static` lifetimes?  I could imagine, for example, that there might be only a single origin corresponding to the concrete lifetime `'static`, or that it's otherwise treated in some special way.



> in a statement such as `_2 = const {alloc1: *mut fdnode}` the RHS is not issued a loan; is that correct? Polonius appears not to issue any loans here

I'm not sure.  What does Polonius do for `let x = &MY_STATIC`?  Does this issue a loan, or is it treated as a sort of no-op projection, like `&*some_ref`?  It might be better to treat the `const { alloc: *mut T }` as a similar sort of loan or projection (matching whatever Polonius does for `&MY_STATIC`), since unlike rustc we treat `*mut` as if it were a ref and give it an origin.

Note this is separate from the previous point, which is about statics containing refs, like `BAR` here:
```Rust
static FOO: i32 = 1;
static BAR: &'static i32 = &FOO;
```

c2rust-analyze/src/main.rs
```diff
@@ -759,6 +778,7 @@ fn run(tcx: TyCtxt) {
                     &acx,
                     &info.dataflow,
                     &mut asn.perms_mut(),
+                    &mut maps,
```
spernsteiner, line 781: I don't think it's a good idea to share a single `AtomMaps` across the borrowck invocations for all functions.  It lets the map grow quite large and is unnecessary because different functions never interact with each other's atoms.  (This is similar to the reason we have separate `LocalPointerTable`s and `GlobalPointerTable`s for working with `PointerId`s.)  If (after investigating Polonius's `'static` handling) it's actually necessary for all functions to share the global/static part of the `AtomMaps`, I think a better way to go about it is to pass `maps.clone()` by value into each function's `borrowck` invocation, so they no longer share one big map.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -369,7 +396,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                     */
                     assert_eq!(expect_ty.args.len(), 0, "Generic types not yet supported.");
 
-                    let adt_def = self.tcx.adt_def(adt_did);
+                    let adt_def = self.acx.tcx().adt_def(adt_did);
```
spernsteiner, line 399: In places where the `TyCtxt` is used multiple times, I'll usually put a `let tcx = self.acx.tcx();` or similar at the top of the function, so the use sites can avoid a bit of boilerplate (there is only one `tcx` - it doesn't really matter exactly how you got hold of it)

aneksteind: 60be3579

c2rust-analyze/src/main.rs
aneksteind: How would I determine in the `nll` output that polonius is having all function share the global/static part of the `AtomMaps`?

aneksteind:
> Have you checked what Polonius normally does for statics that contain `'static` lifetimes?  I could imagine, for example, that there might be only a single origin corresponding to the concrete lifetime `'static`, or that it's otherwise treated in some special way.

> I'm not sure.  What does Polonius do for `let x = &MY_STATIC`?  Does this issue a loan, or is it treated as a sort of no-op projection, like `&*some_ref`?  It might be better to treat the `const { alloc: *mut T }` as a similar sort of loan or projection (matching whatever Polonius does for `&MY_STATIC`), since unlike rustc we treat `*mut` as if it were a ref and give it an origin.

@spernsteiner I've run polonius on the following example:

```rust
static mut oneshot_fdn: fdnode = fdnode { ctx: &0 };

pub struct fdnode {
    pub ctx: &'static u8,
}

unsafe extern "C" fn server_free() {
    let x = &oneshot_fdn;
}

fn main() {
    unsafe {
        server_free();
    }
}
```

and it appears:
- no loans are issued
- there are two universal regions
- there are various subset relations, but none of them are in terms of the universal regions

aneksteind:
> I'm not sure. What does Polonius do for let x = &MY_STATIC? Does this issue a loan, or is it treated as a sort of no-op projection, like &*some_ref? 

It looks like the MIR for this treats it as such:

```rust
        _2 = const {alloc1: *mut fdnode}; // scope 0 at test_static.rs:8:14: 8:25
                                         // mir::Constant
                                         // + span: test_static.rs:8:14: 8:25
                                         // + literal: Const { ty: *mut fdnode, val: Value(Scalar(alloc1)) }
        _1 = &(*_2);                     // scope 0 at test_static.rs:8:13: 8:25
```

spernsteiner:
It looks like `rustc -Z nll-facts` treats `'static` as a universal/placeholder region, as if it was a parameter with the constraint that it outlives all other parameters.  E.g. `fn f<'a>(...)` is treated like `fn f<'static: 'a, 'a>(...)`.  So there isn't much special handling for `'static` in Polonius itself, aside from generating those extra constraints.  As an implementation detail, the first placeholder region `'_#0r` is always `'static`, then each lifetime parameter of the function gets a placeholder region, and finally there's an extra placeholder region at the end (maybe `'empty`, or maybe a lifetime representing the duration of the function, or maybe something else).  The first region outlives the lifetime parameters, and the parameters outlive the last region.

Accesses of statics (e.g. `let x = FOO;` where `static FOO: i32`) produce no loans and have no constraints relating them to `'static`/`'_#0r`, as you noted.  For non-mut statics, which produce MIR like `_1 = const {alloc2: &'_#3r i32};`, rustc seems to create a fresh region to the rvalue `&i32`, but leaves the region unconstrained, so it can be equal to `'static` or any shorter region as needed.  (The assignment then sets a constrain between the region on the LHS and the region `'_#3r` on the RHS.)  I think it would also be legal to constrain the region to be exactly `'static` (via `subset_base('_#3r, '_#0r); subset_base('_#0r, '_#3r)`), or to constrain it to be "`'static` or shorter".  For mut statics, the MIR looks like `_4 = const {alloc5: *mut i32};`, and subsequent "reborrow" operations like `_6 = &(*_4);` create a fresh region for the `Rvalue::Ref` but leave it unconstrained.

Since we generally assign regions to raw pointers as well as refs, I think we should treat both the mut and non-mut cases the same: assign a region to the RHS, and leave it unconstrained so it can become `'static` if needed (or something shorter if not).

For mentions of `'static` within statics (e.g. `static FOO_REF: &'static i32`), it looks like rustc generates a fresh region each time the static is mentioned, then constrains it to be equal to `'static` (`x <= '_#0r`, `'_#0r <= x`).  This seems like a reasonable approach to me (and in general it's good to stick to the way rustc does things when possible).  Instead of assigning regions to the actual static declarations, we would generate new fresh regions and constraints at each of their uses (MIR `const { allocN: *mut T }` etc.).  Real and hypothetical regions can be treated the same in this context.

I think this also simplifies the `Operand::Constant` case, since we no longer need to look up the actual static `DefId` - we simply generate a fresh region for each position where one can appear in the type, and (since global allocs can only refer to `'static` data) constrain each fresh region to be `'static`.

aneksteind:
@spernsteiner I haven't addressed the `'static` lifetime piece yet, but I've addressed generating origins for statics in https://github.com/immunant/c2rust/pull/975/commits/99adfbbc57a38d02a7714804825563db3d9a5077. Does that seem aligned with what we talked about?

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -64,6 +64,8 @@ pub enum OriginParam {
     /// A hypothesized region derived from a pointer type,
     /// e.g. `'h0` derived from the pointer in `*mut foo`
     Hypothetical(i64),
+    /// The 'static region
+    Static,
```
spernsteiner, line 68: `'static` doesn't make sense as an `OriginParam`.  `OriginParam` represents a lifetime parameter declaration (e.g. the `'a` in `struct S<'a> { ... }`), as opposed to `OriginArg` which is a "lifetime expression" used as an argument when referring to a generic type (e.g. the `'b` in `let x: S<'b> = ...;`).  It's not valid to write `'static` in a declaration position, such as `struct Foo<'static> { ... }`, so it shouldn't have an `OriginParam` variant.

Is `OriginParam::Static` actually used anywhere?  I don't see any mentions outside the `TryFrom` impl below.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -204,10 +205,12 @@ fn run_polonius<'tcx>(
...
-    // Start the origin counter at 3.  This has no effect on the semantics, but makes for easier
-    // diffs between our facts and the facts generated by rustc.
-    for _ in 0..3 {
-        let _ = maps.origin();
+    // the 'static region is always 0
+    let static_origin = maps.origin();
+    // polonius gives an origin to each argument
+    let mut arg_origins = vec![];
+    for _ in 0..mir.arg_count {
+        arg_origins.push(maps.origin());
```
spernsteiner, line 213: Rustc assigns an origin to each lifetime parameter of the function, not to each value parameter, which is what `Body::arg_count` counts.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -161,10 +162,62 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                                *c.ty().kind(),
+                                TyKind::Ref(..) | TyKind::RawPtr(..)
+                            ));
+
+                            // Polonius does not appear to issue loans for statics
+                            let label = Label {
+                                origin: None,
+                                origin_params: &[],
+                                perm,
+                            };
```
spernsteiner, line 215: Rustc doesn't issue loans for these, but it does assign an origin (at least in cases where the `Operand::Constant` for the static has ref type, which we generalize to both ref and raw pointer types).  Rustc also doesn't set any constraints based on the operand or rvalue directly, but the origin produced here can be involved in other constraints, such as the one produced by the enclosing `StatementKind::Assign`.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -161,10 +162,62 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
     }
 
-    pub fn visit_operand(&self, op: &Operand<'tcx>) -> LTy<'tcx> {
+    pub fn visit_operand(&mut self, op: &Operand<'tcx>) -> LTy<'tcx> {
         match *op {
             Operand::Copy(pl) | Operand::Move(pl) => self.visit_place(pl),
             Operand::Constant(ref c) => {
+                if c.ty().is_any_ptr() {
+                    if let Some(alloc_id) = const_alloc_id(c) {
+                        if let Some(did) = find_static_for_alloc(&self.acx.tcx(), alloc_id) {
```
spernsteiner, line 171: I think we should `error!`/`panic!` if `find_static_for_alloc` fails.  If `c` has pointer/reference type, then either it should be null/dangling, in which case `const_alloc_id` will return `None`, or else it should be a pointer to a valid static, in which case `find_static_for_alloc` should return `Some`.

c2rust-analyze/src/borrowck/mod.rs
aneksteind: Yeah, I agree. This is because `Label` has `pub origin_params: &'tcx [(OriginParam, Origin)]` which is used in for e.g. `field_lty`. I guess this issue of `Static` is because of a deficiency in that implementation and it should instead potentially be `pub origin_args: &'tcx [(OriginArg, Origin)]`

c2rust-analyze/src/borrowck/mod.rs
aneksteind: Makes sense, I totally misread your [comment](https://github.com/immunant/c2rust/pull/975#issuecomment-1629506921) explaining this.

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: So we should generate a fresh origin here as well to assign to the pointer type that `c` in `Operand::Constant(ref c)` is?

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: this will fail in cases of string literals I believe, so I'm not sure if this makes sense to do, unless you think it needs to be covered? touched upon this a little in [this comment](https://github.com/immunant/c2rust/pull/975#discussion_r1254559773)

c2rust-analyze/src/borrowck/mod.rs
aneksteind: I've made a change in dd458e0e

c2rust-analyze/src/borrowck/type_check.rs
aneksteind: Added in 6ccfd803

c2rust-analyze/src/borrowck/mod.rs
aneksteind: @spernsteiner I think I've addressed this issue in cde7b43e

aneksteind:
@spernsteiner would you mind taking another look?

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -204,11 +202,30 @@ fn run_polonius<'tcx>(
...
+            .tcx()
+            .hir()
+            .find(acx.tcx().hir().local_def_id_to_hir_id(local_def_id));
+
+        match hir_func_item {
+            Some(Node::Item(item)) => {
+                if let ItemKind::Fn(_, ref generics, _) = item.kind {
+                    for generic in generics.params.iter() {
+                        if matches!(generic.kind, GenericParamKind::Lifetime { .. }) {
+                            func_lifetime_origins.push(maps.origin())
```
spernsteiner, line 221: There are a bunch of relations that rustc would normally add over these origins - please either emit the relevant facts or add comments pointing out that they're missing:

* My read of the [polonius docs](https://rust-lang.github.io/polonius/rules/relations.html#3-placeholder-and-universal_region) is that we should emit `universal_region(origin)` and `placeholder(origin, loan)` (where `loan` is a fresh loan, not connected to anything else) for each of these region parameters and also for `'static` (which is treated like a parameter).
* There should be either `known_placeholder_subset` or `subset_base` relations matching each outlives constraint `'a: 'b` in `tcx.predicates_of(local_def_id)`.
* There should be similar subset relations for the implicit outlives constraints `'static: 'a` for each parameter `'a`.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -347,7 +364,7 @@ fn construct_adt_origins<'tcx>(
     ltcx.arena().alloc_from_iter(origins)
 }
 
-fn assign_origins<'tcx>(
+pub fn assign_origins<'tcx>(
```
spernsteiner, line 367: Does this need to be public?  It looks like the only new use of it in this PR is in `borrowck::type_check`, which can see it regardless.

c2rust-analyze/src/borrowck/type_check.rs
spernsteiner: Ah, good point.  Never mind, then.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -161,10 +164,63 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
+                            let pointer_id = self.acx.type_of(op).label;
+                            let perm = self.hypothesis[pointer_id];
+                            let args = self.ltcx.mk_slice(&[lty]);
+                            assert!(matches!(
+                                *c.ty().kind(),
+                                TyKind::Ref(..) | TyKind::RawPtr(..)
+                            ));
+
+                            // Polonius does not appear to issue loans for statics,
+                            // but assigns an origin
```
spernsteiner, line 213:
```suggestion
                            // Polonius does not appear to issue loans for the outermost ref/ptr in
                            // an address-of-static constant, but it does assign an origin
```
Small edit to clarify why the origin here is handled differently from the origins produced by `assign_origins` above.

c2rust-analyze/src/context.rs
```diff
@@ -300,6 +300,7 @@ pub struct GlobalAnalysisCtxt<'tcx> {
     pub field_ltys: HashMap<DefId, LTy<'tcx>>,
 
     pub static_tys: HashMap<DefId, LTy<'tcx>>,
+    pub static_origins: HashMap<DefId, crate::borrowck::LTy<'tcx>>,
```
spernsteiner, line 303: Looks like this is now unused.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -204,11 +202,30 @@ fn run_polonius<'tcx>(
...
+        let hir_func_item = acx
+            .tcx()
+            .hir()
+            .find(acx.tcx().hir().local_def_id_to_hir_id(local_def_id));
+
+        match hir_func_item {
+            Some(Node::Item(item)) => {
+                if let ItemKind::Fn(_, ref generics, _) = item.kind {
+                    for generic in generics.params.iter() {
+                        if matches!(generic.kind, GenericParamKind::Lifetime { .. }) {
```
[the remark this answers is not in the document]
aneksteind: d9ed884a

c2rust-analyze/src/borrowck/mod.rs
aneksteind: 3303860a

c2rust-analyze/src/context.rs
aneksteind: 91517f0b

c2rust-analyze/src/borrowck/mod.rs
aneksteind: Does this match what you suggest? d6220b75

aneksteind: It seems like there is one remaining known placeholder subset relationship between lifetimes life `arg: &'a &'c i32`. I can look into what it would take to add it to this PR or we can save it for another PR

spernsteiner: That looks about right to me.

The `'c: 'a` bound that's implied by well-formedness of `arg`'s type should definitely show up somewhere.  I think I've seen bounds like that in the output of either `predicates_of` or a related query.  Maybe try looking at `explicit_predicates_of`, `inferred_outlives_of`, and/or `predicates_defined_on`.

It would also be fine to just leave a comment about the limitation, especially if you can't find an easy way to get the predicate from rustc.

spernsteiner (approved):

c2rust-analyze/src/borrowck/mod.rs
aneksteind: `predicates_defined_on` seems to encapsulate the former two, and `predicates_of` is supposed to almost-always be exactly `predicates_defined_on` (according to the documentation), but I don't see that any of these options capture the implicit relationship. i'll keep digging

aneksteind: I've created an issue for this: https://github.com/immunant/c2rust/issues/1002#issue-1816492960

PR #976, opened 2023-07-05
fix mismatch in number of origin parameters for cast result

### LINKED ISSUE #944: analyze: borrowck assertion failure on casts involving hypothetical region params
Test case:
```Rust
// Regression test for an assertion failure in `borrowck::type_check` `do_assign` related to
// mismatched numbers of origin params in casts.

pub struct fdnode {
    pub ctx: *mut u8,
}

unsafe extern "C" fn server_free(fdn: *mut fdnode) {
    let _fdn2 = fdn as *const fdnode;
}
```

Panics on [this assertion](https://github.com/immunant/c2rust/blob/d1e05f3203efa29d1277eead56a33d93d549ac0b/c2rust-analyze/src/borrowck/type_check.rs#L423):
```
thread 'rustc' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', c2rust-analyze/src/borrowck/type_check.rs:423:13
```

This error currently appears in 102 places in lighttpd

aneksteind:
Fixes #944. Previously cast result types were labeled with the default label, which have no `OriginParams`. This PR mimics the behavior of labeling cast results in the special case that the cast was from a null const type, where we relabled the `expected_ty` with fresh origins.

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -316,13 +316,6 @@ impl<'tcx> TypeChecker<'tcx, '_> {
             }
             Rvalue::Cast(CastKind::Pointer(PointerCast::MutToConstPointer), ref op, _ty) => {
                 let op_lty = self.visit_operand(op);
-                // TODO: handle Unsize casts at minimum
-                /*
-                assert!(
-                    !matches!(ty.kind(), TyKind::RawPtr(..) | TyKind::Ref(..)),
-                    "pointer Cast NYI"
-                );
-                */
                 // Here we relabel `expect_ty` to utilize the permissions it carries
                 // but substitute the rest of its `Label`s' parts with fresh origins
                 // Othwerise, this is conceptually similar to labeling the cast target
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -187,6 +187,36 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         origin
     }
 
+    fn relabel_fresh_origins(&mut self, expect_ty: LTy<'tcx>) -> LTy<'tcx> {
+        self.ltcx.relabel(expect_ty, &mut |lty| {
+            let perm = lty.label.perm;
+            match lty.ty.kind() {
+                TyKind::Ref(_, _, _) | TyKind::RawPtr(_) => {
+                    let origin = Some(self.maps.origin());
+                    Label {
+                        origin,
+                        origin_params: &[],
+                        perm,
+                    }
+                }
+                TyKind::Adt(..) => {
+                    let origin_params =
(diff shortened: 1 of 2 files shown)


c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -273,58 +303,31 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
                 );
                 */
-                Label::default()
-            }),
+                // Here we relabel `expect_ty` to utilize the permissions it carries
+                // but substitute the rest of its `Label`s' parts with fresh origins
+                // Othwerise, this is conceptually similar to labeling the cast target
+                // `ty`. We would simply do that, but do not have the information necessary
+                // to set its permissions.
+                self.relabel_fresh_origins(expect_ty)
```
spernsteiner, line 329: I think we probably need to establish some kind of relationships between the fresh origins and the `Operand` of the `Rvalue::Cast`.  Otherwise, in `let z: Z = (x: X) as Y;`, there is a relationship between Y and Z (due to the enclosing `Assign`), but no relationship between X and Y, so X and Y/Z will be considered totally independent.

The null pointer case this is based on is a bit of a special case since it's effectively generating a pointer from thin air rather than deriving a new pointer from an existing one.  A better reference is the `Aggregate` case below, which (at least for tuples) generates some fresh origins but then connects them to existing operands.

spernsteiner: I think `do_assign` happens to mostly work, but it would be better to match on specific `CastKind`s/`TyKind`s and explicitly do the right thing for each one.  (`do_assign` should probably also assert that the two `Ty`s are equivalent, like `dataflow::type_check`'s `do_assign`/`do_equivalence_nested` does.)

aneksteind: Is f3751839 sufficient for now? I think it covers the `lighttpd` failure and checks for equality. I think it might be best if we match on specific `CastKind`s/`TyKind`s as needed.

spernsteiner: Yes, that seems fine for now.  Though it might be slightly cleaner to do the `TyKind::RawPtr` check inside the `MutToConstPointer`, where it's clear why that logic is needed.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -273,48 +304,36 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
             }
+            Rvalue::Cast(CastKind::Pointer(PointerCast::MutToConstPointer), ref op, _ty) => {
+                let op_lty = self.visit_operand(op);
+                // TODO: handle Unsize casts at minimum
+                /*
+                assert!(
+                    !matches!(ty.kind(), TyKind::RawPtr(..) | TyKind::Ref(..)),
+                    "pointer Cast NYI"
+                );
+                */
```
spernsteiner, line 325: This comment and assert don't apply to this new case.

spernsteiner (approved):

PR #978, opened 2023-07-05
(`c2rust-analyze`) Add `known_fns!` for declaring the permissions on ptrs in known (i.e. `libc`) `UnknownDef` `fn`s

kkysen:
I wrote this as a `macro_rules!` macro so that we can write the pointer permission annotations inline with what the normal `extern` declaration would look like.  The macro checks that the declaration matches `libc`'s (or wherever it could be from) and that there are the correct number of `PermissionSet`s on each type corresponding to the number of pointers directly in that type (i.e., the number of `*`).

The syntax for this is an optional `: [PERM1a | PERM1b, PERM2a | PERM2b | PERM2c]` after each type in a `fn` signature, where the order of the `PermissionSet`s corresponds to the pointers in the type in left-to-right order.  For example:

```rust
known_fns! {
    mod libc {

        #[cfg(target_os = "linux")]
        fn __errno_location() -> *mut c_int: [READ | WRITE];

        #[cfg(target_os = "macos")]
        fn __error() -> *mut c_int: [READ | WRITE];

        fn _exit(
            status: c_int,
        ) -> !;

        fn abort() -> !;

        fn abs(
            i: c_int,
        ) -> c_int;

        fn accept(
            socket: c_int,
            address: *mut sockaddr: [WRITE],
            address_len: *mut socklen_t: [READ | WRITE],
        ) -> c_int;

        fn read(
            fd: c_int,
            buf: *mut c_void: [WRITE | OFFSET_ADD],
            count: size_t,
        ) -> ssize_t;

        fn write(
            fd: c_int,
            buf: *const c_void: [READ | OFFSET_ADD],
            count: size_t,
        ) -> ssize_t;

        fn strtol(
            s: *const c_char: [READ | OFFSET_ADD],
            endp: *mut *mut c_char: [WRITE, WRITE | OFFSET_ADD],
            base: c_int,
        ) -> c_long;

    }
};
```

I'm still working on the integration of these `KnownFn`s into `fn_sigs` as `LFnSig`s, but I wanted to open a separate PR for how they are declared in the first place.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -135,7 +135,7 @@ impl PermissionSet {
     // `.union` is used here since it's a `const fn`, unlike `BitOr::bitor`.
     pub const STRING_LITERAL: Self = Self::READ.union(Self::OFFSET_ADD);
 
-    #[allow(unused)]
+    #[cfg(test)]
     pub const fn union_all<const N: usize>(a: [Self; N]) -> Self {
         let mut this = Self::empty();
         let mut i = 0;
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -134,6 +134,17 @@ impl PermissionSet {
     //
     // `.union` is used here since it's a `const fn`, unlike `BitOr::bitor`.
     pub const STRING_LITERAL: Self = Self::READ.union(Self::OFFSET_ADD);
+
+    #[allow(unused)]
+    pub const fn union_all<const N: usize>(a: [Self; N]) -> Self {
+        let mut this = Self::empty();
+        let mut i = 0;
+        while i < N {
+            this = this.union(a[i]);
+            i += 1;
+        }
+        this
+    }
 }
 
 bitflags! {
--- a/c2rust-analyze/src/known_fn.rs
+++ b/c2rust-analyze/src/known_fn.rs
@@ -1,4 +1,3 @@
-#[cfg(test)]
 use crate::context::PermissionSet;
 
 #[allow(unused)]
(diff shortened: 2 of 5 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -134,6 +134,17 @@ impl PermissionSet {
...
+
+    #[cfg(test)]
+    pub const fn union_all<const N: usize>(a: [Self; N]) -> Self {
+        let mut this = Self::empty();
+        let mut i = 0;
+        while i < N {
+            this = this.union(a[i]);
+            i += 1;
+        }
+        this
```
spernsteiner, line 146: Why not use a for loop here?

c2rust-analyze/src/known_fn.rs
```diff
@@ -0,0 +1,362 @@
...
+    const fn checked(self) -> Self {
+        let ty = self.ty.as_bytes();
+        let mut num_ptrs = 0;
+        let mut i = 0;
+        while i < ty.len() {
+            if ty[i] == b'*' {
+                num_ptrs += 1;
+            }
+            i += 1;
+        }
```
spernsteiner, line 41: Why not use a for loop here?

c2rust-analyze/src/known_fn.rs
```diff
@@ -0,0 +1,362 @@
...
+#[cfg(test)]
+macro_rules! known_fn_ty {
+    ($ty:ty: $perms:tt) => {{
+        KnownFnTy {
+            name: "",
+            ty: stringify!($ty),
+            perms: const_slice!(PermissionSet, perms_annotation!($perms)),
+        }
+        .checked()
+    }};
```
spernsteiner, line 62: Seems like this could take the name directly, instead of having a separate `named` method.  The tests below that use `known_fn_ty` without `named` could pass a dummy name instead.

spernsteiner (approved):
Looks good to me.  It's a bit unfortunate that the annotations have to be written out-of-line (`*mut *mut T: [A, B]` rather than `*mut {A} *mut {B} T` or similar), but parsing the latter with `macro_rules` would be a real pain.  This should work fine for the libc use case, where the types can't get too complex.

c2rust-analyze/src/context.rs
kkysen: `for` loops can't be used in `const` `fn`s since they call `Iterator::next`, a `trait` method that can't be `const`.

c2rust-analyze/src/known_fn.rs
kkysen: Same as https://github.com/immunant/c2rust/pull/978#discussion_r1253750006.

c2rust-analyze/src/known_fn.rs
kkysen: I tried that, but (and I forgot the exact reason why) it makes the macro ambiguous or the type inference not work or something.

kkysen:
> It's a bit unfortunate that the annotations have to be written out-of-line (`*mut *mut T: [A, B]` rather than `*mut {A} *mut {B} T` or similar), but parsing the latter with `macro_rules` would be a real pain.

Yeah, I agree that would be ideal, but that would be quite complex to parse and it'd probably mess up other nice-to-have stuff like syntax highlighting.

PR #979, opened 2023-07-06
analyze: recover from failures in shim generation

spernsteiner:
Previously, if `bad` would call `good` through a shim, but generation of `good_shim` fails (for example, because the shim would require an unsupported cast like `*mut T -> &[T]`), then `bad` would still be rewritten to call the nonexistent `good_shim` function.  With this branch, we instead mark `good` as failed so that `good_shim` is no longer needed (shims are only needed for calling non-failed functions) and retry the entire rewriting pass.  Marking a function as failed may require shims to be generated for its callees, but eventually this process will converge once all the functions where shims can't be generated have been marked as failed.

### DIFF
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -800,96 +800,113 @@ fn run(tcx: TyCtxt) {
         );
     }
 
-    // Before generating rewrites, add the FIXED flag to the signatures of all functions that
-    // failed analysis.
-    for did in gacx.iter_fns_failed() {
-        let lsig = gacx.fn_sigs[&did];
-        for sig_lty in lsig.inputs_and_output() {
-            for lty in sig_lty.iter() {
-                let ptr = lty.label;
-                if !ptr.is_none() {
-                    gasn.flags[ptr].insert(FlagSet::FIXED);
-                }
-            }
-        }
-    }
-
     // Buffer debug output for each function.  Grouping together all the different types of info
     // for a single function makes FileCheck tests easier to write.
     let mut func_reports = HashMap::<LocalDefId, String>::new();
 
     // Generate rewrites for all functions.
     let mut all_rewrites = Vec::new();
-    for &ldid in &all_fn_ldids {
-        if gacx.fn_failed(ldid.to_def_id()) {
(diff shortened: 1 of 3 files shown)


c2rust-analyze/src/main.rs
```diff
@@ -800,97 +800,115 @@ fn run(tcx: TyCtxt) {
...
     let mut all_rewrites = Vec::new();
-    for &ldid in &all_fn_ldids {
-        if gacx.fn_failed(ldid.to_def_id()) {
-            continue;
+
+    // It may take multiple tries to reach a state where all rewrites succeed.
+    loop {
+        func_reports.clear();
+        all_rewrites.clear();
+        eprintln!("\n\n ------ START LOOP ------");
```
aneksteind, line 814: nit: I think something like `START REWRITE LOOP` would be better

c2rust-analyze/tests/filecheck/unrewritten_calls_shim_fail.rs
```diff
@@ -0,0 +1,37 @@
+//! --catch-panics
+#![feature(register_tool)]
+#![register_tool(c2rust_analyze_test)]
+// Like `unrewritten_calls.rs`, but specifically testing the case where shim generation fails.
+
+// CHECK-LABEL: fn slice_good1(x: &mut [(i32)])
```
aneksteind, line 6: should this assert that a cast is present to call `slice_bad`?

aneksteind (approved):
LGTM, but if the test could use an assertion for the cast then it'd be good to put that before merging

c2rust-analyze/tests/filecheck/unrewritten_calls_shim_fail.rs
spernsteiner: Cast insertion is already checked by `unrewritten_calls.rs`; this test is only checking the retry behavior when shim generation fails.

c2rust-analyze/src/main.rs
spernsteiner: I cleaned up the debug prints a little bit

PR #980, opened 2023-07-07
(`c2rust-analyze`) Allow calls to `UnknownDef` `KnownFn`s with known ptr perms

kkysen:
This allows `Callee::UnknownDef` calls if they are direct (i.e. not a `fn` ptr), call a foreign `fn` (i.e. from an `extern` block), and are a `KnownFn` (i.e. returned by `const fn all_known_fns`).  They are treated as normal local `fn` calls (i.e. they use `.visit_local_call`, like for `Callee::LocalDef`) as the `PointerId`s from the `LFnSig`s for these foreign `fn`s are constrained by the `PermissionSet`s annotated in the `KnownFn`s.  Note that these `LFnSig`s for the foreign `fn`s are already created by `fn gather_foreign_sigs`.

I've included a test that checks that a call to `libc`'s `access` succeeds without crashing and that the pointer passed to it is successfully inferred to be `READ | OFFSET_ADD` from `access`'s known ptr perms, and thus intended to be rewritten as a slice (though it's not actually rewritten since the foreign `fn` call is `FIXED`).


c2rust-analyze/src/context.rs
```diff
@@ -665,6 +672,21 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
     pub fn iter_fns_failed(&self) -> impl Iterator<Item = DefId> + '_ {
         self.fns_failed.keys().copied()
     }
+
+    pub fn known_fn_ptr_perms<'a>(
+        &'a self,
+    ) -> impl Iterator<Item = (PointerId, PermissionSet)> + PhantomLifetime<'tcx> + 'a {
+        let tcx = &self.tcx;
+        self.fn_sigs
```
aneksteind, line 680: I might not be remembering this correctly, but don't we already keep track of the foreign signatures? Do you need to filter them here?

c2rust-analyze/src/known_fn.rs
```diff
@@ -143,7 +142,64 @@ impl Display for KnownFn {
...
+                    ::log::warn!(
+                        "declared `extern \"C\" fn` type \
+                     \n\tknown_ty: ({known_ty_num_ptrs}) {known_ty}\
+                     \n\tlty: ({lty_num_ptrs}) {lty:?}\
+                "
+                    )
+                }
+                matching
+            })
+            .flat_map(|(lty, known_ty)| {
```
aneksteind, line 193: nit: the iterator calls are clean but a little tough to digest at first glance. can you maybe add a comment to the steps noting the stages of the pipeline and what's being checked?

c2rust-analyze/src/known_fn.rs
```diff
@@ -143,7 +142,64 @@ impl Display for KnownFn {
     }
 }
 
-#[cfg(test)]
+impl KnownFn {
+    pub fn inputs_and_output(&self) -> impl Iterator<Item = &KnownFnTy> {
+        self.inputs.iter().chain([&self.output])
+    }
+
+    pub fn ptr_perms<'a, 'tcx>(
```
aneksteind, line 150: can you add some documentation here explaining what's filtered and how the results are derived

c2rust-analyze/src/main.rs
```diff
@@ -425,6 +426,8 @@ fn run(tcx: TyCtxt) {
         gacx.fn_sigs.insert(ldid.to_def_id(), lsig);
     }
 
+    gather_foreign_sigs(&mut gacx, tcx);
```
aneksteind, line 429: In [this comment](https://github.com/immunant/c2rust/pull/980/files#r1255886959) this is where I was wondering if we already have what you need without needing to filter all fns

c2rust-analyze/src/util.rs
```diff
@@ -77,6 +78,23 @@ pub fn describe_rvalue<'tcx>(rv: &Rvalue<'tcx>) -> Option<RvalueDesc<'tcx>> {
     })
 }
 
+#[derive(Debug)]
+pub enum UnknownCall<'tcx> {
```
aneksteind, line 82: please add documentation

c2rust-analyze/src/util.rs
```diff
@@ -170,28 +188,34 @@ pub fn ty_callee<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> Callee<'tcx> {
     };
 
     match *ty.kind() {
-        ty::FnDef(did, substs) => {
+        ty::FnDef(def_id, substs) => {
```
aneksteind, line 191: nit: `did` is used elsewhere and I believe is common in `rustc` code. It might be beneficial to keep things consistent

c2rust-analyze/tests/filecheck/known_fn.rs
```diff
@@ -0,0 +1,19 @@
...
+
+extern "C" {
+    fn access(
+        path: *const c_char,
+        amode: c_int,
+    ) -> c_int;
+}
+
+// CHECK-LABEL: final labeling for "known_fn"
+pub fn known_fn() {
```
aneksteind, line 11: this looks great, are there some more examples you can add here that appear in lighttpd?

c2rust-analyze/src/known_fn.rs
```diff
@@ -463,3 +519,59 @@ mod tests {
         );
     }
 }
+
+pub const fn all_known_fns() -> &'static [KnownFn] {
+    known_fns! {
+        mod libc {
+
+            fn __errno_location() -> *mut c_int: [READ | WRITE];
```
aneksteind, line 527: do some of these need `#[cfg(not(target_os = "macos"))]`?

c2rust-analyze/src/context.rs
kkysen: Not in a separate map, I don't think. `gather_foreign_sigs` adds its `LFnSig`s to `gacx.fn_sigs`.

aneksteind: Ah right, I had thought about making a separate map but folded it into `fn_sigs` after all

c2rust-analyze/src/context.rs
```diff
@@ -665,6 +672,21 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
+            .iter()
+            .filter(|(did, _)| tcx.def_kind(tcx.parent(**did)) == DefKind::ForeignMod)
+            .filter_map(|(&did, fn_sig)| {
+                let name = tcx.item_name(did);
```
spernsteiner, line 684: It would be a little more robust to use `tcx.symbol_name` here, which accounts for `#[link_name]` attributes.  `extern`-block `fn`s have no substs, so it should be easy to build the `Instance` that `symbol_name` requires.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -541,6 +549,10 @@ pub fn visit<'tcx>(
         tc.constraints.add_all_perms(ptr, perms);
     }
 
+    for (ptr, perms) in acx.gacx.known_fn_ptr_perms() {
+        tc.constraints.add_all_perms(ptr, perms);
+    }
```
spernsteiner, line 554: This is O(N^2), since `dataflow::type_check::visit` runs on (roughly) every function, and `known_fn_ptr_perms` visits every function to check for `KnownFn` permissions.

First, I don't think it's necessary to use constraints to set the permissions on these pointers.  The `AllPerms` constraint just ORs some bits into the `PointerId`'s entry in the `Assignment`, which is easy to do directly without using `DataflowConstraints`.  I think we could set these permissions up front, then run the dataflow analysis, and finally assert that the permissions were unchanged during the analysis (mainly, if the program is demanding more of a libc function's return value than it actually provides, that's a problem).

Second, if we do actually need constraints here, it would be more efficient to build a global baseline `DataflowConstraints` and clone it for each `visit`, rather than reconstructing those constraints each time.

c2rust-analyze/src/context.rs
```diff
@@ -293,6 +293,8 @@ pub struct GlobalAnalysisCtxt<'tcx> {
 
     pub fn_sigs: HashMap<DefId, LFnSig<'tcx>>,
 
+    pub known_fns: HashMap<Symbol, &'static KnownFn>,
```
spernsteiner, line 296: Please add a doc comment explaining what the key means (I believe it's the C/ASM symbol name of the libc function in question)

c2rust-analyze/src/util.rs
```diff
@@ -115,7 +133,7 @@ pub enum Callee<'tcx> {
     /// Or it could a function in another non-local crate, such as `std`,
     /// as definitions of functions from other crates are not available,
     /// and we definitely can't rewrite them at all.
-    UnknownDef { ty: Ty<'tcx> },
+    UnknownDef(UnknownCall<'tcx>),
```
spernsteiner, line 136: It seems like ideally we'd like to have a separate `Callee` variant for known fns, rather than making them a subset of `UnknownDef`.  The full `AnalysisCtxt` isn't available here, but maybe you could have a `lazy_static` `HashSet` of known fn symbols, which would be initialized from the `KnownFn`s table on first use.

Failing that, maybe it would be cleaner to split out only the specific case we care about (foreign fn, no substs, identified by its symbol name) into a separate `Callee` and leave the rest as the old `UnknownDef(Ty)`.  Having a separate `UnknownCall` currently doesn't buy us much, since the `Indirect`, `Unknown`, and `Direct { is_foreign: false }` cases are still all treated the same everywhere.

c2rust-analyze/src/util.rs
spernsteiner: rustc itself is not especially consistent in this regard, though it does seem to favor `def_id` over `did` somewhat.

c2rust-analyze/src/known_fn.rs
kkysen: Added in e27102259801bd511c169829d6d6df03430f613b.

c2rust-analyze/src/known_fn.rs
kkysen: I added it in the `fn`'s docs in e27102259801bd511c169829d6d6df03430f613b.

c2rust-analyze/src/util.rs
kkysen: > rustc itself is not especially consistent in this regard, though it does seem to favor `def_id` over `did` somewhat.

Yeah, I've noticed that as well.  `def_id` is slightly more preferred I think.  Changing it here got rid of the `def_id: did` in a `struct` initializer, so it seemed slightly cleaner.

c2rust-analyze/src/context.rs
```diff
@@ -179,8 +179,8 @@ pub struct LFnSig<'tcx> {
 }
 
 impl<'tcx> LFnSig<'tcx> {
-    pub fn inputs_and_output(&self) -> impl Iterator<Item = LTy<'tcx>> + 'tcx {
-        self.inputs.iter().copied().chain(iter::once(self.output))
+    pub fn inputs_and_output(&self) -> impl Iterator<Item = &LTy<'tcx>> {
+        self.inputs.iter().chain([&self.output])
```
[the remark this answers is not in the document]
kkysen: I had made this change separately before your PR had merged, so I merged them into this during the rebase.  No particular reason I left the `&` other than I initially/randomly chose it for my version and it worked for both call sites.  I can probably change it without having to touch other code much; let me see.

c2rust-analyze/src/context.rs
kkysen: Oh, I didn't see `tcx.symbol_name`.  That seems better.

c2rust-analyze/src/context.rs
kkysen: Yeah, that's what it is.  I'll add docs.

c2rust-analyze/src/util.rs
kkysen: There are a couple of other cases I was considering:
1. `libc` `fn`s that are in-general, known, but we haven't yet added them to `all_known_fns`.  It'd be nice to have a clear error message in this case that this is a `fn` that should be known.  These would be `Direct { is_foreign: true }`
2. non-local Rust `fn`s, such as those from `std`.  These could be known, though we'd need a different approach to them as they may be generic.  These would be `Direct { is_foreign: false }`.
3. `fn` ptrs that we don't handle yet.  These would be `Indirect` and it'd be clear that they belong to the class of `fn` ptrs, which aren't yet handled (as opposed to the other 2 classes above).
4. Anything else that could get through that we aren't anticipating.  These would be `Unknown`.

Maybe we could embed these within `Callee` rather than having a separate `UnknownCall`, but I think the distinction between the 4 (or 5 if we separate #1 into contained in `all_known_fns` and those not) cases is useful and still pretty simple code.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: > First, I don't think it's necessary to use constraints to set the permissions on these pointers. The `AllPerms` constraint just ORs some bits into the `PointerId`'s entry in the `Assignment`, which is easy to do directly without using `DataflowConstraints`.

Ah, I didn't realize we could do that.  That would definitely be better.  What's the API for that?

> finally assert that the permissions were unchanged during the analysis (mainly, if the program is demanding more of a libc function's return value than it actually provides, that's a problem).

Like we do for the string literal permissions, right?  That makes sense.

> Second, if we do actually need constraints here, it would be more efficient to build a global baseline `DataflowConstraints` and clone it for each `visit`, rather than reconstructing those constraints each time.

Yeah, I could definitely do that, though I'd prefer a better way to set the permissions in the first place.

c2rust-analyze/tests/filecheck/known_fn.rs
kkysen: Yeah, there'll definitely be more.  How many is useful to add here?  Are there any particular cases that you think would be helpful to test?

c2rust-analyze/src/util.rs
kkysen: Added in cafff64402597ff4ab4a10949238145456805161.  I also renamed it to `UnknownDefCallee` for clarity in 98898c63b342503957806b1e26da150a216a1e7a.

c2rust-analyze/src/context.rs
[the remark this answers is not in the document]
kkysen: Changed in 82ba8d8baa32fca5a1db9b51356d460f5728e0fe.

c2rust-analyze/src/context.rs
kkysen: Changed in b608247330c051b7b7b86b9f40071080e1ed7d69.  This also changed the key of the `known_fns` map to `&'static str` since `SymbolName::name` is just a `&str`, not a `Symbol`, so the `Symbol` lookup advantage is lost if we have to intern it every time.

c2rust-analyze/src/context.rs
kkysen: Documented in 17d40e86e2b88a35b29f1698272fbb1d57c31e0a.  It's also now changed to a `&'static str` vs. a `Symbol` in b608247330c051b7b7b86b9f40071080e1ed7d69 (see https://github.com/immunant/c2rust/pull/980#discussion_r1257173224).

c2rust-analyze/src/dataflow/type_check.rs
kkysen: > First, I don't think it's necessary to use constraints to set the permissions on these pointers. The `AllPerms` constraint just ORs some bits into the `PointerId`'s entry in the `Assignment`, which is easy to do directly without using `DataflowConstraints`.

I'm still unsure how to do this, so for now I did this:

> Second, if we do actually need constraints here, it would be more efficient to build a global baseline `DataflowConstraints` and clone it for each `visit`, rather than reconstructing those constraints each time.

in 84d8be42363d9ca5ce25878287e3400f261f4372.

kkysen: Nevermind, I think I might've figured out how to do that.  Let me try that way.

kkysen: > First, I don't think it's necessary to use constraints to set the permissions on these pointers. The `AllPerms` constraint just ORs some bits into the `PointerId`'s entry in the `Assignment`, which is easy to do directly without using `DataflowConstraints`. I think we could set these permissions up front, then run the dataflow analysis, 

Done in 5c093c687ca001ad5f13c319aec0f0888fe85b3b.

> and finally assert that the permissions were unchanged during the analysis (mainly, if the program is demanding more of a libc function's return value than it actually provides, that's a problem).

Done in c5444e9c17ee248869b7cc41fff464bae50346b3.

spernsteiner: Yep, that's what I was thinking of

c2rust-analyze/src/util.rs
spernsteiner: Sounds reasonable to me

spernsteiner (approved):

PR #998, opened 2023-07-19
(`c2rust-analyze`) Add almost all `libc` `KnownFn`s used in `lighttpd_rust_amalgamated`

### LINKED ISSUE #865: (`c2rust-analyze`) Hardcode ptr permissions and effects of known `std` and `libc` functions
Most functions in `libc` and Rust `std` only have local effects (see #855).  As per #843, pointers in such function signatures will receive `PermissionSet::pessimistic()` permissions as a pessimistic fallback.  But since most `std` and `libc` functions are well-known and quite common, we can hardcode the correct pointer permissions for them.

For example, for `read`,

```rust
fn read(__fd: c_int, __buf: *mut c_void, __nbytes: size_t) -> ssize_t;
```

we know that `__buf` is written (and multiple bytes) to by `read`, so we can give it precise ptr permissions of `WRITE | OFFSET_ADD`, rather than `PermissionSet::pessimistic()`, which has many more permissions.  These hardcoded permissions don't necessarily have to be fully precise in all cases; they can still be somewhat pessimistic in certain cases and still be correct, but be a lot better than `PermissionSet::pessimistic()`.

By default, we assume that these functions don't have global effects (e.x. accessing globals and other functions through reverse FFI), but in the case that they do, we can mark them as such.

See also https://github.com/immunant/c2rust/pull/855#issuecomment-1456720671.

kkysen:
* Fixes #865.

This adds `known_fns!` annotations for all `libc` `fn`s used in `lighttpd_rust_amalgamated` with a few exceptions:
* vararg `fn`s aren't yet supported (support for some of them is a WIP)
* `libc` `fn`s that are not in the `libc` crate for some reason (will find them)
* `fn`s with `fn` ptr args, as I'm not certain how to handle them yet
* non-`libc` `fn`s used by `lighttpd_rust_amalgamated` (`pcre2` `fn`s)

I've left `TODO`s for these, as well as for some other ptr perms I wasn't certain about and would like to have reviewed.

I've also outlined in https://github.com/immunant/c2rust/issues/865#issuecomment-1638820466 how we might support some of these exceptions.

### DIFF
--- a/c2rust-analyze/src/known_fn.rs
+++ b/c2rust-analyze/src/known_fn.rs
@@ -683,7 +683,7 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
             ) -> *mut c_void: [READ | WRITE];
 
             fn dlsym(
-                handle: *mut c_void: [READ | WRITE], // TODO(kkysen) may not be a pointer
+                handle: *mut c_void: [READ | WRITE], // TODO(kkysen) May not be a `VALID` pointer.
                 symbol: *const c_char: [READ | OFFSET_ADD | NON_NULL],
             ) -> *mut c_void: [READ | WRITE | OFFSET_ADD | OFFSET_SUB | FREE]; // Could be anything
 
@@ -1055,14 +1055,14 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
             ) -> c_int;
 
             fn mmap(
-                // not yet a valid pointer
+                // Not yet a `VALID` pointer.
                 addr: *mut c_void: [NONE],
                 len: size_t,
                 prot: c_int,
                 flags: c_int,
                 fd: c_int,
                 offset: off_t,
-                // TODO(kkysen) not always a pointer, can be -1
+                // TODO(kkysen) Not always a `VALID` pointer, as it can be `-1`.
             ) -> *mut c_void: [READ | WRITE | OFFSET_ADD | NON_NULL];
 
             fn munmap(
--- a/c2rust-analyze/src/known_fn.rs
+++ b/c2rust-analyze/src/known_fn.rs
@@ -579,6 +579,8 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
             #[cfg(target_os = "macos")]
             fn __error() -> *mut c_int: [READ | WRITE | NON_NULL];
 
+            // fn __ctype_b_loc;
+
             fn _exit(
                 status: c_int,
             ) -> !;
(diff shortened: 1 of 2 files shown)


c2rust-analyze/src/known_fn.rs
```diff
@@ -602,28 +622,823 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
...
+                address_len: socklen_t,
+            ) -> c_int;
+
+            fn calloc(
+                nobj: size_t,
+                size: size_t,
+            ) -> *mut c_void: [READ | WRITE | OFFSET_ADD | FREE];
+
+            fn chdir(
+                dir: *const c_char: [READ | OFFSET_ADD | NON_NULL],
```
spernsteiner, line 649: Might be nice to have an alias for the "standard" string-argument permissions, since they show up in so many functions.

c2rust-analyze/src/known_fn.rs
```diff
@@ -602,28 +622,823 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
...
+                errcode: c_int,
+            ) -> *const c_char: [READ | OFFSET_ADD | NON_NULL];
+
+            fn getaddrinfo(
+                node: *const c_char: [READ | OFFSET_ADD],
+                service: *const c_char: [READ | OFFSET_ADD],
+                hints: *const addrinfo: [READ],
+                // TODO(kkysen) Should the 2nd ptr be `FREE`, as it could be manually `free`d,
+                // but is not meant to be `free`d directly, but by `freeaddrinfo`?.
+                res: *mut *mut addrinfo: [WRITE | NON_NULL, WRITE | NON_NULL],
```
spernsteiner, line 801: Good question.  I sort of expect that we'll end up tying `FREE` to `Box<T>`/`Rc<T>`/etc, which don't (yet) support custom deallocators like `freeaddrinfo`.  But maybe we could handle this with an additional permission or flag that translates into a custom "linear (non-affine) box" type that must be explicitly freed with a special function (you can't just let it go out of scope).

c2rust-analyze/src/known_fn.rs
```diff
@@ -602,28 +622,823 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
...
+            fn mmap(
+                // not yet a valid pointer
+                addr: *mut c_void: [NONE],
+                len: size_t,
+                prot: c_int,
+                flags: c_int,
+                fd: c_int,
+                offset: off_t,
+                // TODO(kkysen) not always a pointer, can be -1
+            ) -> *mut c_void: [READ | WRITE | OFFSET_ADD | NON_NULL];
```
spernsteiner, line 1064: Unfortunate.  There's discussion somewhere of a `VALID` permission for handling cases like this, but of course it's not implemented at all yet.

c2rust-analyze/src/known_fn.rs
```diff
@@ -602,28 +622,823 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
...
+            fn stat(
+                path: *const c_char: [READ | OFFSET_ADD | NON_NULL],
+                buf: *mut stat: [WRITE | NON_NULL],
+            ) -> c_int;
+
+            fn strchr(
+                // `WRITE` because the return type is derived from `cs`'s provenance.
+                cs: *const c_char: [READ | WRITE | OFFSET_ADD | NON_NULL],
+                c: c_int,
+            ) -> *mut c_char: [READ | WRITE | OFFSET_ADD];
```
spernsteiner, line 1316: At some point it might be nice to make functions like this pass-through certain permissions like `ptr::offset` does.  If the returned pointer is never written, there's no need to require `WRITE` on the argument.

spernsteiner (approved):
Looks good overall.

c2rust-analyze/src/known_fn.rs
```diff
@@ -602,28 +622,823 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
...
+                src: *const c_void: [READ | OFFSET_ADD | NON_NULL],
+                n: size_t,
+                // Same as `dest`, but while `dest` shouldn't be `NULL`,
+                // it can without error, and thus this can return `NULL`.
+            ) -> *mut c_void: [WRITE | OFFSET_ADD];
+
+            fn memmove(
+                // TODO(kkysen) Note: A `&mut []` for `dest` and `&[]` for `src` would be UB since they can overlap.
+                dest: *mut c_void: [WRITE | OFFSET_ADD | NON_NULL],
+                src: *const c_void: [READ | OFFSET_ADD | NON_NULL],
```
spernsteiner, line 1032: I don't think this is actually a problem.  We can't safely produce overlapping `&mut [T]` and `&[T]`, so we would instead end up with two `&[Cell<T>]` references in possibly-overlapping cases.

kkysen: Ah, that makes more sense.  I'll add this as a comment, too.

c2rust-analyze/src/known_fn.rs
kkysen: That would be very nice.  Basically being generic over mutability (and potentially other permissions, too), right?  We'd have to modify how we do `visit_local_call` with `LFnSig`s, though, so it'll be a ways away, and we'd have to add a way to annotate them as such.  It might make sense to have multiple declarations of the same function with different annotations, the way there are often `foo` and `foo_mut` `fn`s, and then resolve that at call sites (I was thinking of a similar thing for vararg `fn`s).  But that's definitely a question for later.  Want me to open an issue for it so we don't forget later on?

c2rust-analyze/src/known_fn.rs
kkysen: Yeah, it's unfortunate.  I'll leave the `TODO` and if we eventually add a `VALID` permission, we can handle it then.

c2rust-analyze/src/known_fn.rs
kkysen: That sounds like a potentially good way.  I'll add some of that to the comment and leave it at that for now if that sounds good.

c2rust-analyze/src/known_fn.rs
kkysen: Should `NON_NULL` be included in that?  I'd imagine not, but then it's not saving a ton over just writing `READ | OFFSET_ADD`.  And `READ | OFFSET_ADD` could just be `SLICE` to be more generic than `STR`.

c2rust-analyze/src/known_fn.rs
kkysen: I'll open an issue for this, too.  There are also ones like `glob`/`globfree` and `mmap`/`munmap`.  That said, even if we don't track them correctly, memory leaks are safe, and we'll be doing the same memory management as was in C.  It'd be similar to any other resource as well, like `open`, etc. and `close`.

c2rust-analyze/src/known_fn.rs
spernsteiner: I was thinking `READ | OFFSET_ADD | NON_NULL`, since that seems to be the most common case

c2rust-analyze/src/known_fn.rs
kkysen: Opened #1006.

c2rust-analyze/src/known_fn.rs
kkysen: * Opened #1007.

c2rust-analyze/src/known_fn.rs
kkysen: I'm not sure if it's worth it to do this.  If I do `STR = READ | OFFSET_ADD | NON_NULL`, then for a nullable string, I'd have to do `READ | OFFSET_ADD`, which I think is confusing.  And it's not saving so much repetition; we're only saving a max of 2 permissions per ptr.

If we ever add something like `NULL_TERMINATED` as a flag or something, then it would make more sense in that case.

c2rust-analyze/src/known_fn.rs
```diff
@@ -602,28 +622,823 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
...
+            fn bind(
+                socket: c_int,
+                address: *const sockaddr: [READ],
+                address_len: socklen_t,
+            ) -> c_int;
+
+            fn calloc(
+                nobj: size_t,
+                size: size_t,
+            ) -> *mut c_void: [READ | WRITE | OFFSET_ADD | FREE];
```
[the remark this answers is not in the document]
kkysen: > I think the return type here should actually have `OFFSET_SUB` in addition to `OFFSET_ADD`. It's legal and reasonable to do things like `calloc(...).offset(1).offset(-1)`, and it seems difficult to specify something like "you can do `OFFSET_SUB` on this pointer, but only if you do an `OFFSET_ADD` first".

Why couldn't we have something like this for `offset`?

```rust
fn offset<T>(self: *const T: [OFFSET_ADD], count: isize) -> *const T: [OFFSET_ADD | OFFSET_SUB];
```

> (Though really `calloc` should probably be special-cased along with `malloc`, `realloc`, etc at some point.)

I think it already is.  I just added all of the `libc` `fn`s used by `lighttpd_rust_amalgamated` for consistency, even if they are already builtins.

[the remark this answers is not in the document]
kkysen: In dc4c467a, I made the `fn`s that internally do `.offset`s return `OFFSET_SUB`, which I think is clearly correct (e.x. for `strchr`).  I'm still not sure about the others, though.  From what I understood, if I pointer has only `OFFSET_ADD`, it can be made into a slice.  But if it has `OFFSET_SUB`, too, then it can't become a slice.  When a `fn` returns a pointer to the beginning of memory that could be a slice, why would we add `OFFSET_SUB`?

spernsteiner (approved):

PR #999, opened 2023-07-19
(`c2rust-analyze`) Use `SubstsRef` when creating an `Instance` to get the `SymbolName` so that we don't ICE on generic foreign `fn`s like `extern "rust-intrinsic"`s

### LINKED ISSUE #996: analyze: ICE on pointer transmute
```Rust
unsafe fn f(x: *const u8) -> *const i8 {
    std::mem::transmute(x)
}
```

Triggers an internal compiler error:
```
visit_terminator(_0 = std::intrinsics::transmute::<*const u8, *const i8>(move _2) -> [return: bb1, unwind: bb2])
unsafe extern "rust-intrinsic" fn(*const u8) -> *const i8 {std::intrinsics::transmute::<*const u8, *const i8>} is trivial: false
name: "transmute"
callee = UnknownDef(Direct { ty: unsafe extern "rust-intrinsic" fn(*const u8) -> *const i8 {std::intrinsics::transmute::<*const u8, *const i8>}, def_id: DefId(2:1760 ~ core[480a]::intrinsics::{extern#1}::transmute), substs: [*const u8, *const i8], is_foreign: true })
error: internal compiler error: compiler/rustc_middle/src/ty/instance.rs:321:17: Instance::mono: DefId(2:1760 ~ core[480a]::intrinsics::{extern#1}::transmute) has type parameters

thread 'rustc' panicked at 'Box<dyn Any>', /rustc/d394408fb38c4de61f765a3ed5189d2731a1da91/compiler/rustc_errors/src/lib.rs:1392:9
stack backtrace:
   0: std::panicking::begin_panic::<rustc_errors::ExplicitBug>
   1: std::panic::panic_any::<rustc_errors::ExplicitBug>
   2: <rustc_errors::HandlerInner>::bug::<&alloc::string::String>
   3: <rustc_errors::Handler>::bug::<&alloc::string::String>
   4: rustc_middle::ty::context::tls::with_context_opt::<rustc_middle::ty::context::tls::with_opt<rustc_middle::util::bug::opt_span_bug_fmt<rustc_span::span_encoding::Span>::{closure#0}, ()>::{closure#0}, ()>
   5: rustc_middle::util::bug::opt_span_bug_fmt::<rustc_span::span_encoding::Span>
   6: rustc_middle::util::bug::bug_fmt
   7: <rustc_middle::ty::list::List<rustc_middle::ty::subst::GenericArg>>::fill_item::<<rustc_middle::ty::instance::Instance>::mono::{closure#0}>
   8: <rustc_middle::ty::list::List<rustc_middle::ty::subst::GenericArg>>::for_item::<<rustc_middle::ty::instance::Instance>::mono::{closure#0}>
   9: <rustc_middle::ty::instance::Instance>::mono
  10: c2rust_analyze::context::GlobalAnalysisCtxt::known_fn
             at ./src/context.rs:699:43
  11: c2rust_analyze::dataflow::type_check::TypeChecker::visit_call
             at ./src/dataflow/type_check.rs:403:19
```

`transmute` is declared in an `extern` block, but has type parameters, which seems to be confusing our attempt to get its symbol name for checking against known libc functions.  We should probably just exclude `Abi::RustIntrinsic` extern fns from the normal `known_fn` checking.

kkysen:
* Fixes #996.

Use `SubstsRef` when creating an `Instance` to get the `SymbolName` so that we don't ICE on generic foreign `fn`s like `extern "rust-intrinsic"`s.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -29,6 +29,8 @@ use rustc_middle::ty::FieldDef;
 use rustc_middle::ty::GenericArgKind;
 use rustc_middle::ty::GenericParamDefKind;
 use rustc_middle::ty::Instance;
+use rustc_middle::ty::List;
+use rustc_middle::ty::SubstsRef;
 use rustc_middle::ty::Ty;
 use rustc_middle::ty::TyCtxt;
 use rustc_middle::ty::TyKind;
@@ -685,8 +687,8 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         self.fns_failed.keys().copied()
     }
 
-    pub fn known_fn(&self, def_id: DefId) -> Option<&'static KnownFn> {
-        let symbol = self.tcx.symbol_name(Instance::mono(self.tcx, def_id));
+    pub fn known_fn(&self, def_id: DefId, substs: SubstsRef<'tcx>) -> Option<&'static KnownFn> {
+        let symbol = self.tcx.symbol_name(Instance::new(def_id, substs));
         self.known_fns.get(symbol.name).copied()
     }
 
@@ -707,7 +709,7 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             .filter(|(def_id, _)| {
                 self.tcx.def_kind(self.tcx.parent(**def_id)) == DefKind::ForeignMod
             })
-            .filter_map(|(&def_id, fn_sig)| Some((fn_sig, self.known_fn(def_id)?)))
+            .filter_map(|(&def_id, fn_sig)| Some((fn_sig, self.known_fn(def_id, List::empty())?)))
             .flat_map(|(fn_sig, known_fn)| known_fn.ptr_perms(fn_sig))
     }
 }
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -400,7 +400,7 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 def_id,
                 substs,
                 is_foreign: true,
-            }) if self.acx.gacx.known_fn(def_id).is_some() => {
+            }) if self.acx.gacx.known_fn(def_id, substs).is_some() => {
                 // As this is actually a known `fn`, we can treat it as a normal local call.
                 self.visit_local_call(def_id, substs, args, destination);
             }


kkysen:
@spernsteiner, I'm not sure if this is the right approach here yet, but I'd like to see if it fixes the issue at least.  I'm looking a bit more into how `Instance` and `fn compute_symbol_name` work now.

spernsteiner:
Seems like a reasonable approach.  Though I will say I don't think it's worth trying to extend the `known_fn` machinery to handle generic functions like `transmute`.  We should just bail out on those, one way or another.  With the current changes, we bail out because `symbol_name` (I assume) produces some mangled name that won't match anything in the permission lookup tables, but we could also bail out earlier based on the `substs` being non-empty or by looking at the ABI (the result of `tcx.fn_sig(def_id)` include an `abi` field).

kkysen:
> With the current changes, we bail out because `symbol_name` (I assume) produces some mangled name that won't match anything in the permission lookup tables

Yeah, that was the plan.  But as you said, there might be better ways to do this.  Maybe just check if `substs.is_empty()`.  For checking the ABI, there are a lot of different [`Abi`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_target/spec/abi/enum.Abi.html)s.  We'd presumably want to skip any of the Rust-like ones*, i.e. ones that could be generic.  Should we also skip these ones (like `rust-intrinsic`) in `fn gather_foreign_sigs` in the first place?

The possibly generic Rusty ones, from what I can guess:
* `"Rust"`
* `"rust-intrinsic"`
* `"rust-call"`
* `"rust-cold"`

spernsteiner:
> Maybe just check if `substs.is_empty()`.

One thing I'm unsure about is whether this would work with `known_fn_ptr_perms`, which calls `known_fn` but doesn't have a `substs` available to check.



> We'd presumably want to skip any of the Rust-like ones*, i.e. ones that could be generic.

`extern "Rust" { ... }` doesn't allow generic functions in the extern block.  I think the only ABI that allows that is `extern "rust-intrinsic" { ... }` (which is very much a special case).



> Should we also skip these ones (like `rust-intrinsic`) in `fn gather_foreign_sigs` in the first place?

`gather_foreign_sigs` feeds into the "mark FFI-exposed types `FIXED`" pass, right?  We could potentially skip `rust-intrinsic`s there, since they shouldn't have any local (= potentially rewritable) types in their signature anyway.  But `extern "Rust" { ... }` should generally be treated about the same as `extern "C" { ... }`.

kkysen:
> `extern "Rust" { ... }` doesn't allow generic functions in the extern block. I think the only ABI that allows that is `extern "rust-intrinsic" { ... }` (which is very much a special case).

It doesn't?  I didn't realize that.  Do you know where it documents that?  I've had trouble finding much documentation about the different ABIs in general.

spernsteiner:
> Do you know where it documents that?

No idea, but if you try something like `extern "Rust" { fn foo<T>(x: T); }`, you get "error[[E0044]](https://doc.rust-lang.org/stable/error_codes/E0044.html): foreign items may not have type parameters".  Also, it's not clear what it would mean to have such a thing, or how you would implement linking (especially dynamic linking) for one.  `extern "rust-intrinsic"` only allows it because each `rust-intrinsic` function is special-cased within the compiler.

PR #1000, opened 2023-07-19
wip function hypothetical lifetimes

aneksteind:


c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -613,28 +613,138 @@ pub fn gen_ty_rewrites<'tcx>(
...
         .unwrap_or_else(|| panic!("expected def {:?} to be a function", ldid));
 
+    let (origin_params, input_origin_args, output_origin_args) =
+        &acx.gacx.fn_origins.fn_info[&ldid.to_def_id()];
+    let hir_generics = acx.tcx().hir().get_generics(ldid);
+
+    let predicates = acx.tcx().predicates_of(ldid);
+
+    if let Some(generics) = hir_generics {
+        let mut generics_rws: Vec<Rewrite> = vec![];
```
spernsteiner, line 623: This approach, particularly the handling of predicates, seems fragile and also a bit complex.  It would be better to emit a rewrite that only injects the new parameters and leaves everything else unchanged (which includes any existing predicates, as in `fn f<'a, 'b: 'a>(...)`).  We should look through the HIR generics to find the right insertion point, then emit a zero-width rewrite that adds `'h0, ..., 'hN` at that point.  The logic for this is a little nontrivial - the `Span`s are scattered across the HIR `GenericParam`s and `WherePredicate`s, and it needs to handle leading/trailing commas properly when there are no regions, no tys, or no params at all (`fn foo<>(...)` is legal) - but overall, I think it will be cleaner since we won't need to regenerate any predicates.

Functions with no parameter list at all will need to have one added, but that should be straightforward since in that case there are no predicates to worry about.

c2rust-analyze/src/labeled_ty.rs
```diff
@@ -316,7 +316,31 @@ impl<'tcx, L: Copy> LabeledTyCtxt<'tcx, L> {
     where
         F: FnMut(LabeledTy<'tcx, L1>, LabeledTy<'tcx, L2>, &'tcx [LabeledTy<'tcx, L>]) -> L,
     {
-        assert_eq!(lty1.ty, lty2.ty);
+        // If we're comparing two type where one has late-bound regions but the
```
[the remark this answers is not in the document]
spernsteiner: Not sure I understand.  What's wrong with having `'_` in function signatures?  I assume you mean erasing late-bound regions on the signatures we put into `gacx.fn_sigs`?

Either way, I think you could do the region-ignoring comparison here by calling `replace_escaping_bound_vars_uncached` with an appropriate callback function.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -457,9 +461,92 @@ impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
...
+        for (predicate, predicate_span) in predicates.predicates {
+            if matches!(
+                predicate.kind().skip_binder(),
+                PredicateKind::RegionOutlives(OutlivesPredicate(a, b)) if a != b
+            ) && (last_lifetime_span.is_none()
+                || last_lifetime_span.unwrap().hi() < predicate_span.hi())
+            {
+                last_lifetime_span = Some(*predicate_span)
+            }
+        }
```
spernsteiner, line 484: I think this might malfunction on code with an `OutlivesPredicate` in a separate where clause:
```Rust
fn foo<'a, 'b>(x: &'a i32, y: &'b i32)
where 'a: 'b {
}
```
It will set `last_lifetime_span` to the end of the where clause rather than the end of the generic parameter list.

If you look at the HIR `generics.predicates` instead, the [`WhereRegionPredicate`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/hir/struct.WhereRegionPredicate.html) case includes an `in_where_clause` flag you can use to filter these out.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -457,9 +461,92 @@ impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
...
+                    first_generics_span.shrink_to_lo(),
+                    format!("{},", hypothetical_origin_string),
+                ),
+                _ if generics.params.is_empty() => {
+                    (generics.span, format!("<{}>", hypothetical_origin_string))
+                }
+                _ => (generics.span, format!("{}", hypothetical_origin_string)),
+            };
+            self.hir_rewrites
+                .push((hypothetical_origin_span, Rewrite::PrintTy(format_string)));
```
spernsteiner, line 544: Seems like a slight abuse of `PrintTy`, though I guess we can live with it (or maybe rename to something that sounds more general-purpose, like `Print` or `String`)

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -457,9 +461,92 @@ impl<'a, 'tcx> HirTyVisitor<'a, 'tcx> {
...
+                    format!(",{}", hypothetical_origin_string),
+                ),
+                (_, Some(first_generics_span), _) | (_, _, Some(first_generics_span)) => (
+                    first_generics_span.shrink_to_lo(),
+                    format!("{},", hypothetical_origin_string),
+                ),
+                _ if generics.params.is_empty() => {
+                    (generics.span, format!("<{}>", hypothetical_origin_string))
+                }
+                _ => (generics.span, format!("{}", hypothetical_origin_string)),
```
spernsteiner, line 541: I think the `if` condition here is wrong.  If `last_lifetime_span`, `first_ty_span`, and `first_const_span` are all `None`, then we already know the parameter list is empty.  What we actually need to distinguish here are `fn foo()` vs `fn foo<>()`.  Both will have an empty `params` list, so my guess is the only way to distinguish these is to check whether `generics.span` is empty or not (`lo() == hi()`, or maybe there's a `Span` method to check for this specifically).

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: Looks about right.  More specific comments below.

spernsteiner (approved):
Looks about right to me.

Regarding the conflict: in #1005 (merged yesterday), I pulled the struct rewriting logic out of `HirTyVisitor::visit_item` to avoid O(N^2) behavior (each time we ran rewriting on a function, it also generated all struct rewrites) and give `main.rs` more control over what gets rewritten.  You'll probably have to move the new `gen_generics_rws` calls out of `visit_item` as well.

Issue #997, opened 2023-07-18
analyze: when rewriting, omit unused hypothetical regions


spernsteiner:
Sometimes, we add hypothetical regions to a type, but the rewriting we choose for the field types results in those region parameters going unused:

```Rust
struct S {
    ptr: *mut u8,
}

#[c2rust_analyze_test::fixed_signature]
unsafe fn f(x: *mut *mut u8) {}

unsafe fn g(s: *mut S) {
    f(ptr::addr_of_mut!((*s).ptr));
}
```

Rewritten output:
```Rust
struct S<'h0> {
    ptr: *mut u8,
}

#[c2rust_analyze_test::fixed_signature]
unsafe fn f(x: *mut *mut u8) {}

unsafe fn g(s: &(S)) {
    f(core::ptr::addr_of!(*&mut ((*s).ptr)).cast_mut());
}
```

Here, we add a hypothetical region param `'h0` to `S` for it to use in the type of `ptr`, but we end up leaving `ptr` as a raw pointer type, which doesn't need a region.  This causes rustc to report an error: "parameter `'h0` is never used".  We would have similar issues if we chose to rewrite `ptr` into `Box` or `Rc`.

We need a postprocessing pass that can examine the rewrites (`TypeDesc`s) for ADT field types and decide which of the hypothetical regions params should actually be emitted during rewriting.

aneksteind:
@spernsteiner i'm happy to take this one once i'm done with other tasks, if you want to re-assign to me

spernsteiner:
I self-assigned this first because it's the most immediate blocker for being able to filter exactly what gets rewritten (currently lifetimes are always added) and I plan to start on it immediately, and second because I'd like to get a better understanding of how exactly lifetime parameters are handled at the moment, since we'll need to extend that sometime soon to work with functions and possibly lifetime bounds (as in `'a: 'b`).

aneksteind:
Sounds good, and as an FYI I've been working on lifetimes in functions

PR #1013, opened 2023-08-02
Prevent foreign-mentioned field types from getting their hypothetical origins generated / propagated to encapsulating structures

### LINKED ISSUE #994: analyze: rewriter adds hypothetical lifetime args to mentions of non-rewritten structs
```Rust
extern "C" {
    fn epoll_wait(events: *mut epoll_event);
}

pub struct fdevents {
    pub epoll_events: *mut epoll_event,
}

pub struct epoll_event {
    pub ptr: *mut u8,
}
```

Rewritten output:
```Rust
extern "C" {
    fn epoll_wait(events: *mut epoll_event);
}

pub struct fdevents<'h4,'h5> {
    pub epoll_events: &'h4 (epoll_event<'h5>),
}

pub struct epoll_event {
    pub ptr: *mut u8,
}
```

`struct epoll_event` can't be rewritten, since it's exposed to FFI via `fn epoll_wait`.  But the field type in `fdevents` is rewritten as if a lifetime parameter was added to `epoll_event`.

oinoom:
Fixes #994

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -53,7 +53,28 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         self.ltcx.relabel(
                 field_metadata.origin_args,
                 &mut |flty| match flty.kind() {
-                    TyKind::Ref(..) | TyKind::RawPtr(..) => {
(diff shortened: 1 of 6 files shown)


spernsteiner (approved):
Seems fine as a temporary measure.  Eventually this will be subsumed by a fix for #997.

PR #1015, opened 2023-08-03
analyze: omit unused hypothetical lifetimes during rewriting

spernsteiner:
This change removes hypothetical lifetimes for pointers that are rewritten into non-ref types, such as raw pointers or `Box<T>`.  The `AdtMetadataTable` is computed normally at first so it can be used in borrowck, but once all pointer permissions have been determined, it's recomputed with an extra filter that skips creating hypothetical region params for pointers whose `type_desc::Ownership` is not `Imm`, `Cell`, or `Mut` (`&T`, `&Cell<T>`, or `&mut T`).

For example:
```Rust
struct Foo {
   p: *mut u8,
   #[c2rust_analyze_test::fixed_signature]
   q: *mut u8,
}
```
Previously, `Foo` would be rewritten to have two new region parameters, one for `p` and one for `q`.  But the region for `q` would unused: `q` is marked `FIXED`, so it's left as `*mut u8` rather than rewritten to a safe reference type.  With this change, `Foo` is rewritten to have only one region parameter, the one for `p`.  This avoids rustc errors about unused generic parameters in the rewritten code.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -217,6 +217,7 @@ bitflags! {
     }
 }
 
+#[derive(Default)]
 pub struct AdtMetadataTable<'tcx> {
     pub table: HashMap<DefId, AdtMetadata<'tcx>>,
     pub struct_dids: Vec<DefId>,
@@ -392,6 +393,7 @@ pub struct FnSigOrigins<'tcx> {
     pub output: LabeledTy<'tcx, &'tcx [OriginArg<'tcx>]>,
 }
 
+#[derive(Default)]
 pub struct FnOriginMap<'tcx> {
     pub fn_info: HashMap<DefId, FnSigOrigins<'tcx>>,
 }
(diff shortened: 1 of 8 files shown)


kkysen (commented):
For 5a88caa, were you seeing known ptr permissions change?  Which ones?

spernsteiner:
> For [5a88caa](https://github.com/immunant/c2rust/commit/5a88caa4c0d734cf52c067006a21921276207c2e), were you seeing known ptr permissions change? Which ones?

Opened #1017 with the specific cases I saw in lighttpd

spernsteiner:
>are we somehow constraining things, resulting in a different set of rewrites, by giving hypothetical lifetimes to polonius that won't end up being used?

For variables that get rewritten to `Box`/`Rc`, the presence or absence of `UNIQUE` doesn't affect the rewrite.  So I guess the concern here would be that including these lifetimes could add constraints on other variables where `UNIQUE` does matter.  I don't know for sure whether or not this can happen, but I would guess not - borrowing the data inside a box is pretty similar from a borrowck perspective to borrowing the data behind an `&mut`.

If this does turn out to be an issue, we can think about borrowck improvements then (and having a real example to look at will help).

PR #1019, opened 2023-08-09
handle `deconstruct_hir_ty` for function pointers

### LINKED ISSUE #992: analyze: missing hypothetical lifetimes within function pointer types
```Rust
struct NeedsLifetime {
    p: *mut u8,
}

struct Vtable {
    copy: Option<unsafe extern "C" fn(*const NeedsLifetime) -> *mut NeedsLifetime>,
}
```

Rewritten output:
```Rust
struct NeedsLifetime<'h1> {
    p: &'h1 (u8),
}

struct Vtable<'h2,'h1,'h3> {
    copy: Option<unsafe extern "C" fn(*const NeedsLifetime) -> *mut NeedsLifetime>,
}
```

Notice the uses of `NeedsLifetime` within the type of `Vtable::copy` don't pass the necessary lifetime arguments.

oinoom:
Fixes #992. Prior to this we had no support for deconstructing the HIR ty of a function to get its arguments, which caused `handle_ty` to not generate any rewrites for function pointers in struct fields. This adds support for deconstructing the signatures by first asserting that the inputs and output of an HIR function signature match the length of the MIR function signature and then packages the inputs and outputs of the HIR signature into a vector.

### DIFF
--- a/c2rust-analyze/src/rewrite/ty.rs
+++ b/c2rust-analyze/src/rewrite/ty.rs
@@ -404,6 +404,7 @@ fn rewrite_ty<'tcx>(
     hir_rewrites: &mut Vec<(Span, Rewrite)>,
     rw_lty: RwLTy<'tcx>,
     hir_ty: &hir::Ty<'tcx>,
+    adt_metadata: &AdtMetadataTable,
 ) {
     if !rw_lty.ty.is_adt() && rw_lty.label.ty_desc.is_none() && !rw_lty.label.descendant_has_rewrite
     {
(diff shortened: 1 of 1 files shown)


c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -225,6 +227,18 @@ fn deconstruct_hir_ty<'a, 'tcx>(
...
+        (&ty::TyKind::FnPtr(sig), &hir::TyKind::BareFn(bfnty)) => {
+            let args = sig.skip_binder().inputs_and_output;
+            if args.len() != bfnty.decl.inputs.len() + 1 {
+                panic!("mismatched number of function inputs for {sig:?} and {bfnty:?}");
+            }
+
+            let mut v: Vec<_> = bfnty.decl.inputs.iter().collect();
+            if let FnRetTy::Return(return_ty) = bfnty.decl.output {
+                v.push(return_ty)
+            }
```
spernsteiner, line 239: Why's it okay for the length of `v` to vary here?  Is the caller doing a `zip` and we're relying on the implicit truncation?

If we were to emit a rewrite on the implicit `()` return type of some function pointer, will the rewrite properly fail, due to not having anywhere to apply the rewrite?

spernsteiner (approved):

PR #1020, opened 2023-08-11
Test.lighttpd.md5

oinoom:
Initial `algo_md5` module test, with offending portions of the module commented out for now, until they are addressed

### DIFF
--- a/c2rust-analyze/src/c_void_casts.rs
+++ b/c2rust-analyze/src/c_void_casts.rs
@@ -1,7 +1,7 @@
 use std::borrow::Borrow;
 use std::collections::{HashMap, HashSet};
 
-use rustc_middle::mir::LocalKind;
+use rustc_middle::mir::{BasicBlock, LocalKind};
 use rustc_middle::{
     mir::{
         BasicBlockData, Body, LocalDecls, Location, Place, Rvalue, Statement, StatementKind,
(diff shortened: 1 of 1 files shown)


c2rust-analyze/src/c_void_casts.rs
```diff
@@ -403,59 +426,99 @@ impl<'tcx> CVoidCasts<'tcx> {
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /// Performs BFS in the CFG to find where the c_void_ptr
+    /// was cast to that type
+    fn find_and_insert_pred_cast(
```
spernsteiner, line 490: I'm not a fan of using BFS for this - it seems error prone.  Is it the case that the `CVoidPtr` `Place` is always a temp local, which is written only once?  I think it should be for code like `memset(x as *mut c_void, ...)`.  If there's only one assignment, then we don't need to worry about where exactly the cast is located relative to the call, and instead we can just scan the whole body to find it.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -494,7 +494,80 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 let perms = PermissionSet::FREE;
                 self.constraints.add_all_perms(rv_lty.label, perms);
             }
+            Callee::Memcpy => {
```
spernsteiner, line 497: At some point we should extend this to include a pseudo-assignment `*dest = *src;`.  We expect to produce safe rewrites for at least some uses of memcpy, and in those cases we should make sure `dataflow` treats it like the safe version (particularly in cases where `*src` is a pointer or contains one).

c2rust-analyze/src/rewrite/expr/distribute.rs
```diff
@@ -65,7 +65,7 @@ pub fn distribute(
             let origin = match unlower_map.get(&key) {
                 Some(x) => x,
                 None => {
-                    error!("unlower_map has no origin for {:?}", key);
+                    error!("unlower_map has no origin for {:?} {:?}", key, unlower_map);
```
spernsteiner, line 68:
```suggestion
                    error!("unlower_map has no origin for {:?} {:?}", key);
```
Should this be reverted?  It looks like a debug print that would be incredibly verbose in normal usage.

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: `const` vs `mut` maybe matters for `dest = src`, but it shouldn't matter for `*dest = *src`, which is effectively what's happening in `memcpy` (at least when the number of bytes to copy is equal to `size_of::<T>()`).

spernsteiner: Yes, that seems about right

spernsteiner (approved):
Looks good with the recent fixes

Issue #977, opened 2023-07-05
bitfields-derive: Upgrade to syn 2


langston-barrett:
syn is a big package that takes a long time to compile, and it recently had a major version bump. c2rust-bitfields-derive is a transitive dependency of my application that relies on syn 1, meaning I have to compile syn twice to compile my application. It would be great if bitfields-derive supported syn 2, so I only had to build syn once.

domenukk:
+1

bungcip:
current version of c2rust-bitfields-derived already upgraded to syn 2. I think we can close this issue.

kkysen:
* Fixed by #1271.

PR #1021, opened 2023-08-24
Update Syn dependency

domenukk:
Let's see what breaks (tries to address #977 )


domenukk:
Oh well, seems like a lot changed..
https://github.com/dtolnay/syn/releases/tag/2.0.0

PR #1030, opened 2023-09-29
`c2rust transpile`: When casting `bool`s to floats, go through the integral type `u8`

dgherzka:
Previously, C code that cast `bool`s to floating types, like this

```c
#include <stdbool.h>

void cast_stuff(void) {
    bool b = true;
    float x15 = b;
}
```

would try to do so directly in Rust, like this

```rust
#[no_mangle]
pub unsafe extern "C" fn cast_stuff() {
    let mut b: bool = 1 as libc::c_int != 0;
    let mut x15: libc::c_float = b as libc::c_float;
}
```

which isn't allowed, resulting in errors like this

```shell
error[E0606]: casting `bool` as `f32` is invalid
  --> src/casts.rs:31:34
   |
31 |     let mut x15: libc::c_float = b as libc::c_float;
   |                                  ^^^^^^^^^^^^^^^^^^
   |
   = help: cast through an integer first
```

This fixes things by emitting this Rust instead by casting through the integral type `u8`:

```rust
#[no_mangle]
pub unsafe extern "C" fn cast_stuff() {
    let mut b: bool = 1 as libc::c_int != 0;
    let mut x15: libc::c_float = b as u8 as libc::c_float;
}
```

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4275,6 +4275,10 @@ impl<'c> Translation<'c> {
                     let expr =
                         expr.ok_or_else(|| format_err!("Casts to enums require a C ExprId"))?;
                     Ok(self.enum_cast(ty.ctype, enum_decl_id, expr, val, source_ty, target_ty))
+                } else if target_ty_ctype.is_floating_type() && source_ty_kind.is_bool() {
+                    val.and_then(|x| {
+                        Ok(WithStmts::new_val(mk().cast_expr(mk().cast_expr(x, mk().path_ty(vec!["u8"])), target_ty)))
+                    })
                 } else {
                     // Other numeric casts translate to Rust `as` casts,
                     // unless the cast is to a function pointer then use `transmute`.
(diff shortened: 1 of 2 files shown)


dgherzka:
It wasn't clear to me whether this should be implemented as a new CastKind (BooleanToFloating) or treated as a special case of IntegralToFloating.

dgherzka:
Hi @kkysen, pinging you because I see that you are a major contributor to this project. I apologize if you aren't the right person to talk to. What's the process for getting this reviewed? Thanks in advance!

dgherzka:
Sorry for continuing to tag people, I just want to try one more time. @spernsteiner are you the right person to talk to?

kkysen:
Hi @dgherzka!  Sorry for taking a while to review this; I saw your ping but forgot to take a look.  I'll review it now, and your other PRs and issues.

kkysen (approved):
Thanks for this PR!  Sorry we took so long to review it, but we really appreciate it; thanks for fixing this bug!

> It wasn't clear to me whether this should be implemented as a new CastKind (BooleanToFloating) or treated as a special case of IntegralToFloating.

`bool` technically is an integral type, so I think a case of `IntegralToFloating` should be fine, and the new code handling this case fits in nicely.

And the simple test case looks good and is enough to catch this bug, so everything LGTM.  Thanks!

kkysen (commented):
As for casting through `u8` to go `bool` => `u8` => floating type, I'm not sure if casting through `c_int` would be more correct in terms of what C is supposed to do, but casting through `u8` should always be correct, and it seems more idiomatic for Rust, so I think that's fine.  @thedataking, do you think that's fine?  Are there any subtleties in choosing the intermediary integral type?

PR #1031, opened 2023-10-04
dataflow: avoid propagating offset permissions upward from field access

oinoom:
Pointers to structs with array fields being iterated over via `offset` were erroneously being given `OFFSET` permissions. Those permissions should be limited to the fields themselves. This PR introduces a new dataflow constraint that will propagate permissions except for a specified set. When a field projection is detected, offset permissions are no longer propagated to the base pointer.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -945,7 +945,16 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
         }
     }
 
-    /// Returns the [`LTy`] of an [`Rvalue`] and a boolean indicating whether or
+    /// Returns the [`LTy`] of an [`Rvalue`]
+    pub fn type_of_rvalue(&self, rv: &Rvalue<'tcx>, loc: Location) -> LTy<'tcx> {
+        if let Some(lty) = self.rvalue_tys.get(&loc) {
+            return lty;
+        }
+
+        self.derived_type_of_rvalue(rv)
+    }
+
+    /// Returns a boolean indicating whether or
     /// not the `Rvalue` is a reference or pointer containing a field projection.
     /// For example, the following [`Rvalue`]s will satisfy that criteria:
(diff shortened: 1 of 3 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -945,9 +945,12 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
         }
     }
 
-    pub fn type_of_rvalue(&self, rv: &Rvalue<'tcx>, loc: Location) -> LTy<'tcx> {
+    /// Returns the [`LTy`] of an [`Rvalue`] and a boolean indicating whether or
+    /// not the `Rvalue` contains a field projection.
```
spernsteiner, line 949: Can you clarify the precise meaning of "the `Rvalue` contains a field projection"?  Which of these count as containing a field projection?

```Rust
let r1 = x.field;
let r2 = &x.field;
let r3 = &(*p).field;
let r4 = x.field + y;
// ^ Rvalue::Add(Operand::Copy(x.field), ...)
```

c2rust-analyze/src/dataflow/mod.rs
```diff
@@ -81,6 +89,39 @@ impl DataflowConstraints {
...
+                #[allow(bad_style)]
+                let PROPAGATE_DOWN = PermissionSet::UNIQUE;
+                // Permissions that should be propagated "up": if the subset (`a`) has it,
+                // then the superset (`b`) should be given it.
+                #[allow(bad_style)]
+                let PROPAGATE_UP = PermissionSet::READ
+                    | PermissionSet::WRITE
+                    | PermissionSet::OFFSET_ADD
+                    | PermissionSet::OFFSET_SUB
+                    | PermissionSet::FREE;
```
spernsteiner, line 117: To avoid duplicating these constants, could you implement `fn subset` in terms of `subset_except` (with an empty `except` argument)?

spernsteiner (approved):

PR #1037, opened 2023-11-01
support `c2rust transpile <path/to/*.c>` in lieu of `compile_commands.json`

### LINKED ISSUE #76: Allow `c2rust transpile` to accept a single C file without a `compile_commands.json`
We already support adding compiler flags to a transpiler run, so we can do without compile_commands.json if we're doing a one-off translation of a single file.

### LINKED ISSUE #1033: Lack of support for simple projects or the cc crate
I find the the requirement to use `compile_commands.json` incredibly annoying, because it never fits the types of projects I want to convert to Rust, and also requires use of clunky 3rd party non-Rust dependencies and non-Rust build systems.

It so happens that all C code that I dare to port to Rust is small — only a handful of `.c` files, sometimes literally just a single `.c` file or a header-only dependency. And for such trivial code I use the `cc` crate, not something complicated like `meson`.

If a project is too hard to build with the `cc` crate, I consider it also too big and complex to convert to Rust. So I'm frustrated that c2rust *only* supports the types of projects I *don't* want to convert. When I want to convert literally a single `.c` file that has zero dependencies and compiles with `gcc *.c` without issues, I have to jump through so many hoops. I need to re-learn how to use `cmake`, and I need to deal with series of issues in installing the right version of python, and installing its dependencies, and venvs and whatever else it requires. All of that to pass *a single file path* to `c2rust`.

Please, support `c2rust *.c`.

oinoom:
* Fixes #1033.
* Fixed #76.

This adds support for transpiling C sources without needing to generate a `compile_commands.json` with something such as `bear` or `cmake`. The way it works is that a temporary `temp_compile_commands` file is created. One alternative to this approach would have been to alter the AST exporter to use a `FixedCompilationDatabase` but it seemed simpler to support this feature from the front end.

### DIFF
--- a/c2rust/src/bin/c2rust-transpile.rs
+++ b/c2rust/src/bin/c2rust-transpile.rs
@@ -240,7 +240,8 @@ fn main() {
         .any(|path| path.extension() == Some(std::ffi::OsStr::new("json")))
     {
         // More than one file provided and at least one is a JSON file
-        panic!("Multiple JSON files provided. Exactly one compile_commands.json file should be provided, or a list of source files.");
+        panic!("Compile commands JSON and multiple sources provided.
+                Exactly one compile_commands.json file should be provided, or a list of source files, but not both.");
     } else {
         // Handle as a list of source files
         c2rust_transpile::create_temp_compile_commands(&args.compile_commands)


c2rust-transpile/src/lib.rs
```diff
@@ -227,9 +227,44 @@ fn get_module_name(
...
 /// Main entry point to transpiler. Called from CLI tools with the result of
 /// clap::App::get_matches().
-pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]) {
+pub fn transpile(tcfg: TranspilerConfig, source_or_cc_db: &Path, extra_clang_args: &[&str]) {
     diagnostics::init(tcfg.enabled_warnings.clone(), tcfg.log_level);
 
+    let temp_json_path = "compile_commands.json";
+    let build_dir = get_build_dir(&tcfg, source_or_cc_db);
+
+    let cc_db = if source_or_cc_db.to_string_lossy().ends_with(".c") {
```
fw-immunant, line 236: Use [`Path::extension`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.extension).

c2rust-transpile/src/lib.rs
```diff
@@ -227,9 +227,44 @@ fn get_module_name(
 
 /// Main entry point to transpiler. Called from CLI tools with the result of
 /// clap::App::get_matches().
-pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]) {
+pub fn transpile(tcfg: TranspilerConfig, source_or_cc_db: &Path, extra_clang_args: &[&str]) {
     diagnostics::init(tcfg.enabled_warnings.clone(), tcfg.log_level);
 
+    let temp_json_path = "compile_commands.json";
```
fw-immunant, line 233: This should probably live in a temporary directory, e.g. the one from [`env::temp_dir`](https://doc.rust-lang.org/stable/std/env/fn.temp_dir.html).

fw-immunant (approved):
Looks mostly sane. Commented a few cleanups/improvements inline; the only user-significant one is to use a temporary directory for the generated `compile_commands.json`.

Error messages could be more helpful, e.g. including paths and mentioning what the JSON file is for.

fw-immunant:
This probably should have some corresponding doc/diagnostics changes, e.g. the `"Could not find compile_commands.json file at path: {}"` error in `c2rust-transpile.rs` and the docs/name of `Args.compile_commands` in that file.

c2rust-transpile/src/lib.rs
```diff
@@ -227,10 +227,65 @@ fn get_module_name(
...
+            r#"{{
+                "directory": "{}",
+                "command": "clang {}",
+                "file": "{}"
+            }}"#,
+            build_dir.to_str().unwrap(),
+            file_path.to_str().unwrap(),
+            file_path.to_str().unwrap()
+        )
+    };
```
kkysen, line 256: I think a `#[derive(Serialize)]` on a `struct CompileCommand` would be a lot simpler to use and less error-prone. That way we can leave things as `PathBuf`s, etc.

Also, an `arguments` list field instead of a `command` field is usually what automated tooling seems to generate, as spacing isn't a concern there.

c2rust-transpile/src/lib.rs
```diff
@@ -227,10 +227,65 @@ fn get_module_name(
...
+        for entry in read_dir(build_dir.clone()).unwrap() {
+            if let Ok(entry) = entry {
+                let path = entry.path();
+                let ext = path.extension().unwrap_or_default();
+                if ext == "c" || ext == "h" {
+                    let command = format_compile_command(&build_dir, &path);
+                    temp_json.push(command);
+                }
+            }
+        }
```
kkysen, line 272: Do we need to handroll globbing?  There's a lot of intricacies with that that we're gonna get correct, so it might be better to just accept multiple sources and let the shell do the glob expansion.

c2rust-transpile/src/lib.rs
kkysen: Oh, we already have a `#[derive(Serialize, Deserialize)] struct CompileCmd`: https://github.com/immunant/c2rust/blob/c603b25b517303b2f7fbf81bf00d1daac896e364/c2rust-transpile/src/compile_cmds/mod.rs#L12-L33

c2rust-transpile/src/lib.rs
```diff
@@ -227,10 +227,65 @@ fn get_module_name(
 
 /// Main entry point to transpiler. Called from CLI tools with the result of
 /// clap::App::get_matches().
-pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]) {
+pub fn transpile(tcfg: TranspilerConfig, source_or_cc_db: &Path, extra_clang_args: &[&str]) {
     diagnostics::init(tcfg.enabled_warnings.clone(), tcfg.log_level);
 
-    let lcmds = get_compile_commands(cc_db, &tcfg.filter).unwrap_or_else(|_| {
+    let build_dir = get_build_dir(&tcfg, source_or_cc_db);
```
thedataking, line 233: rather than adding a bunch of code to `transpile`, might it make sense to extract this into its own function? If not, a comment or two might be in order.

c2rust-transpile/src/lib.rs
thedataking: I agree. The shell expands wildcards by default so expecting a list of concrete file paths seems simpler.

thedataking (commented):
Once the CLI is agreed upon, don't forget to update the readme.

LegNeato:
I too think it would be cleaner to make `transpile` take `CompileCmd`. That way tests as well as library users can generate one with whatever content they want and pass it in rather than hitting the fs first. All this file and json stuff is not required to actually transpile, so it feels weird to tie them together.

As an aside, there appears to be an existing project that could be leveraged though I am not sure if it makes sense:

https://github.com/rizsotto/json_compilation_db

kkysen:
> I too think it would be cleaner to make `transpile` take `CompileCmd`. That way tests as well as library users can generate one with whatever content they want and pass it in rather than hitting the fs first. All this file and json stuff is not required to actually transpile, so it feels weird to tie them together.

☝️ 

> As an aside, there appears to be an existing project that could be leveraged though I am not sure if it makes sense:
> 
> https://github.com/rizsotto/json_compilation_db

I think the existing https://github.com/immunant/c2rust/blob/c603b25b517303b2f7fbf81bf00d1daac896e364/c2rust-transpile/src/compile_cmds/mod.rs#L12-L33 we have already does what `json_compilation_db` does.

kkysen (commented):
Another suggestion I thought of: most `clang` tooling that I've seen that accepts compilation databases (e.x. `clang-tidy`) also accepts a list of files as normal arguments, and then after `--`, a list of cflags that applies to all of those files and would go after the `clang <file>` in the `command`/`arguments`.  Following that I think would be best for compatibility and familiarity, and is also a bit more flexible since it allows for other arguments (albeit the same ones for every file).

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -84,9 +84,18 @@ struct Args {
...
-    /// Input compile_commands.json file
-    #[clap()]
-    compile_commands: PathBuf,
+    /// Input compile_commands.json file or path to source files
+    #[clap(long, parse(from_os_str), required = false, conflicts_with = "source")]
+    compile_commands: Option<PathBuf>,
+
+    /// Source files to process, only valid if compile_commands.json path is not provided.
+    #[clap(
+        long = "source",
```
kkysen, line 93:
```suggestion
        long,
```

It defaults to the field name, right?

c2rust-transpile/src/lib.rs
```diff
@@ -19,6 +19,7 @@ use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::process;
 
+use compile_cmds::CompileCmd;
```
kkysen, line 22: Nit:

```suggestion
use crate::compile_cmds::CompileCmd;
```

Using `crate::` is helpful since it's portable to other non `lib.rs` files and it's clear it's not an external dependency.

kkysen (commented):
I want to look at the interface with `clang` we're using.  I'm still not sure why we have to create a temporary compile commands.  If we do, though, I think we should use the `tempfile` crate to do so, as that deletes the file on destruction, so it will get deleted if we panic or return early in between, which very well may happen.  Also, the clap CLI seems okay, but I think it'd be better if we can match clang tooling's API for specifying either a compile commands or a list of sources and a list of flags/other arguments, which is more flexible and consistent with existing tooling.

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -84,9 +84,18 @@ struct Args {
...
+        parse(from_os_str),
+        multiple_values = true,
+        required = false
+    )]
+    source: Vec<PathBuf>,
```
kkysen, line 98:
```suggestion
    sources: Vec<PathBuf>,
```

If it accepts multiple sources I think the argument name should be plural, too.

c2rust-transpile/src/lib.rs
```diff
@@ -225,12 +226,44 @@ fn get_module_name(
...
+pub fn create_temp_compile_commands(sources: &[PathBuf]) -> PathBuf {
+    let temp_path = std::env::temp_dir().join("compile_commands.json");
+    let compile_commands: Vec<CompileCmd> = sources
+        .iter()
+        .map(|source_file| {
+            let absolute_path = fs::canonicalize(source_file)
+                .unwrap_or_else(|_| panic!("Could not canonicalize {}", source_file.display()));
+
+            CompileCmd {
+                directory: absolute_path.parent().unwrap().to_path_buf(),
```
kkysen, line 238: clang tooling's `CommonOptionsParser` (what we call), [uses `.` as the default directory when there's no compilation database](https://clang.llvm.org/doxygen/classclang_1_1tooling_1_1FixedCompilationDatabase.html#a1443b7812e6ffb5ea499c0e880de75fc), so we should probably match that.

thedataking:
> Also, the clap CLI seems okay, but I think it'd be better if we can match clang tooling's API for specifying either a compile commands or a list of sources and a list of flags/other arguments, which is more flexible and consistent with existing tooling.

@kkysen this was supposed to be small and quick task narrowly addresses PR #1033. Please do not add additional work.

thedataking:
> The new interface would either be providing `--compile-commands` or `--source`, e.g. `c2rust transpile --source add/*.h`

PR #1033 asked if we could accept source files instead of a `.json` file, why do we need `--compile-commands` or `--source`? Can't we just assume the user is providing source files if the (base of) the file name isn't exactly `compile_commans.json`? That's how I read the request in the PR.

This PR has received more work and review than I expected. I'd like it to be wrapped up in the simplest possible way that satisfied the original goals and doesn't break or extend the existing interface (unless we must).

thedataking (approved):
LGTM. Should we update the docs?

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -226,19 +226,35 @@ fn main() {
...
-        panic!(
-            "Could not find compile_commands.json file at path: {}",
-            cc_json_path.display()
-        )
-    });
+    let compile_commands = if args.compile_commands.len() == 1
+        && args.compile_commands[0].extension() == Some(std::ffi::OsStr::new("json"))
+    {
+        // Only one file provided and it's a compile commands JSON file
+        args.compile_commands[0].clone()
```
kkysen, line 233: Previously we canonicalized this path.  We should preserve that behavior, right?

PR #1040, opened 2023-11-03
Derive `Debug` for structs that allow it

dgherzka:
This is allowed for any struct that does not contain any unions recursively.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -1628,17 +1628,15 @@ impl<'c> Translation<'c> {
                 }
 
                 // Gather up all the field names and field types
-                let (field_entries, contains_va_list, debuggable) =
+                let (field_entries, contains_va_list, can_derive_debug) =
                     self.convert_struct_fields(decl_id, fields, platform_byte_size)?;
-
-                println!("{name} debuggable = {debuggable}");
 
                 let mut derives = vec![];
                 if !contains_va_list {
                     derives.push("Copy");
                     derives.push("Clone");
                 };
-                if debuggable {
+                if can_derive_debug {
                     derives.push("Debug");
                 }
                 let has_bitfields =
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4275,6 +4275,10 @@ impl<'c> Translation<'c> {
                     let expr =
                         expr.ok_or_else(|| format_err!("Casts to enums require a C ExprId"))?;
                     Ok(self.enum_cast(ty.ctype, enum_decl_id, expr, val, source_ty, target_ty))
+                } else if target_ty_ctype.is_floating_type() && source_ty_kind.is_bool() {
+                    val.and_then(|x| {
+                        Ok(WithStmts::new_val(mk().cast_expr(mk().cast_expr(x, mk().path_ty(vec!["u8"])), target_ty)))
+                    })
                 } else {
                     // Other numeric casts translate to Rust `as` casts,
                     // unless the cast is to a function pointer then use `transmute`.
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -1631,10 +1631,22 @@ impl<'c> Translation<'c> {
                 let (field_entries, contains_va_list) =
                     self.convert_struct_fields(decl_id, fields, platform_byte_size)?;
 
+                println!("field_entries = {field_entries:#?}");
+                println!("ast_context = {:#?}", self.ast_context);
+
+
+                // for f in field_entries {
+                // self.ast_context.resolve_type(f.ty)
+                // }
+
                 let mut derives = vec![];
                 if !contains_va_list {
                     derives.push("Copy");
                     derives.push("Clone");
+                    // WIP: only if possible
+                    // - no unions
+                    // - any other reqs?
+                    derives.push("Debug");
                 };
                 let has_bitfields =
                     fields
@@ -1700,7 +1712,7 @@ impl<'c> Translation<'c> {
                     let outer_struct = mk()
                         .span(span)
                         .pub_()
-                        .call_attr("derive", vec!["Copy", "Clone"])
+                        .call_attr("derive", vec!["Copy", "Clone", "Debug"])
                         .meta_item_attr(AttrStyle::Outer, repr_attr)
                         .struct_item(name, vec![outer_field], true);
 
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4277,7 +4277,10 @@ impl<'c> Translation<'c> {
                     Ok(self.enum_cast(ty.ctype, enum_decl_id, expr, val, source_ty, target_ty))
                 } else if target_ty_ctype.is_floating_type() && source_ty_kind.is_bool() {
                     val.and_then(|x| {
-                        Ok(WithStmts::new_val(mk().cast_expr(mk().cast_expr(x, mk().path_ty(vec!["u8"])), target_ty)))
+                        Ok(WithStmts::new_val(mk().cast_expr(
+                            mk().cast_expr(x, mk().path_ty(vec!["u8"])),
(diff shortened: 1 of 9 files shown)


dgherzka:
I actually meant to open this PR against my fork, not upstream. It may be kind of rough, but it would be helpful to know whether I'm at least on the right track!

dgherzka:
Because I opened this against the wrong repo, it also contains https://github.com/immunant/c2rust/pull/1030. Sorry, I'm a bit of a github noob.

dgherzka:
The platform differences in va_list are making it hard to unit test this. On Linux, the struct (S3) has a lifetime parameter, but the return type of the translated fn that returns it doesn't have one, so the build fails (that must be a bug, right?). It's also unsafe to construct the object, so it can't just be a static because nothing will wrap it in `unsafe {}`. I'll see if just declaring pointers helps.

kkysen:
> Because I opened this against the wrong repo, it also contains #1030. Sorry, I'm a bit of a github noob.

Usually I stack PRs, so PR #1 is merging into `main` from `branch1` and PR #2 is merging into `branch1` from `branch2`.  Then when I merge PR #1 and delete `branch1`, it updates PR #2 to be merging into `main` from `branch2`.  That's when the branches belong to the repo, though, and I'm not sure how it works for forks.  When #1030 merges, though, this PR should fix itself, though you may need to rebase.

c2rust-transpile/src/translator/structs.rs
```diff
@@ -296,7 +343,7 @@ impl<'a> Translation<'a> {
         struct_id: CRecordId,
         field_ids: &[CDeclId],
         platform_byte_size: u64,
-    ) -> TranslationResult<(Vec<Field>, bool)> {
+    ) -> TranslationResult<(Vec<Field>, bool, bool)> {
```
kkysen, line 346: We probably want this `(Vec<Field>, bool, bool)` to be a `struct` with named fields at this point.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4275,6 +4284,13 @@ impl<'c> Translation<'c> {
                     let expr =
                         expr.ok_or_else(|| format_err!("Casts to enums require a C ExprId"))?;
                     Ok(self.enum_cast(ty.ctype, enum_decl_id, expr, val, source_ty, target_ty))
+                } else if target_ty_ctype.is_floating_type() && source_ty_kind.is_bool() {
```
ahomescu, line 4287: This seems unrelated to `Debug`. Unrelated changes usually get separate PRs, but we can make an exception this time.

ahomescu: Never mind, I just saw PR #1030 .

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1194,6 +1194,12 @@ struct ConvertedVariable {
     pub init: TranslationResult<WithStmts<Box<Expr>>>,
 }
 
+pub struct ConvertedStructFields {
```
dgherzka, line 1197: This needed to be `pub` because `convert_struct_fields` is, but it seems like both of those should possibly be `pub(crate)`. I didn't want to make that change because I didn't see very many instances of `pub(crate)` in c2rust-transpile. Should I leave that as is or make these crate-private?

dgherzka:
This is now rebased and no longer includes the commits that are already in #1030.

c2rust-transpile/src/translator/mod.rs
kkysen: `pub` should be fine here.  We haven't been careful with the `pub` API for `c2rust-transpile` since it's meant to be used only from the CLI (being able to call it programmatically would be nice, but it's tricky, as the C++ `clang` part of it isn't entirely thread-safe).  So since most of the types here are `pub` but probably shouldn't be, it's fine to make this one `pub`, too.

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -155,6 +155,10 @@ struct Args {
     /// Fail when the control-flow graph generates branching constructs
     #[clap(long)]
     fail_on_multiple: bool,
+
+    /// Derive `Debug` trait for any structs that allow it (i.e., do not recursively contain any unions)
+    #[clap(long)]
+    derive_debug: bool,
```
kkysen, line 161: Can we make the API for this `--derive`, where we give it a list of `trait`s to derive, which at this point are `Clone`, `Copy`, and `Debug`, and `Clone` and `Copy` are done by default?  So the `clap` API would probably be something like this:

```rust
use clap::ValueEnum;

#[derive(Clone, Copy, Debug, ValueEnum)]
enum Derive {
    Clone,
    Copy,
    Debug,
}

struct Args {
    ...
    /// `trait`s to automatically derive where possible
    #[clap(long)]
    derive: Vec<Derive>,
}
```

I'm not getting the `clap` derive API right, but something like that.

dgherzka:
Latest push changes this CLI argument and rebases the branch.

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -164,9 +177,29 @@ enum InvalidCodes {
...
 }
 
+#[derive(Clone, Copy, Debug, ValueEnum)]
+#[clap(rename_all = "PascalCase")]
+enum ExtraDerive {
+    Debug,
+}
+
+impl ExtraDerive {
+    fn to_transpiler_derive(&self) -> Derive {
```
dgherzka, line 187: These are two different types because I didn't want to make c2rust-transpile depend on clap so it could implement ValueEnum. It works out because allowing `--derive Clone` wouldn't make much sense anyway if that's enabled by default.

tests/structs/src/debug_derive.c
```diff
@@ -0,0 +1,29 @@
...
+typedef struct {
+    struct {
+        struct {
+            union {
+                int d;
+                float e;
+            } c;
+        } b;
+    } a;
+} NotDebuggable1;
```
kkysen, line 27: You can test this is not `Debug` using [`static_assertions::assert_not_impl_any!`](https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_not_impl_any.html).  There's also a similar `assert_impl_one!` you can use to do the check you're already doing, that a type is `Debug`.

kkysen: That way you also don't need to generate a value of the types, just the type itself.

kkysen (changes requested):
The way the derives are calculated now (including for the existing ones) is much more elegant (thanks!).  I do have a bunch of suggestions, though, but it looks quite good so far.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1644,10 +1649,27 @@ impl<'c> Translation<'c> {
...
+                            Derive::Debug => can_derive_debug,
+                            Derive::BitfieldStruct => has_bitfields,
+                        };
+                        if can_derive {
+                            Some(derive.to_string())
+                        } else {
+                            None
+                        }
+                    })
+                    .collect_vec();
```
kkysen, line 1671:
```suggestion
                    .collect::<Vec<_>>();
```

Equally simple and doesn't require `itertools`; same with the other `.collect_vec()`.

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -155,6 +157,17 @@ struct Args {
...
+    /// Add extra derived traits to generated structs in addition to the default
+    /// set of derives (Copy, Clone, BitfieldStruct). Specify multiple times to
+    /// add more than one derive. A struct will derive all traits in the set for
+    /// which it is eligible.
+    ///
+    /// For example, a struct containing a union cannot derive Debug, so
+    /// `#[derive(Debug)]` will not be added to that struct regardless of
+    /// whether `--derive Debug` is specified.
+    #[clap(long = "derive", value_enum, value_name = "TRAIT")]
+    extra_derives: Vec<ExtraDerive>,
```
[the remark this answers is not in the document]
dgherzka: It still is

dgherzka:
Marking as a draft while I make the fixes. Thanks for all the feedback!

PR #1061, opened 2024-01-19
Ubuntu 22.04 installation requirements added

Yeaseen:
For `c2rust` to be installed on `Ubuntu 22.04`, the library `libstdc++-12-dev` is a prerequisite. I added this to the Readme file. Also, I discarded any `.DS_store` file to be uploaded on 'GitHub' in the `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_store
 *.tar.xz
 *.tar.xz.sig
 *.tar.gz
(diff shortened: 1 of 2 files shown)


fw-immunant:
Hi, what error do you observe when installing with the current (sans `libstdc++12`) set of dependencies? `libstdc++12` is not a direct dependency of c2rust, so it doesn't make sense to install it explicitly. I believe this is an upstream packaging error, which seems to be tracked here: https://bugs.launchpad.net/ubuntu/+source/llvm-toolchain-14/+bug/2038340

PR #1069, opened 2024-03-18
Demonstrate use of PDG from dynamic analysis in static analysis

fw-immunant:
Turns out we need to pass the profile flags twice: once for the outer invocation of cargo itself and once for the inner invocation of our cargo wrapper. This, along with the `--feature` flags for the inner crate and the addition of an optional dependency, make things work with the same crate ID.

### DIFF
--- a/scripts/pdg.sh
+++ b/scripts/pdg.sh
@@ -11,18 +11,22 @@ SCRIPT_DIR="${CWD}/$(dirname "${SCRIPT_PATH}")"
 # Environment Variables:
 # * `PROFILE` (default `release`):
 #       a `cargo` profile as in `target/$PROFILE`
+# * `NO_USE_PDG` (default empty):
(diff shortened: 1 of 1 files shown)


spernsteiner:
Seems unfortunate that the user's project needs this extra `c2rust-analysis-rt` dependency that they'll never use in normal builds.  How does `c2rust-instrument` go about injecting this dependency?  Could `c2rust-analyze` do the same?

kkysen:
> Seems unfortunate that the user's project needs this extra `c2rust-analysis-rt` dependency that they'll never use in normal builds. How does `c2rust-instrument` go about injecting this dependency? Could `c2rust-analyze` do the same?

`c2rust instrument` has an option to add it as an optional dependency automatically.

fw-immunant:
Dropped the dependency-adding commit as we actually don't need it as @kkysen indicated.

PR #1070, opened 2024-03-25
analyze: add C2RUST_ANALYZE_NO_CARGO env var to disable cargo integration

spernsteiner:
Adding cargo integration made it a bit tricky to access the pre-cargo behavior, in which `c2rust-analyze` would behave like `rustc`.  This behavior was useful for experimenting on small single-file example programs (e.g. `c2rust-analyze foo.rs`).  This commit adds a new env var, `C2RUST_ANALYZE_NO_CARGO`, which can be set to restore the old behavior.

### DIFF
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -202,8 +202,6 @@ impl Cargo {
 
 const RUSTC_WRAPPER_VAR: &str = "RUSTC_WRAPPER";
 const RUST_SYSROOT_VAR: &str = "RUST_SYSROOT";
-/// If this env var is set, we skip all cargo integration and just behave like rustc.
-const NO_CARGO_VAR: &str = "C2RUST_ANALYZE_NO_CARGO";
 
 /// Read a [`PathBuf`] from the [`mod@env`]ironment that should've been set by the [`cargo_wrapper`].
 fn env_path_from_wrapper(var: &str) -> anyhow::Result<PathBuf> {
@@ -274,9 +272,16 @@ fn is_build_script(at_args: &[String]) -> anyhow::Result<bool> {
     Ok(bin_crate_name().is_none() && is_bin_crate(at_args)?)
 }
 
+/// Check whether "no Cargo" mode is enabled.  In this mode, `c2rust-analyze` behaves like `rustc`
+/// instead of like `cargo`.  For example, `C2RUST_ANALYZE_NO_CARGO=1 c2rust-analyze foo.rs` will
+/// process `foo.rs` much like `rustc foo.rs` does.
+fn in_no_cargo_mode() -> bool {
+    env::var_os("C2RUST_ANALYZE_NO_CARGO").is_some()
+}
+
 /// Run as a `rustc` wrapper (a la `$RUSTC_WRAPPER`/[`RUSTC_WRAPPER_VAR`]).
 fn rustc_wrapper() -> anyhow::Result<()> {
-    let no_cargo = env::var_os(NO_CARGO_VAR).is_some();
+    let no_cargo = in_no_cargo_mode();
     let mut at_args = if no_cargo {
         env::args().collect::<Vec<_>>()
     } else {
@@ -427,7 +432,7 @@ fn main() -> anyhow::Result<()> {
     let own_exe = env::current_exe()?;
 
     let wrapping_rustc = env::var_os(RUSTC_WRAPPER_VAR).as_deref() == Some(own_exe.as_os_str())
-        || env::var_os(NO_CARGO_VAR).is_some();
+        || in_no_cargo_mode();
     if wrapping_rustc {
         rustc_wrapper()
     } else {
--- a/c2rust-analyze/src/main.rs
+++ b/c2rust-analyze/src/main.rs
@@ -202,6 +202,8 @@ impl Cargo {
 
 const RUSTC_WRAPPER_VAR: &str = "RUSTC_WRAPPER";
 const RUST_SYSROOT_VAR: &str = "RUST_SYSROOT";
+/// If this env var is set, we skip all cargo integration and just behave like rustc.
+const NO_CARGO_VAR: &str = "C2RUST_ANALYZE_NO_CARGO";
 
 /// Read a [`PathBuf`] from the [`mod@env`]ironment that should've been set by the [`cargo_wrapper`].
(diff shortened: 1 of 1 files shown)


c2rust-analyze/src/main.rs
```diff
@@ -274,10 +276,15 @@ fn is_build_script(at_args: &[String]) -> anyhow::Result<bool> {
 
 /// Run as a `rustc` wrapper (a la `$RUSTC_WRAPPER`/[`RUSTC_WRAPPER_VAR`]).
 fn rustc_wrapper() -> anyhow::Result<()> {
-    let mut at_args = env::args().skip(1).collect::<Vec<_>>();
+    let no_cargo = env::var_os(NO_CARGO_VAR).is_some();
```
ahomescu, line 279: Since you're using this from two places, maybe put it in a helper function?

c2rust-analyze/src/main.rs
```diff
@@ -274,10 +276,15 @@ fn is_build_script(at_args: &[String]) -> anyhow::Result<bool> {
...
+    let mut at_args = if no_cargo {
+        env::args().collect::<Vec<_>>()
+    } else {
+        env::args().skip(1).collect::<Vec<_>>()
```
ahomescu, line 283: Could you add a sanity check that the first argument is indeed `cargo`? Is it supposed to be the path to the binary?

ahomescu (approved):

c2rust-analyze/src/main.rs
kkysen: The first arg will be the current executable, `c2rust-analyze`.

c2rust-analyze/src/main.rs
```diff
@@ -274,10 +276,15 @@ fn is_build_script(at_args: &[String]) -> anyhow::Result<bool> {
...
+    };
```
kkysen, line 284: With the cargo wrapper, the first arg is `c2rust-analyze` and the second arg is `rustc`.  If we're running as a `rustc` wrapper, `rustc` is not going to the first arg, so we need to set it.

kkysen (commented):
Alternatively, what about achieving this similar to how `cargo rustc` works?  That is, `c2rust-analyze rustc ...` runs as `rustc` directly, and we check this by seeing if argv1 is `rustc`.  Then when we invoke `rustc_wrapper`, the logic for `at_args` doesn't need to change, since argv1 is `rustc`, just like when it was invoked by the cargo wrapper.  And for `is_primary_compilation`, we just do the `env::var_os(RUSTC_WRAPPER_VAR).as_deref() == Some(own_exe.as_os_str())` check again (if it's not equal, `is_primary_compilation` should be true).

fw-immunant (approved):
I kind of wish this made `rustc_wrapper` accept `at_args` and `is_primary_package` (or a clearer name for controlling which callbacks set is passed) as arguments instead of adding branches to its body.

Correctness looks fine, but I wish this were documented externally. I don't know if I fully understand how this changes the interface of the program--is it just making it easier to access the behavior that would be provided by `CARGO_PRIMARY_PACKAGE= RUSTC_WRAPPER=/path/to/c2rust-analyze /path/to/c2rust-analyze foo.rs`?

spernsteiner:
>what about achieving this similar to how `cargo rustc` works?

`cargo rustc` still requires a `Cargo.toml`, and chooses which file to compile based on the Cargo targets.  This mode is meant to work like plain `rustc`, which can compile small examples via `rustc foo.rs` without needing any supporting Cargo infrastructure.  This is useful for running our filecheck test cases and similar kinds of examples.

c2rust-analyze/src/main.rs
spernsteiner: I added a helper function and put a little description of the behavior in its doc comment.

spernsteiner:
> That is, `c2rust-analyze rustc ...` runs as `rustc` directly, and we check this by seeing if argv1 is `rustc`.

Ah, I misunderstood what you meant.

I think it would be confusing for `c2rust-analyze <subcmd>` to work like `cargo <subcmd>` except in the specific case where `subcmd == "rustc"`.  If we were to define our own set of subcommands independent of cargo's, then it might make sense to define `c2rust-analyze rustc` this way, but that's a more significant refactor than I wanted to do for this branch (the idea here is just to bring back the old `rustc`-like behavior in the simplest way possible).

c2rust-analyze/src/main.rs
```diff
@@ -272,12 +272,24 @@ fn is_build_script(at_args: &[String]) -> anyhow::Result<bool> {
     Ok(bin_crate_name().is_none() && is_bin_crate(at_args)?)
 }
 
+/// Check whether "no Cargo" mode is enabled.  In this mode, `c2rust-analyze` behaves like `rustc`
```
ahomescu, line 275: Would it be useful to also have a command line flag, e.g., `c2rust-analyze --rustc-mode`?

c2rust-analyze/src/main.rs
```diff
@@ -272,12 +272,24 @@ fn is_build_script(at_args: &[String]) -> anyhow::Result<bool> {
...
+/// instead of like `cargo`.  For example, `C2RUST_ANALYZE_NO_CARGO=1 c2rust-analyze foo.rs` will
+/// process `foo.rs` much like `rustc foo.rs` does.
+fn in_no_cargo_mode() -> bool {
+    env::var_os("C2RUST_ANALYZE_NO_CARGO").is_some()
```
ahomescu, line 279: Does this also return `true` if the variable is empty? I.e. someone runs the tool as `C2RUST_ANALYZE_NO_CARGO= c2rust-analyze`.

spernsteiner: Yes - this is consistent with our handling of other env vars like `C2RUST_ANALYZE_TEST_DONT_CATCH_PANIC`

c2rust-analyze/src/main.rs
spernsteiner: Currently we don't parse the command line arguments in rustc / rustc-wrapper mode - we just pass through whatever args we get to rustc's own argument handling.  All `c2rust-analyze`-specific configuration in these modes is done through env vars.  If we decide in the future to change this, then at that point I think it would make sense to have `--rustc-mode` (or a subcommand, maybe) as well.

PR #1071, opened 2024-03-25
analyze: emit inline annotations for debugging

spernsteiner:
This branch adds a new `annotate` module that allows providing annotations to be attached to specific lines during rewriting, and uses it to show pointer permissions/flags inline in the rewritten code.  For example:

```Rust
#[no_mangle]
pub unsafe extern "C" fn insertion_sort<'h0>(n: libc::c_int, p: &'h0 mut [(libc::c_int)]) {
// 10: p: typeof(_2) = *mut {g0} i32
// 10: p:   g0 = READ | WRITE | UNIQUE | OFFSET_ADD | OFFSET_SUB, (empty)
    let mut i: libc::c_int = 1 as libc::c_int;
    while i < n {
        let tmp: libc::c_int = *&(&(&*(p))[((i as isize) as usize) ..])[0];
        // 15: p.offset(i as isize): typeof(_9) = *mut {l9} i32
        // 15: p.offset(i as isize):   l9 = READ | UNIQUE, (empty)
        // 15: p: typeof(_10) = *mut {l11} i32
        // 15: p:   l11 = READ | UNIQUE | OFFSET_ADD | OFFSET_SUB, (empty)
        let mut j: libc::c_int = i;
        while j > 0 as libc::c_int && *&(&(&*(p))[(((j - 1 as libc::c_int) as isize) as usize) ..])[0] > tmp {
        // 19: p.offset((j - 1 ... size): typeof(_21) = *mut {l23} i32
        // 19: p.offset((j - 1 ... size):   l23 = READ | UNIQUE, (empty)
        // 19: p: typeof(_22) = *mut {l25} i32
        // 19: p:   l25 = READ | UNIQUE | OFFSET_ADD | OFFSET_SUB, (empty)
            *&mut (&mut (p)[((j as isize) as usize) ..])[0] = *&(&(&*(p))[(((j - 1 as libc::c_int) as isize) as usize) ..])[0];
            // 24: p.offset((j - 1 ... size): typeof(_30) = *mut {l34} i32
            // 24: p.offset((j - 1 ... size):   l34 = READ | UNIQUE, (empty)
            // 24: p: typeof(_31) = *mut {l36} i32
            // 24: p:   l36 = READ | UNIQUE | OFFSET_ADD | OFFSET_SUB, (empty)
            // 24: p.offset(j as isize): typeof(_37) = *mut {l43} i32
            // 24: p.offset(j as isize):   l43 = READ | WRITE | UNIQUE, (empty)
            // 24: p: typeof(_38) = *mut {l45} i32
            // 24: p:   l45 = READ | WRITE | UNIQUE | OFFSET_ADD | OFFSET_SUB, (empty)
            j -= 1
        }
        *&mut (&mut (p)[((j as isize) as usize) ..])[0] = tmp;
        // 29: p.offset(j as isize): typeof(_46) = *mut {l54} i32
        // 29: p.offset(j as isize):   l54 = READ | WRITE | UNIQUE, (empty)
        // 29: p: typeof(_47) = *mut {l56} i32
        // 29: p:   l56 = READ | WRITE | UNIQUE | OFFSET_ADD | OFFSET_SUB, (empty)
        i += 1
    }
}
```

This is helpful for debugging, since it means we no longer need to cross-reference line numbers in the debug output with the source code.  It's also easy to add more annotations in the future if it would be useful.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -50,7 +50,7 @@ use rustc_middle::ty::Ty;
 use rustc_middle::ty::TyCtxt;
 use rustc_middle::ty::TyKind;
 use rustc_middle::ty::WithOptConstParam;
-use rustc_span::{Span, Symbol, DUMMY_SP};
+use rustc_span::{Span, Symbol};
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::env;
@@ -1544,7 +1544,13 @@ fn run(tcx: TyCtxt) {
         }
 
         // Emit annotations for fields
-        let span = tcx.def_ident_span(did).unwrap_or(DUMMY_SP);
+        let span = match tcx.def_ident_span(did) {
+            Some(x) => x,
+            None => {
+                warn!("field {:?} has no def_ident_span to annotate", did);
+                continue;
+            }
+        };
         let mut ptrs = Vec::new();
         let ty_str = context::print_ty_with_pointer_labels(field_lty, |ptr| {
             if ptr.is_none() {
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -1581,9 +1581,16 @@ fn run(tcx: TyCtxt) {
 
     // Apply rewrite to all functions at once.
     let mut update_files = rewrite::UpdateFiles::No;
-    if let Ok(val) = env::var("C2RUST_ANALYZE_REWRITE_IN_PLACE") {
-        if val == "1" {
-            update_files = rewrite::UpdateFiles::Yes;
+    if let Ok(val) = env::var("C2RUST_ANALYZE_REWRITE_MODE") {
+        match val.as_str() {
+            "none" => {}
+            "inplace" => {
+                update_files = rewrite::UpdateFiles::InPlace;
+            }
+            "alongside" => {
+                update_files = rewrite::UpdateFiles::Alongside;
+            }
+            _ => panic!("bad value {:?} for C2RUST_ANALYZE_REWRITE_MODE", val),
         }
     }
     rewrite::apply_rewrites(tcx, all_rewrites, annotations, update_files);
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -664,19 +664,19 @@ fn run(tcx: TyCtxt) {
             pointee_type::generate_constraints(&acx, &mir)
         }));
 
-        let pointee_constraints = match r {
-            Ok(x) => x,
+        let mut info = FuncInfo::default();
+        let local_pointee_types = LocalPointerTable::new(acx.num_pointers());
+        info.acx_data.set(acx.into_data());
+
+        match r {
+            Ok(pointee_constraints) => {
+                info.pointee_constraints.set(pointee_constraints);
+            }
             Err(pd) => {
                 gacx.mark_fn_failed(ldid.to_def_id(), pd);
-                continue;
             }
-        };
+        }
 
-        let local_pointee_types = LocalPointerTable::new(acx.num_pointers());
-
-        let mut info = FuncInfo::default();
-        info.acx_data.set(acx.into_data());
-        info.pointee_constraints.set(pointee_constraints);
         info.local_pointee_types.set(local_pointee_types);
         info.recent_writes.set(RecentWrites::new(&mir));
         func_info.insert(ldid, info);
@@ -1101,6 +1101,9 @@ fn run(tcx: TyCtxt) {
                     field_ltys,
                 );
             }));
+
+            info.acx_data.set(acx.into_data());
+
             match r {
                 Ok(()) => {}
                 Err(pd) => {
@@ -1108,8 +1111,6 @@ fn run(tcx: TyCtxt) {
                     continue;
                 }
             }
-
-            info.acx_data.set(acx.into_data());
         }
 
         let mut num_changed = 0;
@@ -1156,6 +1157,9 @@ fn run(tcx: TyCtxt) {
 
             acx.check_string_literal_perms(&asn);
         }));
+
+        info.acx_data.set(acx.into_data());
+
         match r {
             Ok(()) => {}
             Err(pd) => {
@@ -1163,8 +1167,6 @@ fn run(tcx: TyCtxt) {
                 continue;
             }
         }
-
-        info.acx_data.set(acx.into_data());
     }
 
     // Check that these perms haven't changed.
@@ -1310,6 +1312,9 @@ fn run(tcx: TyCtxt) {
                 all_rewrites.extend(expr_rewrites);
                 all_rewrites.extend(ty_rewrites);
             }));
+
+            info.acx_data.set(acx.into_data());
+
             match r {
                 Ok(()) => {}
                 Err(pd) => {
@@ -1317,8 +1322,6 @@ fn run(tcx: TyCtxt) {
                     continue;
                 }
             }
-
-            info.acx_data.set(acx.into_data());
         }
 
         // This call never panics, which is important because this is the fallback if the more
@@ -1456,6 +1459,8 @@ fn run(tcx: TyCtxt) {
         if let Some(report) = func_reports.remove(&ldid) {
             eprintln!("{}", report);
         }
+
+        info.acx_data.set(acx.into_data());
     }
 
     // Print results for `static` items.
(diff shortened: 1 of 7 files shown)


c2rust-analyze/src/analyze.rs
```diff
@@ -1456,6 +1458,54 @@ fn run(tcx: TyCtxt) {
...
+            let span = local_span(decl);
+            let mut ptrs = Vec::new();
+            let ty_str = context::print_ty_with_pointer_labels(acx.local_tys[local], |ptr| {
+                if ptr.is_none() {
+                    return String::new();
+                }
+                ptrs.push(ptr);
+                format!("{{{}}}", ptr)
+            });
+            if ptrs.len() == 0 {
```
ahomescu, line 1494: Small nit here:
```suggestion
            if ptrs.is_empty() {
```

c2rust-analyze/src/annotate.rs
```diff
@@ -0,0 +1,62 @@
...
+        let src = sm
+            .span_to_snippet(span)
+            .unwrap_or_else(|_| "<error>".into());
+        let src = src.split_ascii_whitespace().collect::<Vec<_>>().join(" ");
+        let (src1, src2, src3) = if src.len() > 20 {
+            (&src[..15], " ... ", &src[src.len() - 5..])
+        } else {
+            (&src[..], "", "")
+        };
+        self.m.entry(file_idx).or_insert_with(Vec::new).push((
```
ahomescu, line 42: `or_default()` is shorter than `or_insert_with(Vec::new)`.

c2rust-analyze/src/context.rs
```diff
@@ -1399,3 +1399,127 @@ impl Assignment<'_> {
...
+) -> String {
+    let mut out = String::new();
+    print_ty_with_pointer_labels_into(&mut out, lty, &mut f);
+    out
+}
+
+pub fn print_ty_with_pointer_labels_into<L: Copy>(
+    dest: &mut String,
+    lty: LabeledTy<L>,
+    f: &mut impl FnMut(L) -> String,
```
ahomescu, line 1419: Does this need to be passed by reference here, when it's by value in the other function? Is it because of the recursion?

c2rust-analyze/src/context.rs
```diff
@@ -1399,3 +1399,127 @@ impl Assignment<'_> {
...
+                dest.push_str("&mut ");
+            }
+            if s.len() > 0 {
+                dest.push_str(&s);
+                dest.push_str(" ");
+            }
+            print_ty_with_pointer_labels_into(dest, lty.args[0], f);
+        }
+        FnDef(def_id, _substs) => {
+            write!(dest, "{:?}", def_id).unwrap();
```
ahomescu, line 1483: I don't remember what's in the `DefId`, does this print a function name?

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -580,12 +593,51 @@ impl<'a, F: FnMut(&str)> Sink for RewriteTreeSink<'a, F> {
...
+    v: Vec<Option<usize>>,
+}
+
+impl LineMapBuilder {
+    /// Record that part of input line `i` is found in output line `j`.
+    pub fn record(&mut self, i: usize, j: usize) {
+        if i >= self.v.len() {
+            self.v.resize(i + 1, None);
+        }
+        self.v[i] = match self.v[i] {
```
ahomescu, line 610: `self.v[i] = Some(cmp::min(j, self.v[i].unwrap_or(j)));` or something like that would be shorter.

c2rust-analyze/src/rewrite/mod.rs
```diff
@@ -243,17 +247,78 @@ impl apply::Sink for FormatterSink<'_, '_> {
...
 
+/// Return a copy of `src` with `annotations` added as comments.  If `line_map` is provided, the
+/// line numbers in `annotations` are remapped using `line_map` before the annotations are inserted
+/// in `src`.
+fn add_annotations(
+    src: String,
+    line_map: Option<&[usize]>,
+    mut annotations: Vec<(usize, String)>,
+) -> String {
+    if annotations.len() == 0 {
```
ahomescu, line 258: `is_empty`?

c2rust-analyze/src/context.rs
fw-immunant: Handily we've written a parser for this format in `parse_def_id` - the name is included if available.

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -596,12 +593,51 @@ impl<'a, F: FnMut(&str)> Sink for RewriteTreeSink<'a, F> {
...
 /// Apply rewrites `rws` to the source files covered by their `Span`s.  Returns a map giving the
-/// rewritten source code for each file that contains at least one rewritten `Span`.
+/// rewritten source code for each file that contains at least one rewritten `Span`.  The
+/// `Vec<usize>` is a line map: for each line of the original source code, it gives the line number
+/// in the output `String` of the first output line that contains some part of the input line.
 pub fn apply_rewrites(
     source_map: &SourceMap,
     rws: Vec<(Span, Rewrite)>,
-) -> HashMap<FileName, String> {
+) -> HashMap<FileName, (String, Vec<usize>)> {
```
fw-immunant, line 640: A newtype could maybe clarify this more concisely than docs.

c2rust-analyze/src/context.rs
spernsteiner: Yes, this is the easiest way to avoid issues with the branching recursion.

c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -231,9 +231,6 @@ impl<S: Sink> Emitter<'_, S> {
     fn emit_str(&mut self, s: &str) -> Result<(), S::Error> {
         self.sink.emit_str(s)
     }
-    fn emit_orig_str(&mut self, s: &str, line: usize) -> Result<(), S::Error> {
-        self.sink.emit_orig_str(s, line)
-    }
```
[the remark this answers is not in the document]
spernsteiner: I added it at some point during development, but it wound up being unused in the end.

c2rust-analyze/src/context.rs
spernsteiner: > I don't remember what's in the `DefId`, does this print a function name?

It will print something like `DefId(0:4 ~ insertion_sort[add9]::insertion_sort)`, which is not the prettiest, but should be okay for debug output.

c2rust-analyze/src/annotate.rs
spernsteiner: I prefer the `or_insert_with(Vec::new)` version as a hint to the reader that the map values are `Vec`s.

c2rust-analyze/src/analyze.rs
```diff
@@ -664,19 +660,19 @@ fn run(tcx: TyCtxt) {
...
+        info.acx_data.set(acx.into_data());
+
+        match r {
+            Ok(pointee_constraints) => {
+                info.pointee_constraints.set(pointee_constraints);
+            }
             Err(pd) => {
                 gacx.mark_fn_failed(ldid.to_def_id(), pd);
-                continue;
             }
```
ahomescu, line 673: Was this previously jumping back to line 623? Why isn't the `continue` needed anymore?

c2rust-analyze/src/analyze.rs
```diff
@@ -1156,15 +1153,16 @@ fn run(tcx: TyCtxt) {
...
             acx.check_string_literal_perms(&asn);
         }));
+
+        info.acx_data.set(acx.into_data());
+
         match r {
             Ok(()) => {}
             Err(pd) => {
                 gacx.mark_fn_failed(ldid.to_def_id(), pd);
                 continue;
```
ahomescu, line 1163: Is this one still needed?

c2rust-analyze/src/analyze.rs
```diff
@@ -1492,6 +1542,27 @@ fn run(tcx: TyCtxt) {
             let ty_flags = gasn.flags[pid];
             eprintln!("{name:}: ({pid}) perms = {ty_perms:?}, flags = {ty_flags:?}");
         }
+
+        // Emit annotations for fields
+        let span = tcx.def_ident_span(did).unwrap_or(DUMMY_SP);
```
ahomescu, line 1547: Can the dummy span here be a problem when picking which line you're printing?

c2rust-analyze/src/analyze.rs
spernsteiner: Looks like this change is part of 6d7d3e3dfc253577c25a240888d842d04edc171c "preserve function acx_data on error".  This loop is the one that initializes the `FuncInfo` (which contains `acx_data`) for each function.  Removing the `continue` means that if an error occurs in this part of the analysis, only `info.pointee_constraints` will be affected, and all other `info` fields will still get initialized.

ahomescu (approved):

PR #1073, opened 2024-03-26
analyze: add --rewrite-mode pointwise

spernsteiner:
This adds a new --rewrite-mode called "pointwise", which rewrites each function in isolation, with all other definitions marked `FIXED`.  The static analysis runs only once, then we run multiple rewriting passes using the same analysis results, making this much more efficient than running the whole `c2rust-analyze` tool multiple times.

The rewritten code is output to a separate file for each function.  For example, given `foo.rs` containing a function `bar`, this mode will write to `foo.bar.rs` the new code produced by rewriting only `bar`.

Pointwise mode is the basis for our new "pointwise success rate" metric.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -1199,27 +1199,46 @@ fn run(tcx: TyCtxt) {
         }
     }
 
-    run2(
-        tcx,
-        gacx,
-        gasn,
-        global_pointee_types,
-        func_info,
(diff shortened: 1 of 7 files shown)


fw-immunant (approved):
LGTM, the added loop around `run2` seems like it could benefit from a refactor to multiply less work (or at least output less) but I'm not sure exactly which portions could be reused across all definitions as the analysis does have to vary.

PR #1074, opened 2024-03-29
analyze: add scripts for computing pointwise metrics

spernsteiner:
Adds scripts for computing "pointwise success rate" metrics.  For each function, we run the static analysis and rewrite that function in isolation, producing a new `.rs` file where that function has been rewritten but all other code remains the same. Then we remove the `unsafe` qualifier from the target function and try to compile the code.  The "pointwise success rate" is the number of functions on which this procedure succeeds.

The main entry point is `c2rust-analyze/scripts/run_pointwise_metrics_lighttpd.sh` (as the name suggests, this is designed to compute the success rate on lighttpd specifically).  It uses a few helpers: `pointwise_try_build.sh` tries to remove `unsafe` and compile the rewritten code for a specific function, `pointwise_try_build_unmodified.sh` does the same but on the unmodified, non-rewritten code (used for computing a baseline success rate), and `pointwise_metrics.py` tallies up the results and prints overall counts.

Current output on lighttpd:
```
pointwise:     98/1008 functions passed
unmodified:   149/1008 functions passed
improved 20 functions
broke 71 functions
```

This PR depends on #1073, which implements the `pointwise` rewrite mode in `c2rust-analyze`.


c2rust-analyze/scripts/pointwise_try_build_unmodified.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "func=$func"
+
+filter_errors() {
+    jq 'select(.level == "error") | .message' -r |
+        { grep -v -e '^aborting due to ' -e '^call to unsafe function is unsafe ' || true; }
+}
+
+d="$(dirname "$f")"
+f="$d/${name}_safe_${func}.rs"
+cp "$d/$name.rs" "$f"
```
fw-immunant, line 26: These are the only difference between these two scripts--it would be good to just make these conditional and avoid the duplication.

fw-immunant (changes requested):
Deduplicate the nearly-identical scripts somehow, otherwise looks good.

c2rust-analyze/scripts/pointwise_try_build_unmodified.sh
spernsteiner: I've updated this to deduplicate the similar scripts.

fw-immunant (approved):

PR #1075, opened 2024-04-01
ast-exporter: use major, not full, version for clang resource dir

fw-immunant:
Without this patch, running `c2rust-transpile` built against Clang 17 produces errors caused by an inability to locate `stddef.h`.


fw-immunant:
It looks like this is an upstream change as of Clang 17: https://reviews.llvm.org/D125860

kkysen (approved):

PR #1081, opened 2024-04-12
analyze: initial implementation of NON_NULL static analysis

spernsteiner:
This adds a very basic static analysis for `NON_NULL` within the current `dataflow` framework.  It starts by optimistically assuming that all pointers are `NON_NULL`, and removes the permission from pointers into which a `ptr::null()` or equivalent might flow.  This branch just implements the static analysis, not rewriting.

I'm actually not a huge fan of this design - I think we'd probably get much better results with a path-sensitive analysis that can handle common patterns from C like `if !p.is_null() { let q = (&p).field; /* use q... */ }` by detecting null checks in the CFG.  But the simple path-insensitive version is sufficient for now, and gives us somewhere to plug in the dynamic analysis results.

### DIFF
--- a/c2rust-analyze/src/borrowck/type_check.rs
+++ b/c2rust-analyze/src/borrowck/type_check.rs
@@ -603,7 +603,9 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                         });
                     }
                     Callee::Null { .. } => {
-                        // TODO: do we need to do anything here?
+                        // Just visit the place.  The null pointer returned here has no origin, so
+                        // there's no need to call `do_assign` to set up subset relations.
+                        let _pl_lty = self.visit_place(destination);
                     }
                 }
             }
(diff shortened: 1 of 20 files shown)


c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -602,6 +602,9 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                             self.visit_operand(p)
                         });
                     }
+                    Callee::Null { .. } => {
+                        // TODO: do we need to do anything here?
```
fw-immunant, line 606: I don't think we do. The only thing that comes to mind is calling `visit_place(destination)` as we do for other pointer-valued function calls, but if I'm understanding its definition properly, that call is actually pure and we don't need its result here.

c2rust-analyze/src/util.rs
```diff
@@ -342,6 +345,30 @@ fn builtin_callee<'tcx>(tcx: TyCtxt<'tcx>, did: DefId, substs: SubstsRef<'tcx>)
             Some(Callee::IsNull)
         }
 
+        "null" | "null_mut" => {
+            // The `core::mem::size_of` function.
```
fw-immunant, line 349: Correct this comment.

fw-immunant (approved):
One comment fix needed as mentioned inline, otherwise LGTM.

PR #1086, opened 2024-04-25
Import null pointer information from PDG into static analysis

ahomescu:


ahomescu:
I still need to test this on some actual code (maybe lighttpd?)

c2rust-analyze/src/analyze.rs
```diff
@@ -993,7 +993,10 @@ fn run(tcx: TyCtxt) {
                 if !node_info.unique {
                     perms.remove(PermissionSet::UNIQUE);
                 }
-                // TODO: PermissionSet::NON_NULL
+                if let Some(true) = g.is_null {
+                    // TODO: is this enough?
+                    perms.remove(PermissionSet::NON_NULL);
```
spernsteiner, line 998: Looks right to me.  It's in line with the handling of `UNIQUE`, which is the other forward-flowing permission aside from `NON_NULL`.

c2rust-analyze/src/analyze.rs
```diff
@@ -993,7 +993,10 @@ fn run(tcx: TyCtxt) {
                 if !node_info.unique {
                     perms.remove(PermissionSet::UNIQUE);
                 }
-                // TODO: PermissionSet::NON_NULL
+                if let Some(true) = g.is_null {
```
spernsteiner, line 996: `is_null` should always be `Some` here, right?  Maybe this should use `unwrap()`.

pdg/src/builder.rs
```diff
@@ -231,6 +231,13 @@ pub fn add_node(
         .unwrap_or_else(|| graphs.graphs.push(Graph::new()));
     let node_id = graphs.graphs[graph_id].nodes.push(node);
 
+    // Mark the graph for null pointers
+    let ptr_is_null = ptr.map_or(false, |ptr| ptr == 0);
+    let graph_is_null = &mut graphs.graphs[graph_id].is_null;
+    // Assert that we're not mixing null and non-null pointers
+    assert!(graph_is_null.is_none() || *graph_is_null == Some(ptr_is_null));
+    *graph_is_null = Some(ptr_is_null);
```
spernsteiner, line 239: I wonder if the `Option` for `is_null` can be removed entirely.  It seems like it only exists so `Graph::new()` has a default value that's non-conflicting with both `true` and `false`.  But if this is the only place where new graphs are created, then we could pass `ptr_is_null` to `Graph::new()` and initialize the `is_null` field directly to the correct value.

spernsteiner (approved):

c2rust-analyze/src/analyze.rs
ahomescu: I was wondering if we want the dynamic analysis to also do the opposite: add `NON_NULL` to pointers where the static algorithm removed it (i.e. we never saw a null pointer in the event log). That might be more useful, but also less sound and harder to implement.

spernsteiner: Right - I think that's something to add in a separate PR, since it requires some additional changes.

c2rust-analyze/src/analyze.rs
```diff
@@ -993,7 +993,10 @@ fn run(tcx: TyCtxt) {
                 if !node_info.unique {
                     perms.remove(PermissionSet::UNIQUE);
                 }
-                // TODO: PermissionSet::NON_NULL
+                if g.is_null {
+                    // TODO: is this enough?
+                    perms.remove(PermissionSet::NON_NULL);
+                }
```
ahomescu, line 1000: @spernsteiner If I'm understanding #1088 correctly, the other thing we need to do here is add `NON_NULL` to `updates_forbidden`?

spernsteiner: Roughly, yes.  Though we might want to experiment a bit - for example, it might be easier to understand the results if we set `updates_forbidden` only on function signatures, or only on named variables and not on temporaries.

spernsteiner: I think there can be two nodes with the same `dest` (either in different `Graph`s or in the same one), representing two different observations of that `dest` during the execution.  We want to set `NON_NULL` in `perms` (and in `updates_forbidden`) only if all observations with the same `dest` show it containing a non-null pointer.  So we need some additional logic beyond just doing `perms.insert(NON_NULL); updates_forbidden.insert(NON_NULL);` here.

spernsteiner:
Is this waiting on anything, other than needing a `cargo fmt` to fix the CI?

dynamic_instrumentation/src/instrument.rs
```diff
@@ -340,35 +340,39 @@ impl<'tcx> MutVisitor<'tcx> for RewriteAddressTakenLocals<'tcx> {
...
-                        .arg_var(place.local)
-                        .arg_index_of(field)
-                        .source(place)
-                        .dest_from(proj_dest)
-                        .add_to(self);
-                }
-            }
-        }
-    }
+// TODO(ahomescu): this is completely broken and needs a rewrite
```
fw-immunant, line 343: Hmmm, did I break this all the way back in 2c6b0ff3c0f95c0c4eecd3b02e25bba469b588c9?

fw-immunant (approved):
LGTM overall. The brokenness in `visit_place` is surprising but I agree that the current behavior is wrong as described; see inline comment.

dynamic_instrumentation/src/instrument.rs
ahomescu: I don't think so, this problem seems to have always been here in some form or another.

spernsteiner:
@ahomescu Don't forget to fix the CI. Currently it's failing on the `cargo fmt` style check.

analysis/runtime/src/events.rs
```diff
@@ -30,10 +30,8 @@ pub enum EventKind {
 
     CopyRef,
 
-    /// Field projection. Used for operations like `_2 = &(*_1).0`. Nested field
-    /// accesses like `_4 = &(*_1).x.y.z` are broken into multiple `Node`s, each
-    /// covering one level.
-    Field(Pointer, u32),
+    /// Projection. Used for operations like `_2 = &(*_1).0`.
+    Project(Pointer, Pointer),
```
spernsteiner, line 34: Can you update the comment to say more about the meaning of the two `Pointer` fields? I guess the first one is the base pointer and the second is the result of the projection?

Also, since the text "Nested field accesses are broken into multiple `Node`s" was removed, does that mean a single `Project` might project through multiple levels of structs?  If so, is there any ambiguity in cases like this:

```Rust
struct A {
  x: i32,
  y: i32,
}
struct B {
  z: i32,
  a: A,
}
let p: *const B = ...;
// Suppose p == 0x1000
let q1: *const A = &(*p).a;
// q1 == 0x1004
let q2: *const i32 = &(*p).a.x;
// q2 == 0x1004
```
Note that this distinction matters to the borrow checker in some cases:
```Rust
let mut b: B = ...;
let r1 = &mut b.a.y;
let r2 = &mut b.a.x; // Legal
let r3 = &mut b.a; // Same address as r2, but not legal
*r1 = 1;
```

analysis/runtime/src/events.rs
```diff
@@ -62,7 +60,10 @@ pub enum EventKind {
     StoreAddrTaken(Pointer),
 
     /// The pointer that appears as the address result of addr_of(Local)
-    AddrOfLocal(Pointer, Local),
+    AddrOfLocal(Pointer, Local, u32),
+
+    /// The address of a constant value
+    AddrOfConst(Pointer, u32),
```
spernsteiner, line 66: Can you add comments here explaining the meaning of the various fields?  I guess for `AddrOfLocal` it's `(resulting pointer value, local whose address was taken, size of that local)`?  And similarly for `AddrOfConst` but without an indication of the thing whose address was taken?

spernsteiner, line 66: Can you also update the comment and/or rename the variant to clarify the meaning of `AddrOfConst`?  The name is surprising to me because constants are typically rvalues, not lvalues, and thus can't have their address taken (at least not directly).  From the way it's used, it looks like this is for `Rvalue::Use(Constant)` where a constant of pointer type is copied into a local - maybe `UseConst` would be a better name?  Or `AddrOfStatic`, since the pointee must be a static allocation of some sort (either `&STATIC` or `"foo"`).

dynamic_instrumentation/src/instrument.rs
```diff
@@ -343,43 +343,110 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+                        let _ = proj_iter.next();
+                    }
                 }
+            };
+
+            // We have some other elements between the two projections
+            if deref2_base.projection.len() - deref1_base.projection.len() > 1 {
+                self.loc(location, location, project_fn)
+                    .arg_var(deref1_base)
+                    .arg_addr_of(deref2_base.clone())
```
spernsteiner, line 390: Could you add a comment here with an example of what `deref1_base` and `deref2_base` might look like and what instrumentation should be generated as a result?  In particular, it's not clear to me why it's correct to take the address of `deref2_base` but not `deref1_base` (or maybe I've just misunderstood the meaning of the `Projection` fields).

dynamic_instrumentation/src/instrument.rs
```diff
@@ -343,43 +343,110 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+                // We are just loading the address for the next dereference
+                _ if have_deref2 => Some(load_fn),
+
+                NonMutatingUse(NonMutatingUseContext::Inspect) |
+                NonMutatingUse(NonMutatingUseContext::Copy) |
+                NonMutatingUse(NonMutatingUseContext::Move) => Some(load_fn),
+
+                MutatingUse(MutatingUseContext::Store) |
+                MutatingUse(MutatingUseContext::Call) |
+                MutatingUse(MutatingUseContext::AsmOutput) => Some(store_fn),
```
spernsteiner, line 405: In cases with 3+ derefs, is it correct to use `store_fn` for the derefs in the middle?  My usual intuition is that all `Deref` projections except the last are loads, and the last one depends on the context.  Given `*(*(*p).x).y`, my understanding is that the loop will run twice, once with `p` and `(*p).x` as `deref1_base` and `deref2_base`, and again with `(*p).x` and `(*(*p).x).y`.  But only the last of these is actually the target of a store - nothing stores to the intermediate `(*p).x`, even though it appears as `deref2_base` the first time around the loop.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -343,43 +343,110 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+                self.loc(location, location, ptr_fn)
+                    .arg_addr_of(deref2_base)
+                    .add_to(self);
+            }
+
+            if have_deref2 {
+                // Add an event for the pointer of the second dereference
+                self.loc(location, location, load_value_fn)
+                    .arg_var(deref2_base.clone())
+                    .add_to(self);
```
spernsteiner, line 432: Might be good to have an example here as well.  I wrote a big comment explaining why emitting `LoadValue` here is wrong, before finally realizing that it's correct :)

dynamic_instrumentation/src/instrument.rs
```diff
@@ -485,29 +527,29 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
                     .dest(&dest)
                     .add_to(self);
             }
-            Rvalue::Use(Operand::Constant(..)) => {
+            Rvalue::Use(op @ Operand::Constant(..)) => {
+                let const_ty = match op_ty(op).kind() {
+                    TyKind::RawPtr(TypeAndMut { ty, .. }) => *ty,
+                    TyKind::Ref(_, ty, _) => *ty,
+                    ty @ _ => unimplemented!("Unsupported constant ty {:?}", ty)
+                };
```
spernsteiner, line 542: I guess there's logic somewhere that prevents this from being called on non-pointers?  Otherwise `x = 0` would hit this `unimplemented!` case.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -485,29 +527,29 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
+                let layout = ctx.layout_of(ty::ParamEnv::reveal_all().and(const_ty))
+                    .expect("Failed to compute layout of constant");
+                let size = u32::try_from(layout.size.bytes())
+                    .expect("Failed to convert local size");
```
spernsteiner, line 546: What size does this come up with when `op` has type `&str`?  Zero?  I haven't looked into how the size field is used - is it okay to have a zero there?

spernsteiner (commented):
Initial comments.  I haven't looked at the PDG code yet.

pdg/src/builder.rs
```diff
@@ -57,7 +57,13 @@ impl EventKindExt for EventKind {
...
             FromInt(lhs) => lhs,
             Alloc { ptr, .. } => ptr,
-            AddrOfLocal(lhs, _) => lhs,
+            AddrOfLocal(lhs, _, _) => lhs,
+            AddrOfConst(ptr, _) => ptr,
+            // Corner case: Offset(..) events with a base pointer of zero are special
+            // because the result might be an actual pointer, e.g., c2rust will
+            // emit a pointer increment `a += b` as `a = a.offset(b)` which we need
+            // to ignore here if `a == 0` which is equivalent to `a = b`.
+            Offset(0, _, ptr) => ptr,
```
spernsteiner, line 66: Could this be detected and handled earlier (such as by replacing the `Offset` with a copy operation) to avoid the special case here?  This sounds like the sort of thing that would require matching special cases in several other places, which could easily be forgotten.

pdg/src/builder.rs
```diff
@@ -102,42 +109,102 @@ impl EventKindExt for EventKind {
...
+
+            if need_insert {
+                provenances.insert(ptr, mapping);
+            }
         }
-        AddrOfLocal(ptr, _) => {
+        CopyRef => unimplemented!(),
+        AddrOfLocal(ptr, _, size) => {
+            // TODO: is this a local from another function?
+            mapping.size = size.try_into().ok();
```
spernsteiner, line 194: Can this conversion ever fail?  If not, it seems like you could make the field `size: usize` instead of `Option<usize>` and avoid the `None` case above.

pdg/src/graph.rs
```diff
@@ -281,18 +283,26 @@ impl Display for DisplayNode<'_> {
...
     /// The nodes in the graph.  Nodes are stored in increasing order by timestamp.  The first
     /// node, called the "root node", creates the object described by this graph, and all other
     /// nodes are derived from it.
     #[serde(with = "crate::util::serde::index_vec")]
     pub nodes: IndexVec<NodeId, Node>,
+
+    /// Whether this graph was built from a null pointer. This should be `Some(true)` or
+    /// `Some(false)` in most cases, but we use `None` to represent an uninitialized value
+    /// to detect internal contradictions.
+    pub is_null: bool,
```
spernsteiner, line 297: Comment is out of date

spernsteiner (commented):
Seems generally reasonable.  This approach of collapsing nested projections down to a single byte offset will limit our ability to extract borrowchecker aliasing information from the PDG, if we end up wanting to do that in the future.

Are there any automated tests related to instrumentation or the PDG?  I thought we had one or two, but I could be wrong.

analysis/runtime/src/events.rs
ahomescu: Right, the current encoding does not distinguish sub-fields at offset 0 from their parents. I couldn't think of a case where it mattered, thanks for the example!

We can't break down nested field accesses into multiple events (the old implementation was broken and required some hackiness in the PDG), so I had an idea for an alternative: give each combination of nested subfields its own unique ID, and add that to the event. In your example above, `b.a.y` would get ID 0, `b.a.x` would be 1, and `b.a` would be 2. Then `b.a.x` and `b.a` could be differentiated by that number.

dynamic_instrumentation/src/instrument.rs
ahomescu: Line 397 is supposed to cover this case (`have_deref2` is essentially equivalent to "is not the last deref"). We should get here 3 times with the following `(deref1_base, deref2_base)` pairs:
* `p`, `(*p).x` with `have_deref2==true`, will always be a load because of 397
* `(*p).x)`, `(*(*p).x).y` with `have_deref2==true`, will always be a load because of 397
* `*(*p).x).y`, `*(*(*p).x).y` with `have_deref2==false` because of line 379, will be a store only if the place context needs it. `deref2_base` and `have_deref2` might be confusing here. There is no explicit dereference in the MIR, only the implicit one from reading/writing the place itself (e.g. `(*_p).x.y` is counted as an extra implicit dereference of the final pointer, so 2 in total).

dynamic_instrumentation/src/instrument.rs
ahomescu: Yes, line 475.
```
            _ if !is_region_or_unsafe_ptr(value_ty) => {}
```

dynamic_instrumentation/src/instrument.rs
ahomescu: Added a comment.

analysis/runtime/src/events.rs
spernsteiner: > `b.a.y` would get ID 0, `b.a.x` would be 1, and `b.a` would be 2. Then `b.a.x` and `b.a` could be differentiated by that number.

As long as there's a lookup table somewhere that can map those IDs back to paths (e.g. 0 -> `.a.y`), that sounds fine - it would provide the same information as the nested version.

dynamic_instrumentation/src/instrument.rs
spernsteiner: That makes sense, thanks

analysis/runtime/src/events.rs
ahomescu: I added that, and moved it into #1107.

pdg/src/builder.rs
ahomescu: The field would still need to be `size: Option<usize>` because we have other cases where we get a pointer without a known size. That comes from the `None` on line 268.

Some of this is happening because we get pointers from native libraries (e.g. glibc) which we have no control over. I think the best we can do right now is say "this is a pointer we know nothing about", and maybe add some clever heuristics later.

Re the conversion: I could do `Some(size.try_into().unwrap())`, that might be better for debugging?

analysis/runtime/src/events.rs
ahomescu: The MIR calls them "const", e.g., `_21 = const {alloc1082: *mut *mut fdlog_st}` and `_32 = const b"/usr/local/lib\x00"` and `_63 = const "non-null function pointer"`. They are source-level constants of aggregate types, e.g., strings and arrays.

I think everything from https://rustc-dev-guide.rust-lang.org/mir/index.html#mir-constant-values that's not converted to a literal becomes one of these "const"s.

dynamic_instrumentation/src/instrument.rs
ahomescu: Good catch, yeah `&str` has zero size. That's not good, I need to fix it.

pdg/src/builder.rs
ahomescu: Moved it to the event handlers.

analysis/runtime/src/events.rs
spernsteiner: The RHSs of those examples are all `Operand::Constant`s, and MIR only allows taking the address of a place, not a constant.  rustc handles string literals and such by creating an anonymous static allocation and emitting an `Operand::Constant` whose value is a pointer to that allocation.  It handles accesses to normal static items similarly - take a look at the MIR for [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1629be21656514623d3d3cc4769915f7).

As far as I know, any time you have a constant of pointer/reference type, it must either not point to anything (`123 as *const u8`, `&()`) or must be a pointer to a static allocation (possibly anonymous).  That's why I suggested `AddrOfStatic` as a possible name for this variant.

dynamic_instrumentation/src/instrument.rs
```diff
@@ -531,49 +653,51 @@ impl<'tcx> Visitor<'tcx> for CollectInstrumentationPoints<'_, 'tcx> {
...
-                        .add_to(self);
-                };
+                        .instrumentation_priority(InstrumentationPriority::Early);
+                    if matches!(value, Rvalue::Ref(_, BorrowKind::Mut { .. }, _)) {
+                        b = b.arg_addr_of(*p);
+                    } else {
+                        b = b.arg_var(dest);
+                    }
+                    b = b.arg_var(proj_idx);
+                    b.add_to(self);
```
spernsteiner, line 699: I was confused at first by how different the `if`/`Ref` and `else`/`AddressOf` cases look here.  But after looking at it more, I think `arg_addr_of(*p)` and `arg_var(dest)` produce the same value, but handle the types differently.  In the `AddressOf` case, `dest` already holds the address of `p` as a raw pointer, so you can just pass `dest` to the instrumentation function.  But in the `Ref` case, `dest` holds the address as a reference, so you take the address as a raw pointer instead using `arg_addr_of(*p)`.  Is that correct?  And would it be legal / reasonable to use `arg_addr_of(*p)` unconditionally?

pdg/src/builder.rs
```diff
@@ -102,42 +104,118 @@ impl EventKindExt for EventKind {
...
+                    false
+                })
+                .unwrap_or(true);
+
+            if need_insert {
+                provenances.insert(ptr, mapping);
+            }
         }
-        AddrOfLocal(ptr, _) => {
+        CopyRef => unimplemented!(),
```
spernsteiner, line 202: Is it okay for this to remain unimplemented?  Could this cause trouble when running on partially refactored code that uses safe refs in some places?

spernsteiner (approved):
Looks reasonable overall.  Left a few comments.

dynamic_instrumentation/src/instrument.rs
ahomescu: The old code used the reference, but I'm not sure why. `arg_addr_of(*p)` should always be correct.

pdg/src/builder.rs
ahomescu: The actual problem is the dynamic instrumentation doesn't ever emit `CopyRef` right now. I'm wondering if we should just remove this node kind.

ahomescu: Update: I removed `CopyRef`.

analysis/runtime/src/metadata.rs
```diff
@@ -46,11 +48,17 @@ impl FromIterator<Metadata> for Metadata {
...
             functions.extend(metadata.functions);
+            projections.extend(metadata.projections);
+        }
+        Self {
+            locs,
+            functions,
+            projections,
         }
-        Self { locs, functions }
     }
```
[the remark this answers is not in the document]
ahomescu: I switched from `IndexSet` to `HashMap` with random 64-bit keys. That means different projection sets can be merged together, and the probability of key collision should be low.

analysis/runtime/src/events.rs
ahomescu: > The RHSs of those examples are all `Operand::Constant`s, and MIR only allows taking the address of a place, not a constant. rustc handles string literals and such by creating an anonymous static allocation and emitting an `Operand::Constant` whose value is a pointer to that allocation. It handles accesses to normal static items similarly - take a look at the MIR for [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1629be21656514623d3d3cc4769915f7).

Right, but in that example even the `static` is initialized in MIR to a `const`:
```
static X: i32 = {
    let mut _0: i32;

    bb0: {
        _0 = const 123_i32;
        return;
    }
}
```

The static itself is initialized to the value of a `const` pointer.
 
> As far as I know, any time you have a constant of pointer/reference type, it must either not point to anything (`123 as *const u8`, `&()`) or must be a pointer to a static allocation (possibly anonymous). That's why I suggested `AddrOfStatic` as a possible name for this variant.

One significant difference between constants and statics (at least at the Rust language level): static can be mutable. I think it makes more sense to use `AddrOfConst` for anything that's an immutable constant (even if it's internally a static allocation), and use `AddrOfStatic` for things that are explicitly Rust `static` declarations (and potentially mutable, either via `mut` or e.g. `Cell`).

I'll admit that for the purpose in this PR (determining the size of a pointee) the difference between `const` and `static` isn't too relevant, since statics also need to be `Sized`.

spernsteiner: I just think `AddrOfConst` is a confusing name because things declared `const` in Rust don't have addresses.  This applies to `const C: i32 = 123;`, `fn f<const N: usize>()`, and the new const expressions `const { 2 + 2 }`.  `C`, `N`, and `const { 2 + 2 }` all become `Operand::Constant`s in the MIR, and you can't take the address of an `Operand::Constant`.  (You can write `&C`, `&N`, and `&const { 2 + 2 }` in Rust, but at the MIR level, this actually stores the `Operand::Constant` to a temporary and then takes the address of that temporary.)

>The static itself is initialized to the value of a const pointer.

In this particular example, it happens to be initialized with an `Operand::Constant` of type `i32`, but that isn't always the case.  For example, `static X: i32 = 123 + 456;` produces this MIR:
```Rust
static X: i32 = {
    let mut _0: i32;
    let mut _1: (i32, bool);

    bb0: {
        _1 = AddWithOverflow(const 123_i32, const 456_i32);
        assert(!move (_1.1: bool), "attempt to compute `{} + {}`, which would overflow", const 123_i32, const 456_i32) -> [success: bb1, unwind continue];
    }

    bb1: {
        _0 = move (_1.0: i32);
        return;
    }
}
```

Accessing `X` uses a const pointer:

```Rust
    bb0: {
        _1 = const {alloc1: &i32};
        _0 = (*_1);
        return;
    }
```

Essentially what has happened here is the expression `X` was expanded to `*(&X)`, then `&X` was const-evaluated (since the address of a static is a constant) and produced `{alloc1: &i32}`.  The statement `_1 = const {alloc1: &i32};` stores the address of the static `X` into `_1`, hence my suggestion to name this operation `AddrOfStatic`.

>One significant difference between constants and statics (at least at the Rust language level): static can be mutable.

Compilation of `static mut` works roughly the same way:

```Rust
    bb0: {
        _1 = const {alloc1: *mut i32};
        _2 = _1 as *const () (PtrToPtr);
        _3 = _2 as usize (Transmute);
        _4 = AlignOf(i32);
        _5 = Sub(_4, const 1_usize);
        _6 = BitAnd(_3, _5);
        _7 = Eq(_6, const 0_usize);
        assert(_7, "misaligned pointer dereference: address must be a multiple of {} but is {}", _4, _3) -> [success: bb1, unwind unreachable];
    }

    bb1: {
        _0 = (*_1);
        return;
    }
```

Here it's turned the expression `X` into `*(ptr::addr_of_mut!(X))`, and it const-evaluated `ptr::addr_of_mut!(X)` to `{alloc1: *mut i32}`.  This is again storing the address of the static into `_1`.  (The rest of `bb0` is an alignment check for the raw-pointer dereference - I guess rustc now generates those automatically in debug mode.)

---

In the interest of not spending too much more time on this particular detail, if you still disagree about the naming, just mark this resolved and go ahead with the current one.

ahomescu: Looking at the event code again, I'm thinking maybe it makes more sense to call it `AddrOfSized` anyway, since that's what it actually does. Then it could be used for constants _and_ locals.

PR #1087, opened 2024-04-29
analyze: uncomment line in scripts/run_pointwise_metrics.sh

spernsteiner:
In #1074, I had commented out a line to speed up testing, and accidentally left it that way when I merged the branch.


fw-immunant (approved):
Using `:` no-op for comments in shell scripts is wild, I'd never thought of that. It works better with line continuations but of course isn't syntax-highlit; I missed this when reviewing too.

PR #1088, opened 2024-04-30
analyze: allow overriding dataflow for specific permissions

spernsteiner:
This adds a new internal feature for overriding dataflow analysis for specific permissions of specific pointers.  The `propagate` method of `dataflow` now takes an additional `updates_forbidden` set, which has a `PermissionSet` mask for every `PointerId`, and avoids adding or removing a permission for a `ptr` if the corresponding bit is set in `updates_forbidden[ptr]`.  When `updates_forbidden` is used, the resulting permissions after running `dataflow` might not actually satisfy the dataflow constraints.

This is designed to support the PDG "`NON_NULL` override" feature, where information about nullability from the PDG can override static analysis results, though that feature is not part of the current PR.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -1088,41 +1088,14 @@ fn run(tcx: TyCtxt) {
     // Run dataflow solver and borrowck analysis
     // ----------------------------------
 
-    // For testing, putting #[c2rust_analyze_test::fail_before_analysis] on a function marks it as
-    // failed at this point.
-    for &ldid in &all_fn_ldids {
-        if !util::has_test_attr(tcx, ldid, TestAttr::FailBeforeAnalysis) {
-            continue;
-        }
-        gacx.mark_fn_failed(
-            ldid.to_def_id(),
-            DontRewriteFnReason::FAKE_INVALID_FOR_TESTING,
-            PanicDetail::new("explicit fail_before_analysis for testing".to_owned()),
-        );
-    }
-
-    // For testing, putting #[c2rust_analyze_test::force_non_null_args] on a function marks its
-    // arguments as `NON_NULL` and also adds `NON_NULL` to the `updates_forbidden` mask.
-    for &ldid in &all_fn_ldids {
-        if !util::has_test_attr(tcx, ldid, TestAttr::ForceNonNullArgs) {
-            continue;
-        }
-
-        let info = func_info.get_mut(&ldid).unwrap();
-        let mut asn = gasn.and(&mut info.lasn);
-        let mut updates_forbidden = g_updates_forbidden.and_mut(&mut info.l_updates_forbidden);
-
-        let lsig = &gacx.fn_sigs[&ldid.to_def_id()];
-        for arg_lty in lsig.inputs.iter().copied() {
-            for lty in arg_lty.iter() {
-                let ptr = lty.label;
-                if !ptr.is_none() {
-                    asn.perms_mut()[ptr].insert(PermissionSet::NON_NULL);
-                    updates_forbidden[ptr].insert(PermissionSet::NON_NULL);
-                }
-            }
-        }
-    }
+    apply_test_attr_fail_before_analysis(&mut gacx, &all_fn_ldids);
+    apply_test_attr_force_non_null_args(
+        &mut gacx,
+        &all_fn_ldids,
+        &mut func_info,
+        &mut gasn,
+        &mut g_updates_forbidden,
+    );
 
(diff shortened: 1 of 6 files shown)


c2rust-analyze/src/analyze.rs
```diff
@@ -554,6 +554,8 @@ struct FuncInfo<'tcx> {
     /// get a complete [`Assignment`] for this function, which maps every [`PointerId`] in this
     /// function to a [`PermissionSet`] and [`FlagSet`].
     lasn: MaybeUnset<LocalAssignment>,
+    /// Local part of the `updates_forbidden` mask.
+    l_updates_forbidden: MaybeUnset<LocalPointerTable<PermissionSet>>,
```
ahomescu, line 558: Would it be simpler if we used a separate permission (e.g. `DYNAMIC_NON_NULL`) that doesn't get propagated when it shows up? Would that even work?

c2rust-analyze/src/analyze.rs
```diff
@@ -1095,6 +1101,29 @@ fn run(tcx: TyCtxt) {
         );
     }
 
+    // For testing, putting #[c2rust_analyze_test::force_non_null_args] on a function marks its
```
ahomescu, line 1104: This function is already over 700 lines, could we move this step into a separate function?

c2rust-analyze/src/dataflow/mod.rs
```diff
@@ -118,27 +125,52 @@ impl DataflowConstraints {
...
         rules: &mut R,
+        updates_forbidden: Option<&PointerTable<T>>,
     ) -> Result<bool, String>
     where
         T: PartialEq,
         R: PropagateRules<T>,
     {
         let mut xs = TrackedPointerTable::new(xs.borrow_mut());
 
+        let restrict_updates = |rules: &mut R, ptr, old: &T, new: T| {
```
ahomescu, line 166: You're capturing `updates_forbidden`, why not `rules` too?

c2rust-analyze/src/dataflow/mod.rs
```diff
@@ -373,6 +421,9 @@ trait PropagateRules<T> {
     ) -> (T, T);
     fn all_perms(&mut self, ptr: PointerId, perms: PermissionSet, val: &T) -> T;
     fn no_perms(&mut self, ptr: PointerId, perms: PermissionSet, val: &T) -> T;
+    /// Apply a filter to restrict updates.  The result is similar to `new`, but all flags marked
+    /// in `updates_forbidden` are adjusted to match their `old` values.
+    fn restrict_updates(&mut self, old: &T, new: &T, updates_forbidden: &T) -> T;
```
ahomescu, line 426: Are both implementations of this identical? If so, it might make sense to just provide the implementation here.

c2rust-analyze/src/dataflow/mod.rs
spernsteiner: The closure would hold a borrow on `rules` for its lifetime, conflicting with the calls to `rules.subset(...)` and such below.

c2rust-analyze/src/dataflow/mod.rs
spernsteiner: That would require a bunch of constraints on `T` in the trait definition.  (The constraints are already satisfied by both impls, but would be annoying to write down.)

c2rust-analyze/src/analyze.rs
spernsteiner: I'm not sure what you mean by "doesn't get propagated".

I could imagine propagating `NON_NULL` normally and having a separate, non-propagated `DYNAMIC_NON_NULL` flag that fills in for a missing `NON_NULL` at rewriting time.  But we want the dynamic information to cut off propagation of nullability, which this wouldn't provide.  For example, suppose null flows into `p`, `p` into `q`, and `q` into `r`, and `q` is set to `NON_NULL` with `updates_forbidden`.  Then `r` would also be `NON_NULL` because the nullability of `p` is prevented from propagating through `q` due to `updates_forbidden`.

We could instead have a non-propagated flag that affects the propagation of `NON_NULL` in the same way that `updates_forbidden` does, but I don't think that implementation would be any simpler than the current one.

ahomescu (approved):

c2rust-analyze/src/analyze.rs
spernsteiner: Moved this and the adjacent bit of `TestAttr` handling into their own helper functions: 8851c89.  As we continue working on this code, we can do more refactoring as we go.

PR #1091, opened 2024-05-06
Switch to crossbeam-queue for events

ahomescu:
Use a lock-free queue to store the event list because the event callbacks might be called from inside a signal handler. Most system calls are not safe in that context, so we just spin if the queue is full.


analysis/runtime/src/runtime/backend.rs
```diff
@@ -91,7 +103,7 @@ impl Backend {
         self.flush();
     }
 
-    pub fn run(&mut self, events: impl IntoIterator<Item = Event>) {
+    pub fn run(&mut self, events: Arc<ArrayQueue<Event>>) {
         let (lock, cvar) = &*FINISHED;
         let mut finished = lock.lock().unwrap();
```
kkysen, line 108: What about this lock?

kkysen: There's also `MainThreadRuntime::send_event`, which calls `Mutex::lock`, too, on a `Mutex<Backend>`.  Also, are we sure allocations can't happen in any of these calls?

kkysen (commented):
Did we run into this causing problems?  Signal safety is really annoying.

analysis/runtime/src/runtime/backend.rs
spernsteiner: >are we sure allocations can't happen in any of these calls?

`ArrayQueue` docs say:

>This queue allocates a fixed-capacity buffer on construction, which is used to store pushed elements.... Having a buffer allocated upfront makes this queue a bit faster than SegQueue.

Which sounds like it doesn't allocate on push.

>There's also MainThreadRuntime::send_event, which calls Mutex::lock

@ahomescu, can you clarify specifically which functions should support being called from a signal handler context?

kkysen: > Which sounds like it doesn't allocate on push.

There's a bunch of other functions here that look like they could potentially called inside a signal handler.  I understand this one is specifically designed not to allocate, but I'm not sure about the others.  Like bincode serialization or formatting output, etc.  They might not, but we need to check, right?

spernsteiner: My vague understanding of the design here is that the instrumentation calls we insert push events onto a queue, and a background thread processes those events.  Since we potentially could be instrumenting signal handlers, the instrumentation/push code needs to be lock-free and allocation-free.  But the background thread's processing (including bincode/serialization stuff) doesn't run in a signal handler and so isn't subject to those constraints.

kkysen: That makes sense.  So then only `ExistingRuntime::send_event` needs to be async-signal-safe, meaning `fn run` and `fn write_all` here don't need to be, right?

analysis/runtime/src/runtime/scoped_runtime.rs
```diff
@@ -172,7 +179,8 @@ impl Runtime for BackgroundThreadRuntime {
     /// Initialize the [`BackgroundThreadRuntime`], which includes [`thread::spawn`]ing,
     /// so it must be run post-`main`.
     fn try_init(mut backend: Backend) -> Result<Self, AnyError> {
-        let (tx, rx) = mpsc::sync_channel(1024);
+        let tx = Arc::new(ArrayQueue::new(1048576));
```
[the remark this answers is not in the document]
ahomescu: 1048576 is 1024 * 1024. I could make it a round 1 million elements, not sure it needs to be a power of two.

analysis/runtime/src/runtime/backend.rs
ahomescu: Right, it's exactly what @spernsteiner said: the push code needs to be signal-safe, but the receiver doesn't because it runs in its own thread.

analysis/runtime/src/runtime/scoped_runtime.rs
```diff
@@ -117,16 +115,27 @@ impl Runtime for MainThreadRuntime {
...
+    }
+}
+
 impl ExistingRuntime for BackgroundThreadRuntime {
     fn finalize(&self) {
         // Only run the finalizer once.
         self.finalized.get_or_init(|| {
             // Notify the backend that we're done.
-            self.tx.send(Event::done()).unwrap();
+            self.push_event(Event::done());
```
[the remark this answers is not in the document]
ahomescu: Right, should be fine. I'll go see if I can add a blocking implementation.

spernsteiner:
What's left to do here?

>Edit: Should I go audit all the functions in the runtime and check exactly which ones need to be signal-safe? Shouldn't be too hard.

@ahomescu, did you already do this / do you still plan to?

ahomescu:
> What's left to do here?
> 
> > Edit: Should I go audit all the functions in the runtime and check exactly which ones need to be signal-safe? Shouldn't be too hard.
> 
> @ahomescu, did you already do this / do you still plan to?

I added a bunch of comments on signal safety, and removed a few potentially unsafe `eprintln!` calls.

analysis/runtime/src/handlers.rs
```diff
@@ -2,6 +2,11 @@ use crate::events::{Event, EventKind};
 use crate::mir_loc::MirLocId;
 use crate::runtime::global_runtime::RUNTIME;
 
+// WARNING! Most handlers in this file may be called from a signal handler,
+// so they and all their callees should be signal-safe.
```
kkysen, line 6: Nit: Isn't the technical term `async-signal-safe` specifically?

```suggestion
// so they and all their callees should be async-signal-safe.
```

analysis/runtime/src/runtime/scoped_runtime.rs
```diff
@@ -147,12 +168,14 @@ impl ExistingRuntime for BackgroundThreadRuntime {
...
-                self.tx.send(event).unwrap();
+                // # Signal-safety: `push_event` is safe if `can_sleep == false`
+                self.push_event(event, false);
             }
             Some(()) => {
-                // Silently drop the [`Event`] as the [`BackgroundThreadRuntime`] has already been [`BackgroundThreadRuntime::finalize`]d.
+                // Silently drop the [`Event`] as the [`BackgroundThreadRuntime`]
+                // has already been [`BackgroundThreadRuntime::finalize`]d.
+                //
+                // # Signal-safety: Should be fine even after the end of `main`.
```
kkysen, line 178: Should we note that this is only true for `once_cell::OnceCell`'s (default) `feature = "std"` implementation?  `std::sync::OnceLock`, the supposed replacement, uses a `futex_wait`.

Also, while pre `main` there shouldn't be any signal handlers installed yet, post `main`, any signal handlers installed probably weren't uninstalled, so a post `main` signal seems more likely than a pre `main` one.

kkysen: Oh, nevermind, I see you've removed the code in `AfterMain`.

analysis/runtime/src/runtime/skip.rs
```diff
@@ -45,15 +45,11 @@ pub(super) fn skip_event(event: Event, reason: SkipReason) {
...
+            // # Signal-safety: atomic increments are safe.
             EVENTS_SKIPPED_BEFORE_MAIN.fetch_add(1, Ordering::Relaxed);
         }
         AfterMain => {
-            // This is after `main`, so it's safe to use things like `eprintln!`,
-            // which uses `ReentrantMutex` internally, which may use `pthread` mutexes.
-            WARNED_AFTER_MAIN.get_or_init(|| {
-                eprintln!("skipping {reason}");
-            });
-            eprintln!("skipped event after `main`: {:?}", event.kind);
```
kkysen, line 56: Shouldn't we still be able to print to the unbuffered `stderr` if we make no allocations?  It's useful to have at least some sort of warning here.

ahomescu: I initially thought `write` wasn't signal-safe, but https://man7.org/linux/man-pages/man7/signal-safety.7.html says it is. The first `eprintln!` is probably safe to call (I'd need to check its implementation), but the second one probably isn't because the formatter allocates.

analysis/runtime/src/handlers.rs
ahomescu: Right, I'll update the comments. IMHO it's still weird because async-signal-safe implies the existence of sync-signal-safe, which I haven't heard of. Maybe they mean `signalfd`???

analysis/runtime/src/runtime/skip.rs
kkysen: The second one is more useful and they both format an argument.  The second one can be changed to also `eprintln!("{}", event.kind.as_str())` or something like that, though.

kkysen: If `write` wasn't async-signal-safe that would be very bad.  Then you can't do much at all in the signal handler (like the pipe trick).

ahomescu: I put the old code back in with some changes. The `eprintln!` is still problematic, but I don't see a way to get rid of it.

analysis/runtime/src/runtime/global_runtime.rs
```diff
@@ -40,10 +40,16 @@ impl GlobalRuntime {
...
+    ///
+    /// May be called from a signal handler, so it needs to be async-signal-safe.
     pub fn send_event(&self, event: Event) {
+        // # Async-signal-safety: OnceCell::get() is just a dereference
         match self.runtime.get() {
             None => {
                 // Silently drop the [`Event`] as the [`ScopedRuntime`] isn't ready/initialized yet.
+                //
+                // # Async-signal-safety: unlikely we wind up in a signal handler
+                // before `main`, and `BeforeMain` is async-signal-safe anyway.
```
spernsteiner, line 52:
```suggestion
                // # Async-signal-safety: `skip_event(_, BeforeMain)` is async-signal-safe.
```
I found these "should be safe" comments confusing

analysis/runtime/src/runtime/scoped_runtime.rs
```diff
@@ -147,12 +168,14 @@ impl ExistingRuntime for BackgroundThreadRuntime {
...
-                self.tx.send(event).unwrap();
+                // # Async-signal-safety: `push_event` is safe if `can_sleep == false`
+                self.push_event(event, false);
             }
             Some(()) => {
-                // Silently drop the [`Event`] as the [`BackgroundThreadRuntime`] has already been [`BackgroundThreadRuntime::finalize`]d.
+                // Silently drop the [`Event`] as the [`BackgroundThreadRuntime`]
+                // has already been [`BackgroundThreadRuntime::finalize`]d.
+                //
+                // # Async-signal-safety: Should be fine even after the end of `main`.
```
spernsteiner, line 178:
```suggestion
                // Silently drop the [`Event`] as the [`BackgroundThreadRuntime`]
                // has already been [`BackgroundThreadRuntime::finalize`]d.
                //
                // # Async-signal-safety: `skip_event(_, AfterMain)` is NOT SAFE;
                // however, see the comment in `skip_event` for an explanation
                // of why this will probably be okay in practice.
```

analysis/runtime/src/runtime/backend.rs
```diff
@@ -80,18 +83,32 @@ pub enum Backend {
...
-        for event in events {
+    fn write_all(&mut self, events: Arc<ArrayQueue<Event>>) {
+        let backoff = Backoff::new();
+        loop {
+            let event = match events.pop() {
+                Some(event) => event,
+                None => {
+                    // We can't use anything with a futex here since
+                    // the event sender might run inside a signal handler
+                    backoff.snooze();
```
spernsteiner, line 94:
```suggestion
                    // We can't block on a lock/semaphore here since
                    // it might not be safe for the event sender to wake
                    // us from inside a signal handler
                    backoff.snooze();
```
Took me a minute to figure out how a futex here would relate to the event sender's signal handler

spernsteiner (approved):

PR #1092, opened 2024-05-10
Add a set of comments and question on PDG construction

ahomescu:
I wrote my questions on the PDG builder algorithm in the form of comments. It might be useful to answer (some of) these and leave them in as comments, for future reference.


pdg/src/builder.rs
```diff
@@ -44,6 +44,7 @@ impl EventKindExt for EventKind {
         use EventKind::*;
         Some(match *self {
             CopyPtr(lhs) => lhs,
+            // Question: this takes a pointer as input and produces another?
             Field(ptr, ..) => ptr,
```
spernsteiner, line 48: > Question: this takes a pointer as input and produces another?

That's what I would expect.  In the static analysis, we have a similar notion of `PlaceElem::Field` as a "pointer projection" that takes a pointer to the struct and produces a pointer to the field.  This is similar to (some uses of) `offset`, which takes a pointer to (the first element of) an array and produces a pointer to an element, and also similar to [LLVM's GetElementPtr operation](https://llvm.org/docs/GetElementPtr.html).

pdg/src/builder.rs
```diff
@@ -54,10 +55,12 @@ impl EventKindExt for EventKind {
             StoreValue(ptr) => ptr,
             CopyRef => return None, // FIXME
             ToInt(ptr) => ptr,
+            // Question: this has two pointers, is the input one sufficient?
             Realloc { old_ptr, .. } => old_ptr,
             FromInt(lhs) => lhs,
             Alloc { ptr, .. } => ptr,
             AddrOfLocal(lhs, _) => lhs,
+            // Question: same, is the base pointer enough?
             Offset(ptr, _, _) => ptr,
```
spernsteiner, line 64: >is the base pointer enough?

At one point, we planned to distinguish forward offsets (`OFFSET_ADD`) from backward ones (`OFFSET_SUB`).  Offsets that statically only go forward can be implemented on slices with a subslice operation, but offsets that might go backward need a different type that tracks how far back it can go and remain in bounds.

PR #1094, opened 2024-05-14
Fix Nix

DieracDelta:
The current nix expression doesn't build for me on nixos on 23.11 stable (bunch of various build errors that I can share if useful). This nix flake builds the project without erroring (e.g. `nix develop -c cargo build --release`)  and also brings in the rust toolchain declaratively with nix instead of with rustup.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,183 @@
+{
+  "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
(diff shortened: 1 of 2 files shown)


kkysen (commented):
Hi!  We're not in a position to review and maintain nix things for c2rust, but we may be able to merge this (like the other existing nix things) with no promises about maintainability.  If this fixes nix build issues, that's great!  So thanks for this!

PR #1095, opened 2024-05-16
analyze: add NON_NULL rewrites

spernsteiner:
This branch implements rewriting of nullable pointers (those lacking `PermissionSet::NON_NULL`) to `Option`.  This includes the following kinds of rewrites:

* Type annotations: `Option<&mut T>` instead of `&mut T`
* Casts between nullable and non-nullable via `p.unwrap()` and `Some(p)`.  For most permissions, we implement only one direction because the other is invalid/nonsensical, but here we implement both in order to support "unsound" rewrites involving overridden `NON_NULL` flags (as in #1088).
* Borrows: when casting `p: Option<&mut T>`, we use `p.as_deref_mut().unwrap()` instead of `p.unwrap()` to avoid consuming the original `p`.  (In the code, this is called a "downgrade", since it allows borrowing `Option<Box<T>>` as `Option<&T>` and similar.)
* Pointer projections on nullable pointers.  Where `NON_NULL` pointers would use `&p[0]`, nullable ones use `p.map(|ptr| &ptr[0])`.  Internally, this is represented similar to `Some(&p.unwrap()[0])`, but it's handled specially by `rewrite::expr::convert` to produce a `map` call instead, which passes through `None` without a panic.
* `unwrap()` calls on derefs.  `*p` is rewritten to `*p.unwrap()`, or to `*p.as_deref().unwrap()` if a downgrade/borrow is necessary to avoid moving `p`.
* `ptr::null()` and `0 as *const _` to `None`, and `p.is_null()` to `p.is_none()`.

The new `non_null_rewrites.rs` test case passes, and the rewritten code compiles.

This might be easier to review commit-by-commit.


spernsteiner:
I realized while writing the PR that I never tested field projections (`q = &(*p).field`).  I'll test it but will leave any fixes to a separate PR, since this one is already fairly complex.

c2rust-analyze/src/rewrite/ty.rs
```diff
@@ -333,37 +333,46 @@ impl Convert<hir::Mutability> for mir::Mutability {
...
+    let mut path_parts_iter = path.split("::");
+    let crate_name = path_parts_iter
+        .next()
+        .unwrap_or_else(|| panic!("couldn't find crate name in {path:?}"));
+
+    let krate = tcx
         .crates(())
         .iter()
         .cloned()
         .find(|&krate| tcx.crate_name(krate).as_str() == "core")
```
ahomescu, line 346: `crate_name` here instead of `"core"`?

ahomescu, line 346: `crate_name` here instead of `"core"`?

ahomescu:
Left a couple of comments. This is a large PR, it will take me a while to get through.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -461,6 +501,35 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
+                            }
+                        });
+                    }
+
+                    Callee::Null { .. } => {
+                        self.enter_rvalue(|v| {
+                            if !v.flags[pl_ty.label].contains(FlagSet::FIXED) {
+                                let arg_non_null =
+                                    v.perms[pl_ty.label].contains(PermissionSet::NON_NULL);
+                                if arg_non_null {
```
ahomescu, line 524: Should it just be `assert!(!arg_non_null, "...");`?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -484,14 +553,18 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
             Rvalue::Repeat(ref op, _) => {
                 self.enter_rvalue_operand(0, |v| v.visit_operand(op, None));
             }
-            Rvalue::Ref(_rg, _kind, pl) => {
-                self.enter_rvalue_place(0, |v| v.visit_place(pl));
+            Rvalue::Ref(_rg, kind, pl) => {
+                let mutbl = match kind {
```
ahomescu, line 557:
```suggestion
                let mutbl = matches!(kind, BorrowKind::Mut { .. });
```

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -801,6 +958,58 @@ where
             return Ok(());
         }
 
+        if from.option && from.own != to.own {
+            // Downgrade ownership before unwrapping the `Option` when possible.  This can avoid
+            // moving/consuming the input.  For example, if the `from` type is `Option<Box<T>>` and
+            // `to` is `&mut T`, we start by calling `p.as_mut().as_deref()`, which gives
```
ahomescu, line 964: Wouldn't that be `p.as_deref_mut()`?

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -801,6 +958,58 @@ where
...
+                    }
+                    Ownership::RawMut | Ownership::Cell | Ownership::Mut => {
+                        (self.emit)(RewriteKind::OptionDowngrade {
+                            mutbl: true,
+                            deref: true,
+                        });
+                        from.own = Ownership::Mut;
+                    }
+                    _ => {
+                        // Remaining cases are unsupported.
```
ahomescu, line 983: We could log this.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -801,6 +958,58 @@ where
...
+            }
+        }
+
+        let mut in_option_map = false;
+        if from.option && !to.option {
+            // Unwrap first, then perform remaining casts.
+            (self.emit)(RewriteKind::OptionUnwrap);
+            from.option = false;
+        } else if from.option && to.option {
+            eprintln!("try_build_cast_desc_desc: emit OptionMapBegin");
```
ahomescu, line 995: Just a general comment: the analyzer should use a logger like `log` with e.g. `env_logger` as the back end, so all this debugging output can be silenced (`2>/dev/null` does technically work, but it's less configurable).

We don't need to fix all the existing lines in this PR, but maybe we could switch these? Or alternatively a follow-up PR that switches everything.

c2rust-analyze/src/rewrite/ty.rs
spernsteiner: Fixed

c2rust-analyze/src/type_desc.rs
```diff
@@ -39,10 +39,51 @@ pub enum Quantity {
...
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
+pub struct PtrDesc {
+    pub own: Ownership,
+    pub qty: Quantity,
+    pub option: bool,
+}
+
+impl<'tcx> From<TypeDesc<'tcx>> for PtrDesc {
+    fn from(x: TypeDesc<'tcx>) -> PtrDesc {
+        let TypeDesc {
```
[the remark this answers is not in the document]
spernsteiner: I'm not a fan of this style for function arguments, and I don't think we currently use it anywhere else in the codebase

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -119,6 +119,284 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                assert!(matches!(hir_rw, Rewrite::Identity));
+                Rewrite::Text("None".into())
+            }
+
+            mir_op::RewriteKind::MemcpySafe {
+                elem_size,
+                dest_single,
+                src_single,
+            } => {
+                // `memcpy(dest, src, n)` to a `copy_from_slice` call
```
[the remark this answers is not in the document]
spernsteiner: We could maybe do some of those, but it's out of scope for this PR.  This code is copied/refactored from existing code - it isn't new in this PR.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -119,6 +119,284 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                    Some(Box::new(format_rewrite!("dest"))),
+                )
+            }
+
+            mir_op::RewriteKind::MemsetZeroize {
+                ref zero_ty,
+                elem_size,
+                dest_single,
+            } => {
+                // `memcpy(dest, src, n)` to a `copy_from_slice` call
```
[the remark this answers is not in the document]
spernsteiner: Fixed

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Fixed.  The `Null` case was originally symmetric with the `IsNull` case, but I don't think there's a good reason to have symmetry between the two.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Writing out all the cases ensures that if a new variant appears in a future rustc version, we'll get a compile error and remember to update this match, instead of potentially having it silently do the wrong thing.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Fixed.  I originally implemented rewriting to `p.as_mut().as_deref_mut()`, but this doesn't produce the right output type.  I fixed the rewriting to produce `p.as_deref_mut()`, but forgot to update the comment.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -119,6 +119,284 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+        /// Find the next `OptionMapEnd` and return the parts of `mir_rws` that come strictly
+        /// before it and strictly after it.  Returns `None` if there's an `OptionMapBegin` before
+        /// the next `OptionMapEnd`, or if there's no `OptionMapEnd` found in `mir_rws`.
+        fn split_option_map_rewrites(
+            mir_rws: &[DistRewrite],
+        ) -> Option<(&[DistRewrite], &[DistRewrite])> {
+            for (i, mir_rw) in mir_rws.iter().enumerate() {
+                match mir_rw.rw {
+                    // Bail out if we see nested delimiters.
+                    mir_op::RewriteKind::OptionMapBegin => return None,
```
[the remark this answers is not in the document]
spernsteiner: Added some clarification to the comment in ff2026304.  Bailing out here is not a hard error - it is legal, but suboptimal, to rewrite `OptionMapBegin` as `p` -> `p.unwrap()`.  Also, currently `mir_op` should never generate rewrites that trigger this.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -470,6 +608,45 @@ pub fn convert_cast_rewrite(kind: &mir_op::RewriteKind, hir_rw: Rewrite) -> Rewr
...
+            Rewrite::MethodCall("unwrap".to_string(), Box::new(hir_rw), vec![])
+        }
+        mir_op::RewriteKind::OptionSome => {
+            // `p` -> `Some(p)`
+            Rewrite::Call("std::option::Option::Some".to_string(), vec![hir_rw])
+        }
+
+        mir_op::RewriteKind::OptionMapBegin => {
+            // `p` -> `p.unwrap()`
+            Rewrite::MethodCall("unwrap /*map_begin*/".to_string(), Box::new(hir_rw), vec![])
```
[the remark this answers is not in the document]
spernsteiner: I initially added the `/*map_begin*/` comment for debugging, and I think it's actually useful to leave it in place to help us catch any cases where the `Option::map` rewrite is not working properly.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -119,6 +119,284 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+            return convert_cast_rewrite(rw, hir_rw);
+        }
+        let ex = ex.unwrap();
+
+        // Cases that extract a subexpression are handled here; cases that only wrap the
+        // top-level expression (and thus can handle a non-`Identity` `hir_rw`) are handled by
+        // `convert_cast_rewrite`.
+        match *rw {
+            mir_op::RewriteKind::OffsetSlice { mutbl } => {
+                // `p.offset(i)` -> `&p[i as usize ..]`
```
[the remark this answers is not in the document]
spernsteiner: There's logic in `rewrite::apply` to add parentheses where needed.  It always adds them around `&foo` to avoid this specific issue.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: I switched out the println calls that were added in this PR.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: I updated this comment to clarify: 73652be

spernsteiner:
@ahomescu I've addressed all the previous feedback - any objections to merging this now?

ahomescu (approved):

PR #1101, opened 2024-07-01
Update Cargo.toml

anoraktrend:
fix time bug


theoparis:
I think a better solution would be to update time-macros in c2rust/Cargo.toml.

anoraktrend:
Probably, too tired to try and implement it right now.

kkysen:
Superseded by #1115.

Issue #1097, opened 2024-06-07
analyze: proposal for simplified analysis and rewriting of allocations


spernsteiner:
Correct rewriting of `malloc`/`free` to `Box<T>` requires tracking ownership throughout the program.  This proposal describes an approach that produces correct rewrites with minimal new analysis work, at the cost of introducing run-time panics when the C code does not follow Rust ownership discipline.

The main idea is to rewrite owning pointers not to `Box<T>` but to `Option<Box<T>>`.  The `Option` is `Some` if ownership still resides at this location, and it is `None` if ownership has been transferred to some other location.  Accessing a location after ownership has been transferred away will cause a run-time panic due to unwrapping `None`.

## Definitions

A "pointer location" is a local variable or memory location of pointer type.  MIR temporaries representing the results of intermediate expressions are included in this definition.

A "heap object" is a region of memory allocated by `malloc` or a similar function.  A new object is "live" until it is freed by passing a pointer to the object to `free` or a similar function.

"Ownership" is a relation between objects and pointer locations.  At all points during execution, each live object is owned by one or more pointer locations.  (Single ownership of an object corresponds to `Box<T>`; multiple ownership corresponds to `Rc<T>`.) If a pointer location owns an object, then the value in that location is a pointer to the owned object.  In some situations, copying a value from a source pointer location to a destination may also transfer or copy ownership from the source to the destination.

## Rewrites

This section describes the new rewrites we would perform under this proposal.  The next section describes analysis changes that enable these rewrites.

**Type rewrites:**  Many pointer locations function as "borrowed references" and never hold ownership of an object in any execution of the program.  These are handled by existing rewrites.  For locations that may hold ownership, we add a new rewrite that replaces `*const T`/`*mut T` with `Option<Box<T>>`.

This approach to type rewrites means that for any pointer location that might hold ownership at some points in some executions, we assume that all assignments to that pointer location also grant ownership.  We call such a pointer location an "owning pointer location".  Note that an owning pointer location is not guaranteed to hold ownership throughout its lifetime; ownership may be transferred away during an assignment from the pointer location in question.  This is why we must rewrite to `Option<Box<T>>` rather than `Box<T>`.

**Expression rewrites:**  In every (static) assignment operation `q = p` on pointers, the LHS and RHS can each be either an owning pointer location or a non-owning pointer location.

[...]

PR #1106, opened 2024-07-22
analyze: implement Box<T> rewrites

spernsteiner:
This branch adds rewrites to convert pointers with the `FREE` permission into `Box<T>`.  This follows the dynamic ownership tracking proposal in #1097.

Specific changes:
* Adds a new `dyn_owned: bool` field to `TypeDesc`, which indicates that the pointer type should be wrapped in a dynamic ownership wrapper.
* Extends `ZeroizeType` handling to support zero-initializing some pointer types (necessary to allow `malloc`/`calloc` of structs that contain pointers).
* Adds support for `Box` and `dyn_owned` related casts in `mir_op` and `rewrite::expr::convert`.
* Adds custom rewrite rules for `malloc`, `calloc`, `free`, and `realloc`.

### DIFF
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -286,6 +286,38 @@ impl<'tcx> ConvertVisitor<'tcx> {
                 )
             }
 
+            mir_op::RewriteKind::MallocSafe {
+                ref zero_ty,
+                elem_size,
+                single,
+            } => {
+                // `malloc(n)` -> `Box::new(z)` or similar
+                assert!(matches!(hir_rw, Rewrite::Identity));
+                let zeroize_expr = generate_zeroize_expr(zero_ty);
+                let mut stmts = vec![
+                    Rewrite::Let(vec![("byte_len".into(), self.get_subexpr(ex, 0))]),
+                    Rewrite::Let1(
+                        "n".into(),
+                        Box::new(format_rewrite!("byte_len as usize / {elem_size}")),
+                    ),
+                ];
+                let expr = if single {
+                    stmts.push(Rewrite::Text("assert_eq!(n, 1)".into()));
+                    format_rewrite!("Box::new({})", zeroize_expr)
+                } else {
+                    stmts.push(Rewrite::Let1(
+                        "mut v".into(),
+                        Box::new(Rewrite::Text("Vec::with_capacity(n)".into())),
+                    ));
+                    stmts.push(format_rewrite!(
+                        "for i in 0..n {{\n    v.push({});\n}}",
+                        zeroize_expr,
+                    ));
+                    Rewrite::Text("v.into_boxed_slice()".into())
+                };
+                Rewrite::Block(stmts, Some(Box::new(expr)))
+            }
+
             mir_op::RewriteKind::CellGet => {
                 // `*x` to `Cell::get(x)`
                 assert!(matches!(hir_rw, Rewrite::Identity));
@@ -589,6 +621,27 @@ fn generate_zeroize_code(zero_ty: &ZeroizeType, lv: &str) -> String {
                 .unwrap();
             }
             writeln!(s, "}}").unwrap();
+            s
+        }
+    }
+}
+
+/// Generate an expression to produce a zeroized version of a value.
+fn generate_zeroize_expr(zero_ty: &ZeroizeType) -> String {
+    match *zero_ty {
+        ZeroizeType::Int => format!("0"),
+        ZeroizeType::Bool => format!("false"),
+        ZeroizeType::Array(ref elem_zero_ty) => format!(
+            "std::array::from_fn(|| {})",
+            generate_zeroize_expr(elem_zero_ty)
+        ),
+        ZeroizeType::Struct(ref name, ref fields) => {
+            let mut s = String::new();
+            write!(s, "{} {{\n", name).unwrap();
+            for (name, field_zero_ty) in fields {
+                write!(s, "{}: {},\n", name, generate_zeroize_expr(field_zero_ty),).unwrap();
+            }
+            write!(s, "}}\n").unwrap();
             s
         }
     }
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -576,7 +576,7 @@ fn generate_zeroize_code(zero_ty: &ZeroizeType, lv: &str) -> String {
         ",
             generate_zeroize_code(elem_zero_ty, "(*elem)")
         ),
-        ZeroizeType::Struct(ref fields) => {
+        ZeroizeType::Struct(_, ref fields) => {
             eprintln!("zeroize: {} fields on {lv}: {fields:?}", fields.len());
             let mut s = String::new();
             writeln!(s, "{{").unwrap();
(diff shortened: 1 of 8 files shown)


c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -286,6 +286,126 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                ref zero_ty,
+                elem_size,
+                single,
+            } |
+            mir_op::RewriteKind::CallocSafe {
+                ref zero_ty,
+                elem_size,
+                single,
+            } => {
+                // `malloc(n)` -> `Box::new(z)` or similar
```
ahomescu, line 299: Could you add an assertion here that the size of `zero_ty` is `elem_size`? Or can `zero_ty` be a different type?

spernsteiner: The sizes don't necessarily match.  `elem_size` is the size of the original, unrewritten type, which we use for converting `memcpy`/`memset` byte counts to a number of elements.  `zero_ty` indicates how to zero out the rewritten type, which might have a different size.  (Also, `ZeroizeTy` doesn't actually contain enough detail to allow computing the size.)

ahomescu: > `zero_ty` indicates how to zero out the rewritten type, which might have a different size.

Could that cause compiler errors?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -286,6 +286,119 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                    ],
+                    _ => unreachable!(),
+                };
+                let expr = if single {
+                    stmts.push(Rewrite::Text("assert_eq!(n, 1)".into()));
+                    format_rewrite!("Box::new({})", zeroize_expr)
+                } else {
+                    stmts.push(Rewrite::Let1(
+                        "mut v".into(),
+                        Box::new(Rewrite::Text("Vec::with_capacity(n)".into())),
```
ahomescu, line 326: Could this just be `vec![{zeroize_expr}; n]`?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -566,7 +679,7 @@ fn generate_zeroize_code(zero_ty: &ZeroizeType, lv: &str) -> String {
     match *zero_ty {
         ZeroizeType::Int => format!("{lv} = 0"),
         ZeroizeType::Bool => format!("{lv} = false"),
-        ZeroizeType::Iterable(ref elem_zero_ty) => format!(
```
ahomescu, line 569: Why the change here from `Iterable` to `Array`? I'm assuming `Iterable` is a superset of the other, so what gets left out?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -663,6 +797,33 @@ pub fn convert_cast_rewrite(kind: &mir_op::RewriteKind, hir_rw: Rewrite) -> Rewr
...
+            Rewrite::Call(
+                "std::mem::replace".to_string(),
+                vec![
+                    Rewrite::Ref(Box::new(hir_rw), hir::Mutability::Mut),
+                    Rewrite::Text("Err(())".into()),
+                ],
+            )
+        }
+        mir_op::RewriteKind::DynOwnedWrap => {
+            Rewrite::Call("std::result::Result::<_, ()>::Ok".to_string(), vec![hir_rw])
```
ahomescu, line 814: Couldn't this just be an `Option`?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -286,6 +286,119 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                        ));
+                        stmts.push(format_rewrite!(
+                            "dest_ptr.resize_with(dest_n, || {})",
+                            zeroize_expr,
+                        ));
+                        Rewrite::Text("dest_ptr.into_boxed_slice()".into())
+                    }
+                    (false, true) => {
+                        format_rewrite!(
+                            "src_ptr.into_iter().next().unwrap_or_else(|| {})",
```
ahomescu, line 379: Couldn't we use `unwrap_or_default()`?

c2rust-analyze/src/rewrite/expr/convert.rs
ahomescu: Answered my own question: it's just a rename.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -594,6 +707,27 @@ fn generate_zeroize_code(zero_ty: &ZeroizeType, lv: &str) -> String {
     }
 }
 
+/// Generate an expression to produce a zeroized version of a value.
+fn generate_zeroize_expr(zero_ty: &ZeroizeType) -> String {
```
ahomescu, line 711: Couldn't you use `Default::default` here? Is it because the trait isn't implemented for generic-sized arrays?

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: `vec![x; n]` requires `x` to be `Clone`, so it can't contain e.g. `Option<&mut T>`

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: It was previously called `Iterable` because it could in principle be used for anything with an `iter_mut()` method.  It's now called `Array` because it can only be used for arrays specifically - the `generate_zeroize_expr` function used for initializing allocations can't be generic as the code we use for zeroizing an existing value on `memset(_, 0, _)`.

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: It could, but `DynOwned` can be composed with `Option<T>` if we encounter a nullable owned pointer, and `Option<DynOwned<T>>` is clearer than `Option<Option<T>>`.  But c2rust-analyze doesn't yet have a run-time support library where we could define `DynOwned<T>`, so for now we use the somewhat ugly `Result<T, ()>` in place of `DynOwned<T>`.

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: It's also not implemented for c2rust-transpiled structs

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: No, but there's a related issue of potentially using the wrong size when converting the byte count passed to `malloc`/`memcpy`/`memset` to an element count.  I have a fix for this (at least for the most common cases) in a later branch: fd24d8158dbd7e4f0900661ae8a6afe4bc4bbc38

spernsteiner:
Looks like the Darwin CI error here is the same issue that's being fixed in #1146.  Once that's merged, I'll rebase this and try the CI again.

thedataking:
> Looks like the Darwin CI error here is the same issue that's being fixed in #1146. Once that's merged, I'll rebase this and try the CI again.

You are right. #1146 is merged now.

c2rust-analyze/src/rewrite/expr/convert.rs
ahomescu: Might be useful to add a comment.

ahomescu (approved):

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Added a comment on the translation of `DynOwned` types: 95d40382e5f9f71f4010f7ed211239fc585d6343

PR #1111, opened 2024-08-06
analyze: borrowck performance improvements

spernsteiner:
I found some easy borrowck/Polonius performance improvements while investigating a performance regression on a development branch.

1. Use `BufReader`/`BufWriter` when loading/saving from the `polonius_cache`.  Otherwise `bincode` reads/writes each 8-byte field individually.
2. Don't dump Polonius facts to `inspect/FUNC/*.facts` unless requested with `C2RUST_ANALYZE_DUMP_POLONIUS_FACTS=1`.  This takes a nontrivial amount of time for big functions and is only needed when debugging certain borrowck issues.

Improvements compared to master (all measurements taken on the second run, after populating `polonius_cache/`):
* `cargo test algo_md5`: 25s -> 4.6s
* `cargo test --release algo_md5`: 16s -> 2.1s

### DIFF
--- a/c2rust-analyze/src/borrowck/mod.rs
+++ b/c2rust-analyze/src/borrowck/mod.rs
@@ -19,6 +19,7 @@ use std::collections::HashMap;
 use std::fmt::{Debug, Formatter, Write as _};
 use std::fs::{self, File};
 use std::hash::{Hash, Hasher};
+use std::io::{BufReader, BufWriter};
 
 mod atoms;
 mod def_use;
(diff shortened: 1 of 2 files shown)


fw-immunant (approved):

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -383,7 +384,7 @@ fn run_polonius<'tcx>(
 fn try_load_cached_output(facts_hash: &str) -> Option<Output> {
     let path = format!("polonius_cache/{}.output", facts_hash);
 
-    let f = File::open(&path).ok()?;
+    let f = BufReader::new(File::open(&path).ok()?);
```
kkysen, line 387: Is it faster to just read it all at once?  Sometimes I've found that to be faster.

PR #1113, opened 2024-08-28
Make scripts/pdg.sh more flexible for more use cases

ahomescu:


kkysen (commented):
What is this trying to accomplish exactly?

The intention was to make the script obsolete over time, as the functionality is also available `pdg/src/main.rs`'s `mod tests`.

Issue #1098, opened 2024-06-19
Fails to compile on latest nightly


theoparis:
See https://github.com/time-rs/time/issues/681, it seems like some dependencies need to be updated since it isn't used directly...

Adding the following to Cargo.toml in this repository seems to fix it for the time being.
```
[patch.crates-io]
time = { git = "https://github.com/time-rs/time" }
```

Update: I think the correct fix would actually be updating Cargo.lock and the time-macros crate.

tommythorn:
With the risk of stating the obvious, this is also causing `cargo install c2rust` to fail.

chrysn:
Note that c2rust compiled just fine also on stable Rust. So as theoparis mentioned in the update, unlocking time-rs-macros sounds like the way to go. We need time 0.3.35 to fix the build issue, which needs Rust 1.67, which is more recent than the currently locked nightly.

kkysen:
* Fixed by #1115.
* Will be released after #1116.

PR #1114, opened 2024-08-28
Update Rust nightly version

chrysn:
Closing https://github.com/immunant/c2rust/issues/1098 would be easiest if we didn't have to cater for the very old nightly. Let's try updating that first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.19.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


kkysen:
Hi! We have an old PR #811 that updated to 1.67, though it's old and needs rebasing. If you'd rather, you could rebase that. That said, I'm gonna fix the issue by just removing `git-testament` for now, which removes the `time` dependency entirely.  It's nice to have, so if we update the nightly later, we'll add it back, but that should be pretty simple.

kkysen:
I'm in the middle of it and I'll put up a PR for it shortly.  If you can test it extensively, that'd be great!  (Though I don't anticipate problems with `time` gone).  Then we'll make a new release right after.

PR #1115, opened 2024-08-28
transpile: remove `git-testament` dependency (for now) to remove `time` dependency to build on 1.80.0

### LINKED ISSUE #1109: rust 1.80.0 build issue
at least time crate needs to be updated to 0.3.36

see https://github.com/time-rs/time/issues/693

kkysen:
* Fixes #1109.

Our old nightly (1.65) pins `time-macros` to an old version, which requires `time` to stay on an old version. But since 1.80 broke `time` (due to https://github.com/time-rs/time/issues/693 and https://github.com/rust-lang/rust/issues/127343), we need to update `time`.  Doing this requires `time-macros` to be upgraded, which requires at least rust 1.67. There's an old PR, #811, for this, but it needs rebasing. So for now, removing `git-testament` and its `time` dependency is simpler, and we can add it back later once we update to a newer nightly.

With this fix, both `cargo build -p c2rust` (the pinned nightly) and `cargo +stable build -p c2rust` (1.80.1) now work.

If this is good, and once we merge this, we'll release a new version with the fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -181,12 +181,10 @@ dependencies = [
  "c2rust-transpile",
  "clap 3.2.25",
  "env_logger",
- "git-testament",
  "is_executable",
  "log",
  "regex",
  "shlex",
- "time-macros",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


kkysen:
@chrysn, feel free to review and/or test this.  Hopefully it works for everything you need.

thedataking (approved):

fw-immunant (approved):
LGTM, thanks for untangling these dep issues.

PR #1116, opened 2024-08-28
Release 0.19.0

kkysen:
This
* updates our version to `0.19.0`
* fixes `clippy` warnings
* fixes warnings on `stable` (`1.80.1`), including setting `resolver = "2"`
* sets `rust-version = "1.65"`, our pinned nightly

However, I'm unable to run `cargo update` since the resolver is still MSRV-unaware IIUC.  `cargo outdated` also fails:

```shell
> cargo outdated -R
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `/home/kkysen/.rustup/toolchains/nightly-2022-08-08-x86_64-unknown-linux-gnu/bin/rustc - --crate-name ___ --print=file-names -C link-arg=-fuse-ld=/home/kkysen/work/c++/mold/current/bin/mold --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=split-debuginfo --print=crate-name --print=cfg` (exit status: 1)
  --- stderr
  error: unknown print request `split-debuginfo`
```

Otherwise, this should be everything according to https://github.com/immunant/c2rust/wiki/Release-Process that needs to be checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ marks.*.json
 inspect/
 *.analysis.txt
 analysis.txt
+polonius_cache/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,7 +174,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "anyhow",
  "c2rust-build-paths",
@@ -189,7 +189,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analysis-rt"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "bincode",
  "crossbeam-queue",
@@ -202,7 +202,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analyze"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "anyhow",
  "assert_matches",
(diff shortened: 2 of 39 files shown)


fw-immunant (approved):
LGTM.

chrysn:
Thanks, works great: I've given this the testing treatment our RIOT OS build system is getting (which is not a full build-all test but usually representative enough), and all works fine. I'd love to update the docs to install the latest released version rather than an untagged git commit.

Is there anything that can be done to advance this to a crates.io release?

kkysen:
> Is there anything that can be done to advance this to a crates.io release?

We hit some snags on that, because there's a cargo bug where the workspace `Cargo.lock` is not considered when publishing. It's fixed in the next cargo version, but of course we're pinned, so we'd need to update that.  Or it might be possible to publish with a `Cargo.lock` that doesn't work on the pinned nightly, as everything published is fine on stable (would this cause anyone problems?).

Is the GitHub release enough for now?  We're trying to publish the rest of the crates (most of the smaller dependency ones have been already), we've just been busy with other stuff, and will try to get it done soon.

chrysn:
My impression on this is that anyone who uses the crate through crates.io will not get the nightly-version pinning anyway, but whatever they have installed. People might in theory install it through `cargo +nightly-2022-... install c2rust`, but they can not reliably find out which nightly they should use without cloning the repository in the first place (it's not discoverable through the released version AFAICT), and when they've cloned it, they don't even have to look it up any more but can just install from the checkout that has both the right Cargo.lock and the rust-toolchain file.

PR #1121, opened 2024-09-05
transpile: parenthesize subexpression of reference expressions if necessary

### LINKED ISSUE #1084: Missing parentheses on reference applied to a cast
```c
int main(void) {
    int * x = &(int){0};
}
```

generates

```rust
    let mut x: *mut libc::c_int = &mut 0 as libc::c_int as *mut libc::c_int;
    return 0;
}
pub fn main() {
}
```

but extra parentheses are needed

```rust
    let mut x: *mut libc::c_int = &mut (0 as libc::c_int) as *mut libc::c_int;
```

This can happen in more realistic code when an output function parameter is being ignored `f(&(int){0});`

If someone reads this issue and wonders if this is even legal C, check out [compound literal](https://en.cppreference.com/w/c/language/compound_literal)

> The value category of a compound literal is [lvalue](https://en.cppreference.com/w/c/language/value_category) (its address can be taken). ...

fw-immunant:
* Fixes #1084.

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -985,13 +985,13 @@ impl Builder {
     }
 
     pub fn addr_of_expr(self, e: Box<Expr>) -> Box<Expr> {
-        Box::new(Expr::Reference(ExprReference {
+        Box::new(parenthesize_if_necessary(Expr::Reference(ExprReference {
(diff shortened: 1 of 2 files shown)


kkysen (commented):
Can we add the test from #1084?

kkysen (approved):

PR #1129, opened 2024-09-18
`transpile`: Add SIMD intrinsic `__builtin_ia32_vperm2f128_pd256/_mm256_permute2f128_pd`

Kriskras99:


fw-immunant:
Thanks for the contribution!

This looks sane, but it would be nice to have a bit of background: is there an example .c file that can be used to validate this, and should we add support for other AVX intrinsics at the same time?

Kriskras99:
I ran into this when transpiling soxr. This was the only intrinsic missing for that code base.

I can extract an example usage from that code base to use as a test

Kriskras99:
Also seems the Github Action needs to be updated

kkysen:
> I can extract an example usage from that code base to use as a test

Yeah, a simple test for this would be nice.  Thanks!

PR #1132, opened 2024-09-21
Translate `builtin_assume` and unaligned SIMD intrinsics

thedataking:
Unaligned vector types such as `__m128_u` or `__256_u` not not have any equivalents in Rust so we translate them to their aligned equivalents. This is unsound so we emit warning that the program may be mistranspiled. These fixes are required to transpile the `zstd` compressor on modern x86 hosts.


c2rust-transpile/src/translator/builtins.rs
```diff
@@ -613,6 +613,9 @@ impl<'c> Translation<'c> {
                     )
                 })
             }
+
+            // TODO: provide proper implementation, this is just to get stuff building
+            "__builtin_assume" =>  Ok(self.convert_expr(ctx.used(), args[0])?),
```
kkysen, line 618: When you do implement this, note that Rust (as of 1.81.0) now has `core::hint::assert_unchecked`, which should be the equivalent of `__builtin_assume`.

kkysen (commented):
How come overaligning isn't allowed in all cases?  Padding differences?

kkysen (commented):
There does seem to be some unaligned stuff in `stdarch`. 
 For example, [`_mm_storeu_si128`](https://github.com/rust-lang/stdarch/blob/1b3d2c0c8a652644559bba8176d6af5c1a28e04d/crates/core_arch/src/x86/sse2.rs#L1279).

c2rust-transpile/src/translator/builtins.rs
thedataking: @kkysen do you want to add support for `__builtin_assume` into a separate PR with a test? 

It seems my PR is trying to do two things and only one is blocked on what rust provides, so that's messy. 

WDYT about landing the rest of this PR? Should the warning reference the PR folkert opened?

kkysen: > do you want to add support for `__builtin_assume` into a separate PR with a test?

We're still stuck on the old nightly, so adding support for newer features like this is tricky.  That said, we could merge #1227, which would use `stable` for `c2rust-transpile` and let us start targeting newer Rust versions for just the transpiler.  #1227 didn't work before, but now that CI is on GitHub Actions, it works without much modifications, so it wouldn't be too much work to review and merge that.  If we do that, I can add support for it.

c2rust-transpile/src/translator/simd.rs
```diff
@@ -100,6 +102,21 @@ impl<'c> Translation<'c> {
                     ).into());
                 }
 
+                // Drop the `_u` suffix for unaligned SIMD types as they have
+                // no equivalents in Rust. Warn the user about possible probems.
```
kkysen, line 106: Can we use this for unaligned SIMD types?

[Godbolt](https://godbolt.org/z/PhY635j4d)

```rust
use core::arch::x86_64::__m256;
use core::mem;

#[unsafe(no_mangle)]
pub fn aligned() -> usize {
    mem::align_of::<__m256>()
}

#[repr(packed)]
pub struct UnalignedM256(pub __m256);

#[unsafe(no_mangle)]
pub fn unaligned() -> usize {
    mem::align_of::<UnalignedM256>()
}
```

`#[repr(packed)]` lets you override the alignment and unalign a type.

PR #1134, opened 2024-09-21
transpile: Make cast from `bool` to pointer compile through `size_t`

### LINKED ISSUE #1077: Casting bool to pointer type results in invalid Rust
The fix should be the same as #1030, but presumably with something like `size_t` as the intermediate type.

(Why, you might ask, would anyone ever do this in real code. The answer is that C doesn't have generics, so projects will have, say, a single hash table type with `void *` values, and store booleans in it by casting.)

GPHemsley:
* Similar to #1030.

* Fixes #1077.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4282,6 +4282,13 @@ impl<'c> Translation<'c> {
                             target_ty,
                         )))
                     })
+                } else if target_ty_ctype.is_pointer() && source_ty_kind.is_bool() {
+                    val.and_then(|x| {
+                        Ok(WithStmts::new_val(mk().cast_expr(
+                            mk().cast_expr(x, mk().path_ty(vec!["libc", "size_t"])),
+                            target_ty,
+                        )))
+                    })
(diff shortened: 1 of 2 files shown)


kkysen (approved):
LGTM.  Same exact thing as #1030 except it's through `size_t`.

One thing that we should keep in mind is that `#![feature(strict_provenance)]` will probably be stabilized soon (https://github.com/rust-lang/rust/pull/130350), and at that point, it'd probably be better to be more specific about this.


In #1077, @jorendorff said this was the reason why this is done:

> (Why, you might ask, would anyone ever do this in real code. The answer is that C doesn't have generics, so projects will have, say, a single hash table type with `void *` values, and store booleans in it by casting.)

This seems likely, as a provenance-preserving pointer cast through a `bool` does not make any sense, as `false`/`0` is null and `true`/`1` should almost always be an invalid address as well.  Thus, we should be able to unconditionally use [`ptr::without_provenance`](https://doc.rust-lang.org/std/ptr/fn.without_provenance.html) for this cast once `strict_provenance` is stabilized.

PR #1135, opened 2024-09-21
transpile: Make volatile post-increment assignment compile

### LINKED ISSUE #1064: In the while loop body using a volatile variable, the generated code will lack a semicolon.
ubuntu2310 + llvm-config-15 + C2Rust 0.18.0

input:
```c
#include <stdio.h>

static void Encode(char *str_)
{
  const char *volatile str = str_;
  static const char *HEX = "0123456789ABCDEF";
  while (*str)
  {
    int c = (unsigned char)*str++;
  }
}

int main()
{
  Encode(NULL);
  return 0;
}


```

output:
```rs
unsafe extern "C" fn Encode(mut str_: *mut libc::c_char) {
    let mut str: *const libc::c_char = str_;
    static mut HEX: *const libc::c_char = b"0123456789ABCDEF\0" as *const u8
        as *const libc::c_char;
    while *str != 0 {
        let fresh0 = ::core::ptr::read_volatile::<
            *const libc::c_char,
        >(&str as *const *const libc::c_char);
        ::core::ptr::write_volatile(
            &mut str as *mut *const libc::c_char,
            (::core::ptr::read_volatile::<
                *const libc::c_char,
            >(&str as *const *const libc::c_char))
                .offset(1),
        )
        let mut c: libc::c_int = *fresh0 as libc::c_uchar as libc::c_int;
    }
}
    Encode(0 as *mut libc::c_char);
    return 0 as libc::c_int;
}
pub fn main() {
}

```

GPHemsley:
Fixes #1064.


tests/ints/src/volatile.c
```diff
@@ -44,4 +44,9 @@ void entry3(const unsigned buffer_size, int buffer[])
     buffer[8] = s.buffer[3];
 }
 
-
+void volatile_stuff(void)
+{
+    volatile int x1 = 0;
+    int x2 = x1++;
+    x2;
+}
```
kkysen, line 52: I know you said #1064 doesn't have anything to do with it being a loop, and just being `volatile`, but could we just include the loop version here (in addition) anyways, just erring on the more comprehensive testing side?

kkysen (commented):
Thanks for fixing this!  I'm still trying to understand what exactly is happening here and why the code was fine before for non-volatile writes, as I'm not too familiar with this part of the code.  So still reviewing...

tests/ints/src/volatile.c
GPHemsley: Done.

GPHemsley:
> Thanks for fixing this! I'm still trying to understand what exactly is happening here and why the code was fine before for non-volatile writes, as I'm not too familiar with this part of the code. So still reviewing...

I also don't understand, but I've also added a test case to prove that it was indeed working before.

GPHemsley:
I think it's simply rare that volatile writes like this occur, and so this scenario was not commonly hit.

There isn't a comprehensive test suite for the operator code, so edge cases are not always being tested.

GPHemsley:
As for why `assign_expr()` includes a semicolon when passed through `expr_stmt()` and `call_expr()` doesn't, I suppose I haven't tracked that down yet.

```rust
    pub fn assign_expr(self, lhs: Box<Expr>, rhs: Box<Expr>) -> Box<Expr> {
        Box::new(Expr::Assign(ExprAssign {
            attrs: self.attrs,
            eq_token: Token![=](self.span),
            left: lhs,
            right: rhs,
        }))
    }
```

```rust
    pub fn call_expr(self, func: Box<Expr>, args: Vec<Box<Expr>>) -> Box<Expr> {
        let args = args.into_iter().map(|a| *a).collect();
        Box::new(parenthesize_if_necessary(Expr::Call(ExprCall {
            attrs: self.attrs,
            paren_token: token::Paren(self.span),
            func,
            args,
        })))
    }
```

```rust
    pub fn expr_stmt(self, expr: Box<Expr>) -> Stmt {
        Stmt::Expr(*expr)
    }

    pub fn semi_stmt(self, expr: Box<Expr>) -> Stmt {
        Stmt::Semi(*expr, Token![;](self.span))
    }
```

```rust
    #[cfg_attr(doc_cfg, doc(cfg(feature = "printing")))]
    impl ToTokens for Stmt {
        fn to_tokens(&self, tokens: &mut TokenStream) {
            match self {
                Stmt::Local(local) => local.to_tokens(tokens),
                Stmt::Item(item) => item.to_tokens(tokens),
                Stmt::Expr(expr) => expr.to_tokens(tokens),
                Stmt::Semi(expr, semi) => {
                    expr.to_tokens(tokens);
                    semi.to_tokens(tokens);
                }
            }
        }
    }
```

GPHemsley:
If it helps, this code:
```C
	int i = 0;
	i++;
	int i2 = i++;
	int i3 = i--;
	i--;

	volatile int vi = 0;
	vi++;
	volatile int vi2 = vi++;
	volatile int vi3 = vi--;
	vi--;
```

generates this code:

```rust
    let mut i: libc::c_int = 0 as libc::c_int;
    i += 1;
    i;
    let fresh0 = i;
    i = i + 1;
    let mut i2: libc::c_int = fresh0;
    let fresh1 = i;
    i = i - 1;
    let mut i3: libc::c_int = fresh1;
    i -= 1;
    i;
    let mut vi: libc::c_int = 0 as libc::c_int;
    ::core::ptr::write_volatile(
        &mut vi as *mut libc::c_int,
        ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int) + 1,
    );
    ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int);
    let fresh2 = ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int);
    ::core::ptr::write_volatile(
        &mut vi as *mut libc::c_int,
        ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int) + 1,
    )
    let mut vi2: libc::c_int = fresh2;
    let fresh3 = ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int);
    ::core::ptr::write_volatile(
        &mut vi as *mut libc::c_int,
        ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int) - 1,
    )
    let mut vi3: libc::c_int = fresh3;
    ::core::ptr::write_volatile(
        &mut vi as *mut libc::c_int,
        ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int) - 1,
    );
    ::core::ptr::read_volatile::<libc::c_int>(&vi as *const libc::c_int);
```

`convert_post_increment()` is the only time a call to `volatile_write()` ends up in the `stmts` part of `WithStmts`; other calls have it end up in `val`, which is why the other lines don't have this problem.

PR #1137, opened 2024-09-22
`rust-analyzer`: add `rust-analyzer` component to `rust-toolchain.toml` so that it is in sync with our pinned nightly

kkysen:
The latest `rust-analyzer` only supports `stable` and now breaks, so this pins it to the version on our `nightly`, which won't break on that version.


fw-immunant (approved):
LGTM.

PR #1146, opened 2024-10-22
Fix Darwin CI

fw-immunant:
Darwin CI has been failing on all recent PRs; hopefully this fixes it.


fw-immunant:
This fixes CI; it also adds explicit casts, so if the purpose of this test coverage was to handle *implicit* casts then we should instead take the approach of adding `#pragma clang diagnostic ignore "-Wincompatible-function-pointer-types"` and `#pragma clang diagnostic ignore "-Wint-conversion"` to the appropriate files.

thedataking (approved):

PR #1147, opened 2024-10-22
analyze: add HEAP and STACK permissions

spernsteiner:
This adds new pointer permissions that track whether a pointer may point to the heap or to the stack.  (Statics are counted as stack, since like stack variables, we can't use `Box<T>` for pointers to them.)

As usual for forward-dataflow permissions in our framework, the definitions are a little tricky:
* `HEAP` means "points to the heap, or is null" - in other words, this pointer definitely doesn't point to the stack.
* `STACK` means "points to the stack, or is null" - in other words, this pointer definitely doesn't point to the heap.
* `HEAP | STACK` means the pointer must be null, since null is the only pointer in the intersection of those two sets.  In other words, the pointer points neither to the heap nor to the stack.
* Having neither flag means the pointer could point anywhere.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -138,6 +138,20 @@ bitflags! {
         /// [`_.is_null()`]: core::ptr::is_null
         /// [`_.is_some()`]: Option::is_some
         const NON_NULL = 0x0080;
+
+        /// This pointer points to the heap (or is null).
+        ///
+        /// `HEAP` and `STACK` form a four-element lattice.  `HEAP` means definitely heap, `STACK`
+        /// means definitely stack, neither means it could be either (top), and both means the
+        /// pointer is either definitely null or comes from an unknown source (bottom).
+        const HEAP = 0x0100;
+
+        /// This pointer points to the stack or to static memory (or is null).
+        ///
+        /// Currently we distinguish stack/static vs heap, but don't distinguish stack vs static.
+        /// The reason is that heap pointers can be rewritten to `Box<T>`, but stack and static
+        /// pointers both cannot.
+        const STACK = 0x0200;
     }
 }
 
@@ -167,7 +181,11 @@ impl PermissionSet {
     /// The permissions for a (byte-)string literal.
     //
     // `union_all` is used here since it's a `const fn`, unlike `BitOr::bitor`.
-    pub const STRING_LITERAL: Self = Self::union_all([Self::READ, Self::OFFSET_ADD]);
+    pub const STRING_LITERAL: Self = Self::union_all([Self::READ, Self::OFFSET_ADD, Self::STACK]);
+
+    /// Negative permissions for a (byte-)string literal.  These permissions should be absent from
+    /// all string literals, contrary to the defaults for most pointers.
+    pub const STRING_LITERAL_NEGATIVE: Self = Self::union_all([Self::HEAP]);
 }
 
 bitflags! {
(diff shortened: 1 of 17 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -138,6 +138,20 @@ bitflags! {
...
+        /// means definitely stack, neither means it could be either (top), and both means the
+        /// pointer is either definitely null or comes from an unknown source (bottom).
+        const HEAP = 0x0100;
+
+        /// This pointer points to the stack or to static memory (or is null).
+        ///
+        /// Currently we distinguish stack/static vs heap, but don't distinguish stack vs static.
+        /// The reason is that heap pointers can be rewritten to `Box<T>`, but stack and static
+        /// pointers both cannot.
+        const STACK = 0x0200;
```
ahomescu, line 154: I don't like the ambiguous use of `STACK` here, but I'm having a hard time coming up with an alternative. Maybe `STORAGE`?  Or rename `HEAP`/`STACK` to `DYNAMIC`/`STATIC`? Another option: `ALLOC`/`NO_ALLOC`.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -723,8 +732,9 @@ pub fn visit<'tcx>(
         equiv_constraints: Vec::new(),
     };
 
-    for (ptr, perms) in acx.string_literal_perms() {
+    for (ptr, perms, neg_perms) in acx.string_literal_perms() {
         tc.constraints.add_all_perms(ptr, perms);
+        tc.constraints.add_no_perms(ptr, neg_perms);
```
ahomescu, line 737: There might be a subtle answer I'm missing: what is the difference between "remove permissions", "add no permissions", and "add forbidden permissions"?

ahomescu (approved):

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: The various `add_foo` methods each add a `foo` constraint to the constraint set.  `add_no_perms(ptr, perms)` adds `Constraint::NoPerms(ptr, perms)`, which requires that `ptr` has none of the permissions listed in `perms`.

Maybe "emit" would be clearer than "add" for these methods?

ahomescu: `emit` usually makes me think of code generation, so it's not much better.

I think the distinction between "remove permissions" and "add no permissions" is clearer, but is there one between "add no permissions" and "updates forbidden"?

spernsteiner: `Constraint::NoPerms` is generated when we encounter code that implies that a pointer must not have a certain permission.  For example, on `p = ptr::null()`, we emit (approximately) `Constraint::NoPerms(p, NON_NULL)` - `p` must not have the `NON_NULL` permission.  This is mainly relevant for permissions that are set by default on all pointers and are removed only when required by specific constraints (`NON_NULL` is one such permission).

"Updates forbidden" is not a constraint; instead, it causes the initial value of a permission to be preserved, ignoring any constraints that would require otherwise.  For example, if we have `Constraint::NoPerms(p, NON_NULL)`, but `updates_forbidden[p]` contains `NON_NULL`.  Normally, processing that constraint during `DataflowConstraints::propagate` would remove the `NON_NULL` permission from `p`.  But because `NON_NULL` is listed in `updates_forbidden[p]`, the change to `p`'s `NON_NULL` flag is suppressed.

Issue #1103, opened 2024-07-10
analyze: tracking issue for simple_buffer


spernsteiner:
`simple_buffer` is a simplified buffer implementation (similar to `Vec<u8>`) designed to mimic lighttpd's `buffer` module.

* [ ] borrowck panic: Missing origin params on `Rvalue::Cast` output.  This is triggered by `malloc(...) as *mut buffer`.  `buffer` has one origin param for its internal data pointer, but the `borrowck::LTy` computed for the cast has no origin params.
* [ ] unlower panic: Missing entry for `Rvalue::Ref` arising from `let ref mut fresh = ...;`.  
* [ ] borrowck: `return malloc(...)` removes `UNIQUE` from the output pointer because Polonius reports an error.  
  * Needs more investigation to understand the precise cause of this error
* [ ] borrowck: Drops `UNIQUE` due to overlapping borrows around `realloc`:
  ```C
  buf->data = realloc(buf->data, ...);
  ```
  ```Rust
  let old = (*buf).data;
  let fresh = &mut (*buf).data;
  *fresh = realloc(old, ...);
  ```
* [ ] borrowck: Drops `UNIQUE` due to two-phase-like behavior in `buffer_realloc(buf, (*buf).cap * 2)`.
  * [ ] Check whether lighttpd buffer actually uses this pattern
* [ ] mir_op: `malloc`/`realloc` calls aren't recognized as rewritable via `SafeMalloc`/`SafeRealloc`.
  * Needs more investigation
* [ ] `derive(Copy)` is left on `struct buffer` even after rewriting the `data` pointer to use `Box`.
* [ ] rewrite: Many bad interactions between `Option` (`NULL`) and `Result` (`DynOwned`/`FREE`) rewrites.
  * [ ] Develop test cases for individual issues

spernsteiner:
We can now analyze, rewrite, compile, and run `simple_buffer` after patching it with workarounds for various `c2rust-analyze` issues.  As we fix the issues tracked here, we can remove the workarounds until `simple_buffer` works unmodified.

spernsteiner:
I've implemented fully interprocedural pointee type analysis (on a dev branch - no PR yet), and it now produces correct pointee types for the `data` field and for all `malloc`/`free`/`realloc` calls.  Next I'm planning to look at the borrowck `Rvalue::Cast` issue.

PR #1148, opened 2024-10-22
analyze: misc fixes for simple_buffer

spernsteiner:
This branch has fixes for various minor issues uncovered while testing on `simple_buffer.c` (#1103).

* Fix borrowck handling of casts to `*mut my_struct` when `my_struct` has lifetimes
* Fix various cases of compile errors introduced by `DynOwned` rewrites
* Fix precedence of method calls when generating rewritten source code
* Treat `realloc(NULL, n)` like `malloc(n)`
* Support nullable pointers in the `ZeroizeTy` initializers used for `malloc`/`memset` rewrites

### DIFF
--- a/c2rust-analyze/src/rewrite/apply.rs
+++ b/c2rust-analyze/src/rewrite/apply.rs
@@ -361,7 +361,7 @@ impl<S: Sink> Emitter<'_, S> {
                 })
             }
             Rewrite::MethodCall(ref method, ref receiver_rw, ref arg_rws) => {
-                self.emit(receiver_rw, 0)?;
+                self.emit(receiver_rw, 3)?;
                 self.emit_str(".")?;
                 self.emit_str(method)?;
                 self.emit_parenthesized(true, |slf| {
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -758,6 +758,11 @@ pub fn convert_cast_rewrite(kind: &mir_op::RewriteKind, hir_rw: Rewrite) -> Rewr
             Rewrite::Ref(Box::new(place), mutbl_from_bool(mutbl))
         }
 
+        mir_op::RewriteKind::Deref => {
+            // `p` -> `*p`
+            Rewrite::Deref(Box::new(hir_rw))
+        }
+
         mir_op::RewriteKind::OptionUnwrap => {
             // `p` -> `p.unwrap()`
             Rewrite::MethodCall("unwrap".to_string(), Box::new(hir_rw), vec![])
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -807,7 +807,10 @@ pub fn convert_cast_rewrite(kind: &mir_op::RewriteKind, hir_rw: Rewrite) -> Rewr
             // `p` -> `mem::replace(&mut p, Err(()))`
             Rewrite::Call(
                 "std::mem::replace".to_string(),
-                vec![hir_rw, Rewrite::Text("Err(())".into())],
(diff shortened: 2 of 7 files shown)


c2rust-analyze/src/rewrite/apply.rs
```diff
@@ -361,7 +361,7 @@ impl<S: Sink> Emitter<'_, S> {
                 })
             }
             Rewrite::MethodCall(ref method, ref receiver_rw, ref arg_rws) => {
-                self.emit(receiver_rw, 0)?;
+                self.emit(receiver_rw, 3)?;
```
ahomescu, line 364: What's the magic 3 value here?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -364,9 +365,14 @@ impl<'tcx> ConvertVisitor<'tcx> {
                 }
                 let expr = match (src_single, dest_single) {
                     (false, false) => {
+                        let src = if option {
+                            "src_ptr.unwrap_or(Box::new([]))"
```
ahomescu, line 369: Could this just be `src_ptr.unwrap_or_default()`? Or use `unwrap_or_else` so you're not constructing a Box every time?

c2rust-analyze/tests/filecheck/rewrite_nullable_box.rs
```diff
@@ -0,0 +1,35 @@
...
+
+extern "C" {
+    fn malloc(_: libc::c_ulong) -> *mut libc::c_void;
+    fn free(__ptr: *mut libc::c_void);
+}
+
+#[c2rust_analyze_test::skip_borrowck]
+// CHECK-LABEL: fn test_is_null(
+pub unsafe fn test_is_null(cond: bool) {
+    // CHECK: let mut ptr: core::option::Option<core::result::Result<std::boxed::Box<(i32)>,()>> =
```
ahomescu, line 17: Why is it `Box<(i32)>`? Is that inner tuple needed? (or is it something else)

ahomescu (approved):

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Operator precedence, roughly.  Things that bind less tightly than `.` need to be parenthesized when used as a method receiver.  E.g. `(2 + 2).method()`

ahomescu: Might be useful to add a comment for it, and/or change it to a `const` with a good name.

c2rust-analyze/tests/filecheck/rewrite_nullable_box.rs
spernsteiner: That's not a tuple (1-tuples require a trailing comma, like `(i32,)`).  When reusing part of a type annotation as written in the original source code (in this case, the `i32` from `*mut i32`), we wrap the reused type in parentheses so we don't have to worry about precedence or other interactions with the surrounding rewrite.  In theory, these parentheses should be easy to detect and remove with an "unused parens" lint in the cases where they're not needed.

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Zero-sized boxes don't allocate - that includes boxed zero-length arrays, as in this case.  As for `unwrap_or_default`, I generally prefer `.unwrap_or_else(Vec::new)` etc. to make it clear what the result type is intended to be.

c2rust-analyze/src/rewrite/apply.rs
spernsteiner: Changing all the precedence values to named `const`s might be a good idea.  I think that should be a separate PR, though

PR #1151, opened 2024-10-28
analyze: static/dynamic comparison mode for NON_NULL

spernsteiner:
This adds a mode for comparing static analysis results to the dynamic results from the PDG.  Comparison mode is enabled by setting the env var `C2RUST_ANALYZE_COMPARE_PDG=1`.  When this mode is enabled, c2rust-analyze skips normal rewriting and instead prints a copy of the input code with annotation comments indicating places where static and dynamic NON_NULL results disagree.  Specifically, it annotates cases where either the static analysis computed NON_NULL but the PDG includes at least one observation where the pointer is null, or the dynamic analysis marked the pointer nullable but the PDG contains observations that are all non-null.

In support of the main change, this branch also includes a refactor of the PDG processing code.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -1,7 +1,7 @@
 use crate::annotate::AnnotationBuffer;
 use crate::borrowck;
 use crate::context::{
-    self, AnalysisCtxt, AnalysisCtxtData, DontRewriteFieldReason, DontRewriteFnReason,
+    self, AnalysisCtxt, AnalysisCtxtData, Assignment, DontRewriteFieldReason, DontRewriteFnReason,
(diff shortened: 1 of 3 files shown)


ahomescu:
I have my own changes for the PDG at https://github.com/immunant/c2rust/pull/1158 and https://github.com/immunant/c2rust/tree/ahomescu/analyze-compare-static-dynamic-v2, we should probably de-duplicate our branches. Could you take a look at mine?

ahomescu:
> I have my own changes for the PDG at #1158 and https://github.com/immunant/c2rust/tree/ahomescu/analyze-compare-static-dynamic-v2, we should probably de-duplicate our branches. Could you take a look at mine?

After doing some rebasing, I see that your branch is the same as mine. Closing the latter.

ahomescu (approved):

Issue #755, opened 2022-11-29
`src/ExportResult.hpp:11:10: fatal error: 'algorithm' file not found`


kkysen:
`c2rust`, specifically `c2rust-ast-exporter`, sometimes fails to build with the error:

```shell
src/ExportResult.hpp:11:10: fatal error: 'algorithm' file not found
```

This is meant to be an unclosed issue tracking this error and potential solutions.  See also the very related #380.

Some potential solutions:

* Ubuntu 22:

  ```shell
  sudo apt install libstdc++-12-dev
  ```

  (from https://github.com/immunant/c2rust/issues/380#issuecomment-1312065339)

AdrianTayag:
Also experienced the same issue and was rectified by following the potential solution listed above. Bumping this to be added to the README

CearX:
Thank you very much, this has solved my problem.

Version:
ubuntu 2204
rust-nightly 1.79

PR #1162, opened 2024-11-15
Apply the fixing in #755

ssrlive:


thedataking (changes requested):
Is there a generic package that provides the required headers? If. you look at the other dependencies, we avoid specifying the version in their name such that it works across different systems. We should do the same here.

fw-immunant:
The generic package is just `libstdc++-dev`.

PR #1159, opened 2024-11-12
ci: remove Slack integration for `c2rust-testsuites` CI workflow

kkysen:
This is not very important, and it fails on external PRs, meaning we're unable to run `c2rust-testsuites` on them and catch potential errors.

* Fixes #1138.

PR #1128, opened 2024-09-18
`transpile`: Rewrite `fn compare_src_locs` implementation to have a total order

Kriskras99:
This implementation is simplified compared to the previous one. It is also almost twice as slow in [the exhaustive test](https://github.com/immunant/c2rust/issues/1126#issuecomment-2354876003) (15 vs 25 seconds).
However, in real sort usage the impact should be significantly less.

* Fixes #1126

Issue #1126, opened 2024-09-15
`SrcLoc` sorting is non-transitive and a false total order and equality (panics in 1.81)


Kriskras99:
Seems one (or more) of the types used in the transpiler have a wrong `Ord` implementation.
```
Transpiling cr.c
thread 'main' panicked at core/src/slice/sort/shared/smallsort.rs:860:5:
user-provided comparison function does not correctly implement a total order
stack backtrace:
   0:     0x620e4a93b8da - std::backtrace_rs::backtrace::libunwind::trace::h4f532649ff050af0
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:     0x620e4a93b8da - std::backtrace_rs::backtrace::trace_unsynchronized::h2bd9c4760a46d04c
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x620e4a93b8da - std::sys::backtrace::_print_fmt::h06a2f1333d2c9eb7
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/std/src/sys/backtrace.rs:66:9
   3:     0x620e4a93b8da - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hc39aee40395be66f
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/std/src/sys/backtrace.rs:39:26
   4:     0x620e4a9628db - core::fmt::rt::Argument::fmt::h391dd44a8ecf182f
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/core/src/fmt/rt.rs:177:76
   5:     0x620e4a9628db - core::fmt::write::h718938895d6a6b80
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/core/src/fmt/mod.rs:1178:21
   6:     0x620e4a937d83 - std::io::Write::write_fmt::h97874dbf951485b0
                               at /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/library/std/src/io/mod.rs:1823:15
   7:     0x620e4a93b722 - std::sys::backtrace::BacktraceLock::print::h28b06058dc60edca
[...]

PR #1163, opened 2024-11-18
transpile: fix `curl` transpile breakage while maintaining a total `SrcLoc` order

kkysen:
#1128 fixed the non-total (non-transitive) `SrcLoc` ordering, but accidentally broke the `curl` transpilation test in `c2rust-testsuites`, which is tested in CI, but was broken for external contributors PRs (fixed now in #1159).

In the `curl` transpilation, a couple of `use` imports (`__sigset_t` and `C2RustUnnamed_4`) were missing, cause the build to fail afterward.

I'm still not exactly sure why this fixes the issue while maintaining a transitive, total order, but it passes the total order test and transpiles `curl` correctly now.  Hopefully this is a complete fix, and I didn't just fix a one-off error in `curl`.

* 9679a3be is the real fix.
* a4a052b3 are generally useful for testing and debugging.

### DIFF
--- a/c2rust-ast-exporter/src/clang_ast.rs
+++ b/c2rust-ast-exporter/src/clang_ast.rs
@@ -1,9 +1,10 @@
 use serde_bytes::ByteBuf;
 use serde_cbor::error;
-use std;
 use std::collections::{HashMap, VecDeque};
 use std::convert::TryInto;
+use std::fmt::{Display, Formatter};
 use std::path::{Path, PathBuf};
+use std::{self, fmt};
 
 pub use serde_cbor::value::{from_value, Value};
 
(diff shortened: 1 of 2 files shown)


kkysen:
@Kriskras99, does this still also fix the crash you were seeing in #1126?

Kriskras99:
Still fixes the crash!

kkysen:
> Still fixes the crash!

Awesome!!

fw-immunant (approved):
This looks sane; it's easier to review by looking at the cumulative diff since df42b55eae9ecfd4380004a513a10526ef8776cf: `git diff df42b55eae9ecfd4380004a513a10526ef8776cf..9679a3befb9d5d9431ee7db4d967347edb64536d -- c2rust-transpile/src/c_ast/mod.rs`

PR #1164, opened 2024-11-22
analyze: use a common namespace for all local and global PointerIds

spernsteiner:
Currently, each function has its own namespace for local `PointerId`s: `PointerId::local(3)` in function `f` is distinct from `PointerId::local(3)` in function `g`.  This creates some problems for pointee type inference*.  This branch puts all functions' local `PointerId`s into a common namespace, with each function occupying a different range in that space.  This means each `PointerId` is now globally unique, and code that is analyzing one function can mention `PointerId`s from a different function without ambiguity.

Most analyses still only look at a single function; these use a sparse `PointerTable` that has entries only for global `PointerId`s (that is, `PointerId`s in the range allocated for globals) and for the local `PointerId`s of the current function.  Any analyses that need to consider cross-function local `PointerId`s can use a single large `GlobalPointerTable` instead.

---

\* The specific issue is this: the allocation in lighttpd's `buffer_init` is not initialized inside that function, so correctly inferring the type requires unifying type variables interprocedurally.  The type variables for the pointee analysis are tracked in a `VarTable`, which contains `LTy`s.  We'd like to track variables from all functions in one big `VarTable`, so that we can unify variables that originated in different functions.  However, `LTy`s may contain local `PointerId`s, so a shared `VarTable` could mix up local `PointerId`s from different functions/namespaces, producing nonsensical results.  The fix being applied here is to put all `PointerId`s into a single namespace, so all type variables can be tracked in a common `VarTable`, and there is no longer an obstacle to unifying type variables from different functions.

### DIFF
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -978,6 +978,10 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
             .get(def_id)
             .intersects(DontRewriteFnReason::ANALYSIS_INVALID_MASK)
     }
+
+    pub fn ptr_is_global(&self, ptr: PointerId) -> bool {
+        self.ptr_info.contains(ptr)
+    }
 }
 
 impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
@@ -1060,6 +1064,10 @@ impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
 
     pub fn local_ptr_base(&self) -> u32 {
         self.ptr_info.base()
+    }
+
+    pub fn ptr_is_global(&self, ptr: PointerId) -> bool {
+        self.gacx.ptr_is_global(ptr)
     }
 
     pub fn type_of<T: TypeOf<'tcx>>(&self, x: T) -> LTy<'tcx> {
@@ -1309,7 +1317,7 @@ fn remap_local_ptr_info(
     let mut new_local_ptr_info = LocalPointerTable::<PointerInfo>::new(base, count);
     let mut new_ptr_info = new_global_ptr_info.and_mut(&mut new_local_ptr_info);
     for (old, &new) in map.iter() {
-        if old.index() < old_local_ptr_info.base() {
+        if !old_local_ptr_info.contains(old) {
             // If `old` is global then `new` is also global, and this remapping was handled already
             // by `remap_global_ptr_info`.
             continue;
--- a/c2rust-analyze/src/context.rs
+++ b/c2rust-analyze/src/context.rs
@@ -4,10 +4,7 @@ use crate::borrowck::{AdtMetadata, FieldMetadata, OriginArg, OriginParam};
 use crate::known_fn::{all_known_fns, KnownFn};
 use crate::labeled_ty::{LabeledTy, LabeledTyCtxt};
 use crate::panic_detail::PanicDetail;
-use crate::pointer_id::{
-    GlobalPointerTable, LocalPointerTable, NextGlobalPointerId, NextLocalPointerId, PointerTable,
-    PointerTableMut,
-};
+use crate::pointer_id::{GlobalPointerTable, LocalPointerTable, PointerTable, PointerTableMut};
 use crate::util::{self, describe_rvalue, PhantomLifetime, RvalueDesc};
 use assert_matches::assert_matches;
 use bitflags::bitflags;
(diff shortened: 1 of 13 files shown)


c2rust-analyze/src/context.rs
```diff
@@ -834,8 +834,12 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
         self.fn_origins = fn_origin_args_params(self.tcx, &self.adt_metadata);
     }
 
-    pub fn function_context<'a>(&'a mut self, mir: &'a Body<'tcx>) -> AnalysisCtxt<'a, 'tcx> {
-        AnalysisCtxt::new(self, mir)
+    pub fn function_context<'a>(
+        &'a mut self,
+        mir: &'a Body<'tcx>,
+        base: u32,
```
ahomescu, line 840: Should the name be more detailed, e.g., `ptr_base`?

c2rust-analyze/src/context.rs
```diff
@@ -981,16 +994,21 @@ impl<'tcx> GlobalAnalysisCtxt<'tcx> {
...
+    pub fn ptr_is_global(&self, ptr: PointerId) -> bool {
+        self.ptr_info.contains(ptr)
+    }
 }
 
 impl<'a, 'tcx> AnalysisCtxt<'a, 'tcx> {
     pub fn new(
         gacx: &'a mut GlobalAnalysisCtxt<'tcx>,
         mir: &'a Body<'tcx>,
+        base: u32,
```
ahomescu, line 1007: `local_ptr_base`?

c2rust-analyze/src/equiv.rs
```diff
@@ -141,9 +143,9 @@ impl<'g> EquivSet<'g> {
 
     fn set_parent(&self, x: PointerId, parent: PointerId) {
         // Local items can point to global ones, but not vice versa.
-        if x.is_global() {
-            debug_assert!(parent.is_global());
-        }
+        //if x.is_global() {
```
ahomescu, line 146: Do you have a fix for this? You could just delete it.

c2rust-analyze/src/pointer_id.rs
```diff
@@ -60,21 +59,55 @@ impl fmt::Debug for PointerId {
...
+        self.index().hash(hasher);
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct NextLocalPointerId(u32);
 
 impl NextLocalPointerId {
-    pub fn new() -> NextLocalPointerId {
+    pub fn _new() -> NextLocalPointerId {
```
ahomescu, line 96: Why the rename?

c2rust-analyze/src/pointer_id.rs
```diff
@@ -302,18 +401,29 @@ impl<T> GlobalPointerTable<T> {
...
 }
 
+pub struct PointerRange {
+    base: u32,
+    len: u32,
+}
+
+impl PointerRange {
+    pub fn contains(&self, ptr: PointerId) -> bool {
+        // If `ptr.index() < self.base`, the subtraction will wrap to a large number in excess of
```
ahomescu, line 421: Does this require that `base + len` not overflow?

ahomescu: Update: I see above that you check for this. Maybe add a comment here?

ahomescu (approved):

c2rust-analyze/src/pointer_id.rs
spernsteiner: The function is currently unused, but I've left it in place for completeness of the API (`NextGlobalPointerId` has a `fn new()`, so the local version should too) and in case it's needed again in the future.  To suppress the dead code warning, I prefer a leading underscore rather than `#[allow(dead_code)]` because it makes it more likely that anyone who starts using the function in the future will also remember to un-suppress the warning.

c2rust-analyze/src/context.rs
spernsteiner: Changed both `base` arguments to `local_ptr_base` (b68e56876)

c2rust-analyze/src/equiv.rs
spernsteiner: Fixed in b68e56876

c2rust-analyze/src/pointer_id.rs
spernsteiner: Added a comment in b68e56876

I think this `contains` method actually works when `base + len` overflows, but some of the pointers near the top of the range would be unrepresentable as a `PointerId`.

PR #1170, opened 2024-11-27
`transpile`: emit zeroed array expressions as `[0; N]`

folkertdev:
an input like 

```c
int foo() {
    int x[16] = {};
    return 0;
}
```

was emitted as

```rust
    let mut bar: [libc::c_int; 16] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```

but now emits

```rust
    let mut bar: [libc::c_int; 16] = [0; 16];
```

The current version only supports floating and integral types: in more recent rust versions it would be easy to make this work for e.g. structs: just use a `const { ... }` block. But c2rust targets older versions, and there (without introducing extra constants) the expression's type must be copy, and I could not find a convenient way to tell whether a `CTypeId` will derive/implement `Copy`.

this came up when working with the output of `creduce`, which loves zeroing arrays.

### DIFF
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -190,6 +190,14 @@ impl<'c> Translation<'c> {
                 if is_string {
                     let v = ids.first().unwrap();
                     self.convert_expr(ctx.used(), *v)
+                } else if ids.len() == 0 {
+                    // this was likely a C array of the form `int x[16] = {}`,
+                    // we'll emit that as [0; 16].
+                    let len = mk().lit_expr(mk().int_unsuffixed_lit(n as u128));
+                    self.implicit_default_expr(ty, ctx.is_static)?
+                        .and_then(|default_value| {
+                            Ok(WithStmts::new_val(mk().repeat_expr(default_value, len)))
+                        })
                 } else {
                     Ok(ids
                         .iter()
--- a/tests/arrays/src/arrays.c
+++ b/tests/arrays/src/arrays.c
@@ -6,8 +6,13 @@ static char *foo = "mystring";
 void entry(const unsigned buffer_size, int buffer[const])
 {
     int arr[1][1] = { 1 };
-
     arr[0][0] += 9;
+
+    int arr2[16] = {};
+    arr2[15] += 9;
+
+    struct {char* x; int y;} arr3[1] = {};
+    arr3[0].y += 9;
 
     int i = 0;
 


thedataking:
Thanks Folkert, I like this idea. Your comment indicates the output Rust is likely correct – but not necessarily so. Any examples where this would fail? If so, should this be guarded by a flag like other risky/experimental features?

Also, it would be great if you could add a test or two.

folkertdev:
I believe the current change is correct, across rust versions. It just isn't as general as it could be.

For a "repeat" expression `[E; N]`, the expression `E` must be either

- a value of a type that is `Copy`
- a constant value (so a path to a `const` or, in newer versions, a `const { ... }` expression)

The current versions restricts the type to be numeric `ty.is_integral_type() || ty.is_floating_type()`, because we know that all of those numeric types `impl Copy`, and hence will work.

But e.g. a `typedef struct { int x, int y } Point` still gets emitted as a long array literal right now.

For such a struct, c2rust will actually emit a `#[derive(Clone, Copy)]` down the line, so if that information were available then an `[E; N]` style array could be emitted for any type that is `impl Copy`. That "does this impl `Copy`?" information must be somewhere, I just could not find it.

> Also, it would be great if you could add a test or two.

I added something basic, what else do you want to test and where?

kkysen:
@folkertdev, thanks for this improvement!  It seems like a sound transformation to me.

> For such a struct, c2rust will actually emit a `#[derive(Clone, Copy)]` down the line, so if that information were available then an `[E; N]` style array could be emitted for any type that is `impl Copy`. That "does this impl `Copy`?" information must be somewhere, I just could not find it.

I'm pretty sure that all C types are implicitly `Copy`, so `c2rust` should mark all such transpiled types `#[derive(Clone, Copy)]`.  I think we can include these, not just primitives.

That said, the CI is failing:

```shell
error[E0599]: no method named `is_integral_type` found for struct `c_ast::CTypeId` in the current scope
   --> c2rust-transpile/src/translator/literals.rs:193:48
    |
193 |                 } else if ids.len() == 0 && ty.is_integral_type() || ty.is_floating_type() {
    |                                                ^^^^^^^^^^^^^^^^ method not found in `c_ast::CTypeId`
    |
   ::: c2rust-transpile/src/c_ast/mod.rs:15:1
    |
15  | pub struct CTypeId(pub u64);
    | ------------------ method `is_integral_type` not found for this struct

error[E0599]: no method named `is_floating_type` found for struct `c_ast::CTypeId` in the current scope
   --> c2rust-transpile/src/translator/literals.rs:193:73
    |
193 |                 } else if ids.len() == 0 && ty.is_integral_type() || ty.is_floating_type() {
    |                                                                         ^^^^^^^^^^^^^^^^ method not found in `c_ast::CTypeId`
    |
   ::: c2rust-transpile/src/c_ast/mod.rs:15:1
    |
15  | pub struct CTypeId(pub u64);
    | ------------------ method `is_floating_type` not found for this struct
```

Did you see this?

kkysen (commented):
Oh, that's a good find.  I didn't think about VLAs and unsized types.

> I think that would not come up here because we're working with a `CTypeKind::ConstantArray`, but I'm not 100% sure.

I would think so, too.  If the C compiles, an unsized type shouldn't be allowed in an array, and therefore the array's element type should be `Copy`.

kkysen (approved):
LGTM now, but could someone else take a quite look to make sure there's not a case I'm overlooking where this might not work? @fw-immunant maybe?

fw-immunant:
I *think* this is fine, but I would want to test it against a case like `struct {char* x; int y;} x[1] = {};`.

fw-immunant (approved):

folkertdev:
I added that test case

PR #1172, opened 2024-12-02
analyze: allow pointee analysis to work interprocedurally

spernsteiner:
This branch modifies the pointee-type analysis to use a single global `VarTable` for all functions, which allows pointee inference variables produced in one function to be resolved in another.  This is necessary for lighttpd's `buffer` module, which allocates a `buffer` object in one function (producing an inference variable for the `malloc`) and initializes it in a different function (where the inference variable can now be resolved).

The pointee analysis previously involved an "export" operation that ran after processing each function to convert the analysis results to a form that makes sense in other functions.  This meant that the results couldn't contain any inference variables or mention any local `PointerId`s.  Any entries in the results that violated these constraints would be erased, and the result set would be marked "incomplete".  With this branch, both constraints are lifted, as both inference variables and local `PointerId`s are representable in the global scope.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -627,12 +627,6 @@ fn run(tcx: TyCtxt) {
         assert!(loop_count <= 1000);
         let old_global_pointee_types = global_pointee_types.clone();
 
-        // Clear the `incomplete` flags for all global pointers.  See comment in
-        // `pointee_types::solve::solve_constraints`.
-        for (_, tys) in global_pointee_types.iter_mut() {
-            tys.incomplete = false;
-        }
-
         for &ldid in &all_fn_ldids {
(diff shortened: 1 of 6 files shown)


c2rust-analyze/src/analyze.rs
```diff
@@ -2274,15 +2270,10 @@ fn print_function_pointee_types<'tcx>(
...
+            if tys.tys.is_empty() {
                 continue;
             }
-            debug!(
-                "  pointer {:?}: {:?}{}",
-                ptr,
-                tys.ltys,
-                if tys.incomplete { " (INCOMPLETE)" } else { "" }
-            );
+            debug!("  pointer {:?}: {:?}", ptr, tys.tys,);
```
ahomescu, line 2276: Is the extra comma needed?

ahomescu (approved):

PR #1173, opened 2024-12-02
analyze: run equiv pass before pointee

spernsteiner:
This branch changes the order of analysis passes to run the equivalence class analysis and `PointerId` renumbering before running the pointee type analysis.  In particular, this improves the pointee results in cases involving `fresh` references, such as this new test case:

```Rust
let mut p = 0 as *mut libc::c_void;
let fresh = &mut p;
*fresh = malloc(4);
```

Most of the changes here are refactorings to move specific analysis passes from `run()` into helper functions, which makes them easier to reorder.

### DIFF
--- a/c2rust-analyze/src/dataflow/type_check.rs
+++ b/c2rust-analyze/src/dataflow/type_check.rs
@@ -64,6 +64,18 @@ impl<'tcx> TypeChecker<'tcx, '_> {
         self.constraints.add_subset_except(dest, src, except);
     }
 
+    /// Add `Constraint::AllPerms`, which requires `ptr` to have all of the permissions listed in
+    /// `perms`.
+    fn add_all_perms(&mut self, ptr: PointerId, perms: PermissionSet) {
+        self.constraints.add_all_perms(ptr, perms);
+    }
+
+    /// Add `Constraint::NoPerms`, which requires `ptr` to have none of the permissions listed in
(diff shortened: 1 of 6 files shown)


c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -31,40 +31,79 @@ use rustc_middle::ty::{SubstsRef, Ty, TyKind};
...
-        self.equiv_constraints.push((a, b));
+        if let Some(ref mut equiv_constraints) = self.equiv_constraints {
+            equiv_constraints.push((a, b));
+        }
+    }
+
+    fn pointee_type(&self, ptr: PointerId) -> Option<LTy<'tcx>> {
+        self.pointee_types
+            .as_ref()
+            .and_then(|pointee_types| pointee_types[ptr].get_sole_lty())
```
ahomescu, line 106: Will this panic if `ptr` is out of bounds?

ahomescu (approved):

c2rust-analyze/src/dataflow/type_check.rs
spernsteiner: It will panic if `ptr` is `PointerId::NONE` (the same is true of the code this is replacing).  I added an assert just to make it clear that this should only be called on actual pointers: 7f1d501

PR #1174, opened 2024-12-02
analyze: script for automatically fixing some compile errors

spernsteiner:
This branch adds a Python script that automatically fixes some kinds of compiler errors by processing the output of `rustc --error-format json`.  It applies all `MachineApplicable` suggestions, plus two other kinds of fixes that don't come with a `MachineApplicable` suggestion (in our current nightly) but are important for lighttpd `buffer`:

* "Lifetime may not live long enough": For each function, `c2rust-analyze` adds a lifetime parameter for each reference in the arguments and return type, but doesn't add any outlives bounds (e.g. `'a: 'b`) to relate those lifetimes.  On functions that return a borrowed reference into one of their arguments, this results in a borrow checker error.  The error comes with a human-readable suggestion like "consider adding the following bound: `'a: 'b`", and also marks the declaration of the first lifetime (`'a`); the auto-fix script parses this output and inserts the suggested `: 'b` bound after the declaration.
* "The trait `Copy` may not be implemented for this type": When `c2rust-analyze` converts a field from raw pointer type to `&mut T` or `Box<T>`, it doesn't remove `derive(Copy)` from the struct.  This causes a compile error because `&mut T` and `Box<T>` don't implement `Copy`.  The error marks the location of the `Copy` token in `#[derive(Copy)]`; the auto-fix script removes that token (and the subsequent comma, if necessary).  Note this can produce a trailing comma (`#[derive(Clone, Copy)]` -> `#[derive(Clone, )]`) or an empty `derive()` (`#[derive(Copy)]` -> `#[derive()]`), both of which are legal but unidiomatic.

### DIFF
--- a/c2rust-analyze/scripts/auto_fix_errors.py
+++ b/c2rust-analyze/scripts/auto_fix_errors.py
@@ -1,5 +1,5 @@
 import argparse
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 import json
 import re
 import sys
@@ -231,13 +231,10 @@ def main():
         ))
 
     # Group fixes by file
-    fixes_by_file = {}
+    fixes_by_file = defaultdict(list)
     for fix in fixes:
-        file_fixes = fixes_by_file.get(fix.file_path)
-        if file_fixes is None:
-            fixes_by_file[fix.file_path] = [fix]
-        else:
-            file_fixes.append(fix)
+        fixes_by_file[fix.file_path].append(fix)
+    fixes_by_file = dict(fixes_by_file)
     for file_fixes in fixes_by_file.values():
         file_fixes.sort(key=lambda fix: fix.start_byte)
 
--- a/c2rust-analyze/scripts/auto_fix_errors.py
+++ b/c2rust-analyze/scripts/auto_fix_errors.py
@@ -1,5 +1,5 @@
 import argparse
-from dataclasses import dataclass
+from collections import namedtuple
 import json
 import re
 import sys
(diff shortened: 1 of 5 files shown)


c2rust-analyze/scripts/auto_fix_errors.py
```diff
@@ -0,0 +1,286 @@
...
+            file_path=rm.file_path,
+            line_number=rm.line_number,
+            start_byte=rm.start_byte,
+            end_byte=fix_end_byte,
+            new_text='',
+            message=MSG_DERIVE_COPY,
+        ))
+
+    # Group fixes by file
+    fixes_by_file = {}
```
ahomescu, line 234: Could you simplify this using `fixes_by_file = defaultdict(list)`?

c2rust-analyze/scripts/auto_fix_errors.py
```diff
@@ -0,0 +1,286 @@
...
+RemoveDeriveCopy = namedtuple('RemoveDeriveCopy', (
+    'file_path',
+    'line_number',
+    # Byte offset of the start of the token `Copy`.
+    'start_byte',
+    # Byte offset of the end of the token `Copy`.
+    'end_byte',
+))
+
+MSG_LIFETIME_BOUND = 'lifetime may not live long enough'
```
ahomescu, line 54: When I ran the analyzer on amalgamated lighttpd, I got a few other common errors:
* `error[E0106]: missing lifetime specifiers`
* ```error[E0277]: `*mut i8` cannot be shared between threads safely``` caused by rewriting some `static mut` globals into `static`.
* ```error[E0392]: parameter `'h16` is never used``` and more like it; basically just unused lifetimes
* `error[E0726]: implicit elided lifetime not allowed here`

Some or all of these could also be handled by your tool (E0392 specifically comes to mind). Should we add them to this PR, or a follow-up?

ahomescu (approved):

c2rust-analyze/scripts/auto_fix_errors.py
```diff
@@ -0,0 +1,286 @@
...
+))
+
+RemoveDeriveCopy = namedtuple('RemoveDeriveCopy', (
+    'file_path',
+    'line_number',
+    # Byte offset of the start of the token `Copy`.
+    'start_byte',
+    # Byte offset of the end of the token `Copy`.
+    'end_byte',
+))
```
kkysen, line 52: nit: I think it's a lot more idiomatic to use `@dataclass`es instead of `namedtuple`.

spernsteiner: I originally used `@dataclass` here, but switched to `namedtuple` because CI was failing - the Ubuntu 18 runner has Python 3.6, but dataclasses were added in 3.7

c2rust-analyze/scripts/auto_fix_errors.py
spernsteiner: Reverse-engineering a correct fix from the error json is nontrivial, so handling more errors should be left to a separate PR.

c2rust-analyze/scripts/auto_fix_errors.py
kkysen: Oh that's really annoying. Python 3.6 was EOL 3 years ago.

c2rust-analyze/scripts/auto_fix_errors.py
spernsteiner: Done in [7098d9e](https://github.com/immunant/c2rust/pull/1174/commits/7098d9e8a732db4cf7020bb48764e16df00c5a07)

PR #1175, opened 2024-12-02
analyze: more def list options

spernsteiner:
This adds more command line options for controlling which defs get rewritten.  Specifically, it adds two new options:

* `--force-rewrite-defs-list FILE`: Reads from `FILE` a list of `DefId`s that should always be rewritten, even if related defs encounter analysis or rewriting errors.
* `--skip-pointee-defs-list FILE`: Reads from `FILE` a list of `DefId`s for which pointee analysis should be skipped.

Both options take input in the same format as the existing `--fixed-defs-list FILE` option.

These options were useful for getting the `buffer` module to rewrite successfully while running `c2rust-analyze` against the entire lighttpd codebase.

This PR also fixes a bug that caused `DefId`s to be erroneously rejected when parsing the `--fixed-defs-list` file.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -528,6 +528,14 @@ fn get_force_rewrite_defs() -> io::Result<HashSet<DefId>> {
     Ok(force_rewrite)
 }
 
+fn get_skip_pointee_defs() -> io::Result<HashSet<DefId>> {
+    let mut skip_pointee = HashSet::new();
+    if let Ok(path) = env::var("C2RUST_ANALYZE_SKIP_POINTEE_LIST") {
+        read_defs_list(&mut skip_pointee, &path)?;
+    }
+    Ok(skip_pointee)
+}
+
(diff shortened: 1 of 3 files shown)


c2rust-analyze/src/analyze.rs
```diff
@@ -425,7 +425,7 @@ fn parse_def_id(s: &str) -> Result<DefId, String> {
     };
 
     let rendered = format!("{:?}", def_id);
-    if &rendered != s {
+    if &rendered != orig_s {
```
ahomescu, line 428: Was this a bug previously?

ahomescu (approved):

c2rust-analyze/src/analyze.rs
spernsteiner: Yes - it would erroneously reject some `DefId`s when parsing the `--fixed-defs-list` file.  I edited the PR description (which will be included in the eventual merge commit) to mention this

PR #1176, opened 2024-12-02
analyze: memcpy rewrite improvements

spernsteiner:
This branch has several improvements to the rewriting of `memcpy` and `memset`:

* Support `memcpy` on `Quantity::Single` references (`&T`/`&mut T`, as opposed to `&[T]`/`&mut [T]`).  The reference is automatically converted to a 1-element slice.  The `copy_from_slice` call will panic with an out-of-bounds error if the number of elements to copy is not 0 or 1.
* Support `memcpy` on `Option` references.  If the source or destination pointer is `None`, then this will panic if the element count is nonzero, and otherwise will do nothing.
* Use `mem::size_of::<T>()` instead of the original size of `T` for converting `memcpy`/`memset` byte counts to element counts.

The `size_of` change is a bit subtle.  In C, if `sizeof(struct foo) == 16`, it's legal to `memcpy` an array of `struct foo` using `n * 16` as the byte length (or `n * SIZE_OF_FOO`, where `SIZE_OF_FOO` is `#define`'d to be 16) instead of `n * sizeof(struct foo)`.  This presents a problem for us because `c2rust-analyze` may change the size of `foo` when rewriting its pointer fields.  After rewriting, the `n * 16` version computes a byte length based on the original size for `struct foo`, while the `n * sizeof(struct foo)` computes it using the rewritten size.  For converting the byte length to an element count, we previously used the original size, which works for the `n * 16` version but not for `n * sizeof(struct foo)`.  This branch changes the element count calculation to divide by the rewritten size instead, which works for `n * sizeof(struct foo)` but not for `n * 16`.  The hope is that `n * sizeof(struct foo)` is more common, since it's usually preferred in C for portability reasons.

### DIFF
--- a/c2rust-analyze/src/rewrite/expr/convert.rs
+++ b/c2rust-analyze/src/rewrite/expr/convert.rs
@@ -224,7 +224,7 @@ impl<'tcx> ConvertVisitor<'tcx> {
             }
 
             mir_op::RewriteKind::MemcpySafe {
-                elem_size,
+                ref elem_ty,
                 dest_single,
                 dest_option,
                 src_single,
@@ -241,7 +241,7 @@ impl<'tcx> ConvertVisitor<'tcx> {
                 ]));
                 stmts.push(Rewrite::Let(vec![(
                     "n".into(),
-                    format_rewrite!("byte_len as usize / {elem_size}"),
+                    format_rewrite!("byte_len as usize / std::mem::size_of::<{elem_ty}>()"),
                 )]));
                 let mut convert = |var: &str, is_mut, is_single, is_option| {
                     let single_to_slice = if is_single {
@@ -298,7 +298,7 @@ impl<'tcx> ConvertVisitor<'tcx> {
 
             mir_op::RewriteKind::MemsetZeroize {
                 ref zero_ty,
-                elem_size,
+                ref elem_ty,
                 dest_single,
             } => {
                 // `memset(dest, 0, n)` to assignments that zero out each field of `*dest`
@@ -320,7 +320,7 @@ impl<'tcx> ConvertVisitor<'tcx> {
                         ]),
                         Rewrite::Let(vec![(
                             "n".into(),
-                            format_rewrite!("byte_len as usize / {elem_size}"),
+                            format_rewrite!("byte_len as usize / std::mem::size_of::<{elem_ty}>()"),
                         )]),
                         format_rewrite!("assert_eq!(val, 0, \"non-zero memset NYI\")"),
                         zeroize_body,
@@ -331,12 +331,12 @@ impl<'tcx> ConvertVisitor<'tcx> {
 
             mir_op::RewriteKind::MallocSafe {
                 ref zero_ty,
-                elem_size,
+                ref elem_ty,
                 single,
             }
             | mir_op::RewriteKind::CallocSafe {
                 ref zero_ty,
-                elem_size,
+                ref elem_ty,
                 single,
             } => {
                 // `malloc(n)` -> `Box::new(z)` or similar
@@ -347,7 +347,9 @@ impl<'tcx> ConvertVisitor<'tcx> {
                         Rewrite::Let(vec![("byte_len".into(), self.get_subexpr(ex, 0))]),
                         Rewrite::Let1(
                             "n".into(),
-                            Box::new(format_rewrite!("byte_len as usize / {elem_size}")),
+                            Box::new(format_rewrite!(
+                                "byte_len as usize / std::mem::size_of::<{elem_ty}>()"
+                            )),
                         ),
                     ],
                     mir_op::RewriteKind::CallocSafe { .. } => vec![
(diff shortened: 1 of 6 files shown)


c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -304,15 +347,19 @@ impl<'tcx> ConvertVisitor<'tcx> {
                         Rewrite::Let(vec![("byte_len".into(), self.get_subexpr(ex, 0))]),
                         Rewrite::Let1(
                             "n".into(),
-                            Box::new(format_rewrite!("byte_len as usize / {elem_size}")),
+                            Box::new(format_rewrite!(
+                                "byte_len as usize / std::mem::size_of::<{elem_ty}>()"
```
ahomescu, line 351: Should we have a check here (maybe an assertion) that `byte_len` is a multiple of the element size?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -224,38 +224,81 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                            format_rewrite!("&mut []")
+                        } else {
+                            format_rewrite!("&[]")
+                        };
+                        Rewrite::Match(
+                            Box::new(Rewrite::Text(var.into())),
+                            vec![
+                                (format!("Some({var})"), single_to_slice),
+                                (
+                                    "None".into(),
```
ahomescu, line 267: This is `None => Some(&[])`, right? A comment might help.

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -224,38 +224,81 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                let mut stmts = Vec::with_capacity(5);
+
+                stmts.push(Rewrite::Let(vec![
+                    ("dest".into(), self.get_subexpr(ex, 0)),
+                    ("src".into(), self.get_subexpr(ex, 1)),
+                    ("byte_len".into(), self.get_subexpr(ex, 2)),
+                ]));
+                stmts.push(Rewrite::Let(vec![(
+                    "n".into(),
+                    format_rewrite!("byte_len as usize / std::mem::size_of::<{elem_ty}>()"),
```
ahomescu, line 244: Same comment as below: should we assert here (or at least warn) if `byte_len` is not a multiple of the element size?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -224,38 +224,81 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
 
             mir_op::RewriteKind::MemcpySafe {
-                elem_size,
+                ref elem_ty,
                 dest_single,
+                dest_option,
                 src_single,
+                src_option,
             } => {
                 // `memcpy(dest, src, n)` to a `copy_from_slice` call
```
ahomescu, line 233: The actual call is `memcpy(dest, src, byte_len)` right?

c2rust-analyze/src/rewrite/expr/convert.rs
```diff
@@ -224,38 +224,81 @@ impl<'tcx> ConvertVisitor<'tcx> {
...
+                            ],
+                        )
+                    } else {
+                        single_to_slice
+                    };
+                    stmts.push(Rewrite::Let1(var.into(), Box::new(rhs)));
+                };
+                convert("dest", true, dest_single, dest_option);
+                convert("src", false, src_single, src_option);
+                stmts.push(Rewrite::MethodCall(
```
ahomescu, line 282: `dest[..n].copy_from_slice(&src[..n])` is the rewritten code?

ahomescu (approved):

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Added an assertion for this and similar cases: dfecf02

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Added an assertion in dfecf02

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Added a comment to show the example output in 4808c78

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Updated comment in 4808c78

c2rust-analyze/src/rewrite/expr/convert.rs
spernsteiner: Yes - I added a comment showing this in 4808c78

PR #1179, opened 2024-12-02
analyze: move `Option<&mut T>` on last use instead of reborrowing

spernsteiner:
Currently, we reborrow `Option<&mut T>` at most use sites by calling `p.as_deref_mut()`.  This creates a new reference that's tied to the lifetime of the `Option`, rather than the lifetime of the original reference.  This is fine for temporaries and local variables, but creates a problem when returning the result from the current function:

```Rust
fn f(x: Option<&mut Foo>) -> Option<&mut i32> {
    x.as_deref_mut().map(|x| &mut x.field)
}
```

This produces a borrowck error because `as_deref_mut` has the signature `(&'b mut Option<&'a mut T>) -> Option<&'b mut T>`, meaning the result must not outlive `x`.

The fix here is to consume `x`:

```Rust
fn f(x: Option<&mut Foo>) -> Option<&mut i32> {
    // No `.as_deref_mut()` - `x` is moved into `map()`
    x.map(|x| &mut x.field)
}
```

In this case, `map` has the signature `(Option<&'a mut T>, [closure]) -> Option<&'a mut i32>`, preserving the lifetime of the input reference.

This branch adds an analysis to identify the "last use" of each local variable and modifies the rewriting rules around `Option<T>` to omit `.as_deref_mut()` at the last use.

### DIFF
--- a/c2rust-analyze/src/last_use.rs
+++ b/c2rust-analyze/src/last_use.rs
@@ -1,0 +1,365 @@
+//! Analysis to find the last use of a MIR local.  For non-`Copy` types, the last use can be made
+//! into a move instead of a clone or borrow.
+//!
+//! We specifically use this when handling borrows of non-`Copy` pointers like `Option<&mut T>`.
+//! At most use sites, we produce `ptr.as_deref_mut().unwrap() ...`, which borrows `ptr` instead of
+//! moving it.  However, this construction produces a reference with a shorter lifetime, matching
+//! the lifetime of the local `ptr`.  When returning a borrow from a function, we instead want to
+//! emit `ptr.unwrap() ...`, which consumes `ptr` but produces a reference with the same lifetime
+//! as the reference in `ptr`.
+
+use std::collections::{HashMap, VecDeque};
+
+use rustc_index::bit_set::BitSet;
+use rustc_index::vec::IndexVec;
+use rustc_middle::mir::traversal;
+use rustc_middle::mir::{
+    BasicBlock, BasicBlockData, Body, Local, Location, Operand, Place, Rvalue, Statement,
+    StatementKind, Terminator, TerminatorKind,
+};
+
+/// A single MIR `Statement` or `Terminator` may refer to multiple `Place`s in different
+/// `Operand`s.  This type identifies a particular `Place` of interest within the `Statement`
+/// or`Terminator`.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
+pub enum WhichPlace {
+    Lvalue,
+    Operand(usize),
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct LastUse {
+    /// `(location, which_place)` is present in this map if the indicated `Place` is the last use
+    /// of that `Place`'s `Local`.
+    ///
+    /// Note that there may be more than one "last use" of a given local, if it is used, written,
+    /// and then used again.
+    ///
+    /// The `Local` value is used only for debugging; it isn't needed for generating rewrites.
+    set: HashMap<(Location, WhichPlace), Local>,
+}
+
+impl LastUse {
+    pub fn is_last_use(&self, loc: Location, which: WhichPlace) -> bool {
+        self.set.contains_key(&(loc, which))
+    }
+
+    pub fn iter<'a>(&'a self) -> impl Iterator<Item = (Location, WhichPlace, Local)> + 'a {
+        self.set
+            .iter()
+            .map(|(&(loc, which), &local)| (loc, which, local))
+    }
+}
+
+struct Action {
+    kind: ActionKind,
+    local: Local,
+    location: Location,
+    which_place: WhichPlace,
+}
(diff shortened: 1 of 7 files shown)


c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -383,6 +406,44 @@ impl<'a, 'tcx> ExprRewriteVisitor<'a, 'tcx> {
...
     }
 
+    fn current_sub_loc_as_which_place(&self) -> Option<last_use::WhichPlace> {
+        // This logic is a bit imprecise, but should be correct in the cases where we actually call
+        // it (namely, when the `Place`/`Rvalue` is simply a `Local`).
+        let mut which = None;
+        for sl in &self.sub_loc {
+            match *sl {
+                SubLoc::Dest => {
+                    which = Some(last_use::WhichPlace::Lvalue);
```
ahomescu, line 416: Can there be multiple `sub_loc`s that set `which` in this loop? Do you always want the last one?

If you only need one, you could exit early with `return`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
```diff
@@ -1404,6 +1515,14 @@ pub struct CastBuilder<'a, 'tcx, PT1, PT2, F> {
     perms: &'a PT1,
     flags: &'a PT2,
     emit: F,
+    /// If set, the cast builder is allowed to move the expression being cast.  For example, we
+    /// might emit `ptr.map(|p| &mut *p).unwrap()` instead of `ptr.as_deref_mut().unwrap()` when
+    /// `can_move` is set; the former moves out of `ptr` but avoids an additional borrow, making it
+    /// suitable to be used as the result expression of a function.
+    can_move: bool,
+    /// If set, the cast builder will emit a downgrade/borrow operation even for no-op casts, if
+    /// the thing being cast can't be moved (`!can_move`) and also can't be copied.
```
ahomescu, line 1524: Are there only 3 states between the 2 variables: can-move, can-copy, must-borrow? Maybe using an `enum` here would be better.

ahomescu (approved):

c2rust-analyze/src/last_use.rs
```diff
@@ -0,0 +1,365 @@
...
+    Lvalue,
+    Operand(usize),
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct LastUse {
+    /// `(location, which_place)` is present in this map if the indicated `Place` is the last use
+    /// of that `Place`'s `Local`.
+    ///
+    /// Note that there may be more than one "last use" of a given local, if it is used, written,
```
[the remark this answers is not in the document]
spernsteiner: Yes - added a comment in ab616f362 that explains the various possibilities

c2rust-analyze/src/last_use.rs
```diff
@@ -0,0 +1,365 @@
...
+                    eprintln!("  enqueue {pred:?}");
+                    queue.push_back(pred);
+                }
+            }
+        }
+    }
+
+    debug_assert!(dirty.is_empty());
+
+    // Make a final pass over each block to find the `LastUse` of each variable.  Any time we see
```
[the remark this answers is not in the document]
spernsteiner: I mentioned an example like this in the comment added in ab616f362, and also more generally explained the connection to liveness analysis in 28cfbd450.

ahomescu (approved):

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: There are indeed only three possible behaviors - `can_move && borrow` produces the same cast as `can_move && !borrow`.  However, the code that sets the flags handles the two independently (and in particular, sometimes produces the `can_move && borrow` case).  So I think it's better to leave the code as-is, with the logic mapping the four cases to three outcomes centralized in the builder.

c2rust-analyze/src/last_use.rs
```diff
@@ -0,0 +1,365 @@
...
+    // an `ActionKind::Use` where the local is live before but dead after, we record that action as
+    // a "last use".
+    let mut last_use = LastUse {
+        set: HashMap::new(),
+    };
+    for (bb, actions) in block_actions.iter_enumerated() {
+        let mut live = BitSet::new_empty(mir.local_decls.len());
```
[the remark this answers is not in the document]
spernsteiner: The amount of duplicated code is small enough that I think it's fine to leave it as-is.  The tricky logic of mapping MIR statements to defs and uses of locals is already factored out in the building of the `block_actions`.

c2rust-analyze/src/rewrite/expr/mir_op.rs
spernsteiner: Added a comment in 8f40e87fe91b83f96a642fc168c683edd253ecf1 - there can indeed be multiple matching `sub_loc`s, and we want the last one.

PR #1180, opened 2024-12-02
analyze: misc fixes for lighttpd buffer (part 2/2)

spernsteiner:
More small fixes for lighttpd's `buffer` module:

* In `dataflow`, require `WRITE` permission for the no-op reborrow projection `&mut *p`.  This projection requires `p: &mut T` rather than `p: &T`.
* Add a script for collecting successfully rewritten defs.  The script reads the output of `c2rust-analyze` to identify defs for which analysis and rewriting succeeded without errors, and extracts only those defs into a separate file.  This is useful for checking whether the results of rewriting compile after testing on a particular module when there are compile errors elsewhere in the codebase.
* Fix some bugs in the pointwise metrics scripts

### DIFF
--- a/c2rust-analyze/scripts/extract_working_defs.py
+++ b/c2rust-analyze/scripts/extract_working_defs.py
@@ -1,0 +1,115 @@
+"""
+Script for extracting defs on which rewriting succeeded.  Currently, this is
+specific to lighttpd_rust_amalgamated, but it could be generalized to work on
+other projects as well.  The script identifies "working defs" by reading the
+log of a `c2rust-analyze` run and collecting defs that have no reported
+"analysis failed" errors and are not listed in the `--fixed-defs-list`.  It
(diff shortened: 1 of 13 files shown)


c2rust-analyze/scripts/extract_working_defs.py
```diff
@@ -0,0 +1,115 @@
+"""
```
ahomescu, line 1:
```suggestion
#!/usr/bin/env python3
"""
```

ahomescu (approved):

c2rust-analyze/scripts/extract_working_defs.py
spernsteiner: I added the `#!` and also ran `chmod +x` on both `extract_working_defs.py` and `auto_fix_errors.py`, in 74d3226325e273ab353c441a73e26bf7e5c7723c

PR #1181, opened 2024-12-02
analyze: add option to skip borrowck

spernsteiner:
This adds a new command-line option `--skip-borrowck` that completely disables the borrowck pass.  All pointers are given the `UNIQUE` permission, allowing them to be rewritten to `&mut T` instead of `&Cell<T>`.  This is useful when the code uses language features that aren't supported by borrowck and it's known (or assumed) that it doesn't require any `Cell` rewrites.

### DIFF
--- a/c2rust-analyze/src/analyze.rs
+++ b/c2rust-analyze/src/analyze.rs
@@ -785,6 +785,8 @@ fn run(tcx: TyCtxt) {
         }
     }
 
+    let skip_borrowck_everywhere = env::var("C2RUST_ANALYZE_SKIP_BORROWCK").as_deref() == Ok("1");
+
     // Load permission info from PDG
     let pdg_compare = env::var("C2RUST_ANALYZE_COMPARE_PDG").as_deref() == Ok("1");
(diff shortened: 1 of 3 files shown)


c2rust-analyze/src/analyze.rs
```diff
@@ -2269,7 +2281,7 @@ fn pdg_update_permissions<'tcx>(
                     perms.insert(PermissionSet::OFFSET_SUB);
                 }
                 if !node_info.unique {
-                    perms.remove(PermissionSet::UNIQUE);
+                    //perms.remove(PermissionSet::UNIQUE);
```
ahomescu, line 2284: Should this be enabled for `!skip_borrowck_everywhere`?

spernsteiner: Fixed in 0ec40f6.

I originally developed this branch by just commenting out a bunch of borrowck/`UNIQUE`-related code.  I guess I missed this part when converting it to be toggleable based on a command-line option.

ahomescu (approved):

PR #1182, opened 2024-12-03
analyze: add test based on lighttpd buffer

spernsteiner:
Adds a test case based on lighttpd's `buffer` module.  This includes FileCheck assertions on several of the functions to make sure we're getting the expected type and expression rewrites.

### DIFF
--- a/c2rust-analyze/tests/filecheck/lighttpd_buffer.rs
+++ b/c2rust-analyze/tests/filecheck/lighttpd_buffer.rs
@@ -1,0 +1,2872 @@
+//! --use-manual-shims
+#![feature(rustc_private)]
+#![feature(register_tool)]
+#![register_tool(c2rust_analyze_test)]
+
+/*
+This file should compile as 100% safe code after running `c2rust-analyze` and appropriate cleanup.
+```sh
+# Run `c2rust-analyze` with standard options
+cargo test lighttpd_buffer
+
+# From `lighttpd_buffer.rs.analysis.txt`, manually copy the lines between the `==== BEGIN` and
+# `==== END` markers into a separate file, `test_buffer.rs`.
+
+# Auto-fix remaining compile errors
+rustc --crate-type rlib --error-format json test_buffer.rs 2>errors.json
+python3 scripts/auto_fix_errors.py errors.json
+# Run again to fix newly uncovered errors
+rustc --crate-type rlib --error-format json test_buffer.rs 2>errors.json
+python3 scripts/auto_fix_errors.py errors.json
+
+# Remove `unsafe`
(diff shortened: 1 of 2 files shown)


c2rust-analyze/tests/filecheck/lighttpd_buffer.rs
```diff
@@ -0,0 +1,2872 @@
...
+        __tp: *const tm,
+    ) -> size_t;
+    */
+}
+
+#[no_mangle]
+#[cold]
+#[c2rust_analyze_test::fixed_signature]
+pub unsafe extern "C" fn ck_assert_failed(
+    /* TODO: all ck_assert_failed arguments are commented out due to bad casts at call sites
```
ahomescu, line 80: Should we open an issue on this?

c2rust-analyze/tests/filecheck/lighttpd_buffer.rs
```diff
@@ -0,0 +1,2872 @@
...
+pub struct const_iovec {
+    pub iov_base: *const libc::c_void,
+    pub iov_len: size_t,
+}
+pub type buffer_encoding_t = libc::c_uint;
+pub const ENCODING_MINIMAL_XML: buffer_encoding_t = 3;
+pub const ENCODING_HTML: buffer_encoding_t = 2;
+pub const ENCODING_REL_URI_PART: buffer_encoding_t = 1;
+pub const ENCODING_REL_URI: buffer_encoding_t = 0;
+// CHECK-LABEL: fn buffer_is_blank<'h0>(mut b: &'h0 (buffer)) -> libc::c_int {
```
ahomescu, line 140: How difficult would it be to avoid emitting the extra parentheses?

```suggestion
// CHECK-LABEL: fn buffer_is_blank<'h0>(mut b: &'h0 buffer) -> libc::c_int {
```

spernsteiner: This is the same issue as https://github.com/immunant/c2rust/pull/1148#discussion_r1850802900.  It would be nontrivial to fix within c2rust-analyze.  Whether the parens are needed or not depends on what kind of HIR node is inside them, but in `rewrite::apply` where the parens are actually being inserted, the HIR node isn't available (`rewrite::apply` operates on the source code as a string rather than HIR).

ahomescu (approved):

c2rust-analyze/tests/filecheck/lighttpd_buffer.rs
spernsteiner: Opened #1189

PR #1187, opened 2024-12-08
fix: clippy::not_unsafe_ptr_arg_deref

Nugine:
https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref

### DIFF
--- a/analysis/runtime/src/handlers.rs
+++ b/analysis/runtime/src/handlers.rs
@@ -161,8 +161,8 @@ pub fn addr_of_local(mir_loc: MirLocId, ptr: usize, local: u32, size: u32) {
     });
 }
 
-pub fn addr_of_sized<T: ?Sized>(mir_loc: MirLocId, ptr: *const T) {
-    let size = unsafe { core::mem::size_of_val(&*ptr) };
+pub unsafe fn addr_of_sized<T: ?Sized>(mir_loc: MirLocId, ptr: *const T) {
+    let size = core::mem::size_of_val(&*ptr);
     RUNTIME.send_event(Event {
         mir_loc,
         kind: EventKind::AddrOfSized {
--- a/analysis/runtime/src/handlers.rs
+++ b/analysis/runtime/src/handlers.rs
@@ -161,8 +161,10 @@ pub fn addr_of_local(mir_loc: MirLocId, ptr: usize, local: u32, size: u32) {
     });
 }
 
+/// # Safety
+/// `ptr` must be dereferenceable.
 pub unsafe fn addr_of_sized<T: ?Sized>(mir_loc: MirLocId, ptr: *const T) {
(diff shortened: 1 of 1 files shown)


fw-immunant:
Surface rust never calls this function, so it's not really meaningful to mark it `unsafe`.

analysis/runtime/src/handlers.rs
```diff
@@ -161,8 +161,8 @@ pub fn addr_of_local(mir_loc: MirLocId, ptr: usize, local: u32, size: u32) {
     });
 }
 
-pub fn addr_of_sized<T: ?Sized>(mir_loc: MirLocId, ptr: *const T) {
-    let size = unsafe { core::mem::size_of_val(&*ptr) };
+pub unsafe fn addr_of_sized<T: ?Sized>(mir_loc: MirLocId, ptr: *const T) {
+    let size = core::mem::size_of_val(&*ptr);
```
kkysen, line 165:
```suggestion
/// # Safety
/// `ptr` must be dereferenceable.
pub unsafe fn addr_of_sized<T: ?Sized>(mir_loc: MirLocId, ptr: *const T) {
    let size = core::mem::size_of_val(unsafe { &*ptr });
```

kkysen (commented):
It doesn't matter if this is called normally or not.  It's unsound to mark this as safe.  Seems like it passes all of the tests marked `unsafe`, so that seems fine.

fw-immunant:
My point is that there is no calling `unsafe` block that would place the other half (caller-side) of the `SAFETY` comment here, as calls to this function are [inserted by safe code that modifies MIR](https://github.com/immunant/c2rust/blob/f106b9bbc00233eefc0eea19785839f8bfd9a1bb/dynamic_instrumentation/src/instrument.rs#L644-L652) regardless of this function's unsafety. If our MIR transformations are incorrect, the object program may exhibit UB; we're operating past Rust's safety boundary here.

Nugine:
This PR is just a style fix for [`clippy::not_unsafe_ptr_arg_deref`](https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref). I see this error when using the latest nightly toolchain.
Feel free to accept or close this PR.

kkysen:
> My point is that there is no calling `unsafe` block that would place the other half (caller-side) of the `SAFETY` comment here, as calls to this function are [inserted by safe code that modifies MIR](https://github.com/immunant/c2rust/blob/f106b9bbc00233eefc0eea19785839f8bfd9a1bb/dynamic_instrumentation/src/instrument.rs#L644-L652) regardless of this function's unsafety. If our MIR transformations are incorrect, the object program may exhibit UB; we're operating past Rust's safety boundary here.

That part I understand, but we still shouldn't define unsound functions.  And it's simple to fix it this way, and we can still grep for this `unsafe fn`.

PR #1192, opened 2024-12-20
fix typos

jayvdb:


kkysen (approved):

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -752,7 +752,7 @@ impl<'c> Translation<'c> {
                     // Extract expression
                     let in_expr = in_expr.map(|(i, operand)| (i, operand.expression));
 
-                    // For inouts, change the dirspec to include 'in'
```
fw-immunant, line 755: NAK. An inout is not an input, it's a bidirectional (in and out) argument; see above where we also mention that we are handling inout arguments.

fw-immunant (changes requested):
The change in `assembly.rs` incorrectly alters the sense of the comment there. Please revert that hunk.

c2rust-transpile/src/translator/assembly.rs
kkysen: Oh whoops I didn't see that.

jayvdb:
I have also added a .typos.toml , config for https://crates.io/crates/typos-cli ; happy to remove if not desired, or move into a .config/ in order to reduce top level files.

fw-immunant (approved):
LGTM now.

PR #1197, opened 2025-01-27
Fix proc macro punctuation panic

fw-immunant:
Should fix #1142.

PR #1202, opened 2025-02-21
correctly construct _ as Ident rather than Punctuation

chrysn:
This fixes another bug like https://github.com/immunant/c2rust/pull/1197: `_` is no puctuation -- another case of https://github.com/dtolnay/proc-macro2/issues/475, fall-out of https://github.com/dtolnay/proc-macro2/issues/470.

## Testing

This enables building RIOT sources again without `--locked`. I'm not entirely sure, but I think the offending assembly was

```c
  __ASM ("rev %0, %1" : __CMSIS_GCC_OUT_REG (result) : __CMSIS_GCC_USE_REG (value) );
```

which gets turned into

```rust
    asm!(
        "rev {0}, {1}", lateout(reg) result, inlateout(reg) value => _,
        options(preserves_flags, pure, readonly)
    );
```

where the `_` is the identifier (formerly punctuation) which c2rust trips over.

### DIFF
--- a/c2rust-transpile/src/translator/assembly.rs
+++ b/c2rust-transpile/src/translator/assembly.rs
@@ -913,6 +913,7 @@ impl<'c> Translation<'c> {
             // Emit input and/or output expressions, separated by "=>" if both
             push_expr(&mut tokens, mk().paren_expr(constraints_ident));
             if let Some(in_expr) = in_expr {
+                let in_expr_span = in_expr.span();
                 push_expr(&mut tokens, in_expr);
                 if out_expr.is_some() {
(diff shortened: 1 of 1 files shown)


chrysn:
Note that I did not put much thought into the span that is put in. Looks natural to use this one, could also have used `span` for less precision, but given this is all just used to generate a file (and not directly compiler input), I'm not sure it matters at all what is in there -- at least I wouldn't know how to trigger anything where it makes a difference and thus can be tested what the span in there means.

fw-immunant (approved):
LGTM. I don't think getting the span perfectly minimal is a big deal here.

PR #1203, opened 2025-02-24
Release 0.20.0

kkysen:
Following as much from [c2rust/wiki/Release-Process](https://github.com/immunant/c2rust/wiki/Release-Process) as still works/applies, plus some extra additions, I ...
* [ ] `cargo outdated -R`: doesn't work due to unknown print request `split-debuginfo`
* [ ] `cargo update`: doesn't work because of old nightly
* [ ] `cargo clippy`:
  * [x] `cargo clippy --fix`
  * [x] Other trivial `cargo clippy` fixes
* [x] `cargo fmt --check`
* [x] Updated versions from `0.19.0` to `0.20.0`.
* [x] Specified missing versions for relative dependencies.
* [x] `cargo build`
* [x] `cargo +stable build -p c2rust`
* [x] Set `publish = false` in crates that are unstable due to `#![feature(rustc_private)]`.

Then I will publish the crates in this order with `cargo +stable -p $package`:
* `c2rust-asm-casts`
* `c2rust-ast-builder`
* `c2rust-ast-printer`
* `c2rust-build-paths`
* `c2rust-ast-exporter`
* `c2rust-bitfields-derive`
* `c2rust-bitfields`
* `c2rust-transpile`
* `c2rust`

or if this works, then I'll use this:

```
cargo +nightly publish --workspace --dry-run -Z package-workspace --exclude c2rust-analysis-rt --exclude c2rust-instrument --exclude c2rust-pdg --exclude c2rust-analyze
```

I need to use `cargo +stable` for this because the pinned nightly does not correctly resolve the `Cargo.lock` for a workspace, and thus tries to incompatibly update it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -180,7 +180,7 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "c2rust"
-version = "0.19.0"
+version = "0.20.0"
 dependencies = [
  "anyhow",
  "c2rust-build-paths",
@@ -195,7 +195,7 @@ dependencies = [
 
 [[package]]
 name = "c2rust-analysis-rt"
-version = "0.19.0"
+version = "0.20.0"
 dependencies = [
  "bincode",
  "crossbeam-queue",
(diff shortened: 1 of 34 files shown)


fw-immunant (approved):
LGTM. 🚀

PR #1206, opened 2025-03-13
docs/readme: Update `bear` arguments

javierhonduco:
The build command should be passed after `--`, see https://github.com/rizsotto/Bear?tab=readme-ov-file#how-to-use


javierhonduco:
Seems like some GitHub actions need to be updated

kkysen (approved):
Thanks!  The CI is broken (not sure if it still is), but a doc fix isn't going to affect any tests, so I'll merge this in.

PR #1214, opened 2025-04-05
build: support cmake 4.0

chenrui333:
bump `cmake_minimum_required` to support cmake 4.0 build

```
    CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
      Compatibility with CMake < 3.5 has been removed from CMake.
```


kkysen (approved):
Thanks!

PR #1218, opened 2025-04-28
transpile: use a repeat expression for arrays with a single zero initializer

folkertdev:
translate

```c
unsigned short data[DATA_SIZE] = {0};
```

to 

```rust
let mut data: [libc::c_ushort; 65535] = [0 as libc::c_int as libc::c_ushort; 65535];
```

instead of a literal with 65535 zeros.

I also simplified the logic. Although https://github.com/GaloisInc/C2Rust/issues/40 is a dead link I believe the second commit is correct.

### DIFF
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -180,6 +180,27 @@ impl<'c> Translation<'c> {
             CTypeKind::ConstantArray(ty, n) => {
                 // Convert all of the provided initializer values
 
+                let to_array_element = |id: CExprId| -> TranslationResult<_> {
+                    self.convert_expr(ctx.used(), id)?.result_map(|x| {
+                        // Array literals require all of their elements to be
+                        // the correct type; they will not use implicit casts to
+                        // change mut to const. This becomes a problem when an
+                        // array literal is used in a position where there is no
+                        // type information available to force its type to the
+                        // correct const or mut variation. To avoid this issue
+                        // we manually insert the otherwise elided casts in this
+                        // particular context.
+                        if let CExprKind::ImplicitCast(ty, _, CastKind::ConstCast, _, _) =
+                            self.ast_context[id].kind
+                        {
+                            let t = self.convert_type(ty.ctype)?;
+                            Ok(mk().cast_expr(x, t))
+                        } else {
+                            Ok(x)
+                        }
+                    })
+                };
+
                 // We need to handle the 4 cases in `str_init.c` with identical initializers:
                 // * `ptr_extra_braces`
                 // * `array_extra_braces`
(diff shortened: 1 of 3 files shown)


folkertdev:
Looks like you need to update your CI to use a newer ubuntu version (we usually use `ubuntu-latest`, but maybe there is a reason you pin a specific version?)

> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

kkysen:
> Looks like you need to update your CI to use a newer ubuntu version (we usually use `ubuntu-latest`, but maybe there is a reason you pin a specific version?)
> 
> > This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see [actions/runner-images#11101](https://github.com/actions/runner-images/issues/11101)

I forget exactly why but the latest Ubuntu might've broke something in CI (it did for rav1d recently).

folkertdev:
Ok well maybe `22.04` still works, or maybe latest got updated with a fix? (I can try that too)

I'm also not sure what's up with the azure runners, but it doesn't look like the changes in the PR would cause those errors:

https://dev.azure.com/immunant/c2rust/_build/results?buildId=4343&view=logs&j=f7d03aff-a499-5cb4-81d5-3cd2462ec94c&t=2246cd89-1657-535f-9a56-0b87376126d6&l=282

Maybe pinning `once_cell` to a slightly older version works there?

thedataking:
> I'm also not sure what's up with the azure runners

I am going to switch us off the Azure runners as soon as I get the time to do so. If you've tested locally and are convinced your change is not the cause of the breakage, that's good enough for me. Feel free to merge without CI passing. And thanks for contributing this improvement!

thedataking:
Quick note @folkertdev : I would like to avoid PR that changes the transpiler AND the CI. Those should be separate PRs. (Testing on this PR is fine though.)

folkertdev:
ah, I'll split that out then. I'd just like CI to check my work, not everything runs locally.

folkertdev:
I misinterpreted what C actually does earlier. The usage of

```c
int foo[] = { 0 }
```

is [extremely widespread](https://github.com/search?q=%22%7B+0+%7D%22++language%3AC&type=code), but at least with modern compilers it is equivalent in practice to

```c
int foo[] = { }
```

Anyway, that is not at all similar to `int foo[16] = { 1 }` , which does not repeat the `1` 16 times, but instead zeroes out the remainder of the array. 

So this has now been fixed to still detect the extremely common `{ 0 }`  scenario, but leave other incomplete initializers alone.

---

I ran test file locally and diffed the outputs. I'm not able to run any of the tests though, e.g. this fails

```
> ./scripts/test_translator.py --only-directories="arrays" tests
arrays:
 [ FAILED ] translate incomplete_arrays.c
incomplete_arrays.c:5:8: error: unknown type name 'uint32_t'
    5 | extern uint32_t SOME_INTS[];
      |        ^
warning: c2rust: Encountered unsupported BuiltinType kind 489 for type <dependent type>
```

That file does `#include <stdint.h>`, so I'm not sure what the problem is. Anyway,  in terms of testing that is kind of the best I can do I think?

kkysen:
@folkertdev, I merged your CI PR #1219 now, so you can rebase this now if you want and hopefully more CI works now.  I didn't want to rebase for your in case it messes anything up you might've been working on.

folkertdev:
Well I was playing with more complicated zero initializers (e.g. `struct {short; int y;} arr6[4] = { 0 };`, but I don't really know enough about C (or C++, and the AST you get from clang seems to use a bunch of C++ terminology) to be sure that the logic is correct.

So let's merge this when CI is happy, and then we'll see if the more complex case comes up.

bungcip:
@folkertdev some test like https://github.com/immunant/c2rust/blob/71c95a2e11b18887b48a2fdcc3fbb2cd4f34c98a/tests/items/src/test_fn_attrs.rs#L24 appear to test the output (basically just `include_str!` and check if string contains suitable output)

kkysen:
> Btw, wouldn't something like `cargo insta` (or some other snapshot testing solution) be really useful for testing the output of `c2rust-transpile`? I've basically been doing that manually. Unless I'm missing something, the current tests really only test that the `c2rust-transpile` does not panic (and maybe that it generates rust that compiles?). What it generates exactly does not appear to be tracked?

Yes, it definitely would.  I've used it in newer tests like for `c2rust-pdg` and it worked great, but we haven't had time to change our existing transpiler tests.  It would be super helpful to have the translated output checked in and reviewable.

tests/arrays/src/arrays.c
```diff
@@ -14,6 +14,16 @@ void entry(const unsigned buffer_size, int buffer[const])
...
 
+    int arr4[16] = {0};
+    arr4[15] += 9;
+
+    struct {short; int y;} arr5[1] = { { 1, 2 } };
+    arr5[0].y += 9;
+
+    // excess elements
+    int arr6[2] = { 1, 2, 3 };
+    int arr7[0] = { 1234 };
```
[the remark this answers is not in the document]
folkertdev: What actually runs the logic in `test_arrays.rs`? just a `cargo test` does not seem to do it, the functions in that file appear to never be actually called, and they are not marked `#[test]`.

c2rust-transpile/src/translator/literals.rs
```diff
@@ -173,61 +173,75 @@ impl<'c> Translation<'c> {
...
+                    )
+                };
+
+                match ids {
+                    [] => {
+                        // this was likely a C array of the form `int x[16] = {}`,
+                        // we'll emit that as [0; 16].
+                        let len = mk().lit_expr(mk().int_unsuffixed_lit(n as u128));
+                        Ok(self
+                            .implicit_default_expr(ty, ctx.is_static)?
```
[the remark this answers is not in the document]
folkertdev: The implementation of `self.implicit_default_expr` already uses `zero_initializer` for struct/union/enum.

c2rust-transpile/tests/snapshots/arrays.c
folkertdev, line 1: This is a copy of `arrays.c`, with some additional tests added at the bottom. I've left the original for now because it actually gets executed. I could remove it though if the snapshots are sufficient.

c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
```diff
@@ -0,0 +1,408 @@
...
+        {
+            let mut init = C2RustUnnamed_1 {
+                x: 1 as std::ffi::c_int as std::ffi::c_short,
+                y: 2 as std::ffi::c_int,
+            };
+            init
+        },
+        C2RustUnnamed_1 { x: 0, y: 0 },
+        C2RustUnnamed_1 { x: 0, y: 0 },
+    ];
```
folkertdev, line 128: in more recent versions this could be e.g.

```rust
let mut many_struct: [C2RustUnnamed_1; 3] = const { 
    let mut base = [C2RustUnnamed_1 { x: 0, y: 0 }; 3];
    base[0] = C2RustUnnamed_1 {
        x: 1 as std::ffi::c_int as std::ffi::c_short,
        y: 2 as std::ffi::c_int,
    };

    base
};
```

but idk, it's extra work and may not actually come up in practice.

folkertdev (commented):
I've rebased and added snapshot tests. Unless CI finds something I think this one is ready.

fw-immunant:
It would be nice to rebase and merge this now that #1265 has landed. I would suggest removing `buffer` from the snapshot test because it only exists for testing runtime behavior, and makes the snapshot much larger and uglier.

@kkysen: Since you recently touched initializer code, could you take a look? You should be able to push to [folkertdev:array-default-value](https://github.com/folkertdev/c2rust/tree/array-default-value).

c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
kkysen: Yeah, this is definitely cleaner (of course, like you, I'm not sure how worth it it is).  How recently was support for this added, though?  I know the `const {}` blocks are new, but you don't need that for this, right?

c2rust-transpile/tests/snapshots/arrays.c
kkysen: I moved things to add `arrays.c` and its snapshot in a separate commit so we can more easily see the diff that this PR creates in it.  Otherwise, it's all good.  I'll just leave the existing `arrays.c` for now.

kkysen:
> I would suggest removing `buffer` from the snapshot test because it only exists for testing runtime behavior, and makes the snapshot much larger and uglier.

`buffer` is useful for testing various array things, similar to some of the tests I added in `str_init.c`.  So while not needed for this PR, I would prefer to keep it in the snapshot.

c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
folkertdev: True, you just have to name the `const` item, but I guess if you nest it in a block the name can never conflict?

kkysen: Here in normal scope, nothing should need to be `const`.  In `const` scope (e.g. a `const` or `static` item), then it'll need to be `const`, but it shouldn't need the explicit `const {`, just a normal `{`, since it already knows to evaluate it in `const` mode.

kkysen (approved):
LGTM now.  @folkertdev, I made a few small changes/reorganizations, so if it still looks good to you, I'll merge it.

c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
folkertdev: Oh yeah, if the value is `Copy`, which it would be here, then `const` isn't even needed.

folkertdev:
Yes this looks good, thanks!

c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
kkysen: Ah, true, although everything from C should be `Copy`.

fw-immunant:
> > I would suggest removing `buffer` from the snapshot test because it only exists for testing runtime behavior, and makes the snapshot much larger and uglier.
> 
> `buffer` is useful for testing various array things, similar to some of the tests I added in `str_init.c`. So while not needed for this PR, I would prefer to keep it in the snapshot.

My point is that codegen of accesses to `buffer` doesn't exercise our translation of initializers; leaving it out would keep this test case more focused. But not a blocking concern.

kkysen:
> My point is that codegen of accesses to `buffer` doesn't exercise our translation of initializers; leaving it out would keep this test case more focused. But not a blocking concern.

Ah, I see what you mean.  Fixed in 102770d24.

PR #1212, opened 2025-04-03
GH Actions: bump tagged versions of actions/cache

brk:
Recent workflow runs have been failing because actions/cache@v2 was recently decommissioned. See
https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down for details.

PR #1219, opened 2025-04-29
CI updates (Ubuntu 22, LLVM 15, pinned `once_cell`)

folkertdev:
you should probably also just merge

- #1214
- #1212

pinning the `once_cell` version is a bit of a hack. It would not be needed if the rust version would update to `1.70.0`: https://doc.rust-lang.org/beta/std/cell/struct.OnceCell.html

The github CI action still fails to translate curl even with these changes, and it doesn't really give any error message. So I'm not sure what to do with that.

### DIFF
--- a/.github/workflows/internal-testsuite.yml
+++ b/.github/workflows/internal-testsuite.yml
@@ -79,7 +79,7 @@ jobs:
         sudo apt-get -qq update
         sudo apt-get -qq install    \
             libbrotli-dev           \
-            libclang-12-dev         \
+            libclang-15-dev         \
             libgcrypt20             \
             libreadline-dev         \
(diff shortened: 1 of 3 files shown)


kkysen (approved):
Thanks for helping fix this!

I merged #1212 and #1214, so I rebased this PR on top of those now.

For the `once-cell` thing, we'd love to update our nightly, but it's gotten increasingly difficult as it's gotten more and more out of date.  Best would probably be to split `c2rust-transpile` and its stable dependencies into a separate workspace from `c2rust-analyze` and its unstable dependencies, but this is a good enough fix for now.  I'll try to see and fix what else is still broken in CI after this.

Issue #1220, opened 2025-05-02
Split `c2rust-transpile` (and deps) and `c2rust-analyze` (and deps) into two separate workspaces


kkysen:
Right now, `c2rust-transpile` and its dependencies can all build on `stable`, while `c2rust-analyze` and its dependencies all require a pinned `nightly-2022-08-08`.  However, only one `rust-toolchain.toml` is allowed per directory (and thus workspace), so even the stable crates like `c2rust-transpile` are forced to be compiled with the old nightly.  This presents numerous problems that interfere with development and maintenance.

The majority of external contributions we get are for the transpiler, and they often get mired in this issue.  This includes both PRs we've received, as well as others I've discussed `c2rust` with in-person who wanted to use `c2rust transpile` and contribute, but the old nightly and the issues caused by it dissuaded them.

The same goes for ongoing maintenance.  We only release `c2rust-transpile` and its dependencies (and the simple `c2rust` binary), and the old nightly significantly interferes with releasing new `c2rust` versions and maintaining things as is.  For example, we can no longer `cargo publish`, as the pinned nightly's `cargo` has a bug (https://github.com/rust-lang/cargo/issues/11148), which has been fixed (https://github.com/rust-lang/cargo/pull/11477), but in `1.68.0`, so we can't use it.  Thus, we have to `cargo +stable publish -p ${stable-c2rust-crate}` (https://github.com/immunant/c2rust/pull/1116#issuecomment-2338813261).  We also can't `cargo update` (#1116), as that updates to newer versions that don't support our old nightly (which also presents frustrating problems in incrementally updating our nightly).  Newer `cargo` versions can use MSRV-aware resolution, but we're stuck on an old `cargo`.  Thus, `cargo install c2rust` is more fragile and needs `--locked`.  We also need to pin certain dependencies like `once_cell` (#1219) due to their interaction with the `rustc` version and `std`.

Upgrading from our increasingly ancient nightly also grows harder every day.  Keeping up with the latest nightly is difficult enough on its own when we are close behind, as each nightly can break internal APIs we use, but the farther we get behind, we also face issues in dependencies (`once_cell`, `syn`, and Rust tooling like `cargo publish`).  #811 is old work trying to update to a newer nightly, but it was never merged, and now there are more issues with it (merge conflicts, but also issues with dependencies not allowing incremental updates to the nightly version, and forcing larger all-at-once updates to the next Rust version).

[...]

PR #1225, opened 2025-05-07
Fixes to `rust-toolchain.toml` handling

joshtriplett:
This fixes a few aspects of the `rust-toolchain.toml` handling, to make it
easier to make future changes such as limiting the use of `rust-toolchain.toml`
to the tools that use rustc internals.

Fix a few obsolete references to `rust-toolchain` rather than
`rust-toolchain.toml`, since c2rust doesn't use the former.

Use the `rustfmt` component rather than `rustfmt-preview`, since the latter is
obsolete (rustfmt has not been a "preview" for a long time).

Finally, make `c2rust-transpile` use a separate file for the generated
`rust-toolchain.toml`. When c2rust-transpile translates C variadics, it
generates code using the (currently) nightly-only feature `c_variadics`, so it
emits a `rust-toolchain.toml` file to use nightly. Rather than copy this file
from the `rust-toolchain.toml` used to build c2rust itself, use a separate
file.
    
This also means the generated `rust-toolchain.toml` can stop requiring
toolchain components that the generated code doesn't need, such as rustc-dev.
(It still requires `rustfmt`, as c2rust may run `cargo fmt` on the generated
code if invoking `c2rust-refactor`.)


kkysen (approved):
Thanks!

kkysen:
> future changes such as limiting the use of `rust-toolchain.toml` to the tools that use rustc internals.

What do you mean by this?

kkysen:
> I mean the implementation of #1220 : attempting to build only the tools needing rustc internals with the nightly toolchain, and the rest with the stable toolchain.

Ahh.  I thought you were referring to some future `rustc`/`rustup` change, lol.  #1220/#1227 is totally good.

PR #1227, opened 2025-05-07
Split crates requiring nightly Rust into separate `unstable` workspace

joshtriplett:


scripts/docker_build.sh
```diff
@@ -26,10 +26,7 @@ IMAGES["fedora:34"]="1"
 build_image() {
     BASE_IMAGE=${1}
     IMAGE_TAG=$(echo ${BASE_IMAGE} | tr -s :/ - ) # replace colons and slashes with hyphens
-
-    # pull the rust version out of ../rust-toolchain.toml to keep things synced
-    RUST_TOOLCHAIN_FILE="$SCRIPT_DIR/../rust-toolchain.toml"
-    RUST_VER=$($SCRIPT_DIR/query_toml.py toolchain.channel $RUST_TOOLCHAIN_FILE)
+    RUST_VER=stable
```
kkysen, line 29: CI probably needs both stable and nightly since we run tests for both `c2rust-transpile` and the rustc internal dependent ones in CI.

Cargo.lock
kkysen, line 1: Can we keep the original lock file for now and update it separately?  Seems like this is what CI is failing on.  You did update CI to use stable, but we do still need nightly in CI in addition to stable to run the nightly-dependent tests.

kkysen:
Also I merged #1225 so you can rebase now.

Cargo.lock
joshtriplett: @kkysen I think what this is showing is that it isn't going to be possible to solve this with `exclude`. (Or at least, doing so would require also excluding a lot of CI, and I assume ) It'll take actually splitting into a separate workspace (with its own lockfile), and thus implementing all of #1220.

I'll work on that.

joshtriplett:
Currently attempting a full implementation of #1220. Still WIP; will work on ensuring CI passes.

joshtriplett:
Side note: `analysis/tests/minimal/reference_pdg.bc` was checked in despite being `.gitignore`d. It doesn't *look* like that test is using "compare" mode, so I don't think that file is being used, only written to. Nonetheless, I moved it across to `unstable/analysis/tests/minimal/reference_pdg.bc`. If there's any potential cleanup there, it can happen separately from this PR.

joshtriplett:
@kkysen For some reason, the Azure pipeline doesn't seem to be kicking off for pushes anymore, and I'm not sure why. I *think* I've fixed most of the issues at this point, but I can't test that until the Azure pipeline runs again.

kkysen:
> Side note: `analysis/tests/minimal/reference_pdg.bc` was checked in despite being `.gitignore`d. It doesn't _look_ like that test is using "compare" mode, so I don't think that file is being used, only written to. Nonetheless, I moved it across to `unstable/analysis/tests/minimal/reference_pdg.bc`. If there's any potential cleanup there, it can happen separately from this PR.

It's used in a test here: https://github.com/immunant/c2rust/blob/4b7691e44578859b7739a46ff5dceb165d131272/c2rust-analyze/tests/analyze.rs#L74
As long as you keep the relative paths the same, it should still work fine.  We in general gitignored `*.bc` files but added this one manually for this one test.

joshtriplett:
(Apologies that this change stalled out. Some other things came up. I'll get back to it eventually, but someone else who wants to work on it should not hesitate to pick it up and run with it.)

kkysen (commented):
> (Apologies that this change stalled out. Some other things came up. I'll get back to it eventually, but someone else who wants to work on it should not hesitate to pick it up and run with it.)

Thanks again for working on this before!  We may pick it up, as after we moved off of Azurue in #1275, @fw-immunant noticed that this can pass CI with some minimal extra changes.  It does certainly help make many things easier, and allows us to implement things like `__builtin_assume` (https://github.com/immunant/c2rust/pull/1132#discussion_r2230807333).  However, we're also resurrecting `c2rust-refactor`, which was an even older nightly, and being upgraded to the currently pinned nightly, so I'm not sure if that will interact in any way.  It's probably still a good idea, though, as long as we're still able to have `c2rust-transpile` target the pinned nightly in its output (like the tests are made to do in this PR).

kkysen (commented):
> (Apologies that this change stalled out. Some other things came up. I'll get back to it eventually, but someone else who wants to work on it should not hesitate to pick it up and run with it.)

Thanks again for working on this before!  We may pick it up, as after we moved off of Azurue in #1275, @fw-immunant noticed that this can pass CI with some minimal extra changes.  It does certainly help make many things easier, and allows us to implement things like `__builtin_assume` (https://github.com/immunant/c2rust/pull/1132#discussion_r2230807333).  However, we're also resurrecting `c2rust-refactor`, which was an even older nightly, and being upgraded to the currently pinned nightly, so I'm not sure if that will interact in any way.  It's probably still a good idea, though, as long as we're still able to have `c2rust-transpile` target the pinned nightly in its output (like the tests are made to do in this PR).

ahomescu:
`provision_mac.sh` is out of date, do we want to fix it?

fw-immunant:
> `provision_mac.sh` is out of date, do we want to fix it?

We don't use it in CI, and I'm not sure it's used by any developers either. But it's also a short, simple script and shouldn't be hard to update.

ahomescu:
We discussed this in an Immunant meeting and we're not super comfortable with changing the top-level repository structure this much. I'll try to add a `rust-toolchain.toml` file in the transpile subdirectory instead.

Issue #1226, opened 2025-05-07
Move CI from Azure Pipelines to GitHub Actions


joshtriplett:
https://github.com/immunant/c2rust/issues/477#issue-1289711171 mentions moving CI to GitHub Actions instead of Azure Pipelines. I'm filing this issue to ask: is doing that still the plan, and would PRs doing that be welcome?

kkysen:
Yup, that's still the plan, we just have had time to get around to do it yet.  PRs would definitely be welcome.  Thanks for all the help!

PR #1232, opened 2025-05-08
Move Linux CI from Azure Pipelines to GitHub Actions

joshtriplett:
This simplifies the CI setup substantially. In particular, this installs packages as part of the CI setup, rather than in a separate manual container build step where modifications by contributors are ineffective until the container images are regenerated.

Now, changes to provisioning will take immediate effect, allowing upgrades to new versions to be contributed and tested easily.

This does not move macOS over, because I don't have sufficient experience with macOS to easily debug the issues that occur when attempting to do so.

* Fixes Linux part of #1226.


joshtriplett:
@kkysen Would you mind safelisting `Swatinem/rust-cache` for GitHub Actions?

kkysen:
> @kkysen Would you mind safelisting `Swatinem/rust-cache` for GitHub Actions?

Do you know how to do that?

joshtriplett:
> > @kkysen Would you mind safelisting `Swatinem/rust-cache` for GitHub Actions?
> 
> Do you know how to do that?

I believe it'll be in the GitHub actions settings, either for the project or for the organization. There's a setting for where it can run actions from.

kkysen:
> > > @kkysen Would you mind safelisting `Swatinem/rust-cache` for GitHub Actions?
> > 
> > 
> > Do you know how to do that?
> 
> I believe it'll be in the GitHub actions settings, either for the project or for the organization. There's a setting for where it can run actions from.

@thedataking, can you see if you're able to do this?  I don't see where I can, or I probably don't have permission to do it.  See the error message here: https://github.com/immunant/c2rust/actions/runs/14911291371.

azure-pipelines.yml
```diff
@@ -1,120 +0,0 @@
...
-      debian10:
-        containerImage: immunant/c2rust:debian-buster-latest
-      debian11:
-        containerImage: immunant/c2rust:debian-bullseye-latest
-      fedora34:
-        containerImage: immunant/c2rust:fedora-34-latest
-      ubuntu20:
-        containerImage: immunant/c2rust:ubuntu-focal-latest
-      ubuntu18:
-        containerImage: immunant/c2rust:ubuntu-bionic-latest
```
kkysen, line 34: Can we test on all of these distros and versions in GitHub Actions, too, instead of just Ubuntu 24?

kkysen: Oh wait, GitHub Actions only supports Ubuntu, right?  So we'll have to drop testing on all the others.  @thedataking is this okay?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  test-linux:
+    runs-on: ubuntu-24.04
```
kkysen, line 11: I'm wondering if we should just use `ubuntu-latest`.  We often get people finding issues with updated distros with newer LLVM/clang versions, and it'd be nicer to catch that in CI rather than having users need to open an issue, even if CI might suddenly break.

scripts/provision_mac.sh
kkysen, line 1: Can we fully inline this in `ci.yml`?  And the other provision scripts, too.

kkysen (changes requested):

.github/workflows/ci.yml
joshtriplett: @kkysen Sure, I can do that.

scripts/provision_mac.sh
joshtriplett: I'd be happy to do that as an additional step, once this builds at all on macOS. Would you mind if that's a follow-up PR?

azure-pipelines.yml
joshtriplett: @kkysen It's theoretically possible to run a custom container of an arbitrary distro on GitHub Actions, but one of the huge benefits of this change to the CI is avoiding the separate (and manual) step of building containers, which makes it impossible for contributors to modify and test (because contributors can't replace those containers). With the new approach, all the packages are installed on the fly, and it's much easier to iterate on the result.

I think that usability improvement would be worth making this change.

scripts/provision_mac.sh
kkysen: Of course

joshtriplett:
@kkysen I think this is ready to merge now.

scripts/provision_mac.sh
```diff
@@ -23,8 +23,8 @@ export HOMEBREW_NO_AUTO_UPDATE=1
 brew install -q python cmake ninja gpg llvm@17 bash
 
 # Python 3 packages
-python3 -m pip install --user --upgrade pip
-python3 -m pip install --user -r "$SCRIPT_DIR/requirements.txt"
+python3 -m pip install --break-system-packages --user --upgrade pip
+python3 -m pip install --break-system-packages --user -r "$SCRIPT_DIR/requirements.txt"
```
kkysen, line 27: Are these still needed if this PR is just moving Linux CI now?

kkysen (approved):
LGTM but I'd still like @thedataking to take a final look, particularly about moving from testing on multiple distros to just Ubuntu.  I think it's likely fine, and GitHub Actions have numerous benefits, but I just want to double check.

scripts/provision_mac.sh
joshtriplett: No, though they'll be needed by anyone trying to port to newer macOS with newer pip. But I should remove them from this PR, good catch.

thedataking: If this PR doesn't move macOS CI over, it shouldn't touch mac-only files. That can be considered when re-enabling macOS.

kkysen: Yeah, they're definitely okay in general for another PR.  Although I'd personally prefer if we move all of our python dependency management to `uv` to make things simpler and more reliable.

kkysen:
Also I'm re-running CI to see how long a cached CI takes, although that should improve a bunch with `rust-cache`.

thedataking:
@joshtriplett appreciate the work to fix our CI. Thanks! 

> Would you mind safelisting Swatinem/rust-cache for GitHub Actions?

I'd prefer to use the "vanilla" cache module unless there is some clear benefit. Wouldn't the pattern below work?

```
      - name: cache rust crates
        uses: actions/cache@v4
        with:
          path: |
            ~/.cargo/bin/
            ~/.cargo/registry/index/
            ~/.cargo/registry/cache/
            ~/.cargo/git/db/
            target/
          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```

kkysen:
> * Fixes Linux part of #1226.

I updated this.  We still have macOS to do, right?

joshtriplett:
> @joshtriplett appreciate the work to fix our CI. Thanks!
> 
> > Would you mind safelisting Swatinem/rust-cache for GitHub Actions?
> 
> I'd prefer to use the "vanilla" cache module unless there is some clear benefit. Wouldn't the pattern below work?
> 
> ```
>       - name: cache rust crates
>         uses: actions/cache@v4
>         with:
>           path: |
>             ~/.cargo/bin/
>             ~/.cargo/registry/index/
>             ~/.cargo/registry/cache/
>             ~/.cargo/git/db/
>             target/
>           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
> ```

Caching correctly is a lot more complicated than that, and also involves caching more than that. The cache needs to depend on the toolchain, the job, etc. And the cache should include the results of compiling the dependencies, so that most builds are relatively fast. Doing it right requires logic on par with what rust-cache already does.

If you want to lock it down to a specific version of rust-cache, we could do that, and be very deliberate about upgrading it.

thedataking:
Thanks for educating me. Khyber mentioned that rust-lang also uses this action which eases my concern somewhat.

> If you want to lock it down to a specific version of rust-cache, we could do that, and be very deliberate about upgrading it.

OK, v2 should be whitelisted now. Could we have a comment in the workflow to remind us why we're not using Github's cache action?

kkysen (approved):
All good now, thank you!!

PR #1235, opened 2025-05-08
Support `__float128`, which shows up in clang headers on some systems

joshtriplett:
Without this, translation can fail with:

`warning: c2rust: Encountered unsupported BuiltinType kind 478 for type __float128`

followed by:

`thread 'main' panicked at 'Type conversion not implemented for TagTypeUnknown expecting 3', c2rust-transpile/src/c_ast/conversion.rs:827:22`

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1687,6 +1687,8 @@ pub enum CTypeKind {
     // ARM Scalable Vector Extension types
     // TODO: represent all the individual types in AArch64SVEACLETypes.def
     UnhandledSveType,
+
+    Float128,
 }
 
 impl CTypeKind {
@@ -1713,6 +1715,7 @@ impl CTypeKind {
             UInt128 => "unsigned __int128",
             Half => "half",
             BFloat16 => "bfloat16",
+            Float128 => "__float128",
             _ => unimplemented!("Printer::print_type({:?})", self),
         }
     }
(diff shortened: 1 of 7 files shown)


joshtriplett:
This fixes *all* the test failures I get when trying to test c2rust locally on latest Debian.

c2rust-transpile/src/convert_type.rs
```diff
@@ -330,6 +330,7 @@ impl TypeConverter {
             CTypeKind::Int128 => Ok(mk().path_ty(mk().path(vec!["i128"]))),
             CTypeKind::UInt128 => Ok(mk().path_ty(mk().path(vec!["u128"]))),
             CTypeKind::BFloat16 => Ok(mk().path_ty(mk().path(vec!["bf16"]))),
+            CTypeKind::Float128 => Ok(mk().path_ty(mk().path(vec!["f128", "f128"]))),
```
kkysen, line 333: Can you combine this with the above `CTypeKind::LongDouble` arm?

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1713,6 +1715,7 @@ impl CTypeKind {
             UInt128 => "unsigned __int128",
             Half => "half",
             BFloat16 => "bfloat16",
+            Float128 => "__float128",
```
kkysen, line 1718: This is probably just used for debug output so it's okay, but it's possible we might conflate this with `_Float128` (which we don't handle yet, but should).  That said, `__float128` and `_Float128` are both IEEE 754 binary128, so us having one `CTypeKind::Float128` type is probably good enough.

kkysen (approved):
Just one little nit but otherwise it LGTM.  Thanks for fixing this, too!

c2rust-transpile/src/convert_type.rs
joshtriplett: @kkysen I can. The main reason I hadn't was to allow for the theoretical possibility of handling targets where C's `long double` is not `f128`. However, it does seem unlikely that those would ever be supported, and if they *are* supported, we can always re-split the arms. Fixing.

c2rust-transpile/src/c_ast/mod.rs
joshtriplett: I do think it makes sense to map them both to the same underlying type, yes.

kkysen (approved):

PR #1241, opened 2025-05-09
Allow running multiple transpiles in parallel

jameysharp:
If c2rust-transpile needs to generate a temporary compile_commands.json, previously it used a fixed path for it across all invocations. Instead, create a uniquely-named temporary directory for it, so that parallel transpiles don't clobber each other.

I've used a two-year-old release of tempfile to avoid adding a bunch of duplicate versions of other crates to Cargo.lock, but presumably some upgrades ought to happen sooner or later.

### DIFF
--- a/c2rust/src/bin/c2rust-transpile.rs
+++ b/c2rust/src/bin/c2rust-transpile.rs
@@ -1,7 +1,7 @@
 use clap::{Parser, ValueEnum};
 use log::LevelFilter;
 use regex::Regex;
-use std::{fs, path::PathBuf};
+use std::{ffi::OsStr, fs, path::PathBuf};
 
 use c2rust_transpile::{Diagnostic, ReplaceMode, TranspilerConfig};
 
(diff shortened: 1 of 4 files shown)


c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -226,28 +226,29 @@ fn main() {
         tcfg.emit_modules = true
     };
 
-    let mut created_temp_compile_commands = false;
+    let mut temp_compile_commands = None;
```
kkysen, line 229: Could you add `dir` to this name?

kkysen (changes requested):
Thanks for fixing this!  Some small changes preferred, but the tempdir part looks good.

kkysen:
> I've used a two-year-old release of tempfile to avoid adding a bunch of duplicate versions of other crates to Cargo.lock, but presumably some upgrades ought to happen sooner or later.

This is fine.  We're (@joshtriplett is) in the process of splitting out the unstable crates (#1227, #1220) so that we can freely update our dependencies for the stable crates like `c2rust-transpile`, among other benefits, so we should be able to upgrade dependencies like `tempfile` soon enough.

kkysen (approved):

jameysharp:
Thanks for reviewing! I've made the changes you requested. I don't understand why you prefer buffering the entire temporary `compile_commands.json` in memory instead of using a fixed-size buffer, but that's okay, I've reverted that change as you asked. How does this look to you now?

I tried to understand why the c2rust-testsuite CI task failed last time but it didn't look like it had anything to do with my changes, I think. Is that an existing issue?

By the way, I was looking at this while trying to figure out how to introduce snapshot testing for c2rust-transpile, because I found that approach super helpful when working on Wasmtime. Now I see you've been discussing that recently in https://github.com/immunant/c2rust/pull/1218#issuecomment-2845890078 so I'm glad to see that might be welcome.

kkysen:
> Thanks for reviewing! I've made the changes you requested. I don't understand why you prefer buffering the entire temporary `compile_commands.json` in memory instead of using a fixed-size buffer, but that's okay, I've reverted that change as you asked. How does this look to you now?

Thanks!  Looks all good.  Buffering it in memory and writing in one go is a bit simpler and since the compile commands will be very small, it's not really gonna make a difference.  It'll probably be marginally faster, but that's not really very important here; it's moreso that it's just a bit simpler.

> I tried to understand why the c2rust-testsuite CI task failed last time but it didn't look like it had anything to do with my changes, I think. Is that an existing issue?

Yeah, that's an existing issue that's been failing on curl for a little while.  We're still trying to figure that one out, but I haven't had too much time to look into it more lately.

> By the way, I was looking at this while trying to figure out how to introduce snapshot testing for c2rust-transpile, because I found that approach super helpful when working on Wasmtime. Now I see you've been discussing that recently in https://github.com/immunant/c2rust/pull/1218#issuecomment-2845890078 so I'm glad to see that might be welcome.

That'd definitely be welcome!  We have a few `insta` snapshot tests for some newer tests, but if you'd like to help migrate the older transpiler tests, that'd be super helpful.

PR #1243, opened 2025-05-12
Bump bindgen to be able to handle CXCursor_LinkageSpec in LLVM18+

spastorino:
Current master branch gets stuck trying to compile `c2rust-ast-exporter` in Archlinux.
The problem is that the bindgen version used by c2rust doesn't handle the new Clang 18+'s `CXCursor_LinkageSpec` reported by `clang::getCursorKindForDecl`.

https://github.com/rust-lang/rust-bindgen/pull/2689 fixes the issue, so I've bumped the bindgen version to the minimum that works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,24 +136,24 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.65.1"
+version = "0.69.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfdf7b466f9a4903edc73f95d6d2bcd5baf8ae620638762244d3f60143643cc5"
+checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
 dependencies = [
- "bitflags",
+ "bitflags 2.9.0",
  "cexpr",
  "clang-sys",
+ "itertools",
  "lazy_static",
  "lazycell",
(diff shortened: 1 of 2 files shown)


kkysen (commented):
Thanks!

Although this is causing a new CI error here: https://github.com/immunant/c2rust/pull/1243/checks#step:11:75.

The `curl` test in the `c2rust testsuite` CI job has been failing for a while, but the compile error in `zstd` here is new.

The `thread 'main' panicked at 'unsupported proc macro punctuation character ' '', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.95/src/lib.rs:813:13` error is usually from `proc-macro2`/`syn` being incompatibly changed in `Cargo.lock`, so I'm guessing that's the issue?  Otherwise the change should be all good, and we should also be able to fix the punctuation character issue independently.

kkysen:
#1244 fixes that CI issue with `proc-macro2` `1.0.95`, so after that merges, you should be able to rebase this on top of that and then I can merge this.

kkysen:
#1244 is merged now, so you can go ahead.

spastorino:
I'm just checking with my cellphone but should I do anything? as you didn't update proc_macro2 in your PR. So this should merge clean?.
We need to re-run CI.

kkysen (approved):

PR #1244, opened 2025-05-12
transpile: remove invalid `Punct::new(' ')` in `fn convert_asm` and update `proc-macro2`

kkysen:
`' '` is not a valid punctuation, and newer versions of `proc_macro2` don't allow it.  However, we don't seem to need the space here at all, as `zstd` (from `c2rust-testsuite`) still transpiles and compiles successfully without it.

This also updates to `proc-macro` `1.0.95` (from `1.0.86`) to test this.

#1243 should be able to be merged after this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1180,9 +1180,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.86"
+version = "1.0.95"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e719e8df665df0d1c8fbfd238015744736151d4445ec0836b8e628aae103b77"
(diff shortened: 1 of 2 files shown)


fw-immunant (approved):
LGTM, compare the output of this sample:
```rust
fn main() {
    use proc_macro2::{TokenStream, TokenTree, Punct, Spacing::*};

    let mut tokens = vec![];
    tokens.push(TokenTree::Punct(Punct::new(',', Alone)).into());
    let result = quote::quote!(out(foo));
    tokens.push(result);
    //tokens.push(TokenTree::Punct(Punct::new(' ', Alone)).into());
    tokens.push(quote::quote!(_));
    let s: TokenStream = FromIterator::from_iter(tokens.into_iter());
    println!("{}", s);
}
```

```
, out (foo) _
```

PR #1246, opened 2025-05-14
Introduce some snapshot tests of c2rust-transpile

jameysharp:
Here's a quick implementation of snapshot testing for c2rust-transpile. I'm open to bike-shedding, such as discussing how the test files are laid out, or any other requests. I thought I'd just put together a quick proof of concept to start discussion with.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,15 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
+version = "0.7.20"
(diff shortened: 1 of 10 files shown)


kkysen (commented):
Thanks!  This looks great as a proof of concept.  I do think we'd like to keep the existing tests and their structure, though, if you'd like to work on that.  Also, sorry I took a little while to review this.  It's very welcome.

jameysharp:
No worries! I have my own distractions to deal with anyway.

When you say you'd "like to keep the existing tests and their structure", what do you mean by that? I guess if you could lay out more concretely what kind of structure you have in mind for these tests, that would help me.

Issue #1242, opened 2025-05-12
Not compiling with LLVM 20.1.0 or later


shirok1:
See https://github.com/llvm/llvm-project/pull/122289

```
  Cloning into 'tinycbor_build'...
  HEAD is now at d393c16 Parser: fix reading it->extra on big endian when bytesNeeded == 1
  /Users/shiroki/.cargo/git/checkouts/c2rust-2471865f8518dc65/0757fe4/c2rust-ast-exporter/src/AstExporter.cpp:2217:61: error: too many arguments to function call, expected 0, have 1
   2217 |                                  array, D->getBitWidthValue(*this->Context));
        |                                         ~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~~~~
  /opt/homebrew/Cellar/llvm/20.1.4_1/include/clang/AST/Decl.h:3166:12: note: 'getBitWidthValue' declared here
   3166 |   unsigned getBitWidthValue() const;
        |            ^
  /Users/shiroki/.cargo/git/checkouts/c2rust-2471865f8518dc65/0757fe4/c2rust-ast-exporter/src/AstExporter.cpp:2309:25: warning: enumeration value 'Binary' not handled in switch [-Wswitch]
   2309 |             switch (SL->getKind()) {
        |                     ~~~~^~~~~~~~~
  /Users/shiroki/.cargo/git/checkouts/c2rust-2471865f8518dc65/0757fe4/c2rust-ast-exporter/src/AstExporter.cpp:2309:25: note: add missing switch cases
   2309 |             switch (SL->getKind()) {
        |                         ^
  1 warning and 1 error generated.
  gmake[3]: *** [CMakeFiles/clangAstExporter.dir/build.make:79: CMakeFiles/clangAstExporter.dir/AstExporter.cpp.o] Error 1
  gmake[2]: *** [CMakeFiles/Makefile2:421: CMakeFiles/clangAstExporter.dir/all] Error 2
  gmake[1]: *** [CMakeFiles/Makefile2:428: CMakeFiles/clangAstExporter.dir/rule] Error 2
  gmake: *** [Makefile:254: clangAstExporter] Error 2
```

Simply removing `*this->Context` will make the compile success.

shirok1:
I think we can follow https://github.com/iovisor/bcc/pull/5194 and use a macro here

PR #1248, opened 2025-05-18
Fix getBitWidthValue compilation error on LLVM 20+

shirok1:
Introduced by upstream in https://github.com/llvm/llvm-project/pull/122289

Ref: https://github.com/iovisor/bcc/pull/5194

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2213,8 +2213,12 @@ class TranslateASTVisitor final
 
                          // 2. Encode bitfield width if any
                          if (D->isBitField()) {
+#if LLVM_VERSION_MAJOR >= 20
+                             cbor_encode_uint(array, D->getBitWidthValue());
+#else
                              cbor_encode_uint(
                                  array, D->getBitWidthValue(*this->Context));
+#endif
                          } else {
                              cbor_encode_null(array);
                          };


shirok1:
Closes #1242

shirok1:
Ping @kkysen for approving workflows

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2213,8 +2213,12 @@ class TranslateASTVisitor final
 
                          // 2. Encode bitfield width if any
                          if (D->isBitField()) {
+#if LLVM_VERSION_MAJOR >= 20
+                             cbor_encode_uint(array, D->getBitWidthValue());
```
kkysen, line 2217: One little nit: could you use a bit width variable inside the `#if` and then a single `cbor_encode_uint` call outside of it?

kkysen (commented):
Thanks! This looks great and simple, just one small style nit.

c2rust-ast-exporter/src/AstExporter.cpp
shirok1: Sorry but I'm not familiar with code here and not sure about how to name this variable (`bitWidth` is already occupied in this scope)

kkysen: Okay, I did it here on your PR (bc656a7).

kkysen (approved):

PR #1257, opened 2025-06-30
`transpile`: Support unaligned reads and writes

folkertdev:
This is very work-in-progress, but I think I need some pointers here. I'll add some inline comments.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -1641,10 +1641,29 @@ class TranslateASTVisitor final
 
     bool VisitUnaryOperator(UnaryOperator *UO) {
         std::vector<void *> childIds = {UO->getSubExpr()};
-        encode_entry(UO, TagUnaryOperator, childIds, [UO](CborEncoder *array) {
+
+        encode_entry(UO, TagUnaryOperator, childIds, [this, UO](CborEncoder *array) {
             cbor_encode_string(array, UO->getOpcodeStr(UO->getOpcode()).str());
             cbor_encode_boolean(array, UO->isPrefix());
+
+            if (UO->getOpcode() == UO_Deref) {
+                QualType eltTy = UO->getSubExpr()->getType()->getPointeeType();
+                CharUnits align = Context->getTypeAlignInChars(eltTy);
+
+                if (auto *TD = eltTy->getAs<TypedefType>()) {
+                    QualType naturalTy = TD->getDecl()->getUnderlyingType();
+                    CharUnits naturalAlign = Context->getPreferredTypeAlignInChars(naturalTy);
+
+                    if (align < naturalAlign) {
+                        cbor_encode_int(array, align.getQuantity());
+                    } else {
+                        cbor_encode_int(array, -1);
+                    }
+
+                }
+            }
         });
+
         return true;
     }
 
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1230,20 +1230,20 @@ pub enum CastKind {
 /// Represents a unary operator in C (6.5.3 Unary operators) and GNU C extensions
 #[derive(Debug, Clone, Copy)]
 pub enum UnOp {
-    AddressOf,     // &x
-    Deref,         // *x
-    Plus,          // +x
-    PostIncrement, // x++
-    PreIncrement,  // ++x
-    Negate,        // -x
-    PostDecrement, // x--
-    PreDecrement,  // --x
-    Complement,    // ~x
-    Not,           // !x
-    Real,          // [GNU C] __real x
-    Imag,          // [GNU C] __imag x
-    Extension,     // [GNU C] __extension__ x
-    Coawait,       // [C++ Coroutines] co_await x
+    AddressOf,                 // &x
+    Deref { unaligned: bool }, // *x
+    Plus,                      // +x
+    PostIncrement,             // x++
+    PreIncrement,              // ++x
+    Negate,                    // -x
+    PostDecrement,             // x--
+    PreDecrement,              // --x
+    Complement,                // ~x
+    Not,                       // !x
+    Real,                      // [GNU C] __real x
+    Imag,                      // [GNU C] __imag x
+    Extension,                 // [GNU C] __extension__ x
+    Coawait,                   // [C++ Coroutines] co_await x
 }
 
 impl UnOp {
@@ -1251,7 +1251,7 @@ impl UnOp {
         use UnOp::*;
         match self {
             AddressOf => "&",
-            Deref => "*",
+            Deref { .. } => "*",
             Plus => "+",
             PreIncrement => "++",
             PostIncrement => "++",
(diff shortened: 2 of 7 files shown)


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -1641,10 +1641,29 @@ class TranslateASTVisitor final
...
+                    if (align < naturalAlign) {
+                        cbor_encode_int(array, align.getQuantity());
+                    } else {
+                        cbor_encode_int(array, -1);
+                    }
+
+                }
+            }
         });
+
```
folkertdev, line 1666: I have LLM'd my way to a solution here, this is probably terrible c++ code.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -903,7 +918,23 @@ impl<'c> Translation<'c> {
...
+                                    let mut val = 
+                                        mk().unary_expr(UnOp::Deref(Default::default()), val);
+
+                                    // If the type on the other side of the pointer we are dereferencing is volatile and
+                                    // this whole expression is not an LValue, we should make this a volatile read
+                                    if lrvalue.is_rvalue() && cqual_type.qualifiers.is_volatile
+                                    {
+                                        val = self.volatile_read(val, cqual_type)?
+                                    }
+                                    Ok(val)
```
folkertdev, line 935: I'm confused by the value of `lrvalue`. I'd expect `ptr` to be an `lvalue` in e.g. `*ptr = value`, and an `rvalue` in the case `return *ptr`, but both are apparently `lvalue`s. Hence, I don't see how to differentiate between reads and writes here.

c2rust-transpile/tests/snapshots/snapshots__transpile@aligned_read_write.c.snap
```diff
@@ -0,0 +1,39 @@
...
+    unused_assignments,
+    unused_mut
+)]
+pub type __uint32_t = std::ffi::c_uint;
+pub type uint32_t = __uint32_t;
+pub type unaligned_uint32 = uint32_t;
+#[no_mangle]
+pub unsafe extern "C" fn unaligned_read(mut ptr: *const std::ffi::c_void) -> uint32_t {
+    return *(ptr as *const unaligned_uint32);
+}
```
folkertdev, line 21: so this is still wrong. I can make this one work, but then the unaligned write below is no longer correct.

c2rust-transpile/src/translator/operators.rs
folkertdev: adding to my confusion, this will not actually emit a `read_volatile`. Maybe I misunderstand what the attribute does here but I think that's a bug?

```c
uint32_t volatile_read(volatile void* ptr) {
    return *((volatile unaligned_uint32*)ptr);
}
```

folkertdev: hmm yeah this just seems wrong https://c.godbolt.org/z/sPcMM4b1T. The `lrvalue` is always `LValue` (I think a pointer dereference is just always an lvalue currently?), so that volatile read is never emitted.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -1641,10 +1641,29 @@ class TranslateASTVisitor final
...
+                if (auto *TD = eltTy->getAs<TypedefType>()) {
+                    QualType naturalTy = TD->getDecl()->getUnderlyingType();
+                    CharUnits naturalAlign = Context->getPreferredTypeAlignInChars(naturalTy);
+
+                    if (align < naturalAlign) {
+                        cbor_encode_int(array, align.getQuantity());
+                    } else {
+                        cbor_encode_int(array, -1);
+                    }
```
kkysen, line 1661: Overalignment can be useful to handle as well, so maybe just `cbor_encode_uint` both the actual and natural alignment?

c2rust-ast-exporter/src/AstExporter.cpp
kkysen: I'm not super familiar with libclang either.  Looks like it might work?  I can look into it more later.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1230,28 +1230,28 @@ pub enum CastKind {
...
+    PreIncrement,              // ++x
+    Negate,                    // -x
+    PostDecrement,             // x--
+    PreDecrement,              // --x
+    Complement,                // ~x
+    Not,                       // !x
+    Real,                      // [GNU C] __real x
+    Imag,                      // [GNU C] __imag x
+    Extension,                 // [GNU C] __extension__ x
+    Coawait,                   // [C++ Coroutines] co_await x
```
kkysen, line 1246: The formatting here is annoying 😩.  Could you add another commit that just makes all of the same-line `//` comments into proper doc comments above the variant with a newline after each one?  Or I could if you prefer.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -822,7 +837,7 @@ impl<'c> Translation<'c> {
 
                 match arg_kind {
                     // C99 6.5.3.2 para 4
-                    CExprKind::Unary(_, c_ast::UnOp::Deref, target, _) => {
+                    CExprKind::Unary(_, c_ast::UnOp::Deref { unaligned: _ }, target, _) => {
```
kkysen, line 840:
```suggestion
                    CExprKind::Unary(_, c_ast::UnOp::Deref { .. }, target, _) => {
```

If we don't care about the alignment, could it just be this?  Like you did for the other `match`.

c2rust-transpile/src/translator/operators.rs
kkysen: Hmm, that's a bug, right?  I can open an issue for it.

kkysen: Is the volatile code here just for trying to figure out how to emit `read_aligned` correctly?

kkysen: Oh, nevermind, it's already existing code.

kkysen: I think in `return *ptr`, both `ptr` and `*ptr` are lvalues in C.  See https://godbolt.org/z/bxrs7dczs.  `return *ptr` is the rvalue.  So the rvalue part of it comes not from the dereference, but from what's done with the dereference, if that makes sense.  Like in `*p + 1`, the whole `*p + 1` expression is an rvalue, but the `*p` itself is still an lvalue because `*p` is writable as well, i.e., `*p = 2` is well-formed while `*p + 1 = 2` is not.

folkertdev: Right, that is what makes this code confusing: I don't think it can ever be an `rvalue` in a way that is relevant, so that branch is unreachable.

Also in general that means we don't have great way of distinguishing whether a `*` means a read, write, or both. That is fine when a C `*` is translated with a rust `*`, but when you want to use the methods, it's a problem.

folkertdev: also there are already a bunch of open issues wrt volatile reads and writes. They just don't actually currently work as far as I can tell.

PR #1259, opened 2025-06-30
`transpile`: impl `__builtin_rotateright*` as `core::intrinsics::rotate_right`

### LINKED ISSUE #1251: Unimplemented builtin __builtin_rotateright32
am using c2rust transpile [libtomcrypt](https://github.com/libtom/libtomcrypt), but c2rust report that:

```
Transpiling rc6.c
error: Failed to translate rc6_ecb_decrypt: Unimplemented builtin __builtin_rotateright32
--> /home/xxx/libtomcrypt/src/ciphers/rc6.c:197:8
--> /home/xxx/libtomcrypt/src/ciphers/rc6.c:197:8
--> /home/xxx/libtomcrypt/src/ciphers/rc6.c:197:16
```

kkysen:
* Fixes #1251.


fw-immunant:
If we had separate `#[test] fn transpile_linux_only()` and `#[test] fn transpile_portable()`, would that avoid the `insta` glob conflict while letting us keep macOS coverage for the other tests?

kkysen:
> The transpiler changes look good, but I'd really like to avoid decreasing test coverage if at all possible.

If we can get cross compiling in tests working well, which I can work on, then we can, but otherwise it'll be difficult to generate the macOS snapshots locally not on macOS.

fw-immunant:
> > If we had separate `#[test] fn transpile_linux_only()` and `#[test] fn transpile_portable()`, would that avoid the `insta` glob conflict while letting us keep macOS coverage for the other tests?
> 
> Nope, that's what created the glob conflict. Insta removes the common prefix of all files per glob, so they end up conflicting that way. That's what I tried to do initially as it's simpler, but it doesn't work.

What's the scope of these glob conflicts (test, test suite, etc.)? Are there temporary files involved? This is weird and I don't think I understand what restrictions we're working under for adding/verifying snapshot tests. I didn't see anything in the insta docs about this--is there something to read?

fw-immunant:
> > The transpiler changes look good, but I'd really like to avoid decreasing test coverage if at all possible.
> 
> If we can get cross compiling in tests working well, which I can work on, then we can, but otherwise it'll be difficult to generate the macOS snapshots locally not on macOS.

Can we generate the snapshots from the macOS CI machine (as something we could copy-paste from output of a failing run), or is that not how insta works?

kkysen:
> Can we generate the snapshots from the macOS CI machine (as something we could copy-paste from output of a failing run), or is that not how insta works?

Yeah, we definitely could, but we'd have to do so every time a test changes, which would make development a lot more painful.

kkysen:
> What's the scope of these glob conflicts (test, test suite, etc.)? Are there temporary files involved? This is weird and I don't think I understand what restrictions we're working under for adding/verifying snapshot tests. I didn't see anything in the insta docs about this--is there something to read?

Each call to `insta::glob!` finds all matched paths, removes them from the `*.snap` file path.  So `snapshots/rotate.c` and `snapshots/linux/rotate.c` would both result in a `snapshots/snapshots__transpile@rotate.c.snap` file (assuming there are other `*.c` files in each of those directories).  These then overwrite each other.

PR #1261, opened 2025-06-30
`transpile`: emit `assert_unchecked(ptr.is_aligned())` for `__builtin_assume_aligned`

kkysen:
Previously, this silently emitted nothing, but now `ptr.is_aligned()` and `core::hint::assert_unchecked` are stable.  However, since we're still on an old pinned nightly, I didn't add a test here.


fw-immunant:
Code changes look good, but it seems like our Rust in CI is too old for this.

kkysen:
Closing for now because our Rust version is too old.

Issue #1245, opened 2025-05-13
Fix `curl` transpilation failing in CI in `c2rust-testsuite`


kkysen:

bungcip:
this is what i get when running in my system
```
~/c2rust-testsuite/tests/curl$ ./make.sh

...
<cut off>
...

openldap.c:82:17: error: conflicting types for 'ldap_connect'; have 'CURLcode(struct connectdata *, _Bool *)'
   82 | static CURLcode ldap_connect(struct connectdata *conn, bool *done);
      |                 ^~~~~~~~~~~~
In file included from openldap.c:39:
/usr/include/ldap.h:1555:1: note: previous declaration of 'ldap_connect' with type 'int(LDAP *)' {aka 'int(struct ldap *)'}
 1555 | ldap_connect( LDAP *ld );
      | ^~~~~~~~~~~~
  CC       libcurl_la-gopher.lo
make[2]: *** [Makefile:2096: libcurl_la-openldap.lo] Error 1
```

bungcip:
this is because curl 7.64.1 cannot be compiled with latest ldap. this pull request https://github.com/curl/curl/pull/7005 already fixed it.  so we can fix it either by update curl, or pin libldap2-dev to old version

bungcip:
after updating curl to 8.13.0 and editing `configure.sh` to add `--with-openssl` parameter. now I can run it and now get more error. At least the error is from `c2rust-transpile` 

```

...
<cut off>
...

 [-Wclang-ast]
warning: Missing type 94387132983040 for node: AstNode { tag: TagTypedefDecl, children: [], loc: SrcSpan { fileid: 179, begin_line: 103, begin_column: 1, end_line: 103, end_column: 37 }, type_id: Some(94387132983040), rvalue: LValue, macro_expansions: [], macro_expansion_text: Some("__ATOMIC_SEQ_CST"), extras: [Text("atomic_long"), Bool(false)] }
Exported Clang AST was invalid. Check warnings above for unimplemented features.
--> /usr/lib/llvm-18/lib/clang/18/include/stdatomic.h:103:1
 [-Wclang-ast]
[...]

PR #1263, opened 2025-07-03
`transpile`: impl `__builtin_ia32_pause` and `__builtin_arm_yield`

thedataking:
* Part of #1245.

Required to transpile modern curl (see https://github.com/curl/curl/issues/9058).

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -357,9 +357,14 @@ impl<'c> Translation<'c> {
                 })
             }
 
-            "__builtin_ia32_pause" => {
+            "__builtin_ia32_pause" | "__builtin_arm_yield" => {
                 // `spin_loop()` is implemented as `_mm_pause()` (the `pause` instruction) on `x86`/`x86_64`,
                 // but it's the safe and cross-platform version of it, so prefer it.
+                // On `arm`, it's implemented as `yield`, although on `aarch64`,
+                // it's implemented as `isb` instead as this is more efficient.
+                // See <https://github.com/rust-lang/rust/commit/c064b6560b7ce0adeb9bbf5d7dcf12b1acb0c807>.
+                // `core::arch::aarch64::__yield()` could be used instead,
+                // but it's unstable (`#![feature(stdarch_arm_hints)]`), so it's not ideal.
                 let spin_loop = mk().abs_path_expr(vec!["core", "hint", "spin_loop"]);
                 let call = mk().call_expr(spin_loop, vec![]);
                 self.convert_side_effects_expr(
--- a/c2rust-transpile/tests/snapshots/pause.c
+++ b/c2rust-transpile/tests/snapshots/pause.c
@@ -1,3 +1,0 @@
-void pause(void) {
-    __builtin_ia32_pause();
-}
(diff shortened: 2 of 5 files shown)


c2rust-transpile/tests/snapshots/pause.rs
```diff
@@ -0,0 +1,18 @@
+#![allow(
+    dead_code,
+    mutable_transmutes,
+    non_camel_case_types,
+    non_snake_case,
+    non_upper_case_globals,
+    unused_assignments,
+    unused_mut
+)]
+#![feature(stdsimd)]
```
kkysen: So `fn import_simd_function` unconditionally adds `#![feature(stdsimd)]` even for stable functions like `core::arch::x86_64::_mm_pause`, which isn't ideal, but should probably be fixed separately.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -375,6 +375,16 @@ impl<'c> Translation<'c> {
...
+            "__builtin_ia32_pause" => {
+                let fn_name = "_mm_pause";
+                self.import_simd_function(fn_name)?;
+                let ident = mk().ident_expr(fn_name);
+                let call = mk().call_expr(ident, vec![]);
+                Ok(WithStmts::new(
+                    vec![mk().semi_stmt(call)],
+                    self.panic_or_err("No value for _mm_pause"),
+                ))
+            }
```
kkysen: This isn't really a SIMD function, so I'd move it out of this group of builtins.

kkysen: That said, I'd prefer to emit the safe and cross-platform `core::hint::spin_loop` instead.  It directly calls `_mm_pause` on x86/x86_64, but also supports other platforms as well: https://github.com/rust-lang/rust/blob/d6120810e56387730b2e86115471354c8084ca4a/library/core/src/hint.rs#L269-L304.

Furthermore, while the x86/x86_64 `_mm_pause` is stable now, the arm/aarch64 equivalent `__yield` is not and requires `#![feature(stdarch_arm_hints)]` (`aarch64`) and `#![feature(stdarch_arm_neon_intrinsics)]` (`arm`).  `__builtin_ia32_pause` is obviously x86/x86_64 only, but I think it'd be better to support the more cross-platform version that makes implementing other architectures' equivalents a lot simpler.

kkysen (changes requested):
As discussed, I'll make these changes myself.

fw-immunant (approved):
LGTM.

c2rust-transpile/src/translator/builtins.rs
thedataking: Discussed offline with @kkysen: I'm not convinced we should try to emit `spin_loop` here in the interest of portability for two reasons:

1. The `curl` code which needs this feature emits `yield` on the `aarch64` code path but it seems `spin_loop` lowers to `isb` on `aarch64` so now we've introduced a subtle codegen difference between the C and Rust version. This makes cross checking more difficult.

2. When the programmer emits x86-specific intrinsics, it is likely on code paths that are inherently target/toolchain/OS specific and, IMHO, it is not worth it to emit more portable intrinsic rather than doing something "less surprising".

PR #1265, opened 2025-07-04
transpile: handle all cases of extra braces in string initializers

### LINKED ISSUE #1264: `curl`: Transpiler emits single-dimensional initializer for 2D entries in `struct alpn_spec`
Curl contains this struct and its static initializer (`lib/vtls/vtls.c`, `lib/vtls/vtls_int.h`):

```C
#define ALPN_ENTRIES_MAX 3
#define ALPN_NAME_MAX    10

struct alpn_spec {
    char    entries[ALPN_ENTRIES_MAX][ALPN_NAME_MAX];
    size_t  count; /* number of entries */
};

static const struct alpn_spec ALPN_SPEC_H11 = {
    { ALPN_HTTP_1_1 },
    1
};

```

The code emitted is:

```Rust
static mut ALPN_SPEC_H11: alpn_spec = unsafe {
    {
        let mut init = alpn_spec {
            entries: *::core::mem::transmute::<
                &[u8; 10],
                &mut [std::ffi::c_char; 10],
            >(b"http/1.1\0\0"),
            count: 1 as std::ffi::c_int as size_t,
        };
        init
    }
};
```

which does not match the required `[[c_char; 10]; 3]` type.

## Reproducing

- Check out https://github.com/immunant/c2rust-testsuite/pull/14
- Run `c2rust-testsuite && PATH=$PATH:$C2RUST_TOP/target/release ./test.py --verbose curl`

kkysen:
* Fixes #1264.

We need to handle the 4 cases in `str_init.c` with identical initializers:
* `ptr_extra_braces`
* `array_extra_braces`
* `array_of_ptrs`
* `array_of_arrays`

All 4 have different types, but the same initializer, which is possible because C allows extra braces around any initializer element.  For non-string literal elements, the clang AST already fixes this up, but doesn't for string literals, so we need to handle them specially.  The default logic handles all except `array_extra_braces`.

`array_extra_braces` is uniquely identified by:
* there being only one element in the initializer list
* the element type of the array being `CTypeKind::Char` (w/o this, `array_of_arrays` is included)
* the expr kind being a string literal (`CExprKind::Literal` of a `CLiteral::String`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -895,17 +895,15 @@ dependencies = [
 
 [[package]]
 name = "insta"
-version = "1.29.0"
+version = "1.43.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a28d25139df397cbca21408bb742cf6837e04cdbebf1b07b760caf971d6a972"
+checksum = "154934ea70c58054b556dd430b99a98c2a7ff5309ac9891597e339b5c28f4371"
 dependencies = [
  "console",
  "globset",
- "lazy_static",
- "linked-hash-map",
+ "once_cell",
  "similar",
  "walkdir",
- "yaml-rust",
 ]
 
 [[package]]
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -173,24 +173,32 @@ impl<'c> Translation<'c> {
             CTypeKind::ConstantArray(ty, n) => {
                 // Convert all of the provided initializer values
 
-                // Need to check to see if the next item is a string literal,
-                // if it is need to treat it as a declaration, rather than
-                // an init list. https://github.com/GaloisInc/C2Rust/issues/40
-                let mut is_string = false;
-
-                if ids.len() == 1 {
-                    let v = ids.first().unwrap();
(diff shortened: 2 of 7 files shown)


fw-immunant (approved):
Code changes LGTM, but history is a bit messy. Could you squash down the changes to test cases so we just have the before (adding test cases, some commented out) and after (all passing) states? Or even split out the `insta` update commit and squash everything else.

fw-immunant:
I'm also curious, will this interact with https://github.com/immunant/c2rust/pull/1218?

kkysen (commented):
> Code changes LGTM, but history is a bit messy. Could you squash down the changes to test cases so we just have the before (adding test cases, some commented out) and after (all passing) states? Or even split out the `insta` update commit and squash everything else.

Okay, I can do that. I didn't squash the renaming function commit since it was a bit messy with the snapshots, but I can do that if you prefer. Squashing the comment commit should be simple.

> I'm also curious, will this interact with #1218?

I'm not sure. Let me check.

kkysen:
> I'm also curious, will this interact with #1218?

Yes, but it should be pretty easy to combine them.  The `if is_string_literal(*single)` case just needs to be updated with the new logic here.

PR #1266, opened 2025-07-05
Translate C integer types to portable Rust types

fw-immunant:
This substantially improves our fidelity of type translation for common integral types. We now recognize types like `size_t` and `uint8_t` and convert them to the appropriate platform-independent Rust types. This is nontrivial because C-the-language does not have exact-width types and instead defines them via platform-dependent mappings to types with platform-dependent sizes, such as `unsigned long` and `unsigned char`. We have to preëmpt Clang to introduce a distinction between these types and "native" C ones when we convert the AST after exporting; to make effective use of this type information in translation we need to pass the expected type of the translated expression downward through conversion of expressions and statements.

PR #1267, opened 2025-07-05
Various small cleanups

fw-immunant:
Split out of #1266 to keep it smaller, mostly intended to facilitate future development.


fw-immunant:
No semantic changes, and tests pass modulo the known Curl issue (#1245).

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2845,7 +2846,7 @@ impl<'c> Translation<'c> {
                             return true;
                         }
 
-                        // None assignments don't prove enough type information unless there are follow-up assignments
+                        // None assignments don't provide enough type information unless there are follow-up assignments
```
kkysen, line 2849:
```suggestion
                        // `None` assignments don't provide enough type information unless there are follow-up assignments
```

nit: A bit clearer since None could read as an English word, too.

PR #1268, opened 2025-07-09
More driveby cleanups

fw-immunant:
No semantic changes except for an improved error message when the `compile_commands.json` passed an as argument does not exist.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -233,6 +233,7 @@ impl<'c> Translation<'c> {
         }
     }
 
+    /// Translate an assignment binary operator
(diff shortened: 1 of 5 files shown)


c2rust-transpile/src/translator/operators.rs
```diff
@@ -262,7 +263,7 @@ impl<'c> Translation<'c> {
         )
     }
 
-    /// Translate an assignment binary operator
+    /// Translate an assignment binary operator, providing a pre-translated RHS expression
```
```suggestion
    /// Translate an assignment binary operator given a pre-translated RHS expression
```

I think this makes more sense. Provided sounds like the method returns that.

kkysen (approved):
LGTM except the providing wording is confusing.

PR #1270, opened 2025-07-10
Allow platform-dependent snapshot test results

fw-immunant:
This is needed for #1266.


c2rust-transpile/tests/snapshots/snapshots__transpile-linux.snap
```diff
@@ -0,0 +1 @@
+snapshots__transpile-linux@rotate.c.snap
```
kkysen: What generates this?

c2rust-transpile/tests/snapshots.rs
```diff
@@ -77,28 +61,37 @@ fn transpile(c_path: &Path) {
...
+
+    let name = platform
+        .map(|platform| ["transpile", platform].join("-"))
+        .unwrap_or("transpile".into());
+
+    insta::assert_snapshot!(name, &rs, &debug_expr);
 
     let status = Command::new("rustc")
-        .args(&["--crate-type", "lib", "--edition", "2021"])
+        .args(&["--crate-type", "lib", "--edition", "2021", "-o", "-"])
```
kkysen, line 72: Ah this is a lot simpler way of doing this.  This writes to stout?  How does that work vs `/dev/null`?

c2rust-transpile/tests/snapshots/snapshots__transpile-macos@rotate.c.snap
```diff
@@ -0,0 +1,28 @@
+---
```
kkysen, line 1: How do we generate the macOS snapshots on Linux easily?

c2rust-transpile/tests/snapshots.rs
fw-immunant: `/dev/null` would be a more obvious bit-bucket, but rustc attempts to create temporary files in the output directory and can't for `/dev`.

c2rust-transpile/tests/snapshots/snapshots__transpile-macos@rotate.c.snap
fw-immunant: There's no way to do so from Linux without a macOS VM; we need to compile against Apple's proprietary (though nominally open-source) headers. If we want to support macOS, this is the price we pay.

We could improve the workflow by having snapshot .new files produced as an artifact from our macOS CI pipeline.

c2rust-transpile/tests/snapshots/snapshots__transpile-linux.snap
fw-immunant: I'm not sure whether you mean what creates this string or what creates the file `snapshots__transpile-linux@rotate.c.snap`. The string itself is from me running `ln -s` to introduce the symlink. The target file is created by insta itself inside `insta::glob` if multiple files are globbed, along with other files for the other glob matches.

If `insta::glob` only matches one file, it puts it at a path with no `@input.filename` portion, hence needing the symlink.

kkysen: Ah, these are symlinks.  `insta` does that removing common prefix thing here again.  I'd rather just put a dummy `empty.c` file in `snapshots/platform-specific`, and then once we add another real test we can remove it.

c2rust-transpile/tests/snapshots.rs
kkysen: What gets printed to stdout with `-`?

fw-immunant: The `--emit=link` contents, whatever they are. In this case the output is not captured, so it doesn't really matter what it is.

c2rust-transpile/tests/snapshots/snapshots__transpile-linux.snap
fw-immunant: Done.

PR #1272, opened 2025-07-11
ci: allow reusing testsuite workflow

fw-immunant:
Just making sure this doesn't break CI on this repo, then we can see if it enables what we want on the testsuite side.


fw-immunant:
No breakage here, and the job on the testsuite side seems to work. Merging.

PR #1274, opened 2025-07-12
fix compilation error on clang/llvm 21

bungcip:
GCCAsmStmt::getOutputConstraint() now return `std::string` instead of `StringRef`. this change is introduced in https://github.com/llvm/llvm-project/pull/131003


kkysen (approved):

kkysen:
Thanks!

PR #1275, opened 2025-07-14
ci: move macOS `azure-pipelines.yml` to GitHub Actions with a matrix

### LINKED ISSUE #1278: Port macOS `azure-pipelines.yml` to GitHub Actions
The existing Azure Pipelines Darwin/macOS is x86_64, so we should use x86_64 macOS in GitHub Actions initially as well.

kkysen:
* Fixes #1278.


.github/workflows/ci.yml
```diff
@@ -17,7 +23,10 @@ jobs:
       # cached, what gets used as part of the key, and what additional handling
       # happens to make the cache reliable and smaller.
       - uses: Swatinem/rust-cache@v2
-      - run: |
+
+      - name: Install packages (Ubuntu)
+        if: runner.os == 'Linux'
```
fw-immunant, line 28: If there's a straightforward way to branch on the distro rather than the OS here, it would be cleaner. But I don't see any such thing in the GH [context docs](https://docs.github.com/en/actions/reference/contexts-reference#runner-context), from a cursory look.

kkysen: Yeah, I was looking for that, but I wasn't able to find a way using `if runner.*` like this. But we're only using one distro per OS anyways, and GitHub Actions only has Ubuntu images as well (which is why we removed testing the other Linux distro), so I think this should be fine for now.

fw-immunant:
Actually, why didn't the testsuite run here?

It seems to be back for subsequent PRs. Maybe GH knows that we only changed one `.yaml` file here so it omitted testing unrelated pipelines.

kkysen:
> Actually, why didn't the testsuite run here?
> 
> It seems to be back for subsequent PRs. Maybe GH knows that we only changed one `.yaml` file here so it omitted testing unrelated pipelines.

Wdym? Isn't it running here? https://github.com/immunant/c2rust/actions/runs/16271520856

fw-immunant:
> Wdym? Isn't it running here? https://github.com/immunant/c2rust/actions/runs/16271520856

Not that I see--we should have 3 jobs: test on macOS, test on Linux, and c2rust-testsuite, but I only see the former two there. It [seems to be](https://github.com/immunant/c2rust/actions/) running for subsequent PRs and merges though.

kkysen:
https://github.com/immunant/c2rust/actions/runs/16271520844 is the corresponding `c2rust-testsuite` run.  That said, I didn't touch `c2rust-testsuite` at all so I'm not sure why it would change.

fw-immunant:
Huh, it didn't exist before. I guess it just took a long time to queue.

PR #1277, opened 2025-07-14
ci: use `awalsh128/cache-apt-pkgs-action@latest` to cache `apt install`

kkysen:
We already use this GitHub Action on other projects like IA2 and rav1d.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,37 +59,36 @@ jobs:
 
       - name: Install packages (Ubuntu)
         if: runner.os == 'Linux'
-        run: |
-          sudo apt-get update
-          sudo apt-get install \
-            clang \
-            clang-tools \
-            cmake \
-            curl \
-            git \
-            gperf \
-            libbrotli-dev \
-            libclang-${{ matrix.clang-version }}-dev \
-            libgcrypt20 \
-            libreadline-dev \
-            libidn2-dev \
-            libldap2-dev \
-            libncurses5-dev \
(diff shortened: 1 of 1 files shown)


fw-immunant (approved):

kkysen:
@thedataking, can you allow `awalsh128/cache-apt-pkgs-action` to use, since we already use it in IA2 and rav1d? Also, if you're already doing that, can you also allow `dtolnay/rust-toolchain`? It's used by most `rust-lang/` repos.

thedataking:
I've whitelisted `awalsh128/cache-apt-pkgs-action`. 

Why do you need `dtolnay/rust-toolchain`? I'd like to understand the need before we polish our CI actions further.

(we have `Swatinem/rust-cache@v2` enabled in case that's relevant.)

kkysen (commented):
> Why do you need `dtolnay/rust-toolchain`? I'd like to understand the need before we polish our CI actions further.
> 
> (we have `Swatinem/rust-cache@v2` enabled in case that's relevant.)

We actually don't. rust-cache covers it, I just had to reverse the order of the steps.  So we should be all good now.  Thanks!

thedataking:
> Still seems to be blocked 😞

I think I forgot to press save. My bad, very sorry 😞

kkysen:
> I think I forgot to press save. My bad, very sorry 😞

😆 https://github.com/immunant/c2rust/actions/runs/16516133789 I tried closing and reopening the PR, but it/s still denying it.

PR #1280, opened 2025-07-14
ci: test on aarch64 macOS

### LINKED ISSUE #1279: Run CI on arm64 macOS
If there's a suitable GitHub Actions runner for arm64 Linux, we should test on that as well.

kkysen:
* Fixes #1279.

This starts testing in CI on aarch64 macOS, our first aarch64 target to test.

Varargs are broken on aarch64 (#1281), so this PR also disables varargs tests on aarch64.

However, doing this didn't work with the current `test_translator.py`, which textually searches the Rust test file for `fn test_*`s.  `#[cfg(not(target_arch = "aarch64"))]` is the simplest and most straightforward way to disable such tests on aarch64, but parsing this from Python and generating a test runner in `main.rs` that's platform-dependent is tricky.  Since Rust's built-in test runner already does much of this, and does it better, I switched `test_translator.py` to simply run `cargo test`.  This means marking all of the test functions with `#[test]` and generating a mostly empty `lib.rs` instead of a `main.rs`.  We also lose the ability to distinguish between expected failures and successes, as `cargo test` treats them the same, but I think this is fine.  It also means we can use the idiomatic `#[should_panic]` instead of parsing the non-standard `// xfail` ourselves.

I can also break this part out into a separate PR if you think that'd be helpful.


scripts/test_translator.py
```diff
@@ -269,9 +271,8 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
                 elif (filename.startswith("test_") and ext == ".rs" and
                       files.search(filename)):
                     rs_test_file = self._read_rust_test_file(path)
-
                     self.rs_test_files.append(rs_test_file)
-
+
```
fw-immunant: Nit: Trailing whitespace added.

scripts/run_ci_checks.sh
```diff
@@ -48,7 +48,9 @@ test() {
 # so changing `RUSTFLAGS` will not trigger a full rebuild.
 test-translator() {
     unset RUSTFLAGS
-    ./scripts/test_translator.py tests/
+    uv venv
+    uv pip install -r ./scripts/requirements.txt
+    uv run ./scripts/test_translator.py tests/
```
fw-immunant, line 53: I assume this improves or fixes something, but I'm not well versed enough in python packaging land to know what. Could you give some rationale/motivation in the commit message? Why are these uv changes needed/desirable?

tests/items/src/test_varargs.rs
```diff
@@ -1,8 +1,11 @@
 //! feature_c_variadic,
 
+use crate::varargs::rust_call_printf;
+// See #1281. Varargs don't yet work on aarch64.
+#[cfg(not(target_arch = "aarch64"))]
```
fw-immunant, line 5: Do we actually know whether this is true for aarch64 in general, or just the case for aarch64 macOS? I would err or the side of a more specific exception if we haven't narrowed it down.

fw-immunant (approved):
A couple questions inline; when addressed this is good to merge.

scripts/run_ci_checks.sh
kkysen: No, this was just already updated in `ci.yml` and I forgot to update it here, too, so I just updated it here as well now when I was running things locally and realized it diverged.

tests/items/src/test_varargs.rs
kkysen: Linux aarch64 is much more broken currently, so I'm not sure.  When I eventually fix #1283, I'll check.

fw-immunant: Ok, added a note over there.

scripts/run_ci_checks.sh
fw-immunant: Ok, just explain in the commit message why we're making this change.

kkysen: Whoops, thought I had.

kkysen: Fixed in 1d7346cf6.

scripts/test_translator.py
kkysen: Fixed in 5e1797d28.

PR #1282, opened 2025-07-14
[DNM] test arm64 vaarg fix in CI

fw-immunant:
I can't validate this hunch locally, so opening a PR for CI.


fw-immunant:
Ok, looks like this is definitely not the right fix.

PR #1283, opened 2025-07-15
ci: test on arm64 linux

kkysen:
Just testing.


fw-immunant:
Note to see if we can enable varargs tests on aarch64 Linux when this lands.

fw-immunant:
It looks to me like this is a toolchain problem. Probably using a stable toolchain where possible would help, as suggested in #1227.

kkysen:
> It looks to me like this is a toolchain problem. Probably using a stable toolchain where possible would help, as suggested in #1227.

That should fix compiling the transpiler, but not the rest of the codebase where the errors are from.

PR #1286, opened 2025-07-17
Fix skipping of Paren expressions in typed AST iterators

fw-immunant:
I found this bug while working on #1266: it was causing our explicit type propagation pass to stop at any parenthesized expression.

Fixing the parenthesization results in some other failures, which the second commit addresses with a mostly-stub implementation of a few new AST nodes.


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -1749,6 +1749,33 @@ class TranslateASTVisitor final
...
+#if CLANG_VERSION_MAJOR >= 9
+    bool VisitBuiltinBitCastExpr(BuiltinBitCastExpr *E) {
+        auto children = E->children();
+        std::vector<void *> childIds(std::begin(children), std::end(children));
+        encode_entry(E, TagBuiltinBitCastExpr, childIds);
+        return true;
+    }
+#endif
+
+#if CLANG_VERSION_MAJOR >= 9
```
kkysen: You want these all separate `#if`s?

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -1706,6 +1706,52 @@ impl ConversionContext {
                     self.expr_possibly_as_stmt(expected_ty, new_id, node, e)
                 }
 
+                ASTEntryTag::TagBuiltinBitCastExpr => {
+                    let expression_old = node.children[0].expect("Expected expression for bitcast");
+                    let expression = self.visit_expr(expression_old);
+
+                    let typ_old = node.type_id.expect("Expected type for bitcast");
+                    let typ = self.visit_qualified_type(typ_old);
```
kkysen: nit: Seems like we generally use `ty` not `typ` here.

kkysen (approved):

c2rust-ast-exporter/src/AstExporter.cpp
fw-immunant: Probably not. `BuiltinBitCastExpr` was definitely introduced in Clang 9.0, but I need to verify the version requirements of the other nodes. If they were introduced together I'll consolidate the `#if`s.

fw-immunant: It looks like both were introduced long before Clang 4.0 (fe31481f6829eaebf40205a1695a7aa68f0de36c and 5d41378146ad784843bedd870ffa6dda3966edc4 in 2010 and 2011 respectively), so we don't need any guards for the other two. Will remove.

PR #1287, opened 2025-07-17
Revert "transpile: snapshot: build to stdout rather than creating+deleting rlib"

kkysen:
This reverts commit e09c41f0a4028fd2f5b2dd70deaed943cdd11a37.

This doesn't actually write to stdout, it writes to the file `-`. That's not what we want, so just revert to deleting the file.


fw-immunant:
Good catch, my bad. On the other hand, `rustc --emit link=/dev/null` *does* seem to do what we want.

PR #1288, opened 2025-07-17
ci: run on both Ubuntu 22.04 and ubuntu-latest

fw-immunant:
We were previously running the testsuite on 22.04 and in-repo tests on ubuntu-latest (currently 24.04). Now, run both on both. This is particularly useful coverage because 22.04 uses Clang 15, which we do support but which differs substantially in some behaviors from later versions.

Split out from #1266.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
           - runner: macos-13
             os: macOS
             arch: x86_64
+      fail-fast: false
     runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
           - runner: macos-13
             os: macOS
             arch: x86_64
+            clang-version: 17
       fail-fast: false
     runs-on: ${{ matrix.runner }}
     steps:
@@ -72,7 +73,7 @@ jobs:
         if: runner.os == 'macOS'
         run: |
           # `bash` needed b/c macOS ships with bash 3, which doesn't support arrays properly
-          brew install -q cmake ninja gpg llvm@17 bash
+          brew install -q cmake ninja gpg llvm@${{ matrix.clang-version }} bash
 
       - uses: astral-sh/setup-uv@v6
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: ci
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
 
 jobs:
@@ -13,6 +13,11 @@ jobs:
           - runner: ubuntu-latest
             os: Linux
             arch: x86_64
+            clang-version: 18
+          - runner: ubuntu-22.04
+            os: Linux
+            arch: x86_64
+            clang-version: 15
           - runner: macos-13
             os: macOS
             arch: x86_64
@@ -39,7 +44,7 @@ jobs:
             git \
             gperf \
             libbrotli-dev \
-            libclang-dev \
+            libclang-${{ matrix.clang-version }}-dev \
             libgcrypt20 \
             libreadline-dev \
             libidn2-dev \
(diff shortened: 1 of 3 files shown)


fw-immunant:
See https://github.com/immunant/c2rust/actions/runs/16352406150/job/46202286312?pr=1288 for the error prevented by the `test_translator.py` change.

.github/workflows/ci.yml
```diff
@@ -13,9 +13,15 @@ jobs:
           - runner: ubuntu-latest
             os: Linux
             arch: x86_64
+            clang-version: 18
```
kkysen, line 16: It's a bit more convenient if you do it as


```suggestion
            clang-version: clang-18
```

so the default name says `clang-18` instead of just `18`.

scripts/test_translator.py
```diff
@@ -234,8 +234,11 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
         self.target = None
 
         # include the compiler resource directory in compile_commands.json
-        _, stdout, _ = clang["-print-resource-dir"].run(retcode=None)
-        self.clang_resource_dir = " \"-I{}/include\",".format(stdout.strip())
+        if sys.platform == "darwin":
+            _, stdout, _ = clang["-print-resource-dir"].run(retcode=None)
+            self.clang_resource_dir = " \"-I{}/include\",".format(stdout.strip())
```
kkysen, line 243: nit: Easier to read IMO.

```suggestion
            self.clang_resource_dir = f" \"-I{stdout.strip()}/include\",")
```

kkysen: Oh, I guess you didn't actually change this.  Either way's fine then.

scripts/test_translator.py
```diff
@@ -234,8 +234,11 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
         self.target = None
 
         # include the compiler resource directory in compile_commands.json
-        _, stdout, _ = clang["-print-resource-dir"].run(retcode=None)
-        self.clang_resource_dir = " \"-I{}/include\",".format(stdout.strip())
+        if sys.platform == "darwin":
```
kkysen, line 241: So we skip setting `clang_resource_dir` on linux, but it only seems to be an issue on Ubuntu 22/clang 15, not later versions.  Should that be checked here?  Although, without this on Ubuntu 24 seems to work as well.

fw-immunant: We should never have to pass the resource directory explicitly--it's built into the compiler toolchain. For some reason, we do need to do so on macOS 12 to work around an issue with SIMD intrinsics: cc4c73f1b6526b7d87cd77c6af7efe91212d8422

We could always pass the resource directory, but on systems that have multiple Clang versions installed, we don't know which to dispatch to to ask for the resource directory matching the version of Clang that c2rust is built against. So instead, we limit the scope of this workaround to only macOS.

scripts/test_translator.py
fw-immunant: As unreadable as this is, I think it's probably best not to mix f-strings with the `.format()` used in the rest of the file.

.github/workflows/ci.yml
fw-immunant: The version alone is probably better in case we need to use it as part of `llvm-config-N` or such. I realize the GitHub test matrix labels are a bit opaque with more minimal values, but looking at the matrix itself makes it clear what's going on.

scripts/test_translator.py
kkysen: Okay, that makes sense. Could you put that in a comment?

.github/workflows/ci.yml
kkysen: Hmm, true. We can also just set the name explicitly ourselves.

fw-immunant: Oh, I didn't realize. Let's just do that then.

.github/workflows/ci.yml
```diff
@@ -13,9 +13,17 @@ jobs:
...
+          - runner: ubuntu-22.04
+            os: Linux
+            arch: x86_64
+            clang-version: 15
           - runner: macos-13
             os: macOS
             arch: x86_64
+            clang-version: 17
+      fail-fast: false
+    name: test (${{ matrix.runner }}, Clang ${{ matrix.clang-version }})
```
kkysen:
```suggestion
    name: test (${{ matrix.runner }}, ${{ matrix.os }}, ${{ matrix.arch}}, Clang ${{ matrix.clang-version }})
```

PR #1290, opened 2025-07-18
test_translator: remove `#![feature(nll)]`, stabilized in 1.63

kkysen:
Our old pinned nightly is 1.65, so this was already stabilized even back then.

### DIFF
--- a/scripts/test_translator.py
+++ b/scripts/test_translator.py
@@ -402,7 +402,6 @@ class TestDirectory:
             "extern_types",
             "simd_ffi",
             "stdsimd",
-            "nll",
             "linkage",
             "register_tool",
         ])


fw-immunant (approved):
Code change is fine, but as an aside I'd like to understand better what the source of truth for compatibility of our generated code is. Do you happen to know offhand? I can't find anything about this in our docs.

fw-immunant:
Is this the toolchain we expect transpiled code to compile against? https://github.com/immunant/c2rust/blob/2adadd104dff57b8bf8225b76d832a00ed09eda4/c2rust-transpile/src/build_files/generated-rust-toolchain.toml#L1-L3

kkysen (commented):
> Is this the toolchain we expect transpiled code to compile against?
> 
> https://github.com/immunant/c2rust/blob/2adadd104dff57b8bf8225b76d832a00ed09eda4/c2rust-transpile/src/build_files/generated-rust-toolchain.toml#L1-L3

Yes, everything is pinned to this nightly and is detected whenever we run cargo/rustc.

kkysen:
The exception is when we publish, as that's broken on the old nightly. I use stable to publish.

PR #1291, opened 2025-07-21
transpile: save platform-specific `.rs` files, too

kkysen:
We already save platform-specific `.snap` files for snapshot tests, but the `.rs` files still clash, and we aren't checking them in for all platforms.  This moves the `.rs` files to `.{platform}.rs`.  And since the crate name can't have a `.` in it, we need to explicitly set the crate name to the file stem now.

This also cleans up some other little stuff.  `dummy.c` isn't needed anymore since we now have 2 other platform-specific snapshots.  Also, set the `.rlib` path with `-o` explicitly instead of guessing what the output file will be named.


fw-immunant:
A separate PR is fine. Could it precede this one to simplify the diffstats for everything that's in flight?

fw-immunant (approved):
Let's go ahead and merge this. I'd prefer a cleaner history without redundant generated `.rs` files but we should end up in the same final repo state either way.

Issue #123, opened 2019-06-27
Redeclaration with incompatible signature may use unexpected signature


TheDan64:
Newlib does a little bit more complicated version of this:
```c
#include <limits.h>

double __exp10(unsigned x);

int main(void) {
  __exp10(1U);

  return 0;
}
```

Where `double __exp10(unsigned x)` is defined in another local file.

When translating this code you get a warning from C:
```shell
/home/dkolsoi/repos/c2rust/test.c:3:8: warning: incompatible redeclaration of library function
      '__exp10' [-Wincompatible-library-redeclaration]
double __exp10(unsigned x);
       ^
/home/dkolsoi/repos/c2rust/test.c:3:8: note: '__exp10' is a builtin with type 'double (double)'
```

But the translated code:
```rust
extern "C" {
    #[no_mangle]
    fn __exp10(_: libc::c_double) -> libc::c_double;
}
unsafe fn main_0() -> libc::c_int { __exp10(1u32); return 0i32; }
```
Will fail to compile because it uses the builtin signature which takes a double and not an int.

Interestingly, passing the `-ffreestanding` flag to clang via the transpiler will bypass this issue and use the correct signature:
```rust
extern "C" {
    #[no_mangle]
    fn __exp10(x: libc::c_uint) -> libc::c_double;
}
#[export_name = "main"]
pub unsafe extern "C" fn main_0() -> libc::c_int {
    __exp10(1u32);
    return 0i32;
}
```

(also interesting, but not problematic, note that it's using `x` not `_` in the param sig)

PR #1292, opened 2025-07-21
transpile: revert `__builtin_ia32_pause`/`__builtin_arm_yield` to emit `_mm_pause` and `__yield` instead of cross-platform `spin_loop`

kkysen:
Fixes <https://github.com/immunant/c2rust/pull/1263#discussion_r2193513798>.

This switches back to emitting the more direct but unstable translations of `_mm_pause` and `__yield`.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -381,10 +381,10 @@ impl<'c> Translation<'c> {
             }
 
             "__builtin_ia32_pause" => {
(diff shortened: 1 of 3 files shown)


kkysen (commented):
> The `import_simd_function` can also cause `stdstimd` to be used. Would it make sense to open an issue capturing the TODOs you noted for deprecation and reference that issue in both locations where `stdsimd` might be pulled in? E.g.
> 
> ```
> // TODO(#123):  In Rust 1.88 (probably earlier), `#![feature(stdsimd)]` has been removed
> ```
> 
> You can easily test if it was pulled in 1.88 or 1.87 using compiler-explorer if you want.

Tracked in #1298 now.

Issue #1262, opened 2025-07-02
Consider using native f128 support


fw-immunant:
The [f128 crate](https://github.com/jkarns275/f128) is deprecated (per its readme) as native support for the type is available in Nightly. We may want to wait until this is stable, but should track the feature's stabilization to decide when to move. This might fix bugs such as #829.

fw-immunant:
Apparently the f128 crate also does not work on macOS; see b800f4da322445b6450297bcae9571b537bb406a.

PR #1294, opened 2025-07-22
transpile: do not allow f128::new in const contexts

### LINKED ISSUE #829: `long double` #defines incompatible with `--translate-const-macros`
Given:

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <math.h>

#define _PI_180 (M_PIl / 180.0)
#define _180_PI (180.0 / M_PIl)

int main(void) {
    printf("%LF\n", _PI_180);
    printf("%LF\n", _180_PI);
    return 0;
}
```

The generated code (with `--translate-const-macros`) is:

```rust
use ::c2rust_out::*;
extern "C" {
}
pub const M_PIl: f128::f128 = f128::f128::new(3.141592653589793238462643383279502884);
pub const _PI_180: f128::f128 = M_PIl / f128::f128::new(180.0f64);
pub const _180_PI: f128::f128 = f128::f128::new(180.0f64) / M_PIl;
    printf(b"%LF\n\0" as *const u8 as *const libc::c_char, _PI_180);
    printf(b"%LF\n\0" as *const u8 as *const libc::c_char, _180_PI);
    return 0 as libc::c_int;
}
pub fn main() {
}
```

This does not compile, I have abbreviated the output as it just repeats of the same things (several times):

```
error[E0015]: cannot call non-const fn `f128::new::<f64>` in constants
error[E0015]: cannot call non-const operator in constants
```

Software versions:
* Arch Linux (rolling release distro).
* glibc 2.37-2
* clang 15.0.7-1
* rustc 1.67.1 (from rustup)
* c2rust 0.17.0 (from cargo install)

fw-immunant:
Fixes #829. This is a limitation of the f128 crate itself; if changed upstream or if we begin to translate f128 to the native Rust type, we can begin translating f128 in const contexts again (see #1262).


c2rust-transpile/src/translator/literals.rs
```diff
@@ -116,6 +116,13 @@ impl<'c> Translation<'c> {
                 };
                 let val = match self.ast_context.resolve_type(ty.ctype).kind {
                     CTypeKind::LongDouble => {
+                        if ctx.is_const {
+                            return Err(format_translation_err!(
```
kkysen, line 120: Do these errors bubble up and indicate that we can't translate this as a const macro? Or does the transpiler just error?

fw-immunant: The former.

tests/macros/src/define.c
kkysen, line 1: I don't think we test with `--translate-const-macros` yet, so could you instead make this a snapshot test after #1296?  A snapshot test is also easier to understand for something like this where it's helpful to see the output.

fw-immunant: The `macros` `define.c` exercises that flag explicitly; see its first line.

kkysen: Ah, okay.

kkysen: A snapshot test is still easier to read but this is good enough coverage.

Issue #1295, opened 2025-07-22
Implement conservative subset of `--translate-const-macros`


fw-immunant:
`--translate-const-macros` in its full generality is very hard to implement, because it's quite difficult to determine if nontrivial expessions will be const-safe; see [this comment](https://github.com/immunant/c2rust/issues/803#issuecomment-1544754766). It seems like it may not be worth the effort to make the implementation much more complex because it will never cover all possible uses of even non-function-like C macros.

That said, it's annoying that we don't have a way to pick the low-hanging fruit here without attempting to tackle the thornier issues. We should implement a minimal implementation of similar functionality that handles `#define NAME literal` and then in a follow-up expand to handle a minimal conservative subset of expressions such as variable references and binary/unary/paren expressions. By staying extremely conservative we can sidestep the `const` safety question entirely for now.

We might implement this as a mode/argument for the existing `--translate-const-macros` flag, but in any case we want whatever we come up with to be enabled by default, which means it needs solid testing.

kkysen:
To expand this further beyond literals, I think it may be better to implement this moreso outside of the transpiler.  That is, we currently try to translate const macros to rust `const`s.  To do so, we need to ensure all of the operations are `const`-compatible.  And some are `unsafe`, so we need to add `unsafe {}` blocks for them as well.  If these fail, then we just translate it as nothing.

Instead, I think it would be better to translate all const macros (and probably fn macros) as non-`const` `unsafe fn`s in Rust.  We should be able to do this in far more cases, as we don't have to worry if operations are `const`, and we can use `unsafe` operations as well.  Then we just replace the macro expansions with a call to this function instead of a `const` item.

For example, in the #853 case, this will successfully translate it even though the operations are not `const`.

Then we can, optionally, do a secondary analysis to make things as `const` and safe as possible, which is also more generally applicable.  We can add `const` to any `fn` where it's possible and then bubble that up (see [`clippy::missing_const_for_fn`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn)).  And we could do the same for removing `unsafe` from `fn`s that are already safe.

@fw-immunant, what do you think?

Or I can do the `const` analysis in the transpiler.  And I can also either do an `unsafe` analysis, or just put everything in an `unsafe` block even if it may be safe (like we do for some functions that are completely safe right now).

kkysen:
[...]

PR #1296, opened 2025-07-23
transpile: enable `--translate-const-macros` in snapshot tests

kkysen:
`--translate-const-macros` doesn't fully work in all cases yet (see #1295), but it works for our current snapshot tests, so enabling it while we implement #1295 (a minimal version) will help with testing.

### DIFF
--- a/c2rust-transpile/tests/snapshots.rs
+++ b/c2rust-transpile/tests/snapshots.rs
@@ -37,7 +37,7 @@ fn config() -> TranspilerConfig {
         enabled_warnings: Default::default(),
         emit_no_std: false,
         output_dir: None,
-        translate_const_macros: false,
+        translate_const_macros: true,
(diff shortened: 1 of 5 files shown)


fw-immunant:
Shouldn't we update our snapshots in the same PR that implements the desired functionality? This seems premature.

fw-immunant:
> I'm not sure why more testing of what we already currently support is premature. I'll improve it, but this works right now already.

This switches testing from a configuration that we do expect to be commonly used (no `--translate-const-macros`) to one that we don't expect to be commonly used (with `--translate-const-macros`). If we had the conservative subset implemented, it would make sense to test that because it will be on by default. But this change alters the configuration used in testing to differ from the default configuration that most users will use; that's why it seems premature. Even after #1295 is complete, I don't expect most users will enable the full `--translate-const-macros` flag.

PR #1299, opened 2025-07-24
scripts: restore `provision_mac.sh` (removed in #1275) and remove docker scripts

### LINKED ISSUE #1297: Dangling reference to provision_mac.sh
https://github.com/immunant/c2rust/blob/2adadd104dff57b8bf8225b76d832a00ed09eda4/docs/README-developers.md?plain=1#L12-L14 deleted in 807b88dc74233328eecc72a476f2562b1d19c915.

kkysen:
* Fixes #1297.


fw-immunant (changes requested):
Docker support and the `provision_mac.sh` script are both suggested as installation options in our `README-developers.md` file; merging this as-is trades one dangling reference for another. I don't know how much maintenance our Docker support gets, but it seems like a good thing to have a reproducible means to build c2rust. I would leave the Docker support around for now.

kkysen (commented):
Sure, I can remove the docker changes from here. I'm just not sure it even still works.

fw-immunant:
Oh, I didn't see the discussion in the bug. The docker stuff can go if we aren't planning to support it.

PR #1300, opened 2025-07-24
Remove snapshot .rs files

fw-immunant:
These are outputs of snapshot testing, and are redundant with their corresponding `.snap` files; tracking them is just clutter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ inspect/
 *.rs.fixed
 analysis.txt
 polonius_cache/
+
+# Outputs of c2rust-transpile snapshot tests
+c2rust-transpile/tests/snapshots/**/*.rs


.gitignore
```diff
@@ -52,3 +52,6 @@ inspect/
 *.rs.fixed
 analysis.txt
 polonius_cache/
+
+# Outputs of c2rust-transpile snapshot tests
+c2rust-transpile/tests/snapshots/**/*.rs
```
kkysen, line 57: So the `.rs` files are significantly easier, IMO, to review since their diffs are syntax highlighted by GitHub and other tools like `delta`. So I prefer to keep them for that reason. I understand the extra files are mostly redundant, but they also don't have a big drawback either. If there's a way to syntax highlighted the `.snap` files, though, then I'd be totally fine with this (let me know if there is; I might just not be aware).

fw-immunant: In the hypothetical opposite direction, we wouldn't add these files to the repository just to have syntax highlighting on diffs for snapshots. Our repo contents should make sense on their own; we shouldn't track temporary files in git.

thedataking: > If there's a way to syntax highlighted the .snap files, though, then I'd be totally fine with this (let me know if there is; I might just not be aware).

We can try adding the following to `.gitattributes`:

```
# Tell GitHub to treat test‑snapshot files as Rust
c2rust-transpile/tests/snapshots/*.snap linguist-language=Rust
```

(You might have to adjust the glob expression; I'm shooting from the hip here.)

Here's the [docs](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github).

> Our repo contents should make sense on their own; we shouldn't track temporary files in git.

I agree. This is a case where we should follow the principle of least surprise.

kkysen: > We can try adding the following to `.gitattributes`:
> 
> ```
> # Tell GitHub to treat test‑snapshot files as Rust
> c2rust-transpile/tests/snapshots/*.snap linguist-language=Rust
> ```
> 
> (You might have to adjust the glob expression; I'm shooting from the hip here.)

Oh that's very clever. Let me try that and see if it can work.

thedataking: @kkysen sounds good. This might interest you https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#using-gitattributes

PR #1301, opened 2025-07-25
transpile: syntax highlight `*transpile*.c.snap`s as Rust

kkysen:
There are some issues with GitHub caching old renders it seems, but it seems like this can work.  I'm not exactly sure which of the `linguist-language=Rust`, `diff=Rust`, and `linguist-detectable` are necessary due to the difficulty in caching and testing, so this adds all three, which I know can work.

This tests the diffs by running `rustfmt` on each generated `*.rs` file, which we should probably do anyways for consistency.


thedataking:
You could try adding `*.snap diff=rust` to `.gitattributes`. That may give you local diffs with syntax highlighting. 

If that's doesn't do it, I think we're going to have to live without syntax highlighting, unfortunately. Please help Frances get his cleanup PR landed in either case.

kkysen:
> You could try adding `*.snap diff=rust` to `.gitattributes`. That may give you local diffs with syntax highlighting.

I have this already, in a slightly different form.  Is `rust`/`Rust` case-sensitive?

```
c2rust-transpile/tests/snapshots/*transpile*.c.snap diff=Rust
```

thedataking:
> I have this already, in a slightly different form. Is `rust`/`Rust` case-sensitive?

Don't know. The [gitattributes docs](https://git-scm.com/docs/gitattributes) use lowercase `rust` so I did too.

Did this work for you locally? You can always use `gh` from the command line to pull down PRs and review them locally... even if it is more convenient not have this extra step. Pretty sure you also have the option to tell VSCode how to treat `*.snap` files.

PR #1305, opened 2025-07-31
transpile: add `--translate-const-macros conservative`

### LINKED ISSUE #1295: Implement conservative subset of `--translate-const-macros`

kkysen:
* Fixes #1295.

This is the most conservative subset of `--translate-const-macros`, which I've split into `none`, `conservative`, and `experimental` vs. the previous `true` and `false`.  `conservative` should always work, and is intended to become the default (or I could make it now).  `experimental` tries to do as much as possible, but may break.

For `--translate-const-macros`, `conservative` for now just means literals, i.e. `CExprKind::Literal`s.  This is the most basic initial implementation, although there are lots of other low-hanging fruit.  Literals are the only `CExprKind` that is non-recursive, which makes it a lot simpler to reason about.


thedataking:
nit: `minimal` makes sense from an implementation view. From a user perspective, `safe`, `stable`, or `conservative` is probably more useful since stuff should move from `experimental` to `minimal` thus making it a misnomer down the road.

kkysen (commented):
> nit: `minimal` makes sense from an implementation view. From a user perspective, `safe`, `stable`, or `conservative` is probably more useful since stuff should move from `experimental` to `minimal` thus making it a misnomer down the road.

I chose `minimal` since Rust features themselves have a history of stabilizing `min_*` features like `min_const_generics` or `min_specialization`, which are likewise the minimal version of a larger feature that still needs a lot more work to fully stabilize. 

Open to a different name for `experimental`, though, but I think `min` or `minimal` is good because it already has an understood meaning in the Rust community.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -36,8 +36,8 @@ fn config() -> TranspilerConfig {
         enabled_warnings: Default::default(),
         emit_no_std: false,
         output_dir: None,
-        translate_const_macros: false,
-        translate_fn_macros: false,
+        translate_const_macros: TranslateMacros::Minimal,
+        translate_fn_macros: TranslateMacros::Minimal,
```
fw-immunant: This should still match the default (at this point, `None`). We should change the settings used for snapshots at the same time as we change the settings used for command-line transpilation.

fw-immunant (changes requested):
Mostly looks good, but we should make the default `minimal`/`conservative` for both the CLI and snapshot tests at the same time. As far as I can tell this PR changes the latter but not the former.

fw-immunant:
Re: naming, I don't have strong preferences; the semantics (especially sane defaults) are what's important here.

c2rust-transpile/tests/snapshots.rs
kkysen: Make it the default in the CLI in this PR?  I'll make it the default for `test_translator.py`, too, but I just want to make sure you mean change the default CLI behavior as well.

kkysen: I had left `test_translator.py` testing `experimental` as that's its previous behavior.

fw-immunant: I mean that the default for snapshots and the CLI should be kept in sync. It's up to you whether we change that default (from `None` to `Minimal`/`Conservative`) in this PR or later.

The `test_translator.py` change is, as you say, semantics-preserving. We shouldn't need any other changes to `test_translator.py` here afaik.

kkysen: > I mean that the default for snapshots and the CLI should be kept in sync. It's up to you whether we change that default (from `None` to `Minimal`/`Conservative`) in this PR or later.

Well we need to change it for snapshots in this PR in order to test it.  So I guess that means immediately changing the behavior for the CLI, which I can do.

kkysen: Fixed in 469e4c3.

Issue #853, opened 2023-02-24
Missing casts in const macro translation


DCNick3:
A but I found when working on [unsafe-libopus](https://github.com/DCNick3/unsafe-libopus)

Apparently, when const macro translation is enabled, some casts are dropped, which leads to a change in semantics.

## Reproducible case

```c
#include <stdio.h>

#define silk_int16_MIN ((short)0x8000)

int test() { return silk_int16_MIN; }

int main() {
    printf("%d\n", test());
}
```

Without `--translate-const-macros`:
```rust
#![allow(dead_code, mutable_transmutes, non_camel_case_types, non_snake_case, non_upper_case_globals, unused_assignments, unused_mut)]
extern "C" {
    fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
}
#[no_mangle]
pub unsafe extern "C" fn test() -> libc::c_int {
    return 0x8000 as libc::c_int as libc::c_short as libc::c_int;
}
unsafe fn main_0() -> libc::c_int {
    printf(b"%d\n\0" as *const u8 as *const libc::c_char, test());
    return 0;
}
pub fn main() {
    unsafe { ::std::process::exit(main_0() as i32) }
}
```

Outputs `-32768`

With `--translate-const-macros`:
```rust
#![allow(dead_code, mutable_transmutes, non_camel_case_types, non_snake_case, non_upper_case_globals, unused_assignments, unused_mut)]
extern "C" {
    fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
}
pub const silk_int16_MIN: libc::c_int = 0x8000 as libc::c_int;
#[no_mangle]
pub unsafe extern "C" fn test() -> libc::c_int {
    return silk_int16_MIN;
}
unsafe fn main_0() -> libc::c_int {
    printf(b"%d\n\0" as *const u8 as *const libc::c_char, test());
    return 0;
}
pub fn main() {
    unsafe { ::std::process::exit(main_0() as i32) }
}
```

Outputs `32768`

## The bug

When no constants are generated the value is correctly cast to `c_short` before being widened back: `0x8000 as libc::c_int as libc::c_short as libc::c_int`

When constant translation is enabled, however, the `0x8000` constant is not cast to the `short` type and stored directly as an int, which is incorrect

GPHemsley:
This appears to be because `canonical_macro_replacement()` is calling `resolve_expr_type_id()`, which then discards the intermediate types:

[...]

PR #1306, opened 2025-08-02
transpile: expand `--translate-const-macros conservative` to a lot more `CExprKind`s

### LINKED ISSUE #803: Problematic macro translation with --translate-const-macros

kkysen:
* Fixes #803.

We do this by recursively checking whether an expression is `const`.  This is generally done in a conservative manner, modulo the `ExplicitCast` bug (#853), non-`const` `sizeof(VLA)`s, and `f128`'s non-`const` methods (#1262).  For #853, this does generate incorrect code even on `conservative`, but I'll work on fixing that next.

Statements are not handled yet.

Also, because we now allow certain operations that are `unsafe`, like ptr arithmetic, we wrap all `const`s in an `unsafe` block.  This is similar to how all `fn`s we generate are marked `unsafe fn`s even if they don't contain any `unsafe` operations within them.  We can improve this, but for now, this works and is correct.

Also, the output was being non-deterministic due to the usage of `HashMap`s for macro maps like `macro_invocations`, so I changed these to `IndexMap`s that are order-preserving.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
@@ -29,7 +29,7 @@ pub struct fn_ptrs {
     pub fn2: Option<unsafe extern "C" fn(std::ffi::c_int) -> std::ffi::c_int>,
 }
 pub type zstd_platform_dependent_type = std::ffi::c_long;
-pub const NESTED_INT: std::ffi::c_int = 0xffff as std::ffi::c_int;
+pub const NESTED_INT: std::ffi::c_int = unsafe { 0xffff as std::ffi::c_int };
 #[no_mangle]
 pub unsafe extern "C" fn local_muts() {
     let mut literal_int: std::ffi::c_int = 0xffff as std::ffi::c_int;
@@ -323,7 +323,7 @@ pub static mut global_const_str_concatenation: [std::ffi::c_char; 18] = unsafe {
 pub unsafe extern "C" fn test_fn_macro(mut x: std::ffi::c_int) -> std::ffi::c_int {
     return x * x;
 }
-pub const TEST_CONST2: std::ffi::c_int = 2 as std::ffi::c_int;
+pub const TEST_CONST2: std::ffi::c_int = unsafe { 2 as std::ffi::c_int };
 #[no_mangle]
 pub unsafe extern "C" fn reference_define() -> std::ffi::c_int {
     let mut x: std::ffi::c_int = 1 as std::ffi::c_int;
@@ -344,8 +344,8 @@ pub static mut fns: fn_ptrs = {
 };
 #[no_mangle]
 pub static mut p: *const fn_ptrs = unsafe { &fns as *const fn_ptrs };
-pub const ZSTD_WINDOWLOG_MAX_32: std::ffi::c_int = 30 as std::ffi::c_int;
-pub const ZSTD_WINDOWLOG_MAX_64: std::ffi::c_int = 31 as std::ffi::c_int;
+pub const ZSTD_WINDOWLOG_MAX_32: std::ffi::c_int = unsafe { 30 as std::ffi::c_int };
+pub const ZSTD_WINDOWLOG_MAX_64: std::ffi::c_int = unsafe { 31 as std::ffi::c_int };
 #[no_mangle]
 pub unsafe extern "C" fn test_zstd() -> U64 {
     return (if ::core::mem::size_of::<zstd_platform_dependent_type>() as std::ffi::c_ulong
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
@@ -29,223 +29,190 @@ pub struct fn_ptrs {
     pub fn2: Option<unsafe extern "C" fn(std::ffi::c_int) -> std::ffi::c_int>,
 }
 pub type zstd_platform_dependent_type = std::ffi::c_long;
-pub const NESTED_INT: std::ffi::c_int = unsafe { 0xffff as std::ffi::c_int };
+pub const true_0: std::ffi::c_int = unsafe { 1 as std::ffi::c_int };
+pub const LITERAL_INT: std::ffi::c_int = unsafe { 0xffff as std::ffi::c_int };
+pub const LITERAL_BOOL: std::ffi::c_int = unsafe { true_0 };
+pub const LITERAL_FLOAT: std::ffi::c_double = unsafe { 3.14f64 };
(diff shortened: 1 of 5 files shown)


fw-immunant:
What variation in output did we see due to nondeterministic iteration? Was it reordered output definitions or something more subtle/messy? Definitely worthwhile to make our output deterministic here, I'm just curious how this manifested.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
```diff
@@ -29,287 +29,245 @@ pub struct fn_ptrs {
     pub fn2: Option<unsafe extern "C" fn(std::ffi::c_int) -> std::ffi::c_int>,
 }
 pub type zstd_platform_dependent_type = std::ffi::c_long;
-pub const NESTED_INT: std::ffi::c_int = 0xffff as std::ffi::c_int;
+pub const true_0: std::ffi::c_int = unsafe { 1 as std::ffi::c_int };
```
kkysen, line 34: @fw-immunant, note that this does achieve a lot of what #290 does.  It's similar to the portable types, stuff.  There, we'll now emit a `uint32_t` type alias instead of a `u32` directly, and this will emit a `true_0` `const` instead of a `true` literal directly.  Both could then be refactored pretty easily.  I'm not exactly sure how #290 works, but the advantage here is that it matches where the C used the `true` and `false` macros, and if C uses 0 and non-zero directly, then it keeps that behavior.  And then I'm not sure how #290 handles `int`s vs `_Bool`s either.

We can also resurrect #295 to use `r#true` raw identifiers instead if that's helpful.

fw-immunant:
One thing I'd like to include here is documentation (just a comment probably) relating the conservative/pessimistic analysis in `is_const_expr` with the less-conservative (aka unsound) analysis used by `--translate-const-macros=experimental`: in the latter case, we simply call `ctx.set_const(true)` and translate the expression. This has the effect of making various checks of (effectively) the form `if ctx.is_const() { return error }` trigger at various points in translating the expression where we recognize it to be incompatible with `const`. I don't think we need to merge the `is_const_expr` approach with this one, as the latter is fail-open (and does so in practice, e.g. in the translation of function call expressions which results in #803). But we should note for readers that there are effectively two implementations of this logic, and if/when we attempt to improve the `experimental` mode it would be by adding more logic of this form.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
fw-immunant: There definitely is some overlap in spirit but I think it's probably worthwhile to address this from both sides--this PR will clean up some uses of `#define` symbols, but it seems like we're still mostly translating them as integers.

Do you know what might help us translate `LITERAL_BOOL` as `const LITERAL_BOOL: bool = ...;`? It seems like `recreate_const_macro_from_expansions` might be able to realize that all uses are boolifying it.

kkysen: > Do you know what might help us translate `LITERAL_BOOL` as `const LITERAL_BOOL: bool = ...;`? It seems like `recreate_const_macro_from_expansions` might be able to realize that all uses are boolifying it.

I'm not sure, but I'll look into it.

kkysen:
> This has the effect of making various checks of (effectively) the form `if ctx.is_const() { return error }` trigger at various points in translating the expression where we recognize it to be incompatible with `const`. I don't think we need to merge the `is_const_expr` approach with this one, as the latter is fail-open (and does so in practice, e.g. in the translation of function call expressions which results in #803).

So you're saying the existing approach of checking `if ctx.is_const()` is incorrect, right?

fw-immunant:
> > This has the effect of making various checks of (effectively) the form `if ctx.is_const() { return error }` trigger at various points in translating the expression where we recognize it to be incompatible with `const`. I don't think we need to merge the `is_const_expr` approach with this one, as the latter is fail-open (and does so in practice, e.g. in the translation of function call expressions which results in #803).
> 
> So you're saying the existing approach of checking `if ctx.is_const()` is incorrect, right?

It's currently known to be incorrect, e.g. it doesn't check function calls for constness, resulting in #803. But it's not fundamentally wrong to implement constness checking that way if we covered every possible case during translation, and for an implementation that wants to cover all cases I think it's probably the right way to implement it--we don't want a separate full AST traversal for each analysis because they would need to stay synchronized with respect to the particular way we translate every idiom so they can reason about the Rust we're generating. For now I think it's fine to have two parallel approaches (a conservative one and an aspirationally precise one) as long as we document what's going on.

fw-immunant (approved):
The comments in the last commit need some fixes (see my inline comments at https://github.com/immunant/c2rust/commit/604c58a557f18288027aaa4c30ba3cb78ad2157e), but that can be addressed in a follow up. These changes look good, but we should avoid making all translations of const macros unsafe now that we're doing them by default. The goal of the conservative const macro translation effort is to generate fully idiomatic code for cases simple enough for us to do so, so we should avoid regressing translation of the simplest cases like `#define FOO 5`.

So this can merge into `kkysen/const-macros-unsafe-block` if you like, but I'd rather not merge that branch into master while it adds unsafe to the simplest const macro translations. Merging `fw/const-less-unsafe` into this branch (or into `kkysen/const-macros-unsafe-block` after this merges into it) would be my suggested route to that.

The stacked PR workflow is confusing but hopefully the above makes sense.

kkysen (commented):
> The comments in the last commit need some fixes (see my inline comments at [604c58a](https://github.com/immunant/c2rust/commit/604c58a557f18288027aaa4c30ba3cb78ad2157e)), but that can be addressed in a follow up.

Fixed in [`81551c0..ecddc95`](https://github.com/immunant/c2rust/compare/81551c042638a1a8da67563e7e1646b6a5091a17..ecddc955d30f4a40135f279a6cdc050093229a2e).

PR #1308, opened 2025-08-07
Apply clippy fixes to stable crates

brk:
This applies a selection of fixes (mostly auto-applied, some manually implemented or suppressed) from Rust 1.88.0 which remain backwards-compatible with c2rust's pinned 1.65 nightly.

I've split them into separate commits under the assumption that it makes review slightly easier. I assume this branch would be squash merged.

### DIFF
--- a/c2rust-ast-exporter/src/lib.rs
+++ b/c2rust-ast-exporter/src/lib.rs
@@ -109,6 +109,7 @@ unsafe fn marshal_result(result: *const ffi::ExportResult) -> HashMap<String, Ve
         // Convert CBOR bytes
         let csize = *res.sizes.offset(i);
         let cbytes = *res.bytes.offset(i);
+        #[allow(clippy::unnecessary_cast /*, reason = "needed on x86_64-unknown-linux-gnu" */)]
         let bytes = slice::from_raw_parts(cbytes, csize as usize);
         let mut v = Vec::new();
         v.extend_from_slice(bytes);
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -2420,8 +2420,8 @@ mod tests {
                 let b = locs[j];
                 for c in locs.iter().take(n) {
                     let ab = ctx.compare_src_locs(&a, &b);
-                    let bc = ctx.compare_src_locs(&b, &c);
-                    let ac = ctx.compare_src_locs(&a, &c);
+                    let bc = ctx.compare_src_locs(&b, c);
+                    let ac = ctx.compare_src_locs(&a, c);
                     if ab == bc {
                         let [ab, bc, ac] = [ab, bc, ac].map(|ord| match ord {
                             Ordering::Less => "<",
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1602,49 +1602,23 @@ impl BinOp {
     #[rustfmt::skip]
     pub fn input_types_same(&self) -> bool {
         use BinOp::*;
-        self.all_types_same() || match self {
-            Less => true,
-            Greater => true,
-            LessEqual => true,
-            GreaterEqual => true,
-            EqualEqual => true,
-            NotEqual => true,
-
-            And => true,
-            Or => true,
-
-            AssignAdd => true,
-            AssignSubtract => true,
-            AssignMultiply => true,
-            AssignDivide => true,
-            AssignModulus => true,
-            AssignBitXor => true,
-            AssignShiftLeft => true,
-            AssignShiftRight => true,
-            AssignBitOr => true,
-            AssignBitAnd => true,
-
-            Assign => true,
-            _ => false,
-        }
+        self.all_types_same() || matches!(self,
+            Less | Greater | LessEqual | GreaterEqual | EqualEqual | NotEqual
+            | And | Or
+            | AssignAdd | AssignSubtract | AssignMultiply | AssignDivide | AssignModulus
+            | AssignBitXor | AssignShiftLeft | AssignShiftRight | AssignBitOr | AssignBitAnd
+            | Assign
+        )
     }
 
     /// Does the rust equivalent of this operator have type (T, T) -> T?
     /// This ignores cases where one argument is a pointer and we translate to `.offset()`.
     pub fn all_types_same(&self) -> bool {
         use BinOp::*;
-        match self {
-            Multiply => true,
-            Divide => true,
-            Modulus => true,
-            Add => true,
-            Subtract => true,
-
-            BitAnd => true,
-            BitXor => true,
-            BitOr => true,
-            _ => false,
-        }
+        matches!(
+            self,
+            Multiply | Divide | Modulus | Add | Subtract | BitAnd | BitXor | BitOr
+        )
     }
 }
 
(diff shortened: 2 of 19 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1600,6 +1600,7 @@ impl BinOp {
 
     /// Does the rust equivalent of this operator have type (T, T) -> U?
     #[rustfmt::skip]
+    #[allow(clippy::match_like_matches_macro)]
```
kkysen: I think you can change this one to a `matches!`, unless there's a reason you didn't.  And when/if you do, we should be able to remove the `#[rustfmt::skip]` as well.

c2rust-ast-exporter/src/lib.rs
```diff
@@ -95,6 +95,7 @@ extern "C" {
     fn clang_version() -> *const c_char;
 }
 
+#[allow(clippy::unnecessary_cast)]
```
kkysen: I believe you can put this on the specific line that needs it.  Also, can you add a `reason = ""`, but commented out with `/* , reason = "" */` since it's not stabilized yet (but will be soon after merging #1227)?  Something like "needed on aarch64-apple-darwin" seems good.

kkysen: Also, if `cfg_attr` works for this to only enable it when needed, even better.  Not sure that works, but it might.

c2rust-transpile/src/translator/assembly.rs
```diff
@@ -17,17 +17,20 @@ enum ArgDirSpec {
     InLateOut,
 }
 
-impl ToString for ArgDirSpec {
-    fn to_string(&self) -> String {
+impl std::fmt::Display for ArgDirSpec {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
kkysen: Can you import `Display`, `Formatter`, and `fmt::Result`?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4884,6 +4884,7 @@ impl<'c> Translation<'c> {
     }
 
     /// Convert a boolean expression to a boolean for use in && or || or if
+    #[allow(clippy::collapsible_match)]
```
kkysen: Is there a reason this one's skipped?  From a cursory glance, it seems like it'd be good.

kkysen (commented):
> I've split them into separate commits under the assumption that it makes review slightly easier. I assume this branch would be squash merged.

It does help with reviewing, so thanks!  I'd rather just merge the separate commits, as they seem separate for a reason (e.x. some auto-applied, some manual, etc.).  Could you squash in the `cargo fmt` to whichever commit it applies to, though?  Thanks!

c2rust-transpile/src/rust_ast/set_span.rs
```diff
@@ -186,7 +186,7 @@ set_span_impl!(enum ImplItem, s via
 
 impl SetSpan for Block {
     fn set_span(&mut self, s: Span) {
-        if self.stmts.is_empty() == false {
+        if !self.stmts.is_empty() {
```
bungcip, line 189: just my opinion, but i think clippy `bool_comparison` lint is wrong here. `!self.stmts.is_empty()` is double negation and using `== false` is more obvious. 
(https://github.com/rust-lang/rust-clippy/issues/4717)

c2rust-transpile/src/c_ast/mod.rs
brk: The separation of operators into comparisons, and/or, and assignments seemed meaningful. But if you're happy with using `matches!` that's easy enough.

c2rust-transpile/src/translator/assembly.rs
brk: Sure thing.

Quick question: would you like changes made as separate commits that are then `absorb`ed before merging, or should I just force-push a modified sequence of commits?

c2rust-transpile/src/translator/mod.rs
brk: As-is:

```rs
                if let Expr::Binary(ExprBinary { op, .. }) = *unparen(arg) {
                    match op {
                        BinOp::Or(_)
                        | BinOp::And(_)
                        | BinOp::Eq(_)
                        | BinOp::Ne(_)
                        | BinOp::Lt(_)
                        | BinOp::Le(_)
                        | BinOp::Gt(_)
                        | BinOp::Ge(_) => {
                            if target {
```

Collapsed:

```rs
                if let Expr::Binary(ExprBinary {
                    op:
                        BinOp::Or(_)
                        | BinOp::And(_)
                        | BinOp::Eq(_)
                        | BinOp::Ne(_)
                        | BinOp::Lt(_)
                        | BinOp::Le(_)
                        | BinOp::Gt(_)
                        | BinOp::Ge(_),
                    ..
                }) = *unparen(arg)
                {
                    if target {
```

Subjective, of course, but I think the formatting of the collapsed version is worse than the un-collapsed version. Keeping the `ExprBinary` and `unparen(arg)` on the same line seems desirable. And the collapsed version has three mostly-empty lines.

c2rust-transpile/src/rust_ast/set_span.rs
brk: The argument in the issue you linked is about readability with long method chains, which doesn't seem to apply here. And I'm not sure what you mean by double negation -- "empty" is a state predicate, so "not empty" is a single negation. 

In any case, I have no particular preference between addressing the clippy lint via its suggested fix or via an explicit `allow`.

bungcip: i think its ok to follow default suggestion from clippy in `c2rust` project. I write like that `if self.stmts.is_empty() == false {` because habit and i usualy set `bool_comparison` to allow in all my rust project 😉

c2rust-transpile/src/c_ast/mod.rs
kkysen: You could do 4 `matches!` then.  It's a pretty simple function so I'm not too worried.

c2rust-transpile/src/translator/assembly.rs
kkysen: Small things like this you can amend commits.  There's no hard rule, though, so use your judgement.  If it seems more substantial you can just add a new commit.  And just try to keep each commit passing CI, which helps with bisecting if we need to do that.

c2rust-transpile/src/rust_ast/set_span.rs
kkysen: `== false` feels wrong to me, although `match` `bool` I like even though `clippy` doesn't.  Not really a big deal, though, so I just follow `clippy`.

c2rust-transpile/src/translator/mod.rs
kkysen: Both of these look okay to me.  I don't have too strong of a preference here.  If you prefer to leave it as is, can you move the `#[allow]` to the one expression and leave a (commented out for now) `reason`?  You can also leave a comment that this should be an `#[expect]`, but our old nightly doesn't support that yet.

kkysen:
Sorry, can you rebase?  I just merged a fix/reversion for CI.

brk:
@kkysen I believe the re-pushed commits address your requests. I also rebased against the previous master branch. Will rebase again!

kkysen (approved):
LGTM now! Thanks!

PR #1311, opened 2025-08-07
transpile: const macros: do not resolve expr type

### LINKED ISSUE #853: Missing casts in const macro translation

fw-immunant:
Fixes #853. Still needs tests.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2184,11 +2184,8 @@ impl<'c> Translation<'c> {
                     .kind
                     .get_type()
                     .ok_or_else(|| format_err!("Invalid expression type"))?;
-                let (expr_id, ty) = self
-                    .ast_context
-                    .resolve_expr_type_id(id)
-                    .unwrap_or((id, ty));
-                let expr = self.convert_expr(ctx, expr_id, None)?;
+                let ty = self.ast_context[id].kind.get_type().unwrap_or(ty);
+                let expr = self.convert_expr(ctx, id, None)?;
 
                 // Join ty and cur_ty to the smaller of the two types. If the
                 // types are not cast-compatible, abort the fold.


fw-immunant:
After cherry-picking:
d0027046b9c2a73c105a132c477e87c0c2a55c9c
6e4ebd41f37539c800e39376c7aaf821ac7a4c45
498f13040e21ee86e1df103c259445d7a9ae6ba6
f8d0e9ec64f6950787d19d81a683ebbf8812bfab

onto this branch and running `c2rust-transpile --overwrite-existing --translate-const-macros c2rust-transpile/tests/snapshots/macros.c` I see that test translated as follows:

```rust
pub const silk_int16_MIN: std::ffi::c_short = 0x8000 as std::ffi::c_int
    as std::ffi::c_short;
#[no_mangle]
pub unsafe extern "C" fn test_silk_int16_MIN() -> std::ffi::c_int {
    let mut _null: std::ffi::c_char = (*::core::mem::transmute::<
        &[u8; 1],
        &[std::ffi::c_char; 1],
    >(b"\0"))[(silk_int16_MIN as std::ffi::c_int + 0x8000 as std::ffi::c_int) as usize];
    return silk_int16_MIN as std::ffi::c_int;
}
```

That looks sane to me (the const is -32768). Am I missing something?

kkysen:
Hmm, that is correct.  I just wasn't seeing that in the snapshot.  If you rebase onto that branch and check in the updated snapshot, it should be all good.  It just wasn't working for me for some reason.

kkysen:
@fw-immunant, can I rebase this?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2174,11 +2174,8 @@ impl<'c> Translation<'c> {
                     .kind
                     .get_type()
                     .ok_or_else(|| format_err!("Invalid expression type"))?;
-                let (expr_id, ty) = self
-                    .ast_context
-                    .resolve_expr_type_id(id)
-                    .unwrap_or((id, ty));
-                let expr = self.convert_expr(ctx, expr_id, None)?;
+                let ty = self.ast_context[id].kind.get_type().unwrap_or(ty);
```
kkysen: What's the difference between this and the above

```rust
                let ty = self.ast_context[id]
                    .kind
                    .get_type()
                    .ok_or_else(|| format_err!("Invalid expression type"))?;
```

?

fw-immunant: Oh, good catch--there really isn't one. I guess this can be dropped; all we needed to do was remove the unnecessary `resolve_expr_type_id`.

fw-immunant:
Here's my reasoning around this change:

`resolve_expr_type_id` is documented to "Resolve true expression type, iterating through any casts and variable references.". From what I can tell, it does this correctly, though one has to consider any outer casts not to be real parts of the expression. So for an expression like `(double)(char)500` it will give back the expression and type of `500`.

We call this in `recreate_const_macro_from_expansions`, presumably to have a better chance of looking at not the surrounding casts in context but instead at the type of the macro expansion itself. But the expression pointed at by this is not necessarily going to have the same value as the original one; `500 != (char)500`. So it isn't valid to rewrite all uses of a macro with the expression that `resolve_expr_type_id` gives back.

If we avoid calling `resolve_expr_type_id`, we stick with an expression that has the correct value, but it may have a type unduly influenced by implicit casts in its context, e.g. what we see in the changed snapshot where the `true` macro is translated as a `double` due to its usage in a floating-point arithmetic expression in `MIXED_ARITHMETIC LITERAL_INT`.

I'll look into whether there's some way to determine which casts are caused by the expansion context as opposed to being part of the macro itself, as the former as the only ones we should strip off to generate the translation for the macro body.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2525,22 +2521,30 @@ impl<'c> Translation<'c> {
...
-                            is_null
-                        }
-                    }
-                }))
+                            if ctx.is_const {
+                                return Err(format_translation_err!(
+                                    None,
+                                    "cannot check nullity of pointer in `const` context",
+                                ));
+                            }
```
kkysen: [`pointer::is_null`](https://doc.rust-lang.org/core/primitive.pointer.html#method.is_null) is `const` now as of 1.84.  Should probably at least note that here.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4424,6 +4424,12 @@ impl<'c> Translation<'c> {
             | CastKind::FloatingToIntegral
             | CastKind::IntegralToFloating
             | CastKind::BooleanToSignedIntegral => {
+                if kind == CastKind::PointerToIntegral && ctx.is_const {
+                    return Err(format_translation_err!(
+                        None,
+                        "cannot observe pointer values in `const` context",
+                    ));
+                }
```
kkysen: Do we need to add this to `fn is_const_expr` as well?  I hadn't considered this.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2089,8 +2089,9 @@ impl CTypeKind {
             PtrDiff => "ptrdiff_t",
             WChar => "wchar_t",
 
-            _ => unimplemented!("Printer::print_type({:?})", self),
+            _ => return format!("Printer::print_type({:?})", self),
```
kkysen: This should happen in the `impl Display` instead (where we currently just call `self.as_str()`).  `fn as_str` implies a cheap conversion to `&str`, not allocating a `String`.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
kkysen, line 1: The output here looks a whole lot nicer now.

fw-immunant:
This is now passing tests again and should be ready for review/merge.

c2rust-transpile/src/c_ast/mod.rs
fw-immunant: Commit removed from this PR, it was just WIP for debugging.

c2rust-transpile/src/translator/mod.rs
fw-immunant: Done.

c2rust-transpile/src/translator/mod.rs
fw-immunant: I'm not sure. Is there a test case we could add that would make the problem clear?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2525,22 +2521,31 @@ impl<'c> Translation<'c> {
...
+                                mk().method_call_expr(e, "is_some", vec![])
+                            } else {
+                                mk().method_call_expr(e, "is_none", vec![])
+                            }
                         } else {
-                            is_null
-                        }
-                    }
-                }))
+                            // TODO: `ptr::is_null` becomes stably const in Rust 1.84.
```
```suggestion
                            // TODO: `pointer::is_null` becomes stably const in Rust 1.84.
```

It's an inherent method, not in the `std::ptr` module.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2525,22 +2521,31 @@ impl<'c> Translation<'c> {
...
+                                    None,
+                                    "cannot check nullity of pointer in `const` context",
+                                ));
+                            }
+                            let is_null = mk().method_call_expr(e, "is_null", vec![]);
+                            if negated {
+                                mk().unary_expr(UnOp::Not(Default::default()), is_null)
+                            } else {
+                                is_null
+                            }
```
kkysen: How come we can't use `fn match_bool` here?

c2rust-transpile/tests/snapshots/macrocase.c
```diff
@@ -0,0 +1,21 @@
+#define ZSTD_d_format ZSTD_d_experimentalParam1
```
kkysen, line 2: Can you also add a comment saying where this is from (like the file or something)?

c2rust-transpile/tests/snapshots/snapshots__transpile-linux@macros.c.snap
```diff
@@ -37,7 +37,7 @@ pub const POS: [std::ffi::c_char; 3] =
 pub unsafe extern "C" fn memcpy_str_literal(mut out: *mut std::ffi::c_char) {
     memcpy(
         out as *mut std::ffi::c_void,
-        POS.as_ptr() as *const std::ffi::c_void,
+        b"\"]\0" as *const u8 as *const std::ffi::c_char as *const std::ffi::c_void,
```
kkysen: What's happening here?  `POS` is still being generated, but not used, and this is a worse translation, right?

c2rust-transpile/tests/snapshots/snapshots__transpile@macrocase.c.snap
```diff
@@ -0,0 +1,34 @@
...
+pub const ZSTD_d_format: std::ffi::c_uint = unsafe { 1000 as std::ffi::c_uint };
+#[no_mangle]
+pub unsafe extern "C" fn ZSTD_dParam_getBounds(mut dParam: ZSTD_dParameter) -> std::ffi::c_int {
+    let mut bounds: std::ffi::c_int = 0 as std::ffi::c_int;
+    match dParam as std::ffi::c_uint {
+        100 => {
+            bounds = 1 as std::ffi::c_int;
+            return bounds;
+        }
+        1000 => {
```
kkysen, line 26: What's (not) happening with the patterns here?

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
```diff
@@ -47,34 +47,23 @@ pub const LITERAL_STRUCT: S = unsafe {
...
-    }
-};
+pub const NESTED_CHAR: std::ffi::c_int = unsafe { LITERAL_CHAR };
+pub const NESTED_STR: [std::ffi::c_char; 6] = unsafe { LITERAL_STR };
+pub const NESTED_STRUCT: S = unsafe { LITERAL_STRUCT };
 pub const PARENS: std::ffi::c_int = unsafe { NESTED_INT * (LITERAL_CHAR + true_0) };
 pub const PTR_ARITHMETIC: *const std::ffi::c_char = unsafe {
-    LITERAL_STR
-        .as_ptr()
+    (b"hello\0" as *const u8 as *const std::ffi::c_char)
```
kkysen: It seems str literals have regressed.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -504,6 +504,48 @@ impl ConversionContext {
...
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+            move |e| case_exprs.contains(&e)
+        };
+
+        let mut macro_invocations = IndexMap::new();
+        std::mem::swap(
```
kkysen: nit: I think we usually just `use std::mem`.

```suggestion
        mem::swap(
```

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -504,6 +504,48 @@ impl ConversionContext {
             }
         }
 
+        // Adjust macro expansions to skip any implicit casts added at the expansion site. These
+        // casts should not be considered part of the macro when we re-fold const macros.
+
+        // We *do* need to consider implicit casts in case expressions to be part of the macro,
+        // because the type of the `match` scrutinee must match the type of the cases in Rust.
+        // If we translated switch-case labels to use const macros, those would need to prefer the
+        // type of the match scrutinee, to which the macro is implicitly cast in case expressions.
```
kkysen, line 513: It seems match patterns aren't getting translated as macros at all, though.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -504,6 +504,48 @@ impl ConversionContext {
...
+                })
+                .collect();
+            move |e| case_exprs.contains(&e)
+        };
+
+        let mut macro_invocations = IndexMap::new();
+        std::mem::swap(
+            &mut macro_invocations,
+            &mut self.typed_context.macro_invocations,
+        );
```
kkysen:
```suggestion
        let mut macro_invocations = mem::take(&mut macro_invocations);
```

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -433,6 +433,19 @@ impl TypedAstContext {
         }
     }
 
+    /// Find underlying expression beneath any implicit casts.
+    pub fn beneath_implicit_casts(&self, expr_id: CExprId) -> CExprId {
+        let expr = &self.index(expr_id).kind;
+        use CExprKind::*;
+        match expr {
+            ImplicitCast(_, subexpr, _, _, _) => {
```
kkysen, line 441: If it's only one match arm and not a ton of them, a glob import probably isn't worth it.

```suggestion
        match expr {
            CExprKind::ImplicitCast(_, subexpr, _, _, _) => {
```

kkysen: Could also be an `if let`.  This likely won't change, right, as it's pretty specific to implicit casts?

c2rust-transpile/src/translator/mod.rs
kkysen: Let me see.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4414,6 +4414,12 @@ impl<'c> Translation<'c> {
                 val.and_then(|x| {
                     let intptr_t = mk().path_ty(vec!["libc", "intptr_t"]);
                     let intptr = mk().cast_expr(x, intptr_t.clone());
+                    if ctx.is_const {
+                        return Err(format_translation_err!(
+                            None,
+                            "cannot transmute integers to Option<fn ...> in `const` context",
+                        ));
+                    }
```
kkysen, line 4451: Same as for the others, we should probably add a test for these `const` forbidden things.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4702,7 +4702,7 @@ impl<'c> Translation<'c> {
             CExprKind::DeclRef(_, decl_id, _) if variants.contains(&decl_id) => {
                 return val.map(|x| match *unparen(&x) {
                     Expr::Cast(ExprCast { ref expr, .. }) => expr.clone(),
-                    _ => panic!("DeclRef {:?} of enum {:?} is not cast", expr, enum_decl),
+                    _ => mk().cast_expr(x, target_ty),
```
kkysen: Can you add this backup handling to `fn enum_cast`'s docs?

c2rust-transpile/src/translator/mod.rs
fw-immunant: I think we plausibly could; I wasn't looking for refactorings here, and the code is slightly different (here we only handle pointer types).

c2rust-transpile/src/c_ast/conversion.rs
fw-immunant: Done.

c2rust-transpile/tests/snapshots/macrocase.c
fw-immunant: Done.

c2rust-transpile/src/translator/mod.rs
fw-immunant: These have coverage in the test suite, but we could add a snapshot test.

c2rust-transpile/tests/snapshots/snapshots__transpile-linux@macros.c.snap
fw-immunant: Addressed in "transpile: const macros: check for const macro expansion when translating two levels deep in casts".

Basically, translation of `ArrayToPointerDecay` casts was looking through the next level of expression nesting for literals and skipping the chance to instead translate that literal as an existing macro, which only happens if it gets `convert_expr` called on it.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
fw-immunant: Addressed in "transpile: const macros: check for const macro expansion when translating two levels deep in casts".

c2rust-transpile/src/translator/mod.rs
fw-immunant: Done.

c2rust-transpile/tests/snapshots/macrocase.c
fw-immunant: Done.

c2rust-transpile/tests/snapshots/snapshots__transpile@macrocase.c.snap
fw-immunant: We need to avoid translating switch cases to const macros, because those consts may have more interesting types than plain integers, but we cannot cast those types to the same type as the switch-case scrutinee (which is always an integer) within Rust patterns. The right way to support this would require changing the type of the generated match's scrutinee; it would be nice to do, but for now this tests that we haven't introduced a regression.

c2rust-transpile/src/c_ast/conversion.rs
fw-immunant: Yes; this logic prevents switch cases from being translated as const macros, because we need to do other work before that's possible.

c2rust-transpile/src/translator/mod.rs
fw-immunant: I added a comment inline and restricted this to only the `Expr::Path` case corresponding to a const macro.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4516,9 +4516,22 @@ impl<'c> Translation<'c> {
 
                 let is_const = pointee.qualifiers.is_const;
 
+                // Handle literals by looking at the next level of expr nesting. Avoid doing this
+                // for expressions that will be translated as const macros, because emitting the
+                // name of the const macro only occurs if we process the expr_id with a direct call
+                // to `convert_expr`.
```
kkysen, line 4556: When/where does this happen?  That might've been part of what was tripping me up in adding `override_ty`s to const macros.

fw-immunant: At the beginning of `convert_expr`:
```rs
    pub fn convert_expr(
        &self,
        mut ctx: ExprContext,
        expr_id: CExprId,
        override_ty: Option<CQualTypeId>,
        let Located {
            loc: src_loc,
            kind: expr_kind,
        } = &self.ast_context[expr_id];

        trace!(
            "Converting expr {:?}: {:?}",
            expr_id,
            self.ast_context[expr_id]
        );

        if let Some(converted) = self.convert_const_macro_expansion(ctx, expr_id, override_ty)? {
            return Ok(converted);
        }
```

c2rust-transpile/src/c_ast/conversion.rs
kkysen: Ah, I misunderstood.

c2rust-transpile/src/translator/mod.rs
kkysen: Oh, sorry, I meant the opposite.  When do we not go through `fn convert_expr` to process an `expr_id`?

c2rust-transpile/src/translator/mod.rs
kkysen: We should add a snapshot test, too, because the `tests/` tests run with `--translate-const-macros experimental`, not `conservative`.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4498,9 +4516,22 @@ impl<'c> Translation<'c> {
...
                 let expr_kind = expr.map(|e| &self.ast_context.index(e).kind);
+                let translate_as_macro = expr
+                    .map(|e| {
+                        self.convert_const_macro_expansion(ctx, e, None)
+                            .transpose()
+                            .is_some()
```
kkysen, line 4563: Why do we want to consider `Err`s as translated as macro?

c2rust-transpile/src/translator/mod.rs
fw-immunant: The `tests/` tests only run with `experimental` if they have the `//! translate_const_macros` flag  at their top, but I'll see about a snapshot test.

kkysen: Oh yeah, true.

c2rust-transpile/src/translator/mod.rs
fw-immunant: I think you're right, `.ok().flatten().is_some()` is probably more correct here. I was flying by types a bit too much.

c2rust-transpile/src/translator/mod.rs
fw-immunant: Whenever we look extra levels down in the expression tree and short-circuit instead of recursing in `convert_expr`; there isn't a concise answer here, but it mostly happens toward leaves of the tree.

fw-immunant:
[Hit our intermittent CI failure](https://github.com/immunant/c2rust/actions/runs/18043943959/job/51349928012?pr=1311) (#1365)...

c2rust-transpile/src/translator/mod.rs
fw-immunant: I added a test for this case, but the other two places where I added translation failures in const context seem to be very hard to trigger. Maybe it makes sense to include those commits despite lack of test coverage for now, instead of blocking this indefinitely.

kkysen: Yeah, I think that's fine.

c2rust-transpile/tests/snapshots/snapshots__transpile-linux@sigign.c.snap
```diff
@@ -0,0 +1,23 @@
...
+    unused_mut
+)]
+pub type __sighandler_t = Option<unsafe extern "C" fn(std::ffi::c_int) -> ()>;
+#[no_mangle]
+pub unsafe extern "C" fn foo() {
+    let mut handler: Option<unsafe extern "C" fn(std::ffi::c_int) -> ()> = ::core::mem::transmute::<
+        libc::intptr_t,
+        Option<unsafe extern "C" fn(std::ffi::c_int) -> ()>,
+    >(
+        1 as std::ffi::c_int as libc::intptr_t,
```
kkysen, line 21: How come `SIG_IGN` isn't translated as a `const`?

fw-immunant: It isn't allowed to transmute from an integer to an `Option<fn>` in a const, so this is intentional: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2d1f5c7d28e3bd8b5a9166bcd1ac6611

PR #1313, opened 2025-08-08
Revert "transpile: make `--translate-{const,fn}-macros conservative` the default for the CLI, too"

kkysen:
This reverts commit 6c689e17e89e9feac90d4e44863f0fdf891e1eb6.

This broke CI (`c2rust-testsuite`), which wasn't caught because CI wasn't run on PRs not directly into `master`.

* #1312 fixes this.

### DIFF
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -46,10 +46,10 @@ type TranspileResult = Result<(PathBuf, PragmaVec, CrateSet), ()>;
 #[derive(Default, Debug)]
 pub enum TranslateMacros {
     /// Don't translate any macros.
+    #[default]
     None,
 
     /// Translate the conservative subset of macros known to always work.
-    #[default]
     Conservative,
 
     /// Try to translate more, but this is experimental and not guaranteed to work.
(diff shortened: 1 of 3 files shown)


fw-immunant:
Making PRs into branches other than master always confuses me. It means that reviewing requires considering the diff of the PR *and* the diff between the PR's destination and master.

fw-immunant:
Shouldn't we revert the whole PR, not just one commit? Based on the [CI run](https://github.com/immunant/c2rust/actions/runs/16822098042/job/47650990096) it looks like we have actual breakage that we should have fixed before merging the `--translate-const-macros` changes; the particular commit here still looks correct to me in context.

kkysen (commented):
> Shouldn't we revert the whole PR, not just one commit? Based on the [CI run](https://github.com/immunant/c2rust/actions/runs/16822098042/job/47650990096) it looks like we have actual breakage that we should have fixed before merging the `--translate-const-macros` changes; the particular commit here still looks correct to me in context.

The breakage is fundamentally from an earlier commit in the PR, but it's less churn for me to just revert this one, fix it, and then un-revert it.  Can I merge this PR now?  Might not be ideal, but it should just get things done quickly.

kkysen:
I'm going to merge this so CI stops being broken.

PR #1323, opened 2025-08-13
transpile: update `fn convert_expr` docs and remove unused code

kkysen:
Some doc updates I made while doing #1322.  One of the types previously used is no longer used anywhere, so I removed it.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3316,11 +3316,11 @@ impl<'c> Translation<'c> {
                 Ok(result)
             }
 
-            ConstantExpr(_ty, child, value) => {
+            ConstantExpr(ty, child, value) => {
                 if let Some(constant) = value {
                     self.convert_constant(constant).map(WithStmts::new_val)
                 } else {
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -1132,31 +1132,6 @@ pub(crate) fn unparen(expr: &Expr) -> &Expr {
...
-///     Then, `x = n` is a volatile read of `n` but `p = &n` is not. We get this behaviour
-///     by translating the argument of `&` using `ExprUse::LValue` and the right hand side of `=`
-///
-///     using `ExprUse::RValue`.
-///
-///   * handling `volatile` properly
-///
-/// See `Translation::convert_expr` for more details.
-#[derive(Copy, Clone, Debug, PartialOrd, PartialEq, Ord, Eq)]
-pub enum ExprUse {
```
ahomescu, line 1153: Is this completely unused now? In that case, maybe the PR description should be updated since it's not just a documentation change.

kkysen: Yeah, not used anywhere. It was `pub` so it wasn't flagged as unused once we stopped using it.

PR #1334, opened 2025-08-21
transpile: make `TypedAstContext::macro_{invocations,expansions,expansion_test}` into `IndexMap`s instead of `HashMap`s

kkysen:
* Split out of #1306.

Previously, `macro_invocations` was a `HashMap`, and thus iterating through it was unordered, which populated the `Vec<CExprId>` of `macro_expansions` non-deterministically, which then resulted in non-deterministic output from `--translate-const-macros conservative`.

I also changed the other `macro_*` maps to `IndexMap`, as many other maps in `TypedAstContext` are already `IndexMap`s, too, and it's likely that we want these stably ordered and deterministic.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -71,14 +71,14 @@ pub struct TypedAstContext {
     pub label_names: IndexMap<CLabelId, Rc<str>>,
 
     // map expressions to the stack of macros they were expanded from
-    pub macro_invocations: HashMap<CExprId, Vec<CDeclId>>,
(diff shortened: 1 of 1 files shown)


kkysen (commented):
@fw-immunant, I think you previously said this looked good, so if you could review quickly and we can merge it, that'd be great.  I don't want #1306 to get too large, so I'd like to merge the simpler parts of it first.

kkysen (commented):
> It seems like the target branch should be `master`, but this change does seem reasonable.

It was #1337 just to speed up CI as I was rebasing a lot and running a lot of slow CI jobs.  It points into `master` now.

PR #1335, opened 2025-08-21
transpile: unconditionally make `const`s use `unsafe` blocks for `--translate-const-macros conservative`

kkysen:
* Split out of #1306.

Some operations, such as ptr arithmetic, are `unsafe` and can be done in const macros.  So as an initially overly conservative implementation, just make all `const`s use `unsafe` blocks in case `unsafe` operations are used.  This is what we already do for `fn`s, for example, even if all of the operations in a `fn` are safe.  We can improve this, but for now, this works and is correct.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2152,7 +2152,7 @@ impl<'c> Translation<'c> {
         ctx: ExprContext,
         expansions: &[CExprId],
     ) -> TranslationResult<(Box<Expr>, CTypeId)> {
-        let (val, ty) = expansions
+        let (mut val, ty) = expansions
             .iter()
             .try_fold::<Option<(WithStmts<Box<Expr>>, CTypeId)>, _, _>(None, |canonical, &id| {
                 self.can_convert_const_macro_expansion(id)?;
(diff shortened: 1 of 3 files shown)


kkysen (commented):
@fw-immunant, same as https://github.com/immunant/c2rust/pull/1334#pullrequestreview-3138620388.

fw-immunant:
> > This makes our translation of a simple `#define FOO 5` less idiomatic, which seems contrary to the goal of providing fully idiomatic translation for the simplest cases of `#define`s.
> 
> I'm not sure if the goal was to provide fully idiomatic translation for the simplest `#define`s. We've never strived for `c2rust transpile` emitting fully idiomatic code, but rather, as idiomatic as reasonably possible. I'd like to merge this first as the simplest way to translate more const macros, and then we can work on things like #1339 on top of this (I'll leave more comments there). You made #1339 point at #1306, which points at #1335 (this PR), which seems like the right ordering to me, so if that's what you intended, then can we merge this PR now?

That's what I understood from the phrase "low-hanging fruit"--the simple cases where we can get the exact right answer (fully idiomatic translation) without necessarily having a solution that can handle the full generality of inputs. Maybe @thedataking could clarify what our priorities are here.

I'd rather not merge this because it makes translation of simple uses of `#define` noticeably less idiomatic, which will be visible to any users of master (e.g. godbolt.org afaiu). #1339 reverts this change, and my intent was to merge 1339 into its target branch rather than master--if I wanted to merge it into master I would have set its PR target to that.

#1306 is approved and can merge into this branch whenever you like.

PR #1338, opened 2025-08-21
transpile: expand `--translate-const-macros conservative` to handle statements

kkysen:
This handles all statements except for `goto`s and their labels.

`goto`s are tricky, because they can be non-local and jump out of the context of the macro.  A `goto` and its labels are `const` if the whole state machine we compile to has all `const` statements, but determining what that is exactly is trickier, and might depend on the context in which the macro is used.  This is probably fairly uncommon, so we just assume it's not `const` for now.

Note that in C, labels are for `goto`s.  There are no labeled `break`s and `continue`s.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -691,9 +691,52 @@ impl TypedAstContext {
         }
     }
 
-    pub fn is_const_stmt(&self, _stmt: CStmtId) -> bool {
-        // TODO
-        false
+    pub fn is_const_stmt(&self, stmt: CStmtId) -> bool {
+        let is_const = |stmt| self.is_const_stmt(stmt);
+        let is_const_expr = |expr| self.is_const_expr(expr);
+
+        use CStmtKind::*;
+        match self[stmt].kind {
+            Label(stmt) => is_const(stmt),
+            Case(expr, stmt, _const_expr) => is_const_expr(expr) && is_const(stmt),
+            Default(stmt) => is_const(stmt),
+            Compound(ref stmts) => stmts.iter().copied().all(is_const),
+            Expr(expr) => is_const_expr(expr),
+            Empty => true,
+            If {
+                scrutinee,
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -691,9 +691,52 @@ impl TypedAstContext {
...
+                condition,
+                increment,
+                body,
+            } => {
+                init.map_or(true, is_const)
+                    && condition.map_or(true, is_const_expr)
+                    && increment.map_or(true, is_const_expr)
+                    && is_const(body)
+            }
+            Goto(label) => is_const(label),
```
ahomescu: If I'm reading this correctly, a `goto` is const if the target statement (the one right after the label) is also const? That seems like a pretty rough approximation, not the actual rule we want.

kkysen: What's the rule?  Isn't it `const` if the statement is `const`?  The goto label itself doesn't make anything not `const`.

ahomescu: Example I had in mind:
```C
int main(void) {
foo: 3;
    printf("something!\n");
    goto foo;
}
```

Here the algorithm will look at `goto foo` and `foo: 3;` and conclude they're both `const`, and I'm wondering if maybe the `goto` shouldn't be because of the `printf`.

Otoh I think the algorithm winds up doing the right thing, because you can't jump outside of the current function so either the whole thing is const, or none of it is.

kkysen: After discussing this offline, we decided we'll just skip `goto`s and their labels for now, treating them as non-`const`.  Determining if it's `const` requires knowing if the surrounding context is also `const`, which is trickier to determine, and for a macro, might not even be available.  `goto`s in macros are probably also fairly uncommon, so we shouldn't miss too many cases from skipping this.

This is done in bba6c01.

ahomescu (approved):

PR #1339, opened 2025-08-22
Avoid `unsafe` in const macro translations except where necessary

fw-immunant:
Atop #1306, as I'd rather not regress all const translations to be `unsafe` in master now that we translate const macros by default.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2700,9 +2700,9 @@ impl<'c> Translation<'c> {
                 let mut init = init?;
 
                 stmts.append(init.stmts_mut());
-                // A const context is not "already unsafe" the way code within a `fn` is (since we
-                // translate all fns as unsafe). Therefore, in const context expose any underlying
-                // unsafety in the initializer with an unsafe block.
+                // A `const` context is not "already unsafe" the way code within a `fn` is
+                // (since we translate all `fn`s as `unsafe`). Therefore, in `const` contexts,
+                // expose any underlying unsafety in the initializer with an `unsafe` block.
                 let init = if ctx.is_const {
                     init.to_unsafe_pure_expr()
                         .expect("init should not have any statements")
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3764,21 +3764,26 @@ impl<'c> Translation<'c> {
                         };
 
                         let lhs = self.convert_expr(ctx.used(), arr, None)?;
-                        Ok(lhs.map(|lhs| {
+                        lhs.and_then(|lhs| {
                             // stmts.extend(lhs.stmts_mut());
                             // is_unsafe = is_unsafe || lhs.is_unsafe();
 
                             // Don't dereference the offset if we're still within the variable portion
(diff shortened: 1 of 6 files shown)


kkysen (commented):
See https://github.com/immunant/c2rust/pull/1335#pullrequestreview-3150243961, too, but in general, what's the difference between the `unsafe` block for a `const` and the `unsafe` on an `unsafe fn` (which in the latest edition is meant to also be an `unsafe` block in the `unsafe fn`, too)?  They seem exactly the same to me, and I think we should approach them the same.  So a change like this (which I haven't looked through all of the details of yet) makes sense to me if it also applies to an `unsafe fn`.  Note that there are often a bunch of `unsafe fn`s that are fully safe and don't need the `unsafe`, and removing the `unsafe` for a `const` that doesn't need it and remove the `unsafe` for a `fn` that doesn't need it seem like the exact same problem.

fw-immunant:
> See [#1335 (review)](https://github.com/immunant/c2rust/pull/1335#pullrequestreview-3150243961), too, but in general, what's the difference between the `unsafe` block for a `const` and the `unsafe` on an `unsafe fn` (which in the latest edition is meant to also be an `unsafe` block in the `unsafe fn`, too)? They seem exactly the same to me, and I think we should approach them the same. So a change like this (which I haven't looked through all of the details of yet) makes sense to me if it also applies to an `unsafe fn`. Note that there are often a bunch of `unsafe fn`s that are fully safe and don't need the `unsafe`, and removing the `unsafe` for a `const` that doesn't need it and remove the `unsafe` for a `fn` that doesn't need it seem like the exact same problem.

The difference is that there is no unsafe block surrounding definitions of `const`s prior to #1335 (nor after the first commit of this PR, `transpile: revert set consts unsafe`). This change (``transpile: in const contexts, use `unsafe` around unsafe initializers``) is really about following the same logic for initializers as we already follow for default zero values, [several lines below](https://github.com/immunant/c2rust/blob/7e193ffd1a38d8f67a45733a5d942270df23ec5f/c2rust-transpile/src/translator/mod.rs#L2715-L2720): prior to this commit, our translation here follows the `TranslationContext`'s notion of whether we're in a const context for the default zero values but doesn't for initializers.

When we change the edition we target, it will make sense to use `WithStmts::to_unsafe_pure_expr` in more places and we won't need to dispatch on the constness of the context to decide between `to_unsafe_pure_expr` and `into_value` anymore. The precision that this PR adds w/r/t which generated expressions are unsafe will also be useful for the translation of function bodies at that point.

fw-immunant:
History is cleaned up now; this should be good to merge.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2708,7 +2707,15 @@ impl<'c> Translation<'c> {
                 let mut init = init?;
 
                 stmts.append(init.stmts_mut());
-                let init = init.into_value();
+                // A const context is not "already unsafe" the way code within a `fn` is (since we
+                // translate all fns as unsafe). Therefore, in const context expose any underlying
```
```suggestion
                // translate all fns as unsafe). Therefore, in const context, expose any underlying
```

kkysen (approved):

c2rust-transpile/src/translator/mod.rs
kkysen: Fixed in 6bef46c.

PR #1340, opened 2025-08-22
Remove extraneous separator from paths with arguments

brk:
`compute_size_of_ty()` constructs a path using `path_segment_with_args()`, so that e.g. `sizeof(int)` gets translated to `::core::mem::size_of::<libc::c_int>()`.

But this code in ast-builder was adding an extra separator token to the path when a path segment has an argument. The resulting token stream for the call example given before thus corresponded to `::core::mem::size_of::<libc::c_int> :: ()`. You can verify this by adding a debug print log to show the converted value of a conditional and then translating code like `(sizeof(int) > 0) ? 0 : 0)`. I haven't dug into why, but (unless I'm mistaken about what's doing what here) `prettyplease` masks this bug by not including the syntactically incorrect trailing separator in its string output.

The incorrect syntax can be exposed when converting the invalid path to a token stream. Or, put another way, passing such expressions to a macro exposes the bad syntax.

Without this patch, translating

```c
#include <stddef.h>
#include <stdio.h>
struct s { int p[512]; };
int foo(unsigned char x) {
	return offsetof(struct s, p[sizeof(int) + x]);
}
int main(int argc, char** argv) {
    printf("%d\n", foo(argc));
    return 0;
}
```

gives `foo` as

```rs
#[no_mangle]
pub unsafe extern "C" fn foo(mut x: libc::c_uchar) -> libc::c_int {
    return offset_of!(
        s, p[(::core::mem::size_of:: < libc::c_int > :: () as libc::c_ulong)
                                                  // ^^ this guy right here!
        .wrapping_add(x as libc::c_ulong) as usize]
    ) as libc::c_ulong as libc::c_int;
}
```

Anyways, it appears that every use of `path_segment_with_args` in c2rust comes at the end of a path. I don't know if the extra separator would be needed for segments that are not at the end. Maybe there are other ways of getting paths with args that I didn't look for? I tried running the `c2rust-testsuite` but it seems to want a very specific machine config. So testing in CI it is...

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -315,13 +315,7 @@ impl<S: Make<PathSegment>> Make<Path> for Vec<S> {
     fn make(self, mk: &Builder) -> Path {
         let mut segments = Punctuated::new();
         for s in self {
-            let segment = s.make(mk);
-            let has_params = !segment.arguments.is_empty();
-            segments.push(segment);
-            // separate params from their segment with ::
-            if has_params {
-                segments.push_punct(Default::default());
-            }
+            segments.push(s.make(mk));
         }
         Path {
             leading_colon: None,
--- a/c2rust-transpile/tests/tokens.rs
+++ b/c2rust-transpile/tests/tokens.rs
@@ -1,0 +1,13 @@
+use c2rust_ast_builder::mk;
+use syn::__private::ToTokens;
+
+#[test]
+fn test_tokenize() {
+    let tys = vec![mk().path_ty(vec!["t"])];
+    let args = mk().angle_bracketed_args(tys);
+    let path_segment = mk().path_segment_with_args("x", args);
+    assert_eq!(path_segment.to_token_stream().to_string(), "x :: < t >");
+
+    let path = mk().path(vec![path_segment]);
+    assert_eq!(path.to_token_stream().to_string(), "x :: < t >");
+}


bungcip:
can you add the problematic code in snapshot test?

brk:
I looked in to this but it was a big yak shave. `std::mem::offset_of` wasn't stabilized until 1.77, so it's not usable on the the pinned nightly. And even with `self.use_crate(ExternCrate::Memoffset);` it seems that insta tests don't get access to extern crates. I'm not sure why. The errors look like this:

```
error[E0432]: unresolved import `memoffset`
  --> tests/snapshots/path_segment_with_args_in_macro.rs:10:5
   |
10 | use memoffset::offset_of;
   |     ^^^^^^^^^ use of undeclared crate or module `memoffset`

error: cannot determine resolution for the macro `offset_of`
  --> tests/snapshots/path_segment_with_args_in_macro.rs:18:12
   |
18 |     return offset_of!(
   |            ^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying macro imports
```

FWIW the snapshot file I used was

```
$ cat c2rust-transpile/tests/snapshots/path_segment_with_args_in_macro.c
struct s { int p[512]; };
int foo(unsigned char x) {
	return __builtin_offsetof(struct s, p[sizeof(int) + x]);
}
```

I think even with those issues fixed, there would be another problem: unlike the C macro (or, at least, glibc's implementation thereof), `std::mem::offset_of` does not support array indexing, so the snapshot test would not compile. The array indexing is needed to trigger the non-constant path of offsetof handling.

kkysen (commented):
> I looked in to this but it was a big yak shave. `std::mem::offset_of` wasn't stabilized until 1.77, so it's not usable on the the pinned nightly. And even with `self.use_crate(ExternCrate::Memoffset);` it seems that insta tests don't get access to extern crates.

It does seem like an `insta` test is pretty tricky here.  Can you add a non-`insta` under `tests/` somewhere?

> I'm not sure why.

For the `insta` tests, we just transpile them to a single file that we compile with `rustc` instead of transpiling them to a whole crate compiled with `cargo`.  Since resolving dependencies is tricky without `cargo` and we just haven't set the tests up to compile whole crates, it doesn't work.  No fundamental reason why we can't set it up to transpile the whole crate, though; we just haven't done that yet.

> ```rust
> #[no_mangle]
> pub unsafe extern "C" fn foo(mut x: libc::c_uchar) -> libc::c_int {
>     return offset_of!(
>         s, p[(::core::mem::size_of:: < libc::c_int > :: () as libc::c_ulong)
>                                                   // ^^ this guy right here!
>         .wrapping_add(x as libc::c_ulong) as usize]
>     ) as libc::c_ulong as libc::c_int;
> }
> ```

Also, how did you get this to compile without a `use memoffset::offset_of`? I don't see it: https://godbolt.org/z/T8feYMKjb.

c2rust-ast-builder/src/builder.rs
```diff
@@ -315,13 +315,7 @@ impl<S: Make<PathSegment>> Make<Path> for Vec<S> {
     fn make(self, mk: &Builder) -> Path {
         let mut segments = Punctuated::new();
         for s in self {
-            let segment = s.make(mk);
-            let has_params = !segment.arguments.is_empty();
-            segments.push(segment);
-            // separate params from their segment with ::
-            if has_params {
-                segments.push_punct(Default::default());
-            }
```
kkysen, line 324: Maybe this is for a turbofish?  Not too familiar with this code, though.

brk:
> It does seem like an `insta` test is pretty tricky here. Can you add a non-`insta` under `tests/` somewhere?

Done. 

> Also, how did you get this to compile without a `use memoffset::offset_of`? I don't see it: https://godbolt.org/z/T8feYMKjb.

Just to clarify, the test case I showed wasn't originally meant to compile! The goal was to show that the extraneous Colon2 token was being produced, but the only way to show that without modifications to c2rust's code was to arrange for the translated code to construct a macro around the problematic path, and c2rust doesn't generate many macros directly. Quick grepping showed the non-constant `offset_of!` case. Unfortunately the broken-ness of the `offset_of` handling masked the brokenness of the extra Colon2!

c2rust-transpile/tests/tokens.rs
kkysen, line 1: Oh, I hadn't considered testing this with a unit test (I was thinking a test under `tests/`, which runs `c2rust transpile` but transpiles to a whole crate).  But this works very well.

c2rust-transpile/tests/tokens.rs
```diff
@@ -0,0 +1,13 @@
+use syn::__private::ToTokens;
+
+#[test]
+fn test_tokenize() {
+    use c2rust_ast_builder::mk;
```
kkysen: nit: Can you move this to the top-level?

c2rust-transpile/tests/tokens.rs
```diff
@@ -0,0 +1,13 @@
...
+    let tys = vec![mk().path_ty(vec!["t"])];
+    let args = mk().angle_bracketed_args(tys);
+    let pathsegment = mk().path_segment_with_args("x", args);
```
kkysen: nit: `path_segment`

kkysen (approved):
The test looks good, thanks!  Can you just fix a couple of the nits in the test?  I'll merge then.

c2rust-transpile/tests/tokens.rs
brk: The bug only happens with `path_segment_with_args`, not with `path_segment`.

kkysen: Oh I meant the variable name.

kkysen (commented):
If you rebase, CI should be fixed now that #1351 was merged.

kkysen (approved):

PR #1285, opened 2025-07-16
Nix refactor

RossSmyth:
Hello,

This nix devshell was a bit jank. I cleaned it up, and also added a nix derivation.

1. Make a ./nix directory to put Nix-specific stuff in
2. Create a patch for grabbing tinycbor from nixpkgs, since fetching in the Nix sandbox is not allowed

This is fine because the version in nixpkgs is the same version that was being fetched via git anyways.

4. Create a derivation for building c2rust in the nix sandbox at ./nix/package.nix
5. Create a wrapper derivation for running c2rust at ./nix/default.nix

This provides things that c2rust needs in its environment to run, such as setting up PATH and other things.

7. Make a ./shell.nix file

This is basically the same as before. It is hooked up to the flake too.

8. Add some flake checks

This allows to check that it builds when updating and other things.

9. Add overlay in ./default.nix

This is very useful, because it allows for people to directly import c2rust as an overlay into their NixOS config, home-manager config, or devshells.

10. Update Nix instructions in README 

I am unsure what sort of scope the project would like for Nix integration. This is basically the minimum of being able to build it on Nix easily. There are other things that could be done:

1. Run Nix checks in CI
2. Automatically update lock file

There is a TODO for a test failures:
https://github.com/immunant/c2rust/blob/21200930e97f18466c26b5f40a61160593643961/c2rust-analyze/tests/analyze.rs#L47
```log
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `/build/source/target/x86_64-unknown-linux-gnu/release/c2rust-analyze /nix/store/ipjksldzmw00pwndfc8v4wy99q83k8z2-cargo-auditable-0.6.5/bin/cargo-auditable rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 101)
  --- stderr
  error: multiple input filenames provided (first two filenames are `rustc` and `-`)

thread 'main' panicked at '
[ERROR @ c2rust-analyze/src/main.rs:241 @ c2rust_analyze]: error (exit status: 101) running: "/nix/store/k5cb6zl45g4hpwzkjj41qzwq29zy2w1x-cargo-1.65.0-nightly-2022-08-08-x86_64-unknown-linux-gnu/bin/cargo" "check" "--manifest-path" "../analysis/tests/lighttpd-minimal/Cargo.toml"
[...]

PR #1341, opened 2025-08-22
Fix nix build for c2rust.

Chobbes:
This commit adds a number of improvements to the nix build for c2rust the main highlights are:

- A c2rust package is provided, allowing nix users to install and run c2rust itself, instead of just getting a development environment from nix.
- Update LLVM / clang to version 18.
- Test cases pass in the c2rust dev environment now.

Some fixes include...

- Removing oxalica and just using fenix. Mixing rust overlays in nix was causing issues.
- Use the nix tinycbor package to build c2rust. The c2rust cmake files add an external dependency on tinycbor, which causes sandboxed nix builds to fail because they cannot fetch the package. This involves patching the CMakeLists.txt file, and using pkg-config to find the system tinycbor library. Note: in the dev shell cmake will still fetch the external tinycbor.
- Use the C2RUST_TINYCBOR_PATH environment variable, if it exists, to find tinycbor in c2rust-ast-exporter's build.rs
- Use bindgenHook to make sure Rust's bindgen can find header files.
- Update the test_translator.py script to include nix flags in compiler_commands.json Messing with bindgenHook, disable checks so package builds.


c2rust-ast-exporter/build.rs
```diff
@@ -150,6 +150,15 @@ fn build_native(llvm_info: &LLVMInfo) {
     };
 
     // Statically link against 'clangAstExporter' which requires 'tinycbor'
```
kkysen: Could you move this comment to the line right before `cargo:rustc-link-lib=static=clangAstExporter`?  I think that makes a bit more sense.

scripts/test_translator.py
```diff
@@ -244,11 +244,20 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
         # are broken without it on macOS 12.
         # limit this to macOS because if we do happen to have multiple versions of Clang around, we
         # don't know which to use here, and using the wrong can one break things badly
-        if sys.platform == "darwin":
+        if sys.platform == "darwin" or os.environ.get("C2RUST_USE_NIX") == "1":
```
kkysen: I think you could just add this as a normal argument to `argparse`.  It should be more easily discoverable that way.

scripts/test_translator.py
```diff
@@ -244,11 +244,20 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
...
         else:
-            self.clang_resource_dir = ""
+            self.clang_extra_flags = ""
+
+        # The compile_commands.json needs to have extra flags to tell
+        # it where to find libraries when using nix.
+        if os.environ.get("C2RUST_USE_NIX") == "1":
+            nix_compile_flags = os.environ.get("BINDGEN_EXTRA_CLANG_ARGS", "") + os.environ.get("NIX_CFLAGS_COMPILE", "")
+            nix_compile_flags = nix_compile_flags.replace('-isystem ', '-I').replace('-idirafter ', '-I')
+            nix_compile_flags = ['"{}"'.format(f) for f in nix_compile_flags.split() if f.startswith("-I")]
```
kkysen: nit:

```suggestion
            nix_compile_flags = ['"{}"'.format(flag) for flag in nix_compile_flags.split() if flag.startswith("-I")]
```

I think `f` used in close conjunction with `.format` is a bit confusing (I've seen `f` used for the format string arg name before).

scripts/test_translator.py
```diff
@@ -244,11 +244,20 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
...
+            self.clang_extra_flags = " \"-I{}/include\",".format(stdout.strip())
         else:
-            self.clang_resource_dir = ""
+            self.clang_extra_flags = ""
+
+        # The compile_commands.json needs to have extra flags to tell
+        # it where to find libraries when using nix.
+        if os.environ.get("C2RUST_USE_NIX") == "1":
+            nix_compile_flags = os.environ.get("BINDGEN_EXTRA_CLANG_ARGS", "") + os.environ.get("NIX_CFLAGS_COMPILE", "")
+            nix_compile_flags = nix_compile_flags.replace('-isystem ', '-I').replace('-idirafter ', '-I')
```
kkysen: I think we're not the most consistent already, but I think `"` double quotes are more commonly used.  Could you stick to them (except for things like `'"{}"'`)?

c2rust-ast-exporter/build.rs
```diff
@@ -150,6 +150,15 @@ fn build_native(llvm_info: &LLVMInfo) {
...
+    // Use a custom tinybor directory via an environment variable
+    if let Ok(tinycbor_dir) = env::var("C2RUST_TINYCBOR_PATH") {
```
kkysen: From what I've seen (correct me if I'm wrong), `TINYCBOR_DIR` would be more idiomatic for the environment variable here.

scripts/test_translator.py
```diff
@@ -244,11 +244,20 @@ def __init__(self, full_path: str, files: 're.Pattern', keep: List[str], log_lev
...
         # limit this to macOS because if we do happen to have multiple versions of Clang around, we
         # don't know which to use here, and using the wrong can one break things badly
-        if sys.platform == "darwin":
+        if sys.platform == "darwin" or os.environ.get("C2RUST_USE_NIX") == "1":
             _, stdout, _ = clang["-print-resource-dir"].run(retcode=None)
-            self.clang_resource_dir = " \"-I{}/include\",".format(stdout.strip())
+            self.clang_extra_flags = " \"-I{}/include\",".format(stdout.strip())
         else:
-            self.clang_resource_dir = ""
+            self.clang_extra_flags = ""
```
kkysen: We lose the naming of `clang_resource_dir` here, so could you leave it as a `clang_resource_dir` local variable here for this `if`/`else` and then add a separate `self.clang_extra_flags = clang_resource_dir + nix_compile_flags` line after determining each of them?

kkysen (commented):
Thanks for the improvements here!  To reiterate what I've told some other people, we don't have the resources to maintain and review any of the Nix-related code here, but we're happy to accept fixes.  I've reviewed the parts of this that touch the main c2rust files (`build.rs` and `test_translator.py`), but I haven't looked at the nix-specific files.

Also, I'm not sure if you've seen @RossSmyth's #1285, so I just wanted to make sure the two of you are aware of each other's nix changes and can coordinate anything necessary.

c2rust-ast-exporter/build.rs
Chobbes: Of course, no problem!

scripts/test_translator.py
Chobbes: @kkysen I personally think an environment variable is better in this situation because the environment variable will be automatically set by nix so it will be set for the nix build, and it will automatically be set for anybody using the development shell provided by nix. Happy to be persuaded otherwise, though! Do you still think it makes sense to change this?

scripts/test_translator.py
Chobbes: Fair enough, I'll incorporate this change!

scripts/test_translator.py
Chobbes: Yep, no problem. I'll update my changes to prefer double quotes.

c2rust-ast-exporter/build.rs
Chobbes: @kkysen I don't think you're wrong, and I even originally had `TINYCBOR_PATH`. I was just worried that it would be more likely to clash, which probably isn't a real concern. I can change this.

scripts/test_translator.py
Chobbes: Okay, I've made this change (well, I just do `self.clang_extra_flags += nix_compile_flags` in the nix if statement. I'm happy to make `nix_compile_flags` an empty string in a branch or something instead, but I thought it might be more clear that you're not getting weird nix flags unless you're actually using nix)

Chobbes:
@kkysen thanks for the review! I have pushed the fixes for most of the things you brought up, and I left a couple of them unresolved in order to get your opinion. The main one is the use of the environment variable for the test script --- I think this is actually the right thing to do here because nix will automatically set the environment variable for you, which I think is better than having an argument to the test script that you have to know to set on nix systems. The other is a stylistic concern, which I'm happy to defer to your judgement entirely on.

Yes, I had gotten this to work under nix a week or so ago with all of the test cases passing, and just got the time to clean this up on Friday. Only then did I see @RossSmyth's https://github.com/immunant/c2rust/pull/1285, which seems to have had activity from between me getting things working locally and cleaning things up and making my PR. I commented on that PR when I noticed, but haven't heard back yet.

scripts/test_translator.py
kkysen: If this is something automatically set by nix, I think it's fine then.  Could you maybe move this to a `C2RUST_USE_NIX` global set in `main` near the arg parsing, though?  I just want it to be more easily discoverable.  Or something in the `--help` message, too.  Same for the other environment variables.

scripts/test_translator.py
kkysen: This seems good, thanks!

scripts/test_translator.py
Chobbes: Ah, okay, I think that's a good idea. I pushed a change for this, let me know if you want any changes!

Rua:
I tried this out and got this:
```
$ nix run github:Chobbes/c2rust/nix-fixed -- transpile -o . --overwrite-existing compile_commands.json
i_sound.c:27:10: fatal error: 'errno.h' file not found
```

Building `c2rust` succeeds, but while running the transpile it can't find C's header files.

scripts/test_translator.py
```diff
@@ -46,6 +46,9 @@
     'cc_db', 'c_obj', 'c_lib', 'rust_src',
 ]
 
+# Whether or not to use nix for c2rust dependencies This adds paths to
```
kkysen:
```suggestion
# Whether or not to use nix for c2rust dependencies. This adds paths to
```

scripts/test_translator.py
```diff
@@ -616,6 +631,11 @@ def main() -> None:
         choices=intermediate_files + ['all'], default=[],
         help="Which intermediate files to not clear"
     )
+    parser.add_argument(
+        '--use-nix', dest='use_nix', action='store_true',
+        default=os.environ.get('C2RUST_USE_NIX')=='1',
```
kkysen:
```suggestion
        default=os.environ.get('C2RUST_USE_NIX') == '1',
```

Chobbes:
> I tried this out and got this:
> 
> ```
> $ nix run github:Chobbes/c2rust/nix-fixed -- transpile -o . --overwrite-existing compile_commands.json
> i_sound.c:27:10: fatal error: 'errno.h' file not found
> ```
> 
> Building `c2rust` succeeds, but while running the transpile it can't find C's header files.

Yes! This is expected (though maybe not ideal) and the `compile_commands.json` needs to have the paths to search for the libraries. For example:

```
"arguments": [ "cc", "-D_FORTIFY_SOURCE=0", "-I/nix/store/y3lzh1bym1q2h0d09byc6lxsiya3izqm-clang-wrapper-18.1.8/resource-root/include","-I/nix/store/74qjr01q87nwfl0dbsr1s45p8crw3q1f-glibc-2.40-66-dev/include", "-I/nix/store/qnwxpk0in4bm43q2qnykvkjxa9qhqd0z-gcc-14.3.0/include/c++/14.3.0", "-I/nix/store/qnwxpk0in4bm43q2qnykvkjxa9qhqd0z-gcc-14.3.0/include/c++/14.3.0/x86_64-unknown-linux-gnu", "-I/nix/store/pqy858gyg7d8c9p7k587023sardl4wfh-compiler-rt-libc-18.1.8-dev/include", "-I/nix/store/pqy858gyg7d8c9p7k587023sardl4wfh-compiler-rt-libc-18.1.8-dev/include", "-I/nix/store/3yb3a900z5r0x678d8pp46gwab8bbki5-clang-18.1.8-dev/include", "-I/nix/store/3yb3a900z5r0x678d8pp46gwab8bbki5-clang-18.1.8-dev/include", "-I/nix/store/fhdpk4fpgpdkkf8pdl6px5lmnp42z3fz-llvm-18.1.8-dev/include", "-I/nix/store/fhdpk4fpgpdkkf8pdl6px5lmnp42z3fz-llvm-18.1.8-dev/include", "-I/nix/store/z1l05nn4xyaxv25f9pvi7bkmw6jmb48c-ncurses-6.5-dev/include", "-I/nix/store/z1l05nn4xyaxv25f9pvi7bkmw6jmb48c-ncurses-6.5-dev/include", "-I/nix/store/qjh11kgxv245166f27017hwx2fvmwh8p-zlib-1.3.1-dev/include", "-I/nix/store/qjh11kgxv245166f27017hwx2fvmwh8p-zlib-1.3.1-dev/include", "-I/nix/store/7qggxrwirl36a4cgab3q5kzbira2gbqi-tinycbor-0.6.1/include", "-I/nix/store/7qggxrwirl36a4cgab3q5kzbira2gbqi-tinycbor-0.6.1/include", "-I/nix/store/b97gjl5zygaf6vzxsa1lf7jih8q4xi00-openssl-3.5.1-dev/include", "-I/nix/store/b97gjl5zygaf6vzxsa1lf7jih8q4xi00-openssl-3.5.1-dev/include", "-I/nix/store/pqy858gyg7d8c9p7k587023sardl4wfh-compiler-rt-libc-18.1.8-dev/include", "-I/nix/store/pqy858gyg7d8c9p7k587023sardl4wfh-compiler-rt-libc-18.1.8-dev/include", "-I/nix/store/3yb3a900z5r0x678d8pp46gwab8bbki5-clang-18.1.8-dev/include", "-I/nix/store/3yb3a900z5r0x678d8pp46gwab8bbki5-clang-18.1.8-dev/include", "-I/nix/store/fhdpk4fpgpdkkf8pdl6px5lmnp42z3fz-llvm-18.1.8-dev/include", "-I/nix/store/fhdpk4fpgpdkkf8pdl6px5lmnp42z3fz-llvm-18.1.8-dev/include", "-I/nix/store/z1l05nn4xyaxv25f9pvi7bkmw6jmb48c-ncurses-6.5-dev/include", "-I/nix/store/z1l05nn4xyaxv25f9pvi7bkmw6jmb48c-ncurses-6.5-dev/include", "-I/nix/store/qjh11kgxv245166f27017hwx2fvmwh8p-zlib-1.3.1-dev/include", "-I/nix/store/qjh11kgxv245166f27017hwx2fvmwh8p-zlib-1.3.1-dev/include", "-I/nix/store/7qggxrwirl36a4cgab3q5kzbira2gbqi-tinycbor-0.6.1/include", "-I/nix/store/7qggxrwirl36a4cgab3q5kzbira2gbqi-tinycbor-0.6.1/include", "-I/nix/store/b97gjl5zygaf6vzxsa1lf7jih8q4xi00-openssl-3.5.1-dev/include", "-I/nix/store/b97gjl5zygaf6vzxsa1lf7jih8q4xi00-openssl-3.5.1-dev/include", "-c", "variable_arrays.c" ],
```

You can pull these from the `BINDGEN_EXTRA_CLANG_ARGS` and `NIX_CFLAGS_COMPILE` environment variables. This is done for the `compile_commands.json` in the test suite:

```
            nix_compile_flags = os.environ.get("BINDGEN_EXTRA_CLANG_ARGS", "") + os.environ.get("NIX_CFLAGS_COMPILE", "")
            nix_compile_flags = nix_compile_flags.replace("-isystem ", "-I").replace("-idirafter ", "-I")
            nix_compile_flags = ['"{}"'.format(flag) for flag in nix_compile_flags.split() if flag.startswith("-I")]
            nix_compile_flags = ", ".join(nix_compile_flags) + ","
            self.clang_extra_flags += nix_compile_flags
```

I'm not sure if there's a better solution in general, there's some discussions about this problem in nix:

- https://discourse.nixos.org/t/using-nix-infrastructure-to-reliably-generate-compile-commands-json/21961/10

I'm not super familiar with the whole `compile_commands.json` thing, it seems like it makes sense to specify where the libraries are in nix there, but maybe there's a way to provide a wrapper for `c2rust` that would be more convenient.

Rua:
I added the flags in `NIX_CFLAGS_COMPILE` to the JSON (replacing with -I as needed like in the script) and that got it a bit further. But it can't find `stddef.h` now. I guess that file is in another location not listed?

Also, `BINDGEN_EXTRA_CLANG_ARGS` appears to be empty on my system.

Chobbes:
> I added the flags in `NIX_CFLAGS_COMPILE` to the JSON (replacing with -I as needed like in the script) and that got it a bit further. But it can't find `stddef.h` now. I guess that file is in another location not listed?
> 
> 
> 
> Also, `BINDGEN_EXTRA_CLANG_ARGS` appears to be empty on my system.

You might want to also add the -I flags for `clang --print-resource-dir`. You'll need the `BINDGEN_EXTRA_CLANG_ARGS` from the dev shell. You can also set `NIX_DEBUG=1` and compile your file and it should print all of the nix paths and you can use that.

Rua:
Ok, finally figured it out! Turns out I was almost there, I just needed to change this arg
```
-I/nix/store/qnwxpk0in4bm43q2qnykvkjxa9qhqd0z-gcc-14.3.0/lib/gcc/x86_64-unknown-linux-gnu/14.3.0/include-fixed
```
to say `include` at the end instead.

Chobbes:
> Ok, finally figured it out! Turns out I was almost there, I just needed to change this arg
> 
> ```
> 
> -I/nix/store/qnwxpk0in4bm43q2qnykvkjxa9qhqd0z-gcc-14.3.0/lib/gcc/x86_64-unknown-linux-gnu/14.3.0/include-fixed
> 
> ```
> 
> to say `include` at the end instead.

Fantastic! Yeah, I'm not sure, but it might make sense to be able to add some flags for the includes to c2rust directly, and then we can make a wrapper like the nix clang wrapper that automatically includes some of the system libraries. I figured for now it makes sense to just explicitly include all of the stuff because thats sort of what it seems like you're supposed to do, but it's definitely not the most convenient and the default compile_commands.json generated by the c2rust translator tests leaves some things implicit, I think.

scripts/test_translator.py
Chobbes: Updated.

scripts/test_translator.py
Chobbes: Updated this!

Rua:
Is this ready to be merged now? I'm on Nix, and have a few PRs going that are really awkward to work with without these fixes! :D

Rua:
@Chobbes When running some of the tests, I get errors saying it can't find `stdlib.h`. Is that something you can fix?

EDIT: I think it's in this function somewhere, which generates compile commands for the tests. Since your changes added Nix-specific options to the compile commands elsewhere, I figure they need to be added here too?
https://github.com/immunant/c2rust/blob/master/c2rust-transpile/src/lib.rs#L246-L283

kkysen:
> Is this ready to be merged now? I'm on Nix, and have a few PRs going that are really awkward to work with without these fixes! :D

Yes, I think so.

> @Chobbes When running some of the tests, I get errors saying it can't find `stdlib.h`. Is that something you can fix?

Unless something is broken.  If it's all good, I can merge.

Rua:
It seems like it might be a Nix-specific issue relating to the tests. The PR as it stands now is an improvement for sure, and I'd be happy to see it merged now, but hopefully Chobbes will look into this problem afterwards.

Chobbes:
@Rua I can't look today, but I can look on Monday. All of the tests were passing for me with cargo in the dev shell, though. How are you running the tests?

Rua:
```
$ nix develop
$ cargo clean
$ cargo test -p c2rust-transpile --release
   [Compiling stuff...]
    Finished release [optimized] target(s) in 45.39s
     Running unittests src/lib.rs (target/release/deps/c2rust_transpile-109e8649cb12d2ca)

running 5 tests
test renamer::tests::forgets ... ok
test renamer::tests::scoped ... ok
test renamer::tests::simple ... ok
test rust_ast::validate_repr ... ok
test c_ast::tests::test_compare_src_locs_ord ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/snapshots.rs (target/release/deps/snapshots-61d349fc3a38e06a)

running 1 test
/home/rua/code/c2rust/c2rust-transpile/tests/snapshots/arrays.c:1:10: fatal error: 'stdlib.h' file not found
    1 | #include <stdlib.h>
      |          ^~~~~~~~~~
1 error generated.
Error while processing /home/rua/code/c2rust/c2rust-transpile/tests/snapshots/arrays.c.
error: multiple input filenames provided (first two filenames are `+nightly-2022-08-08` and `tests/snapshots/arrays.rs`)

test transpile_all ... FAILED

failures:

---- transpile_all stdout ----
Transpiling arrays.c
thread 'transpile_all' panicked at 'assertion failed: status.unwrap().success()', c2rust-transpile/tests/snapshots.rs:125:5


failures:
    transpile_all

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.19s

error: test failed, to rerun pass '-p c2rust-transpile --test snapshots'
```

Chobbes:
@Rua thanks! I guess I had been running the `./scripts/test_translator.py` script directly after running `cargo build --release`, and I'm not sure how that all plugs into `cargo test`. That's probably where the differences are arising, though.

Chobbes:
@Rua ah, yeah, okay, I didn't know about these cargo test cases. It looks like these rely on the `create_temp_compile_commands` function in `c2rust-transpile/src/lib.rs` file to create the `compile_commands.json` file. I think this needs to be modified in a similar way to the python test script, or maybe there's a better way to wrap things so we don't need to do that. I'm not sure what the best approach is.

Rua:
@Chobbes how are things going with this?

Chobbes:
@Rua to be honest, I was kind of hoping that somebody would chime in with a better solution to dealing with the compile_commands.json thing. I don't think it would be hard to replace it in this test case with a specific compile_commands.json that we could patch in the nix version... But I'm wondering if maybe the c2rust binary should be wrapped to automatically include these flags? I don't use LSPs or anything normally, so I'm actually not very familiar with compile_commands.json and what the best way to handle it within nix is.

Rua:
I have been wondering about that too, and what the "correct" approach is to the problem. For now, I think this PR is a good improvement as it allows me to run c2rust and the non-snapshot tests. A more proper/long-term fix can be done another time.

Chobbes:
Yeah, I'd be happy to discuss this more! I get the sense that it's "normal" to wrap compilers like this in nix (but I'm not certain), but I'm a little fuzzy on the connection between c2rust and the compile_commands.json --- I guess I would have thought that if clang was appropriately wrapped c2rust would find the libraries too, but that seems to not be the case.

Also in a perfect world, maybe c2rust would have flags to specify extra stuff that's missing from the compile_commands.json so it's easier to wrap?

Rua:
@kkysen can this be merged?

Chobbes:
Thanks @kkysen !

Rua:
@Chobbes I'm noticing something odd now that this is merged. In my VSCodium, rust-analyzer no longer reports any clippy warnings, only rustc warnings. But if I run `cargo clippy` myself I get a lot of warnings.

I don't get any logged errors from the rust-analyzer plugin, it behaves as if everything is "normal". I have the rust-analyzer plugin set to use `clippy` as the check command, so it can't be that.

PR #1343, opened 2025-08-23
transpile: refactor `single_attr` and `call_attr` to just use `mk().meta_namevalue()` and `mk().meta_path()` instead of creating structure itself

bungcip:
refactor `single_attr` and `call_attr` so Meta creation is unified in `mk().meta_*` function


kkysen (approved):
Thanks for this simplification!

Issue #1326, opened 2025-08-15
Feature request: disable optimization


mernst:
c2rust does certain optimizations.  For example, it treats the source code `("hello", x+y)` the same as `x+y`.  This generally reduces the size of the Rust program, which is good.  However, in some cases, I want all the code to be output.  Would it be possible to add a command-line option that disables optimizations?  Thanks!

kkysen:
We don't intend to perform any optimizations, but I think this might be caused by skipping some expressions with no side effects.  We'll see if we can fix this.

PR #1344, opened 2025-08-24
Dont write unused unary expr when operator is increment/decrement

bungcip:
currently, c2rust transpiled this code
```c
void increment(){
    int i = 5;
    i++;
    i--;
    --i;
    ++i;
}
```
to this
```rust
#[no_mangle]
pub unsafe extern "C" fn increment() {
    let mut i: std::ffi::c_int = 5 as std::ffi::c_int;
    i += 1;
    i;
    i -= 1;
    i;
    i -= 1;
    i;
    i += 1;
    i;
}
```
which contain useless `i` expression. this is because UnaryExpr can contain side effect expression which cannot be ignored, ~~however not all unary expr can contain side effects. this pull request limit to only Negate , Plus, Complement, and Not~~ . However we can limit it to only expr which is not pure.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -1034,13 +1034,19 @@ impl<'c> Translation<'c> {
             }
         }?;
 
-        // Unused unary operators (`-foo()`) may have side effects, so we need
-        // to add them to stmts.
-        if ctx.is_unused() {
+        // Some unused unary operators (`-foo()`) may have side effects, so we need
+        // to add them to stmts when name is not increment/decrement operator.
+        if ctx.is_unused()
+            && !matches!(
+                name,
+                c_ast::UnOp::PreDecrement
+                    | c_ast::UnOp::PreIncrement
+                    | c_ast::UnOp::PostDecrement
+                    | c_ast::UnOp::PostIncrement
+            )
+        {
             let v = unary.clone().into_value();
-            unary
-                .stmts_mut()
-                .push(Stmt::Expr(*v, Some(Default::default())));
+            unary.add_stmt(mk().semi_stmt(v));
         }
         Ok(unary)
     }
--- a/c2rust-transpile/tests/snapshots/exprs.c
+++ b/c2rust-transpile/tests/snapshots/exprs.c
@@ -1,0 +1,35 @@
+int puts(const char *str);
+
+static int side_effect(){
+    puts("the return of side effect");
+    return 10;
+}
+
+void unary_without_side_effect(){
+    int i = 5;
+    -i;
+    +i;
+    ~i;
+    !i;
+    &i;
+    *&i;
+    i++;
+    i--;
+    --i;
+    ++i;
+}
+
+void unary_with_side_effect(){
+    char* arr[1] = {0};
+
+    -side_effect();
+    +side_effect();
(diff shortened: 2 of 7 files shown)


kkysen (commented):
Oh, actually, this test is OS specific since the printf declarations are slightly different. It's an annoying difference, so you could just try to rewrite it without using printf for the side effect (or just declaring it yourself).

bungcip:
ok, now testcase is not using printf. let see the this PR pass the testsuite.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -1034,9 +1034,9 @@ impl<'c> Translation<'c> {
             }
         }?;
 
-        // Unused unary operators (`-foo()`) may have side effects, so we need
-        // to add them to stmts.
-        if ctx.is_unused() {
+        // Some unused unary operators (`-foo()`) may have side effects, so we need
+        // to add them to stmts when arg expr is not pure
```
```suggestion
        // to add them to stmts when `arg` is not pure.
```

c2rust-transpile/tests/snapshots/exprs.c
```diff
@@ -0,0 +1,33 @@
...
+}
+
+void unary_without_side_effect(){
+    int i = 5;
+    -i;
+    +i;
+    ~i;
+    !i;
+    &i;
+    *&i;
```
kkysen, line 15: So these unary ops now emit nothing, and while that's generally correct (I haven't thought through the exact semantics of the `*&` one yet), it may also delete code that was intentionally written in C, similar to #1326.  For things like `i++`, removing the `i;` and only keeping the `i += 1;` makes perfect sense, but I'm not sure for these other unary operators.

kkysen (commented):
Thanks for the help on this!

c2rust-transpile/tests/snapshots/exprs.c
bungcip: hmm, i thought that c2rust have policy to removing unused things, i saw some other code that skip generating rust ouput when ctx.is_unused() is true, if this is wrong I will the change the code to remove in increment expression only.

bungcip:
hmm, i dont understand why mac CI fails even though i dont touch build configuration

c2rust-transpile/tests/snapshots/exprs.c
bungcip: ok, i force push new code that limit to not write unused code only when operator is increment/decrement

c2rust-transpile/src/translator/operators.rs
bungcip: new code will write unused expression when operator is not increment/decrement. so the comment is outdated now.

kkysen (commented):
> hmm, i dont understand why mac CI fails even though i dont touch build configuration

Sorry, there was a separate CI failure it seems.  If you rebase, CI should be fixed now that #1351 was merged.

c2rust-transpile/tests/snapshots/exprs.c
kkysen, line 1: Also, do you mind splitting this into a separate commit?  That is, one commit that adds this test and its snapshot.  Then the second commit that makes the main change and updates the snapshot.  That way it's clearer what exactly changes.  Thanks!

kkysen (approved):

PR #1346, opened 2025-08-25
transpile: add support for `CTypeKind::UnaryType`s in `--translate-const-macros conservative`

### LINKED ISSUE #1345: `sizeof` isn't translated as `CTypeKind::Size`
Currently, `sizeof`s (i.e. `CTypeKind::UnaryType(_, UnTypeOp::Sizeof, _, _)`) aren't marked as the portable type `CTypeKind::Size`, but as the non-portable `CTypeKind::ULong` (at least on Linux).  Thus, when there's a `sizeof` in a const macro, we get translations like this:

```rust
pub const SIZE: c_ulong = unsafe { ::core::mem::size_of::<[std::ffi::c_int; 10]>() };
```

Which fails to compile since `mem::size_of` returns `usize`.

kkysen:
* Fixes #1345.

This adds support for `CTypeKind::UnaryType` in `--translate-const-macros conservative`.  Previously, these `UnaryType`s received the non-portable type of `c_ulong`, which failed to compile since their Rust translations like `mem::size_of` return `usize`.  On the C side, `sizeof` expressions are indeed supposed to have type `size_t`, so they should get the portable type `CTypeKind::Size`.  This adds this type in `fn bubble_expr_types`, which might not be the right place to do so, since that's currently for fixing up the types from clang 15.  This type fix could instead be moved to a separate pass as well.  In doing so, I also cleaned up some minor stuff in `fn bubble_expr_types`, which should be easier to review if you ignore whitespace changes.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -640,12 +640,10 @@ impl TypedAstContext {
             // Unary ops should be `const`.
             // TODO handle `f128` or use the primitive type.
             Unary(_, _, expr, _) => is_const(expr),
-            UnaryType(_, _, _, _) => false, // TODO disabled for now as tests are broken
             // Not sure what a `None` `CExprId` means here
             // or how to detect a `sizeof` of a VLA, which is non-`const`,
             // although it seems we don't handle `sizeof(VLAs)`
             // correctly in macros elsewhere already.
-            #[allow(unreachable_patterns)]
             UnaryType(_, _, expr, _) => expr.map_or(true, is_const),
             // Not sure what a `OffsetOfKind::Variable` means.
             OffsetOf(_, _) => true,
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -894,71 +894,74 @@ impl TypedAstContext {
             fn children(&mut self, id: SomeId) -> Vec<SomeId> {
                 immediate_children_all_types(self.ast_context, id)
             }
+
             fn post(&mut self, id: SomeId) {
-                if let SomeId::Expr(e) = id {
-                    let new_ty = match self.ast_context.c_exprs[&e].kind {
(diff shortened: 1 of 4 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -888,76 +885,90 @@ impl TypedAstContext {
...
-                    ) {
-                        *ty = new_ty;
-                    };
-                }
+                    }
+                    CExprKind::Unary(_ty, op, e, _idk) => op.expected_result_type(
+                        self.ast_context,
+                        self.ast_context.c_exprs[&e].kind.get_qual_type().unwrap(),
+                    ),
+                    CExprKind::Paren(_ty, e) => self.ast_context.c_exprs[&e].kind.get_qual_type(),
```
ahomescu, line 945: Why did this need to change? The new line is too long, the old block made it more readable.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -888,76 +885,90 @@ impl TypedAstContext {
...
+                        };
+                        let ty = self
+                            .ast_context
+                            .type_for_kind(&kind)
+                            .expect("CTypeKind::Size should be size_t");
+                        Some(CQualTypeId::new(ty))
+                    }
+                    _ => return,
+                };
+                let ty = self
```
ahomescu, line 961: Isn't this the same expression, with a refactoring? Do you need to change this?

ahomescu (approved):

c2rust-transpile/src/c_ast/mod.rs
kkysen: I just decreased the indentation one level, which generally makes things more readable, but this particular line is up to rustfmt.

c2rust-transpile/src/c_ast/mod.rs
kkysen: It was just harder to read since the end of the method chain and if block were far apart.

Issue #22, opened 2018-08-24
Ways to clean up C string literals before writing output?


anp:
This is pretty hard to parse even once you get used to it: `b"contents\x00" as *const u8 as *const c_char`. One way to reduce the noise while remaining correct would be to rely on a crate that lets you declare a C string literal like `cstr-macro`.

rinon:
Going through and triaging old issues. This has been largely taken care of by refactoring passes that clean up the resulting rust code. I'll dig up some details and examples.

ahomescu:
We could revisit this now that C string literals are in the language.

Rua:
@ahomescu @kkysen I wanted to try to implement this, but as it stands, C-string literals are not even available as an unstable feature in the nightly currently used by c2rust. It would need a nightly from June 2023 at the earliest, preferably August.

ahomescu:
I'm not against bumping the nightly if it's only for transpiler output (c2rust-refactor is stuck on 2022-08-08 and it seemed like a lot of work to move). Adding @thedataking for visibility.

PR #1349, opened 2025-08-27
transpile: support `CastKind::ArrayToPointerDecay`s in `--translate-const-macros conservative`

### LINKED ISSUE #1348: array to ptr decays don't emit ptr type casts in const macros
For example, for a `memcpy` of a `char s[N]` array, `.as_ptr()` is emitted for the array to ptr decay, but the ptr cast from `c_char` to `c_void` isn't done.

kkysen:
* Fixes #1348.

This is done by adding a ptr cast to the target pointer type if the element types aren't the same, where the element type is like the pointee type, but also includes array types that will decay to pointers.  `fn CTypeKind::element_ty` is added for that.

I also cleaned up some stuff in `fn convert_cast`, as the naming was pretty irregular, repetitive, and confusing.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile-linux@macros.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile-linux@macros.c.snap
@@ -31,11 +31,13 @@ pub unsafe extern "C" fn size_of_const() -> std::ffi::c_int {
     return SIZE as std::ffi::c_int;
 }
 pub const SIZE: usize = unsafe { ::core::mem::size_of::<[std::ffi::c_int; 10]>() };
+pub const POS: [std::ffi::c_char; 3] =
+    unsafe { *::core::mem::transmute::<&[u8; 3], &[std::ffi::c_char; 3]>(b"\"]\0") };
 #[no_mangle]
 pub unsafe extern "C" fn memcpy_str_literal(mut out: *mut std::ffi::c_char) {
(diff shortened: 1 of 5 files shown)


c2rust-transpile/tests/snapshots/snapshots__transpile-linux@macros.c.snap
```diff
@@ -31,11 +31,13 @@ pub unsafe extern "C" fn size_of_const() -> std::ffi::c_int {
     return SIZE as std::ffi::c_int;
 }
 pub const SIZE: usize = unsafe { ::core::mem::size_of::<[std::ffi::c_int; 10]>() };
+pub const POS: [std::ffi::c_char; 3] =
+    unsafe { *::core::mem::transmute::<&[u8; 3], &[std::ffi::c_char; 3]>(b"\"]\0") };
```
ahomescu, line 35: As a follow-up, we might be able to clean these up by switching from byte literals to C string literals. I found an [old issue](https://github.com/immunant/c2rust/issues/22) on this.

ahomescu (approved):

c2rust-transpile/tests/snapshots/snapshots__transpile-linux@macros.c.snap
kkysen: Yup, definitely something I'd like to do.

PR #1328, opened 2025-08-19
Update c2rust-refactor to nightly-2022-08-08

ahomescu:
Update the refactoring tool to build and run with the same Rust nightly as the transpiler.

PR #1350, opened 2025-08-28
cleanup AST builder functionality

bungcip:
- remove unused function & impl
  - `impl<'a> Make<Unsafety> for &'a str`
  - `impl<'a> Make<Constness> for &'a str`
  - `impl Make<TokenStream> for Vec<String>`
  - `impl Make<PathArguments> for ParenthesizedGenericArguments`
  - `fn parenthesized_args`
- refactor `mk().vis()` usage so we can remove `impl<'a> Make<Visibility> for &'a str`
- refactor `mk().set_mutbl()` usage so we can remove `impl<'a> Make<Mutability> for &'a str`
- refactor `mk().unary_expr()` usage so we can remove `impl<'a> Make<UnOp> for &'a str`
- unify `impl Make<TokenStream> for Vec<&str>`, `impl Make<TokenStream> for Vec<u64>`, and `impl Make<TokenStream> for Vec<Meta>` functionality to use `fn comma_separated`

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -2,6 +2,7 @@
 
 use std::str;
 
+use itertools::intersperse;
 use proc_macro2::{Literal, Punct, Spacing, Span, TokenStream, TokenTree};
 use std::default::Default;
 use std::iter::FromIterator;
@@ -161,21 +162,14 @@ fn punct_box<T, P: Default>(x: Vec<Box<T>>) -> Punctuated<T, P> {
     Punctuated::from_iter(x.into_iter().map(|x| *x))
 }
 
-fn comma_separated<T, F>(items: Vec<T>, f: F) -> TokenStream
+fn comma_separated<I, T>(items: I) -> TokenStream
 where
-    F: Fn(&mut TokenStream, T),
+    I: Iterator<Item = T>,
+    T: ToTokens + Clone,
 {
-    let mut tokens = TokenStream::new();
-    let mut first = true;
-    for item in items {
-        if !first {
-            TokenTree::Punct(Punct::new(',', Spacing::Alone)).to_tokens(&mut tokens);
-        } else {
-            first = false;
-        }
-        f(&mut tokens, item);
-    }
-    tokens
+    let items = items.map(|items| items.to_token_stream());
+    let comma = TokenTree::Punct(Punct::new(',', Spacing::Alone)).into_token_stream();
+    intersperse(items, comma).collect()
 }
 
 pub trait Make<T> {
@@ -286,27 +280,19 @@ impl Make<TokenStream> for Vec<TokenTree> {
 
 impl Make<TokenStream> for Vec<&str> {
     fn make(self, _mk: &Builder) -> TokenStream {
-        comma_separated(self, |tokens, s| {
-            let tt = TokenTree::Ident(Ident::new(s, Span::call_site()));
-            tt.to_tokens(tokens);
-        })
+        comma_separated(self.iter().map(|&s| Ident::new(s, Span::call_site())))
     }
 }
 
 impl Make<TokenStream> for Vec<u64> {
     fn make(self, _mk: &Builder) -> TokenStream {
-        comma_separated(self, |tokens, s| {
-            let tt = TokenTree::Literal(Literal::u64_unsuffixed(s));
-            tt.to_tokens(tokens);
-        })
+        comma_separated(self.iter().map(|&s| Literal::u64_unsuffixed(s)))
     }
 }
 
 impl Make<TokenStream> for Vec<Meta> {
     fn make(self, _mk: &Builder) -> TokenStream {
-        comma_separated(self, |tokens, s| {
-            s.to_tokens(tokens);
-        })
+        comma_separated(self.iter())
     }
 }
 
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -161,6 +161,23 @@ fn punct_box<T, P: Default>(x: Vec<Box<T>>) -> Punctuated<T, P> {
     Punctuated::from_iter(x.into_iter().map(|x| *x))
 }
 
+fn comma_separated<T, F>(items: Vec<T>, f: F) -> TokenStream
+where
+    F: Fn(&mut TokenStream, T),
+{
+    let mut tokens = TokenStream::new();
+    let mut first = true;
+    for item in items {
+        if !first {
+            TokenTree::Punct(Punct::new(',', Spacing::Alone)).to_tokens(&mut tokens);
+        } else {
+            first = false;
+        }
+        f(&mut tokens, item);
+    }
+    tokens
+}
+
 pub trait Make<T> {
     fn make(self, mk: &Builder) -> T;
 }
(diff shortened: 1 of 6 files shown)


c2rust-transpile/src/cfg/structures.rs
```diff
@@ -667,12 +667,13 @@ impl StructureState {
 ///   * Negating something of the form `!<expr>` produces `<expr>`
 ///
 fn not(bool_expr: &Expr) -> Box<Expr> {
+    use syn::UnOp;
```
kkysen, line 670: If this can be hoisted to the top-level, that'd be preferable.

c2rust-ast-builder/src/builder.rs
```diff
@@ -336,82 +290,29 @@ impl Make<TokenStream> for Vec<TokenTree> {
     }
 }
 
-impl Make<TokenStream> for Vec<String> {
```
kkysen, line 333: With an iterator-based `fn comma_separated`, you can simplify this to:

```rust
impl Make<TokenStream> for Vec<&str> {
    fn make(self, _mk: &Builder) -> TokenStream {
        comma_separated(self.iter().map(|&s| Ident::new(s, Span::call_site())))
    }
}

impl Make<TokenStream> for Vec<u64> {
    fn make(self, _mk: &Builder) -> TokenStream {
        comma_separated(self.iter().map(|&s| Literal::u64_unsuffixed(s)))
    }
}

impl Make<TokenStream> for Vec<Meta> {
    fn make(self, _mk: &Builder) -> TokenStream {
        comma_separated(self.iter())
    }
}
```

c2rust-ast-builder/src/builder.rs
```diff
@@ -161,6 +161,23 @@ fn punct_box<T, P: Default>(x: Vec<Box<T>>) -> Punctuated<T, P> {
...
+    for item in items {
+        if !first {
+            TokenTree::Punct(Punct::new(',', Spacing::Alone)).to_tokens(&mut tokens);
+        } else {
+            first = false;
+        }
+        f(&mut tokens, item);
+    }
+    tokens
+}
```
kkysen:
```suggestion
fn comma_separated<I, T>(items: I) -> TokenStream
where
    I: Iterator<Item = T>,
    T: ToTokens + Clone,
{
    let items = items.map(|items| items.to_token_stream());
    let comma = TokenTree::Punct(Punct::new(',', Spacing::Alone)).into_token_stream();
    intersperse(items, comma).collect()
}
```
If you add `itertools` to `c2rust-ast-builder` (which we already use for `c2rust-transpile`), this could be a lot simpler.

c2rust-ast-builder/src/builder.rs
```diff
@@ -252,47 +247,6 @@ impl Make<Extern> for Abi {
...
-        }
-    }
-}
-
-impl<'a> Make<UnOp> for &'a str {
-    fn make(self, _mk: &Builder) -> UnOp {
-        match self {
-            "deref" | "*" => UnOp::Deref(Default::default()),
-            "not" | "!" => UnOp::Not(Default::default()),
-            "neg" | "-" => UnOp::Neg(Default::default()),
```
kkysen, line 290: I'm not sure getting rid of all of these is helpful.  `"!"` is pretty easy to read and understand (although things like `"not"` and the other non-standard spellings for things seem bad).  It has the downside of not being as easily searchable since it's not tied to a symbol like `UnOp::Not`, but it's also easier to read and write in a bunch of ways.

c2rust-transpile/src/cfg/structures.rs
bungcip: maybe we can create a new module, lets call it `transform`, which contain all helper function that transform a rust AST to another rust AST. so we can put helper function like this `not()` and `unparen()` in there. However i prefer doing that in another PR rather than this

c2rust-ast-builder/src/builder.rs
bungcip: if it is ok to add new dependencies crate, i will add it tomorrow.   cannot wait for https://github.com/rust-lang/rust/issues/79524 to become stable😃

c2rust-ast-builder/src/builder.rs
bungcip: I think is more readable for all unary_expr() usage  in transpile to just use `mk().deref_expr()`, `mk().neg_expr()`, and `mk().not_expr()`. I can create follow up PR for the refactor

c2rust-ast-builder/src/builder.rs
kkysen: Yeah, you can use `itertools`.  It shouldn't be an issue since we already use it in `c2rust-transpile`.

c2rust-ast-builder/src/builder.rs
fw-immunant: I believe the original motivation for this was to isolate the actual types of the syntax library from the code using the builder API as much as possible. This was nice when we changed from using the rustc syntax tree to `syn`. But I don't think we anticipate any such switch in the future, and I don't think we desperately need to keep this coupling loose anymore. I'm slightly in favor of the `not_expr()`/etc. refactor you suggest, but I don't think using `UnOp` enum values in core translator code is verboten at this point.

ahomescu:
https://github.com/immunant/c2rust/pull/1328 restores c2rust-refactor which also uses the AST builder. I need to check whether it's using any of the methods this PR is removing.

bungcip:
@ahomescu, correct me if i am wrong, but i think c2rust-refactor dont use c2rust-ast-builder. it seems to use rustc_ast

c2rust-ast-builder/src/builder.rs
bungcip:
```
$ cargo build
  Downloaded pathdiff v0.2.3
  Downloaded cfg-if v1.0.3
  ....
  Downloaded linux-raw-sys v0.9.4
  Downloaded 114 crates (12.7 MB) in 31.19s (largest was `linux-raw-sys` at 2.3 MB)
error: package `clap v4.5.47` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.65.0-nightly
```

just simply adding itertools in c2rust-ast-builder, and running usual `cargo update` + `cargo build` fails when somehow clap suddenly upgraded to `4.5` 🤣. maybe I will just wait until https://github.com/immunant/c2rust/pull/1227 landed

kkysen: We can't use `cargo update`.  But just adding `itertools` should work; I prototyped it and it seemed to work fine.

ahomescu:
> @ahomescu, correct me if i am wrong, but i think c2rust-refactor dont use c2rust-ast-builder. it seems to use rustc_ast

Right, the new PR completely forks the AST builder used by the transpiler (based on syn) from the one used by c2rust-refactor.

c2rust-ast-builder/src/builder.rs
```diff
@@ -160,6 +161,15 @@ fn punct<T, P: Default>(x: Vec<T>) -> Punctuated<T, P> {
 fn punct_box<T, P: Default>(x: Vec<Box<T>>) -> Punctuated<T, P> {
     Punctuated::from_iter(x.into_iter().map(|x| *x))
 }
+fn comma_separated<I, T>(items: I) -> TokenStream
```
kkysen, line 165: nit:

```suggestion

fn comma_separated<I, T>(items: I) -> TokenStream
```

c2rust-ast-builder/src/builder.rs
```diff
@@ -330,97 +277,29 @@ impl Make<TokenStream> for Vec<TokenTree> {
...
-            }
-
-            meta.to_tokens(&mut tokens);
-        }
-
-        tokens
+        comma_separated(self.iter())
     }
 }
-
```
kkysen, line 411: How come this newline is deleted?

kkysen (approved):
LGTM!  Could you just fix up a couple of the newline separators that got messed up?

c2rust-ast-builder/src/builder.rs
bungcip: hmm, maybe some `rebase` shenanigan😅. I will fix it now

bungcip: done

c2rust-ast-builder/src/builder.rs
bungcip: done

Issue #301, opened 2020-09-18
C without undefined behaviour is translated to Rust code with undefined behaviour


jrmuizel:
```c
struct Foo {
    int len;
};

void dec(Foo* f) {
    f->len--;
}

void bar() {
    Foo f = {5};
    struct Foo *fp = &f;
    dec(fp);
    f.len = 6;
    dec(fp);
}
```

translates to

```rust
#[derive(Copy, Clone)]
#[repr(C)]
pub struct Foo {
    pub len: libc::c_int,
}
#[no_mangle]
pub unsafe extern "C" fn dec(mut f: *mut Foo) { (*f).len -= 1; }
#[no_mangle]
pub unsafe extern "C" fn bar() {
    let mut f: Foo = { let mut init = Foo{len: 5 as libc::c_int,}; init };
    let mut fp: *mut Foo = &mut f;
    dec(fp);
    f.len = 6 as libc::c_int;
    dec(fp);
}
```
which when run under miri gives:
```
error: Undefined Behavior: no item granting read access to tag <untagged> found in borrow stack.
  --> src/main.rs:10:49
   |
10 | pub unsafe extern "C" fn dec(mut f: *mut Foo) { (*f).len -= 1; }
   |                                                 ^^^^^^^^^^^^^ no item granting read access to tag <untagged> found in borrow stack.
   |
   = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still experimental
   = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
           
   = note: inside `dec` at src/main.rs:10:49
note: inside `bar` at src/main.rs:17:5
  --> src/main.rs:17:5
   |
17 |     dec(fp);
   |     ^^^^^^^
note: inside `main` at src/main.rs:21:14
  --> src/main.rs:21:14
   |
21 |     unsafe { bar() }
   |              ^^^^^
   = note: inside `<fn() as std::ops::FnOnce<()>>::call_once - shim(fn())` at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
   = note: inside `std::sys_common::backtrace::__rust_begin_short_backtrace::<fn(), ()>` at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys_common/backtrace.rs:137:18
[...]

Issue #1217, opened 2025-04-12
Incorrect Value Read from Struct Initialized with `PTHREAD_RWLOCK_INITIALIZER`


Yeaseen:
This bug arises when using a **compound literal** to initialize a `struct` containing a `pthread_rwlock_t`, and passing a pointer to that struct to a thread. While this is safe and valid in C, the Rust code generated by **C2Rust** mishandles the struct's **memory lifetime** or **layout**, resulting in corrupted field values at runtime.

This issue is **distinct from [#1216]**, which involves `PTHREAD_MUTEX_INITIALIZER` and leads to a runtime crash. In contrast, this bug causes **silent data corruption** without triggering a panic or assertion failure.

## Source C code:

```c
#include <pthread.h>
#include <stdio.h>

typedef struct {
    pthread_rwlock_t lock;
    int val;
} S;

void *reader(void *arg) {
    S *s = arg;
    pthread_rwlock_rdlock(&s->lock);
    printf("val: %d\n", s->val);
    pthread_rwlock_unlock(&s->lock);
    return NULL;
}

int main() {
    pthread_t t;
    S *s = (S[]){ { PTHREAD_RWLOCK_INITIALIZER, 42 } };
    pthread_create(&t, NULL, reader, s);
    pthread_join(t, NULL);
    return 0;
}
```

## C Output:

```
val: 42
```

<details>
<summary>Click to view translated Rust code</summary>
```rust
#![allow(
    dead_code,
    mutable_transmutes,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
use ::transpiled_code::*;
extern "C" {
    fn pthread_create(
        __newthread: *mut pthread_t,
        __attr: *const pthread_attr_t,
        __start_routine: Option::<
            unsafe extern "C" fn(*mut libc::c_void) -> *mut libc::c_void,
        >,
        __arg: *mut libc::c_void,
    ) -> libc::c_int;
    fn pthread_join(
        __th: pthread_t,
        __thread_return: *mut *mut libc::c_void,
    ) -> libc::c_int;
    fn pthread_rwlock_rdlock(__rwlock: *mut pthread_rwlock_t) -> libc::c_int;
    fn pthread_rwlock_unlock(__rwlock: *mut pthread_rwlock_t) -> libc::c_int;
    fn printf(_: *const libc::c_char, _: ...) -> libc::c_int;
}
#[derive(Copy, Clone)]
[...]

PR #1355, opened 2025-09-02
transpile: Use raw borrows instead of normal borrows

### LINKED ISSUE #743: Static string is not borrowed correctly when used multiple times
_(note, I probably don't know c2rust enough to correctly describe the problem)_

When transpiling the c library [h3](https://github.com/uber/h3), there exists a static string (in a macro?):

https://github.com/uber/h3/blob/139e5bff48ec16e5af745136d514a5b104869427/src/apps/testapps/testH3CellAreaExhaustive.c#L53

Which is transpiled like so:

```rust
let mut pos: [libc::c_char; 49] = *::core::mem::transmute::<&[u8; 49], &mut [libc::c_char; 49]>(
        b"distance between cell centers should be positive\0",
    );
```

I don't believe it should be mutable as the string is never changed. The variable is later referenced multiple times like this:

```rust
pos.as_mut_ptr()
```

which throws a borrow checker error. I was able to fix the issue by just inlining the static string at the `pos.as_mut_ptr()` callsite.

Code in https://github.com/LegNeato/unsafe-h3lib/commit/386d0f40ecb96216fbb89e090017318873ccb215#r89264842 (note, takes a bit to load)

### LINKED ISSUE #1224: Addressing `static mut` Function Pointer Translation in C2Rust for Safe Rust Semantics
## Original C Code

```c
#include <stdio.h>
int funA() { return 10; }

int (*func_pointer[1])() = {funA};

int func1() {
    return (*func_pointer[0])();
}

int main() {
    int result = func1();
    printf("Result: %d\n", result);
    return 0;
}
```

## Translated Rust Code by C2Rust (with warning)

```rust
#[no_mangle]
pub unsafe extern "C" fn funA() -> libc::c_int {
    return 10 as libc::c_int;
}

#[no_mangle]
pub static mut func_pointer: [Option::<unsafe extern "C" fn() -> libc::c_int>; 1] = unsafe {
    [
        Some(
            ::core::mem::transmute::<
                unsafe extern "C" fn() -> libc::c_int,
                unsafe extern "C" fn() -> libc::c_int,
            >(funA),
        ),
    ]
};

#[no_mangle]
pub unsafe extern "C" fn func1() -> libc::c_int {
    return ::core::mem::transmute::<
        _,
        fn() -> libc::c_int,
    >(
        (Some(
            (*func_pointer.as_mut_ptr().offset(0 as libc::c_int as isize))
                .expect("non-null function pointer"),
        ))
            .expect("non-null function pointer"),
    )();
}

    return 0 as libc::c_int;
}
```

### Warning on `cargo build`

```text
warning: creating a mutable reference to mutable static is discouraged
  --> src/runner.rs:33:15
   |
33 |             (*func_pointer.as_mut_ptr().offset(0 as libc::c_int as isize))
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static

    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
   = note: mutable references to mutable statics are dangerous; it's undefined behavior if any other pointer to the static is used or if any other reference is created for the static while the mutable reference lives
   = note: `#[warn(static_mut_refs)]` on by default
```

## Modified Rust Code (No Warning, Same Behavior)

```rust
use std::cell::UnsafeCell;

#[no_mangle]
pub static func_pointer: FuncPointerWrapper = FuncPointerWrapper(UnsafeCell::new([Some(funA)]));

#[repr(transparent)]
pub struct FuncPointerWrapper(pub UnsafeCell<[Option<unsafe extern "C" fn() -> libc::c_int>; 1]>);

unsafe impl Sync for FuncPointerWrapper {}

#[no_mangle]
pub unsafe extern "C" fn funA() -> libc::c_int {
    10
}

#[no_mangle]
pub unsafe extern "C" fn func1() -> libc::c_int {
    let ptr_array = &*func_pointer.0.get();
    ptr_array[0].expect("non-null function pointer")()
}
```

## Issue I tried to overcome:

- C2Rust generates `static mut` and accesses it using a **mutable reference**, triggering the `static_mut_refs` warning.
- This is undefined behavior if any other reference exists while the mutable one is active.

So, I:
- Replaced `static mut` with a safe `static` using `UnsafeCell`.
- Wrapped the `UnsafeCell` in a transparent wrapper struct (`FuncPointerWrapper`).
- Used `unsafe impl Sync` to allow `static` usage, acknowledging and controlling the unsafe access explicitly.

Feel free to close this if this is not important. @kkysen

### LINKED ISSUE #1239: C2Rust Emits Unsafe Reference to Packed Field Causing Rust Error E0793
C2Rust incorrectly emits a reference to a field inside a `#pragma pack(1) struct`, which results in an unsafe and invalid reference in Rust. This triggers Rust compiler error `E0793`, as references to potentially unaligned fields are undefined behavior in Rust.

## C code:

```c
#include <stdlib.h>
#pragma pack(push, 1)
struct S {
    int x;
};
#pragma pack(pop)

int func() {
    struct S* s = malloc(sizeof(struct S));
    return &s->x == &s->x;
}

int main(){
    return 0;
}
```

## Translated Rust Code:

```rust
#[derive(Copy, Clone)]
#[repr(C, packed)]
pub struct S {
    pub x: libc::c_int,
}
#[no_mangle]
pub unsafe extern "C" fn func() -> libc::c_int {
    let mut s: *mut S = malloc(::core::mem::size_of::<S>() as libc::c_ulong) as *mut S;
    return (&mut (*s).x as *mut libc::c_int == &mut (*s).x as *mut libc::c_int)
        as libc::c_int;
}
```

## Rust build error:

```text
error[E0793]: reference to packed field is unaligned
  --> src/runner.rs:22:13
   |
22 |     return (&mut (*s).x as *mut libc::c_int == &mut (*s).x as *mut libc::c_int)
   |             ^^^^^^^^^^^
   |
   = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
   = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
   = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)

error[E0793]: reference to packed field is unaligned
  --> src/runner.rs:22:48
   |
22 |     return (&mut (*s).x as *mut libc::c_int == &mut (*s).x as *mut libc::c_int)
   |                                                ^^^^^^^^^^^
   |
   = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
   = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
   = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)

For more information about this error, try `rustc --explain E0793`.
```

### LINKED ISSUE #1352: Potential unsoundness caused by emitting borrows
In some cases, c2rust emits borrows when translating C code. The borrows may be explicit (with `&`) or implicit by calling a method on the value. Any kind of borrowing has the potential for unsoundness, because borrows can invalidate pointers that were created before the borrow. Using an invalidated pointer is UB.

One particular case of this I noticed is the use of `as_ptr` and `as_mut_ptr` for array accesses. Since those borrow the whole array, they invalidate any existing pointers into the array.

I don't know if this is a known issue, but I didn't find any issues or PRs about this exact problem. I did find #301 which is related, but doesn't go into the general issue of borrowing.

Rua:
* Depends on #1446.
* Fixes #1352.
* Fixes #1239.
* Fixes #1224.
* Fixes #743.

I noticed that there was some logic to determine whether reference-to-pointer decay happens automatically, or needs a cast. I don't know if that's still needed at all, I can strip it out if not. I never dealt with inline assembly before, so I wasn't sure how to handle references vs pointers in that context. I left it as it is for now.

### DIFF
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -142,7 +142,17 @@ impl<'c> Translation<'c> {
                     Some(mk().array_ty(mk().ident_ty("u8"), mk().lit_expr(len as u128)));
             }
             needs_cast = true;
-        } else {
+        }
+        // Values that translate into temporaries can't be raw-borrowed in Rust,
+        // and must be regular-borrowed first.
+        // Borrowing in a static/const context will extend the lifetime to static.
+        else if arg_is_macro
+            || ctx.is_static
+                && matches!(
+                    arg_expr_kind,
+                    Some(CExprKind::Literal(..) | CExprKind::CompoundLiteral(..))
+                )
+        {
             let arg_cty_kind = &self.ast_context.resolve_type(arg_cty.ctype).kind;
 
             if is_array_decay {
@@ -167,6 +177,15 @@ impl<'c> Translation<'c> {
                 if ctx.decay_ref.is_yes() || needs_cast {
                     ref_cast_pointee_ty = Some(self.convert_pointee_type(arg_cty.ctype)?);
                 }
+            }
+        } else {
+            self.use_feature("raw_ref_op");
+            val = val.map(|val| mk().set_mutbl(mutbl).raw_borrow_expr(val));
+
+            if is_array_decay {
+                // TODO: Call `ptr::as_[mut]_ptr` instead once that is available.
+                // (`array_ptr_get` feature added to nightly in January 2024)
+                needs_cast = true;
             }
         }
 
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile-aarch64@vm_x86.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile-aarch64@vm_x86.c.snap
@@ -11,6 +11,7 @@ input_file: c2rust-transpile/tests/snapshots/arch-specific/vm_x86.c
     unused_assignments,
     unused_mut
 )]
+#![feature(raw_ref_op)]
 #[derive(Copy, Clone)]
 #[repr(C)]
 pub struct vm_t {
@@ -52,8 +53,8 @@ pub unsafe extern "C" fn VM_CallCompiled(
     *(image.offset(programStack as isize) as *mut byte as *mut ::core::ffi::c_int) =
         -(1 as ::core::ffi::c_int);
     entryPoint = (*vm).codeBase.offset((*vm).entryOfs as isize);
-    opStack =
-        (stack.as_mut_ptr() as *mut ::core::ffi::c_int).offset(16 as ::core::ffi::c_int as isize);
+    opStack = (&raw mut stack as *mut byte as *mut ::core::ffi::c_int)
+        .offset(16 as ::core::ffi::c_int as isize);
     *opStack = 0 as ::core::ffi::c_int;
     opStackOfs = 0 as ::core::ffi::c_int;
     if opStackOfs != 1 as ::core::ffi::c_int
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile-x86_64-linux@varargs.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile-x86_64-linux@varargs.c.snap
@@ -11,7 +11,7 @@ input_file: c2rust-transpile/tests/snapshots/arch-os-specific/varargs.c
     unused_assignments,
     unused_mut
 )]
-#![feature(c_variadic)]
+#![feature(c_variadic, raw_ref_op)]
 extern "C" {
     fn printf(__format: *const ::core::ffi::c_char, ...) -> ::core::ffi::c_int;
     fn vprintf(
@@ -129,8 +129,8 @@ pub unsafe extern "C" fn valist_struct_pointer_member(
     let mut b: vastruct = vastruct {
         args: ::core::mem::MaybeUninit::uninit().assume_init(),
     };
-    let mut p: *mut vastruct = &mut a;
-    let mut q: *mut vastruct = &mut b;
+    let mut p: *mut vastruct = &raw mut a;
+    let mut q: *mut vastruct = &raw mut b;
     (*p).args = args.clone();
     (*q).args = (*p).args.clone();
     vprintf(fmt, (*p).args.as_va_list());
@@ -156,7 +156,7 @@ pub unsafe extern "C" fn borrowed_valist(mut count: size_t, mut args: ...) {
     let mut ap: ::core::ffi::VaListImpl;
     ap = args.clone();
     while count > 0 as size_t {
-        print_int(&mut ap);
+        print_int(&raw mut ap);
         count = count.wrapping_sub(1);
     }
 }
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile-x86_64@vm_x86.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile-x86_64@vm_x86.c.snap
@@ -11,7 +11,7 @@ input_file: c2rust-transpile/tests/snapshots/arch-specific/vm_x86.c
     unused_assignments,
     unused_mut
 )]
-#![feature(asm)]
+#![feature(asm, raw_ref_op)]
 use ::core::arch::asm;
 #[derive(Copy, Clone)]
 #[repr(C)]
@@ -54,8 +54,8 @@ pub unsafe extern "C" fn VM_CallCompiled(
     *(image.offset(programStack as isize) as *mut byte as *mut ::core::ffi::c_int) =
         -(1 as ::core::ffi::c_int);
     entryPoint = (*vm).codeBase.offset((*vm).entryOfs as isize);
-    opStack =
-        (stack.as_mut_ptr() as *mut ::core::ffi::c_int).offset(16 as ::core::ffi::c_int as isize);
+    opStack = (&raw mut stack as *mut byte as *mut ::core::ffi::c_int)
+        .offset(16 as ::core::ffi::c_int as isize);
     *opStack = 0 as ::core::ffi::c_int;
     opStackOfs = 0 as ::core::ffi::c_int;
     asm!(
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@arrays.c.snap
@@ -11,6 +11,7 @@ input_file: c2rust-transpile/tests/snapshots/arrays.c
     unused_assignments,
     unused_mut
 )]
+#![feature(raw_ref_op)]
 #[derive(Copy, Clone)]
 #[repr(C)]
 pub struct C2RustUnnamed {
@@ -42,7 +43,7 @@ pub static mut static_char_ptr: *mut ::core::ffi::c_char =
     b"mystring\0" as *const u8 as *const ::core::ffi::c_char as *mut ::core::ffi::c_char;
 #[no_mangle]
 pub static mut static_void_ptr: *mut ::core::ffi::c_void =
-    unsafe { static_char_array.as_ptr() as *mut ::core::ffi::c_char as *mut ::core::ffi::c_void };
+    unsafe { &raw const static_char_array as *mut ::core::ffi::c_char as *mut ::core::ffi::c_void };
 #[no_mangle]
 pub unsafe extern "C" fn entry() {
     let mut int_2d: [[::core::ffi::c_int; 1]; 1] = [[1 as ::core::ffi::c_int]];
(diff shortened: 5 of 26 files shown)


Rua:
Hmm, the CI failed on `&raw mut` not being allowed in statics. I thought I had tested that...

Rua:
I asked around on The #dark-arts channel of the Rust Discord, and they assured me that `&raw mut` is allowed in statics. It seems that in *current* Rust, `&raw mut` is allowed in statics, but in the nightly being used for c2rust, it was not yet allowed?

bungcip:
c2rust-transpile still use old nightly ( i think `2022-08-08`), and in those version, `&raw mut` still unstable. I expect there is still bug in those version

Rua:
In that nightly version, the `addr_of` and `addr_of_mut` macros are already stabilised. And internally, those macros just call `&raw const` and `&raw mut`. So any bugs would have affected stable code.

Rua:
Sadly I am not able to make this work right now. I'm running into niche cases where a temporary is borrowed and its lifetime is extended. Rust does not extend the lifetime when using the raw borrow operator, only the normal borrow operator. This is making a few of the tests still fail.

If there is a way to know exactly in which cases the borrow operator references a temporary, then that could potentially solve the issue.

Rua:
I think I found a way to deal with the issue, at least in the case of compound literals. That makes this pretty much ready I think, all the tests run successfully.

Two questions still remain:
- `DecayRef` doesn't seem needed anymore. Should I remove it?
- At https://github.com/immunant/c2rust/blob/master/c2rust-transpile/src/translator/named_references.rs#L127-L132 there is a bit of code that takes a `ref mut`. Changing that to a pointer may still be good, but if I do so, the `binary_conditionals` test stops compiling, so I've left it.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2979,12 +2979,8 @@ impl<'c> Translation<'c> {
                 }
             }
             _ => {
-                let addr_lhs = mk().set_mutbl(mutbl).addr_of_expr(lhs);
-
-                let lhs_type = self.convert_type(lhs_type.ctype)?;
-                let ty = mk().set_mutbl(mutbl).ptr_ty(lhs_type);
-
-                mk().cast_expr(addr_lhs, ty)
+                self.use_feature("raw_ref_op");
```
kkysen: Are we able to use `addr_of!` and `addr_of_mut!` instead of `&raw const` and `&raw mut` for now?  As far as I'm aware, their semantics are the same, and `&raw` just stabilized the native syntax.  Once we merge #1227, we can switch to `&raw`, or maybe add a target version (so that things like `c2rust-refactor` can keep consuming the old output).

tests/conditionals/src/test_conditionals.rs
```diff
@@ -83,11 +85,11 @@ pub fn test_unused_conditional() {
...
+pub fn test_else_if_chain() {
     unsafe {
-        assert_eq!(entry4(0) , rust_entry4(0));
-        assert_eq!(entry4(10) , rust_entry4(10));
-        assert_eq!(entry4(20) , rust_entry4(20));
-        assert_eq!(entry4(30) , rust_entry4(30));
+        assert_eq!(entry4(0), rust_entry4(0));
+        assert_eq!(entry4(10), rust_entry4(10));
+        assert_eq!(entry4(20), rust_entry4(20));
+        assert_eq!(entry4(30), rust_entry4(30));
```
kkysen: Can you put the formatting changes in another PR?  I can merge that right away.

c2rust-transpile/src/translator/mod.rs
kkysen: On the other hand, the `&raw` syntax is a lot more concise, which is nice, and we could also just remove `#![feature(raw_ref_op)]` when switching to newer versions.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4512,25 +4508,23 @@ impl<'c> Translation<'c> {
...
-                                }));
-                            }
-
-                            Ok(call)
+                            let target_ty = self.convert_type(ty.ctype)?;
+
+                            Ok(val.map(|x| {
+                                self.use_feature("raw_ref_op");
+                                let borrow = mk().set_mutbl(mutbl).raw_borrow_expr(x);
+                                // TODO: Change to call `ptr::as_[mut]_ptr` once that is available.
```
kkysen: Can you add when this is available?  Are we waiting for it to be `const`?

c2rust-transpile/src/translator/operators.rs
```diff
@@ -941,37 +943,33 @@ impl<'c> Translation<'c> {
...
-                                return Ok(addr_of_arg);
-                            }
-                        }
+                    self.use_feature("raw_ref_op");
+                    if ctx.is_static && matches!(mutbl, Mutability::Mutable) {
+                        // TODO: The currently used nightly doesn't allow `&raw mut` in static
+                        // initialisers, but the latest version does.
+                        // So we take a `&raw const` and then cast.
+                        // Remove this exemption when the version is updated.
+                        val = mk().cast_expr(mk().raw_borrow_expr(val), ty);
```
kkysen: Can we combine this logic with the other use of it in its own function?  Especially since it's something we know is going to have to be updated in the future.

c2rust-transpile/src/translator/mod.rs
Rua: > On the other hand, the `&raw` syntax is a lot more concise, which is nice, and we could also just remove `#![feature(raw_ref_op)]` when switching to newer versions.

Yes, that was my intention. The users will probably just want to use the newest (stable) Rust version, and this way they don't have to convert everything themselves.

c2rust-transpile/src/translator/mod.rs
Rua: It was added in January 2024 and is still unstable right now. So for the nightly used in c2rust, it just doesn't exist at all yet.

c2rust-transpile/src/translator/operators.rs
Rua: Probably not easily, as the surrounding code is rather different.

Rua:
Sorry for all the code churn. The commits kinda got messed up by rebasing and I decided to just redo everything from the start.

Currently, the CI is failing because it's trying to raw-borrow what is being translated to a `const` in Rust. In C, this is a macro that expands to a literal string. I think because it's a macro, the special exception for string literals here is no longer triggering, and it's instead being processed as a regular array-to-pointer decay. But because the macro expansion is being stored as a  `const`, this doesn't work.
https://github.com/immunant/c2rust/blob/master/c2rust-transpile/src/translator/mod.rs#L4503

Is there a way to test whether an array-to-pointer decay expression is a macro expansion that was translated into a Rust `const` like this? Then I can add a special case to handle that situation.

fw-immunant:
> Sorry for all the code churn. The commits kinda got messed up by rebasing and I decided to just redo everything from the start.
> 
> Currently, the CI is failing because it's trying to raw-borrow what is being translated to a `const` in Rust. In C, this is a macro that expands to a literal string. I think because it's a macro, the special exception for string literals here is no longer triggering, and it's instead being processed as a regular array-to-pointer decay. But because the macro expansion is being stored as a `const`, this doesn't work. https://github.com/immunant/c2rust/blob/master/c2rust-transpile/src/translator/mod.rs#L4503
> 
> Is there a way to test whether an array-to-pointer decay expression is a macro expansion that was translated into a Rust `const` like this? Then I can add a special case to handle that situation.

`ctx.is_const` on the `AstContext` indicates whether translation is currently in a `const` context. Translation can branch on that to either choose a different (const-compatible) idiom for expressing the C computation or to fail (return `Err(_)`) and the caller will attempt to translate the code without creating a `const`.

Rua:
I'm marking this as a draft for now, because there's still some issues to be figured out regarding temporaries being borrowed. Probably depends on a solution for #1217 and #1379.

Rua:
> `ctx.is_const` on the `AstContext` indicates whether translation is currently in a `const` context. Translation can branch on that to either choose a different (const-compatible) idiom for expressing the C computation or to fail (return `Err(_)`) and the caller will attempt to translate the code without creating a `const`.

I added a check for `ctx.is_const`, but it seems that the value is `false`, so the CI is generating raw-borrow errors instead of using `as_ptr`.

EDIT: I figured out I needed `translate_as_macro` instead. Now the tests are almost passing.

Rua:
I haven't abandoned this PR, but as I worked on it I kept finding more things that needed fixing first before this can work:
- #1481
- #1446
- #1443
- #1434
- #1407

kkysen (commented):
> I haven't abandoned this PR, but as I worked on it I kept finding more things that needed fixing first before this can work:

Thanks for the continued work!  This is definitely a big issue, and fixing it would make our output significantly more sound, so it's very much appreciated!

Rua:
At long last this is finally passing all the tests! I think it's ready to be reviewed now.

kkysen (commented):
I'm rebasing this after I merged #1494 since they overlap a bunch, and it's easier to see what's happening here post #1494.

c2rust-transpile/src/translator/pointers.rs
```diff
@@ -160,14 +170,23 @@ impl<'c> Translation<'c> {
...
                     ref_cast_pointee_ty = Some(self.convert_pointee_type(arg_cty.ctype)?);
                 }
             }
+        } else {
+            self.use_feature("raw_ref_op");
+            val = val.map(|val| mk().set_mutbl(mutbl).raw_borrow_expr(val));
+
+            if is_array_decay {
+                // TODO: Call `ptr::as_[mut]_ptr` instead once that is available.
+                // (`array_ptr_get` feature added to nightly in January 2024)
```
kkysen:
```suggestion
                // TODO: Call `pointer::as_[mut]_ptr` instead once that is available.
                // (`#![feature(array_ptr_get)]` added in nightly-2024-03-17.)
```

I think `pointer` is more often used for inherent ptr methods, as `std::ptr` is a separate module.

https://github.com/rust-lang/rust/pull/119411/ was merged on 2024-03-17, so I think that's the correct date.  Also, I prefer using `#![feature(...)]` and `nightly-yyyy-mm-dd` so that it's more easily greppable.

kkysen:
Oh, I also think this fixes #301.  You or I should also add a test for that.

kkysen:
I added the test from #301, but it still triggers UB in `miri` ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9b399f2b4436b9954a1c8c73aff4e8d8)) even with the `&raw mut` instead of `&mut`.

workingjubilee:
Per Ralf:
> Specifically, when you have a local variable f and write `f.len = ...`, then (like if you had used a mutable reference) this asserts to the compiler that f is the unique way to access this memory, so any outstanding other pointers (safe references or unsafe raw pointers) now become invalid. This is crucial to enable the compiler to perform optimizations based on these aliasing rules.

pointers have to be recreated after that field access.

kkysen:
> Per Ralf:
> 
> > Specifically, when you have a local variable f and write `f.len = ...`, then (like if you had used a mutable reference) this asserts to the compiler that f is the unique way to access this memory, so any outstanding other pointers (safe references or unsafe raw pointers) now become invalid. This is crucial to enable the compiler to perform optimizations based on these aliasing rules.
> 
> pointers have to be recreated after that field access.

What about if it was `(*f).len =` instead (where `f` is a `*mut _`)?  Is there no way to have multiple `*mut`s that stay valid as long as there is no concurrent access?

PR #1359, opened 2025-09-04
Store `alloca` vectors in top function stope

### LINKED ISSUE #759: Incorrect translation of alloca in block scope
Tested with c2rust 0.16.0

```c
#include <stdio.h>
#include <string.h>

int one = 1;

int main(void) {
  char *p;
  if (one) {
    p = __builtin_alloca(100);
  }
  strcpy(p, "Hello, world!");
  puts(p);
}
```
This is a program that has no memory errors. The `if (one)` is guaranteed to be entered, though the compiler cannot see that, and the result of `__builtin_alloca` persists until the function ends, even after the block has ended.

It is translated to the following Rust code:
```rust
#![register_tool(c2rust)]
#![feature(register_tool)]
use ::c2rust_out::*;
extern "C" {
    fn puts(__s: *const libc::c_char) -> libc::c_int;
    fn strcpy(_: *mut libc::c_char, _: *const libc::c_char) -> *mut libc::c_char;
}
#[no_mangle]
pub static mut one: libc::c_int = 1 as libc::c_int;
    let mut p: *mut libc::c_char = 0 as *mut libc::c_char;
    if one != 0 {
        let mut fresh0 = ::std::vec::from_elem(
            0,
            100 as libc::c_int as libc::c_uint as usize,
        );
        p = fresh0.as_mut_ptr() as *mut libc::c_char;
    }
    strcpy(p, b"Hello, world!\0" as *const u8 as *const libc::c_char);
    puts(p);
    return 0;
}
pub fn main() {
}
```
Note here the `let mut fresh0 = ::std::vec::from_elem(0, 100 as libc::c_int as libc::c_uint as usize);` in block scope. Here, unlike in the C program, the vec will be dropped when the block is exited, and `p` is left a dangling pointer.

A possible corrected translation would be to make `fresh0` a function-scope variable.

Rua:
Fixes #759. A long-standing issue, where memory for `alloca` would be freed too early due to being declared in the wrong scope. This PR changes it so that all allocations are stored in a single `Vec<Vec<u8>>` declared at the top of the function. A new `Vec<u8>` of the appropriate size is pushed to this for every `alloca` call. This way, the allocations will not be dropped until the end of the function, just like in C.


fw-immunant (approved):
The translation strategy seems sane and the changes look good. It would be great to have a test here, though; the one from #759 would probably suffice.

Rua:
I had a look at the existing tests. Before this PR, the translated output of code using `alloca` would compile; it just had UB due to a dangling pointer. I don't immediately see a way of testing that code has no dangling pointers.

fw-immunant:
If running the code crashes, that would show a regression; otherwise, we could also add a snapshot test (see `c2rust-transpile/tests/snapshots/`).

tests/builtins/src/alloca.c
```diff
@@ -0,0 +1,14 @@
...
+
+int alloca_hello(void) {
+    char *p;
+    if (one) {
+        p = __builtin_alloca(100);
+    }
+    strcpy(p, "Hello, world!");
+    puts(p);
+    return 0;
+}
```
kkysen: nit:

```suggestion
}

```

c2rust-transpile/tests/snapshots/alloca.c
```diff
@@ -0,0 +1,20 @@
...
+        *alloca1 = val1;
+    }
+
+    if (TRUE) {
+        alloca2 = alloca(sizeof(int));
+        *alloca2 = val2;
+    }
+
+    return *alloca1 + *alloca2;
+}
```
kkysen:
```suggestion
}

```

kkysen (commented):
One little nit, can you end files with a newline?  I can merge after that.

Rua:
Does anything else need to be done here?

kkysen (changes requested):
Could you remove the merge commits and rebase instead?

Rua:
Ok, done.

PR #1360, opened 2025-09-04
Avoid mutable transmute in string literals

### LINKED ISSUE #1154: Stop emitting code that triggers the mutable_transmutes lint
Transmuting from a reference to a mutable reference is insta-UB: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8da5074ba78a22ba108e198273c547ba

```
error: Undefined Behavior: writing to alloc1 which is read-only
 --> src/main.rs:5:17
  |
5 |         let _ = core::mem::transmute::<&[u8; 10], &mut [libc::c_char; 10]>(b"\0\0\0\0\0\0\0\0\0\0");
  |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ writing to alloc1 which is read-only
  |
  = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
  = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
  = note: BACKTRACE:
  = note: inside `main` at src/main.rs:5:17: 5:100
```

C2rust currently does such transmutes for code like `char mode2[10] = "";` where it produces `let mut mode2: [libc::c_char; 10] = *core::mem::transmute::<&[u8; 10], &mut [libc::c_char; 10]>(b"\0\0\0\0\0\0\0\0\0\0");` even though `let mut mode2: [libc::c_char; 10] = [0; 10];` works just fine.

Rua:
Fixes #1154, going with the fix proposed in https://github.com/immunant/c2rust/issues/743#issuecomment-1358127467.

The original code would set the array size to 1 if it wasn't statically determined by the type. I'm not sure in which situations that would happen, it still seems suspicious. But to avoid breaking anything I don't understand, I separated it off, and applied my fix only to the case where the size is known.


c2rust-transpile/src/translator/mod.rs
```diff
@@ -1237,7 +1237,6 @@ impl<'c> Translation<'c> {
                 "non_camel_case_types",
                 "non_snake_case",
                 "dead_code",
-                "mutable_transmutes",
```
kkysen, line 1236: The unknown length/1 case is still doing a mutable transmute, right?

I think we should just drop that case.  I can't tell what it's for either.

c2rust-transpile/src/translator/literals.rs
```diff
@@ -153,30 +153,23 @@ impl<'c> Translation<'c> {
...
             CLiteral::String(ref val, width) => {
                 let mut val = val.to_owned();
-
                 let num_elems = match self.ast_context.resolve_type(ty.ctype).kind {
-                    // Match the literal size to the expected size padding with zeros as needed
-                    CTypeKind::ConstantArray(_elem_ty, size) => size,
-                    // zero terminator
-                    _ => 1,
+                    CTypeKind::ConstantArray(_elem_ty, num_elems) => num_elems,
+                    _ => panic!("String literal with unknown size: {:?}", val),
```
kkysen:
```suggestion
                    kind => panic!("String literal with unknown size: {val:?}, kind = {kind:?}"),
```

Rua:
Done!

PR #1364, opened 2025-09-16
Fix the reorganize_definitions refactoring

ahomescu:
Some fixes to the `reorganize_definitions` transform for c2rust-refactor to make it work again.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -701,9 +701,18 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
                         // FIXME: we should also check if items overlap
                         mod_items.extend(new_items.into_iter());
                     } else {
-                        let new_mod = mk().mod_(new_items);
-                        // TODO: do this again when we switch back to outline modules
-                        //new_mod.inline = inline;
+                        // TODO: we should outline the modules, but there
+                        // is currently an issue with the pretty-printer
+                        // where it both writes an output file, and the inline
+                        // module in the parent
+                        //
+                        //let modb = if inline {
+                        //    mk.inline()
+                        //} else {
+                        //    mk()
+                        //};
+                        let modb = mk().inline();
+                        let new_mod = modb.mod_(new_items);
                         let new_mod_item = mk()
                             .pub_()
                             .id(mod_info.id)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -937,13 +946,69 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
-                                    uses[namespace].insert(u.ident(), mod_id);
+                                    uses[namespace].insert(u.ident(), (mod_id, item.id, def_id));
                                 }
                             }
                         }
                     }
                     true
                 });
+
+                // If any items are imported into the current module,
```
ahomescu: This is an improvement, not a bug fix. I'm debating whether to split this into a separate PR, wdyt?

fw-immunant: If it can be separate and would shrink this PR, I'd say go for it. It makes it easier to pick out improvements (vs. restored functionality) in the changelog when we cut the next release.

kkysen: Yeah I agree, too. Smaller PRs would be much appreciated.

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        match mod_kind {
+            ModKind::Loaded(_items, Inline::Yes, _spans) => {
+                // TODO: handle #[path="..."]
+            }
+
+            ModKind::Loaded(_items, Inline::No, _spans) => {
+                // We shouldn't be seeing any loaded modules at this point
+                panic!("unexpected loaded module: {i:?}");
+            }
```
fw-immunant, line 39: The conjunction of these two comments is confusing: are inline modules not considered loaded modules (such that we might see them)? Or is it that specifically we shouldn't see any non-inline loaded modules yet?

c2rust-refactor/src/context.rs
```diff
@@ -48,6 +48,64 @@ impl<'a, 'tcx> RefactorCtxt<'a, 'tcx> {
...
+    fn nested_visit_map(&mut self) -> Self::Map {
+        self.hir_map
+    }
+
+    fn visit_id(&mut self, id: HirId) {
+        if let Some(ns) = hir_id_to_span(id, self.hir_map) {
+            let _old_id = self.span_to_hir_map.insert(ns, id);
+            // Sometimes we get different HirId's with the same span, e.g., Expr
+            // TODO: we still should assert here
+            //assert!(old_id.is_none(), "span {ns:?} already has id {old_id:?} != {id:?}");
```
fw-immunant: Should this `TODO` be addressed before merging, or is there a reason it can't be addressed yet?

c2rust-refactor/src/ast_manip/span_maps.rs
```diff
@@ -22,6 +22,12 @@ pub enum SpanNodeKind {
     Item,
     ForeignItem,
     AssocItem,
+    // We need a separate kind for ExprKind::Path because
+    // the rustc macro expander emits `*foo` expressions where
+    // both the outer and inner expressions have the same span
+    // We disambiguate by assigning (span, Expr) to the outer one,
+    // and (span, PathExpr) to the inner expression.
+    PathExpr,
```
fw-immunant, line 30: Is this similar to the `span_fix` module that `driver.rs` claims (in a 6-year-old comment) we should be using?

c2rust-refactor/src/ast_manip/span_maps.rs
```diff
@@ -0,0 +1,170 @@
...
+        let ns = NodeSpan { span, kind };
+        let old_ns = self.0.node_id_to_span_map.insert(id, ns);
+        let _old_id = self.0.span_to_node_id_map.insert(ns, id);
+
+        assert!(
+            old_ns.is_none(),
+            "id {id:?} already has span {old_ns:?} != {ns:?}"
+        );
+        // Some spans can show up in multiple nodes
+        //assert!(old_id.is_none(), "span {ns:?} already has id {old_id:?} != {id:?}");
```
kkysen, line 77: Isn't this redundant?

c2rust-refactor/src/context.rs
```diff
@@ -48,6 +48,70 @@ impl<'a, 'tcx> RefactorCtxt<'a, 'tcx> {
...
+        Some(Node::Ty(ty)) => Some(NodeSpan::new(ty.span, Ty)),
+        // We do not have a SpanNodeKind for TyBinding
+        // We do not have a SpanNodeKind for TraitRef
+        Some(Node::Pat(pat)) => Some(NodeSpan::new(pat.span, Pat)),
+        Some(Node::Arm(arm)) => Some(NodeSpan::new(arm.span, Arm)),
+        Some(Node::Block(block)) => Some(NodeSpan::new(block.span, Block)),
+        // We do not have a SpanNodeKind for Ctor
+        // We do not have a SpanNodeKind for Lifetime
+        // We do not have a SpanNodeKind for GenericParam
+        // We do not have a SpanNodeKind for Infer
```
kkysen: Are the comments for other `Node` variants?  It might make more sense then to keep the individual matches `=> None` instead of `_ => None`.

c2rust-refactor/src/context.rs
```diff
@@ -918,7 +1007,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
             (Some(ty1), Some(ty2)) => return self.structural_eq_tys(ty1, ty2),
             _ => {}
         }
-        match (self.cx.try_resolve_ty(ty1), self.cx.try_resolve_ty(ty1)) {
+        match (self.cx.try_resolve_ty(ty1), self.cx.try_resolve_ty(ty2)) {
```
kkysen, line 1014: Was this just a typo before?

c2rust-refactor/src/macros.rs
```diff
@@ -8,15 +8,25 @@ macro_rules! match_or {
...
+        $crate::match_or_else!([$e] $arm_pat => $arm_body;
+            |x| panic!("expected {}, got {:?}", stringify!($arm_pat), x))
     };
     ([$e:expr] $($arm_pat:pat => $arm_body:expr),*) => {
-        $crate::match_or!([$e] $($arm_pat => $arm_body),*;
-            panic!("expected one of: {}", stringify!($($arm_pat),*)))
+        $crate::match_or_else!([$e] $($arm_pat => $arm_body),*;
+            |x| panic!("expected one of: {}, got {:?}", stringify!($($arm_pat),*), x))
     };
 }
```
kkysen, line 31: What do we need this for?  Especially if it's not used, I'd rather not add to `expect!` given that `assert_matches!` does the same and is much more standardized.

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl<'a> MutVisitor for LoadModules<'a> {
+    fn flat_map_item(&mut self, mut i: P<Item>) -> SmallVec<[P<Item>; 1]> {
+        let Item {
+            ref mut attrs,
+            ref span,
+            ref ident,
+            kind: ItemKind::Mod(_, ref mut mod_kind),
+            ..
+        } = *i else {
```
kkysen:
```suggestion
        let Item {
            attrs,
            span,
            ident,
            kind: ItemKind::Mod(_, mod_kind),
            ..
        } = &mut *i else {
```

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -0,0 +1,84 @@
...
+            ModKind::Loaded(_items, Inline::No, _spans) => {
+                // We shouldn't be seeing any loaded modules at this point
+                panic!("unexpected loaded module: {i:?}");
+            }
+
+            ModKind::Unloaded => {
+                // Look for dir_path/foo.rs, then try dir_path/foo/mod.rs
+                let mut mod_file_path = self.dir_path.join(ident.as_str()).with_extension("rs");
+                if !self.source_map.file_exists(&mod_file_path) {
+                    mod_file_path = self.dir_path.join(ident.as_str()).with_file_name("mod.rs");
```
kkysen: This doesn't do `${dir_path}/${ident}/mod.rs`, it does `${dir_path}/mod.rs`.

c2rust-refactor/src/ast_manip/span_maps.rs
```diff
@@ -0,0 +1,170 @@
...
+impl NodeSpan {
+    pub fn new(span: Span, kind: SpanNodeKind) -> Self {
+        Self { span, kind }
+    }
+}
+
+#[derive(Default, Clone)]
+pub struct AstSpanMaps {
+    pub node_id_to_span_map: HashMap<NodeId, NodeSpan>,
+    pub span_to_node_id_map: HashMap<NodeSpan, NodeId>,
```
kkysen: Why do we use `HashMap` sometimes and `FxHashMap` other times?

c2rust-refactor/src/ast_manip/span_maps.rs
kkysen: Oh this is `old_id` not `old_ns`, nevermind.

c2rust-refactor/src/context.rs
ahomescu: I think so, yes. It doesn't make sense to match a type against itself.

c2rust-refactor/src/macros.rs
ahomescu: `expect!` is used in a ton of places in the project, it would be nice to get rid of it but that's a separate side project. While it's in use, I thought it would be useful for it to output better error messages (and I needed that during development).

c2rust-refactor/src/ast_manip/span_maps.rs
ahomescu: That module seems to fix two very specific things: the output of `format!`, and spans of items with attributes. I don't think we're seeing either of those here, for starters because the transpiler afaik doesn't emit `format!`.

c2rust-refactor/src/ast_manip/load_modules.rs
ahomescu: Inline modules are "loaded" from the root file, e.g., `lib.rs`. I updated the comment.

c2rust-refactor/src/context.rs
ahomescu: The line immediately above explains why we can't right now: we get different `HirId`s with the same span. We should still assert, but we'd need to refine the condition first.

c2rust-refactor/src/ast_manip/span_maps.rs
ahomescu: `HashMap` is the standard Rust type, while `FxHashMap` is the more optimized internal rustc one. I guess we could consistently use `FxHashMap` everywhere, there's no reason not to.

c2rust-refactor/src/context.rs
fw-immunant: Ah, that makes sense now. OK.

c2rust-refactor/src/ast_manip/span_maps.rs
fw-immunant: I wasn't sure if a shared underlying issue with spans of macro expansions was coming up in both situations. If it doesn't seem to you like they're the same problem, then no need to look further.

c2rust-refactor/src/ast_manip/span_maps.rs
kkysen: I think `FxHashMap` is also not randomly seeded.

PR #1368, opened 2025-09-19
fix removing literal suffixes from vec![10u32, 20u32, 30u32]

oinoom:
Fixes a bug where LazyTokenStreams were not being properly ignored

### DIFF
--- a/c2rust-refactor/gen/ast.txt
+++ b/c2rust-refactor/gen/ast.txt
@@ -32,7 +32,7 @@ struct Crate {
 
 #[rewrite_print_recover] #[rewrite_seq_item] #[rewrite_extra_strategies=item_header]
 #[nonterminal] #[extend_span]
-struct Item { ident, #[match=ignore] attrs, id, kind, vis, span, tokens }
+struct Item { ident, #[match=ignore] attrs, id, kind, vis, span, #[equiv_mode=ignore] #[match=ignore] #[rewrite_ignore] tokens }
 enum ItemKind {
     ExternCrate(name),
     Use(vp),
@@ -98,7 +98,7 @@ enum UseTreeKind {
 struct UseTree { kind, prefix, span }
 
 #[extend_span]
-struct AssocItem { id, ident, #[match=ignore] attrs, vis, kind, span, tokens }
+struct AssocItem { id, ident, #[match=ignore] attrs, vis, kind, span, #[equiv_mode=ignore] #[match=ignore] #[rewrite_ignore] tokens }
 enum AssocItemKind {
     Const(defaultness, ty, init),
     Fn(fn),
@@ -122,7 +122,7 @@ struct FieldDef { #[match=ignore] attrs, id, span, vis, ident, ty, is_placeholde
 
 struct ForeignMod { unsafety, abi, #[mac_table_seq] items }
 #[rewrite_print_recover] #[rewrite_seq_item] #[extend_span]
-struct ForeignItem { ident, #[match=ignore] attrs, kind, id, span, vis, tokens }
+struct ForeignItem { ident, #[match=ignore] attrs, kind, id, span, vis, #[equiv_mode=ignore] #[match=ignore] #[rewrite_ignore] tokens }
 enum ForeignItemKind {
     Fn(fn),
     Static(ty, mutbl, init),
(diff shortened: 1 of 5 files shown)


ahomescu:
This fixes the issue? Was a macro invocation also the problem in the transpiled code?

If the problem is `LazyTokenStream`, the problem might be that we're missing an attribute for the matcher for that node in https://github.com/immunant/c2rust/blob/ahomescu/restore_refactor/c2rust-refactor/gen/ast.txt#L590.

c2rust-refactor/tests/remove_literal_suffixes_token_stream/old.rs
```diff
@@ -0,0 +1,15 @@
+// Regression input for the token-stream mismatch caused by editing vec! literals.
+
+fn main() {
+    let ints = vec![10u32, 20u32, 30u32];
```
ahomescu, line 4: Are we adding, or deleting a token stream here?

oinoom: My impression is that, because we're editing the tokens, the cache of tokens is invalidated and rustc generates a new token stream. That mismatch in `None` token stream and `Some(TokenStream)` is the structural mismatch that caused a panic before and this changes.

> This fixes the issue?

yes

> If the problem is LazyTokenStream, the problem might be that we're missing an attribute for the matcher for that node

I'm not sure how to address this question. I think the crash was happening before any token stream comparison happened, i.e. before it recursed into `Some(TokenStream)`. Are you saying there could be some attribute that results in comparing two `Some(_)` values rather than `(None, Some)` such that a crash wouldn't have happened?

ahomescu: > Are you saying there could be some attribute that results in comparing two Some(_) values rather than (None, Some) such that a crash wouldn't have happened?

It shouldn't be doing the comparison in the first place, can you figure out where it's happening? We have `LazyTokenStream` set to ignore
```
#[equiv_mode=ignore] #[match=ignore] #[rewrite_ignore]
flag LazyTokenStream;
```
so the two streams should always compare as equal.

ahomescu: Working theory: I added those ignore flags on the `LazyTokenStream` type but the fields are actually `Option<LazyTokenStream>` so the ignores aren't applied as I expected. The solution would be to move the ignore attributes to the `tokens` fields everywhere:
```
    #[equiv_mode=ignore] #[match=ignore] #[rewrite_ignore] tokens,
```

oinoom: @ahomescu I believe I've addressed this in https://github.com/immunant/c2rust/commit/d0ef71c634e8e77a3bdb10b083d3209d4df896ba

c2rust-refactor/gen/ast.txt
```diff
@@ -32,7 +32,7 @@ struct Crate {
 
 #[rewrite_print_recover] #[rewrite_seq_item] #[rewrite_extra_strategies=item_header]
 #[nonterminal] #[extend_span]
-struct Item { ident, #[match=ignore] attrs, id, kind, vis, span, tokens }
+struct Item { ident, #[match=ignore] attrs, id, kind, vis, span, #[match=ignore] tokens }
```
ahomescu: `LazyTokenStream` has a few more ignores (see line 589). I'm thinking we
* Add all of them everywhere we have `tokens`, and
* Remove them from the `flag` definition.

c2rust-refactor/src/collapse/mac_table.rs
```diff
@@ -197,10 +197,26 @@ impl<'ast> MacTable<'ast> {
     }
 }
 
+#[derive(Clone, Debug, Default)]
```
ahomescu: Do we still need this?

c2rust-refactor/src/collapse/mac_table.rs
```diff
@@ -467,14 +501,86 @@ impl<T: CollectMacros + Debug> CollectMacros for Option<T> {
                 <T as CollectMacros>::collect_macros(old, new, cx);
             }
             (&None, &None) => {}
-            (_, _) => panic!(
-                "mismatch between unexpanded and expanded ASTs \n  old: {:?}\n  new: {:?}",
-                old, new
-            ),
+            (&None, &Some(ref new_val)) => {
```
ahomescu: Do we need this anymore?

PR #1371, opened 2025-09-19
transpile: separate top-level decl converting and emitting

kkysen:
Right now, we sort the top-level decls once by their source order (although we do this largely backwards, a separate issue) before we convert them and then emit them (insert in the code).  This means that changing the conversion order also changes the order that they're emitted in, which we don't want to do, so this separates the two steps.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -986,8 +986,10 @@ impl TypedAstContext {
         }
     }
 
-    pub fn sort_top_decls(&mut self) {
-        // Group and sort declarations by file and by position
+    /// Sort the top-level declarations by file and source location
+    /// so that we preserve the ordering of all declarations in each file.
+    /// This preserves the order when we emit the converted declarations.
+    pub fn sort_top_decls_for_emitting(&mut self) {
         let mut decls_top = mem::take(&mut self.c_decls_top);
         decls_top.sort_unstable_by(|a, b| {
             let a = self.index(*a);
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -688,18 +688,27 @@ pub fn translate(
             }
         }
 
-        // Export top-level value declarations
-        for top_id in &t.ast_context.c_decls_top {
-            use CDeclKind::*;
-            let needs_export = match t.ast_context[*top_id].kind {
-                Function { is_implicit, .. } => !is_implicit,
-                Variable { .. } => true,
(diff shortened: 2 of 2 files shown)


fw-immunant:
Interesting, the testsuite failure shows us rewriting an expansion of the `NTLMv2_BLOB_LEN` macro, but we shouldn't be treating this macro as successfully converted to a `const` because it references a free (at the definition site) variable:

```C
#define NTLMv2_BLOB_LEN       (44 -16 + ntlm->target_info_len + 4)
```

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -973,8 +973,10 @@ impl TypedAstContext {
         }
     }
 
-    pub fn sort_top_decls(&mut self) {
-        // Group and sort declarations by file and by position
+    /// Sort the top-level declarations by file and source location
+    /// so that we preserve the ordering of all declarations in each file.
+    /// This preserves the order when we emit the converted declarations.
+    pub fn sort_top_decls_for_emitting(&mut self) {
```
ahomescu, line 992: Nitpick: I'm not sure this function name needed to change.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -973,8 +973,10 @@ impl TypedAstContext {
...
         let mut decls_top = mem::take(&mut self.c_decls_top);
         decls_top.sort_unstable_by(|a, b| {
```
ahomescu, line 994: `compare_src_locs` seems to be doing a lot more than expected, but it looks fine. I'm not sure where the reordering or reversal are coming from.

ahomescu (approved):

c2rust-transpile/src/c_ast/mod.rs
kkysen: It's to contrast with `fn sort_top_decls_for_converting` in #1372.

c2rust-transpile/src/c_ast/mod.rs
kkysen: I'm moving the re-ordering to the where it's necessary (emitting) rather than converting.  This can be bundled with #1372, but clearly there's a bug in just this, so it helps to isolate the two.

kkysen (commented):
> Interesting, the testsuite failure shows us rewriting an expansion of the `NTLMv2_BLOB_LEN` macro, but we shouldn't be treating this macro as successfully converted to a `const` because it references a free (at the definition site) variable:
> 
> ```c
> #define NTLMv2_BLOB_LEN       (44 -16 + ntlm->target_info_len + 4)
> ```

This should be skipped by #1387, so merging this now.

PR #1373, opened 2025-09-21
Include libc dependency only when needed

### LINKED ISSUE #1362: Why is `use ::libc;` included in every file?
c2rust includes a `use ::libc;` at the top of every file, regardless of whether `libc` is used in that file or not. That results in a lot of warnings for unused imports.

From what I can tell, the offending code is this bit, which includes a `use` line for every external crate under certain circumstances:
https://github.com/immunant/c2rust/blob/master/c2rust-transpile/src/translator/mod.rs#L1033-L1040

What puzzles me is why this is done. It's not needed to `use` external crates in every file, you can just access them by name where needed. I was looking into removing this behaviour in a PR, but only if it's not actually needed for something.

Rua:
* Fixes #1362

Since most C types are now imported from `std::ffi`, there is less need to always include the `libc` library. This changes it so that it's only included when actually needed.

To implement this, `TypeConverter` now tracks external crates, separately from `Translator`. The two lists are merged afterwards. While working on this, I also spotted a bug that `f128` wasn't being included from `TypeConverter` and a few other places, so I included fixes for that as well.

### DIFF
--- a/c2rust-transpile/src/convert_type.rs
+++ b/c2rust-transpile/src/convert_type.rs
@@ -2,8 +2,10 @@ use crate::c_ast::CDeclId;
 use crate::c_ast::*;
 use crate::diagnostics::TranslationResult;
 use crate::renamer::*;
+use crate::{CrateSet, ExternCrate};
 use c2rust_ast_builder::{mk, properties::*};
 use failure::format_err;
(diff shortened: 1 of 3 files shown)


kkysen (approved):
Thanks!!

PR #1375, opened 2025-09-22
Ensure crate name is a valid identifier

Rua:
Transpile would panic if given an `output_dir` whose name was not a valid identifier. This ensures that it's converted to a valid name if needed.

### DIFF
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -145,7 +145,7 @@ impl TranspilerConfig {
         self.output_dir
             .as_ref()
             .and_then(|dir| dir.file_name())
-            .map(|fname| str_to_ident_checked(fname.to_string_lossy(), true))
+            .map(|fname| str_to_ident_checked(fname.to_string_lossy().as_ref(), true))
             .unwrap_or_else(|| "c2rust_out".into())
     }
 }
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/lib.rs
```diff
@@ -231,7 +231,10 @@ fn get_module_name(
         file.file_name()
     };
     let fname = &fname.unwrap().to_str().map(String::from);
-    let mut name = str_to_ident_checked(fname, check_reserved).unwrap();
+    let mut name = fname
+        .as_ref()
+        .map(|fname| str_to_ident_checked(fname, check_reserved))
+        .unwrap();
```
kkysen:
```suggestion
    let fname = fname.unwrap().to_str().unwrap();
    let mut name = str_to_ident_checked(fname, check_reserved);
```

kkysen: We don't need to allocate an extra `String` here since it'll just get converted back to a `&str` right away.

c2rust-transpile/src/lib.rs
```diff
@@ -144,7 +144,8 @@ impl TranspilerConfig {
     fn crate_name(&self) -> String {
         self.output_dir
             .as_ref()
-            .and_then(|x| x.file_name().map(|x| x.to_string_lossy().into_owned()))
+            .and_then(|dir| dir.file_name())
+            .map(|fname| str_to_ident_checked(fname.to_string_lossy(), true))
```
kkysen:
```suggestion
            .map(|fname| str_to_ident_checked(fname.to_string_lossy().as_ref(), true))
```

c2rust-transpile/src/lib.rs
```diff
@@ -206,16 +207,15 @@ fn str_to_ident<S: AsRef<str>>(s: S) -> String {
...
-    // module names cannot contain periods or dashes
-    filename.as_ref().map(str_to_ident).map(|module| {
-        // make sure the module name does not clash with keywords
-        if check_reserved && RESERVED_NAMES.contains(&module.as_str()) {
-            format!("r#{}", module)
-        } else {
-            module
-        }
-    })
+fn str_to_ident_checked<S: AsRef<str>>(s: S, check_reserved: bool) -> String {
```
kkysen:
```suggestion
fn str_to_ident_checked(s: &str, check_reserved: bool) -> String {
```

Just to simplify things, generics aren't really that helpful here, so can we just use `&str`?  `fn str_to_ident` needs to be updated to `&str` as well.

kkysen (changes requested):
Thanks for fixing this, too!  Suggested a couple of simplifications.

PR #1381, opened 2025-09-27
transpile: use `core::ffi::c_*` instead of `std::ffi::c_*`

### LINKED ISSUE #1356: --emit-no-std broke (regression after 0.20)
At some time after 0.20, `--emit-no-std` broke. Minimal example:

`compile_commands.json`:

```json
[
  {
    "arguments": [
      "any-cc",
      "empty.h"
    ],
    "directory": "/tmp/test",
    "file": "issue.h"
  }
]
```

`issue.h`:

```c
int x;
```

command:

```sh-command
$ c2rust transpile compile_commands.json --emit-no-std
```

output in `issue.rs`:

```rust
#![allow([elided])]
#![no_std]
#[no_mangle]
pub static mut x: std::ffi::c_int = 0;
```

In 0.20, this produced:
```rust
#![allow([elided])]
#![no_std]
#[no_mangle]
pub static mut x: libc::c_int = 0;
```

kkysen:
* Fixes #1356.

This leaves alone `std` items that could be `alloc`, like `std::vec::from_elem` or `std::ffi::CString::new` (needs a newer `rustc` 1.63), but it should address most cases.


kkysen (commented):
@chrysn, does this look good for your use case?

PR #1384, opened 2025-10-01
ci: run testsuite python2 transpilation

fw-immunant:
users have run into issues with the testsuite's python2, so we should test it in CI

more coverage is better, and it's not clear why this wasn't being included before

### DIFF
--- a/.github/workflows/internal-testsuite.yml
+++ b/.github/workflows/internal-testsuite.yml
@@ -96,8 +96,11 @@ jobs:
         sudo apt-get -qq install    \
             build-essential         \
             libbrotli-dev           \
+            libbz2-dev              \
             libclang-${{ matrix.clang-version }}-dev \
+            libdb-dev               \
             libgcrypt20             \
+            libgdbm-dev             \
             libreadline-dev         \
             libidn2-dev             \
             libldap2-dev            \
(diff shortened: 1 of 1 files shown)


.github/workflows/internal-testsuite.yml
```diff
@@ -129,7 +133,7 @@ jobs:
         find testsuite -type f -name compile_commands.json -delete
         export PATH=$PWD/target/release:$HOME/.local/bin:$PATH
         echo "PATH=$PATH"
-        python3 testsuite/test.py curl json-c lua nginx zstd
+        python3 testsuite/test.py curl json-c lua nginx zstd python2
```
kkysen: Lol python3 running python2 looks funny.

fw-immunant:
After [disabling a test that tries to contact a nonexistent network host](https://github.com/immunant/c2rust-testsuite/commit/5b5a098a39d5431383f46410b838e24321a4cb25) the testsuite now passes. I don't think the changes from the `kkysen/const-macros-skip-fn-ptrs` branch are necessary for this PR, as that `test_imaplib` case was the [only one failing](https://github.com/immunant/c2rust/actions/runs/18205360382/job/51834319866) for 620d453df3007ac25e5f596bcc28ea0ccace4664.

PR #1385, opened 2025-10-02
ci: enable `libxml2`

kkysen:


fw-immunant:
This should probably merge before #1384 which will conflict (textually, not semantically) with it.

Issue #1366, opened 2025-09-19
Invalid CLZ macro in libxml2


ahomescu:
Transpiling libxml2 creates an invalid `CLZ` macro that's a redefinition of the C-specific `__builtin_clzll`:
```rust
    #[c2rust::src_loc = "78:9"]
    pub const CLZ: unsafe extern "C" fn(std::ffi::c_ulonglong) -> std::ffi::c_int =
        unsafe { __builtin_clzll };
```

fw-immunant:
Shouldn't this be failing in CI, since the testsuite includes libxml2? What commit does this occur with?

ahomescu:
Reproduced on cbf4e0a20a54fcea886813e9f76394d049e7b75b, let me try current master.

ahomescu:
Still happening on 973811e17d14cbb96cb84a6ff3298bc3171be240.

ahomescu:
> Shouldn't this be failing in CI, since the testsuite includes libxml2? What commit does this occur with?

I'm seeing this on libxml2 which [seems to be disabled in CI](https://github.com/immunant/c2rust/blob/master/.github/workflows/internal-testsuite.yml#L132).

kkysen:
I think this is happening because `__builtin_clzll` is an intrinsic that we normally translate as `.leading_zeros()`, but this isn't working when the function and the call are separated through a const macro.

fw-immunant:
> I think this is happening because `__builtin_clzll` is an intrinsic that we normally translate as `.leading_zeros()`, but this isn't working when the function and the call are separated through a const macro.

Yeah, we correctly transpile calls of builtins but not taking their value.

ahomescu:
I think this has been fixed, libxml2 seems to work fine.

kkysen:
Oh I was leaving this open because the underlying issue isn't fixed yet.

Issue #1321, opened 2025-08-13
transpile: const macros aren't translated with portable types


kkysen:
This is an offshoot of #1320.  After implementing the partial fix described in #1320 of inserting casts at const macro use sites, we still want to properly translate the macro with a portable type, as otherwise the type isn't portable once again, just like before #1266.

fw-immunant:
This does seem to actually cause us to transpile some code to Rust that doesn't compile (minimized from python2's `bytes_methods.c`):
```C
#include <string.h>

#define Py_MEMCPY memcpy
void f(void* x) {
	size_t len = 0;
	Py_MEMCPY(x, x, len);
}
```
transpiles to:
```rust
#![allow(
    dead_code,
    mutable_transmutes,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
extern "C" {
    fn memcpy(
        __dest: *mut std::ffi::c_void,
        __src: *const std::ffi::c_void,
        __n: size_t,
    ) -> *mut std::ffi::c_void;
}
pub type size_t = usize;
pub const Py_MEMCPY: unsafe extern "C" fn(
    *mut std::ffi::c_void,
    *const std::ffi::c_void,
    std::ffi::c_ulong,
) -> *mut std::ffi::c_void = unsafe { memcpy };
#[no_mangle]
pub unsafe extern "C" fn f(mut x: *mut std::ffi::c_void) {
    let mut len: size_t = 0 as size_t;
    memcpy(x, x, len);
}
```

Which results in:
```
error[E0308]: mismatched types
  --> memcpy_macro.rs:22:39
   |
22 | ) -> *mut std::ffi::c_void = unsafe { memcpy };
   |                                       ^^^^^^ expected fn pointer, found fn item
   |
   = note: expected fn pointer `unsafe extern "C" fn(_, _, u64) -> _`
                 found fn item `unsafe extern "C" fn(_, _, usize) -> _ {memcpy}`

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0308`.
```

Rua:
Considering a possible solution to this. As I understand it, the problem is that the `override_ty`s of all the expansion sites are not available when translating a `MacroObject`. So there needs to be a way to query the `override_ty` of an arbitrary expression, so that it doesn't need to be passed down. Thinking that could be solved with a preprocessing pass that goes through the whole tree, and fills a map of `CExprId` to `override_ty`?

PR #1386, opened 2025-10-02
transpile: const macros: skip fn ptr types, as they don't work with portable types

kkysen:
The workaround for non-portable types for const macros is to insert a use-site cast to the portable type, but this works differently for fn ptrs, so while #1321 is fixed, just skip fn ptr const macros.

This should get #1384 to pass CI by working around this issue there, and it should also workaround #1366, although #1366 in `libxml2` appears to have been fixed/worked around from some other change, as #1385 now passes CI.


fw-immunant:
I don't think this commit alters the behavior of the `Py_MEMCPY` test case it adds relative to `master`. I think that's because the scope of #1387 was larger than intended, covering all `DeclRef`s, not just those that reference variables (cf. functions).

fw-immunant:
> Ah, this PR used to before the `DeclRef` PR. I'm pretty sure we had tested that and it fixed that issue. Should we still merge this?

I think the motivation behind this is still sound, but a test case for this would need to involve function pointer types without mentioning any decls, which is probably tricky while avoiding casts that are forbidden during const execution.

PR #1388, opened 2025-10-02
Fix z3 in CI for restored c2rust-refactor

fw-immunant:
I think this requires less fiddly setup with newer versions of the `z3` crate, but we can't use those because they switch to the 2024 edition, which our pinned nightly from 2023 doesn't know about.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,8 @@ jobs:
       - name: cargo test --release --workspace
         run: |
           export RUSTFLAGS="-D warnings"
+          # add homebrew library path for z3
+          export RUSTFLAGS="$RUSTFLAGS -Clink-arg=-L/opt/homebrew/lib -Clink-arg=-Wl,-rpath,/opt/homebrew/lib"
           export RUSTDOCFLAGS="-D warnings"
           cargo test --release --workspace
       - name: Test translator
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
             librtmp-dev
             libssl-dev
             libtool
+            libz3-dev
             llvm
             llvm-dev
             luarocks


ahomescu:
Don't we also have a problem on mac OS? I couldn't figure out how to solve that one.

ahomescu:
Couldn't this have been a pull request from https://github.com/immunant/c2rust/pull/1328, so it only shows your new commits?

ahomescu:
> Couldn't this have been a pull request from #1328, so it only shows your new commits?

I mean, I _can_ cherry-pick your fixes once they work, but PR-against-PR might have been cleaner.

fw-immunant:
> > Couldn't this have been a pull request from #1328, so it only shows your new commits?
> 
> I mean, I _can_ cherry-pick your fixes once they work, but PR-against-PR might have been cleaner.

I just opened a PR to run CI; I can set target to your branch if you want. I wasn't sure if this would need a rebase/etc. first and wanted to avoid noise in the original PR, or I would have just pushed to that branch.

ahomescu:
> > > Couldn't this have been a pull request from #1328, so it only shows your new commits?
> > 
> > 
> > I mean, I _can_ cherry-pick your fixes once they work, but PR-against-PR might have been cleaner.
> 
> I just opened a PR to run CI; I can set target to your branch if you want. I wasn't sure if this would need a rebase/etc. first and wanted to avoid noise in the original PR, or I would have just pushed to that branch.

Just nitpicking and trying to make our lives easier, appreciate the help! I'm not sure when I'll get around to rebase my original branch, but I could roll your fixes into it once they pass CI.

fw-immunant:
> I'm not sure when I'll get around to rebase my original branch, but I could roll your fixes into it once they pass CI.

Despite GH's claims to the contrary, a simple `git rebase master` seems to work without any intervention locally.

fw-immunant:
macOS CI now works... but for some reason the testsuite is broken.

fw-immunant:
e2fbf8c9eb8022b45b97d331d796935296e036c5 [failed the z3 test](https://github.com/immunant/c2rust/actions/runs/18205368160/job/51834346515?pr=1388#step:11:580) claiming quickcheck failed:
```
---- transform::casts::tests::verify_double_cast stdout ----
thread 'transform::casts::tests::verify_double_cast' panicked at '[quickcheck] TEST FAILED. Arguments: (PointerWidth(64), [Int(32, true), Int(32, false), Size(true)])', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-0.9.2/src/tester.rs:178:28
```

Upon further investigation, it looks like this is a real test failure, but not macOS-related. That CI run just got lucky with quickcheck seeds. I'll submit a follow-up PR.

ahomescu (approved):

PR #1391, opened 2025-10-03
Narrow const macro `DeclRef` exclusion to only variables

fw-immunant:
Mentioning enum variants, etc. in const contexts doesn't have the same scoping problems as variables. This change narrows the exception to const macro translation for `DeclRef`s and makes it more distinct from what #1386 covers.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3338,13 +3338,15 @@ impl<'c> Translation<'c> {
                     .ok_or_else(|| format_err!("Missing declref {:?}", decl_id))?
                     .kind;
                 if ctx.is_const {
-                    // TODO Determining which declarations have been declared within the scope of the const macro expr
-                    // vs. which are out-of-scope of the const macro is non-trivial,
-                    // so for now, we don't allow const macros referencing any declarations.
-                    return Err(format_translation_err!(
-                        self.ast_context.display_loc(src_loc),
-                        "Cannot yet refer to declarations in a const expr",
-                    ));
+                    if let CDeclKind::Variable { .. } = decl {
+                        // TODO Determining which variables have been declared within the scope of the const macro expr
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -3338,13 +3338,15 @@ impl<'c> Translation<'c> {
...
                     .kind;
                 if ctx.is_const {
-                    // TODO Determining which declarations have been declared within the scope of the const macro expr
-                    // vs. which are out-of-scope of the const macro is non-trivial,
-                    // so for now, we don't allow const macros referencing any declarations.
-                    return Err(format_translation_err!(
-                        self.ast_context.display_loc(src_loc),
-                        "Cannot yet refer to declarations in a const expr",
-                    ));
+                    if let CDeclKind::Variable { .. } = decl {
```
kkysen, line 3341: I didn't do this initially because I wasn't sure about other declarations within a function and how that interacted with global scope, so I wanted to err on the safer side, since this is supposed to be temporary.

c2rust-transpile/tests/snapshots/macros.c
```diff
@@ -405,3 +405,7 @@ unsigned int ntlm_v2_blob_len(struct ntlmdata *ntlm) { return NTLMv2_BLOB_LEN; }
 int late_init_var() {
   return LATE_INIT_VAR;
 }
+
+enum foo { Variant1, };
```
kkysen, line 409: How does this work if this is defined within a function, the macro is defined outside of it, and used within it?

Rua: Doesn't this problem currently exist already, if the macro references a *type* by name? If the macro contained a cast to that enum for example.

EDIT: No apparently it does not, because enums are always placed in global scope by the transpiler. But then that gives a guarantee for `DeclRef`s too.

PR #1392, opened 2025-10-03
Fix quickcheck-derived test case for refactor casts

fw-immunant:
A CI run randomly found this cast sequence that fails. I'm not sure what to do with it at this point, but this PR serves to at least document what's going on.

### DIFF
--- a/c2rust-refactor/src/transform/casts.rs
+++ b/c2rust-refactor/src/transform/casts.rs
@@ -166,7 +166,7 @@ fn check_double_cast<'tcx>(e_ty: SimpleTy, t1_ty: SimpleTy, t2_ty: SimpleTy) -> 
         | (SameWidth, Extend(s))
         | (SameWidth, FromPointer(s))
         | (SameWidth, ToPointer(s))
-            if s == e_ty.is_signed() =>
+            if s == e_ty.is_signed() && s == t1_ty.is_signed() =>
         {
             DoubleCastAction::RemoveInner
         }


ahomescu:
Your branch needs a rebase (maybe once we land the base PR), then I'll take a look.

c2rust-refactor/src/transform/casts/tests.rs
```diff
@@ -63,48 +63,53 @@ fn cast_tys<'bv>(bv: BV<'bv>, tys: &[SimpleTy], pw: PointerWidth) -> BV<'bv> {
...
 thread_local!(static Z3_CONTEXT: Context = Z3_CONFIG.with(|cfg| Context::new(cfg)));
 
-quickcheck! {
-    // Verify `check_double_cast` using QuickCheck and Z3
-    fn verify_double_cast(pw: PointerWidth, tys: Vec<SimpleTy>) -> bool {
-        if tys.len() <= 1 {
-            return true;
-        }
+// Verify `check_double_cast` using Z3
+fn verify_double_cast(pw: PointerWidth, tys: Vec<SimpleTy>) -> bool {
```
ahomescu, line 83: Why did you need to split this?

c2rust-refactor/src/transform/casts/tests.rs
```diff
@@ -63,6 +63,18 @@ fn cast_tys<'bv>(bv: BV<'bv>, tys: &[SimpleTy], pw: PointerWidth) -> BV<'bv> {
 thread_local!(static Z3_CONFIG: Config = { let mut c = Config::new(); c.set_model_generation(true); c });
 thread_local!(static Z3_CONTEXT: Context = Z3_CONFIG.with(|cfg| Context::new(cfg)));
 
+#[test]
+fn test_cast_i32_u32_ssize() {
+    verify_double_cast(
+        PointerWidth(64),
+        vec![
+            SimpleTy::Int(32, true),
```
ahomescu, line 75: `x as i32 as u32 as isize` is not correct to shorten to `x as i32 as usize` because it would sign-extend, but isn't `x as u32 as isize` fine?

c2rust-refactor/src/transform/casts/tests.rs
fw-immunant: So it could be called with fixed arguments from the `test_cast_i32_u32_ssize` test. It seems like there might be a way to give quickcheck a particular seed to work from, but our test output didn't provide that seed so instead we need to pass particular values directly to reproduce the failure.

ahomescu: Ah got it.

c2rust-refactor/src/transform/casts/tests.rs
fw-immunant: I revised the counterexample printing to be clearer:
```
thread 'transform::casts::tests::test_cast_i32_u32_ssize' panicked at '64b counterexample:
#x80000000 =>
 #x0000000080000000 via tys [Int(32, true), Int(32, false), Size(true)] vs.
 #xffffffff80000000 via min_tys [Int(32, true), Size(true)]', c2rust-refactor/src/transform/casts/tests.rs:122:13
```
So it seems that we can't actually shrink the cast sequence in this case. So presumably calling `check_double_cast(Int(32, true), Int(32, false), Size(true))` should actually return `KeepBoth`. I don't know if it's `check_double_cast` or `cast_kind` that has the real logic bug though.

fw-immunant: Actually, looking a bit closer--the first cast (i32 -> u32) will be assigned `SameWidth` by `cast_kind`, which seems sane. The second (u32 -> isize) will be assigned `ToPointer(true)`, which also seems right. But in this case `check_double_cast`'s match has this arm:
```rust
        (Extend(_), Extend(s))
        | (SameWidth, Extend(s))
        | (SameWidth, FromPointer(s))
        | (SameWidth, ToPointer(s))
            if s == e_ty.is_signed() =>
        {
            DoubleCastAction::RemoveInner
        }
```

`s` is true here, and `e_ty` (i32) is signed, so it returns `RemoveInner`. But this ignores the fact that the first cast does change the interpreted value used by the second cast. So it seems like this guard is wrong, or needs to be refined.

c2rust-refactor/src/transform/casts/tests.rs
```diff
@@ -60,51 +60,79 @@ fn cast_tys<'bv>(bv: BV<'bv>, tys: &[SimpleTy], pw: PointerWidth) -> BV<'bv> {
     tys.windows(2).fold(bv, |y, w| cast_bv(y, w[0], w[1], pw))
 }
 
-thread_local!(static Z3_CONFIG: Config = Config::new());
+thread_local!(static Z3_CONFIG: Config = { let mut c = Config::new(); c.set_model_generation(true); c });
```
ahomescu: Huh the `Config` setters do not return `self`, that would have made this much nicer.

c2rust-refactor/src/transform/casts.rs
```diff
@@ -166,7 +166,7 @@ fn check_double_cast<'tcx>(e_ty: SimpleTy, t1_ty: SimpleTy, t2_ty: SimpleTy) ->
         | (SameWidth, Extend(s))
         | (SameWidth, FromPointer(s))
         | (SameWidth, ToPointer(s))
-            if s == e_ty.is_signed() =>
+            if s == e_ty.is_signed() && s == t1_ty.is_signed() =>
```
ahomescu, line 169: _I think_ for the `SameWidth` case we don't care if `s == e_ty.is_signed()`, but we do for the left `Extend`. Maybe split those cases up?

ahomescu (approved):

c2rust-refactor/src/transform/casts/tests.rs
kkysen: Also, nit, but could you format this more like normal Rust?  I think rustfmt is turned off because it's in a macro.

c2rust-refactor/src/transform/casts.rs
fw-immunant: If you mean:
```diff
             DoubleCastAction::RemoveBoth
         }
 
-        (Extend(_), Extend(s))
-        | (SameWidth, Extend(s))
-        | (SameWidth, FromPointer(s))
-        | (SameWidth, ToPointer(s))
-            if s == e_ty.is_signed() && s == t1_ty.is_signed() =>
+        (Extend(_), Extend(s)) if s == e_ty.is_signed() && s == t1_ty.is_signed() => {
+            DoubleCastAction::RemoveInner
+        }
+        (SameWidth, Extend(s)) | (SameWidth, FromPointer(s)) | (SameWidth, ToPointer(s))
+            if s == e_ty.is_signed() =>
         {
             DoubleCastAction::RemoveInner
         }
```

then it seems like we do care; the test comes up red:

```
test transform::casts::tests::test_cast_i32_u32_ssize ... FAILED
test transform::casts::tests::quickcheck_double_cast ... ok

failures:

---- transform::casts::tests::test_cast_i32_u32_ssize stdout ----
thread 'transform::casts::tests::test_cast_i32_u32_ssize' panicked at '64b counterexample:
#x80000000 =>
 #x0000000080000000 via tys [Int(32, true), Int(32, false), Size(true)] vs.
 #xffffffff80000000 via min_tys [Int(32, true), Size(true)]', c2rust-refactor/src/transform/casts/tests.rs:122:13
```

fw-immunant:
I'd like to go ahead and merge this after CI, as this bug is the cause of occasional intermittent failures in CI ([example](https://github.com/immunant/c2rust/actions/runs/19643883354/attempts/1)).

If we want to make the check less conservative again in the `SameWidth` case, we can figure out in exactly which cases it's sound to do so as a follow-up. I don't think this PR regresses the output of this refactoring pass much in practice.

c2rust-refactor/src/transform/casts.rs
ahomescu: Eesh, you're right, `(SameWidth, Extend(s))` will flip between zero- and sign-extension.

ahomescu:
I can't think of another way to simplify it, so LGTM, let's merge it.

PR #1393, opened 2025-10-05
disable warning from clang in snapshot test

bungcip:
disable warning from clang phase when transpiling c code in snapshot test


kkysen (approved):
Oh great!  I didn't realize there was already an argument to `c2rust_transpile` specifically for this.  I had thought we had to create our own `compile_commands.json`, but this is a lot simpler.  Thanks!

PR #1394, opened 2025-10-05
[WIP] Visit and handle `CountAttributedType` in `c2rust-ast-exporter`

bungcip:
still WIP.  because `__counted_by__` only exist in clang 18, i just want to see the compile error when compiled in older clang version.


bungcip:
hmm, clang 18 also fail. in my system (clang 21) this PR can be compiled...

kkysen (commented):
For the refactor commits like f89084e and 13bae2c, just because they're so whitespace-heavy and make a big diff (even though semantically they're not large), could you put them in their own PR?  You can wait until this PR is ready to review, though, in case any other changes are needed.

bungcip:
tested on clang 18, 19, and 21. not testing on 20, hopefully its working because i can compile it in 21.

PR #1397, opened 2025-10-07
Document current relooper algorithm and do minor cleanup

randomPoison:
I've gone through the current relooper algorithm and added documentation and comments to clarify what the algorithm is currently doing. The module docs now have an overview of what the algorithm does, and I've added many verbose comments to help understand what each step of the algorithm is doing and why we do the things we do. I've aired on the side of verbosity for many of these comments, trying to capture all of the information I needed to understand the algorithm.

I've also done a small number of cleanup changes in places where the code was doing something unnecessary or was expressed in a way that I found hard to follow. ~~None of the changes should change behavior, expect for the one place where I added a `panic!` to surface an error case that was previously being silently ignored.~~

Note that this is not meant to be reviewed one commit at a time. Intermediate commits sometimes have incorrect comments that I fix in later commits.

### DIFF
--- a/c2rust-transpile/src/cfg/loops.rs
+++ b/c2rust-transpile/src/cfg/loops.rs
@@ -180,12 +180,13 @@ impl<Lbl: Hash + Eq + Clone> LoopInfo<Lbl> {
         // be widened.
         for entry in entries {
             loop {
-                let (in_loop, parent_id) = self.loops.get(&loop_id).unwrap_or_else(|| {
-                    panic!(
-                        "Found loop ID {:?} with no corresponding loop info",
-                        loop_id,
-                    )
-                });
+                // NOTE: We currently bail out in the case where there is no loop data
+                // corresponding to the current loop ID, but it's unclear if that's the right
+                // thing to do. In theory we should always have loop data corresponding to a
+                // loop ID, but the `nested_goto.c` test definitely ends up with a loop ID that
+                // doesn't have corresponding info. We should investigate why that happens and
+                // determine if it's valid or not.
+                let (in_loop, parent_id) = self.loops.get(&loop_id)?;
 
                 // If our current loop contains the entry, move on to the next entry. Otherwise
                 // move on to the next wider loop if there is one.


randomPoison:
Ah, I see my error case that should never happen has happened. I'll look into that and likely just revert the `panic!` I added.

c2rust-transpile/src/cfg/relooper.rs
```diff
@@ -292,59 +418,122 @@ impl RelooperState {
...
                 .map(|(lbl, bb)| (lbl.clone(), bb.successors()))
                 .collect();
 
+            // Calculate the transitive closure of the successor map, i.e. the full set of
+            // labels reachable from each block, not including itself. Then flip the edges
+            // to get a map from a label to the set of labels that can reach it.
             let strict_reachable_from = flip_edges(transitive_closure(&successor_map));
+
+            // Flip the edges of the successor map to get a map of immediate predecessors
+            // for each block.
```
fw-immunant, line 448: These comments don't seem to add information beyond exactly what the code is doing, so I'd remove them in favor of concision.

c2rust-transpile/src/cfg/relooper.rs
```diff
@@ -292,59 +418,122 @@ impl RelooperState {
...
             let predecessor_map = flip_edges(successor_map);
 
             (predecessor_map, strict_reachable_from)
         };
 
-        // Try to match an existing branch point (from the initial C). See `MultipleInfo` for more
-        // information on this.
+        // Try to match an existing branch point (from the initial C).
+        //
```
fw-immunant, line 455: Remove empty comment line..

c2rust-transpile/src/cfg/relooper.rs
```diff
@@ -292,59 +418,122 @@ impl RelooperState {
...
+            // Flip the edges of the successor map to get a map of immediate predecessors
+            // for each block.
             let predecessor_map = flip_edges(successor_map);
 
             (predecessor_map, strict_reachable_from)
         };
 
-        // Try to match an existing branch point (from the initial C). See `MultipleInfo` for more
-        // information on this.
+        // Try to match an existing branch point (from the initial C).
```
fw-immunant, line 454: Leave the reference to `MultipleInfo`, but fix the existing `/// TODO: document me` in its documentation.

c2rust-transpile/src/cfg/relooper.rs
```diff
@@ -292,59 +418,122 @@ impl RelooperState {
...
+        //
+        // We do this before creating a loop to better handle the cases where we have
+        // multiple entries that are part of disjoint loops. The loop analysis that
```
fw-immunant, line 457: What do you mean by "disjoint" here? In the comment at the top of `multiples.rs`, it shows two CFG edges (the not-taken side of an `if` and the exit of a `while` loop) that coincide. But these two constructs are nested, not disjoint.

c2rust-transpile/src/cfg/relooper.rs
```diff
@@ -292,59 +418,122 @@ impl RelooperState {
...
+        // branches, this logic will recreate them, avoiding the problem of merging
+        // loops accidentally.
         let mut recognized_c_multiple = false;
         if let Some(ref multiple_info) = self.multiple_info {
             let entries_key = entries.iter().cloned().collect();
             if let Some((join, arms)) = multiple_info.get_multiple(&entries_key) {
                 recognized_c_multiple = true;
 
-                for (entry, content) in arms {
+                // Search through the arms of the C multiple to verify that our current CFG matches
```
fw-immunant, line 471: Calling this a "C multiple" is confusing. Most of the code just calls this a `Multiple`, even though this information *does* come from C.

c2rust-transpile/src/cfg/relooper.rs
```diff
@@ -292,59 +418,122 @@ impl RelooperState {
...
                             }
 
                             to_visit.extend(bb.successors())
                         }
                     }
 
-                    // Check we've actually visited all of the expected content
+                    // Check that we've only visited the expected nodes. If we've visited any nodes
+                    // not in `content`, then the CFG differs from the C multiple and the search
+                    // fails.
```
fw-immunant, line 506: The logic here does seem to do what you describe, not what the comment said before. Should it be doing what it said before (ensuring that all of `content` is in `visited` rather than vice-versa), though? Intuitively I thought we wanted to find an exact match rather than a (possibly proper) subset.

The opposite direction, ensuring that nodes we visit are always in `content`, seems to be covered by the `if !content.contains(&lbl) {` check a few lines earlier.

fw-immunant (commented):
Various comments inline.

randomPoison:
A lot of this documentation is now out of date with all of the reworkings to relooper, so I don't think it's worth trying to clean this up to land first. I'll try to remember to address the feedback here as part of the upcoming PR(s) with the relooper changes.

PR #1399, opened 2025-10-07
reorganize_definitions: Shorten paths that are imported into the current module

ahomescu:
Add an extra step to `reorganize_definitions` that checks for every path if it is already imported into the current module, and uses the short import instead of the fully qualified path. E.g.
```Rust
use crate::foo::Bar;
// Was: fn foo(x: crate::foo::Bar) {}
fn foo(x: Bar) {}
```


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -946,13 +946,69 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
                             }
                         }
                     }
                     true
                 });
+
+                // If any items are imported into the current module,
+                // use the short paths, e.g., bar instead of crate::foo::bar
+                //
+                // TODO: this could be a separate transform
```
kkysen: What transform is it part of?  Reorganize definitions?  It definitely seems super useful on its own.

ahomescu:
> I agree with @kkysen that it would be nice to have this in a separate pass.

I'm not against splitting it, I just have to figure out how to do that. The current implementation uses a lot of internal state from `reorganize_definitions`.

fw-immunant:
While this would be really useful as its own transform, since it appears that would be a substantial effort I'd vote in favor of going ahead and merging as-is this to benefit from it as part of `reorganize_definitions` in the meantime.

ahomescu:
> @ahomescu, did you see my earlier comment here [#1408 (comment)](https://github.com/immunant/c2rust/pull/1408#discussion_r2423094412)?

Sorry, I missed it until now. Do we still want to try that?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -956,13 +956,69 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                                    uses[namespace].insert(u.ident(), (mod_id, item.id, def_id));
                                 }
                             }
                         }
                     }
                     true
                 });
+
+                // If any items are imported into the current module,
+                // use the short paths, e.g., bar instead of crate::foo::bar
```
kkysen:
```suggestion
                // use the short paths, e.g., `bar` instead of `::crate::foo::bar`.
```

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -956,13 +956,69 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                            let path_def_mod_id = remapped_paths
+                                .get(&id)
+                                .map(|(mod_def_id, _)| *mod_def_id)
+                                .or_else(|| path_def_id.as_local().map(|ldid| {
+                                    let parent = self.cx.ty_ctxt().parent_module_from_def_id(ldid);
+                                    self.cx.hir_map().local_def_id_to_node_id(parent)
+                                }));
+                            if Some(item_mod_id) == path_def_mod_id {
+                                // Current path matches one of the definitions
+                                // in the current module, shorten it
```
kkysen:
```suggestion
                                // in the current module; shorten it.
```

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -956,13 +956,69 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                                path.segments.drain(0..path.segments.len() - 1);
```
kkysen:
```suggestion
                                path.segments.drain(..path.segments.len() - 1);
```

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -956,13 +956,69 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                                continue;
+                            }
+
+                            // For now, only shorten paths to uses within the same module
+                            let use_hir_id = self.cx.hir_map().node_to_hir_id(*use_id);
+                            let use_mod_ldid = self.cx.ty_ctxt().parent_module(use_hir_id);
+                            if item_mod_ldid == use_mod_ldid {
+                                // Current path matches one of the uses
+                                // in the current module, shorten it
+                                path.segments.drain(0..path.segments.len() - 1);
```
kkysen:
```suggestion
                                // in the current module; shorten it.
                                path.segments.drain(..path.segments.len() - 1);
```

ahomescu:
Most recent offline discussion: @kkysen and I agreed that it makes sense to split this off into a separate transform, which would also fix some bugs with scoping and aliases.

ahomescu:
I moved this to a separate transform.

ahomescu:
@kkysen @fw-immunant This is now a separate transform, it passed all tests and is already approved. Any more feedback? I'll give it another day or two and then merge it.

c2rust-refactor/src/ast_manip/util.rs
```diff
@@ -172,6 +172,34 @@ pub fn use_idents(tree: &UseTree) -> Vec<Ident> {
...
+impl UseInfo {
+    /// Retrieve the list of Idents and their NodeIds defined by the given UseTree
+    pub fn from_use_tree(tree: &UseTree, id0: NodeId) -> Vec<Self> {
+        match &tree.kind {
+            UseTreeKind::Simple(_, id1, id2) => vec![Self {
+                ident: tree.ident(),
+                ids: vec![id0, *id1, *id2],
+            }],
+            // TODO: support globs but the AST doesn't tell us
+            // which identifiers are imported right now
```
kkysen, line 193: Does `c2rust transpile` emit glob imports?  If it does, we could also probably remove that (I also don't like them, at least at global scope).

ahomescu: There is exactly one instance of `use_glob_item` in the repository: each top-level source for a binary imports all the symbols from the library crate using a glob, e.g., `use ::repo::*;` in libxml2's `runtest.rs`.

We could try to remove that, but I'd prefer doing it in a separate PR (tons of other stuff might break).

ahomescu: Addressed in https://github.com/immunant/c2rust/pull/1540.

ahomescu:
I don't see any more feedback, so I'll go ahead and merge this.

PR #1404, opened 2025-10-09
refactor: fix some `Cargo.toml` things

kkysen:
I noticed `c2rust-macros` was still `exclude`d even though it's used by `c2rust-refactor`, which is now no longer `exclude`d.  Then I noticed a couple other leftover things in `Cargo.toml`s.

### DIFF
--- a/c2rust-refactor/Cargo.toml
+++ b/c2rust-refactor/Cargo.toml
@@ -43,10 +43,6 @@ z3 = "0.12.0"
 quickcheck = "0.9.0"
 rand = "0.7"
 
-[lib]
-name = "c2rust_refactor"
-path = "src/lib.rs"
-
 [features]
 default = []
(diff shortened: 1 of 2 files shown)


c2rust-refactor/Cargo.toml
```diff
@@ -43,10 +43,6 @@ z3 = "0.12.0"
 quickcheck = "0.9.0"
 rand = "0.7"
 
-[lib]
```
ahomescu, line 46: Is this just the default now, or why is it safe to remove?

kkysen: Yeah, as long as I've used Rust, this has never been necessary.  If `src/lib.rs` is present, it's the `lib` by default.  Same with `src/main.rs` and a binary.

PR #1405, opened 2025-10-09
Refactor visit node to reduce rightward drift

bungcip:
`visit_node` in `conversion.rs` is very big, this PR try to split the function and use early return to reduce rightward drift

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -962,1167 +962,1103 @@ impl ConversionContext {
         use self::node_types::*;
 
         if expected_ty & TYPE != 0 {
-            self.convert_node_as_type(untyped_context, node_id, new_id, expected_ty);
-        } else {
-            // Convert the node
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -583,8 +583,7 @@ impl ConversionContext {
         })
     }
 
-    /// Visit one node.
-    fn visit_node(
+    fn convert_node_as_type(
```
bungcip, line 586: bikeshedding welcome for the function name

kkysen: `fn visit_type_node` perhaps?  Otherwise this LGTM.

bungcip: `visit_node_type` already exist, so naming it `visit_type_node` will make it more confusing😅

kkysen: Hmm, `node_as_type` is okay then, but I think `visit` should be kept.

Issue #1379, opened 2025-09-26
String literal macro translation results in potentially dangling pointer


Rua:
When a string literal `#define` is translated, it becomes a Rust `const` array of `c_char`:

```rust
pub const LITERAL_STR: [std::ffi::c_char; 6] =
    unsafe { ::core::mem::transmute::<[u8; 6], [std::ffi::c_char; 6]>(*b"hello\0") };
```

A Rust `const`, in principle, behaves like a temporary, so you can't take its address under normal circumstances. Because of lifetime extension, in places where the `const` then has a reference taken, Rust will extend the lifetime of the reference as long as needed. But because references are generally cast straight to pointers in c2rust generated code, Rust cannot know how long the lifetime of the reference needs to be. This results in a potentially dangling pointer:

```rust
let literal_str_ptr: *const std::ffi::c_char = LITERAL_STR.as_ptr();
```

Asking in #dark-arts on the Rust Community Discord, they suggested using a `const` reference instead:
```rust
pub const LITERAL_STR: &[std::ffi::c_char; 6] =
    unsafe { &::core::mem::transmute::<[u8; 6], [std::ffi::c_char; 6]>(*b"hello\0") };
```

Because `const` references are guaranteed to point to static data, this ensures that pointers can never dangle. Alternatively, translating these as `static` instead of `const` would also fix the issue.

This seems to be essentially the same bug as #1217, but manifesting itself in the context of string literal macros rather than compound literals. The problem is the same either way: the temporary may get dropped while still being pointed to.

kkysen:
Thanks for uncovering this!

> Asking in #dark-arts on the Rust Community Discord, they suggested using a `const` reference instead:
>
> ```rust
> pub const LITERAL_STR: &[std::ffi::c_char; 6] =
>     unsafe { &::core::mem::transmute::<[u8; 6], [std::ffi::c_char; 6]>(*b"hello\0") };
> ```
>
> Because `const` references are guaranteed to point to static data, this ensures that pointers can never dangle.

This seems like a good approach to me.  I would think we could do this everywhere, and since it's a `const`, if a use site needs the array (not a reference), that should work out, too.

> Alternatively, translating these as `static` instead of `const` would also fix the issue.

I think `const` is still the right approach here, since it's a closer translation to a C macro, and C can use a `static` itself if it wants that behavior.

[...]

PR #1407, opened 2025-10-11
transpile: Extend lifetime of compound literals whose address is taken

### LINKED ISSUE #1217: Incorrect Value Read from Struct Initialized with `PTHREAD_RWLOCK_INITIALIZER`

Rua:
- Fixes #1217.

I've only looked at fixing it in function context, as it's not possible to add statements to static initializers ("Expected no side-effects in static initializer" panic). But I'm not sure if the issue exists for statics anyway, because Rust is forced to extend the lifetime of the temporary to static in such a context.

There is also an issue with how this line compiles in the snapshot test:
```c
char *static_char_ptr = (char[]) { "hello" };
```

It gets translated into an array in which the first element is the translated string, and then five zero integer literals. That's clearly broken and doesn't compile in Rust, so I commented out that line. Since my fix doesn't change how statics are handled, I believe that's an existing bug in c2rust that needs its own fix.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3559,7 +3559,9 @@ impl<'c> Translation<'c> {
                     CastKind::BitCast | CastKind::PointerToIntegral | CastKind::NoOp => {
                         ctx.decay_ref = DecayRef::Yes
                     }
-                    CastKind::FunctionToPointerDecay | CastKind::BuiltinFnToFnPtr => {
+                    CastKind::ArrayToPointerDecay
+                    | CastKind::FunctionToPointerDecay
+                    | CastKind::BuiltinFnToFnPtr => {
                         ctx.needs_address = true;
                     }
                     _ => {}
@@ -4067,7 +4069,41 @@ impl<'c> Translation<'c> {
 
             Paren(_, val) => self.convert_expr(ctx, val, override_ty),
 
-            CompoundLiteral(_, val) => self.convert_expr(ctx, val, override_ty),
+            CompoundLiteral(qty, val) => {
+                let val = self.convert_expr(ctx, val, override_ty)?;
+
+                if !ctx.needs_address() || ctx.is_static || ctx.is_const {
+                    // Statics and consts have their intermediates' lifetimes extended.
+                    return Ok(val);
+                }
+
+                // C compound literals are lvalues, but equivalent Rust expressions generally are not.
+                // So if an address is needed, store it in an intermediate variable first.
+                let fresh_name = self.renamer.borrow_mut().fresh();
+                let fresh_ty = self.convert_type(override_ty.unwrap_or(qty).ctype)?;
+
+                val.and_then(|val| {
+                    let fresh_stmt = {
+                        let mutbl = if qty.qualifiers.is_const {
+                            Mutability::Immutable
+                        } else {
+                            Mutability::Mutable
+                        };
+
+                        let local = mk().local(
+                            mk().set_mutbl(mutbl).ident_pat(&fresh_name),
+                            Some(fresh_ty),
+                            Some(val),
+                        );
+                        mk().local_stmt(Box::new(local))
+                    };
+
+                    Ok(WithStmts::new(
+                        vec![fresh_stmt],
+                        mk().ident_expr(fresh_name),
+                    ))
+                })
+            }
 
             InitList(ty, ref ids, opt_union_field_id, _) => {
                 self.convert_init_list(ctx, ty, ids, opt_union_field_id)
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@compound_literals.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@compound_literals.c.snap
@@ -29,17 +29,22 @@ pub const CHAR_ARRAY: [::core::ffi::c_char; 6] =
 #[no_mangle]
 pub unsafe extern "C" fn local_compound_literals() {
     let mut single_int: ::core::ffi::c_int = 42 as ::core::ffi::c_int;
-    let mut single_int_ptr: *mut ::core::ffi::c_int =
-        &mut (42 as ::core::ffi::c_int) as *mut ::core::ffi::c_int;
-    let mut int_ptr_to_array: *mut ::core::ffi::c_int =
-        [42 as ::core::ffi::c_int, 9001 as ::core::ffi::c_int].as_mut_ptr();
-    let mut char_ptr_to_array: *mut ::core::ffi::c_char =
-        (::core::mem::transmute::<[u8; 6], [::core::ffi::c_char; 6]>(*b"hello\0")).as_mut_ptr();
+    let mut fresh0: ::core::ffi::c_int = 42 as ::core::ffi::c_int;
+    let mut single_int_ptr: *mut ::core::ffi::c_int = &mut fresh0 as *mut ::core::ffi::c_int;
+    let mut fresh1: [::core::ffi::c_int; 2] =
+        [42 as ::core::ffi::c_int, 9001 as ::core::ffi::c_int];
+    let mut int_ptr_to_array: *mut ::core::ffi::c_int = fresh1.as_mut_ptr();
+    let mut fresh2: [::core::ffi::c_char; 6] =
+        ::core::mem::transmute::<[u8; 6], [::core::ffi::c_char; 6]>(*b"hello\0");
+    let mut char_ptr_to_array: *mut ::core::ffi::c_char = fresh2.as_mut_ptr();
+    let mut fresh3: [::core::ffi::c_int; 2] =
+        [42 as ::core::ffi::c_int, 9001 as ::core::ffi::c_int];
     let mut int_array_ptr: *mut [::core::ffi::c_int; 2] =
-        &mut [42 as ::core::ffi::c_int, 9001 as ::core::ffi::c_int] as *mut [::core::ffi::c_int; 2];
+        &mut fresh3 as *mut [::core::ffi::c_int; 2];
+    let mut fresh4: [::core::ffi::c_char; 6] =
+        ::core::mem::transmute::<[u8; 6], [::core::ffi::c_char; 6]>(*b"hello\0");
     let mut char_array_ptr: *mut [::core::ffi::c_char; 6] =
-        &mut ::core::mem::transmute::<[u8; 6], [::core::ffi::c_char; 6]>(*b"hello\0")
-            as *mut [::core::ffi::c_char; 6];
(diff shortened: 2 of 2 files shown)


c2rust-transpile/tests/snapshots/compound_literals.c
```diff
@@ -0,0 +1,11 @@
+int static_single_int = (int) { 42 };
+int *static_single_int_ptr = &((int) { 42 });
+int *static_int_array_ptr = (int[]) { 42, 9001 };
+//char *static_char_ptr = (char[]) { "hello" };
```
kkysen: Can you add some of this explanation inline here?

> There is also an issue with how this line compiles in the snapshot test:
> ```c
> char *static_char_ptr = (char[]) { "hello" };
> ```
> 
> It gets translated into an array in which the first element is the translated string, and then five zero integer literals. That's clearly broken and doesn't compile in Rust, so I commented out that line. Since my fix doesn't change how statics are handled, I believe that's an existing bug in c2rust that needs its own fix.

c2rust-transpile/tests/snapshots/compound_literals.c
kkysen, line 1: Can you add this snapshot in a separate commit so that there's a commit with its diff from the fixes you made?  Thanks!

kkysen: I did this in 14713ae.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4070,7 +4072,41 @@ impl<'c> Translation<'c> {
 
             Paren(_, val) => self.convert_expr(ctx, val, override_ty),
 
-            CompoundLiteral(_, val) => self.convert_expr(ctx, val, override_ty),
+            CompoundLiteral(qty, val) => {
+                let val = self.convert_expr(ctx, val, override_ty)?;
+
+                if !ctx.needs_address() || ctx.is_static {
```
```suggestion
                if !ctx.needs_address() || ctx.is_static {
                    // Statics have their intermediates' lifetimes extended.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4070,7 +4072,41 @@ impl<'c> Translation<'c> {
...
+                    return Ok(val);
+                }
+
+                // C compound literals are lvalues, but equivalent Rust expressions generally
+                // are not. So if an address is needed, store it in an intermediate variable
+                // first.
```
kkysen: nit:

```suggestion
                // C compound literals are lvalues, but equivalent Rust expressions generally are not.
                // So if an address is needed, store it in an intermediate variable first.
```

Rua:
I added extra tests to ensure that compound literals in macros don't undergo the same treatment as normal. Because they can be borrowed as statics, same as in #1379. However, I notice that the macro isn't used in the `macro_int_array_ptr` and `macro_char_array_ptr` cases, it's expanded to its definition and then given a fresh variable. Any idea why the macro is bypassed here?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4070,7 +4072,41 @@ impl<'c> Translation<'c> {
 
             Paren(_, val) => self.convert_expr(ctx, val, override_ty),
 
-            CompoundLiteral(_, val) => self.convert_expr(ctx, val, override_ty),
+            CompoundLiteral(qty, val) => {
+                let val = self.convert_expr(ctx, val, override_ty)?;
+
+                if !ctx.needs_address() || ctx.is_static {
+                    // Statics have their intermediates' lifetimes extended.
```
```suggestion
                if !ctx.needs_address() || ctx.is_static || ctx.is_const {
                    // Statics and consts have their intermediates' lifetimes extended.
```

c2rust-transpile/tests/snapshots/compound_literals.c
```diff
@@ -0,0 +1,22 @@
+int static_single_int = (int) { 42 };
+int *static_single_int_ptr = &((int) { 42 });
+int *static_int_array_ptr = (int[]) { 42, 9001 };
+// Currently generates broken Rust code, see
+// https://github.com/immunant/c2rust/issues/1410
+//char *static_char_array_ptr = (char[]) { "hello" };
+
+#define SINGLE_INT ((int) { 42 })
```
kkysen, line 11: Can you move the macros to the test commit? Or a new commit in between.

kkysen: I think you added the fresh changes to the original test commit, too.

Rua:
This should be ready to go now.

kkysen (approved):
One of the snapshot lines was off in the first commit, but I fixed it, so this LGTM now.  Thanks again!

PR #1408, opened 2025-10-11
transpile: Always use absolute paths when referring to crates

### LINKED ISSUE #447: Qualified paths like `libc::c_int` should be globally qualified like `::libc::c_int` to avoid name clashes

Rua:
- Fixes #447.

### DIFF
--- a/c2rust-transpile/src/rust_ast/item_store.rs
+++ b/c2rust-transpile/src/rust_ast/item_store.rs
@@ -36,34 +36,47 @@ impl MultiImport {
 }
 
 #[derive(Debug, Default)]
-pub struct PathedMultiImports(IndexMap<Vec<String>, MultiImport>);
+pub struct PathedMultiImports(IndexMap<(bool, Vec<String>), MultiImport>);
 
 impl PathedMultiImports {
     pub fn new() -> Self {
         Self::default()
     }
 
-    pub fn get_mut(&mut self, path: Vec<String>) -> &mut MultiImport {
-        self.0.entry(path).or_insert(MultiImport::new())
+    pub fn get_mut(&mut self, is_absolute: bool, path: Vec<String>) -> &mut MultiImport {
+        self.0
+            .entry((is_absolute, path))
+            .or_insert(MultiImport::new())
     }
 
     pub fn into_items(self) -> Vec<Box<Item>> {
-        fn build_items((mut path, imports): (Vec<String>, MultiImport)) -> Box<Item> {
-            let mut leaves = imports.leaves;
-            let attrs = imports.attrs.unwrap_or_else(mk);
+        self.0
+            .into_iter()
+            .map(|((is_absolute, mut path), imports)| {
+                let mut leaves = imports.leaves;
+                let attrs = imports.attrs.unwrap_or_else(mk);
 
-            if leaves.len() == 1 {
-                path.push(leaves.pop().unwrap());
+                if leaves.len() == 1 {
+                    path.push(leaves.pop().unwrap());
 
-                let path = mk().abs_path(path);
-                attrs.use_simple_item(path, None as Option<Ident>)
-            } else {
-                let path = mk().abs_path(path);
-                attrs.use_multiple_item(path, leaves.into_iter())
-            }
-        }
+                    let path = if is_absolute {
+                        mk().abs_path(path)
+                    } else {
+                        mk().path(path)
+                    };
 
-        self.0.into_iter().map(build_items).collect()
+                    attrs.use_simple_item(path, None as Option<Ident>)
+                } else {
+                    let path = if is_absolute {
+                        mk().abs_path(path)
+                    } else {
+                        mk().path(path)
+                    };
+
+                    attrs.use_multiple_item(path, leaves.into_iter())
+                }
+            })
+            .collect()
     }
 }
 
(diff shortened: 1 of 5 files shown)


c2rust-transpile/src/convert_type.rs
```diff
@@ -382,22 +381,22 @@ impl TypeConverter {
                 let new_name = self
                     .resolve_decl_name(decl_id)
                     .ok_or_else(|| format_err!("Unknown decl id {:?}", decl_id))?;
-                Ok(mk().path_ty(mk().path(vec![new_name])))
+                Ok(mk().path_ty(vec![new_name]))
```
kkysen, line 384: Stuff like this is a separate, non-functional refactoring, right? Could you put it in a separate commit if so?

c2rust-transpile/tests/snapshots/snapshots__transpile@alloca.c.snap
```diff
@@ -12,29 +12,29 @@ input_file: c2rust-transpile/tests/snapshots/alloca.c
     unused_mut
 )]
 #[no_mangle]
-pub static mut TRUE: core::ffi::c_int = 1 as core::ffi::c_int;
+pub static mut TRUE: ::core::ffi::c_int = 1 as ::core::ffi::c_int;
```
kkysen, line 15: @ahomescu, is #1399 able to handle stuff like this and still shorten the paths with imports?  Also, is it or another refactor able to remove the `::` where not needed?  Would that be difficult at all? 

The change here is correct and handles some edge cases, but it is longer, so it'd be nice to be able to refactor out the unnecessary ones afterward.

Rua:
> I think the CI failings are because you didn't update the os- and arch-specific snapshots that aren't run locally.

The three snapshot CI runs are failing because the leading double colon `::` is expected by the snapshot (because I added it) but, for some reason, not actually generated by the transpiler. And this has me rather puzzled at the moment. For the Clang 18 test, `snapshots__transpile-x86_64@vm_x86.c.snap` gives this diff:

```
    6     6 │     unused_assignments,
    7     7 │     unused_mut
    8     8 │ )]
    9     9 │ #![feature(asm)]
   10       │-use ::core::arch::asm;
         10 │+use core::arch::asm;
   11    11 │ #[derive(Copy, Clone)]
   12    12 │ #[repr(C)]
   13    13 │ pub struct vm_t {
   14    14 │     pub programStack: ::core::ffi::c_int,
```

I can find only one place in the whole codebase that adds an import with the string "asm":
https://github.com/immunant/c2rust/blob/8d87265ed7cd5d0416702d7fee1659ce63e014e6/c2rust-transpile/src/translator/assembly.rs#L1062

All items added to the item store are eventually fed through `PathedMultiImports::into_items`, where I have added lines to make absolute paths:
https://github.com/immunant/c2rust/blob/8d87265ed7cd5d0416702d7fee1659ce63e014e6/c2rust-transpile/src/rust_ast/item_store.rs#L55-L64

And yet, the generated code does not have an absolute path. It's as if `abs_path` there is having no effect.

kkysen:
Ah, I figured out what it is.  It's `rustfmt`.  It's deleting the `::`.

Rua:
It seems to be fixed, but it is now getting stuck on the empty `Ident` issue fixed by #1411.

kkysen:
Ugh, why did this get closed?

fw-immunant:
> Ugh, why did this get closed?

GH seems to suggest it was closed because the target branch was deleted.

kkysen:
> > Ugh, why did this get closed?
> 
> GH seems to suggest it was closed because the target branch was deleted.

Ugh, GitHub being dumb again.  Normally when you delete the target branch, it changes it to `master` (or whatever the target branch of the PR that was closed is), but apparently when it's an external PR, it just closes the PR.

Rua:
Hmm strange, the CI is still failing on the same error. Is it perhaps using an old version?

EDIT: But the MacOS test passed... huh?

kkysen:
> Hmm strange, the CI is still failing on the same error. Is it perhaps using an old version?
> 
> EDIT: But the MacOS test passed... huh?

I'm not sure, but the error message is slightly different. Might be a different error?

Rua:
Transpile is able to generate a new Cargo.toml when transpiling, and it will fill it with dependencies used by the transpiled code, including `c2rust-bitfields`. Is it possible that the file it generates there refers to the stable/crates.io version of `c2rust-bitfields`, and not the one in the current directory tree? If so, that could explain what's going on maybe.

kkysen (commented):
> Transpile is able to generate a new Cargo.toml when transpiling, and it will fill it with dependencies used by the transpiled code, including `c2rust-bitfields`. Is it possible that the file it generates there refers to the stable/crates.io version of `c2rust-bitfields`, and not the one in the current directory tree? If so, that could explain what's going on maybe.

I think you're right.  And it seems to be using `c2rust-bitfields= "0.3"` from what I can tell, which is very old.

kkysen (commented):
@Rua, this should be fixed once #1420 merges (and the stack of PRs in front of it as well).

Rua:
This is now passing the tests, but the Clang 15 test seems to be failing for an unrelated reason.

EDIT: Seems to be fixed again.

c2rust-transpile/tests/snapshots/snapshots__transpile@alloca.c.snap
ahomescu: Maybe, I can give it a try. I'm not sure you can run `reorganize_definitions` multiple times on the same code base, but we can find out.

PR #1411, opened 2025-10-12
bitfields-derive: strip `::` prefix from type if present

Rua:
The code was testing whether the initial `::` was present, but was not taking it out of the string/iterator. That would then lead to a panic when creating the `Ident` on an empty string.

### DIFF
--- a/c2rust-bitfields-derive/src/lib.rs
+++ b/c2rust-bitfields-derive/src/lib.rs
@@ -114,9 +114,8 @@ fn filter_and_parse_fields(field: &Field) -> Vec<Result<BFFieldAttr, Error>> {
 fn parse_bitfield_ty_path(field: &BFFieldAttr) -> Path {
     let mut segments = Punctuated::new();
     let mut segment_strings = field.ty.split("::").peekable();
-    let leading_colon = segment_strings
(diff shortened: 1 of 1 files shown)


Rua:
When you call `split` on `::core::ffi::c_int` you get `["", "core", "ffi", "c_int"]`. Further down in that function, it then creates an `Ident` out of each piece.

Where would I place tests for bitfields-derive?

c2rust-bitfields-derive/src/lib.rs
```diff
@@ -111,29 +112,19 @@ fn filter_and_parse_fields(field: &Field) -> Vec<Result<BFFieldAttr, Error>> {
...
-            Span::call_site().into(),
-        ]))
-    } else {
-        None
-    };
-
     let mut segments = Punctuated::new();
     let mut segment_strings = field.ty.split("::").peekable();
+    let colon = Token![::]([Span::call_site(), Span::call_site()]);
+    let leading_colon = segment_strings.next_if_eq(&"").map(|_| colon);
```
fw-immunant, line 118: Never knew about `Peekable::next_if_eq`!

Rua:
Looks good yes!

PR #1414, opened 2025-10-14
Re-enable `--reorganize-definitions` flag

fw-immunant:
It's currently a no-op. We'll also want to make it run by default, which is where it seems more likely we'll run into CI/release trouble due to our complex rustc version requirements.

### DIFF
--- a/c2rust/src/bin/c2rust-transpile.rs
+++ b/c2rust/src/bin/c2rust-transpile.rs
@@ -124,9 +124,17 @@ struct Args {
     #[clap(long)]
     reduce_type_annotations: bool,
 
-    /// Output file in such a way that the refactoring tool can deduplicate code
-    #[clap(short = 'r', long)]
-    reorganize_definitions: bool,
+    /// Disable `--reorganize-definitions`
+    #[clap(long = "no-reorganize-definitions", hidden = true, action = clap::ArgAction::SetFalse)]
+    reorganize_definitions: bool, // NB, this *is* the right field; this flag sets it to *false*.
+
+    /// Output file in such a way that the refactoring tool can deduplicate code (enabled by default; disable with `--no-reorganize-definitions`)
+    #[clap(
+        short = 'r',
+        long = "reorganize-definitions",
+        conflicts_with = "reorganize-definitions"
+    )]
+    _no_reorganize_definitions: bool, // Field is unused, but flag mutually excludes the negative sense above.
 
     /// Extra arguments to pass to clang frontend during parsing the input C file
     #[clap(multiple = true, last(true))]


c2rust-transpile/src/lib.rs
```diff
@@ -460,8 +460,43 @@ fn get_extra_args_macos() -> Vec<String> {
...
+    let cmd_path = std::env::current_exe().expect("Cannot get current executable path");
+    let mut cmd_path = cmd_path.as_path().canonicalize().unwrap();
+    cmd_path.pop(); // remove current executable
+    cmd_path.push(format!("c2rust-refactor"));
+    assert!(cmd_path.exists(), "{:?} is missing", cmd_path);
+    let args = [
+        "--cargo",
+        "--rewrite-mode",
+        "inplace",
+        "rename_unnamed",
```
ahomescu, line 488: I had completely forgotten we had this, it probably would have fixed the issue from the libxml2 demo.

c2rust/Cargo.toml
```diff
@@ -21,10 +21,12 @@ log = "0.4"
 regex = "1.3"
 shlex = "1.3"
 c2rust-transpile = { version = "0.21.0", path = "../c2rust-transpile" }
+c2rust-refactor = { version = "0.21.0", path = "../c2rust-refactor" }
```
ahomescu, line 24: Does this build? I think when I tried it a few months ago there were some crate version conflicts between c2rust-refactor and the transpiler.

fw-immunant:
Fixed a couple issues where the tests were overly brittle; now in `cargo test` only `tests/simd.x86_64` seems to fail; many errors around not-found types, such as:
```
error[E0412]: cannot find type `__m128` in this scope
```

fw-immunant:
The problem seems to be that imports of needed types for SIMD get placed into submodules and then these types are not in scope for their later usage:
```C
#include <immintrin.h>

void simd_fn_codegen(__m128i i)
{
    i = _mm_aesdeclast_si128(i, i);
}
```

transpiles with `--reorganize-definitions` to:

```Rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
#![register_tool(c2rust)]
#![feature(register_tool, stdsimd)]
#[c2rust::header_src = "/usr/lib/clang/21/include/mmintrin.h:1"]
pub mod mmintrin_h {
    #[cfg(target_arch = "x86")]
    pub use ::core::arch::x86::__m128i;
    #[cfg(target_arch = "x86_64")]
    pub use ::core::arch::x86_64::__m128i;
}
#[c2rust::header_src = "/usr/lib/clang/21/include/__wmmintrin_aes.h:1"]
pub mod __wmmintrin_aes_h {
    #[cfg(target_arch = "x86")]
    pub use ::core::arch::x86::_mm_aesdeclast_si128;
    #[cfg(target_arch = "x86_64")]
    pub use ::core::arch::x86_64::_mm_aesdeclast_si128;
}
#[cfg(target_arch = "x86")]
pub use ::core::arch::x86::_mm_setzero_si128;
#[cfg(target_arch = "x86_64")]
pub use ::core::arch::x86_64::_mm_setzero_si128;
#[no_mangle]
#[c2rust::src_loc = "3:1"]
pub unsafe extern "C" fn rust_simd_fn_codegen(mut i: __m128i) {
    i = _mm_aesdeclast_si128(i, i);
}
```

fw-immunant:
Ah, there's another problem here: if `-e/--emit-build-files` is not passed (e.g. as when invoked from `test_translator.py`), the transpiler skips invoking the refactorer entirely, but does still change its output to include `mod`s for headers and the various `c2rust::src_loc`/`c2rust::header_src` attributes. In this case the output does not compile at all, so even passing `-e` just results in the transpiler balking with:
```
warning: Reorganizing definitions failed: Crate does not compile.
```

So it seems we need to fix this in the transpiler, not the refactorer, and may also need to alter how the transpiler is invoked by `test_translator.py` if we want those tests to actually benefit from the refactorer.

fw-immunant:
The next set of errors in CI are the testsuite failing due to `__sigset_t` and `C2RustUnnamed` types not being imported for some reason.

fw-immunant:
Now I'm seeing `c2rust-refactor: error while loading shared libraries: librustc_driver-14a53782e7072fd8.so: cannot open shared object file: No such file or directory`. Presumably a library path issue of some kind.

c2rust/Cargo.toml
kkysen: This will break `cargo install c2rust` on stable, so I don't think we want this.

fw-immunant: So how do we want to handle this? Should we support building both with or without c2rust-refactor as a dependency (e.g. via a cargo feature and `cfg`)? Cc @thedataking

kkysen:
Split into and superseded by #1451 and #1452.

c2rust/Cargo.toml
kkysen: I don't think having it as a dependency adds very much.  Is what I did in #1451 good for this?

fw-immunant: Yes, I hadn't yet gone through your follow-ups when I posted this. :+1:

PR #1416, opened 2025-10-17
exporter: print src loc for all warnings, including in type encoding

kkysen:


fw-immunant (approved):
Code changes look fine, but I don't fully understand the motivation. Is this a pure refactor, or fixing some particular place where our output was useless/incorrect before?

kkysen (commented):
> Is this a pure refactor, or fixing some particular place where our output was useless/incorrect before?

I was getting errors, like the "Encountered unsupported BuiltinType kind" one, where there was no source location.  `clang::FullSourceLoc()` was being used, but that's an invalid location.  So this threads through the right source location for that and any other errors that occur in the `TypeEncoder`.

PR #1418, opened 2025-10-19
transpile: update dependencies in generated `Cargo.toml`

kkysen:
These have been missed for a long time, so we've been using 6-year-old versions this whole time.

### DIFF
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -182,8 +182,12 @@ impl ExternCrateDetails {
 impl From<ExternCrate> for ExternCrateDetails {
     fn from(extern_crate: ExternCrate) -> Self {
         match extern_crate {
-            ExternCrate::C2RustBitfields => Self::new("c2rust-bitfields", "0.21", true),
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/lib.rs
```diff
@@ -182,8 +182,8 @@ impl ExternCrateDetails {
 impl From<ExternCrate> for ExternCrateDetails {
     fn from(extern_crate: ExternCrate) -> Self {
         match extern_crate {
-            ExternCrate::C2RustBitfields => Self::new("c2rust-bitfields", "0.3", true),
-            ExternCrate::C2RustAsmCasts => Self::new("c2rust-asm-casts", "0.2", true),
+            ExternCrate::C2RustBitfields => Self::new("c2rust-bitfields", "0.21", true),
```
ahomescu: Can we get the version of the transpiler here, and make sure it always matches?

kkysen: Oh sure, I think we can do that, too.  I'll add a commit.

kkysen: Done in 15ea0e9.

PR #1419, opened 2025-10-19
transpile: remove `~` in `Cargo.toml.hbs` causing trailing whitespace to be removed

kkysen:
I think this was probably a typo in the `Cargo.toml.hbs` template.  The `~` means strip trailing whitespace, causing us to emit things like `c2rust-bitfields= "0.3"` instead of `c2rust-bitfields = "0.3"`, which is what `c2rust-testsuite` matches on:

https://github.com/immunant/c2rust-testsuite/blob/5b5a098a39d5431383f46410b838e24321a4cb25/tests/templates.py#L31-L32
```sh
    sed --in-place --regexp-extended "s|c2rust-bitfields = \"([0-9.]+)\"|c2rust-bitfields = { version = \"\1\", path = \"$C2RUST_DIR/c2rust-bitfields\" }|" "$SCRIPT_DIR/repo/Cargo.toml"
    sed --in-place --regexp-extended "s|c2rust-asm-casts = \"([0-9.]+)\"|c2rust-asm-casts = { version = \"\1\", path = \"$C2RUST_DIR/c2rust-asm-casts\" }|" "$SCRIPT_DIR/repo/Cargo.toml"
```

PR #1421, opened 2025-10-20
refactor: update to `cargo v0.65`, matching our pinned nightly of Rust 1.65

kkysen:
Our workspace `Cargo.toml` uses workspace inheritance, which was stabilized in Rust 1.64, so is fine with our pinned nightly of Rust 1.65. But `c2rust-refactor` was using `cargo v0.62` (corresponding to Rust 1.62), so when `c2rust-refactor` was run on a `Cargo.toml` that referenced `c2rust-bitfields` with a relative path to our repo, it broke on this (which #1419 and #1420 fix/enable).

Thus, we need to update to at least `cargo v0.64`, and preferably `v0.65`. `cargo v0.65` depends on `toml_edit v0.14.3`, which depends on `kstring v2.0.0`. By default, `kstring v2.0.1` is selected (since Rust 1.65 doesn't support the MSRV-aware resolver), but it has an MSRV of Rust 1.73, too new.
Luckily, though, `kstring v2.0.0` has an MSRV of Rust 1.59, so we pin it.

### DIFF
--- a/c2rust-refactor/src/lib.rs
+++ b/c2rust-refactor/src/lib.rs
@@ -283,7 +283,6 @@ fn get_rustc_cargo_args(target_type: CargoTarget) -> Vec<RustcArgs> {
 
             let args = cmd
                 .get_args()
-                .iter()
                 .map(|os| os.to_str().unwrap().to_owned())
                 .collect();
             let mut g = self.pkg_args.lock().unwrap();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -454,6 +454,7 @@ dependencies = [
  "flamer",
  "indexmap",
  "json",
+ "kstring",
  "libc",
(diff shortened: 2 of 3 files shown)


ahomescu (approved):

c2rust-refactor/src/lib.rs
```diff
@@ -283,7 +283,6 @@ fn get_rustc_cargo_args(target_type: CargoTarget) -> Vec<RustcArgs> {
 
             let args = cmd
                 .get_args()
-                .iter()
```
fw-immunant, line 286: Is this a drive-by cleanup or actually required by this version change?

kkysen: Required by the cargo update.

PR #1424, opened 2025-10-21
include refactor step in CI

oinoom:


fw-immunant:
The testsuite ref specified here is broken, you should use current master (https://github.com/immunant/c2rust-testsuite/commit/5b5a098a39d5431383f46410b838e24321a4cb25).

PR #1427, opened 2025-10-22
refactor: help the user find a transform

fw-immunant:


c2rust-refactor/src/command.rs
```diff
@@ -572,7 +572,14 @@ impl RefactorState {
                 s
             }));
 
-        let mut cmd = self.cmd_reg.get_command(cmd_name, &args)?;
+        let mut cmd = self.cmd_reg.get_command(cmd_name, &args).map_err(|mut e| {
```
kkysen, line 575: Clap doesn't do this already?  Through `--help` or something.

fw-immunant: Clap doesn't know about the transforms; there isn't an enum or such, but they're registered in the `Registry::commands` `HashMap`.

kkysen: Ah. I'd rather add them to clap then.

fw-immunant: I mostly did this because I didn't know which transforms existed, so I don't intend to spend more time on it. But I think it does help users even as-is.

kkysen: Okay.  Can you leave an issue open to fix it properly, though?

kkysen (approved):
This is fine as a simple intermediate improvement, but let's leave an issue open to fix this properly (integrated with `clap` and `--help`).

PR #1429, opened 2025-10-22
pdg: check in `Cargo.lock`s for analysis tests

kkysen:
We had a blanket rule to ignore all but the top-level `Cargo.lock`. I think this was intended for `tests/`, but `tests/.gitignore` already does this, too. These `Cargo.lock`s are needed, or else our pinned nightly (Rust 1.65) `cargo` will select `syn v2.0`, which can't be built on Rust 1.65. This is needed by `c2rust-instrument`, so upgrading the generated nightly here isn't feasible. All of `c2rust-instrument` would have to be upgraded, which is more involved.

Furthermore, having the `Cargo.lock` should avoid `cargo` having to fetch the whole registry every time, which might help speed up CI.

There's another similar `syn v2.0` error in `test_translator.py`.  I'll fix that separately, but `cargo test` and the pdg tests should be good now.


fw-immunant (approved):
Fixes the `cargo test` phase of CI and seems reasonable. Thanks for digging through all our complex versioning/dependency setup, it really seems like a rat's nest.

PR #1430, opened 2025-10-23
tests: copy `generated-rust-toolchain.toml` to test directories

kkysen:
In `test_translator.py`, we `c2rust-transpile` `*.c` files individually, so `--emit-build-files` doesn't work as intended, since it'd be used per-file and emit in the wrong directory (in `src/`, not parallel to the `Cargo.toml`).  We could change things to transpile all `*.c` files at once, but that'd be a more involved change.  So we just link `generated-rust-toolchain.toml` to `rust-toolchain.toml`.

This should finally fully fix the CI errors due `c2rust-bitfields-derive` resolving `syn v2.0.107`, which can't build on our pinned nightly.

### DIFF
--- a/scripts/test_translator.py
+++ b/scripts/test_translator.py
@@ -497,7 +497,7 @@ class TestDirectory:
         # We `c2rust-transpile` `*.c` files individually, so `--emit-build-files` doesn't work
         # (if it's generated, it's in the wrong directory and may be different for each transpiled file).
         # We could also change things to transpile all `*.c` files at once, but that's more involved.
-        generated_rust_toolchain = Path(__file__).parent / "../c2rust-transpile/src/build_files/generated-rust-toolchain.toml"
+        generated_rust_toolchain = Path(c.TRANSPILE_CRATE_DIR) / "src/build_files/generated-rust-toolchain.toml"
         rust_toolchain = Path(self.full_path) / "rust-toolchain.toml"
         rust_toolchain.unlink(missing_ok=True)
         rust_toolchain.symlink_to(generated_rust_toolchain)


kkysen (commented):
Yay it worked!

scripts/test_translator.py
```diff
@@ -493,6 +493,16 @@ def run(self) -> List[TestOutcome]:
 
         self.generated_files["rust_src"].append(lib_file)
 
+        # Copy `generated-rust-toolchain.toml`.
+        # We `c2rust-transpile` `*.c` files individually, so `--emit-build-files` doesn't work
+        # (if it's generated, it's in the wrong directory and may be different for each transpiled file).
+        # We could also change things to transpile all `*.c` files at once, but that's more involved.
+        generated_rust_toolchain = Path(__file__).parent / "../c2rust-transpile/src/build_files/generated-rust-toolchain.toml"
```
ahomescu: Isn't this a template? The current version doesn't look like it but I think it was supposed to be at some point.

kkysen: Right now we just copy it in `c2rust-transpile`, too:

```rust
fn emit_rust_toolchain(tcfg: &TranspilerConfig, build_dir: &Path) {
    let output_path = build_dir.join("rust-toolchain.toml");
    let output = include_str!("generated-rust-toolchain.toml").to_string();
    maybe_write_to_file(&output_path, output, tcfg.overwrite_existing);
}
```

fw-immunant: This should probably use `c.TRANSPILE_CRATE_DIR / "src/build_files/generated-rust-toolchain.toml"`.

scripts/test_translator.py
```diff
@@ -493,6 +493,16 @@ def run(self) -> List[TestOutcome]:
...
+        rust_toolchain = Path(self.full_path) / "rust-toolchain.toml"
+        rust_toolchain.unlink(missing_ok=True)
+        rust_toolchain.symlink_to(generated_rust_toolchain)
```
fw-immunant, line 503: Just a nit, this is racy in case something else creates the file between these two operations, but I guess pathlib doesn't have a "force" argument for `symlink_to`. I hate the filesystem.

scripts/test_translator.py
fw-immunant: @ahomescu I think you may be thinking of the `Cargo.toml`, which is a handlebars template. This file was added in 3a4e872893c28b93c179d9c1654045b6fcf32895, changing the previous behavior which was to copy `../../rust-toolchain.toml`.

fw-immunant (approved):
A couple comments inline. Only the `TRANSPILE_CRATE_DIR` one is blocking.

It would be nice to clearly document in the README what Rust version transpiled code targets, whether we state it concretely or point to the relevant `rust-toolchain.toml`.

scripts/test_translator.py
kkysen: I can link to a temporary and do an atomic rename, but I wasn't sure if that was worth it.  I don't think there's a way to create a symlink (or hard link) with replacing the file atomically.

fw-immunant: Probably not worth the trouble.

scripts/test_translator.py
kkysen: Fixed in 16ce5ee.

PR #1432, opened 2025-10-24
refactor: fix `remove_literal_suffixes` panics on transpiled C bitfields

oinoom:
The `remove_literal_suffixes` transform panicked on libxml2 when it hit key trees that weren't strictly one-child, producing assertion failures such as:

  - assertion failed: ch_kt.len() == 1
  - expected node, found: Leaf

As context, transform maintains a key-tree representation of types while it tries to unify literal types:

- Leaf -- scalar types (e.g., i32, u8)
- Node([...]) --aggregates or wrappers (structs, tuples, arrays, refs, function signatures)
- Empty -- unknown types or cases we intentionally skip

The code assumed wrapper nodes always exposed exactly one child and that aggregates had predictable shapes. Transpiled C bitfields break those assumptions in a few ways: 

 1. Setters/getters whose type collapses to a single Leaf (we used to panic because children() expected a node), and
 2. Tuple-typed helpers (e.g., (usize, usize) ranges) that hand unify_expr_child a Node([child0, child1]), tripping the "len == 1" assertion
 3. In libxml2's bitfield code, the generated setter calls (e.g., set_mon(...)) don't always have a fully resolved id, so type_dependent_def_id comes back None.

This PR alters the transform to tolerate these cases:

- LitTyKeyNode::children() now returns None for Leaf and Empty, avoiding the panic when inference collapses a node
- Spots that assumed exactly one child (unify_expr_child, array/slice/pattern handling, etc.) now only consume one child if it exists and otherwise fall back to best-effort inference instead of asserting
- Method-call handling guards tables.type_dependent_def_id(hir_id) and logs misses, which happen in generated bitfield setters/getters

Tested with `c2rust-refactor remove_literal_suffixes --cargo --lib` on libxml2, which ran successfully

### DIFF
--- a/c2rust-refactor/src/transform/literals.rs
+++ b/c2rust-refactor/src/transform/literals.rs
@@ -11,6 +11,7 @@ use rustc_span::symbol::Symbol;
 use rustc_ast::visit::{self, Visitor};
 use rustc_span::Span;
 use rustc_type_ir::sty;
+use log::debug;
 
 use std::cell::Cell;
 use std::collections::{HashMap, HashSet};
@@ -519,11 +520,12 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
         }
     }
 
+    /// Unify a child expression when the parent key tree wraps a single element.
     fn unify_expr_child(&mut self, e: &Expr, kt: LitTyKeyTree<'kt, 'tcx>) {
-        if let Some(ch_kt) = kt.get().children() {
-            assert!(ch_kt.len() == 1);
-            self.visit_expr_unify(e, ch_kt[0]);
+        if let Some(&[child_kt]) = kt.get().children() {
+            self.visit_expr_unify(e, child_kt);
         } else {
+            // Not a single-child wrapper: visit without unification
             self.visit_expr(e);
         }
     }
@@ -537,15 +539,15 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
                 // We really want the subexpressions to at least unify with
                 // each other, so we need to either get the key tree
                 // from the caller, or create a new common one for all subexpressions
-                let elem_kt = if let Some(ch) = kt.get().children() {
-                    assert!(ch.len() == 1);
-                    ch[0]
-                } else {
-                    exprs.first()
-                        .and_then(|e| self.cx.opt_node_type(e.id))
-                        .map(|ty| self.ty_to_key_tree(ty, false))
-                        .unwrap_or_else(|| self.new_empty_node())
-
+                let elem_kt = match kt.get().children() {
+                    Some(&[elem_kt]) => elem_kt,
+                    _ => {
+                        // Best effort: reuse the first element type when present.
+                        exprs.first()
+                            .and_then(|e| self.cx.opt_node_type(e.id))
+                            .map(|ty| self.ty_to_key_tree(ty, false))
+                            .unwrap_or_else(|| self.new_empty_node())
+                    }
                 };
                 for e in exprs {
                     self.visit_expr_unify(e, elem_kt);
@@ -556,7 +558,7 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
                 let callee_key_tree = self.expr_ty_to_key_tree(callee);
                 self.visit_expr_unify(callee, callee_key_tree);
                 // TODO: check if generic
-                if let Some(&[ref input_key_trees @ .., output_key_tree]) = callee_key_tree.get().children() {
+                if let LitTyKeyNode::Node(&[ref input_key_trees @ .., output_key_tree]) = callee_key_tree.get() {
                     for (arg_expr, arg_key_tree) in args.iter().zip(input_key_trees.iter()) {
                         self.visit_expr_unify(arg_expr, arg_key_tree);
                     }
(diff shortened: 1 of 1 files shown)


fw-immunant:
With respect to the PR description:
- Why would the types of setters/getters (which are methods) collapse to a `Leaf`, if function types are represented as `Node`? Is this because they don't have a "resolved id"?
- For that matter, why *don't* generated bitfield setters have a resolved id? Is this a phase ordering problem where that info simply isn't available yet, or is it a symptom of something being missing or broken earlier in the compilation/analysis pipeline?
- Is the core problem here that some types considered "wrappers" were expected to have only one child type, but now have more? Which types are these, and what changed?

oinoom:
Thanks @fw-immunant, I should have looked into why this was happening to begin with, and not just assumed an implementation mistake. It started with the child count assertion violation and I followed that path to fix all of the subsequent failures without seeing if it should have actually had a child count of 1 all along. I'll look into this more and get back to you

oinoom:
@fw-immunant @ahomescu it seems that all of this stems from the fact that c2rust-bitfield, macro-expanded code has distinct AST node ids, and this code walks that AST grabbing the node types with opt_node_type. opt_node_type calls self.hir_map().opt_node_to_hir_id(id) which looks up HIR by span. The issue with this is that multiple node ids have the same span, so that map we look up HIR types in will potentially point to the wrong type (the visitor that constructs the map will overwrite HIR with the same span). The result are mismatches such as field types instead of structs (Leafs instead of Nodes), tuples instead of references to tuples (two children vs. one child), and types of method call parameters instead of the method calls themselves (the reason for the "unresolved" id). I didn't write out the getters/setters point as well as I could have -- it's the expressions in the getters/setters that encounter this issue, the definitions themselves seem to not have a problem.

It seems like we can either try to make the mapping one to one somehow, or address it in a way similar to the way I've done in this PR. Maybe there is a better approach / third option?

fw-immunant:
> It seems like we can either try to make the mapping one to one somehow, or address it in a way similar to the way I've done in this PR. Maybe there is a better approach / third option?

It seems like we should either fix this mapping or substantially change how we interpret it; trying to be robust against broken invariants like the first pass of this PR did seems likely to lead to subtler bugs later on. Is it possible to augment the spans somehow to disambiguate which node we actually want?

ahomescu:
> I'll admit I don't fully understand this code.

I wrote this code, I'll take a look at the PR.

ahomescu:
> opt_node_type calls self.hir_map().opt_node_to_hir_id(id) which looks up HIR by span. The issue with this is that multiple node ids have the same span, so that map we look up HIR types in will potentially point to the wrong type (the visitor that constructs the map will overwrite HIR with the same span).

Right, rustc used to give us a very accurate mapping between AST and HIR IDs, but they made that internal so I had to replicate that in parallel in https://github.com/immunant/c2rust/pull/1364. There is a corner case in `SpanToHirMap` where the map returns a different ID depending on the context (path expression or some other kind, see `span_maps.rs`), you could extend that approach to other contexts if it helps.

c2rust-refactor/src/transform/literals.rs
```diff
@@ -519,11 +520,12 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
...
 
+    /// Unify a child expression when the parent key tree wraps a single element.
     fn unify_expr_child(&mut self, e: &Expr, kt: LitTyKeyTree<'kt, 'tcx>) {
-        if let Some(ch_kt) = kt.get().children() {
-            assert!(ch_kt.len() == 1);
-            self.visit_expr_unify(e, ch_kt[0]);
+        if let Some(&[child_kt]) = kt.get().children() {
+            self.visit_expr_unify(e, child_kt);
         } else {
+            // Not a single-child wrapper: visit without unification
```
ahomescu, line 528: The old code would assert and crash on purpose so we could see what's not implemented yet. If you get rid of that, it would be useful to at least log nodes with multiple children.

c2rust-refactor/src/transform/literals.rs
```diff
@@ -556,7 +558,7 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
                 let callee_key_tree = self.expr_ty_to_key_tree(callee);
                 self.visit_expr_unify(callee, callee_key_tree);
                 // TODO: check if generic
-                if let Some(&[ref input_key_trees @ .., output_key_tree]) = callee_key_tree.get().children() {
+                if let LitTyKeyNode::Node(&[ref input_key_trees @ .., output_key_tree]) = callee_key_tree.get() {
```
ahomescu, line 561: Same as above: the panic was intentional.

ahomescu:
> What exactly determines the structure of this tree: the structure of types themselves (related via having other types as fields, type arguments, etc.),

That's right, the recursive structure of the types themselves.

> but what's the benefit of including LitTyKeyNode::Empty in these trees if we don't do any unfication when we see it?

Consistency and simplicity. We can have an `Empty` node for types we don't care about, instead of somehow marking a missing child from the parent.

> The result are mismatches such as field types instead of structs (Leafs instead of Nodes), tuples instead of references to tuples (two children vs. one child), and types of method call parameters instead of the method calls themselves (the reason for the "unresolved" id).

Could you provide some examples here? It might help us get a better idea where the issues are.

oinoom:
Every expression and sub expression for the expanded macros has the same span, even for different macro lines.

```rust
// BEFORE
#[repr(C)]
#[derive(Copy, Clone, BitfieldStruct)]
pub struct _xmlSchemaValDate {
    pub year: core::ffi::c_long,
    #[bitfield(name = "mon", ty = "core::ffi::c_uint", bits = "0..=3")]
    #[bitfield(name = "day", ty = "core::ffi::c_uint", bits = "4..=8")]
    #[bitfield(name = "hour", ty = "core::ffi::c_uint", bits = "9..=13")]
    #[bitfield(name = "min", ty = "core::ffi::c_uint", bits = "14..=19")]
    pub mon_day_hour_min: [u8; 3],
    ...
}

// AFTER
#[repr(C)]
#[derive(Copy, Clone, BitfieldStruct)]
pub struct _xmlSchemaValDate {
    pub year: core::ffi::c_long,
    #[bitfield(name = "mon", ty = "core::ffi::c_uint", bits = "0..=3")]
    #[bitfield(name = "day", ty = "core::ffi::c_uint", bits = "4..=8")]
    #[bitfield(name = "hour", ty = "core::ffi::c_uint", bits = "9..=13")]
impl _xmlSchemaValDate {
  pub fn set_mon(&mut self, int: core::ffi::c_uint) {
      use c2rust_bitfields::FieldType;
      let field = &mut self.mon_day_hour_min;
      let (lhs_bit, rhs_bit) = (0usize, 3usize);
      int.set_field(field, (lhs_bit, rhs_bit));
  }
  pub fn mon(&self) -> core::ffi::c_uint {
      use c2rust_bitfields::FieldType;
      type IntType = core::ffi::c_uint;
      let field = &self.mon_day_hour_min;
      let (lhs_bit, rhs_bit) = (0usize, 3usize);
      <IntType as FieldType>::get_field(field, (lhs_bit, rhs_bit))
  }
  pub fn set_day(&mut self, int: core::ffi::c_uint) {
      use c2rust_bitfields::FieldType;
      let field = &mut self.mon_day_hour_min;
      let (lhs_bit, rhs_bit) = (4usize, 8usize);
      int.set_field(field, (lhs_bit, rhs_bit));
  }
  pub fn day(&self) -> core::ffi::c_uint {
      use c2rust_bitfields::FieldType;
      type IntType = core::ffi::c_uint;
      let field = &self.mon_day_hour_min;
      let (lhs_bit, rhs_bit) = (4usize, 8usize);
      <IntType as FieldType>::get_field(field, (lhs_bit, rhs_bit))
  }
  pub fn set_hour(&mut self, int: core::ffi::c_uint) {
      use c2rust_bitfields::FieldType;
      let field = &mut self.mon_day_hour_min;
      let (lhs_bit, rhs_bit) = (9usize, 13usize);
      int.set_field(field, (lhs_bit, rhs_bit));
  }
  pub fn hour(&self) -> core::ffi::c_uint {
      use c2rust_bitfields::FieldType;
      type IntType = core::ffi::c_uint;
      let field = &self.mon_day_hour_min;
      let (lhs_bit, rhs_bit) = (9usize, 13usize);
      <IntType as FieldType>::get_field(field, (lhs_bit, rhs_bit))
  }
  ...
}
```

ahomescu:
`c2rust-bitfields-derive` seems to use `Span::call_site` a lot, maybe the fix is to replace all of those with proper spans.

oinoom:
I have a WIP [here](https://github.com/immunant/c2rust/pull/1445)

oinoom:
@fw-immunant @ahomescu I made a more informed attempt in #1445

PR #1433, opened 2025-10-24
refactor: avoid reordering attributes when restoring them

### LINKED ISSUE #1426: c2rust-refactor moves derive attributes
Given this input file, "derive.rs":
```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]

#[derive(Copy, Clone)]
#[repr(C)]
pub struct S {}
```

running `c2rust-refactor --rewrite-mode diff remove_unused_labels -- derive.rs` produces:

```

--- old/derive.rs
+++ new/derive.rs
@@ -7,7 +7,7 @@
     unused_mut
 )]

-#[derive(Copy, Clone)]
-#[repr(C)]
+
+#[repr(C)]#[derive(Copy, Clone)]
 pub struct S {}
```

This appears to be an issue with roundtripping attributes, as it also occurs with various other transforms that aren't supposed to modify attributes, such as `test_one_plus_one`. This adds lots of noise to the diffs generated and makes it harder to identify what transforms are actually intending to change.

This change appears to only occur if the `derive` attribute precedes the `repr` one.

fw-immunant:
* Fixes #1426

including in the presence of multiple `#[derive(...)]` attributes interleaved with other attributes. No test at present; I'll look into adding one.

### DIFF
--- a/c2rust-refactor/tests/snapshots.rs
+++ b/c2rust-refactor/tests/snapshots.rs
@@ -109,6 +109,11 @@ fn test_rename_unnamed() {
     test_refactor("rename_unnamed");
 }
 
+#[test]
+fn test_reorder_derives() {
+    test_refactor_named("noop", "reorder_derives.rs");
+}
+
 #[cfg(target_os = "linux")] // `statvfs` and `statfs64` are Linux only.
 #[test]
 fn test_reorganize_definitions() {
--- a/c2rust-refactor/tests/snapshots/reorder_derives.rs
+++ b/c2rust-refactor/tests/snapshots/reorder_derives.rs
@@ -1,0 +1,26 @@
+#![allow(
+    dead_code,
+    non_camel_case_types,
+    non_snake_case,
+    non_upper_case_globals,
+    unused_assignments,
+    unused_mut
+)]
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct S0 {}
+
+#[derive(Clone)]
+#[repr(u8)]
+#[derive(Copy)]
+#[derive(Debug)]
+pub struct S1 {}
+
+#[cfg(not(test))]
+#[derive(Debug, Copy, Clone)]
+#[repr(u32)]
+pub struct S2 {}
+
+#[derive(Debug)]
+pub struct S3 {}
--- a/c2rust-refactor/tests/snapshots/snapshots__refactor-noop-reorder_derives.rs.snap
+++ b/c2rust-refactor/tests/snapshots/snapshots__refactor-noop-reorder_derives.rs.snap
@@ -1,0 +1,29 @@
+---
+source: c2rust-refactor/tests/snapshots.rs
+expression: c2rust-refactor noop --rewrite-mode alongside -- tests/snapshots/reorder_derives.rs --edition 2021 -Awarnings
+---
+#![allow(
+    dead_code,
+    non_camel_case_types,
+    non_snake_case,
+    non_upper_case_globals,
+    unused_assignments,
+    unused_mut
+)]
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct S0 {}
+
+#[derive(Clone)]
+#[repr(u8)]
+#[derive(Copy, Debug)]
+pub struct S1 {}
+
+#[cfg(not(test))]
+#[derive(Debug, Copy, Clone)]
+#[repr(u32)]
+pub struct S2 {}
+
+#[derive(Debug)]
+pub struct S3 {}


fw-immunant:
I added a test but I'm not sure that these c2rust-refactor tests are actually run in CI.

It seems like we're not running `scripts/test_rust_refactor.py`.

c2rust-refactor/src/collapse/macros.rs
```diff
@@ -338,18 +338,20 @@ impl<'a> MutVisitor for CollapseMacros<'a> {
...
+        Iterator::eq(
+            a.tokens().to_tokenstream().into_trees(),
+            b.tokens().to_tokenstream().into_trees(),
+        )
     }
 
-    if let Some(attr) = crate::util::find_by_name(old_attrs, sym::cfg) {
-        if !crate::util::contains_name(&new_attrs, sym::cfg) {
-            new_attrs.push(attr.clone());
+    // If the original item had a `#[derive]` attr, transfer it to the new one.
```
ahomescu: Or `cfg` right?

c2rust-refactor/src/collapse/macros.rs
```diff
@@ -338,18 +338,20 @@ impl<'a> MutVisitor for CollapseMacros<'a> {
...
+    for old_attr in old_attrs.iter().rev() {
```
ahomescu, line 349: No objection to the new code, just curious: what was wrong with the old approach?

c2rust-refactor/tests/reorder_derives/run.sh
kkysen, line 1: Can this just be written as a Rust `#[test]` instead?  `cargo test` is run in CI.

kkysen: `test_rust_refactor.py` is pretty small (unlike `test_translator.py`), and it makes a lot of things more seamless to keep it all in `cargo test`.

c2rust-refactor/src/collapse/macros.rs
fw-immunant: The reason to change from the old approach here is that it assumed there could not be multiple `derive` or `cfg` attrs (dropping all but one of each), and only considered the attribute name but not its contents, causing it to treat `#[cfg(foo)]` being present in `new_attrs` as a reason not to copy over `#[cfg(bar)]` from `old_attrs`.

The ordering fix is contained in the final hunk of the patch.

c2rust-refactor/tests/reorder_derives/run.sh
fw-immunant: We do want to test end-to-end refactorer behavior here if possible, so if you have a suggestion on how to express that as a `#[test]` (e.g. setting up integrated snapshot testing for the refactorer), I'm all ears.

c2rust-refactor/src/collapse/macros.rs
fw-immunant: Fixed.

c2rust-refactor/src/collapse/macros.rs
ahomescu: Got it 👍

fw-immunant:
> Is this why stuff like `#[c2rust::src_loc = "125:8"]#[derive(Copy, Clone)]` is happening? I noticed that while testing crisp.

Yeah, this PR fixes that weird formatting.

ahomescu:
We should land this.

c2rust-refactor/tests/reorder_derives/run.sh
ahomescu: I propose we land this, and then clean up the test as a follow-up.

kkysen: The snapshot infrastructure should be ready to go now, though.

fw-immunant: I've changed the test to use this new infra, though it does need the named test support since this isn't testing a pass of its own.

c2rust-refactor/tests/snapshots.rs
```diff
@@ -109,6 +109,11 @@ fn test_rename_unnamed() {
     test_refactor("rename_unnamed");
 }
 
+#[test]
+fn test_reorder_derives() {
+    test_refactor_named("test_one_plus_one", "reorder_derives.rs");
```
kkysen: Where's `reorder_derives.rs`?  Also, would this test work with a different transform?  `test_one_plus_one` isn't working right now.

fw-immunant: Forgot to add it to the commit, fixed now. Any transform (or an identity transform, if we had one) should work--this test doesn't rely on `test_one_plus_one` actually rewriting its target expression.

kkysen: Is there a `noop` one? I think I saw it somewhere.

fw-immunant: Ah, there is. Using that now.

PR #1434, opened 2025-10-25
transpile: Use literal directly when taking address of string literal

Rua:
This adds a special case for the address-of operator when its operand is a string literal. The code generated after #1360 was problematic because it results in a borrow of a temporary (from the transmute), which leaves dangling pointers.

I plan to factor out the common code between `convert_cast` and `convert_unary_operator`, but I think I'll do that in a followup PR and first make sure that this one is ok.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -938,10 +938,10 @@ impl<'c> Translation<'c> {
                 // In this translation, there are only pointers to functions and
                 // & becomes a no-op when applied to a function.
 
-                let arg = self.convert_expr(ctx.used().set_needs_address(true), arg, None)?;
+                let val = self.convert_expr(ctx.used().set_needs_address(true), arg, None)?;
 
                 if self.ast_context.is_function_pointer(ctype) {
-                    Ok(arg.map(|x| mk().call_expr(mk().ident_expr("Some"), vec![x])))
+                    Ok(val.map(|x| mk().call_expr(mk().ident_expr("Some"), vec![x])))
                 } else {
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/operators.rs
```diff
@@ -950,13 +950,58 @@ impl<'c> Translation<'c> {
...
+                        .is_some();
+
+                    if let (
+                        &CExprKind::Literal(
+                            literal_cqual_type,
+                            CLiteral::String(ref bytes, element_size @ 1),
+                        ),
+                        true,
+                        false,
+                    ) = (
```
fw-immunant: I think a comment is warranted here to explain exactly why we translate this case differently; something like ``// Because we translate string literals into temporary `c_char` arrays, we should not take the address of such a translation. Instead, cast the string literal directly.``

I also don't fully follow why `pointee_ty.qualifiers.is_const` matters here. Could you also comment to explain that?

fw-immunant (changes requested):
Semantically this change looks good, but please add a comment explaining the logic of the added branch.

c2rust-transpile/src/translator/operators.rs
Rua: I took the `is_const` logic from the code used for array-to-pointer decay, which already had the same exception:
https://github.com/immunant/c2rust/blob/860770d9e68c68f3453f57775d98ac3c367e6ef2/c2rust-transpile/src/translator/mod.rs#L4630

Admittedly I'm not sure if it's actually needed, but then it's probably not needed in the other case either. I can remove it if desired.

Issue #1435, opened 2025-10-28
Integer literal overflow in C causes Rust compilation error


HurryPeng:
C source code (from [LibmCS](https://gitlab.com/gtd-gmbh/libmcs/-/blob/1.2.0/libm/mathd/sqrtd.c?ref_type=tags#L63)):
```c
int32_t sign = 0x80000000U;
```
C2Rust 0.21.0 translates it to:
```rust
pub type int32_t = __int32_t;
pub type __int32_t = i32;
let mut sign: int32_t = 0x80000000 as int32_t;
```
which causes this compilation error:
```
error: literal out of range for `i32`
  --> src/mathd/sqrtd.rs:42:29
   |
42 |     let mut sign: int32_t = 0x80000000 as int32_t;
   |                             ^^^^^^^^^^
   |
   = note: `#[deny(overflowing_literals)]` on by default
   = note: the literal `0x80000000` (decimal `2147483648`) does not fit into the type `i32` and will become `-2147483648i32`
   = help: consider using the type `u32` instead
```
C2Rust 0.20.0 translates that to:
```rust
pub type int32_t = __int32_t;
pub type __int32_t = libc::c_int;
let mut sign: int32_t = 0x80000000 as libc::c_uint as int32_t;
```
which compiles.

ahomescu:
I wonder if `let mut sign: int32_t = 0x80000000u32 as int32_t;` would compile fine.

PR #1436, opened 2025-10-28
transpile: tests: add `out_of_range_int.c` test

kkysen:
A test for #1435.  It was broken by #1266 (specifically, 7eaa39f, which also doesn't compile).  It seems we're not handling the `U` suffix, which requires a temporary cast to `c_uint` before getting coerced to `int32_t`/`i32`.

4a4c0cc is the current snapshot, while 36d829e is what it was before #1266, which still compiles.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@out_of_range_int.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@out_of_range_int.c.snap
@@ -5,17 +5,18 @@ input_file: c2rust-transpile/tests/snapshots/out_of_range_int.c
 ---
 #![allow(
     dead_code,
+    mutable_transmutes,
     non_camel_case_types,
     non_snake_case,
     non_upper_case_globals,
     unused_assignments,
     unused_mut
 )]
-pub type __int32_t = i32;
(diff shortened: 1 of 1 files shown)


fw-immunant:
Ah, my bad re: 7eaa39f63219567572695331192a78733ee1e506 not compiling--it depends on 67e6eebce73df5c2eeba319a696d0673db7c6541 but got reordered prior to it during cleanup of that PR.

I believe the problem here is referenced by the doc comment for `literal_kind_matches_ty`:
> Return whether the literal can be directly translated as this type. This does not check that the literal fits into the type's range of values (as doing so is in general dependent on the target platform), just that it is the appropriate kind for the type.

For fixed-size types like `int32_t`, we can relatively easily check the literal against their fixed bounds, but for types like `libc::c_int` we need to be able to determine the bounds for `int` on the target platform.

kkysen (commented):
> for types like `libc::c_int` we need to be able to determine the bounds for `int` on the target platform.

See #1442 now, but I chose to use the narrowest bound for platform-dependent types like `int` so that the resulting code works on all platforms.

kkysen:
* Superseded by #1442.

PR #1439, opened 2025-10-30
transpile: Rework `is_va_list` for better accuracy

### LINKED ISSUE #1353: `va_start` type mismatch
When transpiling this C code

```C
void I_Error (char *error, ...)
{
    va_list argptr;
    va_start (argptr,error);
```

the following code is emitted
```rust
#[no_mangle]
pub unsafe extern "C" fn I_Error(mut error: *mut std::ffi::c_char, mut args: ...) {
    let mut argptr: va_list = 0 as *mut std::ffi::c_char;
    argptr = args.clone();
```

The last line creates a compile error, because `args` is a Rust type, `std::ffi::VaListImpl`, and doesn't match the type of `argptr`.

Rua:
* Fixes #1353.

This reworks the implementation of `is_va_list`, so that it is more accurate (less false positives/negatives). It's split into two subfunctions, one that detects `va_list` based on the `__builtin_va_list` typedef, and the other that detects based on the architecture-specific struct definition. Both methods have situations in which they don't work, so they are used together.

This has been confirmed to work on:
- Linux x86-64
- Linux x86-32 (was previously broken)

MacOS aarch64 is still failing if I enable it, because things like `va_start` seem to be unimplemented. Testing for Linux aarch64 will depend on #1283.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -347,81 +347,120 @@ impl TypedAstContext {
         .is_some()
     }
 
-    /// Follow a chain of typedefs and return true iff the last typedef is named
-    /// `__builtin_va_list` thus naming the type clang uses to represent `va_list`s.
-    pub fn is_builtin_va_list(&self, typ: CTypeId) -> bool {
-        match self.index(typ).kind {
-            CTypeKind::Typedef(decl) => match &self.index(decl).kind {
-                CDeclKind::Typedef {
-                    name: name_,
-                    typ: ty,
-                    ..
-                } => {
-                    if name_ == "__builtin_va_list" {
-                        true
-                    } else {
-                        self.is_builtin_va_list(ty.ctype)
-                    }
-                }
+    /// Returns whether `typ` is a chain of typedefs that ends in `__builtin_va_list`,
+    /// thus naming the type clang uses to represent `va_list`s.
+    /// This works on all architectures, but does not work in situations where typedefs are
+    /// resolved/bypassed, such as with array-to-pointer decay.
+    pub fn is_builtin_va_list(&self, mut typ: CTypeId) -> bool {
+        loop {
+            // Skip over Elaborated types
+            let mut kind = &self.index(typ).kind;
+
+            while let &CTypeKind::Elaborated(typ) = kind {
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -347,83 +347,126 @@ impl TypedAstContext {
...
-                // The mechanism applies: va_list is a `struct __va_list { ... }` as per
-                // https://documentation-service.arm.com/static/5f201281bb903e39c84d7eae
-                // ("Procedure Call Standard for the Arm Architecture Release 2020Q2, Document
-                // number IHI 0042J") Section 8.1.4 "Additional Types"
-                self.is_va_list_struct(typ)
+                name == "__va_list_tag"
             }
 
-            kind => unimplemented!("va_list type {:?} not yet implemented", kind),
+            // TODO:
```
ahomescu: We probably don't care about these, PNaCl was replaced by Wasm.

Rua: I've removed the TODO.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4601,7 +4601,7 @@ impl<'c> Translation<'c> {
                 // and to be a pointer as a function argument we would get
                 // spurious casts when trying to treat it like a VaList which
                 // has reference semantics.
-                if self.ast_context.is_va_list(target_cty.ctype) {
+                if self.ast_context.is_va_list(source_cty.ctype) {
```
ahomescu: You are accounting for decay above on line 440, so why is this change needed?

Rua: I made the change in case there are other cases that produce a `va_list` as the target type, but that have a different source type. I think it may not be necessary, let's try it.

kkysen (commented):
Can you add the example from #1353 as a snapshot test?  You can do that as a commit after the fixes, as prior to the fix, it wouldn't compile (if I'm understanding that correctly).  Thanks!

Rua:
It was only broken for x86-32, which the tests wouldn't catch anyway. Also since the same thing was suggested in https://github.com/immunant/c2rust/pull/1443#pullrequestreview-3468980085 I'll go with that one instead.

PR #1442, opened 2025-11-01
transpile: fix out-of-range literals

### LINKED ISSUE #1435: Integer literal overflow in C causes Rust compilation error

kkysen:
* Fixes #1435.
* Supersedes #1436.

This first reverts the part of 7eaa39f that broke #1435, which is omitting casts from literals where the type already matches without checking if the literal is within that type's range.  Then tests are added and then `fn literal_kind_matches_ty` is fixed to also check the type's range.  If the literal may be out of range on any possible platform, then a cast remains.

I didn't add a float to these tests since floats go through a different code path (there's no cast generated for them).  So for them, we need to explicitly generate a cast if it's out of range.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -2202,7 +2202,8 @@ impl CTypeKind {
         match *self {
             Void => false,
 
-            Bool => (0..=1).contains(&value),
+            // Kind of an integer type, but would definitely need an explicit cast.
+            Bool => false,
 
             // Can be signed or unsigned, so choose the minimum range of each.
             Char => (u8::MIN as u64..=i8::MAX as u64).contains(&value),


c2rust-transpile/src/translator/literals.rs
```diff
@@ -72,16 +72,18 @@ impl<'c> Translation<'c> {
         mk().cast_expr(value, target_ty)
     }
 
-    /// Return whether the literal can be directly translated as this type. This does not check
-    /// that the literal fits into the type's range of values (as doing so is in general dependent
-    /// on the target platform), just that it is the appropriate kind for the type.
+    /// Return whether the literal can be directly translated as this type.
     pub fn literal_kind_matches_ty(&self, lit: &CLiteral, ty: CQualTypeId) -> bool {
```
fw-immunant: At this point, a name like `literal_matches_ty` would make more sense as it's now analyzing the value, not just the kind.

fw-immunant:
For CI, looks like a macOS snapshot needs updating.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2183,6 +2183,177 @@ impl CTypeKind {
...
+            Int64 => in_range::<i64>(value),
+            Int128 => in_range::<i128>(value),
+
+            UInt8 => in_range::<u8>(value),
+            UInt16 => in_range::<u16>(value),
+            UInt32 => in_range::<u32>(value),
+            UInt64 => in_range::<u64>(value),
+            UInt128 => in_range::<u128>(value),
+
+            // There's no guarantee on pointer size, but `NULL` should work.
```
ahomescu, line 2238: It's guaranteed to be at least as large as a `u16`, but I'm not sure that's useful here. It's rare that you'll need a constant pointer that's within the first 64KB.

Edit: Never mind, I realized this is for C pointers.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2183,6 +2183,177 @@ impl CTypeKind {
...
+            Short => in_range::<i16>(value),
+            Int => in_range::<i16>(value),
+            Long => in_range::<i32>(value),
+            LongLong => in_range::<i64>(value),
+
+            // `unsigned int` is at least `u16` and `unsigned long` is at least `u32`.
+            UChar => in_range::<u8>(value),
+            UShort => in_range::<u16>(value),
+            UInt => in_range::<u16>(value),
+            ULong => in_range::<u32>(value),
```
ahomescu, line 2223: Depends on the ABI: LP64 (Linux and macOS) also uses 64 bits for `long`, while Windows uses 32. If you're just being conservative here and trying to handle all platforms, I'd add a comment here.

kkysen: Oh, is this not enough?  I can try to add more from the PR description, where I explained more.

```rust
    /// Whether `value` is guaranteed to be in this integer type's range.
    /// Thus, minimum ranges are used (e.x. [`i16::MIN`] for [`Self::Int`]).
```

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2183,6 +2183,177 @@ impl CTypeKind {
...
+            IntMax => in_range::<i64>(value),
+            UIntMax => in_range::<u64>(value),
+
+            // `size_t` is at least a `u16`, and similar for `ssize_t` and `ptrdiff_t`.
+            Size => in_range::<u16>(value),
+            SSize => in_range::<i16>(value),
+            PtrDiff => in_range::<i16>(value),
+
+            // Floats, see `Self::guaranteed_float_in_range`.
+            Float => false,
```
ahomescu, line 2251: Technically this still works if the integer fits into the mantissa, i.e., 24 bits.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2183,6 +2183,177 @@ impl CTypeKind {
...
+            Double => false,
```
ahomescu, line 2252: Larger mantissa here: 53 bits.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2183,6 +2183,177 @@ impl CTypeKind {
...
+            ConstantArray(_, _) => false,
+            IncompleteArray(_) => false,
+            VariableArray(_, _) => false,
+            TypeOf(_) => false,
+            TypeOfExpr(_) => false,
+            Function(_, _, _, _, _) => false,
+            Typedef(_) => false,
+            Decayed(_) => false,
+            Elaborated(_) => false,
+            Paren(_) => false,
```
ahomescu, line 2271: Couldn't this be recursive?

c2rust-transpile/src/translator/literals.rs
```diff
@@ -72,16 +72,18 @@ impl<'c> Translation<'c> {
...
         let ty_kind = &self.ast_context.resolve_type(ty.ctype).kind;
         match *lit {
-            CLiteral::Integer(..) if ty_kind.is_integral_type() && !ty_kind.is_bool() => true,
+            CLiteral::Integer(value, _) if ty_kind.is_integral_type() && !ty_kind.is_bool() => {
```
ahomescu, line 77: You handle `Bool` in the `guaranteed_integer_in_range`, why is the `is_bool` check here?

c2rust-transpile/src/c_ast/mod.rs
kkysen: It's not meant to be called on floats at all.

c2rust-transpile/src/c_ast/mod.rs
kkysen: It's only meant to be called on integers.

c2rust-transpile/src/translator/literals.rs
kkysen: I can remove it from `fn guaranteed_integer_in_range` if that's clearer. It's handled differently here.

c2rust-transpile/src/translator/literals.rs
kkysen: Good point.  Fixed in c91d5ef.

c2rust-transpile/src/c_ast/mod.rs
ahomescu: Couldn't it be called on something like `123456 as f32`?

kkysen: No, we never go down that code path.

c2rust-transpile/src/c_ast/mod.rs
ahomescu: Yeah you could elaborate a bit, maybe add a line on where differences could come from.

c2rust-transpile/src/c_ast/mod.rs
ahomescu: Got it, that's surprising.

c2rust-transpile/src/c_ast/mod.rs
kkysen: Clarified the docs in 2e94402.

c2rust-transpile/src/c_ast/mod.rs
kkysen: Clarified in the docs in 813e8e3.

c2rust-transpile/src/translator/literals.rs
kkysen: Removed in bc89539.

PR #1443, opened 2025-11-01
transpile: Translate `va_list` with correct types

### LINKED ISSUE #1440: Pointer to `va_list` is translated wrong
When the `va_list` type appears by value in a function parameter, c2rust translates it as a Rust `VaList`. That's correct. But when a function takes `va_list *` by pointer, then that gets translated as Rust `*mut VaList`, which is *not* correct. To see why, consider what such a pointer actually points to on architectures that use a dedicated struct type for `va_list`, such as x86-64 or Linux aarch64:
- In C, `va_list` is a struct or array containing a struct. A pointer `va_list *` is thus a pointer to this struct or array. Dereferencing the pointer gives you the struct.
- In Rust, `VaList` does not contain the struct, which is in `VaListImpl`. `VaList` only contains a reference to the struct. Thus, dereferencing `*mut VaList` gives you the `VaList`, which is not the struct but that reference. You need to dereference `VaList` again to get the struct inside `VaListImpl`.

Clearly, these two are not equivalent and therefore not ABI compatible. The correct translation would be to use `VaListImpl` normally, but `VaList` if and only if C passes `va_list` *by value*.  `va_list *` by pointer should always be translated to `*mut VaListImpl`.

Rua:
- Depends on #1491.
- Fixes #1440.

This changes how the `va_list` type is translated: as `VaListImpl` by default, and as `VaList` only if passed by value to a function. A borrowed `va_list *` is now translated as `*mut VaListImpl` rather than `*mut VaList` as before. I've also added a test for it.

### DIFF
--- a/c2rust-transpile/src/convert_type.rs
+++ b/c2rust-transpile/src/convert_type.rs
@@ -250,7 +250,10 @@ impl TypeConverter {
     ) -> TranslationResult<Box<Type>> {
         let barefn_inputs = params
             .iter()
-            .map(|x| mk().bare_arg(self.convert(ctxt, x.ctype).unwrap(), None::<Box<Ident>>))
+            .map(|x| {
+                let ty = self.convert_function_param(ctxt, x.ctype).unwrap();
+                mk().bare_arg(ty, None::<Box<Ident>>)
+            })
             .collect::<Vec<_>>();
 
         let output = match ret {
@@ -322,7 +325,7 @@ impl TypeConverter {
         ctype: CTypeId,
     ) -> TranslationResult<Box<Type>> {
         if self.translate_valist && ctxt.is_va_list(ctype) {
-            let ty = mk().abs_path_ty(vec!["core", "ffi", "VaList"]);
+            let ty = mk().abs_path_ty(vec!["core", "ffi", "VaListImpl"]);
             return Ok(ty);
         }
 
@@ -445,6 +448,22 @@ impl TypeConverter {
 
             ref t => Err(format_err!("Unsupported type {:?}", t).into()),
         }
+    }
+
+    // Variant of `convert` that handles types that need to be converted differently if they
+    // are the type of a function parameter.
+    pub fn convert_function_param(
+        &mut self,
+        ctxt: &TypedAstContext,
+        ctype: CTypeId,
+    ) -> TranslationResult<Box<Type>> {
+        if ctxt.is_va_list(ctype) {
+            // va_list parameters are translated as VaList rather than VaListImpl
+            let ty = mk().abs_path_ty(vec!["core", "ffi", "VaList"]);
+            return Ok(ty);
+        }
+
+        self.convert(ctxt, ctype)
     }
 
     /// Add the given parameters to a K&R function pointer type,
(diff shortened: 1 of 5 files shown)


fw-immunant:
The changes in the second commit cause the following diff to the transpiled output for the `items` test:
```patch
--- a/varargs.rs
+++ b/varargs.rs
@@ -45,7 +45,7 @@
     mut format: *const ::core::ffi::c_char,
     mut ap: ::core::ffi::VaList,
 ) {
-    vprintf(format, ap.as_va_list() as ::core::ffi::VaList);
+    vprintf(format, ap.as_va_list());
 }
 #[no_mangle]
 pub unsafe extern "C" fn rust_call_vprintf(
@@ -156,7 +156,7 @@
     vprintf(fmt, ap.as_va_list());
 }
 #[no_mangle]
-pub unsafe extern "C" fn rust_print_int(mut ap: *mut ::core::ffi::VaList) {
+pub unsafe extern "C" fn rust_print_int(mut ap: *mut ::core::ffi::VaListImpl) {
     printf(
         b"%d\0" as *const u8 as *const ::core::ffi::c_char,
         (*ap).arg::<::core::ffi::c_int>(),
@@ -167,7 +167,7 @@
     let mut ap: ::core::ffi::VaListImpl;
     ap = args.clone();
     while count > 0 as size_t {
-        rust_print_int(&mut ap.as_va_list());
+        rust_print_int(&mut ap);
         count = count.wrapping_sub(1);
     }
 }
```

These seem reasonable, but I'm not an expert on our variadics transpilation; I don't know enough to say why the previous translation is incorrect. Cc @ahomescu?

Rua:
It comes down to how `VaList` is defined in Rust:
https://doc.rust-lang.org/src/core/ffi/va_list.rs.html#121-170

On platforms where C's `va_list` is a simple pointer, Rust's `VaList` is simply that pointer. But on platforms where `va_list` is a struct (or 1-element array containing a struct), Rust's `VaList` is a pointer to that struct rather than the struct itself. Thus, `VaList` always consists of a single pointer, whereas `va_list` does not and can be larger than `VaList`. Rust's `VaListImpl`, on the other hand, always has the same definition as `va_list`.

ahomescu (approved):

ahomescu:
This change looks good to me, I don't think we accounted for `va_list*` in our original implementation. Afaict the decay should only happen for pure `va_list` parameters, not `va_list*`.

kkysen (commented):
Do we also want `tests/items/src/varargs.c` as a snapshot test?  Clearly it was useful during review to see what the diff was there.

Rua:
> Do we also want `tests/items/src/varargs.c` as a snapshot test?

The definition of `__builtin_va_list` is different for different architectures and OS. So that would need an addition to the test mechanism to handle snapshots that are specific to both. Alternatively, `__builtin_va_list` could just be omitted from the transpiler output altogether, since the transpiled code isn't meant to be using it anyway. I don't know what causes it to be emitted even when unused.

kkysen (commented):
> > Do we also want `tests/items/src/varargs.c` as a snapshot test?
> 
> The definition of `__builtin_va_list` is different for different architectures and OS. So that would need an addition to the test mechanism to handle snapshots that are specific to both.

That should be simpler to do after #1444.  Well actually, it's not too tricky to add now, but it'd have to be modified for #1444.

> Alternatively, `__builtin_va_list` could just be omitted from the transpiler output altogether, since the transpiled code isn't meant to be using it anyway. I don't know what causes it to be emitted even when unused.

If this can work, that'd also be great.  I don't think we need to block this PR on that, though, but it would be a nice follow-up.

Rua:
I've now rebased this PR onto #1491, which adds a snapshot test for varargs.

kkysen (approved):

PR #1445, opened 2025-11-03
Fix NodeId->HirId lookups by carrying structural context

oinoom:
c2rust-refactor would fail due to invalid AST node <-> HIR mappings. This was caused by insufficient disambiguation due to all c2rust-bitfield-derived expanded code sharing the same span with call_site. This PR addresses the aforemented issue by threading structural context through the AST and HIR walkers so NodeId-HirId lookups stay precise even when spans collide.

- Added `child_slot`, `NodeContextKey`, and `StructuralContext` helpers to record hierarchical positions for AST nodes; mirrored the same traversal on the HIR side and stitched the maps together during `HirMap` construction
- Collapsed wrapper nodes derived accessors and helper shims would erase array/tuple shells on the expression side, leaving `Leaf` nodes that conflicted with the HIR/type view where the same values remained wrapped (e.g. `[u8; 3]` -> `Node([Leaf(u8)])`). We now detect those single-child wrappers, peel just enough layers to expose the matching field list, and otherwise panic with a targeted message if the arity still disagrees.

I tried going down another path with more-specific spans, such as using quote_spanned!(field.field_name.span()), using fresh spans with Span::def_site().located_at(old_span), and using attribute metadata (with meta_name_value.lit.span(), but kept running into issues and so abandoned that path.

### DIFF
--- a/c2rust-refactor/src/ast_manip/span_maps.rs
+++ b/c2rust-refactor/src/ast_manip/span_maps.rs
@@ -1,7 +1,150 @@
 use rustc_ast::visit::{self, AssocCtxt, Visitor};
 use rustc_ast::*;
 use rustc_data_structures::fx::FxHashMap;
+use rustc_span::source_map::symbol::Symbol;
 use rustc_span::Span;
+use smallvec::SmallVec;
+
+/// Shared enumeration of child positions within a parent node.
+/// The AST span mapper records a path of these slot values while walking each expression, and
+/// the HIR visitor in `context.rs` reconstructs the same path.  Later, lookups compare both the
+/// span and the path so that nodes which reuse the same span (for example, the derive-generated
+/// BitfieldStruct helpers that originally crashed with a NodeId -> HirId panic) still resolve to
+/// distinct HirIds.  Keeping the numbering identical on both sides is therefore required.
+pub mod child_slot {
+    // Expression slots - Binary operators
+    pub const BINARY_LHS: u16 = 0;
+    pub const BINARY_RHS: u16 = 1;
+
+    // Expression slots - Unary operators
+    pub const UNARY_OPERAND: u16 = 0;
+
+    // Expression slots - Function calls (base values for dynamic computation)
+    pub const CALL_CALLEE: u16 = 0;
+    pub const CALL_ARG_BASE: u16 = 1;
+    // CallArg0 = 1, CallArg1 = 2, ... (computed dynamically)
+
+    // Expression slots - Method calls
+    pub const METHOD_RECEIVER: u16 = 0;
+    pub const METHOD_ARG_BASE: u16 = 1;
+    // MethodArg0 = 1, MethodArg1 = 2, ... (computed dynamically)
+
+    // Expression slots - Tuples
+    pub const TUPLE_ELEM_BASE: u16 = 0;
+    // TupleElem0 = 0, TupleElem1 = 1, ... (computed dynamically)
+
+    // Expression slots - Arrays
+    pub const ARRAY_ELEM_BASE: u16 = 0;
+    // ArrayElem0 = 0, ArrayElem1 = 1, ... (computed dynamically)
+
+    // Expression slots - Struct construction
+    pub const STRUCT_FIELD_BASE: u16 = 0;
+    // StructField0 = 0, StructField1 = 1, ... (computed dynamically)
+
+    // Expression slots - Control flow
+    pub const IF_CONDITION: u16 = 0;
+    pub const IF_THEN_BRANCH: u16 = 1;
+    pub const IF_ELSE_BRANCH: u16 = 2;
+
+    pub const MATCH_SCRUTINEE: u16 = 0;
+    pub const MATCH_ARM_BASE: u16 = 1;
+    // MatchArm0 = 1, MatchArm1 = 2, ... (computed dynamically)
+
+    // Expression slots - Field access and indexing
+    pub const FIELD_BASE: u16 = 0;
+    pub const INDEX_BASE: u16 = 0;
+    pub const INDEX_INDEX: u16 = 1;
+
+    // Block slots (statements use stmt_index; tail expression gets explicit slot)
+    pub const BLOCK_TAIL: u16 = 0x0100;
+
+    // Pattern slots (base values for dynamic computation)
+    pub const PAT_TUPLE_ELEM_BASE: u16 = 0;
+    pub const PAT_STRUCT_FIELD_BASE: u16 = 0;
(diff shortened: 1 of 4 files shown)


c2rust-refactor/src/transform/literals.rs
```diff
@@ -295,6 +294,21 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
...
+                // the same aggregate, so peel the wrapper instead of flagging a mismatch.
+                self.unify_key_trees_internal(ch[0], kt2, seen);
+            }
+
+            (LitTyKeyNode::Leaf(_), LitTyKeyNode::Node(ref ch)) if ch.len() == 1 => {
+                // Symmetric case: the expression retained the wrapper (e.g. literal array) while
+                // the type-derived tree collapsed to the inner key.
+                self.unify_key_trees_internal(kt1, ch[0], seen);
+            }
+
```
fw-immunant, line 321: In meeting I mentioned it would be good to explain not just what happens in this individual cases but why we need to alter the logic in `unify_key_trees_internal` at all despite having improved our `NodeId`->`HidId` mapping.

c2rust-refactor/src/context.rs
```diff
@@ -733,6 +965,10 @@ impl<'hir> HirMap<'hir> {
             return *id;
         }
 
+        // TODO: consult `context_to_node_id_map` using the structural fingerprint recorded by
+        // `SpanToHirMapper`.  Falling straight back to `opt_local_def_id` still conflates siblings
+        // that share a span (e.g. derive helpers), so HIR -> NodeId lookups remain lossy until we
+        // thread the same context key through this path.
```
fw-immunant: As mentioned in meeting, it would be nice to clarify what uses `hir_to_node_id` and where we might see the consequences of not using the structural fingerprint here.

oinoom: Added some more context [here](https://github.com/immunant/c2rust/pull/1445/files#diff-bf29b093d220f1ee2cc2b4cfc0c926c0f66f9003de930e2bdc0c83567de758d4R951)

c2rust-refactor/src/transform/literals.rs
oinoom: More context [here](https://github.com/immunant/c2rust/pull/1445/files#diff-a0675c51322315600dec202b09d803bbc1d98a993a530f2e37f8ebfb149230f8R298)

c2rust-refactor/src/ast_manip/span_maps.rs
```diff
@@ -1,7 +1,150 @@
...
+            n < 128,
+            "closure parameter index overflow: {} exceeds maximum 127",
+            n
+        );
+        dynamic(CLOSURE_PARAM_BASE, n)
+    }
+}
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum SpanNodeKind {
```
ahomescu, line 150: Is this deprecated now? Could we replace it with your new context key?

c2rust-refactor/src/ast_manip/span_maps.rs
```diff
@@ -42,55 +185,266 @@ impl NodeSpan {
...
+    }
+}
+
 #[derive(Default, Clone)]
 pub struct AstSpanMaps {
     pub node_id_to_span_map: FxHashMap<NodeId, NodeSpan>,
     pub span_to_node_id_map: FxHashMap<NodeSpan, NodeId>,
+    /// Extended mapping: (NodeSpan, NodeContextKey) -> NodeId
+    /// This provides structural disambiguation for macro-generated nodes
+    pub context_to_node_id_map: FxHashMap<(NodeSpan, NodeContextKey), NodeId>,
```
ahomescu, line 263: Could these have replaced the old maps altogether?

c2rust-refactor/src/ast_manip/span_maps.rs
```diff
@@ -42,55 +185,266 @@ impl NodeSpan {
...
     }
 }
 
 #[derive(Default, Clone)]
-struct AstSpanMapper(AstSpanMaps);
+struct AstSpanMapper {
+    maps: AstSpanMaps,
+    /// Tracks block/owner/child-slot stacks while walking the AST
+    ctx: StructuralContext<NodeId>,
+    /// Depth of attribute visitation (>0 means we're inside an attribute)
```
ahomescu, line 346: Are attributes nested? What does that look like in Rust code?

c2rust-refactor/src/context.rs
```diff
@@ -689,23 +889,38 @@ impl<'hir> HirMap<'hir> {
...
+        if let Some(node_span) = self.ast_span_maps.node_id_to_span_map.get(&id) {
+            if let Some(context_key) = self.ast_span_maps.node_id_to_context_map.get(&id) {
+                if let Some(hir_id) = self
+                    .context_to_hir_map
+                    .get(&(*node_span, context_key.clone()))
+                {
+                    return Some(*hir_id);
+                }
+            }
+            return self.span_to_hir_map.get(node_span).copied();
```
ahomescu, line 908: Is there a case where a span would be in the old map but not the new context one?

c2rust-refactor/src/ast_manip/span_maps.rs
oinoom: We probably could, but the way I wrote this was to augment the usage of this so it may be a good chunk of work -- which I'm not opposed to but I don't think it'd be a simple substitution.

c2rust-refactor/src/ast_manip/span_maps.rs
oinoom: span_to_node_id_map used by the HIR->AST mapping doesn't use the context key to disambiguate at all, and I haven't tested it with that disambiguation, so it's not clear that it can be replaced with it yet. We could at least see if the context key could replace the span key in the AST->HIR mapping

c2rust-refactor/src/ast_manip/span_maps.rs
oinoom: In libxml the most common case is cfg_attr:

```
#![cfg_attr(feature = "xchecks", plugin(c2rust_xcheck_plugin(config_file = "../xchecks/libxml2_rust.yaml",
                                                             djb2_names_file = "../xchecks/djb2_names_libxml2.yaml")))]
```

c2rust-refactor/src/context.rs
oinoom: There should always be at least one in the new context map, but there may be more since the span is ambiguous

PR #1446, opened 2025-11-04
transpile: Factor out common function for address-of and array-decay

### LINKED ISSUE #1238: `(void*)array` Cast Still Emits Invalid `*mut _` Despite Fix for Scalar-to-void* Casts (Related to #1077)
While issue [#1077](https://github.com/immunant/c2rust/issues/1077) has been resolved — allowing safe casts `(bool to void*)` like `void* x = (void*)b;` to correctly emit Rust code — the related pattern of `casting arrays to void*` still triggers incorrect output in the translated Rust. Rust error code is also different here.

## C code

```c
int array[10];
void* ptr = (void*)array;

int main() {
    return 0;
}
```

## Translated Rust code:

```rust
#[no_mangle]
pub static mut array: [libc::c_int; 10] = [0; 10];
#[no_mangle]
pub static mut ptr: *mut libc::c_void = unsafe {
    array.as_ptr() as *mut _ as *mut libc::c_void
};
```

## Rust build error:

```text
error[E0641]: cannot cast to a pointer of an unknown kind
  --> src/runner.rs:15:23
   |
15 |     array.as_ptr() as *mut _ as *mut libc::c_void
   |                       ^^^^^^ needs more type information
   |
   = note: the type information given here is insufficient to check whether the pointer cast is valid

For more information about this error, try `rustc --explain E0641`.
```

### LINKED ISSUE #1240: Compound Literal Assigned to `volatile int*` Triggers Invalid Pointer Cast in Transpiled Rust
## Description

C2Rust generates invalid Rust code when a `volatile int*` is initialized using a compound literal. The transpiler emits a pointer cast involving `*mut _`, which leads to a compilation failure in Rust due to an ambiguous pointer type.

This issue **does not occur** when the same code uses `int*` or `const int*` instead.

## C code:

```c
volatile int *vs = { (int[]){1, 2, 3} }; 
```

## Translated Rust Code:

```rust
pub static mut vs: *mut libc::c_int = [1, 2, 3].as_ptr() as *mut _ as *mut libc::c_int;
    return 0 as libc::c_int;
}
```

## Cargo build failure:

```text
error[E0641]: cannot cast to a pointer of an unknown kind
  --> src/runner.rs:12:61
   |
12 | pub static mut vs: *mut libc::c_int = [1, 2, 3].as_ptr() as *mut _ as *mut libc::c_int;
   |                                                             ^^^^^^ needs more type information
   |
   = note: the type information given here is insufficient to check whether the pointer cast is valid

For more information about this error, try `rustc --explain E0641`.
```

Rua:
Depends on #1478.

- Fixes #1240.
- Fixes #1238.

This combines handling of the address-of and array-decay cast into a common function.

In the current code, this check for whether the created pointer type has a different pointee than the actual desired pointee type is only applied to array-to-pointer decays but not address-of. I left it that way, but perhaps it's also good to have it for address-of?
https://github.com/immunant/c2rust/blob/bbeb6de0e6d779a98c45947c7edc4ae77cb55ae4/c2rust-transpile/src/translator/mod.rs#L4837-L4843

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4708,84 +4708,12 @@ impl<'c> Translation<'c> {
                     return Ok(val);
                 }
 
-                let pointee = self
-                    .ast_context
-                    .get_pointee_qual_type(target_cty.ctype)
-                    .unwrap_or_else(|| panic!("dereferencing a non-pointer"));
-
-                let is_const = pointee.qualifiers.is_const;
-
-                // Handle literals by looking at the next level of expr nesting. Avoid doing this
-                // for expressions that will be translated as const macros, because emitting the
-                // name of the const macro only occurs if we process the expr_id with a direct call
-                // to `convert_expr`.
-                let expr_kind = expr.map(|e| &self.ast_context.index(e).kind);
-                let translate_as_macro = expr
-                    .map(|e| {
-                        self.convert_const_macro_expansion(ctx, e, None)
-                            .ok()
-                            .flatten()
-                            .is_some()
-                    })
-                    .unwrap_or(false);
-                match expr_kind {
-                    Some(&CExprKind::Literal(
-                        literal_cqual_type,
-                        CLiteral::String(ref bytes, element_size @ 1),
-                    )) if is_const && !translate_as_macro => {
-                        let target_ty = self.convert_type(target_cty.ctype)?;
-
-                        let bytes_padded = self.string_literal_bytes(
-                            literal_cqual_type.ctype,
-                            bytes,
-                            element_size,
-                        );
-                        let element_ty = mk().ident_ty("u8");
-                        let bytes_literal = mk().lit_expr(bytes_padded);
-                        let val = mk().cast_expr(bytes_literal, mk().ptr_ty(element_ty));
-                        let val = mk().cast_expr(val, target_ty);
-                        Ok(WithStmts::new_val(val))
-                    }
-                    _ => {
-                        // Variable length arrays are already represented as pointers.
-                        if let CTypeKind::VariableArray(..) = source_ty_kind {
-                            Ok(val)
-                        } else {
-                            let method = if is_const || ctx.is_static {
-                                "as_ptr"
-                            } else {
-                                "as_mut_ptr"
-                            };
-
-                            let call = val.map(|x| mk().method_call_expr(x, method, vec![]));
-
-                            // If the target pointee type is different from the source element type,
-                            // then we need to cast the ptr type as well.
-                            let call = match source_ty_kind.element_ty() {
-                                None => call,
-                                Some(source_element_ty) if source_element_ty == pointee.ctype => {
-                                    call
-                                }
-                                Some(_) => {
-                                    let target_ty = self.convert_type(target_cty.ctype)?;
-                                    call.map(|ptr| mk().cast_expr(ptr, target_ty))
-                                }
-                            };
-
-                            // Static arrays can now use as_ptr. Can also cast that const ptr to a
-                            // mutable pointer as we do here:
-                            if ctx.is_static && !is_const {
-                                return Ok(call.map(|val| {
-                                    let inferred_type = mk().infer_ty();
-                                    let ptr_type = mk().mutbl().ptr_ty(inferred_type);
-                                    mk().cast_expr(val, ptr_type)
-                                }));
-                            }
-
-                            Ok(call)
-                        }
-                    }
+                // Variable length arrays are already represented as pointers.
+                if let CTypeKind::VariableArray(..) = source_ty_kind {
+                    return Ok(val);
                 }
+
+                self.convert_address_of(ctx, expr, source_cty, target_cty, val, true)
             }
 
             CastKind::NullToPointer => {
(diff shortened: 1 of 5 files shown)


Rua:
Could this PR (and #1478) be reviewed soon? It's becoming rather a blocker for other things at the moment.

ahomescu:
#1478 seems fine to me, I approved it. For this one, you might need to add some reviewers (I'll ask tomorrow if anyone wants to volunteer).

Rua:
> For this one, you might need to add some reviewers (I'll ask tomorrow if anyone wants to volunteer).

I don't think regular users can add/request reviewers on this repo? Unless I'm missing something.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3085,13 +3093,8 @@ impl<'c> Translation<'c> {
...
         if is_static && !pointee.qualifiers.is_const {
-            let mut qtype = pointee;
-            qtype.qualifiers.is_const = true;
-            let ty_ = self
-                .type_converter
-                .borrow_mut()
-                .convert_pointer(&self.ast_context, qtype)?;
-            zero = mk().cast_expr(zero, ty_);
+            let ty_ = self.convert_pointee_type(pointee.ctype)?;
+            zero = mk().cast_expr(zero, mk().ptr_ty(ty_));
```
fw-immunant, line 3097: Should this still be translating the pointee type with `is_const` set to `true`?

c2rust-transpile/src/translator/operators.rs
```diff
@@ -1009,13 +1009,8 @@ impl<'c> Translation<'c> {
...
                             if let Mutability::Mutable = mutbl {
-                                let mut qtype = pointee_ty;
-                                qtype.qualifiers.is_const = true;
-                                let ty_ = self
-                                    .type_converter
-                                    .borrow_mut()
-                                    .convert_pointer(&self.ast_context, qtype)?;
-                                addr_of_arg = mk().cast_expr(addr_of_arg, ty_);
+                                let ty_ = self.convert_pointee_type(pointee_ty.ctype)?;
+                                addr_of_arg = mk().cast_expr(addr_of_arg, mk().ptr_ty(ty_));
```
fw-immunant: Same concern as the other hunk: should this still be translating the pointee type with `is_const` set to `true`?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4708,84 +4708,12 @@ impl<'c> Translation<'c> {
...
-                            // mutable pointer as we do here:
-                            if ctx.is_static && !is_const {
-                                return Ok(call.map(|val| {
-                                    let inferred_type = mk().infer_ty();
-                                    let ptr_type = mk().mutbl().ptr_ty(inferred_type);
-                                    mk().cast_expr(val, ptr_type)
-                                }));
-                            }
-
-                            Ok(call)
```
fw-immunant, line 4781: What happened to this hunk in the factoring? Is there a reason it's no longer necessary? It seems like `infer_ty` is no longer emitted, based on the change in snapshot output.

Ideally this kind of non-obvious change during a refactor would be noted in the commit message.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
```diff
@@ -370,7 +370,7 @@ pub static mut fns: fn_ptrs = {
     init
 };
 #[no_mangle]
-pub static mut p: *const fn_ptrs = unsafe { &fns as *const fn_ptrs };
```
fw-immunant, line 373: It's strange that this is unsafe, but that isn't changed by these commits. Just a note to self to look into this.

c2rust-transpile/src/translator/mod.rs
Rua: I think that question properly applies to the changes made in #1478? This PR depends on that other PR first, but I split them because #1479 also needs those changes.

c2rust-transpile/src/translator/mod.rs
Rua: The intermediate cast that's inserted there is now handled by setting `ref_cast_pointee_ty` if `needs_cast` is `true`:
https://github.com/immunant/c2rust/blob/5a55ba952ee03e18a01f4ec86477864de5241f82/c2rust-transpile/src/translator/mod.rs#L4950-L4958

The logic is that an intermediate cast is required if a final cast is also being done, because you can't cast a reference to just any pointer type; only to the corresponding pointer type (or to its element in the case of arrays). I agree that it's not exactly obvious that it's doing the same thing, the refactoring changes it around quite a bit.

I did remove `infer_ty` because that was the source of the problem in the two issues that this PR fixes (listed in the first post). So now there is always an explicit type given, which is generated by `convert_pointee_type`.

EDIT: I've added an extra line to the commit message now.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
Rua: It's borrowing a `static mut`, so I think it needs to be?

c2rust-transpile/src/translator/mod.rs
fw-immunant: Ok, thanks for the clarification.

c2rust-transpile/tests/snapshots/snapshots__transpile@macros.c.snap
fw-immunant: Ah, I didn't realize that it was `static mut`.

c2rust-transpile/src/translator/mod.rs
fw-immunant: It does indeed. I'll move this comment over there.

fw-immunant:
Modulo comments on #1478, which this includes as its first commit, LGTM.

PR #1447, opened 2025-11-07
transpile: Do not leak `main` argument strings

### LINKED ISSUE #649: Generated wrapper `main` leaks memory of `args`

Rua:
- Fixes #649.

I opted to store the final strings as `Vec<u8>` rather than as `CString`, because the latter does not have a way to get a mutable pointer/reference without leaking it. This is because of the nul termination invariant that `CString` must uphold.

### DIFF
--- a/c2rust-transpile/src/translator/main_function.rs
+++ b/c2rust-transpile/src/translator/main_function.rs
@@ -4,6 +4,7 @@
 //! Rust.
 
 use super::*;
+use c2rust_ast_builder::CaptureBy;
 use failure::format_err;
 use proc_macro2::{TokenStream, TokenTree};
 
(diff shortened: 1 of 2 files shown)


kkysen (commented):
Can you add a snapshot test for this, too?

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
```diff
@@ -0,0 +1,51 @@
...
+    return 0 as ::core::ffi::c_int;
+}
+pub fn main() {
+    let mut args_strings: Vec<Vec<u8>> = ::std::env::args()
+        .map(|arg| {
+            ::std::ffi::CString::new(arg)
+                .expect("Failed to convert argument into CString.")
+                .into_bytes_with_nul()
+        })
+        .collect();
```
kkysen, line 28: I think it's better to store a `Vec<CString>` here.  It's clearer what it's supposed to represent that way, and we can call `CString::raw` on it (`c_string.into_raw().cast::<c_char>()`).

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
```diff
@@ -0,0 +1,51 @@
...
+)]
+unsafe fn main_0(
+    mut argc: ::core::ffi::c_int,
+    mut argv: *mut *mut ::core::ffi::c_char,
+    mut envp: *mut *mut ::core::ffi::c_char,
+) -> ::core::ffi::c_int {
+    return 0 as ::core::ffi::c_int;
+}
+pub fn main() {
+    let mut args_strings: Vec<Vec<u8>> = ::std::env::args()
```
kkysen, line 22: Also, we can do this separately, but it's probably better to use `args_os()` here instead of `args()`, as otherwise we can only handle a limited UTF-8 subset of what the C program could.  One issue, though, is that we'd want to call [`OsString::into_encoded_bytes`](https://doc.rust-lang.org/std/ffi/struct.OsString.html#method.into_encoded_bytes), but that was only stabilized in Rust 1.74, and our generated nightly currently targets Rust 1.70.

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
Rua: But that would reintroduce the leak. I used `Vec<u8>` because it lets you get a mutable reference without leaking, which `CString` does not.

kkysen: How exactly?  I might've not been clear enough.  I'm suggesting we store `args_strings: Vec<CString>` and `args_ptrs: Vec<*mut c_char>`.  `CString` is owning just like `Vec<u8>` is.

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
Rua: I'd prefer doing that in a separate PR. Is it acceptable to use unstable features in c2rust?

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
Rua: But how do you get from `CString` to `*mut c_char`?

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
kkysen: Yeah, a separate PR for that would be good.  `c2rust-transpile` needs to build on `stable`, so we can't use unstable features there, but this would be an as-of-Rust 1.70 unstable (but as-of-Rust 1.74 stable) feature for the generated code.  Using unstable features in the generated code is okay, as we already do for a bunch of things.  This one is more minor, but we know it's been stabilized upstream already, so it should be fine.

c2rust-transpile/tests/snapshots/snapshots__transpile@main.c.snap
kkysen: Oh sorry, `.into_raw()` is wrong since it takes `self`.  But you can use `.as_ptr()` from `CStr` that takes `&self`.

Rua: That gives you an immutable pointer, and casting that to mutable could introduce UB if the user code writes to it.

kkysen: Hmm, that's true.  That's weird that `CString` doesn't have any `&mut self` APIs.  Storing `Vec<u8>`s makes sense then.

c2rust-transpile/src/translator/main_function.rs
```diff
@@ -53,45 +54,96 @@ impl<'c> Translation<'c> {
...
-                stmts.push(mk().semi_stmt(mk().for_expr(
-                    mk().ident_pat("arg"),
-                    mk().call_expr(args_fn, vec![]),
-                    mk().block(vec![mk().semi_stmt(mk().method_call_expr(
-                        mk().path_expr(vec!["args"]),
-                        "push",
-                        vec![mk().method_call_expr(
-                            mk().method_call_expr(
-                                mk().call_expr(
-                                    // TODO(kkysen) change `"std"` to `"alloc"` after `#![feature(alloc_c_string)]` is stabilized in `1.63.0`
```
kkysen, line 75: Can you keep this comment?

kkysen: Actually, I think we can switch to it now.  https://github.com/rust-lang/rust/pull/99277 was merged on 2022-07-16, before 2022-08-08.

c2rust-transpile/src/translator/main_function.rs
```diff
@@ -53,45 +54,96 @@ impl<'c> Translation<'c> {
...
+                            .path_segment_with_args(
+                                "Vec",
+                                mk().angle_bracketed_args(vec![mk().ident_ty("u8")]),
+                            )])]),
+                    )]);
+                    // |arg| {
+                    //     (::std::ffi::CString::new(arg))
+                    //         .expect("Failed to convert argument into CString.")
+                    //         .into_bytes_with_nul()
+                    // }
```
kkysen:
```suggestion
                    // map_arg = |arg| {
                    //     (::std::ffi::CString::new(arg))
                    //         .expect("Failed to convert argument into CString.")
                    //         .into_bytes_with_nul()
                    // };
```

c2rust-transpile/src/translator/main_function.rs
```diff
@@ -53,45 +54,96 @@ impl<'c> Translation<'c> {
...
+                        ReturnType::Default,
+                        mk().method_chain_expr(
+                            cstring_call,
+                            vec![
+                                (mk().path_segment("expect"), vec![expect_arg]),
+                                (mk().path_segment("into_bytes_with_nul"), vec![]),
+                            ],
+                        ),
+                    );
+                    // (args_fn)
```
kkysen:
```suggestion
                    // let args_strings = (args_fn)
```

c2rust-transpile/src/translator/main_function.rs
```diff
@@ -53,45 +54,96 @@ impl<'c> Translation<'c> {
...
+                        Movability::Movable,
+                        vec![mk().ident_pat("arg")],
+                        ReturnType::Default,
+                        mk().cast_expr(
+                            mk().method_call_expr(mk().ident_expr("arg"), "as_mut_ptr", vec![]),
+                            mk().mutbl()
+                                .ptr_ty(mk().abs_path_ty(vec!["core", "ffi", "c_char"])),
+                        ),
+                    );
+                    // ::core::iter::once(::core::ptr::null_mut())
```
kkysen:
```suggestion
                    // chain_arg = ::core::iter::once(::core::ptr::null_mut());
```

c2rust-transpile/src/translator/main_function.rs
```diff
@@ -53,45 +54,96 @@ impl<'c> Translation<'c> {
...
+
+                stmts.push(mk().local_stmt(Box::new({
+                    // Vec<*mut ::core::ffi::c_char>
+                    let ty = mk().path_ty(vec![mk().path_segment_with_args(
+                        "Vec",
+                        mk().angle_bracketed_args(vec![mk()
+                            .mutbl()
+                            .ptr_ty(mk().abs_path_ty(vec!["core", "ffi", "c_char"]))]),
+                    )]);
+                    // |arg| arg.as_mut_ptr() as *mut ::core::ffi::c_char
```
kkysen:
```suggestion
                    // map_arg = |arg| arg.as_mut_ptr() as *mut ::core::ffi::c_char;
```

c2rust-transpile/src/translator/main_function.rs
```diff
@@ -53,45 +54,96 @@ impl<'c> Translation<'c> {
...
-                )));
-                stmts.push(mk().semi_stmt(mk().method_call_expr(
-                    mk().path_expr(vec!["args"]),
-                    "push",
-                    vec![
-                        mk().call_expr(mk().abs_path_expr(vec!["core", "ptr", "null_mut"]), vec![]),
-                    ],
-                )));
+                    );
+                    // args_strings
```
kkysen:
```suggestion
                    // let args_ptrs = args_strings
```

c2rust-transpile/src/translator/main_function.rs
Rua: Changing `std` to `alloc` doesn't seem to work:
> failed to resolve: use of unresolved module or unlinked crate `alloc`

I guess that it's not as simple as just swapping out one for the other.

kkysen: Ah, I think we also need to add `extern crate alloc;`.  We can do that separately then, and just leave the comment here for now.

PR #1449, opened 2025-11-08
transpile: Avoid unnecessary parentheses around calls or fields

Rua:
This adds some exceptions to avoid unncessary parentheses in expressions of the form `a.b.c()`, `a().b()`, `a().b` or `a[b].c()`.


Rua:
In hindsight, I see that this could have been accomplished more straightforwardly by just giving these operators the same precedence level. Oh well.

PR #1450, opened 2025-11-10
refactor: fix `canonicalize_refs` in libxml transforms

oinoom:
`canonicalize_refs` used `cx.hir_map().node_to_hir_id` and unconditionally called `typeck(parent)`. When `MutVisitNodes` descends into attribute tokens or items without bodies (e.g., `extern` decls) the HIR map lacks a node or typeck tables, so the walk panicked with `span_bug!("can't type-check body …")`.

This PR swaps to `opt_node_to_hir_id` so attribute-only nodes short-circuit before we ever try to fetch HIR expressions, keeping the transform aligned with what rustc actually lowers, and checks `maybe_body_owned_by(parent)` before querying typeck tables; this skips extern items and other definitions without bodies, avoiding the `primary_body_of` unwrap panic while still running the transform over real bodies.

### DIFF
--- a/c2rust-refactor/src/transform/canonicalize_refs.rs
+++ b/c2rust-refactor/src/transform/canonicalize_refs.rs
@@ -15,9 +15,21 @@ struct CanonicalizeRefs;
 impl Transform for CanonicalizeRefs {
     fn transform(&self, krate: &mut Crate, _st: &CommandState, cx: &RefactorCtxt) {
         MutVisitNodes::visit(krate, |expr: &mut P<Expr>| {
-            let hir_id = cx.hir_map().node_to_hir_id(expr.id);
+            // `MutVisitNodes` walks into attributes even though rustc never lowers their bodies to HIR.
+            // We skip such nodes by probing with `opt_node_to_hir_id`; if we ever need an attribute-only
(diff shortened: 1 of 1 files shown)


c2rust-refactor/src/transform/canonicalize_refs.rs
```diff
@@ -15,9 +15,21 @@ struct CanonicalizeRefs;
 impl Transform for CanonicalizeRefs {
     fn transform(&self, krate: &mut Crate, _st: &CommandState, cx: &RefactorCtxt) {
         MutVisitNodes::visit(krate, |expr: &mut P<Expr>| {
-            let hir_id = cx.hir_map().node_to_hir_id(expr.id);
+            // `MutVisitNodes` walks into attributes even though rustc never lowers their bodies to HIR.
```
ahomescu, line 18: Are these expressions _inside attributes_? Otherwise how would we hit this for `P<Expr>`?

ahomescu (approved):

c2rust-refactor/src/transform/canonicalize_refs.rs
oinoom: Yeah, ones like `#[c2rust::src_loc = "28:1"]` have a rhs that's an expression

PR #1451, opened 2025-11-10
transpile: re-enable `--reorganize-definitions`

kkysen:
In contrast to #1414, which this supersedes, this doesn't add `c2rust-refactor` as a dependency of `c2rust`, which messes up `librustc_driver*.so` dynamic dependencies and makes `cargo install c2rust` not work by default (stable) anymore.


kkysen:
* Superseded by https://github.com/immunant/c2rust-testsuite/pull/32.

PR #1452, opened 2025-11-10
transpile: fixes before re-enabling `--reorganize-definitions`

kkysen:
Most of the changes from #1414 with #1451 split out.  In combination with #1451, supersedes #1414.

### DIFF
--- a/tests/items/src/test_fn_attrs.rs
+++ b/tests/items/src/test_fn_attrs.rs
@@ -28,42 +28,152 @@ pub fn test_fn_attrs() {
     // static __attribute__((always_inline)) void always_inline_static(void) {}
     // static __attribute__((__noinline__)) void noinline_static(void) {}
     // static void inline inline_static(void) {}
-    assert!(src.contains("#[inline(always)]\nunsafe extern \"C\" fn rust_always_inline_static"));
-    assert!(src.contains("#[inline(never)]\nunsafe extern \"C\" fn rust_noinline_static"));
-    assert!(src.contains("#[inline]\nunsafe extern \"C\" fn rust_inline_static"));
-    assert!(src.contains("#[inline]\nunsafe extern \"C\" fn rust_alt_kw_inline_static"));
-    assert!(src.contains("#[inline]\nunsafe extern \"C\" fn rust_gnu_inline_static"));
-    assert!(src.contains("#[cold]\nunsafe extern \"C\" fn rust_cold_used_attrs"));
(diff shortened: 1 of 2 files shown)


fw-immunant:
We should probably avoid entirely disabling the caching of zero values; I think I noticed c2rust-transpile taking significantly longer after that change. I believe the rest of these fixes were ready to merge.

fw-immunant:
Cleanups in the final commit look good.

kkysen:
> We should probably avoid entirely disabling the caching of zero values; I think I noticed c2rust-transpile taking significantly longer after that change. I believe the rest of these fixes were ready to merge.

I removed 828c6974b from this PR for now.  I'll fix that in its own PR, and I also split out the non-enable-by-default parts of `--reorganize-definitions` in #1454 so that it can merge without these fixes.  Is this PR good to go now?

fw-immunant:
> > We should probably avoid entirely disabling the caching of zero values; I think I noticed c2rust-transpile taking significantly longer after that change. I believe the rest of these fixes were ready to merge.
> 
> I removed 828c6974b from this PR for now. I'll fix that in its own PR, and I also split out the non-enable-by-default parts of `--reorganize-definitions` in #1454 so that it can merge without these fixes. Is this PR good to go now?

The only hesitation I have about merging this as-is is that the codepaths it alters are not tested by CI, because they only apply when `--reorganize-definitions` is passed to the transpiler.

kkysen:
They already passed in CI before, though.  And we have other refactorer code that isn't tested in CI yet, still a WIP.

tests/items/src/test_fn_attrs.rs
```diff
@@ -57,8 +63,7 @@ pub fn test_fn_attrs() {
 
     if cfg!(not(target_os = "macos")) {
         // aliased_fn is aliased to the inline_extern function
-        assert!(src.contains(
-            "extern \"C\" {\n    #[link_name = \"inline_extern\"]\n    fn aliased_fn();"
-        ));
+        let aliased_fn_syntax = |public| format!("extern \"C\" {{\n    #[link_name = \"inline_extern\"]\n    {}fn aliased_fn();", public);
```
ahomescu: Could we use raw string literals here (and below) to make this more readable?

kkysen: Yeah.  Fixed in d4b5e979f.  I used raw strings for all of them, not just the ones touched in this diff.

PR #1455, opened 2025-11-11
transpile: defer+cache imports alongside zero initializers

fw-immunant:
I think this is the right way to fix this. I tried adding a `WithStmtsAndImports` type instead, to avoid larger shared state, but it ended up infecting almost the entire transpiler.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -5369,19 +5369,13 @@ impl<'c> Translation<'c> {
     }
 
     fn import_type(&self, ctype: CTypeId) {
-        for Import {
-            decl_id,
-            ident_name,
-        } in self.imports_for_type(ctype)
-        {
-            self.add_import(decl_id, &ident_name)
-        }
+        self.add_imports(&self.imports_for_type(ctype))
     }
 
-    fn imports_for_type(&self, ctype: CTypeId) -> Vec<Import> {
+    fn imports_for_type(&self, ctype: CTypeId) -> IndexSet<Import> {
         use self::CTypeKind::*;
 
-        let mut imports = vec![];
+        let mut imports = IndexSet::default();
 
         let type_kind = &self.ast_context[ctype].kind;
         match type_kind {
@@ -5407,7 +5401,7 @@ impl<'c> Translation<'c> {
             | Reference(CQualTypeId { ctype, .. })
             | BlockPointer(CQualTypeId { ctype, .. })
             | TypeOf(ctype)
-            | Complex(ctype) => imports = self.imports_for_type(*ctype),
+            | Complex(ctype) => imports.extend(self.imports_for_type(*ctype)),
             Enum(decl_id) | Typedef(decl_id) | Union(decl_id) | Struct(decl_id) => {
                 let mut decl_id = *decl_id;
                 // if the `decl` has been "squashed", get the corresponding `decl_id`
@@ -5420,7 +5414,7 @@ impl<'c> Translation<'c> {
                     .borrow()
                     .resolve_decl_name(decl_id)
                     .expect("Expected decl name");
-                imports.push(Import {
+                imports.insert(Import {
                     decl_id,
                     ident_name,
                 });
@@ -5431,7 +5425,7 @@ impl<'c> Translation<'c> {
 
                 // Rust doesn't use void for return type, so skip
                 if *type_kind != Void {
-                    imports = self.imports_for_type(*ctype);
+                    imports.extend(self.imports_for_type(*ctype));
                 }
 
                 // Param Types


fw-immunant:
This is atop the "enable" branch, despite being a prerequisite fix, so that it gets test coverage in CI.

fw-immunant:
The testsuite is failing in a way that indicates this isn't sufficient to capture all required imports. We might need to handle import deferral nestably.

fw-immunant:
Ready for review now.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3476,13 +3482,11 @@ impl<'c> Translation<'c> {
...
-                        self.add_import(cur_file, decl_id, &rustname);
-                        // match decl {
-                        //     CDeclKind::Variable { is_defn: false, .. } => {}
-                        //     _ => self.add_import(cur_file, decl_id, &rustname),
-                        // }
-                    }
+                    self.add_import(decl_id, &rustname);
+                    // match decl {
+                    //     CDeclKind::Variable { is_defn: false, .. } => {}
+                    //     _ => self.add_import(cur_file, decl_id, &rustname),
```
```suggestion
                    //     _ => self.add_import(decl_id, &rustname),
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4993,15 +4994,29 @@ impl<'c> Translation<'c> {
...
             return Ok(init.clone());
-        }*/
+        }
+
+        let func_name = self
+            .function_context
+            .borrow()
+            .name
+            .clone()
+            .unwrap_or("<none>".to_owned());
```
```suggestion
        let func_name = &self.function_context.borrow().name;
        let func_name = func_name.as_deref().unwrap_or("<none>");
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4993,15 +4994,29 @@ impl<'c> Translation<'c> {
...
-        if let Some(file_id) = self.cur_file.get() {
-            self.import_type(type_id, file_id);
-        }
+        self.import_type(type_id);
+
+        let name = self
+            .type_converter
+            .borrow()
+            .resolve_decl_name(decl_id)
+            .unwrap_or("???".to_owned());
```
```suggestion
        let name = self
            .type_converter
            .borrow()
            .resolve_decl_name(decl_id);
        let name = name.as_deref().unwrap_or("???");
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -261,10 +267,11 @@ pub struct Translation<'c> {
     // Translation state and utilities
     type_converter: RefCell<TypeConverter>,
     renamer: RefCell<Renamer<CDeclId>>,
-    zero_inits: RefCell<IndexMap<CDeclId, WithStmts<Box<Expr>>>>,
+    zero_inits: RefCell<IndexMap<CDeclId, (WithStmts<Box<Expr>>, IndexSet<Import>)>>,
     function_context: RefCell<FuncContext>,
     potential_flexible_array_members: RefCell<IndexSet<CDeclId>>,
     macro_expansions: RefCell<IndexMap<CDeclId, Option<MacroExpansion>>>,
+    deferred_imports: RefCell<Vec<IndexSet<Import>>>,
```
kkysen, line 278: Could you add some docs on what the `Vec` means here?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5291,9 +5317,63 @@ impl<'c> Translation<'c> {
...
+        for Import {
+            decl_id,
+            ident_name,
+        } in imports
+        {
+            self.add_import(*decl_id, &ident_name)
+        }
+    }
+
+    /// Delay the addition of imports until `stop_deferring_imports` is called.
```
```suggestion
    /// Delay the addition of imports until [`Self::stop_deferring_imports`] is called.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5291,9 +5317,63 @@ impl<'c> Translation<'c> {
...
+    fn defer_imports(&self) {
+        self.deferred_imports.borrow_mut().push(Default::default())
+    }
+
+    /// Return whether imports are currently being deferred for later emission.
+    fn deferring_imports(&self) -> bool {
+        !self.deferred_imports.borrow().is_empty()
+    }
+
+    /// Apply and clear the last set of deferred imports. Returns the imports applied.
```
kkysen, line 5355:
```suggestion
    /// Apply and clear the last set of deferred imports. Returns the imports applied.
    /// Do nothing and return no imports if imports were not deferred.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5318,33 +5398,36 @@ impl<'c> Translation<'c> {
...
                 // Rust doesn't use void for return type, so skip
                 if *type_kind != Void {
-                    self.import_type(*ctype, decl_file_id);
+                    imports = self.imports_for_type(*ctype);
                 }
 
                 // Param Types
                 for param_id in params {
-                    self.import_type(param_id.ctype, decl_file_id);
+                    imports.extend(self.imports_for_type(param_id.ctype))
```
kkysen, line 5433: Why call `.extend` sometimes and `imports = ` other times?  It works now with the order, but seems brittle to rearranging/modifying things instead of just always doing `.extend` (`.append` would be better, but it's not in our older version of `indexmap`).

Also, I think it'd be more efficient and clearer if you just passed down an `imports: &mut IndexSet<Import>` in `fn imports_for_type` and called it at the top-level with `let mut imports = Default::default()` and `&mut imports`.

kkysen (commented):
The overall logic makes sense to me.  I had a question about the `fn import_for_type` implementation, though, and then a few nitpicks of other things.

kkysen:
Also, I force pushed to rebase since there was some issue with outdated CI that was changed on `master` already.

c2rust-transpile/src/translator/mod.rs
fw-immunant: Arguably, this should maybe panic; it's a programming error to mismatch calls to start/stop deferring imports. I'll change it and document that.

c2rust-transpile/src/translator/mod.rs
fw-immunant: This should probably be homogeneous (I'll switch to `.extend` everywhere), but I don't think efficiency is critical enough here to justify making the interface expose mutation.

fw-immunant: Done.

c2rust-transpile/src/translator/mod.rs
fw-immunant: Done.

c2rust-transpile/src/translator/mod.rs
fw-immunant: Done.

PR #1456, opened 2025-11-11
readme: update section on cross-transpilation to explain how to cross-arch transpile

kkysen:


README.md
```diff
@@ -291,8 +291,21 @@ or email us at [c2rust@immunant.com](mailto:c2rust@immunant.com).
...
+For example, on an `aarch64-linux-gnu` host, to cross-transpile to `x86_64-linux-gnu`,
+you can run
+
+```sh
+sudo apt install gcc-x86-64-linux-gnu # install cross-compiler, which comes with a sysroot
+c2rust transpile ${existing_args[@]} -- --target=x86_64-linux-gnu --sysroot=/usr/x86_64-linux-gnu
+```
+
+These extra args are passed to the `libclangTooling` that `c2rust-transpile` uses.
+You sometimes also need to pass extra headers, as sometimes headers are installed globally
```
ahomescu: Double "sometimes", the second one can be "occasionally" or another synonym.

README.md
```diff
@@ -291,8 +291,21 @@ or email us at [c2rust@immunant.com](mailto:c2rust@immunant.com).
 > I translated code on platform X, but it didn't work correctly on platform Y.
 
 We run the C preprocessor before translation to Rust.
-This specializes the code to the host platform.
-For this reason, we do not support cross compiling translated code at the moment.
+This specializes the code to the target platform (usually the host platform).
+We do, however, support cross-architecture transpilation with a different sysroot
+(cross-OS transpilation is more difficult because
+it can be difficult to get a sysroot for the target OS).
+For example, on an `aarch64-linux-gnu` host, to cross-transpile to `x86_64-linux-gnu`,
```
ahomescu, line 355: I'd flip these around, I think x86_64 hosts are more common than aarch64.

kkysen: This is what I tested, though, so I am certain this works.

kkysen: I can flip them, but I just haven't actually tested that way.

README.md
kkysen: Fixed in 0326dd8.

README.md
ahomescu: Okay, never mind then.

PR #1457, opened 2025-11-11
refactor: keep attr-only statements and cloned lets printable

oinoom:
Several transforms (`sink_lets`, `remove_redundant_let_types`) were panicking whenever the rewrite engine had to reparse snippets that no longer looked like full Rust statements.

Two cases showed up when testing: 
- while relocating `#[derive]` lines, the transform temporarily splices only the attribute text (for example just `#[derive(Copy, Clone)]`) while the enclosing struct body is rewritten at a higher level. Rust’s parser refuses that input, so `PrintParse<Stmt>` panicked before recovery finished
- newly cloned `let` bindings inherit the builder’s default `DUMMY_SP`, so pretty-printing emits an empty string even though the original code had a real initializer. The AST/text round-trip fails and subsequent rewrites abort

This PR:
- detects attribute-only snippets in `PrintParse<Stmt>` up front, reparses them by temporarily appending a dummy struct, and returns a `StmtKind::Empty` that carries the combined attribute span, which keeps the rewriter from panicking while the surrounding struct/item is printed elsewhere
- adds a helper to the AST builder that reuses the child node’s span whenever a synthesized statement would otherwise inherit `DUMMY_SP`. Newly cloned locals/exprs retain their original source ranges, so `rewrite_at_impl` can pull real snippets instead of an empty buffer

### DIFF
--- a/c2rust-refactor/src/ast_builder/builder.rs
+++ b/c2rust-refactor/src/ast_builder/builder.rs
@@ -360,6 +360,21 @@ pub struct Builder {
 
 #[allow(dead_code)]
 impl Builder {
+    /// Pick the span we should attach to a freshly minted statement.
+    ///
+    /// We want new statements produced by transforms such as `sink_lets` or
+    /// `fold_let_assign` to retain the source span of the code they were cloned
+    /// from so that later pretty-print/reparse passes can recover the original
+    /// text instead of seeing an empty buffer from a `DUMMY_SP`.
+    #[inline]
+    fn stmt_span(&self, fallback: Span) -> Span {
+        if self.span == DUMMY_SP {
+            fallback
+        } else {
+            self.span
+        }
+    }
+
     pub fn new() -> Builder {
         Builder {
             vis: Visibility {
@@ -1552,10 +1567,11 @@ impl Builder {
         L: Make<P<Local>>,
     {
         let local = local.make(&self);
+        let stmt_span = self.stmt_span(local.span);
         Stmt {
             id: self.id,
             kind: StmtKind::Local(local),
-            span: self.span,
+            span: stmt_span,
         }
     }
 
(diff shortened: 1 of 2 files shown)


fw-immunant (approved):
Seems reasonable. Is there an easy way to observe the change in behavior? It would be nice to have a test case if possible.

c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -675,6 +791,27 @@ where
     T: PrintParse + RecoverChildren + Splice + MaybeGetNodeId,
 {
     let printed = add_comments(new.to_string(), new, &rcx);
+    let mut printed = printed;
```
ahomescu: Why not add `mut` to the previous line?

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -360,6 +360,21 @@ pub struct Builder {
 
 #[allow(dead_code)]
 impl Builder {
+    /// Pick the span we should attach to a freshly minted statement.
+    ///
+    /// We want new statements produced by transforms such as `sink_lets` or
+    /// `fold_let_assign` to retain the source span of the code they were cloned
```
ahomescu, line 366: Could we just fix the transforms then? Why couldn't they pass in the spans?

ahomescu (commented):
Did you try this together with https://github.com/immunant/c2rust/pull/1433?

oinoom:
> Did you try this together with https://github.com/immunant/c2rust/pull/1433?

@ahomescu I didn't know this existed, should I have tried that?

oinoom:
@ahomescu  @fw-immunant  I'm trying a different approach in #1469 based on this comment. If that looks good, I'll merge it and close this one

> Could we just fix the transforms then? Why couldn't they pass in the spans?

PR #1458, opened 2025-11-11
transpile: Run `rustfmt` on generated .rs files

### LINKED ISSUE #742: Use `rustfmt` when translating
`rustfmt` used to be part of the refactoring tool but currently isn't part of refactoring _or_ translation. I believe that formatting with `rustfmt` should be part of the translation tool. Doing so will make reading the generated code nicer, makes it so the transpiler code doesn't ever need to care about formatting, and matches user expectations.

Rua:
- Fixes #742.

Rather than running `cargo fmt`, this runs the `rustfmt` binary directly so that it can be used on individual files, and also when there are no build files.

### DIFF
--- a/c2rust-transpile/src/build_files/mod.rs
+++ b/c2rust-transpile/src/build_files/mod.rs
@@ -226,7 +226,10 @@ fn emit_build_rs(
     let output = reg.render("build.rs", &json).unwrap();
     let output_path = build_dir.join("build.rs");
     let path = maybe_write_to_file(&output_path, output, tcfg.overwrite_existing)?;
-    rustfmt(&output_path, build_dir);
+
+    if !tcfg.disable_rustfmt {
+        rustfmt(&output_path, build_dir);
+    }
 
     Some(path)
 }
@@ -256,7 +259,10 @@ fn emit_lib_rs(
     let output_path = build_dir.join(file_name);
     let output = reg.render("lib.rs", &json).unwrap();
     let path = maybe_write_to_file(&output_path, output, tcfg.overwrite_existing)?;
-    rustfmt(&output_path, build_dir);
+
+    if !tcfg.disable_rustfmt {
+        rustfmt(&output_path, build_dir);
+    }
 
     Some(path)
 }
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -97,6 +97,7 @@ pub struct TranspilerConfig {
     pub output_dir: Option<PathBuf>,
     pub translate_const_macros: TranslateMacros,
     pub translate_fn_macros: TranslateMacros,
+    pub disable_rustfmt: bool,
     pub disable_refactoring: bool,
     pub preserve_unused_functions: bool,
     pub log_level: log::LevelFilter,
@@ -517,14 +518,18 @@ fn reorganize_definitions(
     }
 
     invoke_refactor(build_dir)?;
-    // fix the formatting of the output of `c2rust-refactor`
-    let status = Command::new("cargo")
-        .args(["fmt"])
-        .current_dir(build_dir)
-        .status()?;
-    if !status.success() {
-        warn!("cargo fmt failed, code may not be well-formatted");
+
+    if !tcfg.disable_rustfmt {
+        // fix the formatting of the output of `c2rust-refactor`
+        let status = Command::new("cargo")
+            .args(["fmt"])
+            .current_dir(build_dir)
+            .status()?;
+        if !status.success() {
+            warn!("cargo fmt failed, code may not be well-formatted");
+        }
     }
+
     Ok(())
 }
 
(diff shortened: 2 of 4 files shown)


c2rust-transpile/src/build_files/mod.rs
```diff
@@ -225,7 +225,10 @@ fn emit_build_rs(
     });
     let output = reg.render("build.rs", &json).unwrap();
     let output_path = build_dir.join("build.rs");
-    maybe_write_to_file(&output_path, output, tcfg.overwrite_existing)
+    let path = maybe_write_to_file(&output_path, output, tcfg.overwrite_existing)?;
+    rustfmt(&output_path, build_dir);
```
kkysen: We should only run this if `maybe_write_to_file` succeeded, since we don't want to change any files if `--overwrite-existing` isn't set.

c2rust-transpile/src/lib.rs
```diff
@@ -667,3 +669,15 @@ fn get_output_path(
...
+fn rustfmt(output_path: &Path, build_dir: &Path) {
+    let status = Command::new("rustfmt")
+        .args(["--edition", "2021"])
+        .arg(output_path)
+        .current_dir(build_dir)
+        .status();
+
+    if !status.map_or(false, |status| status.success()) {
+        warn!("rustfmt failed, code may not be well-formatted");
+    }
```
kkysen, line 713: Can we handle these errors differently?  The command will fail if `rustc` isn't found or something, and the status will fail if `rustc` exits with an error.  That said, they shouldn't really fail, so I think we can just panic here, maybe with an `.expect`.

Also, if we're calling `rustfmt` here, we can remove the `rustfmt` run in `snapshots.rs`.

c2rust-transpile/src/build_files/mod.rs
Rua: `maybe_write_to_file` returns `None` if it skipped out of overwriting the file, and the statement contains `?` which early returns in that case. So the code after it only runs if the writing succeeded.

c2rust-transpile/src/lib.rs
Rua: I don't think a panic is appropriate, as this is just an extra convenience for the user and not a hard requirement of the transpilation process. The transpilation should still succeed if `rustfmt` is not installed or doesn't complete for whatever reason. Though unlikely, `rustfmt` may not be installed even if `cargo` and `rustc` are.

c2rust-transpile/src/build_files/mod.rs
kkysen: Ah, oops, I missed that `?`.

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -252,8 +255,10 @@ fn emit_lib_rs(
 
     let output_path = build_dir.join(file_name);
     let output = reg.render("lib.rs", &json).unwrap();
+    let path = maybe_write_to_file(&output_path, output, tcfg.overwrite_existing)?;
+    rustfmt(&output_path, build_dir);
```
ahomescu: Could we make this optional (with a command line argument)?

kkysen: Why would we not do it, though?

ahomescu: The user might want a custom formatting configuration. They could add that to `rustfmt.toml`, but that creates a race between adding that file and running the transpiler.

Another reason is performance: what if the user wants to run the transpiler, a bunch of refactoring steps, _then_ run the formatter?

And what if the user just doesn't want to run it at all?

kkysen: > The user might want a custom formatting configuration. They could add that to `rustfmt.toml`, but that creates a race between adding that file and running the transpiler.

They can still run that after the transpiler (which would run `rustfmt`).

> Another reason is performance: what if the user wants to run the transpiler, a bunch of refactoring steps, _then_ run the formatter?

There are plenty of places in the transpiler we don't care about performance, and `rustfmt` is very fast.  It's never going to be the bottleneck compared to the rest of the transpiler.

> And what if the user just doesn't want to run it at all?

We already don't make any guarantees about the formatting of the output.

kkysen: I realized that `rustfmt` could occasionally have effects, like eating a comment, and it might interact with how comment insertion will be done, so leaving it as an on-by-default option is better actually.

Rua:
Does anything need to be changed here? I'm not sure what the final outcome of the discussions was.

ahomescu:
> Does anything need to be changed here? I'm not sure what the final outcome of the discussions was.

It would still be nice to have a command line argument to disable this (but on by default).

Rua:
Ok, done!

c2rust-transpile/src/build_files/mod.rs
ahomescu: LGTM, thanks!

kkysen (approved):

PR #1459, opened 2025-11-12
docs: update out-of-date info in `README.md`, `docs/`, and `manual/`

kkysen:
This updates the `README.md`, `docs/`, and `manual/` to

* re-add `c2rust refactor` now that it's been resurrected.
* remove known limitations that have now been fixed (inline assembly, C11 atomics, packed structs).
* remove references to things that have been removed (cross-checks, Docker, Azure CI, `rustfmt-preview`, Lua scripting).
* recommend `godbolt.org` instead of `c2rust.com`, as it's actually up-to-date and much nicer to use.
* recommend `bear` over `intercept-build`, as it's much more up-to-date and not Linux-only anymore.
* update installation instructions to use `--locked`, which we get a lot of issues about.
* update all the even older nightly references to `nightly-2022-08-08`, except for the generated `nightly-2023-04-15`, which is on purpose and up-to-date.
* update everything else to use `uv`, too (`test_translator.py` and `c2rust-testsuite/test.py` already were, our main test scripts).

It doesn't, however,

* update `c2rust.com/manual`, as I'm not sure how to do that.  But the `README.md` does now warn that `c2rust.com/manual` is currently very out-of-date and to refer to the in-tree sources instead.
* touch any of `manual/c2rust-refactor/doc/{ldoc.css, scripting_api.html}`.
* update most dependencies.  Many aren't needed or used anymore, or use the wrong versions, but determining that needs more CI testing (our main source of truth).

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,12 +50,21 @@ so refer to the in-tree [./manual/](./manual/) for more up-to-date instructions.
 ### Prerequisites
 
 C2Rust requires LLVM 7 or later with its corresponding clang compiler and libraries.
-Python 3.6 or later, CMake 3.5 or later, and openssl (1.0) are also required. These prerequisites may be installed with the following commands, depending on your platform:
+Python (through `uv`), CMake 3.5 or later and openssl (1.0) are also required.
+These prerequisites may be installed with the following commands, depending on your platform:
+
+Python:
+
+```sh
+curl -LsSf https://astral.sh/uv/install.sh | sh
+uv venv
+uv pip install -r scripts/requirements.txt
+```
 
 - **Ubuntu 18.04, Debian 10, and later:**
 
     ```sh
-    apt install build-essential llvm clang libclang-dev cmake libssl-dev pkg-config python3 git
+    apt install build-essential llvm clang libclang-dev cmake libssl-dev pkg-config git
     ```
 
 Depending on the LLVM distribution, the `llvm-dev` package may also be required.
@@ -64,7 +73,7 @@ For example, the official LLVM packages from [apt.llvm.org](https://apt.llvm.org
 - **Arch Linux:**
 
     ```sh
-    pacman -S base-devel llvm clang cmake openssl python
+    pacman -S base-devel llvm clang cmake openssl
     ```
 
 - **NixOS / nix:**
@@ -77,7 +86,7 @@ For example, the official LLVM packages from [apt.llvm.org](https://apt.llvm.org
 
     ```sh
     xcode-select --install
-    brew install llvm python3 cmake openssl
+    brew install llvm cmake openssl
     ```
 
 The C2Rust transpiler now builds using a stable Rust compiler.
@@ -302,7 +311,7 @@ is very similar, but not always as up-to-date and comprehensive as `bear`.
 but a standalone version can be easily installed via `pip` with:
 
 ```sh
-pip install scan-build
+uv tool install scan-build
 ```
 
 #### ... with `compiledb`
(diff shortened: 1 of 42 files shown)


README.md
```diff
@@ -185,12 +214,23 @@ c2rust transpile --binary myprog path/to/compile_commands.json
...
 These modules can be compiled together into a single static Rust library or binary.
 
+You can run with `--reorganize-definitions` (which invokes `c2rust-refactor`),
+which should deduplicate definitions and directly import them
+with `use`s instead of through the C API.
+
+The refactorer can also be run on its own to run other refactoring passes:
+
+```sh
+c2rust refactor $transform --cargo
```
ahomescu:
```suggestion
c2rust refactor --cargo $transform
```
Options usually come first, I'd change this to make it more like what people are used to.

README.md
```diff
@@ -225,46 +265,63 @@ file inside of `<build_dir>`.
...
 
-#### ... with `intercept-build`
+#### ... with `bear`
 
-`intercept-build` (part of the [scan-build tool](https://github.com/rizsotto/scan-build))
-is recommended for non-`cmake` projects.
-`intercept-build` is bundled with `clang` under `tools/scan-build-py`,
-but a standalone version can be easily installed via `pip` with:
+[`bear`](https://github.com/rizsotto/Bear) is recommended for projects whose build systems
+don't generated `compile_commands.json` automatically
```
ahomescu:
```suggestion
don't generate `compile_commands.json` automatically
```

scripts/provision_arch.sh
```diff
@@ -5,9 +5,11 @@ set -e
 SCRIPT_DIR="$(dirname "$0")"
 
 pacman -Sy
-pacman -S --quiet --noconfirm pkgconf make cmake ninja llvm clang python python-pip ncurses git diffutils luarocks libffi strace
+pacman -S --quiet --noconfirm pkgconf make cmake ninja llvm clang ncurses git diffutils luarocks libffi strace
 
-pip3 install --no-cache-dir --disable-pip-version-check -r $SCRIPT_DIR/requirements.txt
+curl -LsSf https://astral.sh/uv/install.sh | sh
```
ahomescu, line 10: It's in `extra` now so `pacman` should support it, do we still need the manual download? https://archlinux.org/packages/extra/x86_64/uv/

ahomescu (changes requested):

scripts/provision_arch.sh
kkysen: Ah, I didn't realize.  `uv` doesn't list `pacman` as a supported installation method (https://docs.astral.sh/uv/getting-started/installation/), though, so I didn't see.  But also, we don't even really use this script anymore, so I'd rather just keep things consistent (we install `uv` the same way everywhere).

README.md
kkysen: I had it this way because that's the order in https://github.com/immunant/c2rust-testsuite/blob/7345f356cbf4b5f1851f78cb6719aa212e398716/tests/templates.py#L70-L73, but this is more common for options in general, so I'll switch.

README.md
kkysen: Fixed in [`f311083..87eebdf`](https://github.com/immunant/c2rust/compare/f31108342256aa1aba8301732ba2bdc40f170eec..87eebdf80c9194ce25f21e05bd308c6c59639ed4).

README.md
kkysen: Fixed in [`f311083..87eebdf`](https://github.com/immunant/c2rust/compare/f31108342256aa1aba8301732ba2bdc40f170eec..87eebdf80c9194ce25f21e05bd308c6c59639ed4).

scripts/provision_arch.sh
ahomescu: Okay if all distros use the same method, then yeah we can keep it consistent.

PR #1462, opened 2025-11-18
Fix Rust compiler version mismatches in c2rust-refactor

ahomescu:


c2rust-analyze/tests/filecheck/alloc.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(extern_types)]
+#![feature(label_break_value)]
```
kkysen, line 2: I thought this was stabilized in Rust 1.65, which 2022-08-08 is already.

ahomescu: Discussed offline: nightly versions usually advertise the version _after_ them, not before.

ahomescu:
I updated it so it fails loudly (panics) if the argument list is empty again.

c2rust-analyze/tests/filecheck/alloc.rs
ahomescu: https://blog.rust-lang.org/2022/11/03/Rust-1.65.0/ was released a while after our nightly.

fw-immunant:
@ahomescu, should these commits end up on `master`? This only merged into your [ahomescu/fix_remove_refs](https://github.com/immunant/c2rust/tree/ahomescu/fix_remove_refs) branch.

ahomescu:
Ooops yeah I landed them too quickly.

PR #1463, opened 2025-11-18
Fix name collisions in reorganize_definitions

ahomescu:


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1100,6 +1108,22 @@ impl MovedDecl {
             },
         }
     }
+
+    fn ident_mut<'a>(&'a mut self) -> &'a mut Ident {
```
kkysen: Do you need the lifetime here?  I thought this is the default.

ahomescu: I was having some lifetime issues and thought this would fix it, but you're right that lifetime elision produces the same thing.

PR #1464, opened 2025-11-19
transpile: Add support for snapshot tests that are both arch and os specific

Rua:
I've added varargs.c as a snapshot test to ensure that it works correctly.


Rua:
Not sure where it is going wrong here. The Linux snapshot test seems to be looking for the snapshot file
```
c2rust-transpile/tests/snapshots/snapshots__transpile-linux-x86_64.snap
```

while it is named

```
c2rust-transpile/tests/snapshots/snapshots__transpile-linux-x86_64@varargs.c.snap
```

The file name of the C source is falling off the end for some reason?

EDIT: After some investigation I found that that's just how `insta` behaves when the directory only contains one file. Not sure what the best way forward is.

PR #1465, opened 2025-11-21
refactor: Fix name collisions in `reorganize_definitions`

ahomescu:
Rename identically-named identifier (appending a suffix) for identifiers with different definitions, e.g.
```rust

    pub const EINPROGRESS: ::core::ffi::c_int = 115;
    pub const EINPROGRESS: ::core::ffi::c_int = 115 as ::core::ffi::c_int;
```

from libxml2. The transform now renames the second definition to `EINPROGRESS_1` and updates all uses.

This is a copy of #1463 since GitHub closed that one after I reordered the branches.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -1,6 +1,6 @@
 use derive_more::From;
 use indexmap::IndexMap;
-use log::{debug, trace};
+use log::{debug, trace, warn};
 use smallvec::SmallVec;
 use std::collections::{HashMap, HashSet, hash_map::Entry};
 use std::mem;
@@ -18,7 +18,7 @@ use rustc_ast::ptr::P;
 use rustc_ast_pretty::pprust::{self, PrintState, item_to_string};
 use rustc_span::symbol::{Ident, kw};
 use rustc_data_structures::map_in_place::MapInPlace;
-use rustc_span::{BytePos, DUMMY_SP};
+use rustc_span::{BytePos, DUMMY_SP, Symbol};
 use smallvec::smallvec;
 
 use crate::ast_manip::util::{is_relative_path, join_visibility, namespace, split_uses, is_exported, is_c2rust_attr};
@@ -28,7 +28,6 @@ use crate::driver::Phase;
 use crate::{expect, match_or};
 use crate::path_edit::fold_resolved_paths_with_id;
 use crate::RefactorCtxt;
-use crate::util::Lone;
 use crate::ast_builder::mk;
 
 use super::externs;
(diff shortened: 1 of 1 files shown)


fw-immunant:
> Do we want to keep these as separate definitions when their values are the same? Is it better to try to fix this from the transpiler side?

I don't have a preference on whether we merge definitions like these, but separately it would be nice to improve this on the transpiler side, because its root cause is the same as why our handling of `switch` cases emits integers rather than uses of `const`s when cases reference `#define`'d values. But that fix could be a bit tricky; we'd need some special handling to keep types in agreement everywhere.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -553,10 +552,19 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
         idents.map(|idents| {
-            for (ident, items) in idents.into_iter() {
-                for item in items {
+            for items in idents.into_values() {
+                for (idx, mut item) in items.into_iter().enumerate() {
+                    if idx > 0 {
+                        let ident = item.ident();
+                        // Append a number suffix to this item if
+                        // there are multiple items with the same name
+                        let name = format!("{0}_{idx}", ident.name.as_str());
```
kkysen: I don't think the `{0}` syntax is used very often.

```suggestion
                        let name = format!("{}_{idx}", ident.name.as_str());
```

```suggestion
                        let name = ident.name.as_str();
                        let name = format!("{name}_{idx}");
```

ahomescu: https://github.com/rust-lang/rust-clippy/pull/15287 I was thinking of this recent change from 1.89 when I wrote it like this, but that's a much newer rustc/clippy than what we use. Your second suggestion looks clearer, I'll use that.

ahomescu:
I didn't touch the transpiler because afaict the two values of `EINPROGRESS` in the example come in as completely different clang AST nodes. If you want to unify the two implementations, go for it, but that does feel like a more significant refactoring to me. Meanwhile, I think the current refactoring fix is also worthwhile to land (otherwise we might hit the problem in other places as well).

kkysen:
Maybe there should be a warning about this? Is it at all possible to check if the actual values are the same and merge them in that case?  Or maybe run remove redundant casts first.

ahomescu:
> Maybe there should be a warning about this?

I could add a warning, sure.

> Is it at all possible to check if the actual values are the same and merge them in that case?

We would basically need to run a compile-time evaluator for that, and that's if the expressions are even constant in the first place.

> Or maybe run remove redundant casts first.

Sure, but that only fixes the subset where casts are the problem. What about e.g. comparing `42` with `42 + 0`?

kkysen:
> > Or maybe run remove redundant casts first.
> 
> Sure, but that only fixes the subset where casts are the problem. What about e.g. comparing `42` with `42 + 0`?

Yeah, we definitely need the general fix as done here, but I think the majority of these that the transpiler produces will be from casts.  So we should do that, too (run remove redundant casts first).

kkysen (approved):

ahomescu:
I added a warning.

PR #1467, opened 2025-11-21
refactor: Fix `remove_unnecessary_refs` to ignore pointers

ahomescu:
Change the behavior of remove_unnecessary_refs to not remove dereferences of pointers because those are generally required.

### DIFF
--- a/c2rust-refactor/src/transform/canonicalize_refs.rs
+++ b/c2rust-refactor/src/transform/canonicalize_refs.rs
@@ -1,6 +1,7 @@
 use rustc_middle::ty::adjustment::{Adjust, AutoBorrow, AutoBorrowMutability};
 use rustc_ast::{Crate, Expr, ExprKind, Mutability, UnOp};
 use rustc_ast::ptr::P;
+use rustc_type_ir::sty;
 
 use crate::ast_builder::mk;
 use crate::ast_manip::MutVisitNodes;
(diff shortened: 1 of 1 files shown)


fw-immunant (commented):
Code changes look reasonable, but is there an example to look at to better understand the motivation for this change? This refactoring pass runs in CI for the testsuite--are there cases that were broken there and now fixed?

ahomescu:
> Code changes look reasonable, but is there an example to look at to better understand the motivation for this change? This refactoring pass runs in CI for the testsuite--are there cases that were broken there and now fixed?

That's right, we have `remove_unnecesary_refs` enabled in CI now and there are a bunch of places where it failed, e.g.
```
   --> src/xzlib.rs:263:17
    |
263 |         *have = (have).wrapping_add(ret as ::core::ffi::c_uint);
    |         -----   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `*mut u32`
    |         |
    |         expected due to the type of this binding
    |
    = note:     expected type `u32`
            found raw pointer `*mut u32`

error[E0599]: no method named `expect` found for raw pointer `*mut Option<unsafe extern "C" fn(*mut c_void, *const i8, ...)>` in the current scope
     --> src/xpath.rs:18203:18
      |
18202 | /             (crate::src::globals::__xmlGenericError())
18203 | |                 .expect(
      | |                 -^^^^^^ method not found in `*mut Option<unsafe extern "C" fn(*mut c_void, *const i8, ...)>`
      | |_________________|
```
and more. 

_I think_ that `&*x` is always fine to elide when `x` is a reference, but never when it's a pointer (correct me if I missed a case).

I had a separate patch for the test suite to disable that transform, but might as well fix it instead.

fw-immunant (approved):

PR #1469, opened 2025-11-21
refactor: preserve spans for rewrites and add empty-print fallback to avoid sink_lets and remove_redundant_let_types

oinoom:
pprust can emit `""` for statements containing `DUMMY_NODE_ID` locals. When the rewrite pipeline reparses that text, it yields zero statements and trips the `lone()` assertion, crashing sink_lets and remove_redundant_let_types.

- those transforms now set the new statement’s span to the original local’s span so source recovery has a real location instead of `DUMMY_SP`
- `rewrite_at_impl` adds a fallback: if pprust pretty-printing returns an empty string (seen for `DUMMY_NODE_ID` locals), it pulls the original source via `span_to_snippet` before reparsing
- span preservation was also added to related rewrites (fold_let_assign and remove_redundant_let_types) so replacements keep usable spans

### DIFF
--- a/c2rust-refactor/src/rewrite/strategy/print.rs
+++ b/c2rust-refactor/src/rewrite/strategy/print.rs
@@ -678,15 +678,11 @@ where
 
     // Fallback: extract source snippet when pretty-printing produces empty output.
     //
-    // The rustc pretty-printer (pprust) produces empty output for AST nodes with
-    // DUMMY_NODE_ID. This commonly occurs when transforms like sink_lets create new
-    // statement wrappers with DUMMY_NODE_ID to distinguish transformed nodes from
-    // originals (see vars.rs:79). Without this fallback, empty pretty-printer output
-    // causes parsing failures (0 statements instead of 1), crashing the rewrite system.
-    //
-    // The fallback extracts source text directly from the node's span using
-    // span_to_snippet(). This requires that statements have valid spans set, which
-    // is why both span preservation and this fallback mechanism are necessary.
+    // pprust can emit "" for Local statements whose inner Local has DUMMY_NODE_ID
+    // (sink_lets sets that intentionally). Reparse of "" yields 0 statements and
+    // hits the lone() assertion. The fallback replaces the empty string with the
+    // original source via span_to_snippet(), assuming the statement carries a real
+    // span (set in vars.rs).
     if printed.trim().is_empty() {
         if let Ok(snippet) = rcx
             .session()


c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -252,7 +252,23 @@ impl Splice for Ty {
...
+        // extend_span_attrs() expands the span backward to include attributes that
+        // appear before the node in source, checking syntax contexts to avoid crossing
+        // macro boundaries.
+        match &self.kind {
+            StmtKind::Local(local) => extend_span_attrs(self.span, &local.attrs),
+            StmtKind::Item(item) => extend_span_attrs(self.span, &item.attrs),
+            StmtKind::Expr(expr) | StmtKind::Semi(expr) => extend_span_attrs(self.span, &expr.attrs),
+            StmtKind::MacCall(mac) => extend_span_attrs(self.span, &mac.attrs),
+            StmtKind::Empty => self.span,
+        }
```
kkysen:
```suggestion
        let attrs = match &self.kind {
            StmtKind::Local(local) => &local.attrs[..],
            StmtKind::Item(item) => &item.attrs[..],
            StmtKind::Expr(expr) | StmtKind::Semi(expr) => &expr.attrs[..],
            StmtKind::MacCall(mac) => &mac.attrs[..],
            StmtKind::Empty => &[],
        };
        extend_span_attrs(self.span, attrs)
```

c2rust-refactor/src/transform/vars.rs
```diff
@@ -415,7 +426,21 @@ impl Transform for FoldLetAssign {
...
+                    //   x = expr;       <- assignment
+                    // into:
+                    //   let x = expr;   <- combined local
+                    //
+                    // The combined Local keeps l.span from the original declaration. Setting
+                    // stmt.span = l.span allows the rewrite system to extract source text using
+                    // the Splice trait. Without this, stmt.span would be DUMMY_SP and source
+                    // extraction would fail.
+                    let stmt_span = l.span;
+                    curs.replace(move |_| mk().span(stmt_span).local_stmt(l));
```
kkysen:
```suggestion
                    curs.replace(move |_| mk().span(l.span).local_stmt(l));
```

kkysen: Also, nit, but `l` is a really confusing variable name, although that was already there before.

c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -674,7 +691,29 @@ fn rewrite_at_impl<T>(old_span: Span, new: &T, mut rcx: RewriteCtxtRef) -> bool
...
+
+    // Fallback: extract source snippet when pretty-printing produces empty output.
+    //
+    // When transforms move macro-expanded code (e.g., from #[derive]), the span may
+    // point to generated code with no source file location. Pretty-printing such nodes
+    // can produce empty output, which causes problems during the reparse/recovery phase.
+    //
+    // If the printed text is empty, try extracting the source directly using
+    // splice_span() (which includes attributes). If source is available, use that
+    // for reparsing instead of the empty pretty-printed text.
```
ahomescu: Could this reinsert back code that the transforms decided to delete?

c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -252,7 +252,24 @@ impl Splice for Ty {
 
 impl Splice for Stmt {
     fn splice_span(&self) -> Span {
-        self.span
+        // Extend statement span to include attributes on the inner node.
+        //
+        // Transforms like sink_lets set stmt.span = local.span, but attributes are
```
ahomescu: You're fixing `sink_lets` below, is this also needed?

oinoom: My understanding is that the attributes need to sink as well, without this the extracted source code span would just be the let statement but with this we cover the attribute as well and so we'll get a match to the pretty printed code (which includes the attribute)

ahomescu: We create a bunch of new `mk()...local_stmt(...)` below, couldn't we copy the attributes there?

oinoom: I'm not exactly sure how that would be done, can you put attributes inside a statement? I thought that was contained at a higher level.

ahomescu: You can put attributes _on_ a statement, wouldn't that work here?

PR #1471, opened 2025-11-24
Fix type error in generated Rust when indexing typedef'd pointers

fw-immunant:
This happens when we index a pointer to one portable type and the result is treated as a different one. In C they're compatible without Clang even including an implicit cast node in the AST, but when we translate to distinct types such as `u64` and `usize`, the generated Rust will contain a type error. We can avoid this by simply adding a cast if necessary.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3957,7 +3957,15 @@ impl<'c> Translation<'c> {
                                 };
 
                             let mul = self.compute_size_of_expr(pointee_type_id.ctype);
-                            Ok(pointer_offset(lhs, rhs, mul, false, true))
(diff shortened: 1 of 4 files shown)


fw-immunant:
> LGTM! Can you add a comment saying where this test is from, though?

The commit message mentions the origin, but since the reduced version is so trivial I'm not sure how much it matters to attribute the test itself.

spernsteiner:
As the origin is non-public, it might be better not to mention it in detail in the public git history

kkysen:
> As the origin is non-public, it might be better not to mention it in detail in the public git history

Is it a secret or just not public?  Also why isn't it not public?  I hadn't even realized.

fw-immunant:
The code is publicly available, but its inclusion in a particular corpus isn't public information. I'll refer to only public info in the commit message.

PR #1472, opened 2025-11-24
Fix Rust compiler version mismatches in c2rust-refactor, rebased

ahomescu:
New PR copying https://github.com/immunant/c2rust/pull/1462 because it landed in the wrong branch.


kkysen:
I think this broke CI: https://github.com/immunant/c2rust/actions/runs/19646285650/job/56262171768 (I guess maybe we should require passing CI before merging lol).

Also, `test.py` took 2-3x times as long as normal.

ahomescu:
> Also, test.py took 2-3x times as long as normal.

This is working as intended, it was doing zero refactoring before.

kkysen:
> This is working as intended, it was doing zero refactoring before.

Ugh, the refactorer is really slow then.  That's a much longer CI time.  More waiting I guess.

PR #1475, opened 2025-11-25
refactor: Implement deep structural comparison of ADTs by comparing all variants individually

### LINKED ISSUE #266: `c2rust transpile` produces `gettimeofday` call with wrong argument type when reorganizing definitions
Steps to reproduce:

- `git clone https://github.com/ZirconiumX/Dorpsgek`
- `intercept-build make`
- `c2rust transpile --binary main --reorganize-definitions compile_commands.json`

Expected result:

Code compiles cleanly

Actual result:

```rust
   --> time.rs:115:68
    |
115 |     ::libc::gettimeofday(&mut t as *mut _ as *mut ::libc::timeval, 0 as *mut libc::c_void);
    |                                                                    ^^^^^^^^^^^^^^^^^^^^^^ expected enum `libc::timezone`, found enum `libc::c_void`
    |
    = note: expected raw pointer `*mut libc::timezone`
               found raw pointer `*mut libc::c_void`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
thread 'main' panicked at 'Could not rebuild crate: could not compile `c2rust_out`.
```

Something to note is that this translates fine without `--reorganize-definitions`.

ahomescu:
* Fixes #266.

Fixes one set of issues with `curl` (edit: and `nginx`) caused by a fake equivalence between `core::ffi::c_void` and `libc::timezone`:

```rust
    --> src/src/tool_util.rs:15:70
     |
15   |     ::libc::gettimeofday(&mut now as *mut _ as *mut ::libc::timeval, crate::__stddef_null_h::NULL);
     |     -------------------- arguments to this function are incorrect    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `*mut timezone`, found `*mut c_void`
     |
     = note: expected raw pointer `*mut timezone`
                found raw pointer `*mut c_void`
note: function defined here
    --> /home/ahomescu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.177/src/unix/linux_like/linux/gnu/mod.rs:1126:12
     |
1126 |     pub fn gettimeofday(tp: *mut crate::timeval, tz: *mut crate::timezone) -> c_int;
     |            ^^^^^^^^^^^^
Som
```

### DIFF
--- a/c2rust-refactor/src/context.rs
+++ b/c2rust-refactor/src/context.rs
@@ -1370,6 +1370,9 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
                     return true;
                 }
 
+                // Iterate separately over the variants and fields (as opposed to the `.all_fields()` flat map).
+                // Otherwise, we may find a false equivalence for empty variants,
+                // or, in general, the flattened iterator could have the same elements but different variant boundaries.
                 def1.variants().len() == def2.variants().len()
                     && def1
                         .variants()


ahomescu:
This fixes https://github.com/immunant/c2rust/issues/266, we should close that bug when this lands.

c2rust-refactor/src/context.rs
```diff
@@ -1317,15 +1321,21 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
                     return true;
                 }
 
-                def1.all_fields().count() == def2.all_fields().count()
+                def1.variants().len() == def2.variants().len()
```
kkysen, line 1376: `.all_fields()` does a flat map over `.variants()` and then `.fields()`, which is what we're doing here. Is the issue that the flattened iterator may have the same elements, but the boundaries of each variant are different?

ahomescu: That's right, plus that it doesn't account correctly for empty variants. For example, it considers the following two enums equivalent:
```rust
enum Foo {
  One,
}
enum Bar {
  One,
  Two,
}
```

kkysen: Maybe add a comment explaining some of this?  Not sure it's worth it, though.  Up to you.

tests/integration/tests/nginx/conf.yml
kkysen, line 1: Oh, does `curl`'s `conf.yml` need updating, too?

kkysen: No, nevermind, there's still other bugs there.

c2rust-refactor/src/context.rs
kkysen: Added in b1e71ca.

PR #1478, opened 2025-11-26
transpile: Split `convert_pointer` into two functions for easier reuse

Rua:
A small refactoring that is needed as common dependency for other PRs.

I've added a `convert_pointee_type` to `Translation` as well, to ensure that the pointee type gets imported when necessary.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3071,6 +3071,14 @@ impl<'c> Translation<'c> {
             .convert(&self.ast_context, type_id)
     }
 
+    fn convert_pointee_type(&self, type_id: CTypeId) -> TranslationResult<Box<Type>> {
+        self.import_type(type_id);
+
+        self.type_converter
+            .borrow_mut()
+            .convert_pointee(&self.ast_context, type_id)
+    }
+
     /// Construct an expression for a NULL at any type, including forward declarations,
     /// function pointers, and normal pointers.
     fn null_ptr(&self, type_id: CTypeId, is_static: bool) -> TranslationResult<Box<Expr>> {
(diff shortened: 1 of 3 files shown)


ahomescu (approved):

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3085,13 +3093,8 @@ impl<'c> Translation<'c> {
...
         if is_static && !pointee.qualifiers.is_const {
-            let mut qtype = pointee;
-            qtype.qualifiers.is_const = true;
-            let ty_ = self
-                .type_converter
-                .borrow_mut()
-                .convert_pointer(&self.ast_context, qtype)?;
-            zero = mk().cast_expr(zero, ty_);
+            let ty_ = self.convert_pointee_type(pointee.ctype)?;
+            zero = mk().cast_expr(zero, mk().ptr_ty(ty_));
```
fw-immunant, line 3097: Should this still be translating the pointee type with `is_const` set to `true`?

(moved from [other PR](https://github.com/immunant/c2rust/pull/1446#discussion_r2582671135))

c2rust-transpile/src/translator/operators.rs
```diff
@@ -1009,13 +1009,8 @@ impl<'c> Translation<'c> {
...
                             if let Mutability::Mutable = mutbl {
-                                let mut qtype = pointee_ty;
-                                qtype.qualifiers.is_const = true;
-                                let ty_ = self
-                                    .type_converter
-                                    .borrow_mut()
-                                    .convert_pointer(&self.ast_context, qtype)?;
-                                addr_of_arg = mk().cast_expr(addr_of_arg, ty_);
+                                let ty_ = self.convert_pointee_type(pointee_ty.ctype)?;
+                                addr_of_arg = mk().cast_expr(addr_of_arg, mk().ptr_ty(ty_));
```
fw-immunant, line 1013: Same concern as the other hunk: should this still be translating the pointee type with `is_const` set to `true`?

(also moved from [other PR](https://github.com/immunant/c2rust/pull/1446#discussion_r2582671662))

fw-immunant (commented):
One question inline, otherwise looks good.

c2rust-transpile/src/translator/mod.rs
Rua: `mk()` creates an immutable pointer, `mk().mutbl()` would create a mutable pointer. The `convert_pointee` function only translates the `CTypeId`, so only T in `*const T`, not the `const` part of it which is part of `CQualTypeId`.

fw-immunant: Ah, that makes sense. Thanks for the explanation.

fw-immunant (approved):
LGTM! Thanks for being patient with the review.

PR #1479, opened 2025-11-26
transpile: Translate null pointers with `ptr::null[_mut]`

### LINKED ISSUE #202: Use `std::ptr::null_mut::<T>()` instead of `0 as *mut T`
c2rust current transpiles `NULL` pointers as `0 as *mut T`, but it would be a bit cleaner and more rust-y to instead produce `std::ptr::null_mut::<T>()`.

Rua:
* Depends on #1478.
* Fixes #202.

This is more idiomatic, and is also necessary for supporting strict(er) provenance in the future.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@empty_init.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@empty_init.c.snap
@@ -19,7 +19,7 @@ pub struct Scope {
 #[no_mangle]
 pub static mut scope: *mut Scope = &{
     let mut init = Scope {
-        next: 0 as *const Scope as *mut Scope,
+        next: ::core::ptr::null::<Scope>() as *mut Scope,
     };
     init
 } as *const Scope as *mut Scope;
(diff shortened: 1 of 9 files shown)


Rua:
The testsuite is crashing with a stack overflow. Since all this PR does is replace a cast with a function call, is that particular test nesting calls so deeply that a few extra is making it overrun?

ahomescu:
This has happened before in the lua test, I made https://github.com/immunant/c2rust-testsuite/pull/37 to compile it in release mode so it doesn't use as much stack.

c2rust-transpile/tests/snapshots/snapshots__transpile@empty_init.c.snap
```diff
@@ -19,7 +19,7 @@ pub struct Scope {
 #[no_mangle]
 pub static mut scope: *mut Scope = &{
     let mut init = Scope {
-        next: 0 as *const Scope as *mut Scope,
+        next: ::core::ptr::null::<Scope>() as *mut Scope,
```
kkysen, line 22: We probably want to fix how we're generating this so that we just use `null_mut` directly, but this is fine for now, and still technically correct.

kkysen: Oh wait, I see why it's happening now.  Nevermind.  Nice fix.

kkysen (approved):
LGTM!  Thanks for this; it's very nice to have a provenance correct translation, and also `miri` won't complain about this stuff anymore if we're looking for other more serious issues.

PR #1481, opened 2025-11-28
transpile: Resolve expression first in address-of and array-decay

### LINKED ISSUE #1480: `__FUNCTION__` macro is mistranslated
Transpiling this code

```C
int main() {
    const char* normal = "foobar";
    const char* function = __FUNCTION__;
}
```

gives
```rust
    let mut normal: *const ::core::ffi::c_char = b"foobar\0" as *const u8
        as *const ::core::ffi::c_char;
    let mut function: *const ::core::ffi::c_char = ::core::mem::transmute::<
        [u8; 5],
        [::core::ffi::c_char; 5],
    >(*b"main\0")
        .as_ptr();
    return 0;
}
```

The `__FUNCTION__` macro should not produce a `transmute` call but simply cast a Rust string, as in the `normal` case. Looking at the AST, the difference appears to be that the latter wraps the string in an additional `Predefined` expression, which is missed because the code for address-of operator and array decay only checks for `CExprKind::Literal` directly.

Rua:
- Fixes #1480.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -451,6 +451,26 @@ impl TypedAstContext {
         }
     }
 
+    /// Returns the expression inside any number of nested parentheses.
+    pub fn resolve_parens(&self, mut expr_id: CExprId) -> CExprId {
+        while let CExprKind::Paren(_, subexpr) = self.index(expr_id).kind {
+            expr_id = subexpr;
+        }
+
+        expr_id
+    }
+
+    /// Unwraps a predefined expression, if there is one.
+    pub fn unwrap_predefined_ident(&self, mut expr_id: CExprId) -> CExprId {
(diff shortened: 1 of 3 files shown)


fw-immunant:
In the third commit, is it necessary to recursively remove parentheses as well as looking through `Predefined`? If not, I would rather see this code simply non-recursively check for the presence of a `Predefined` wrapper, because `Predefined` is never nested and only ever wraps string literals. In that case the function would also be better named something that mentions that it unwraps `Predefined` nodes, e.g. `unwrap_predefined_ident`.

Rua:
> In the third commit, is it necessary to recursively remove parentheses

If the C code is `((__func__))` then certainly?

fw-immunant:
Needs rebase due to conflicts, but should be fine to merge once it gets one.

Rua:
Done, thanks!

PR #1484, opened 2025-12-02
refactor: Update `reorganize_definitions` test and fix some issues

ahomescu:
* Fix reorganize_definitions test: the test (and probably all the refactoring tests) has bitrotted; get it working again
* Add repro of statvfs issue from python2: reproduce the issue encountered by the python2 test on Ubuntu 22.04 where the `statvfs` structure has a private field that we cannot use
* Check visibility when comparing structural equivalence for function signatures
* Deep match visibility when comparing function signatures: updates to the algorithm we use to match equivalent types and function signatures.
* Enable more transforms on the python2 integration test.

When matching between two types to determine whether they can be unified, we now account for differences in visibility because importing a structure with private types from an external crate (e.g. `libc`) makes them unusable to us when trying to construct new values of those types. The motivating example is `statvfs`:

```rust
    pub struct statvfs {
        pub f_bsize: c_ulong,
        pub f_frsize: c_ulong,
        pub f_blocks: crate::fsblkcnt_t,
        pub f_bfree: crate::fsblkcnt_t,
        pub f_bavail: crate::fsblkcnt_t,
        pub f_files: crate::fsfilcnt_t,
        pub f_ffree: crate::fsfilcnt_t,
        pub f_favail: crate::fsfilcnt_t,
        pub f_fsid: c_ulong,
        pub f_flag: c_ulong,
        pub f_namemax: c_ulong,
        __f_spare: [c_int; 6],
    }
```

### DIFF
--- a/c2rust-refactor/src/context.rs
+++ b/c2rust-refactor/src/context.rs
@@ -790,7 +790,7 @@ impl<'a, 'tcx> RefactorCtxt<'a, 'tcx> {
 
     /// Compare two items for type compatibility under the C definition.
     ///
-    /// If `match_vis` is `true`, the visibility of all structure/enum/union
+    /// If `match_vis` is `true`, the visibility of all `struct`/`enum`/`union`
     /// fields (i.e. if they are `pub`) must match between the two items.
     pub fn compatible_types(&self, item1: &Item, item2: &Item, match_vis: bool) -> bool {
         TypeCompare::new(self).compatible_types(item1, item2, match_vis)
@@ -1067,7 +1067,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
 
     /// Compare two items for type compatibility under the C definition.
     ///
-    /// If `match_vis` is `true`, the visibility of all structure/enum/union
+    /// If `match_vis` is `true`, the visibility of all `struct`/`enum`/`union`
     /// fields (i.e. if they are `pub`) must match between the two items.
     pub fn compatible_types(&self, item1: &Item, item2: &Item, match_vis: bool) -> bool {
         use rustc_ast::ItemKind::*;
@@ -1273,7 +1273,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
 
     /// Compare two AST types for structural equivalence, ignoring names.
     ///
-    /// If `match_vis` is `true`, the visibility of all structure/enum/union
+    /// If `match_vis` is `true`, the visibility of all `struct`/`enum`/`union`
     /// fields (i.e. if they are `pub`) must match between the two types.
     fn structural_eq_ast_tys(
         &self,


c2rust-refactor/tests/reorganize_definitions/new.rs
kkysen, line 1: Is this a snapshot?

kkysen: How do I run the test?

c2rust-refactor/src/context.rs
```diff
@@ -1072,6 +1072,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
                     self.cx.opt_node_type(item1.id),
                     self.cx.opt_node_type(item2.id),
                 ) {
+                    (Some(ty1), Some(ty2)) if match_vis => self.structural_eq_tys_with_vis(ty1, ty2),
                     (Some(ty1), Some(ty2)) => self.structural_eq_tys(ty1, ty2),
```
kkysen: nit:

I think it'd be a bit clearer if we just used an `if` `else` here:

```suggestion
                    (Some(ty1), Some(ty2)) => if match_vis {
                        self.structural_eq_tys_with_vis(ty1, ty2)
                    } else {
                        self.structural_eq_tys(ty1, ty2)
                    }
```

Same for the one other place it's done, too.

c2rust-refactor/src/context.rs
```diff
@@ -1139,11 +1140,18 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
...
                     _ => {
                         let mut fields = variant1.fields().iter().zip(variant2.fields().iter());
                         fields.all(|(field1, field2)| {
-                            self.structural_eq_ast_tys(&field1.ty, &field2.ty)
+                            // TODO: either Visibility or VisibilityKind should implement
+                            // PartialEq; until then, the closest we have is `is_pub`
+                            if match_vis && field1.vis.kind.is_pub() != field2.vis.kind.is_pub() {
+                                return false;
+                            }
```
kkysen, line 1172: Can you add docs to the function saying `match_vis` compares with field `pub`ness?  Also, is this just `struct` fields?

c2rust-refactor/src/context.rs
```diff
@@ -1139,11 +1140,18 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
                     self.cx.opt_node_type(item1.id),
                     self.cx.opt_node_type(item2.id),
                 ) {
+                    (Some(ty1), Some(ty2)) if match_vis => self.structural_eq_tys_with_vis(ty1, ty2),
                     (Some(ty1), Some(ty2)) => self.structural_eq_tys(ty1, ty2),
                     _ => {
                         let mut fields = variant1.fields().iter().zip(variant2.fields().iter());
```
kkysen, line 1166: Isn't this not accounting for one variant having more fields than the other?  `.zip` just truncates to the shorter iterator.

c2rust-refactor/tests/reorganize_definitions/new.rs
ahomescu: Go into `c2rust-refactor/tests`, run `run-test.sh reorganize_definitions`. I think we should start adding these tests back to CI, but I wasn't sure if it goes into this PR or do another one.

c2rust-refactor/src/context.rs
ahomescu: It should be all of struct/enum/union, but enums might be buggy.

c2rust-refactor/tests/reorganize_definitions/new.rs
kkysen: It'd be nice to have all the testing just in Rust instead of scattered shell scripts.

ahomescu: I'd like to move all the refactorer tests into `tests/refactor` at the top level, and then we can change the testing there. We should do that separately though.

> It'd be nice to have all the testing just in Rust instead of scattered shell scripts.

I'm not sure how this is supposed to work in this case. The inputs are Rust files that we feed into the refactorer, are you saying we should call that tool from Rust `#[test]` code? I'm not sure how that's simpler or what it buys us over shell scripts.

kkysen: > I'd like to move all the refactorer tests into `tests/refactor` at the top level, and then we can change the testing there. We should do that separately though.

I don't care too much, but our transpiler snapshot tests are under `c2rust-transpile/`, so the way it is now would be the same.

> I'm not sure how that's simpler or what it buys us over shell scripts.

We just run `cargo test` and cargo figures out what we need to build. For example the current shell scripts have a lot of complexity over setting up paths and run only debug mode, whereas with cargo it just does the right thing.

ahomescu: Alright, we can do that but I'd prefer making a separate PR for those (probably larger) changes.

kkysen: Yep, that's totally fine.

c2rust-refactor/src/context.rs
```diff
@@ -1062,8 +1065,11 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
         }
     }
 
-    /// Compare two items for type compatibility under the C definition
-    pub fn compatible_types(&self, item1: &Item, item2: &Item) -> bool {
+    /// Compare two items for type compatibility under the C definition.
+    ///
+    /// If `match_vis` is `true`, the visibility of all structure/enum/union
```
kkysen:
```suggestion
    /// If `match_vis` is `true`, the visibility of all struct/enum/union
```

c2rust-refactor/src/context.rs
```diff
@@ -1223,24 +1261,35 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
...
         let out_ty1 = sig1.output();
         let out_ty2 = sig2.output();
-        self.structural_eq_tys(out_ty1, out_ty2)
+        self.structural_eq_tys_with_vis(out_ty1, out_ty2)
     }
 
     /// Compare two AST types for structural equivalence, ignoring names.
-    fn structural_eq_ast_tys(&self, ty1: &rustc_ast::Ty, ty2: &rustc_ast::Ty) -> bool {
+    ///
+    /// If `match_vis` is `true`, the visibility of all structure/enum/union
```
kkysen:
```suggestion
    /// If `match_vis` is `true`, the visibility of all struct/enum/union
```

c2rust-refactor/src/context.rs
```diff
@@ -1139,11 +1151,27 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
...
+                        if match_vis {
+                            self.structural_eq_tys_with_vis(ty1, ty2)
+                        } else {
+                            self.structural_eq_tys(ty1, ty2)
+                        }
+                    }
+                    _ => 'match_fields: {
+                        if variant1.fields().len() != variant2.fields().len() {
+                            break 'match_fields false;
+                        }
```
kkysen, line 1164: Oh clever!

c2rust-refactor/src/context.rs
kkysen: Fixed in af68648.

c2rust-refactor/src/context.rs
kkysen: Fixed in af68648.

PR #1485, opened 2025-12-02
Add the testsuite as a subtree of c2rust

ahomescu:
Run `git subtree` to copy the entire c2rust-testsuite repository into a subdirectory of c2rust (with full history).

### DIFF
--- a/tests/unit/asm.aarch64/Cargo.toml
+++ b/tests/unit/asm.aarch64/Cargo.toml
@@ -4,4 +4,4 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-c2rust-asm-casts = { path = "../../c2rust-asm-casts", version = "0.21.0" }
+c2rust-asm-casts = { path = "../../../c2rust-asm-casts", version = "0.21.0" }
--- a/tests/unit/asm.x86_64/Cargo.toml
+++ b/tests/unit/asm.x86_64/Cargo.toml
@@ -4,4 +4,4 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-c2rust-asm-casts = { path = "../../c2rust-asm-casts", version = "0.21.0" }
+c2rust-asm-casts = { path = "../../../c2rust-asm-casts", version = "0.21.0" }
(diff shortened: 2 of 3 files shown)


brk:
FWIW [`git-subrepo`](https://github.com/ingydotnet/git-subrepo) might be a useful tool to implement the subtree approach. The main difference is that whereas `git-subtree` keeps the subtree's history as commits within the core repository itself, `git-subrepo` keeps them in the cloned trees. 

So, that is, with `git-subtree`, when cloning `c2rust` from GitHub one gets a cloned repository with a bunch of (possibly squashed) commits from `c2rust-testsuite` which are not descendants of `c2rust`'s root commit, plus a set of merge commits.

With `git-subrepo`, the historical commits of `c2rust-testsuite` are only in the cloned repositories. A new clone of `c2rust` from GitHub will get only the in-tree contents, not the historical commits from the `c2rust-testuite` repo.

Hope the explanation makes sense!

There are pros and cons to each approach, but it's good to know that there is a choice available.

ahomescu:
If we land this, all open PRs from https://github.com/immunant/c2rust-testsuite/pulls would have to be copied over to this repository, or landed there and copied with `git subtree`.

ahomescu:
@brk Does `git-subrepo` ship with git, or would it have to be manually installed? Adding an extra dependency to just download c2rust could be a concern.

brk:
@ahomescu `git-subrepo` does not ship with git, but it also would not need to be installed by `c2rust`'s users. Only the subset of  maintainers who want to push or pull commits between the two repositories would need it.

ahomescu:
> Only the subset of maintainers who want to push or pull commits between the two repositories would need it.

That would be the other reviewers on this PR, I think this needs a group discussion.

randomPoison:
`git subtree` is a one-time operation here right? We merge this then deprecate the c2rust-testsuite repo? It sounds like `git-subrepo` is a thing that'd we'd have to keep using going forward, which seems less convenient than just doing a one time merge of the repos.

It seems like `git-subrepo` adds some useful functionality if you want to keep the repos separate, but afaict we have no reason to want that.

thedataking:
Discussed with Andrei offline: could put the testsuite under `tests/integration` and eventually move the current contents under `tests/` into something like `tests/unit`.

PR #1486, opened 2025-12-03
Enable more refactor transforms on lua

ahomescu:
* Moving https://github.com/immunant/c2rust-testsuite/pull/35


kkysen (approved):
I approved it before already, and it's passing CI here now (it wasn't before).

PR #1487, opened 2025-12-03
Add index to name of refactor diff files to keep track of order of application

ahomescu:
* Moving https://github.com/immunant/c2rust-testsuite/pull/34.

The order which we apply the transforms in is relevant, so prepend an index to the diff name. This follows the style of git patches: `0000_first_patch.diff`, `0001_second_patch.diff`, etc...

### DIFF
--- a/tests/integration/tests/templates.py
+++ b/tests/integration/tests/templates.py
@@ -63,9 +63,16 @@ ARTIFACT_DIR="${ARTIFACT_ROOT}/${PROJECT_NAME}"
 rm -rf "$ARTIFACT_DIR"
 mkdir -p "$ARTIFACT_DIR"
 
+diff_idx=0
 while IFS= read -r transform; do
     [[ -z "$transform" ]] && continue
-    DIFF_FILE="$ARTIFACT_DIR/${transform}.diff"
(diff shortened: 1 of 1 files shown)


tests/integration/tests/templates.py
```diff
@@ -63,9 +63,16 @@
...
 mkdir -p "$ARTIFACT_DIR"
 
+diff_idx=0
 while IFS= read -r transform; do
     [[ -z "$transform" ]] && continue
-    DIFF_FILE="$ARTIFACT_DIR/${transform}.diff"
+
+    # Print the index in 4 digit format with leading zeroes,
+    # so the indices align just like in git outputs
+    diff_idx4=`printf "%04d" "${diff_idx}"`
```
kkysen:
```suggestion
    diff_idx4=$(printf "%04d" "${diff_idx}")
```

ahomescu:
> LGTM, but can you use `$(cmd)` instead of `` `cmd` ``? Also, the diff files have an empty line at the top. Though that might not be from this.

The empty line is probably from c2rust-refactor, not sure what to do about that.

PR #1488, opened 2025-12-03
test.py: only run `*.gen.sh` scripts actually generated during that run

kkysen:
* Moving https://github.com/immunant/c2rust-testsuite/pull/29.

Otherwise, a previously generated `*.gen.sh` file will still be run even if it's no longer supposed to be generated (and isn't anymore).


ahomescu (approved):

tests/integration/tests/__init__.py
```diff
@@ -25,12 +26,15 @@ class Test(object):
         "check": ["check.sh", "test.sh"]
     }
 
-    def __init__(self, directory: str):
-        ff = next(os.walk(directory))[2]
-        self.scripts = set(filter(lambda f: f.endswith(".sh"), ff))
-        self.dir = directory
-        self.conf_file = os.path.join(directory, CONF_YML)
-        self.name = os.path.basename(directory)
+    def __init__(self, directory: Path, generated_scripts: set[Path]):
```
ahomescu, line 27: I think the error you're getting is because you're getting all the scripts in `generated_scripts` here, not just the ones in the current `directory`.

kkysen: Yup that was it.

PR #1495, opened 2025-12-06
transpile: Split off new pointers.rs module

Rua:
Working with pointer-related stuff, I found that related code was kinda scattered all over, making it difficult to find at times. This creates a new pointers.rs module, moves some existing functions there, and creates new functions to wrap some other bits of code. I've tried to not make too many semantic changes.

In `convert_assignment_operator_with_rhs` there was a bit of dead code that I removed. Since `AssignAdd` and `AssignSubtract` are handled by the match arms, the `if` that checks for them below a second time wasn't doing anything (and it did the same thing anyway).

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -513,60 +513,46 @@ impl<'c> Translation<'c> {
                         }
 
                         // Everything else
-                        AssignAdd if pointer_lhs.is_some() => {
-                            let mul = self.compute_size_of_expr(pointer_lhs.unwrap().ctype);
-                            let ptr = pointer_offset(write.clone(), rhs, mul, false, false);
-                            WithStmts::new_unsafe_val(mk().assign_expr(write, ptr))
-                        }
-                        AssignSubtract if pointer_lhs.is_some() => {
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/translator/operators.rs
```diff
@@ -513,60 +513,46 @@ impl<'c> Translation<'c> {
...
-                            {
-                                let mul = self.compute_size_of_expr(pointer_lhs.ctype);
-                                let ptr = pointer_offset(
-                                    write.clone(),
-                                    rhs,
-                                    mul,
-                                    op == AssignSubtract,
-                                    false,
-                                );
-                                WithStmts::new_unsafe_val(mk().assign_expr(write, ptr))
```
kkysen, line 539: This was totally redundant/dead code, right?

kkysen (approved):

c2rust-transpile/src/translator/operators.rs
Rua: Yes.

kkysen:
I just merged #1479 so I think this has a conflict now, but I can merge it after that's resolved.

PR #1497, opened 2025-12-06
transpile: Strip tail return only when transpiling a function

### LINKED ISSUE #1506: transpile: `return` is stripped from end of statement expressions
Transpiling the following:
```C
#include <stdio.h>

void foo() {
    int var = 9001;

    ({
        var = 42;
        return;
    });

    printf("I should be unreachable!");
}
```
results in
```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
#![feature(label_break_value)]
extern "C" {
}
#[no_mangle]
pub unsafe extern "C" fn foo() {
    let mut var: ::core::ffi::c_int = 9001 as ::core::ffi::c_int;
    var = 42 as ::core::ffi::c_int;
    printf(b"I should be unreachable!\0" as *const u8 as *const ::core::ffi::c_char);
}
```

The Rust code is missing the `return`, causing the code below to be executed when it shouldn't be. I've tracked down the cause of this to the code here:
https://github.com/immunant/c2rust/blob/ca313fd66e947b87c72d45dbe941a9d737d51474/c2rust-transpile/src/translator/mod.rs#L4260-L4273

`convert_function_body` calls `convert_cfg` with `true` passed for the `cut_out_trailing_ret` argument. But it should be passing `false` for statement expressions, only `true` for actual functions.

Rua:
- Fixes #1506.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -16,11 +16,11 @@ use proc_macro2::{Punct, Spacing::*, Span, TokenStream, TokenTree};
 use syn::spanned::Spanned as _;
 use syn::{
     AttrStyle, BareVariadic, Block, Expr, ExprBinary, ExprBlock, ExprBreak, ExprCast, ExprField,
-    ExprIndex, ExprParen, ExprUnary, FnArg, ForeignItem, ForeignItemFn, ForeignItemMacro,
-    ForeignItemStatic, ForeignItemType, Ident, Item, ItemConst, ItemEnum, ItemExternCrate, ItemFn,
-    ItemForeignMod, ItemImpl, ItemMacro, ItemMod, ItemStatic, ItemStruct, ItemTrait,
-    ItemTraitAlias, ItemType, ItemUnion, ItemUse, Lit, MacroDelimiter, PathSegment, ReturnType,
-    Stmt, Type, TypeTuple, UseTree, Visibility,
+    ExprIndex, ExprParen, ExprReturn, ExprUnary, FnArg, ForeignItem, ForeignItemFn,
+    ForeignItemMacro, ForeignItemStatic, ForeignItemType, Ident, Item, ItemConst, ItemEnum,
+    ItemExternCrate, ItemFn, ItemForeignMod, ItemImpl, ItemMacro, ItemMod, ItemStatic, ItemStruct,
+    ItemTrait, ItemTraitAlias, ItemType, ItemUnion, ItemUse, Lit, MacroDelimiter, PathSegment,
+    ReturnType, Stmt, Type, TypeTuple, UseTree, Visibility,
 };
 use syn::{BinOp, UnOp}; // To override `c_ast::{BinOp,UnOp}` from glob import.
 
@@ -2424,7 +2424,8 @@ impl<'c> Translation<'c> {
                     _ => panic!("function body expects to be a compound statement"),
                 };
                 let mut converted_body =
-                    self.convert_function_body(ctx, name, body_ids, return_type, ret)?;
+                    self.convert_block_with_scope(ctx, name, body_ids, return_type, ret)?;
+                strip_tail_return(&mut converted_body);
 
                 // If `alloca` was used in the function body, include a variable to hold the
                 // allocations.
@@ -2544,7 +2545,6 @@ impl<'c> Translation<'c> {
         graph: cfg::Cfg<cfg::Label, cfg::StmtOrDecl>,
         store: cfg::DeclStmtStore,
         live_in: IndexSet<CDeclId>,
-        cut_out_trailing_ret: bool,
     ) -> TranslationResult<Vec<Stmt>> {
         if self.tcfg.dump_function_cfgs {
             graph
@@ -2606,12 +2606,11 @@ impl<'c> Translation<'c> {
             &mut self.comment_store.borrow_mut(),
             current_block,
             self.tcfg.debug_relooper_labels,
-            cut_out_trailing_ret,
         )?);
         Ok(stmts)
     }
 
-    fn convert_function_body(
+    fn convert_block_with_scope(
         &self,
         ctx: ExprContext,
         name: &str,
@@ -2622,7 +2621,7 @@ impl<'c> Translation<'c> {
         // Function body scope
         self.with_scope(|| {
             let (graph, store) = cfg::Cfg::from_stmts(self, ctx, body_ids, ret, ret_ty)?;
-            self.convert_cfg(name, graph, store, IndexSet::new(), true)
+            self.convert_cfg(name, graph, store, IndexSet::new())
         })
     }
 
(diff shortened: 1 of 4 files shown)


Rua:
I ran into an unexpected snag with this one: the `stmts_block` helper function is adding back the semicolon if it's missing. Ideally I would just remove that behaviour, but that has the obvious potential to break a lot of things, so does anyone know why it's there and what other things rely on it?

Rua:
I was able to remove the semicolon-adding code from `stmts_block` without any adverse effects in the tests, so let's hope it wasn't needed anywhere else. I think this is ready for review now.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5269,7 +5265,70 @@ impl<'c> Translation<'c> {
...
+                *old_expr = *expr;
+            } else {
+                unreachable!()
+            }
+        }
+
+        // Simple blocks, recurse down
+        Expr::Async(ExprAsync { block, .. })
+        | Expr::Block(ExprBlock { block, .. })
+        // TODO: | Expr::Const(ExprConst { block, .. })
```
kkysen: What happens here?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5269,7 +5265,70 @@ impl<'c> Translation<'c> {
...
+            then_branch,
+            else_branch,
+            ..
+        }) => {
+            strip_tail_return(&mut then_branch.stmts);
+
+            if let Some((_, else_branch)) = else_branch {
+                strip_tail_return_expr(else_branch);
+            }
+        }
```
kkysen: If there's only a `then_branch`, then the types might be different between the two branches, right?  Like this:

```rust
if condition {
    return "hello";
}
```

I guess it can't be the last statement in a function then?  I'm not sure if there are still edge cases in sub-expressions, though.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5269,7 +5265,70 @@ impl<'c> Translation<'c> {
...
+
+        // Recurse down all arms of the `match`
+        Expr::Match(ExprMatch { arms, .. }) => {
+            for Arm { body, .. } in arms {
+                strip_tail_return_expr(body);
+            }
+        }
+
+        _ => (),
```
kkysen: What about other expressions?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5269,7 +5265,70 @@ impl<'c> Translation<'c> {
...
+            // Replace return + expression with the expression.
+            if let Expr::Return(ExprReturn {
+                expr: Some(expr), ..
+            }) = temp
+            {
+                *old_expr = *expr;
+            } else {
+                unreachable!()
+            }
+        }
```
kkysen: Could you combine the two branches?  Since there's a `_ => {}` at the end, if you accidentally miss a case here it won't be caught by the compiler.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -5269,7 +5265,70 @@ impl<'c> Translation<'c> {
...
+fn strip_tail_return_expr(expr: &mut Expr) {
+    match expr {
+        old_expr @ Expr::Return(ExprReturn { expr: None, .. }) => {
+            // Replace standalone return with ()
+            *old_expr = *mk().tuple_expr(vec![]);
+        }
+
+        old_expr @ Expr::Return(ExprReturn { expr: Some(_), .. }) => {
+            // placeholder value to allow swapping
+            let temp = std::mem::replace(old_expr, Expr::Verbatim(Default::default()));
```
kkysen: nit: `use std::mem` and `mem::replace`.

c2rust-transpile/src/translator/mod.rs
kkysen: Actually, could you also move this comment outside this pattern?  For some reason it's breaking `rustfmt` so it's not formatting this function anymore.

Rua: If the last statement is a block, then it recurses into that block to do the conversion on the last `return` in that block, if there is one.

c2rust-transpile/src/translator/mod.rs
Rua: If the function returns a value, then there must be an `else_branch`, but if it returns `void` then there doesn't have to be:
```C
    if condition {
        // stuff
        return;
    }
} // end of function
```

c2rust-transpile/src/translator/mod.rs
Rua: This only covers expression types that meet a specific condition: the final value of the subexpression should determine the value of the whole. For most expression types that is not the case; for example the last expression in a `loop` doesn't become the value of the `loop` statement itself.

c2rust-transpile/src/translator/mod.rs
kkysen: I mean why is the `Expr::Const` TODOed out?

c2rust-transpile/src/translator/mod.rs
kkysen: What if the loop has a `break` for example?

Also, could you add that explanation as a comment?

c2rust-transpile/src/translator/mod.rs
kkysen: Ahh.  Could you add that as a comment, too?

c2rust-transpile/src/translator/mod.rs
Rua: Ah! Because the version of `syn` that's being used in c2rust currently doesn't have support for inline `const` blocks yet.

c2rust-transpile/src/translator/mod.rs
Rua: All that this code does is replace `return` with its expression, where it would not change the meaning. I don't think you can ever replace a `return` inside a `loop` without changing the meaning (I think it would just make the code invalid).

Rua:
In order to solve #1506 I had to add another variant to the `ImplicitReturnType` enum, because with `Void` it was adding a `return` that doesn't belong there. I don't know if that's actually the correct solution here. I don't quite understand the label break stuff in the compound statements translation, and there is likely some interaction with #1509 as well.

Rua:
I don't think this PR will affect that one, but we'll have to see I suppose.

Rua:
@ahomescu the test here is failing since merging #1544. It's inserting a label and a block, seemingly for no reason. Is that a bug in your code or in mine?

EDIT: Seems to occur with current master as well. I've updated the test so that it includes the additional block.

ahomescu:
> @ahomescu the test here is failing since merging #1544. It's inserting a label and a block, seemingly for no reason. Is that a bug in your code or in mine?
> 
> EDIT: Seems to occur with current master as well. I've updated the test so that it includes the additional block.

That's not surprising, #1544 changed a few details about how we emit the expression in a `WithStmts`. We should fix here whatever conflicts that caused.

ahomescu:
There is also a clippy lint for this https://rust-lang.github.io/rust-clippy/master/index.html#needless_return. Could we just run that instead?

Rua:
Can clippy run even when no Cargo.toml has been generated?

Rua:
Ok, I've removed the last commit from this PR, so that it only deals with fixing #1506. The rest can be handled by a future PR that adds a clippy run?

Rua:
I stashed it just in case.

PR #1500, opened 2025-12-08
refactor: unset `$RUSTC_WRAPPER` so it doesn't add an extra `rustc` arg

kkysen:
If `$RUSTC_WRAPPER` is set, `rustc` invocations will be add an extra `"rustc"` arg, which trips up the arg parsing code with an unexpected arg.


fw-immunant:
Out of curiosity, in which situations is `RUSTC_WRAPPER` set?

PR #1501, opened 2025-12-08
Windows building support

kant2002:
This is draft PR to check if there is appetite for that Current problems which I would work on
- I should examine what libc variants has simply different signatures, I fail to recognize that, and ignore too much probably.
- libclangAstExporter currently failed to build with error
```
 unsupported or unknown VisualStudio version: 18.0
  if another version is installed consider running the appropriate vcvars script before building this crate
  ', .....cargo\registry\src\github.com-1ecc6299db9ec823\cmake-0.1.50\src\lib.rs:955:2
```
- `patch` executable should be on PATH. I think you may find it in `C:\Program Files\Git\usr\bin\` on most dev machines.

That probably indication that cmake create should be updated, but that's whole can of worm.

Would be glad to recive feedback on the general direction

### DIFF
--- a/c2rust-build-paths/src/lib.rs
+++ b/c2rust-build-paths/src/lib.rs
@@ -91,13 +91,12 @@ pub fn find_llvm_config() -> Option<PathBuf> {
                         "llvm-config"
                     });
                     let canonicalized_dir = lib_dir.canonicalize();
-                    if canonicalized_dir.is_err() {
+                    lib_dir.canonicalize().unwrap_or_else(|_| {
                         panic!(
-                            "LLVM_LIB_DIR is set but `{}' does not exist",
+                            "LLVM_LIB_DIR is set but `{}` does not exist",
                             lib_dir.display()
                         );
-                    }
-                    canonicalized_dir.unwrap()
+                    })
                 })
         })
         .or_else(|| {


kant2002:
I think the second error is due to cmake missing support for VS 2026 https://github.com/rust-lang/cmake-rs/pull/255

c2rust-build-paths/src/lib.rs
```diff
@@ -84,8 +84,21 @@ pub fn find_llvm_config() -> Option<PathBuf> {
...
                 .map(|mut lib_dir| {
-                    lib_dir.push("../bin/llvm-config");
-                    lib_dir.canonicalize().unwrap()
+                    lib_dir.push("..");
+                    lib_dir.push("bin");
+                    if cfg!(target_os = "windows") {
+                        lib_dir.push("llvm-config.exe");
+                    } else {
+                        lib_dir.push("llvm-config");
+                    }
```
kkysen:
```suggestion
                    lib_dir.push(if cfg!(target_os = "windows") {
                        "llvm-config.exe"
                    } else {
                        "llvm-config"
                    });
```

c2rust-build-paths/src/lib.rs
```diff
@@ -84,8 +84,21 @@ pub fn find_llvm_config() -> Option<PathBuf> {
             env::var_os("LLVM_LIB_DIR")
                 .map(PathBuf::from)
                 .map(|mut lib_dir| {
-                    lib_dir.push("../bin/llvm-config");
-                    lib_dir.canonicalize().unwrap()
+                    lib_dir.push("..");
+                    lib_dir.push("bin");
```
kkysen:
```suggestion
                    lib_dir.push("../bin");
```

c2rust-build-paths/src/lib.rs
```diff
@@ -84,8 +84,21 @@ pub fn find_llvm_config() -> Option<PathBuf> {
...
+                    if cfg!(target_os = "windows") {
+                        lib_dir.push("llvm-config.exe");
+                    } else {
+                        lib_dir.push("llvm-config");
+                    }
+                    let canonicalized_dir = lib_dir.canonicalize();
+                    if canonicalized_dir.is_err() {
+                        panic!(
+                            "LLVM_LIB_DIR is set but `{}' does not exist",
```
kkysen:
```suggestion
                            "LLVM_LIB_DIR is set but `{}` does not exist",
```

c2rust-build-paths/src/lib.rs
```diff
@@ -84,8 +84,21 @@ pub fn find_llvm_config() -> Option<PathBuf> {
...
+                            lib_dir.display()
+                        );
+                    }
+                    canonicalized_dir.unwrap()
```
kkysen:
```suggestion
                    let canonicalized_dir = lib_dir.canonicalize();
                    lib_dir.canonicalize().unwrap_or_else(|_| {
                        panic!(
                            "LLVM_LIB_DIR is set but `{}` does not exist",
                            lib_dir.display()
                        );
                    })
```

c2rust-refactor/Cargo.toml
```diff
@@ -35,6 +35,7 @@ petgraph = "0.4"
 cargo-util = "0.2.1"
 shlex = "1.3"
 kstring = "=2.0.0" # v2.0.0 has a MSRV of 1.59, while v2.0.1 has a MSRV of 1.73, but we're pinned to 1.65.
+libloading = "=0.7.4"
```
kkysen, line 38: Can you add the reason it's pinned?  I assume an MSRV issue, but it's good to have them documented like the others.

c2rust-refactor/build.rs
```diff
@@ -5,7 +5,9 @@ use std::path::Path;
 use std::process::Command;
 
 fn process_ast(mode: &str, dest: &Path) {
-    let mut p = Command::new("gen/process_ast.py")
+    let mut p = Command::new("uv")
+        .arg("run")
+        .arg("gen/process_ast.py")
```
kkysen, line 10: In order to keep these in sync, could you just read the `#!`?  You can just read the first line, strip the `#!/usr/bin/env -S ` prefix (panic if different), and then run what's after that.

c2rust-refactor/src/plugin.rs
```diff
@@ -1,39 +1,35 @@
...
                 unsafe {
-                    let so = dlopen(c_path.as_ptr(), RTLD_LAZY);
-                    if so.is_null() {
+                    let lib = libloading::Library::new(path_str.clone());
+                    if lib.is_err() {
                         panic!("failed to open plugin `{}`", path_str);
                     }
-                    let sym = dlsym(so, sym_name.as_ptr());
-                    if sym.is_null() {
+                    let lib = lib.unwrap();
```
kkysen, line 24:
```suggestion
                    let lib = libloading::Library::new(path_str.clone())
                        .unwrap_or_else(|_| panic!("failed to open plugin `{path_str}`"));
```

c2rust-refactor/src/plugin.rs
```diff
@@ -1,39 +1,35 @@
...
+                    let f: Result<libloading::Symbol<fn(&mut Registry)>, libloading::Error> = lib.get(b"register_commands");
+                    if f.is_err() {
                         panic!(
                             "failed to locate symbol `register_commands` in `{}`",
                             path_str
                         );
                     }
-                    let f: fn(&mut Registry) = mem::transmute(sym);
+                    let f = f.unwrap();
```
kkysen, line 32:
```suggestion
                    let f = lib.get::<fn(&mut Registry)>(b"register_commands")
                        .unwrap_or_else(|_| panic!("failed to locate symbol `register_commands` in `{path_str}`"));
```

c2rust-analyze/src/known_fn.rs
```diff
@@ -607,6 +607,7 @@ pub const fn all_known_fns() -> &'static [KnownFn] {
                 i: c_int,
             ) -> c_int;
 
+            #[cfg(any(target_os = "linux", target_os = "macos"))]
```
kkysen, line 610: Correct me if I'm wrong, but is this the right `cfg` here?  Or do we want something more like `#[cfg(not(windows))]`?  Same for the others.

kkysen (changes requested):
> This is draft PR to check if there is appetite for that

Thanks for working on this!  We're not able to help add Windows support and test and maintain it, but similarly to Nix, if you'd like to work on adding Windows support, we'll welcome it and review and merge your PRs.  This one generally looks good so far.

Another thing is that #1444 adds much better support for cross-OS transpilation.  We're not planning to test it on Windows, but the infrastructure should be there to do so if that seems like a good alternative, too.  We'll likely get back to that cross-transpilation work in the medium-term future.

c2rust-refactor/src/plugin.rs
kant2002: Thanks, I'm new to Rust, just super interesting in actual C to Rust process. So I definitely have to catch up on idioms.

c2rust-refactor/Cargo.toml
kant2002: Yes. it's minimum supported Rust version.

c2rust-analyze/src/known_fn.rs
kant2002: Probably right. I did not found that construct in the documentation, or maybe miss it, so I decide to go with what I see. Also I did not notice that you support FreeBSD (even if not officially), and for that my construct definitely incorrect.

c2rust-refactor/src/plugin.rs
kkysen: Of course, glad to help.

PR #1477, opened 2025-11-26
Add definition splitting to transpiler

fw-immunant:
Currently this is just the C-side implementation; I'll add commits moving the impl of Rust splitting and merging from the crisp repository tomorrow.

PR #1505, opened 2025-12-09
Add crisp tools

fw-immunant:
This moves the Rust file splitting/merging tools from the CRISP repo, branch [split-merge-rust](https://github.com/GaloisInc/Tractor-Crisp/tree/split-merge-rust), along with some follow-up fixes and usability improvements.


fw-immunant:
This precedes #1477, which I'll rebase after this lands.

tools/merge_rust/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::fs::{self, File};
+use std::path::PathBuf;
+use rust_util::collect::FileCollector;
+use rust_util::item_span::item_spans;
+use serde_json;
+use clap::Parser;
+
+#[derive(Parser)]
+/// Merge updated item definitions into a Rust codebase.
+struct Args {
```
kkysen: nit:

```suggestion
/// Merge updated item definitions into a Rust codebase.
#[derive(Parser)]
struct Args {
```

tools/rust_util/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            Error::Io(ref x) => Display::fmt(x, f),
+            Error::Syn(ref x) => Display::fmt(x, f),
+            Error::Str(ref x) => Display::fmt(x, f),
+            Error::At(ref desc, ref inner) => write!(f, "{desc}: {inner}"),
+        }
```
kkysen:
```suggestion
        match self {
            Error::Io(x) => Display::fmt(x, f),
            Error::Syn(x) => Display::fmt(x, f),
            Error::Str(x) => Display::fmt(x, f),
            Error::At(desc, inner) => write!(f, "{desc}: {inner}"),
        }
```

tools/rust_util/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match *self {
+            Error::Io(ref x) => Some(x),
+            Error::Syn(ref x) => Some(x),
+            Error::Str(_) => None,
+            Error::At(_, ref x) => Some(x),
+        }
```
kkysen:
```suggestion
        match self {
            Error::Io(x) => Some(x),
            Error::Syn(x) => Some(x),
            Error::Str(_) => None,
            Error::At(_, x) => Some(x),
        }
```

tools/merge_rust/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    let mut fc = FileCollector::default();
+    fc.parse(&src_root_path, vec![], true).unwrap();
+    for &(ref file_path, ref mod_path, ref ast) in &fc.files {
+        eprintln!("visit {:?}", file_path);
+        let old_src = fs::read_to_string(file_path).unwrap();
+
+        let mut rewrites = Vec::new();
+        for (item_path, lo, hi) in item_spans(mod_path.to_owned(), ast) {
+            let old_snippet = &old_src[lo .. hi];
```
kkysen: Could you run `cargo +stable fmt` on the tools?

kkysen (commented):
I agree, I think we don't need to carefully review this since it's already in crisp.  I did notice a few little things, though.

Also, we have to run `cargo +stable` in each of the tool directories.  Is there a way to avoid that?  Would we need an overriding `rust-toolchain.toml` in each of them that has `channel = "stable"`?

tools/merge_rust/src/main.rs
fw-immunant: Done.

PR #1511, opened 2025-12-13
transpile: Allow `clippy::missing_safety_doc` in generated code

Rua:
Allows the `unused_imports` and `clippy::missing_safety_doc` in all generated code. There's always a lot of these which makes it harder to see genuine warnings.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -1332,7 +1332,14 @@ fn arrange_header(t: &Translation, is_binary: bool) -> (Vec<syn::Attribute>, Vec
         for (key, mut values) in pragmas {
             values.sort_unstable();
             // generate #[key(values)]
-            let meta = mk().meta_list(vec![key], values);
+            let args: Vec<_> = values
+                .into_iter()
+                .map(|path_str| {
+                    let path_vec: Vec<_> = path_str.split("::").collect();
+                    mk().meta_path(path_vec)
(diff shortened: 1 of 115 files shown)


Rua:
I see that in order to support clippy pragmas, I first need to make some changes in various places to allow full paths instead of just strings. But before I do all that, is that actually wanted?

kkysen (commented):
Which `unused_imports` are we generating?  If we can reduce those, that'd also be good.

> I see that in order to support clippy pragmas, I first need to make some changes in various places to allow full paths instead of just strings. But before I do all that, is that actually wanted?

I don't think this is particularly important, but supporting clippy pragmas should be fine I think.

Rua:
I've removed the `unused_imports`, so now it only adds `clippy::missing_safety_doc`.

kkysen (commented):
Wouldn't `clippy::missing_safety_doc` be a good warning to have, though?  When you manually make the codebase safer, you'll want safety docs.  And couldn't you suppress this lint just through `clippy` instead of inline?  So I'm not sure this makes sense, but if you have another use case, let me know, as I might've not considered it.

Rua:
My concern is that there are so many of these that it drowns out any other issues, and safety docs are unlikely to be the first thing the user wants to address. So it just adds a lot of noise. When the user is ready to deal with the safety docs, they can re-enable the lint.

ahomescu (approved):

PR #1515, opened 2025-12-16
transpile: Also resolve extension operators in `unwrap_predefined_ident`

### LINKED ISSUE #1514: Cannot take address of a temporary
Currently when translating json-c and zstd we are producing code that doesn't compile because it's attempting to take the address of a temporary:

```
json-c: cargo.transpile(tests/json-c/cargo.transpile.gen.sh)...... FAIL
error[E0745]: cannot take address of a temporary
   --> src/json_object.rs:731:24
    |
731 |               &raw const ::core::mem::transmute::<
    |  ________________________^
732 | |                 [u8; 58],
733 | |                 [::core::ffi::c_char; 58],
734 | |             >(*b"struct json_object *json_object_get(struct json_object *)\0")
    | |______________________________________________________________________________^ temporary value
```

Note that this made it onto master because there is also [a bug in CI that's causing CI to not report when tests fail](#1513).

cc @Rua

Rua:
- Fixes #1514.

Thanks to @randomPoison for helping me find it.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -559,9 +559,18 @@ impl TypedAstContext {
         expr_id
     }
 
+    /// Returns the expression inside an `__extension__` operator.
+    pub fn resolve_extension(&self, expr_id: CExprId) -> CExprId {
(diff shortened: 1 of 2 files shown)


kkysen (approved):
LGTM!  Thanks for fixing it so quickly!

PR #1517, opened 2025-12-16
Fix error code reporting for test suite failures

randomPoison:
Looks like the issue is that when we did `for result in results` it emptied the iterator, so when we then did the `if not all(result.passed for result in results)` check it didn't actually look at any test results. Seems like pulling the test results into a list first fixes the issue.

I branched this off from before https://github.com/immunant/c2rust/pull/1515 so that we can verify that CI actually reports the failures now. I'll rebase onto master after that's confirmed.

### DIFF
--- a/tests/integration/tests/__init__.py
+++ b/tests/integration/tests/__init__.py
@@ -238,9 +238,8 @@ def run_tests(conf: Config, generated_scripts: set[Path]):
         return TestResult(test=test, passed=passed, time=time)
 
     with ThreadPoolExecutor() as executor:
-        results = executor.map(run, tests)
+        results = list(executor.map(run, tests))
 
-    results = [result for result in results]
     for result in results:
         print(f"{result.test.name} took {result.time}")
     if not all(result.passed for result in results):


tests/integration/tests/__init__.py
```diff
@@ -240,6 +240,7 @@ def run(test: Test) -> TestResult:
     with ThreadPoolExecutor() as executor:
         results = executor.map(run, tests)
```
kkysen:
```suggestion
        results = list(executor.map(run, tests))
```

randomPoison: Ah yup that seems more idiomatic \o/

randomPoison:
welp this didn't actually report any errors because I think when CI runs on a pull request it tests the merged version of the code. I could inject a synthetic failure, but I've tested this locally and confirmed an exit code is set so I think we just merge this.

kkysen:
> welp this didn't actually report any errors because I think when CI runs on a pull request it tests the merged version of the code. I could inject a synthetic failure, but I've tested this locally and confirmed an exit code is set so I think we just merge this.

I think local testing is good enough.

PR #1518, opened 2025-12-16
Support python 3.12 in test suite runner

randomPoison:
The `Generator` class takes 3 type arguments, but in python 3.13 it became possible to just specify the first one and the other 2 would default to `None`. Explicitly adding the `None` arguments allows this to also work on python 3.12.

One of my work machines is on a version of Ubuntu that has 3.12 installed and it looks like it'd be mildly painful to install 3.13. Seems like it'd be reasonable to support 3.12 here, unless there's some reason that'd be a bad idea.

### DIFF
--- a/tests/integration/tests/templates.py
+++ b/tests/integration/tests/templates.py
@@ -102,14 +102,14 @@ def render_script(template: str, out_path: str, params: Dict):
     os.chmod(out_path, stat.S_IREAD | stat.S_IWRITE | stat.S_IEXEC)
 
 
-def autogen_cargo(conf_file, yaml: Dict) -> Generator[Path]:
+def autogen_cargo(conf_file, yaml: Dict) -> Generator[Path, None, None]:
     """
     Yield generated paths.
     """
 
     def render_stage(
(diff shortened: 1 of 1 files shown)


randomPoison:
Yeah I agree it's uglier. Is the test suite setup with uv? idk how any of that works, but if it's easy enough to setup a venv or whatever with uv where I can locally setup 3.13 then I can also try that.

thedataking:
I bumped us down to Python 3.12 over here https://github.com/immunant/c2rust-testsuite/commit/593f753e13e758ba2293cc41e749c0d8bf1d4d0a after we discussed it on Slack.

Please land with 3.12 support as Khyber approved rather spending more time on this matter.

kkysen:
> Is the test suite setup with uv?

Yes.

> if it's easy enough to setup a venv or whatever with uv where I can locally setup 3.13 then I can also try that.

If you increase the version in project.toml, go to that directory it's in, `rm -rf .venv/`, and `uv python install`, I think it should do it automatically.  Might need an extra `uv sync` as well.

PR #1521, opened 2025-12-18
transpile: Factor out more pointer-related functions

Rua:
This factors out functions for two things: casts involving pointers, and null checking. In the case of `==` and `!=` operators in a non-conditional context (in operators.rs), this also adds special handling for regular pointers, where only function pointers were specially handled previously. This means that

```c
int b = ptr == NULL;
```

is now translated with the `is_null` method as well, instead of with the `==` operator and `core::ptr::null`.

### DIFF
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -569,12 +569,12 @@ impl<'c> Translation<'c> {
         &self,
         ctx: ExprContext,
         ptr_type: CTypeId,
-        e: Box<Expr>,
+        val: Box<Expr>,
         is_null: bool,
     ) -> TranslationResult<Box<Expr>> {
         Ok(if self.ast_context.is_function_pointer(ptr_type) {
             let method = if is_null { "is_none" } else { "is_some" };
-            mk().method_call_expr(e, method, vec![])
+            mk().method_call_expr(val, method, vec![])
         } else {
             // TODO: `pointer::is_null` becomes stably const in Rust 1.84.
             if ctx.is_const {
@@ -583,11 +583,11 @@ impl<'c> Translation<'c> {
                     "cannot check nullity of pointer in `const` context",
                 ));
             }
-            let is_null_val = mk().method_call_expr(e, "is_null", vec![]);
+            let val = mk().method_call_expr(val, "is_null", vec![]);
             if !is_null {
-                mk().unary_expr(UnOp::Not(Default::default()), is_null_val)
+                mk().unary_expr(UnOp::Not(Default::default()), val)
             } else {
-                is_null_val
+                val
             }
         })
     }


Rua:
> By the way, when you do refactoring PRs like this, can you try to split out the pure moves vs. the semantic changes more carefully/fine-grained?

How is this now?

Rua:
Does anything else need to be done here?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4434,79 +4434,36 @@ impl<'c> Translation<'c> {
...
-                            None,
-                            "cannot transmute integers to Option<fn ...> in `const` context",
-                        ));
-                    }
-                    Ok(WithStmts::new_unsafe_val(transmute_expr(
-                        intptr_t, target_ty, intptr,
-                    )))
-                })
+            CastKind::IntegralToPointer => {
+                self.convert_integral_to_pointer_cast(ctx, source_cty.ctype, target_cty.ctype, val)
```
kkysen, line 4464: In f97fdaa, this call has 4 args vs. 5 in the function definition, with `expr: Option<CExprId>` missing.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4434,79 +4434,36 @@ impl<'c> Translation<'c> {
...
-                        Ok(WithStmts::new_unsafe_val(transmute_expr(
-                            source_ty, target_ty, val,
-                        )))
-                    })
-                } else {
-                    // Normal case
-                    let target_ty = self.convert_type(target_cty.ctype)?;
-                    Ok(val.map(|val| mk().cast_expr(val, target_ty)))
-                }
+                self.convert_pointer_to_pointer_cast(source_cty.ctype, target_cty.ctype, val)
```
kkysen, line 4460: In f97fdaa, this is passing stuff like `source_cty.ctype` (a `CTypeId`), but the function definition takes `CQualTypeId`.

kkysen: I think you need to initially pass just `source_cty` before the next commit, which changes the function signature to `CTypeId`.

c2rust-transpile/src/translator/pointers.rs
```diff
@@ -453,4 +454,203 @@ impl<'c> Translation<'c> {
...
+                val.and_then(|x| {
+                    if self.ast_context.is_function_pointer(source_cty.ctype) {
+                        Ok(WithStmts::new_unsafe_val(transmute_expr(
+                            source_ty, target_ty, x,
+                        )))
+                    } else {
+                        Ok(WithStmts::new_val(mk().cast_expr(x, target_ty)))
+                    }
+                })
+            }
```
kkysen: Is this supposed to be identical to most of `fn convert_pointer_to_integral_cast`?

Rua: It is nearly identical because both functions were initially based on the same code from mod.rs. The refactor commit after that then removes the inactive parts. I separated the two steps into different commits so that moves and refactors were more clearly visible.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4792,30 +4769,13 @@ impl<'c> Translation<'c> {
...
-                    None,
-                    "cannot check nullity of pointer in `const` context",
-                ));
-            }
-            let mut res = mk().method_call_expr(val, "is_null", vec![]);
-            if target {
-                res = mk().unary_expr(UnOp::Not(Default::default()), res)
-            }
-            res
+        if ty.is_pointer() {
```
kkysen: Can you leave the surrounding `Ok(` so we don't need the extra `Ok(...)`s in a lot of places?  Then you can just use `?` on the `self.convert_pointer_is_null(...)`.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -612,45 +617,27 @@ impl<'c> Translation<'c> {
             c_ast::BinOp::ShiftLeft => mk().binary_expr(BinOp::Shl(Default::default()), lhs, rhs),
 
             c_ast::BinOp::EqualEqual => {
-                // Using is_none method for null comparison means we don't have to
-                // rely on the PartialEq trait as much and is also more idiomatic
```
kkysen, line 616: Can you keep this comment?  I think it still applies.

kkysen: You can de-duplicate it with the same one for `NotEqual`, though.  The bodies are short enough now, that I think a single comment outside of the `=> {}`s are fine to cover both.

c2rust-transpile/src/translator/pointers.rs
```diff
@@ -568,4 +568,34 @@ impl<'c> Translation<'c> {
...
+        ptr_type: CTypeId,
+        e: Box<Expr>,
+        negated: bool,
+    ) -> TranslationResult<Box<Expr>> {
+        Ok(if self.ast_context.is_function_pointer(ptr_type) {
+            if negated {
+                mk().method_call_expr(e, "is_some", vec![])
+            } else {
+                mk().method_call_expr(e, "is_none", vec![])
+            }
```
kkysen: You could do this as a follow-up commit, but could we do this?

```suggestion
            let method = if negated { "is_some" } else { "is_none" };
            mk().method_call_expr(e, method, vec![])
```

c2rust-transpile/src/translator/operators.rs
kkysen: If you want, you could also combine the `EqualEqual` and `NotEqual` arms.

c2rust-transpile/src/translator/pointers.rs
```diff
@@ -573,14 +573,14 @@ impl<'c> Translation<'c> {
         &self,
         ctx: ExprContext,
         ptr_type: CTypeId,
-        e: Box<Expr>,
-        negated: bool,
+        mut val: Box<Expr>,
+        is_true: bool,
```
kkysen: I think the previous way was clearer.  Not sure if this commit helps.

c2rust-transpile/src/translator/mod.rs
kkysen, line 4915: This `Ok(`.

c2rust-transpile/src/translator/pointers.rs
Rua: The use of `negated` felt unclear to me in the context of checking "is null". Is it checking if "is null" is negated, or if its truthiness in the C sense is negated? With the new wording it's hopefully more clear that it's checking if "is null" is true or false.

Rua: I've now renamed it to `is_null` to hopefully make it even more obvious.

kkysen: `is_null` is clear.  That's better.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2726,45 +2726,44 @@ impl<'c> Translation<'c> {
...
                 if self.ast_context.is_null_expr(null_expr) =>
             {
-                null_pointer_case(target, ptr)
+                null_pointer_case(ptr, !target)
             }
 
             CExprKind::Unary(_, c_ast::UnOp::Not, subexpr_id, _) => {
                 self.convert_condition(ctx, !target, subexpr_id)
             }
-
```
kkysen, line 2779: Can you leave this newline?

c2rust-transpile/src/translator/operators.rs
```diff
@@ -611,46 +616,23 @@ impl<'c> Translation<'c> {
...
-                    }
+            // Using is_none method for null comparison means we don't have to
+            // rely on the PartialEq trait as much and is also more idiomatic
+            c_ast::BinOp::EqualEqual | c_ast::BinOp::NotEqual => {
+                let (negated, bin_op) = if let c_ast::BinOp::EqualEqual = op {
+                    (true, BinOp::Eq(Default::default()))
                 } else {
-                    mk().binary_expr(BinOp::Eq(Default::default()), lhs, rhs)
+                    (false, BinOp::Ne(Default::default()))
                 };
```
kkysen: I think this is backwards.  `EqualEqual` used to be `is_none`, but `negated = true` means it will be `is_some`.

kkysen: Let me push a fix.

Rua: Ah, the logic was wrong in the previous commit, but corrected in the next, it was just the variable that got named wrong.

kkysen:
@Rua, I'm going to force push.  I need to untangle some commis.  The follow-ups I suggested last time need to go in separate commits because they are separate changes, and they're much harder to review when combined into the same commits that are repeatedly rebased.

Rua:
Yes, looks good.

Issue #1498, opened 2025-12-07
Translate enums as newtype structs with associated constants?


Rua:
Currently, enums are translated as simple type aliases, with their variants translated as separate `const` values. This translation loses the connection between the variants and the enum they belong to. So I wonder if it would be nicer to translate them as newtype structs with associated constants?

So the following C code
```c
enum MyEnum {
    Variant1,
    Variant2,
    Variant3,
};
```
would become
```rust
#[derive(Clone, Copy)]
#[repr(transparent)]
pub struct MyEnum(pub ::core::ffi::c_uint);

impl MyEnum {
    pub const Variant1: Self = Self(0);
    pub const Variant2: Self = Self(1);
    pub const Variant3: Self = Self(2);
}
```

This matches how, for example, [ash](https://github.com/ash-rs/ash) translates C enums ([example](https://docs.rs/ash/latest/ash/vk/struct.ImageLayout.html)). This way, the information that was present in the C type system is not lost, which is a benefit for later refactoring. The user can refactor it into the original system relatively easily if they want.

kkysen:
I agree this would be a nicer translation.  We would also need to add casts in places where an `enum` is implicitly coerced to an `int` or other `enum`, which does happen in C a bunch.

ahomescu:
C heavily relies on the enums actually being their underlying integer types, so wrapping them on the Rust side will introduce a ton of boilerplate (un)wrapping code. I wonder how ugly it will get, it might not be worth doing if it's too much.

kkysen:
> C heavily relies on the enums actually being their underlying integer types, so wrapping them on the Rust side will introduce a ton of boilerplate (un)wrapping code. I wonder how ugly it will get, it might not be worth doing if it's too much.

You have to do this one way or another to make it more idiomatic Rust, though.  And the type safety is very useful for refactoring, especially now that it will give LLMs errors as feedback.

PR #1525, opened 2025-12-19
transpile: Split off `enums` module

Rua:
Creating another submodule for specific things, to offload the main `mod.rs`. Some of the functions here may be a bit trivial at the moment, but they will be useful to implement #1498.

### DIFF
--- a/c2rust-transpile/src/translator/enums.rs
+++ b/c2rust-transpile/src/translator/enums.rs
@@ -85,44 +85,44 @@ impl<'c> Translation<'c> {
         ctx: ExprContext,
         enum_type_id: CTypeId,
         enum_id: CEnumId,
-        expr: CExprId,
+        expr: Option<CExprId>,
         val: Box<Expr>,
     ) -> TranslationResult<Box<Expr>> {
-        match self.ast_context[expr].kind {
-            // This is the case of finding a variable which is an `EnumConstant` of the same enum
-            // we are casting to. Here, we can just remove the extraneous cast instead of generating
-            // a new one.
-            CExprKind::DeclRef(_, enum_constant_id, _)
-                if self.is_variant_of_enum(enum_id, enum_constant_id) =>
-            {
-                // `enum`s shouldn't need portable `override_ty`s.
-                let expr_is_macro = matches!(
-                    self.convert_const_macro_expansion(ctx, expr, None),
-                    Ok(Some(_))
-                );
-
-                // If this DeclRef expanded to a const macro, we actually need to insert a cast,
-                // because the translation of a const macro skips implicit casts in its context.
-                if !expr_is_macro {
-                    return Ok(self.enum_constant_expr(enum_constant_id));
-                }
-            }
-
-            CExprKind::Literal(_, CLiteral::Integer(i, _)) => {
-                return Ok(self.enum_for_i64(enum_type_id, i as i64));
-            }
-
-            CExprKind::Unary(_, c_ast::UnOp::Negate, subexpr_id, _) => {
-                if let &CExprKind::Literal(_, CLiteral::Integer(i, _)) =
-                    &self.ast_context[subexpr_id].kind
+        if let Some(expr) = expr {
+            match self.ast_context[expr].kind {
+                // This is the case of finding a variable which is an `EnumConstant` of the same
+                // enum we are casting to. Here, we can just remove the extraneous cast instead of
+                // generating a new one.
+                CExprKind::DeclRef(_, enum_constant_id, _)
+                    if self.is_variant_of_enum(enum_id, enum_constant_id) =>
                 {
-                    return Ok(self.enum_for_i64(enum_type_id, -(i as i64)));
-                }
-            }
+                    // `enum`s shouldn't need portable `override_ty`s.
+                    let expr_is_macro = matches!(
+                        self.convert_const_macro_expansion(ctx, expr, None),
+                        Ok(Some(_))
+                    );
 
-            // In all other cases, a cast to an enum requires a `transmute` - Rust enums cannot be
-            // converted into integral types as easily as C ones.
-            _ => {}
+                    // If this DeclRef expanded to a const macro, we actually need to insert a cast,
+                    // because the translation of a const macro skips implicit casts in its context.
+                    if !expr_is_macro {
+                        return Ok(self.enum_constant_expr(enum_constant_id));
+                    }
+                }
+
+                CExprKind::Literal(_, CLiteral::Integer(i, _)) => {
+                    return Ok(self.enum_for_i64(enum_type_id, i as i64));
+                }
+
+                CExprKind::Unary(_, c_ast::UnOp::Negate, subexpr_id, _) => {
+                    if let &CExprKind::Literal(_, CLiteral::Integer(i, _)) =
+                        &self.ast_context[subexpr_id].kind
+                    {
+                        return Ok(self.enum_for_i64(enum_type_id, -(i as i64)));
+                    }
+                }
+
+                _ => {}
+            }
         }
 
         let target_ty = self.convert_type(enum_type_id)?;
@@ -130,7 +130,7 @@ impl<'c> Translation<'c> {
     }
 
     /// Given an integer value this attempts to either generate the corresponding enum
-    /// variant directly, otherwise it transmutes a number to the enum type.
+    /// variant directly, otherwise it converts a number to the enum type.
     fn enum_for_i64(&self, enum_type_id: CTypeId, value: i64) -> Box<Expr> {
         let enum_id = match self.ast_context.resolve_type(enum_type_id).kind {
             CTypeKind::Enum(enum_id) => enum_id,
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4486,7 +4486,13 @@ impl<'c> Translation<'c> {
                     let expr =
                         expr.ok_or_else(|| format_err!("Casts to enums require a C ExprId"))?;
                     val.result_map(|val| {
-                        self.convert_cast_to_enum(ctx, target_cty.ctype, enum_decl_id, expr, val)
+                        self.convert_cast_to_enum(
+                            ctx,
+                            target_cty.ctype,
+                            enum_decl_id,
+                            Some(expr),
+                            val,
+                        )
                     })
                 } else if target_ty_kind.is_floating_type() && source_ty_kind.is_bool() {
                     val.and_then(|x| {
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -199,6 +199,7 @@ impl<'c> Translation<'c> {
 
     fn convert_assignment_operator_aux(
         &self,
+        ctx: ExprContext,
         bin_op_kind: BinOp,
         bin_op: c_ast::BinOp,
         read: Box<Expr>,
@@ -230,7 +231,7 @@ impl<'c> Translation<'c> {
 
                 mk().call_expr(fn_path, args)
             } else {
-                mk().cast_expr(read, lhs_type.clone())
+                mk().cast_expr(read, lhs_type)
             };
             let ty = self.convert_type(compute_res_type_id.ctype)?;
             let mut val = self.convert_binary_operator(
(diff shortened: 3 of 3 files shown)


Rua:
CI seems to be crashing, and I don't think it's caused by any of my changes?

Rua:
It works now!

c2rust-transpile/src/translator/enums.rs
```diff
@@ -0,0 +1,185 @@
...
+            .expect("Enums should already be renamed");
+        self.add_import(enum_id, &enum_name);
+
+        let ty = mk().path_ty(vec![enum_name]);
+        let val = match value {
+            ConstIntExpr::I(value) => signed_int_expr(value),
+            ConstIntExpr::U(value) => mk().lit_expr(mk().int_unsuffixed_lit(value as u128)),
+        };
+
+        Ok(ConvertedDecl::Item(
```
kkysen, line 57: This always returns `Ok` only and there's no `?`, so I think it'd be better to just return `ConvertedDecl` and `Ok`-wrap it on the caller's side.

c2rust-transpile/src/translator/enums.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// Translates a cast where the source type, but not the target type, is an `enum` type.
+    pub fn convert_cast_from_enum(
+        &self,
+        target_cty: CTypeId,
+        val: Box<Expr>,
+    ) -> TranslationResult<Box<Expr>> {
+        // Convert it to the expected integral type.
+        // When modifying this, look at `Translation::convert_cast_to_enum` -
+        // this function assumes `DeclRef`'s to `EnumConstants`'s will translate to casts.
```
kkysen:
```suggestion
        // When modifying this, look at `Translation::convert_cast_to_enum`;
        // this function assumes `DeclRef`s to `EnumConstants`s will translate to casts.
```

c2rust-transpile/src/translator/enums.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    pub fn convert_enum_zero_initializer(
+        &self,
+        type_id: CTypeId,
+    ) -> TranslationResult<WithStmts<Box<Expr>>> {
+        Ok(WithStmts::new_val(self.enum_for_i64(type_id, 0)))
+    }
+
+    /// Translates a cast where the source type, but not the target type, is an `enum` type.
```
kkysen:
```suggestion
    /// Translate a cast where the source type, but not the target type, is an `enum` type.
```

c2rust-transpile/src/translator/enums.rs
```diff
@@ -0,0 +1,185 @@
...
+        val: Box<Expr>,
+    ) -> TranslationResult<Box<Expr>> {
+        // Convert it to the expected integral type.
+        // When modifying this, look at `Translation::convert_cast_to_enum` -
+        // this function assumes `DeclRef`'s to `EnumConstants`'s will translate to casts.
+        let ty = self.convert_type(target_cty)?;
+        Ok(mk().cast_expr(val, ty))
+    }
+
+    /// Translates a cast where the target type is an `enum` type.
```
kkysen:
```suggestion
    /// Translate a cast where the target type is an `enum` type.
```

c2rust-transpile/src/translator/enums.rs
```diff
@@ -0,0 +1,185 @@
...
+    ///
+    /// When translating variable references to `EnumConstant`'s, we always insert casts to the
+    /// expected type. In C, `EnumConstants` have some integral type, _not_ the enum type. However,
+    /// if we then immediately have a cast to convert this variable back into an enum type, we would
+    /// like to produce Rust with _no_ casts. This function handles this simplification.
```
kkysen:
```suggestion
    /// When translating variable references to `EnumConstant`s, we always insert casts to the expected type.
    /// In C, `EnumConstant`s have some integral type, _not_ the enum type.
    /// However, if we then immediately have a cast to convert this variable back into an enum type,
    /// we would like to produce Rust with _no_ casts.
    /// This function handles this simplification.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4621,6 +4585,9 @@ impl<'c> Translation<'c> {
                             Ok(WithStmts::new_unsafe_val(transmute_expr(
                                 source_ty, target_ty, x,
                             )))
+                        } else if let &CTypeKind::Enum(..) = source_ty_kind {
+                            self.convert_cast_from_enum(target_cty.ctype, x)
+                                .map(WithStmts::new_val)
```
kkysen, line 4522: What added this?

c2rust-transpile/src/translator/enums.rs
kkysen: I guess if we want to keep them all returning `TranslationResult`s for consistency, that might be fine, too.

c2rust-transpile/src/translator/mod.rs
Rua: It's not needed for anything currently, but I added it for consistency and explicitness if casts from enums need a different implementation in the future.

kkysen: Hmm, could you move it to a separate commit?

Rua:
@kkysen could you review this again? I have a fix for #1498 ready that depends on this.

kkysen (commented):
> @kkysen could you review this again? I have a fix for #1498 ready that depends on this.

Sorry it's been a while.  I'll look at it now.

c2rust-transpile/src/translator/enums.rs
```diff
@@ -85,49 +85,51 @@ impl<'c> Translation<'c> {
...
-                let expr_is_macro = matches!(
-                    self.convert_const_macro_expansion(ctx, expr, None),
-                    Ok(Some(_))
-                );
-
-                // If this DeclRef expanded to a const macro, we actually need to insert a cast,
-                // because the translation of a const macro skips implicit casts in its context.
-                if !expr_is_macro {
-                    return Ok(self.enum_constant_expr(enum_constant_id));
+        if let Some(expr) = expr {
```
kkysen, line 91: Since almost the whole function is in an `if let Some`, do you want to just leave `fn convert_cast_to_enum` alone (renamed) and add a wrapper `fn convert_cast_to_enum` that takes `Option<CExprId>` and calls the original function?

Rua: I don't think that would make that much sense in this case? The main logic of the function is in the part below the `if let Some`, the branch is just special casing some things.

Rua:
I've split the refactoring commit into a bunch of smaller steps, hopefully it is clearer now.

c2rust-transpile/src/translator/enums.rs
```diff
@@ -178,4 +170,23 @@ impl<'c> Translation<'c> {
...
+        variants.contains(&enum_constant_id)
+    }
+
+    fn enum_integral_type(&self, enum_id: CEnumId) -> CQualTypeId {
+        match self.ast_context[enum_id].kind {
+            CDeclKind::Enum {
+                integral_type: Some(integral_type),
+                ..
+            } => integral_type,
+            _ => panic!("{:?} does not point to an `enum` declaration", enum_id),
```
kkysen:
```suggestion
            _ => panic!("{:?} does not point to an integral `enum` declaration", enum_id),
```

The `integral_type: None` case is now included in `_` now, so it should be in the error message, too; otherwise, it could be pretty confusing.

c2rust-transpile/src/translator/enums.rs
```diff
@@ -86,41 +82,40 @@ impl<'c> Translation<'c> {
...
-                    // If this DeclRef expanded to a const macro, we actually need to insert a cast,
-                    // because the translation of a const macro skips implicit casts in its context.
-                    Expr::Path(..) => mk().cast_expr(x, target_ty),
-                    _ => panic!(
-                        "DeclRef {:?} of enum {:?} is not cast: {x:?}",
-                        expr, enum_id
-                    ),
-                });
+                let expr_is_macro = matches!(
+                    self.convert_const_macro_expansion(ctx, expr, None),
```
kkysen: Do we need the `override_ty` here?

c2rust-transpile/src/translator/enums.rs
```diff
@@ -86,41 +82,40 @@ impl<'c> Translation<'c> {
     /// like to produce Rust with _no_ casts. This function handles this simplification.
     pub fn convert_cast_to_enum(
```
kkysen, line 83: What's the existing Rust syntax you're referring to in ``Don't rely on existing Rust syntax in `convert_cast_to_enum` ``?

Rua: In the original code, it expected to be passed a Rust cast expression, and it would then remove the cast. But that was obviously fragile, so I changed it so it no longer relies on that.

c2rust-transpile/src/translator/enums.rs
Rua: I don't know.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -244,25 +245,21 @@ impl<'c> Translation<'c> {
...
+
+            if let &CTypeKind::Enum(enum_id) = result_ty_kind {
+                val = val.result_map(|val| {
+                    self.convert_cast_to_enum(ctx, lhs_type_id.ctype, enum_id, None, val)
+                })?;
+            } else if compute_lhs_resolved_ty.kind == CTypeKind::LongDouble {
                 // We can't as-cast from a non primitive like f128 back to the result_type
-                if compute_lhs_resolved_ty.kind == CTypeKind::LongDouble {
-                    let resolved_lhs_kind = &self.ast_context.resolve_type(lhs_type_id.ctype).kind;
+                let resolved_lhs_kind = &self.ast_context.resolve_type(lhs_type_id.ctype).kind;
```
kkysen: Isn't this the same as `result_ty_kind`?

c2rust-transpile/src/translator/operators.rs
```diff
@@ -244,25 +245,21 @@ impl<'c> Translation<'c> {
...
+                val = self.f128_cast_to(val, resolved_lhs_kind)?;
+            } else {
+                let result_type = self.convert_type(lhs_type_id.ctype)?;
+                val = val.map(|val| mk().cast_expr(val, result_type));
+            }
```
kkysen: Can the `val` assignment be hoisted out of the block/the value of the `if`/`else if`/`else`?  That's how it was before.

c2rust-transpile/src/translator/operators.rs
kkysen: If it is, probably keep the old name of `resolve_lhs_kind`.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -486,16 +483,21 @@ impl<'c> Translation<'c> {
...
+                                        enum_id,
+                                        None,
+                                        val,
+                                    ))?;
                                 } else {
-                                    val.map(|val| mk().cast_expr(val, expr_type))
-                                };
+                                    let expr_type = self.convert_type(expr_type_id.ctype)?;
+                                    val = val.map(|val| mk().cast_expr(val, expr_type));
+                                }
```
kkysen: Can you do the same here?  Keep the `val` assignment outside the `if` and I think `resolved_ty_kind` would be a more consistent name compared to what was before.

c2rust-transpile/src/translator/enums.rs
kkysen: Could you add a comment saying this?  I think `None` is fine here.

```rust
// `enum`s shouldn't need portable `override_ty`s.
```

c2rust-transpile/src/translator/enums.rs
kkysen: Ah, okay.  Could you add that to the commit message?

c2rust-transpile/src/translator/operators.rs
```diff
@@ -486,16 +482,21 @@ impl<'c> Translation<'c> {
...
                                 )?;
 
-                                let expr_resolved_ty =
-                                    self.ast_context.resolve_type(expr_type_id.ctype);
-                                let is_enum_result = expr_resolved_ty.kind.is_enum();
-                                let expr_type = self.convert_type(expr_type_id.ctype)?;
-                                let val = if is_enum_result {
-                                    val.set_unsafe();
-                                    val.map(|val| transmute_expr(lhs_type, expr_type, val))
+                                let result_ty_kind = &self.ast_context.resolve_type(expr_type_id.ctype).kind;
```
kkysen: `let result_type_id = compute_res_type_id.unwrap_or(expr_type_id);` in this function, so `result` means something else in this function, not just `expr_type_id`.  Can this stay `expr_resolved_ty_kind`?

c2rust-transpile/src/translator/enums.rs
kkysen: Aren't there only two lines in `fn convert_cast_to_enum` that are outside of the `if let Some`?

c2rust-transpile/src/translator/operators.rs
Rua: Sorry, not sure what you want to be changed here.

c2rust-transpile/src/translator/enums.rs
Rua: Yes, but the `if let Some` contains early returns. If that's made into its own function, where would the early returns go?

c2rust-transpile/src/translator/operators.rs
kkysen: Can you rename `result_ty_kind` to `expr_resolved_ty_kind`?

c2rust-transpile/src/translator/enums.rs
kkysen: Oh, true.  Hmm, probably not worth changing anything here then.

PR #1528, opened 2025-12-23
Allow empty include paths

abakst:
Given `main.c`
```c
int main() {
  assert(0);
  return 0;
}
```
and `header.h`:
```
void assert(int x);
````

and `compile_commands.json` derived from 

```clang -c -include header.h main.c```

`c2rust transpile compile_commands.json` panics in `TypedAstContext::include_path` because the include path for `header.h` is `None`.

This PR fixes the panic by first checking if the root include is not `None`.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -207,8 +207,10 @@ impl TypedAstContext {
         if cfg!(debug_assertions) {
             if let Some(root_include) = includes.first() {
                 let file_id = self.file_map[root_include.fileid as usize];
-                let file = &self.files[file_id];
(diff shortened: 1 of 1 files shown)


abakst:
I would happily add a regression test, but I'm not sure exactly where to put it.

abakst:
> > I would happily add a regression test, but I'm not sure exactly where to put it.
> 
> Our testing infrastructure isn't very well setup to test custom compile commands, so I'm not sure this is easily testable, but I think it's simple enough to not need a test. Also, this seems to be in a `cfg!(debug_assertions)`, so are you also encountering this error in a release build?
> 
> Otherwise it LGTM besides whatever rustfmt is complaining about.

No this was in a debug build. Thanks!

abakst:
@kkysen apologies but do I re-trigger the CI actions so that I can merge?

PR #1530, opened 2025-12-30
transpile: Prefix c2rust-generated new variable names with `c2rust_`

### LINKED ISSUE #1524: Prefix all newly generated names with `c2rust_`
Whenever the transpiler generates a new name that is not in the original source code, prefix it with `c2rust_`. This makes it much clearer when reviewing/refactoring the generated code, what is original and what is a transpilation artefact. This would not apply to original names that were _renamed_ due to a naming conflict, only to names that are created anew by the transpiler.

Rua:
- Closes #1524.

### DIFF
--- a/c2rust-transpile/src/renamer.rs
+++ b/c2rust-transpile/src/renamer.rs
@@ -105,12 +105,14 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
     }
 
     pub fn pick_name(&mut self, basename: &str) -> String {
+        check_c2rust_name(basename);
         self.pick_name_in_scope(basename, None)
     }
 
     /// Permanently assign a name that doesn't collide with anything
     /// currently in scope, and also never goes out of scope
     pub fn pick_name_root(&mut self, basename: &str) -> String {
+        check_c2rust_name(basename);
         self.pick_name_in_scope(basename, Some(0))
     }
 
@@ -176,6 +178,10 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
         self.next_fresh += 1;
         self.pick_name(&format!("c2rust_fresh{fresh}"))
     }
+}
+
+fn check_c2rust_name(basename: &str) {
+    assert!(basename.starts_with("c2rust_") || basename.starts_with("C2Rust_"));
 }
 
 #[cfg(test)]


c2rust-transpile/src/translator/builtins.rs
```diff
@@ -372,7 +372,9 @@ impl<'c> Translation<'c> {
                     let mut fn_ctx = self.function_context.borrow_mut();
                     let alloca_allocations_name =
                         &*fn_ctx.alloca_allocations_name.get_or_insert_with(|| {
-                            self.renamer.borrow_mut().pick_name("alloca_allocations")
+                            self.renamer
+                                .borrow_mut()
+                                .pick_name("c2rust_alloca_allocations")
```
kkysen, line 377: Could you also add an assert in `fn pick_name` that it starts with `c2rust_`?

Rua: That would not work for the inner/padding suffixes, which turn `Foo` into `Foo_Inner` and `Foo_PADDING` (`Foo_C2RustInner` and `Foo_C2RustPADDING` after this change).

kkysen: Hmm.  Could you instead check that the name either starts with `c2rust_` or contains `_C2Rust`?

Rua: I think simply checking case-insensitively if it contains `c2rust` would be enough? What matters is ensuring that the name is visibly tagged in some way, it doesn't matter exactly where.

kkysen: That might be enough, but I'd rather be a bit more strict to start with, and if we need to relax it for some reason, we can.  Searching for case-sensitive `c2rust_|_C2Rust` will have a few fewer false positives than a case-insensitive `c2rust` search.

Rua: I decided instead to put `C2Rust_` in front of the type names, to make everything more visually consistent.

c2rust-transpile/src/convert_type.rs
```diff
@@ -182,13 +182,11 @@ impl TypeConverter {
...
-            let mut suffix_name = self
-                .renamer
-                .get(&decl_id)
-                .unwrap_or_else(|| "C2RustUnnamed".to_string());
-            suffix_name += suffix;
-
-            self.renamer.pick_name(&suffix_name)
+            let name = self.renamer.get(&decl_id).map_or_else(
+                || format!("C2RustUnnamed_{}", suffix),
+                |name| format!("C2Rust_{}_{}", name, suffix),
```
kkysen:
```suggestion
                || format!("C2RustUnnamed_{suffix}"),
                |name| format!("C2Rust_{name}_{suffix}"),
```

c2rust-transpile/src/renamer.rs
```diff
@@ -174,7 +174,7 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
     pub fn fresh(&mut self) -> String {
         let fresh = self.next_fresh;
         self.next_fresh += 1;
-        self.pick_name(&format!("fresh{}", fresh))
+        self.pick_name(&format!("c2rust_fresh{}", fresh))
```
kkysen:
```suggestion
        self.pick_name(&format!("c2rust_fresh{fresh}"))
```

c2rust-refactor/src/transform/items.rs
```diff
@@ -56,7 +55,7 @@ impl Transform for RenameRegex {
                 smallvec![i.map(|i| {
                     Item {
                         ident: mk().ident(&new_name),
-                        .. i
+                        ..i
```
kkysen: I think formatting changes got in here somehow.  Not sure if they're from you or from upstream changes.  Could you rebase?

c2rust-transpile/src/renamer.rs
```diff
@@ -105,12 +105,14 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
...
+        assert!(basename.starts_with("c2rust_") || basename.starts_with("C2Rust_"));
         self.pick_name_in_scope(basename, None)
     }
 
     /// Permanently assign a name that doesn't collide with anything
     /// currently in scope, and also never goes out of scope
     pub fn pick_name_root(&mut self, basename: &str) -> String {
+        assert!(basename.starts_with("c2rust_") || basename.starts_with("C2Rust_"));
         self.pick_name_in_scope(basename, Some(0))
     }
```
kkysen, line 117: nit:

Could you refactor `assert!(basename.starts_with("c2rust_") || basename.starts_with("C2Rust_"));` into its own function like `fn check_c2rust_name`?  It's a simple line but I don't want it to get out of sync.

c2rust-refactor/src/transform/items.rs
```diff
@@ -127,8 +127,8 @@ impl Transform for RenameUnnamed {
         let mut renamer: Renamer = Default::default();
         let mut counter: usize = 0;
 
-        let has_unnamed = |ident: &Ident| { ident.as_str().contains("C2RustUnnamed") };
-        let make_name = |counter| { Ident::from_str(&format!("C2RustUnnamed_{}", counter)) };
+        let has_unnamed = |ident: &Ident| ident.as_str().contains("C2Rust_Unnamed");
+        let make_name = |counter| Ident::from_str(&format!("C2Rust_Unnamed_{}", counter));
```
kkysen:
```suggestion
        let make_name = |counter| Ident::from_str(&format!("C2Rust_Unnamed_{counter}"));
```

c2rust-refactor/src/transform/reorganize_definitions.rs
kkysen, line 1: I think a lot of formatting changes are still left in this file.

c2rust-transpile/src/convert_type.rs
```diff
@@ -182,11 +182,11 @@ impl TypeConverter {
...
         self.suffix_names.entry(key).or_insert_with(|| {
-            let name = self.renamer.get(&decl_id).map_or_else(
-                || format!("C2RustUnnamed_{suffix}"),
-                |name| format!("C2Rust_{name}_{suffix}"),
-            );
-            self.renamer.pick_name(&name)
+            let name = self
+                .renamer
+                .get(&decl_id)
+                .unwrap_or_else(|| "Unnamed".to_string());
```
kkysen:
```suggestion
            let name = self.renamer.get(&decl_id);
            let name = name.as_deref().unwrap_or("Unnamed");
```

We really should just change `fn Renamer::get` to return `Option<&str>`, but that's a bit more involved.

c2rust-refactor/src/transform/reorganize_definitions.rs
kkysen: What was happening with the formatting?  If we have unformatted files, we should format them (just separately).

Rua: I have my IDE set to automatically format files on save, so I don't forget. But I didn't notice that the file wasn't formatted before so it introduced a lot of changes. Now I skipped the autoformat.

kkysen: Hmm, weird.  We should format it, but I'm not sure why `cargo fmt` isn't doing.  `rustfmt` directly on `reorganize_definitions` works, but not `cargo fmt`.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -372,7 +372,9 @@ impl<'c> Translation<'c> {
...
                         });
 
                     // alloca_allocations.push(std::vec::from_elem(0, count));
```
kkysen:
```suggestion
                    // c2rust_alloca_allocations.push(std::vec::from_elem(0, count));
```

c2rust-transpile/src/translator/builtins.rs
kkysen:
```suggestion
                    // c2rust_alloca_allocations.last_mut().unwrap().as_mut_ptr()
```

PR #1534, opened 2026-01-05
postprocess: handle comments in conditional code

kkysen:
Since `c2rust transpile` does not transpile conditionally-compiled code  (and we haven't integrated Hayroll yet), LLMs currently (and logically) don't insert comments from that conditionally-compiled code that haven't been transpiled into Rust.  In order for the comparison checking to work, direct the LLM to still insert these comments where that Rust would've gone.  This allows us to run successfully on most of `json-c` (those caches are checked in), although there are still a few other remaining bugs.

### DIFF
--- a/c2rust-postprocess/tests/llm-cache/CommentTransfer/_json_c_set_last_err/9ade63e9af53ecf8c488a382844137e74be4f51ab55c952a8cec0760a15edb10/metadata.toml
+++ b/c2rust-postprocess/tests/llm-cache/CommentTransfer/_json_c_set_last_err/9ade63e9af53ecf8c488a382844137e74be4f51ab55c952a8cec0760a15edb10/metadata.toml
@@ -1,0 +1,65 @@
+transform = "CommentTransfer"
+identifier = "_json_c_set_last_err"
+model = "gemini-3-pro-preview"
+response = """```rust
+#[no_mangle]
+#[c2rust::src_loc = "75:1"]
+pub unsafe extern "C" fn _json_c_set_last_err(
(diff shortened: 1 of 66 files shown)


thedataking (commented):
I'm not convinced this is the way to go. At least not long term.

We have the compile commands, so we could tell the LLM what environment the function was compiled with so it can determine what code is guarded. We could use that information to exclude guarded comments from the validation function.

Are you suggesting this as a short-term fix to make tests pass? 
If so, please put in some TODO comments... specially the LLM prompt to remind ourselves that this is a workaround that needs addressing. 

Edit to unpack my thinking a bit more: I'm working on a PR that adds a pre-processing step that removes gunk before or around a function. I think it is best to have a sequence of very simple LLM calls that are easy to validate. One is the gunk removal (e.g. license comment at top of file), another could be removal of guarded code. If we ask the LLM to do a lot in one go so we will not be able to use some of the simpler and less capable ones and validation gets more involved.

thedataking (changes requested):
Please preserve logic in `remove_backticks` (don't care what you name it) but it should not require backticks to be present, just remove them if they are. Sometimes the LLM emits them (with or without the `rust` part), sometimes, not. We don't want to pay for another roundtip (in time or API costs) if we can just remove backticks if present.

thedataking:
Update from meeting with Galois 1/7/2026: handling conditional compilation is a general problem so we should find a general approach that works not just for c2rust-postprocess but elsewhere too. We can denylist problematic functions for now and once a robust and general approach (probably at the MVIR tools level) is available, we will switch to that.

PR #1535, opened 2026-01-05
postprocess: switch from `gemini-3-pro-preview` to `gemini-3-flash-preview` for the default

kkysen:
`flash` is cheaper and faster and seems to still be sufficiently capable.

~~I did have to adjust the prompt slightly to explicitly say to include the surrounding Rust Markdown code block.  This isn't actually used right now, but it is in #1534, so it seems worth preserving.~~

Edit: The prompt is kept the same now.


thedataking (commented):
I don't think we should require the markdown code blocks. 

Sometimes you get them, sometimes you don't (even for same model), so the way to round-trip less is to just remove them when they're present rather than forcing the model to potentially re-do its work when they're not present.

PR #1540, opened 2026-01-08
transpile: replace glob import used for linking with a normal import

ahomescu:
Remove the one instance of glob imports from auto-generated code, as well as any helper code that generates them.

### DIFF
--- a/c2rust-refactor/src/ast_builder/builder.rs
+++ b/c2rust-refactor/src/ast_builder/builder.rs
@@ -1955,26 +1955,6 @@ impl Builder {
         )
     }
 
-    pub fn use_glob_item<Pa>(self, path: Pa) -> P<Item>
-    where
-        Pa: Make<Path>,
-    {
-        let path = path.make(&self);
-        let use_tree = UseTree {
-            span: DUMMY_SP,
-            prefix: path,
-            kind: UseTreeKind::Glob,
-        };
-        Self::item(
-            Ident::empty(),
-            self.attrs,
-            self.vis,
-            self.span,
-            self.id,
-            ItemKind::Use(use_tree),
-        )
-    }
-
     pub fn foreign_items(self, items: Vec<P<ForeignItem>>) -> P<Item> {
         let fgn_mod = ForeignMod {
             unsafety: self.unsafety,


c2rust-transpile/src/translator/mod.rs
```diff
@@ -1246,7 +1246,7 @@ fn arrange_header(t: &Translation, is_binary: bool) -> (Vec<syn::Attribute>, Vec
                 }
             }
 
-            out_items.push(mk().use_glob_item(mk().abs_path(vec![&t.tcfg.crate_name()])));
+            out_items.push(mk().extern_crate_item(t.tcfg.crate_name().clone(), None));
```
kkysen: This is replacing the glob with an `extern crate`?

kkysen: What are we trying to get in-scope here?

ahomescu: `lib.rs` from the current crate into the binary. For some reason it's not automatically imported, so we have to do it explicitly. Otherwise we get a bunch of linker errors.

kkysen: Hmm.  Can we just import the specific items that are actually used from `lib.rs` in the binary `main.rs` (or whatever it's named)?  `extern crate repo` seems even worse than `use repo::*`.

ahomescu: > `extern crate repo` seems even worse than `use repo::*`.

Why? It used to be common pre-2024 edition.

> Can we just import the specific items that are actually used

We don't know which ones those are, many of those are imported via the FFI.

kkysen: It's an edition 2015 or 2018 thing. It's never or very rarely used in edition 2021, which is what we're generating.

kkysen: We discussed this more offline.  A `use ::repo;` should suffice to link the functions without needing the `::*` glob.  Also, we can add an `#[allow(unused)]` to suppress the unused warning (could you also leave a comment here saying to switch to `#[expect(unused, reason = "...")]` once we upgrade our nightly?), and because I'm not sure if we can add a comment explaining why this is here in the output.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1246,7 +1246,15 @@ fn arrange_header(t: &Translation, is_binary: bool) -> (Vec<syn::Attribute>, Vec
                 }
             }
 
-            out_items.push(mk().use_glob_item(mk().abs_path(vec![&t.tcfg.crate_name()])));
+            // TODO: switch to `#[expect(unused_imports, reason = ...)]` once
+            // we upgrade to a newer nightly that supports it.
```
```suggestion
            // we upgrade to a newer nightly (Rust 1.81) that supports it.
```

Putting the version helps when upgrading, and searching for things that need upgrading.

kkysen: Can you use "Rust "?  That's how we generally write it elsewhere so it's easier to grep.

PR #1543, opened 2026-01-10
postprocess: support multiple transforms

thedataking:
Support multiple transforms such that we can support different post-processing steps such as:
- comment transfer
- assert folding
- relooper cleanup
- reformatting of large tables, etc.

Some transforms might also be broken down into a sequence of simpler transforms. A concrete use case is lua which contains license headers and other matter before function definitions which we do not want to pass so the comment transfer transform.

(Replaces PR #1512 as it was automatically closed when I renamed the branch)


c2rust-postprocess/postprocess/transforms/__init__.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any
+
+from postprocess.transforms.base import AbstractTransform
+from postprocess.transforms.comments import CommentsTransform
+
+
+def get_transform_by_id(id: str, **kwargs: Any) -> AbstractTransform:
```
kkysen: If we're only going to have one call "shape"/API, can we just make these args explicit (i.e., `cache: AbstractCache, model: AbstractGenerativeModel`)?  It makes the APIs a lot easier to understand.

c2rust-postprocess/postprocess/__init__.py
```diff
@@ -124,14 +131,25 @@ def main(argv: Sequence[str] | None = None):
...
+        # maximize cache hits even if the user passed them in a different order
+        transform_ids = sorted(
+            transform_id.strip()
+            for transform_id in args.transforms.split(",")
+            if transform_id.strip()
         )
+        transforms = [
+            get_transform_by_id(transform_id, cache=cache, model=model)
+            for transform_id in transform_ids
+        ]
```
kkysen, line 158: I think `argparse` supports making `--transform` a repeatable argument and restricting it to the available transform names we have.  That way when you do `--help`, it should show the list of possible transforms as well, which can be very useful.

c2rust-postprocess/postprocess/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+from postprocess.exclude_list import IdentifierExcludeList
+from postprocess.utils import get_highlighted_c
+
+
+class AbstractTransform(ABC):
+    """
+    Abstract base class for LLM-driven transforms of c2rust transpiler output.
+    """
+
+    def __init__(self, system_instruction: str, **kwargs: Any):
```
kkysen: Do we need `**kwargs` if we're not using it at all?  Subclasses can have different `__init__` signatures.

c2rust-postprocess/postprocess/transforms/base.py
```diff
@@ -0,0 +1,122 @@
...
+        self._system_instruction = system_instruction
+
+    @property
+    def system_instruction(self) -> str:
+        return self._system_instruction
+
+    # TODO: update signature after factoring out common functionality such
+    #       as matching up definitions, filtering out definitions, etc.
+    @abstractmethod
+    def apply_item(
```
kkysen: Is `item` supposed to be a function, constant, etc?  We use `identifier`/`ident` elsewhere for this terminology, so we might want to match/switch that or be clear about that here in a comment/docstring.

c2rust-postprocess/postprocess/transforms/comments.py
```diff
@@ -53,6 +54,47 @@ def __init__(self, cache: AbstractCache, model: AbstractGenerativeModel):
...
+                    diagnostic += (
+                        f"\nMissing comment in Rust function: {c_comment}. Insert it!"
+                    )
+
+            c_comments_set = set(c_comments)
+            for rust_comment in rust_comments:
+                if rust_comment not in c_comments_set:
+                    diagnostic += (
+                        f"\nExtra comment in Rust function: {rust_comment}. Remove it!"
+                    )
```
kkysen, line 96: I think showing a diff (which #1553 does) is clearer.

c2rust-postprocess/tests/llm-cache/CommentsTransform/partition/bf27c23abbb2f5c2d7dba1891b748252983959f5dc232f35146250b6d2d1fb29/metadata.toml
```diff
@@ -42,11 +39,6 @@ Respond with the Rust function definition with the transferred comments; say not
 
 C function:
 ```c
-/*
- * Lomuto Partition Scheme:
- * Partitions the array so that elements < pivot are on the left,
- * and elements >= pivot are on the right.
- */
```
kkysen: Why are we removing this valid doc comment for `partition`?

c2rust-postprocess/tests/llm-cache/CommentTransfer/partition/bf27c23abbb2f5c2d7dba1891b748252983959f5dc232f35146250b6d2d1fb29/metadata.toml
```diff
@@ -1,16 +1,13 @@
...
-/// Partitions the array so that elements < pivot are on the left, 
-/// and elements >= pivot are on the right.
+response = """/// Partition the subarray around the last element as pivot and return pivot's final index.
 #[no_mangle]
 pub unsafe extern "C" fn partition(
     mut arr: *mut ::core::ffi::c_int,
     mut low: ::core::ffi::c_int,
     mut high: ::core::ffi::c_int,
 ) -> ::core::ffi::c_int {
-    // Partition the subarray around the last element as pivot and return pivot's final index.
```
kkysen, line 13: This seems like a worse translation.  That comment is supposed to be inside the function, and now we're missing the comment on/outside the function.

c2rust-postprocess/tests/llm-cache/TrimTransform/partition/a4e1c68bf25ff0d12996deed99aff2f12fb77a27642c356431d8fb7bd012b82e/response.txt
kkysen, line 1: We need to delete the previous version of this file (which is what #1554 makes easy to do) so that git will show this as a moved file with a diff.  This makes it much easier to review the changes.

c2rust-postprocess/postprocess/transforms/asserts.py
```diff
@@ -0,0 +1,165 @@
...
+            + "```\n\n"
+            + "Rust function:\n```rust\n"
+            + self.rust_function
+            + "```\n"
+        )
+
+
+class AssertsTransform(AbstractTransform):
+    def __init__(self, cache: AbstractCache, model: AbstractGenerativeModel):
+        super().__init__(SYSTEM_INSTRUCTION)
```
kkysen, line 49: Making `SYSTEM_INSTRUCTION` an abstract method might be simpler and more idiomatic.

kkysen (changes requested):
So this PR does a lot of different things, which makes it harder to review altogether at once.  Could we split this into a few separate PRs?

* 8a3e399 seems totally independent.  We can land that on its own without any issue.
* 9595e66 and dde9bea are about supporting multiple transforms, which is what I think this PR was originally titled for.
* 73069ee can be on its own, too, and there are parts of it that also overlap with the diffs in #1553 that I think we can better untangle if it's its own PR.
* 00cb8ec and 0f869ac seem like their own transforms that can be their own PRs.

Also, I don't think the LLM cache is being fully updated on each commit, which makes it much harder to review, as then I can't fully see what that commit changes, as the cache is basically the postprocessor's snapshot tests.

c2rust-postprocess/postprocess/transforms/comments.py
thedataking: PR #1553 says:
> Prints diffs of the C vs. Rust comments when they don't match so we can see what's going wrong.

but the feedback is not for us; it is for the language model. I've done a little experimenting but not enough to know what works best. I expect it is model dependent. I've added a TODO so we can move on and get this PR landed.

thedataking:
Landing this work in multiple smaller PRs per review feedback.

PR #1544, opened 2026-01-10
Fix issue #1509 by improving handling of unused expression side effects

### LINKED ISSUE #1507: About the transaction output bug
Original C code:

```c
assert(writechar('A', &len) == 1);  
assert(len == 1);                    
```

c2rust output:

```rust
if writechar('A'...) == 1 {} else { __assert_fail(...); }
'c_2796: {
    if writechar('A'...) == 1 {} else { __assert_fail(...); }  // ❌ repeat twice!
};
if len == 1 {} else { __assert_fail(...); }
```

There is a bug!

### LINKED ISSUE #1509: About the mistranslation
**Original C Code:**

```c
int main(void) {
    assert(check_add());    // once 
    assert(check_sub());    // once
    return 0;
}
```

**C2rust Output:**

```rust
unsafe fn main_0() -> core::ffi::c_int {
    // first
    if check_add() {} else {
        __assert_fail(...);
    }
    
    // 🔴 second !!!
    'c_1569: {
        if check_add() {} else {
            __assert_fail(...);
        }
    };
    
    // first
    if check_sub() {} else {
        __assert_fail(...);
    }
    
    // 🔴 second !!!
    'c_1529: {
        if check_sub() {} else {
            __assert_fail(...);
        }
    };
    
    return 0 as core::ffi::c_int;
}
```

ahomescu:
* Fixes #1507.
* Fixes #1509.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile-macos@call_only_once.c.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile-macos@call_only_once.c.snap
@@ -1,0 +1,39 @@
+---
+source: c2rust-transpile/tests/snapshots.rs
+expression: cat tests/snapshots/os-specific/call_only_once.macos.rs
+input_file: c2rust-transpile/tests/snapshots/os-specific/call_only_once.c
+---
+#![allow(
+    dead_code,
+    non_camel_case_types,
+    non_snake_case,
+    non_upper_case_globals,
+    unused_assignments,
+    unused_mut
+)]
+extern "C" {
+    fn __assert_rtn(
+        _: *const ::core::ffi::c_char,
+        _: *const ::core::ffi::c_char,
+        _: ::core::ffi::c_int,
+        _: *const ::core::ffi::c_char,
+    ) -> !;
+}
+static mut called: ::core::ffi::c_int = 0 as ::core::ffi::c_int;
+unsafe extern "C" fn called_only_once() -> ::core::ffi::c_int {
+    called += 1;
+    return 1 as ::core::ffi::c_int;
+}
+#[no_mangle]
+pub unsafe extern "C" fn assert_call_only_once() -> ::core::ffi::c_int {
+    if (called_only_once() == 0) as ::core::ffi::c_int as ::core::ffi::c_long != 0 {
+        __assert_rtn(
+            b"assert_call_only_once\0" as *const u8 as *const ::core::ffi::c_char,
+            b"call_only_once.c\0" as *const u8 as *const ::core::ffi::c_char,
+            12 as ::core::ffi::c_int,
+            b"called_only_once()\0" as *const u8 as *const ::core::ffi::c_char,
+        );
+    } else {
+    };
+    return called;
+}


tests/unit/misc/src/call_only_once.c
kkysen, line 1: Can we also just have a snapshot test? It should be easy to see the translation of the assert.

ahomescu:
@kkysen What are the steps for updating the macos snapshots for people running Linux?

c2rust-transpile/tests/snapshots/snapshots__transpile-linux@call_only_once.c.snap
kkysen, line 1: When we were transpiling dav1d, `assert(condition)`s would become a simple `if !condition { panic!(); }`.  What happened to that?  Did we change something, or did libc/clang change something?

ahomescu:
```rust
    if !(::core::mem::size_of::<Dav1dMemPoolBuffer>() as libc::c_ulong
        <= 64 as libc::c_int as libc::c_ulong)
    {
        unreachable!();
    }
```
This is what I found in rav1d for one `assert` call. I have no idea why it has `unreachable!` instead of `__assert_fail`.

ahomescu: https://github.com/memorysafety/rav1d/blob/main/src/include/common/attributes.h#L130 oh that's why!

kkysen: Ahh.  That's a shame.  That's a much easier translation to turn into `assert!`.

ahomescu:
This passes the tests and has snapshots now, any more feedback?

c2rust-transpile/tests/snapshots.rs
```diff
@@ -95,6 +95,9 @@ fn transpile(platform: Option<&str>, c_path: &Path) {
     let rs = fs::read_to_string(&rs_path).unwrap();
     let debug_expr = format!("cat {}", rs_path.display());
 
+    // Replace real paths with placeholders
+    let rs = rs.replace(cwd.to_str().unwrap(), "/c2rust-transpile");
```
kkysen:
```suggestion
    // Make absolute path strings relative
    // so that they compare the same no matter what the CWD is.
    let rs = rs.replace(cwd.to_str().unwrap(), ".");
```

I think it also make more sense to use `.` for replacing the CWD.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4222,7 +4223,7 @@ impl<'c> Translation<'c> {
                 let result_id = substmt_ids[n - 1];
 
                 let name = format!("<stmt-expr_{:?}>", compound_stmt_id);
-                let lbl = cfg::Label::FromC(compound_stmt_id, None);
+                let lbl = cfg::Label::FromC(compound_stmt_id, Some(Rc::from("c2rust_stmt_expr")));
```
kkysen: Should we use `.pick_name`/`.pick_name_in_scope` for this?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4265,10 +4269,19 @@ impl<'c> Translation<'c> {
...
-                Ok(WithStmts::new(stmts, val))
+                if ctx.is_unused() {
+                    Ok(WithStmts::new(
+                        vec![mk().semi_stmt(val)],
+                        self.panic_or_err(
+                            "Compound statement expression is not supposed to be used",
+                        ),
+                    ))
+                } else {
+                    Ok(WithStmts::new(vec![], val))
```
```suggestion
                    Ok(WithStmts::new_val(val))
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4243,20 +4244,23 @@ impl<'c> Translation<'c> {
...
+                                    self.panic_or_err(
+                                        "Compound statement expression is not supposed to be used",
+                                    ),
+                                ));
                             } else {
-                                Vec::new()
-                            };
-
-                            return Ok(WithStmts::new(stmts, val));
+                                return Ok(WithStmts::new(vec![], val));
```
```suggestion
                                return Ok(WithStmts::new_val(val));
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4265,10 +4269,19 @@ impl<'c> Translation<'c> {
...
+                if ctx.is_unused() {
+                    Ok(WithStmts::new(
+                        vec![mk().semi_stmt(val)],
+                        self.panic_or_err(
+                            "Compound statement expression is not supposed to be used",
+                        ),
+                    ))
+                } else {
+                    Ok(WithStmts::new(vec![], val))
+                }
```
```suggestion
                self.convert_side_effects_expr(
                    ctx,
                    WithStmts::new_val(val),
                    "Compound statement expression is not supposed to be used",
                )
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4243,20 +4244,23 @@ impl<'c> Translation<'c> {
...
                             let val = mk().paren_expr(block);
-                            let stmts = if ctx.is_unused() {
-                                vec![mk().semi_stmt(val.clone())]
+                            if ctx.is_unused() {
+                                return Ok(WithStmts::new(
+                                    vec![mk().semi_stmt(val)],
+                                    self.panic_or_err(
+                                        "Compound statement expression is not supposed to be used",
+                                    ),
+                                ));
```
```suggestion
                            return self.convert_side_effects_expr(
                                ctx,
                                WithStmts::new_val(val),
                                "Compound statement expression is not supposed to be used",
                            );
```

c2rust-transpile/src/translator/operators.rs
```diff
@@ -931,17 +931,24 @@ impl<'c> Translation<'c> {
 
         // Some unused unary operators (`-foo()`) may have side effects, so we need
         // to add them to stmts when name is not increment/decrement operator.
+        //
+        // `Extension` is another exception since it's a no-op around the inner expression.
```
```suggestion
        // `UnOp::Extension` (`__extension__`) is another exception since it's a no-op around the inner expression.
```

c2rust-transpile/src/translator/operators.rs
```diff
@@ -931,17 +931,24 @@ impl<'c> Translation<'c> {
...
             )
         {
-            let v = unary.clone().into_value();
-            unary.add_stmt(mk().semi_stmt(v));
+            unary = unary.and_then(|v| {
+                TranslationResult::Ok(WithStmts::new(
+                    vec![mk().semi_stmt(v)],
+                    self.panic_or_err("Unary expression is not supposed to be used"),
+                ))
+            })?;
```
```suggestion
            unary = self.convert_side_effects_expr(
                ctx,
                unary,
                "Unary expression is not supposed to be used",
            )?;
```

You can also get rid of the `ctx.is_unused() &&` since `self.convert_side_effects_expr` checks that already.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -931,17 +931,24 @@ impl<'c> Translation<'c> {
...
+        // `Extension` is another exception since it's a no-op around the inner expression.
         if ctx.is_unused()
             && !matches!(
                 name,
                 c_ast::UnOp::PreDecrement
                     | c_ast::UnOp::PreIncrement
                     | c_ast::UnOp::PostDecrement
                     | c_ast::UnOp::PostIncrement
+                    | c_ast::UnOp::Extension
             )
```
kkysen: Moving this to a `fn UnOp::may_have_side_effects` (and some of the comments) would help clarity I think.

ahomescu: It seems excessive to move a single `if` statement into a separate function. We should split it once it's used in 2-3 different places.

kkysen: Perhaps.  I find it clearer when separate, but not a big deal.  This is fine.

kkysen (approved):
Ideally, we'd split the macos snapshot across the commits like is done for the linux snapshot now, but that's a big pain, and the tests should still pass for each commit on Linux now, so I think it's fine.

PR #1547, opened 2026-01-14
tests/integration: print times for each step and flush to see it in CI

kkysen:
@fw-immunant, this should hopefully fix why it wasn't printing each step until everything was finished.  I also added times for each step (in addition to the total times for each project that we print at the end).


fw-immunant:
> > That said, I don't know if additional explicit timestamps are really necessary given that GH Actions tracks them separately, but it's not a problem to have them.
> 
> Wait where does it track them?

Click "Show timestamps" under the gear icon on the top-right of a CI job log.

PR #1554, opened 2026-01-20
postprocess: add `--gc-cache` to garbage collect any cache entries not used since the last `--gc-cache` call

kkysen:
When there are input changes (like transpiler, refactorer, prompt, etc. changes), the cache becomes outdated and must be recalculated, but the previous entries aren't deleted. This tries to solve that.  It tracks which cache entries are still actively tested/used (this is always done in `llm-cache/.gc`), and then `--gc-cache` deletes everything else. So the normal intended usage is to:
* Run `rm -f llm-cache/.gc`.
* Run all tests, updating the cache with new entries.
* Run with `--gc-cache` to remove the outdated, unused entries.

I tried to do this in a simple way, but it does seem pretty necessary once testing is added to CI and others will need to do the same on their own, instead of me manually deleting the right outdated cache entries.  If there are better/simpler ways to do this, that would also be great.

### DIFF
--- a/c2rust-postprocess/README.md
+++ b/c2rust-postprocess/README.md
@@ -50,6 +50,28 @@ because the C file may not be available and isn't read by `c2rust-postprocess`
 These file paths and function identifiers must match fully.
 They are not regexes or globs.
 
+## LLM Cache
+
+`c2rust-postprocess` caches its LLM responses in `llm-cache/`.
+This is meant to be checked into git so that its shareable, including in CI
+so that we don't have to run a slow, expensive LLM in CI.
+
+### `--gc-cache`
+
+When updating the cache, which may happen when `c2rust-postprocess` is changed,
+or `c2rust-transpile` or `c2rust-refactor` as well, since they affect
+the input to `c2rust-postprocess`, new files are added to the cache.
+However, the old, now unused cache files that they replace remain in the cache.
+To clean these, `--gc-cache` can be used, which deletes all cache files
+that have not been used since the last time `--gc-cache` was used.
+With these deleted, git registers the new files as having moved,
+allowing you to see the diff between the two cache files, which is very useful for review.
+
+The intended way to use `--gc-cache` is to run all of your tests
+and on the last one, add `--gc-cache`.
+This will touch all of the cache files that are actually being used
+and then clean up the ones that are now unused.
+
 # Testing
 
 ## Test prerequisites


kkysen (commented):
I've detached `--gc-cache` from `CommentTransferOptions` now (it was never really needed there) and switched to an mtime-based version, but I haven't gotten around to fully separating `--gc-cache` into a separate script yet, which requires some more work, as part of the script will have to be deduplicated/set up to be imported.

kkysen (commented):
This is now independent of #1553 and rebased on `master`, so it can merge independently.

Also, the mtime and ctime-based implementation was broken because apparently on Linux, ctime is change time, not creation time, so I've switched to a purely mtime-based implementation now and tested it.

c2rust-postprocess/postprocess/cache.py
```diff
@@ -216,11 +249,92 @@ def update(
...
+            removed_any = False
+            for path in dir.iterdir():
+                if path.is_dir():
+                    if walk(path):
+                        removed_any = True
+                else:
+                    if path.stat().st_mtime_ns < oldest_allowed:
+                        try:
+                            path.unlink()
+                            print(f"rm {rel_path(path)}")
```
ahomescu: Do you still want these in?

c2rust-postprocess/postprocess/cache.py
```diff
@@ -216,11 +249,92 @@ def update(
...
+                return False
+
+            # If we have removed something, try to delete the dir.
+            # This only succeeds if the dir is empty, which is what we want.
+            try:
+                dir.rmdir()
+                print(f"rmdir {rel_path(dir)}")
+                return True
+            except OSError as e:
+                if e.errno == ENOTEMPTY:
```
ahomescu, line 319:
```suggestion
                if e.errno != ENOTEMPTY:
                    logging.warning(f"gc_sweep: failed to rmdir {dir}: {e}")                
```

c2rust-postprocess/postprocess/cache.py
```diff
@@ -75,6 +77,19 @@ def flush(self) -> None:  # noqa: B027
         """
         pass
 
+    def gc_init(self) -> None:
```
ahomescu, line 80: Where is this called from?

c2rust-postprocess/postprocess/cache.py
ahomescu: Or at least a `log` call instead of `print`?

kkysen: Yup, good idea.  Fixed now.

c2rust-postprocess/postprocess/cache.py
kkysen: Good catch; fixed now.

c2rust-postprocess/postprocess/cache.py
kkysen: We discussed, and I wanted to emphasize that we're expecting the `ENOTEMPTY` error and skipping that.

c2rust-postprocess/postprocess/__init__.py
```diff
@@ -142,6 +153,9 @@ def main(argv: Sequence[str] | None = None):
             update_rust=args.update_rust,
         )
 
+        if args.gc_cache:
+            cache.gc_sweep()
```
kkysen, line 157: If we want to separate this into a separate script, we'll have to separate out this `cache.gc_sweep()` call, the `cache` creation (`cache = getattr(DirectoryCache, args.cache_scope)()`), and thus `args.cache_scope` as well.  So we'd have to duplicate a bunch of stuff here, or refactor things such that we can import them and only them (like only `--cache-scope`, not the other arguments from `build_arg_parser`).

Issue #1523, opened 2025-12-18
Generate an enum for current_block labels


randomPoison:
Today when we emit `current_block` control flow, we use arbitrary numeric constants as the values for `current_block`. We could instead generate an enum with named variants, using label names taken from the CFG instead of arbitrary numeric constants.

# Details

Today, `current_block` control flow looks something like this

```rust
let mut current_block;
if cond {
    current_block = 123456789;
} else {
    current_block = 987654321;
}
match current_block {
    123456789 => { ... }
    _ => { ... }
}
```

Where we're using the arbitrary numeric constants `123456789` and `987654321` as labels to keep track of and emulate the original C function's control flow. These numeric constants are currently hard to work with for a couple of reasons:

* We don't always reference the number explicitly in code, making it hard to see where a particular label value is actually used. See https://github.com/immunant/c2rust/issues/1522.
* The numbers are arbitrary and have no meaning in relation to the original source code.
* We generate really large numbers that are hard to visually parse.

We can address the latter two issues by generating an enum to use as the type for `current_block`. For example:

```rust
enum CurrentBlock {
    A,
    B,
}
let mut current_block;
if cond {
    current_block = CurrentBlock::A;
} else {
    current_block = CurrentBlock::B;
}
match current_block {
    CurrentBlock::A => { ... }
    CurrentBlock::B => { ... }
}
```

We should generally be able to use the CFG node labels as the variant names for the generated enum, or we can synthetically name them along the lines or `A`, `B`, `C`, etc.

Rua:
While investigating how this might be implemented, I realised that c2rust generates one `match` expression for every `Structure::Multiple`, but is it guaranteed that there is only one? Otherwise, it may be necessary to have multiple `current_block` variables, since they would have different enum types.

randomPoison:
> While investigating how this might be implemented, I realised that c2rust generates one `match` expression for every `Structure::Multiple`, but is it guaranteed that there is only one?

Nope! We can absolutely have multiple `match current_block` in a single function! With the changes in #1558 we generate way fewer `match`es than we do today on master, but it's still very much possible to have multiple instances of irreducible control flow in a single function. In that case, both `match current_block`s are going to have completely different sets of entries, and it would potentially be beneficial to generate a separate enum for each one. We don't have to start there though, and I think your implementation in #1557 is a reasonable starting point.

PR #1556, opened 2026-01-21
transpile: Always explicitly reference `current_block` values

### LINKED ISSUE #1522: Always explicitly reference current_block values
Today, when we generate `match current_block {}` expressions, we pick one of the arms of the match to be the `_` case in order satisfy `match`'s exhaustiveness requirement. This makes reasoning about `current_block` control flow difficult because you can't search for the value of a `current_block` constant to see where it's used.

Instead, we should have an arm that explicitly references the `current_block` value, and then a `_ => unreachable!()` arm to satisfy exhaustiveness.

# Details

Today, when we emit `current_block`, we generate code that looks a bit like this:

```rust
let mut current_block;
if cond {
    current_block = 123456789;
} else {
    current_block = 987654321;
}
match current_block {
    123456789 => { ... }
    _ => { ... }
}
```

If I want to figure out what code each `current_block` value leads to (in practice setting `current_block` may happen very far away from where we `match` on it), I can search the code for `123456789` and see both where it's set and where it's used. But the other label `987654321` doesn't show up in the `match` because it's instead covered by the `_` case.

We do this to satisfy exhaustiveness: `current_block` is just an integer, so if we did

```rust
match current_block {
    123456789 => { ... }
    987654321 => { ... }
}
```

The compiler would complain that our `match` isn't exhaustive. Since we control all the possible values that `current_block` will have, it's valid for us to arbitrarily pick one of these arms and make it the `_` case. However, as noted above, this harms readability and makes it harder to reason about how `current_block` is being used.

Instead we should do this

```rust
match current_block {
    123456789 => { ... }
    987654321 => { ... }
    _ => unreachable!(),
}
```

This would ensure that you can always see all the places where a `current_block` value is used, which would make reasoning about `current_block` control flow easier.

Rua:
- Depends on #1523
- Fixes #1522

I've assumed here that the `entries` field of `Structure::Multiple` is exhaustive. Thus all labels in `entries` minus those in `branches` should be those that were previously covered by the wildcard pattern, and any other values are unreachable. I don't know if that's the case or not; if not, then this may turn out much harder to implement, as I don't know where to recover the missing labels from.

### DIFF
--- a/c2rust-transpile/src/cfg/structures.rs
+++ b/c2rust-transpile/src/cfg/structures.rs
@@ -2,7 +2,8 @@
 
 use super::*;
 use log::warn;
-use syn::{spanned::Spanned as _, ExprBinary, ExprBreak, ExprIf, ExprUnary, Stmt};
+use proc_macro2::TokenStream;
+use syn::{spanned::Spanned as _, ExprBinary, ExprBreak, ExprIf, ExprUnary, MacroDelimiter, Stmt};
 
 use crate::rust_ast::{comment_store, set_span::SetSpan, BytePos, SpanExt};
 
@@ -649,28 +650,12 @@ fn process_cfg(
                 S::mk_loop(Some(label.clone()), body)
             }
 
-            Multiple { entries, branches } => {
-                // If we have entries that aren't one of our branches, our `then` case needs to
-                // be empty so that we fall through to the next structure. Otherwise we pull off
-                // the first branch as the `then` case in order to satisfy the exhaustiveness
-                // requirement for the generated `match`.
-                let mut branch_entries = branches.keys();
-                let then = if entries.iter().all(|entry| branches.contains_key(entry)) {
-                    let then_entry = branch_entries
-                        .next()
-                        .expect("There must be at least one branch");
-                    process_cfg(
-                        &branches[then_entry],
-                        cfg_info,
-                        &next_entries,
-                        loop_context,
-                        break_targets,
-                    )?
-                } else {
-                    S::empty()
-                };
-
-                let cases = branch_entries
+            Multiple {
+                entries: _,
+                branches,
+            } => {
+                let cases = branches
+                    .keys()
                     .map(|entry| {
                         let stmts = process_cfg(
                             &branches[entry],
@@ -683,7 +668,27 @@ fn process_cfg(
                     })
                     .collect::<TranslationResult<_>>()?;
 
-                S::mk_goto_table(cases, Some(then))
+                let then = if cfg_info
+                    .checked_entries
+                    .iter()
+                    .all(|entry| branches.contains_key(entry))
+                {
+                    // If the branches cover all the enum variants in `checked_entries`, then the
+                    // `match` is exhaustive and we don't need a "then" branch.
+                    None
+                } else {
+                    // Otherwise, the branches *should* at least cover all of our entries, so there
+                    // should be no other possible cases. Mark the "then" branch as unreachable.
+                    // See https://github.com/immunant/c2rust/pull/1556#discussion_r3874230125
+                    let unreachable_call = mk().mac_expr(mk().mac(
+                        mk().path(vec!["unreachable"]),
+                        TokenStream::new(),
+                        MacroDelimiter::Paren(Default::default()),
+                    ));
+                    Some(S::mk_singleton(mk().expr_stmt(unreachable_call)))
+                };
+
+                S::mk_goto_table(cases, then)
             }
         };
 
--- a/c2rust-transpile/src/cfg/structures.rs
+++ b/c2rust-transpile/src/cfg/structures.rs
@@ -165,7 +165,10 @@ fn cleanup_labels(
             for (_, case) in cases {
                 cleanup_labels(case, current_loop, encountered_labels);
             }
-            cleanup_labels(then, current_loop, encountered_labels);
+
+            if let Some(then) = then {
+                cleanup_labels(then, current_loop, encountered_labels);
+            }
         }
         Empty | Singleton(_) | Goto(_) => {}
     }
@@ -208,7 +211,10 @@ fn merge_labels(ast: &mut StructuredAST<Box<Expr>, Pat, Label, Stmt>, old: &Labe
             for (_, case) in cases {
                 merge_labels(case, old, new);
             }
-            merge_labels(then, old, new);
+
+            if let Some(then) = then {
+                merge_labels(then, old, new);
+            }
         }
         Exit(_, None) | Empty | Singleton(_) | Goto(_) => {}
     }
@@ -262,7 +268,7 @@ pub trait StructuredStatement: Sized {
     /// Make a `goto` table
     fn mk_goto_table(
         cases: Vec<(Self::L, Self)>, // entries in the goto table
-        then: Self,                  // default case of the goto table
+        then: Option<Self>,          // default case of the goto table
     ) -> Self;
 
     /// Make some sort of loop
@@ -330,7 +336,7 @@ pub enum StructuredASTKind<E, P, L, S> {
     ),
     GotoTable(
         Vec<(L, StructuredAST<E, P, L, S>)>,
-        Box<StructuredAST<E, P, L, S>>,
+        Option<Box<StructuredAST<E, P, L, S>>>,
     ),
     Loop(Option<L>, Box<StructuredAST<E, P, L, S>>),
     Block(L, Box<StructuredAST<E, P, L, S>>),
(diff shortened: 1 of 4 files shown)


c2rust-transpile/src/cfg/structures.rs
```diff
@@ -129,10 +131,11 @@ pub enum StructuredASTKind<E, P, L, S> {
         Box<StructuredAST<E, P, L, S>>,
         Box<StructuredAST<E, P, L, S>>,
     ),
-    GotoTable(
-        Vec<(L, StructuredAST<E, P, L, S>)>,
-        Box<StructuredAST<E, P, L, S>>,
-    ),
+    GotoTable {
+        cases: Vec<(L, StructuredAST<E, P, L, S>)>,
+        then_lbls: Vec<L>,
```
randomPoison: I don't think we need to keep a set of labels for the then arm. iiuc you're combining all of these as an `|` pattern in order to satisfy the exhaustiveness requirement. But this has the drawback that all of the unused labels show up in the generated `match`, which means that if you're looking at some code that does `current_block = Foo;` and you want to find where `Foo` is used, it's going to show up in `match current_block` expressions where it's not actually a valid arm. I think it'd be better to always use `_` as the then pattern.

c2rust-transpile/src/cfg/structures.rs
```diff
@@ -129,10 +131,11 @@ pub enum StructuredASTKind<E, P, L, S> {
...
+        then_body: Box<StructuredAST<E, P, L, S>>,
```
randomPoison: I don't think we need the `then_body` to be a full AST node. There are only 3 possible cases:

- No then branch, i.e. we cover all cases and already satisfy exhaustiveness. This happens if there's only one `match current_block` in the function.
- Unreachable then branch, i.e. our `match` doesn't cover all of the labels. This happens if we have multiple `match current_block` in the function, since each will have different labels.
  - Aside: Down the line we may want to handle this by generating a different enum for each `match current_block`, but I think starting with a single enum per function is the right first step.
- Empty then branch, i.e. we want to be able to fall out of the `match` without running any code. This is representative of fallthrough control flow, either from a `switch` or from `goto`.

So I think maybe something like this would suffice:

```rust
enum Then {
    Empty,
    Unreachable,
}

GotoTable {
    ..
    then: Option<Then>,
}
```

Note that this is also dependent on the changes in #1558, where I removed the `then` field from `Multiple`, so this may be easier to fix after rebasing on top of those changes.

randomPoison (changes requested):
Hello! First off, thank you for putting this together 💜 For the most part this looks good, I just have a couple of comments about how we can simplify things and make the output a bit easier to work with.

Unfortunately, I've currently got a PR up that heavily reworks the relooper implementation (https://github.com/immunant/c2rust/pull/1558), and I think it would be best to wait to merge this until after that refactoring lands. The changes in that PR shouldn't conflict too heavily with the changes you have here from what I can see.

c2rust-transpile/src/cfg/structures.rs
Rua: I haven't changed the `then_body`. It was previously the second element of a tuple variant, my changes just gave it a name for the sake of clarity. `then_lbls` is the new element I added.

c2rust-transpile/src/cfg/structures.rs
Rua: The PR is about adding explicit references in the match arms. If I use `_` instead, doesn't that just keep the original situation? Or am I misunderstanding something?

randomPoison: The goal of #1522 is that the `_` case will either be `unreachable!()` if all entries are present in the list of branches, or empty (`{}`) if there are unhandled entries. That latter case comes from fallthrough patterns in the original C, e.g.

```c
switch (x) {
case 0:
  x++;
case 1:
  x++;
case 2:
  x++;
}
```

With your PR we get a `match` that looks like this:

```rust
match current_block_2 {
    12729336616994636022 => {
        x += 1;
        current_block_2 = 6386683298811446793;
    }
    6386683298811446793 | 17778012151635330486 => {}
    _ => unreachable!(),
}
```

What I intended with #1522 would be more like this:

```rust
match current_block_2 {
    12729336616994636022 => {
        x += 1;
        current_block_2 = 6386683298811446793;
    }
    _ => {},
}
```

That is, all of the **handled** entries show up as match arms, but if there are unhandled entries we should still have an empty wildcard case. So my comment was that we shouldn't need a pattern for the `then` of the goto table because the `then` pattern should always be `_`, and the only difference is whether the body of that arm is `{}` or `unreachable!()`. 

There is an advantage to your approach, which is that `current_block` values that are never handled still show up explicitly in a `{}` arm, but the changes in #1558 make that irrelevant because we no longer generate `match current_block` in those cases. With #1558 I think we will only be generating `match current_block` in cases where all entries to the `Multiple` are handled.

c2rust-transpile/src/cfg/structures.rs
randomPoison: Ah sorry, I think there's a bit of confusion around the intended behavior of #1522, I tried to clarify in [another comment](https://github.com/immunant/c2rust/pull/1556#discussion_r2719112498). As noted there, the goal is that the `_` case will either be either `unreachable!()` or `{}` depending on if there are unhandled entries to the `Multiple`. We want all other branches to be in the `cases` field, so we don't want a `then` field that's a full AST node. Though again, that's kind of a moot point because with #1558 we no longer generate `match current_block` for cases where a `Multiple` can have unhandled entries.

Rua:
I'll wait for #1558 so that I have a better idea what things look like after.

fw-immunant:
This might be less contentious if also phased after #1557, because the compiler will be able to reason about exhaustiveness of cases and there shouldn't be any situations where `_ => unreachable!()` is still necessary (though `_ => {}` could be).

Rua:
> This might be less contentious if also phased after #1557, because the compiler will be able to reason about exhaustiveness of cases and there shouldn't be any situations where `_ => unreachable!()` is still necessary

As I understand it, there may be more than one `match current_block` in a function, and they may not match on the same labels, but the current implementation of #1557 combines all labels into a single large enum. So the `unreachable` branch would still be necessary, and can only be removed once a separate enum is generated for each `match`.

Rua:
@randomPoison I've completely redone this PR and based it on #1557. It seems much more straightforward now?

c2rust-transpile/src/cfg/structures.rs
```diff
@@ -677,6 +665,31 @@ fn process_cfg(
...
+                    let unreachable_call = mk().mac_expr(mk().mac(
+                        mk().path(vec!["unreachable"]),
+                        TokenStream::new(),
+                        MacroDelimiter::Paren(Default::default()),
+                    ));
+                    Some(S::mk_singleton(mk().expr_stmt(unreachable_call)))
+                } else {
+                    // There are entries that are not covered, create an empty "then" branch to
+                    // fall through to the next structure.
+                    Some(S::empty())
```
randomPoison: Hm I'm not sure this case can actually be hit with the new relooper. The only case I know of where we would have entries that aren't covered by the branches are `switch` with fallthrough (or `goto` emulating that behavior), and that always gets translated to block breaks now.

Could you try inserting a `unreachable!` or something here and running the test suite to see if anything actually hits this? You can do that in a commit on this branch and let CI run if that's easier than running the whole integration suite locally.

randomPoison (changes requested):
The new implementation looks good! But we need a test case that actually involves the `_ => unreachable!()` branch, i.e. one that has multiple `match current_block` in the same fn:

<details><summary>Doubly-irreducible test case</summary>

```c
int double_irreducible(int x) {
    if (x % 2 == 0) {
        goto l1;
    } else {
        goto l2;
    }

    while (x < 10) {
l1:
        x += 1;

l2:
        x += 1;
    }

    if (x % 2 == 0) {
        goto a;
    } else {
        goto b;
    }

    while (x < 10) {
a:
        x += 1;

b:
        x += 1;
    }

    return x;
}
```

</details>

c2rust-transpile/src/cfg/structures.rs
```diff
@@ -685,9 +685,17 @@ fn process_cfg(
...
                     Some(S::mk_singleton(mk().expr_stmt(unreachable_call)))
                 } else {
+                    let unreachable_call = mk().mac_expr(mk().mac(
+                        mk().path(vec!["unreachable"]),
+                        TokenStream::from(proc_macro2::TokenTree::Literal(
+                            proc_macro2::Literal::string("Testing for the PR"),
+                        )),
+                        MacroDelimiter::Paren(Default::default()),
+                    ));
+                    Some(S::mk_singleton(mk().expr_stmt(unreachable_call)))
```
randomPoison, line 688: Ah, for testing purposes I meant do an `unreachable!()` in the transpiler itself, to see if this code path can ever be hit during translation. My understanding of relooper is that we'll never hit a case where we have entries that aren't in the branches of the `Multiple`

Rua: Oh right, that works too of course XD

Rua: Doesn't seem like it got tripped. So I could just remove the empty case altogether?

randomPoison: Oh cool, thanks for confirming! Yeah I think you can just turn that second branch into an `else` and remove the empty case entirely.

randomPoison (approved):
This looks good now! Thanks for putting this together 💜

fw-immunant:
This appears to reverse the order of match arms for the snapshot test. Is this intentional?

randomPoison:
Yeah the new ordering makes sense, and will probably be the more sensible layout in most cases. The reason why it reversed the order is because previously we were taking the first branch and moving it to the `_` case at the end. The new setup leaves the first branch at the top.

PR #1557, opened 2026-01-21
transpile: Use enum for `current_block` labels

### LINKED ISSUE #1523: Generate an enum for current_block labels

Rua:
- Fixes #1523

The enum variants now suffice for debugging purposes, so I've removed the `debug_labels` command line option.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -16,7 +16,7 @@ use proc_macro2::{Punct, Spacing::*, Span, TokenStream, TokenTree};
 use syn::spanned::Spanned as _;
 use syn::{
     AttrStyle, BareVariadic, Block, Expr, ExprBinary, ExprBlock, ExprBreak, ExprCast, ExprField,
-    ExprIndex, ExprParen, ExprUnary, FnArg, ForeignItem, ForeignItemFn, ForeignItemMacro,
+    ExprIndex, ExprParen, ExprUnary, Fields, FnArg, ForeignItem, ForeignItemFn, ForeignItemMacro,
     ForeignItemStatic, ForeignItemType, Ident, Item, ItemConst, ItemEnum, ItemExternCrate, ItemFn,
     ItemForeignMod, ItemImpl, ItemMacro, ItemMod, ItemStatic, ItemStruct, ItemTrait,
     ItemTraitAlias, ItemType, ItemUnion, ItemUse, Lit, MacroDelimiter, PathSegment, ReturnType,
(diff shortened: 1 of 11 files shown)


Rua:
I notice that this is failing because transpiled code is referencing enum variants that do not exist. The code currently relies on `get_current_block_labels` to return all the possible labels that could be assigned to `current_block`, which gets them from the union of all the `entries` fields of all `Structure::Multiple` instances. As noted in #1556 I assumed that `entries` covers all possible jump labels, so it's worrying that this does not work.

ahomescu:
I'm not sure how much we'll need this once the relooper changes land; @randomPoison thoughts?

randomPoison:
@ahomescu This still very much necessary! I specifically split out #1523 as work to do post-refactoring. But we will want to wait to land this until after https://github.com/immunant/c2rust/pull/1558 lands.

@Rua the issue you point out in your comment will be easier to address after #1558. In the reworked relooper algorithm we now have a pass that walks the structured CFG and gathers the set of all entries that need `current_block`. That should make it trivial to know which entries need to be in the enum.

randomPoison (commented):
This looks good to me, thank you for putting this together 💜 After looking over the code it's not clear to me why the test is failing, the way you are gathering the set of labels seems correct. But I'm not going to dig into this now because https://github.com/immunant/c2rust/pull/1558 is going to rework the relevant code anyway. Please wait until that PR lands and then rebase your changes. If things are still broken after that point I can help debug the issue 😸

ahomescu:
#1558 has landed, could you rebase this? We can do another review round once it passes the tests.

Rua:
@randomPoison I finally have the time and spoons to look over this. I rebased it but it's still failing on the same points. The snapshot tests aren't catching it, only the testsuite is. Specifically, something about the `PyEval_EvalFrameEx` function inside the Python integration test is tripping it up. But it's very hard to tell what's going on in that function because it's a mess of preprocessor stuff.

randomPoison:
@Rua The issue is in `get_current_block_labels`. When you examine a `Multiple`, you're adding its list of entries to your set of labels, but you're not recursively examining the contents of the `Multiple`'s branches. So if there are other `Multiple`s nested inside one of the branches, you're failing to find them.

Below is a minimal test case that hits this behavior, please add this as a snapshot test to your PR.

<details><summary>Minimal test case that exercises the bug</summary>

```c
int nested_irreducible(int x, int start_with_error) {
    if (x < 0) {
        if (start_with_error)
            goto exit;
    }
    if (start_with_error)
        goto error;
    goto next;

next:
    if (x == 1)
        goto error;
    x -= 1;

error:
    if (x > 0)
        goto next;
    goto exit;

exit:
    return x;
}
```

</details>

That said, I don't think we need your `get_current_block_labels` function at all. We are already have [`gather_cfg_info`](https://github.com/immunant/c2rust/blob/cc604dc84ccc74247a04c5a323c2e0e304ee538f/c2rust-transpile/src/cfg/structures.rs#L422) which gathers our **checked entries**, which are the ones that actually need to be represented in the `CurrentBlock` enum.

For context, before https://github.com/immunant/c2rust/pull/1558 we would use `current_block` for all `Multiple` entries, but after that refactoring we now generate labeled blocks/breaks in many cases. As a result, we don't need an entry in the `CurrentBlock` enum for all `Multiple` entries, we only need them for the checked entries as identified by `gather_cfg_info`. That info is already present in `convert_cfg` where you're generating the `CurrentBlock` enum, so you should be able to remove `get_current_block_labels` and use the checked entry labels already present in `cfg_info`.

Rua:
Ok, seems it finally works now. The CFG stuff is still a bit hocus pocus to me to be honest.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2406,26 +2406,31 @@ impl<'c> Translation<'c> {
         let mut cfg_info = cfg::structures::CfgInfo::default();
         cfg::structures::gather_cfg_info(&relooped, &mut cfg_info);
 
-        let current_block_ident = self
+        let current_block_enum = self
+            .renamer
+            .borrow_mut()
+            .pick_name("C2Rust_CurrentBlock", Namespaces::types());
```
fw-immunant: This is a bikeshed, but: the variable `current_block` is the *current* block; its gets assigned to control which blocks we jump to next, so the type for those would be more aptly called `C2Rust_Block`.

Rua: Fair point, I've renamed it.

fw-immunant (approved):
LGTM! This is much more readable, and I wonder if it optimizes better to branch on small enum values rather than 64b constants.

randomPoison (approved):
Looks great to me! Thanks for all your work putting this together 💜

Issue #330, opened 2021-03-05
Update the Relooper algorithm


RReverser:
C libraries often use `goto exit` / `goto error` as a way to jump within a function to a place that can cleanup resources before exiting when some condition is unmet.

Let me take this example (one of the simpler ones in the wild): https://github.com/kbarbary/sep/blob/d22bef88ded3c5b25f06584aa8a2bc931cad1826/ctest/test_image.c#L142-L212

```c
/* an extremely dumb reader for our specific test FITS file! */
int read_test_image(char *fname, float **data, int *nx, int *ny)
{
  FILE *f;
  char buf[80];  /* buffer to hold a line */
  long pos;
  size_t nbytes, nel, i, elsize, nread;
  char *rawbytes = NULL;
  short *rawdata;
  char tmp;
  int status = 0;
  float bscale, bzero;


  /* hard-code image size & element size */
  *nx = 256;
  *ny = 256;
  elsize = 2;
  nel = *nx * *ny;
  bscale = 2.92273835509;
  bzero = 3713.66692596;


  f = fopen(fname, "r");


  /* read first line and check if it is a FITS file */
  nread = fread(buf, 1, 80, f);
  if (nread != 80) { status = 1; goto exit; }
  if (strncmp(buf, "SIMPLE  =                    T", 30) != 0)
    {
      printf("not a FITS file");
      status = 1;
      goto exit;
    }


  /* read rows until we get to END keyword */
  while (strncmp(buf, "END", 3) != 0)
    {
      nread = fread(buf, 1, 80, f);
      if (nread != 80) { status = 1; goto exit; }
    }


  /* move to next 2880 byte boundary. */
  pos = ftell(f) % 2880;  /* position in "page" */
  if (pos != 0) fseek(f, 2880 - pos, SEEK_CUR);


  /* read raw data bytes */
  nbytes = nel * elsize;
  rawbytes = (char *)malloc(nbytes);
  nread = fread(rawbytes, 1, nbytes, f);
  if (nread != nbytes) { status = 1; goto exit; }


  /* swap bytes in raw data (FITS is big-endian) */
  for (i=0; i<nbytes; i+=2)
    {
      tmp = rawbytes[i];
      rawbytes[i] = rawbytes[i+1];
      rawbytes[i+1] = tmp;
    }
  rawdata = (short *)rawbytes; /* buffer is now little-endian */


  /* convert to float, applying bscale/bzero */
  *data = (float *)malloc(nel * sizeof(float));
[...]

PR #1558, opened 2026-01-21
Rework relooper algorithm for fun and profit

### LINKED ISSUE #1398: Automatically compile c2rust before running tests
When running the tests in this repo (i.e. not the ones in the separate c2rust-testsuite repo), you have to remember to manually run `cargo build --release` before running the tests or the tests will be run using the `c2rust` binary from the last time you built (though we do warn if `target/release/c2rust` isn't present at all). We should automatically run the build before running the tests to remove that manual step and make the local testing process less error-prone.

randomPoison:
This PR reworks the relooper algorithm to replace most usages of `current_block` with labeled blocks/breaks. Addresses #330. Closes https://github.com/immunant/c2rust/issues/1398.

The bulk of the changes are in `relooper.rs` and `structures.rs`:

- In `relooper.rs` the core logic is in the `relooper` function, which takes the unstructured CFG and processes it into the "structured CFG".
- In `structures.rs` we have the logic that takes the structured CFG and turns it into the "strucured AST", which is an intermediate AST that just represents the structured control flow we're translating.
  - I've replaced `structured_cfg_help` with a new `process_cfg` function that is responsible for generating the labeled blocks/breaks.
  - I've also added a secondary cleanup pass `cleanup_labels` that removes redundant labels and extra labeled blocks. This allows the logic in `process_cfg` to be simpler by conservatively labeling all blocks and breaks.

Don't try to review this commit-by-commit, the commit history is NOT clean and shows all of the experimentation and iteration I went through to get to this point. I only have all the commit history intact because I haven't had a need to squash yet, and we should squash this when merging to avoid polluting the commit history.

I'd also recommend not worrying too much about the diff here, and instead just review the new code as if it were all entirely fresh. Many of the parts here have been heavily reworked, and since I'm the only person deeply familiar with what the code was doing before it probably doesn't make sense to worry about how the new code differs from the old version.

~~NOTE: I still need to do a cleanup pass on the unit tests, since a lot of those were slapped together during experimentation. Snapshot tests are all clean, but I may want to add more as I cleanup the unit tests.~~

[218a281](https://github.com/immunant/c2rust/commit/218a2817c62364167202099ac22759ea3823126f) shows the diff between the old behavior and the new behavior in the snapshots.


randomPoison:
Oh joy, we have a subtle difference in behavior between platforms. Looks like the CFG labels that we generate on macos are slightly different than the labels we generate on Ubuntu, likely because the compiler we're using is different and is generating a slightly different CFG.

randomPoison:
Okay the tests should be clean and ready for review as well.

scripts/test_translator.py
```diff
@@ -662,11 +662,24 @@ def main() -> None:
...
+    build_args = ["build", "--release"]
+    retcode, stdout, stderr = cargo[build_args].run(retcode=None)
+
+    if retcode != 0:
+        logging.error("Build failed with return code %d", retcode)
+        logging.error("stdout: %s", stdout)
+        logging.error("stderr: %s", stderr)
+        die("cargo build --release failed", retcode)
+
+    logging.info("Build completed successfully")
```
kkysen, line 709: This is what #1398 tracks, right?

.gitignore
```diff
@@ -55,3 +55,7 @@ polonius_cache/
 
 # Outputs of c2rust-transpile snapshot tests
 c2rust-transpile/tests/snapshots/**/*.rs
+
+# Relooper debug info
+/dumps
+c2rust-transpile/dumps
```
kkysen: Should we add `relooper` in the name somewhere instead of just `dumps`?

tests/unit/gotos/src/test_goto_error.rs
```diff
@@ -0,0 +1,87 @@
+#[test]
+pub fn test_goto_error_a() {
+    use crate::goto_error::rust_goto_error_a;
```
kkysen, line 3: Can these just all be top-level imports?

kkysen: Same for all of the other ones.

c2rust-transpile/src/cfg/mod.rs
```diff
@@ -69,14 +70,26 @@ pub enum Label {
     Synthetic(u64),
 }
 
-impl Label {
-    pub fn pretty_print(&self) -> String {
+impl std::fmt::Display for Label {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
kkysen:
```suggestion
impl Display for Label {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
```

Could we import these?

c2rust-transpile/src/cfg/mod.rs
```diff
@@ -69,14 +70,26 @@ pub enum Label {
...
+            Self::FromC(_, Some(name)) => write!(f, "_{name}"),
+            Self::FromC(id, None) => write!(f, "c_{}", id.0),
+            Self::Synthetic(id) => write!(f, "s_{id}"),
         }
     }
+}
+
+impl std::fmt::Debug for Label {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(self, f)
```
kkysen:
```suggestion
impl Debug for Label {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        Display::fmt(self, f)
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2508,31 +2508,48 @@ impl<'c> Translation<'c> {
...
             self.tcfg.use_c_loop_info,
             self.tcfg.use_c_multiple_info,
             live_in,
         );
 
+        fn dump_structures(relooped: &[cfg::Structure<Stmt>], file_name: &str) {
+            use std::io::Write;
+            std::fs::create_dir_all("dumps").unwrap();
+            let mut file = std::fs::File::create(&file_name).unwrap();
+            write!(&mut file, "{:#?}", relooped).unwrap();
```
```suggestion
            fs::create_dir_all("dumps").unwrap();
            fs::write(&file_name, format!("{relooped:#?}")).unwrap();
```

`File` isn't buffered, so we don't want to use that directly.  It should be simpler to just write all in one go.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2508,31 +2508,48 @@ impl<'c> Translation<'c> {
...
+        }
+
         if self.tcfg.dump_structures {
-            eprintln!("Relooped structures:");
-            for s in &relooped {
-                eprintln!("  {:#?}", s);
+            let file_name = format!("dumps/{}_structures_initial.ron", name);
```
```suggestion
            let file_name = format!("dumps/{name}_structures_initial.ron");
```

I didn't know we use RON.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2508,31 +2508,48 @@ impl<'c> Translation<'c> {
...
+            dump_structures(&relooped, &file_name);
+        }
+
+        if self.tcfg.simplify_structures {
+            relooped = cfg::relooper::simplify_structure(relooped);
+
+            if self.tcfg.dump_structures {
+                let file_name = format!("dumps/{}_structures_simplified.ron", name);
```
```suggestion
                let file_name = format!("dumps/{name}_structures_simplified.ron");
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2508,31 +2508,48 @@ impl<'c> Translation<'c> {
...
         if self.tcfg.json_function_cfgs {
+            std::fs::create_dir_all("dumps").unwrap();
             graph
-                .dump_json_graph(&store, format!("{}_{}.json", "cfg", name))
+                .dump_json_graph(&store, format!("dumps/{name}_cfg.json"))
                 .expect("Failed to write CFG .json file");
         }
 
-        let (lifted_stmts, relooped) = cfg::relooper::reloop(
+        let (lifted_stmts, mut relooped) = cfg::relooper::reloop(
```
kkysen, line 2546:
```suggestion
        let (lifted_stmts, mut relooped) = reloop(
```

Can we just import and directly call stuff like this?  Was there a reason not to?

kkysen (commented):
For the snapshots, could you have one commit where they're tested with the previous relooper, and then a new commit where they're updated so I can see the diffs?  Or if this is already the case, tell me which commits to compare, because there's a lot of them so I'm not sure where to look.

scripts/test_translator.py
randomPoison: Ah yup, I forgot I wrote up an issue for that. I can have this PR close that one (though your notes in there about converting the tests to use `cargo test` are good, we might want to open a separate issue to track that).

.gitignore
randomPoison: It's not necessarily relooper-specific, if there are other parts of the transpile or refactor process where we want to dump large quantities of data for debugging we can drop it in there.

c2rust-transpile/src/translator/mod.rs
randomPoison: This one is only called in one place, so in this case the import doesn't cut down on verbosity meaningfully. Also I think the code was already written that way so I don't see a point in changing it.

tests/unit/gotos/src/test_goto_error.rs
randomPoison: For these it's helpful to import the function names within each test function because that makes it a bit easier to comment out tests for debugging, i.e. if I want to comment out some of the test functions in the original C file, I also have to comment out the corresponding test function and import in the Rust test file. If the import is in the corresponding function I can just comment out the function, whereas if the imports are at the top of the file I have to comment out the test function and the import separately. It's not a big deal either way, but there's also not much of a reason to move the imports.

.gitignore
randomPoison: I did update the comment to make it non-relooper-specific though.

c2rust-transpile/src/translator/mod.rs
randomPoison: What's the issue here with `File` not being buffered? I assume that's some kind of minor performance issue? The docs say to use a buffered write when doing many small writes, but we're just doing one big write so it's not clear to me that it matters here.

c2rust-transpile/src/translator/mod.rs
randomPoison: Ah, it's not really RON, just the `Debug` output for the relooper structured CFG (though that's very close to RON afaict). I just used that file extension to get automatic syntax highlighting (as opposed to having to manually enable syntax highlighting in my editor each time I open the file).

randomPoison:
> For the snapshots, could you have one commit where they're tested with the previous relooper, and then a new commit where they're updated so I can see the diffs? Or if this is already the case, tell me which commits to compare, because there's a lot of them so I'm not sure where to look.

@kkysen There's not a good way to do that without rewriting a lot of history. We didn't have snapshot tests for the CFG stuff before, so they're pretty much all new.

c2rust-transpile/src/translator/mod.rs
kkysen: Ahh. A comment would be useful because that could be pretty confusing.

kkysen:
> There's not a good way to do that without rewriting a lot of history. We didn't have snapshot tests for the CFG stuff before, so they're pretty much all new.

Hmm.  Can you run the snapshots on the previous transpiler in a commit somewhere so we can review the diff?  It's a lot harder to review without that IMO.

randomPoison:
> Can you run the snapshots on the previous transpiler in a commit somewhere so we can review the diff?

Yeah that's easy enough to do, at least. 218a2817c62364167202099ac22759ea3823126f now shows the diff between the old behavior and the new behavior in the snapshots

c2rust-transpile/src/translator/mod.rs
kkysen: I guess if it was already like that.

kkysen:
> Yeah that's easy enough to do, at least. [218a281](https://github.com/immunant/c2rust/commit/218a2817c62364167202099ac22759ea3823126f) now shows the diff between the old behavior and the new behavior in the snapshots

Thanks, that's very helpful.

scripts/test_translator.py
```diff
@@ -91,6 +91,7 @@ def translate(self, cc_db: str, ld_lib_path: str, extra_args: List[str] = []) ->
         args.append("--ignore-c-multiple-info")
         args.append("--ignore-c-loop-info")
         args.append("--ddump-structures")
+        args.append("--json-function-cfgs")
```
fw-immunant: Do we really want this unconditionally? It seems like a debugging aid that should only be produced when desired. When dumps were just to standard error this mattered less, but now we'll be polluting the filesystem. I guess the same goes for `--ddump-structures` if that's also writing to the filesystem now as well.

fw-immunant (approved):
Scoped just to the commit with test diffs from the pre-relooper output: the new output looks improved and I didn't notice any changes in the behavior of translated code.

scripts/test_translator.py
randomPoison: These are no longer in there, so I think you might have been looking at an old diff?

PR #1559, opened 2026-01-21
transpile: add `--c2rust-dir` to resolve relative c2rust dependencies

kkysen:
* Fixes https://github.com/GaloisInc/Tractor-Crisp/issues/38 (once crisp is updated to point to the fix).

Our internal `c2rust` dependencies like `c2rust-bitfields` and `c2rust-asm-casts` need to be patched when building from source/repo (as opposed to crates.io), as the crates.io versions can be out of sync with `HEAD` for `c2rust-transpile`, etc.  We manually patch this in `tests/integration`, CRISP, etc.  This replaces that with a `--c2rust-dir` arg so that `c2rust-transpile` can emit the correct `Cargo.toml` in the first place.

### DIFF
--- a/c2rust-transpile/tests/snapshots.rs
+++ b/c2rust-transpile/tests/snapshots.rs
@@ -1,6 +1,7 @@
 use std::env::current_dir;
 use std::fs;
 use std::path::Path;
+use std::path::PathBuf;
 use std::process::Command;
 
 use c2rust_transpile::{ReplaceMode, TranspilerConfig};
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/build_files/Cargo.toml.hbs
```diff
@@ -33,7 +33,11 @@ name = "{{name}}"
 {{/each}}
 [dependencies]
 {{#each dependencies}}
+{{#if this.path}}
+{{this.name}} = { version = "{{this.version}}", path = "{{this.path}}" }
```
ahomescu, line 37: Does `version` do anything here?

kkysen: Yeah, I believe it's still required by cargo.  It might be a newer requirement, though.

PR #1566, opened 2026-01-24
Improve C decl map handling of `#ifdef`s and multiple declarations

fw-immunant:
When we integrate with Hayroll, we'll want to merge C decl maps from different runs of the transpiler in which different, overlapping subsets of code are enabled via the preprocessor. In prototyping this I ran into cases where we attribute all preprocessor-disabled code immediately preceding a function to that function's source range.

This PR adds logic to our decl map generation to avoid this misattribution.

While we're at it, fix a bug causing reordering of C decls in `c_decls.json` files: the decl map should always match the order in the C source.

This also improves handling of multiple declarations (e.g. `int a, b;`), which libmcs makes particular use of.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -450,6 +450,7 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
     file_path
 }
 
+// TODO: rewrite when we update to Rust 1.80
 // TODO: also remove/ignore /**/ comments?
 /// Identify the preprocessor directive that starts a line of C source code.
 /// For example, `define`, `ifdef`, `if`, etc., without leading `#` or trailing arguments.


c2rust-transpile/src/translator/mod.rs
```diff
@@ -448,12 +448,74 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
...
 
+// TODO: also remove/ignore /**/ comments?
+/// Identify the preprocessor directive that starts a line of C source code.
+fn preprocessor_directive(mut line: &[u8]) -> Option<&[u8]> {
+    assert!(
+        line.iter()
+            .position(|c| *c == b'\n')
+            .unwrap_or(line.len() - 1)
+            == line.len() - 1
+    );
```
```suggestion
    assert!(!line.contains(&b'\n') || line.ends_with(b"\n"));
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -448,12 +448,74 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
...
+    while !line.is_empty() && line[0].is_ascii_whitespace() {
+        line = &line[1..];
+    }
```
kkysen, line 461:
```suggestion
    // TODO Rust 1.80
    // let line = line.trim_ascii_start();
    while !line.is_empty() && line[0].is_ascii_whitespace() {
        line = &line[1..];
    }
```

fw-immunant: Really this whole function could be rewritten more cleanly with newer Rust. But I don't know if a TODO for semantically identical code is worth it.

kkysen: Can you just leave a todo for the newer Rust version then?

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -169,6 +169,18 @@ impl PartialOrd for SrcLocInclude<'_> {
     }
 }
 
+/// The range of source code locations for a C declaration.
+#[derive(Copy, Clone)]
+pub struct CDeclSrcRange {
+    /// The earliest position where this declaration or its documentation might start.
+    pub earliest_begin: SrcLoc,
+    /// A position by which this declaration itself is known to have begun.
+    /// Attributes or return type may possibly precede this position.
```
kkysen, line 184: So it might be the function name, excluding the return type?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -448,12 +448,70 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
...
+    line_offset + (loc.column as usize).saturating_sub(1)
+}
+
+#[test]
+fn test_src_loc_to_byte_offset() {
+    let loc = |l, c| SrcLoc {
+        fileid: 0,
+        line: l,
+        column: c,
+    };
```
kkysen, line 500: nit: I think just `l` is a bad variable name, even though it's very short-lived here.

```suggestion
    let loc = |line, column| SrcLoc {
        fileid: 0,
        line,
        column,
    };
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -448,6 +448,34 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
     file_path
 }
 
+// TODO: also remove/ignore /**/ comments?
+/// Identify the preprocessor directive that starts a line of C source code.
+fn preprocessor_directive(mut line: &[u8]) -> Option<&[u8]> {
+    assert!(!line.contains(&b'\n') || line.ends_with(b"\n"));
```
kkysen, line 458: I wish Rust had an implies/`=>` operator.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -448,12 +448,70 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
...
 
+// TODO: rewrite when we update to Rust 1.80
+// TODO: also remove/ignore /**/ comments?
+/// Identify the preprocessor directive that starts a line of C source code.
+fn preprocessor_directive(mut line: &[u8]) -> Option<&[u8]> {
+    assert!(!line.contains(&b'\n') || line.ends_with(b"\n"));
+    while !line.is_empty() && line[0].is_ascii_whitespace() {
+        line = &line[1..];
+    }
+    if line.starts_with(&[b'#']) {
```
```suggestion
    if line.starts_with(b"#") {
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -448,12 +448,70 @@ fn clean_path(mod_names: &RefCell<IndexMap<String, PathBuf>>, path: Option<&Path
     file_path
 }
 
+// TODO: rewrite when we update to Rust 1.80
+// TODO: also remove/ignore /**/ comments?
+/// Identify the preprocessor directive that starts a line of C source code.
```
kkysen, line 455:
```suggestion
/// Identify the preprocessor directive that starts a line of C source code.
/// For example, `define`, `ifdef`, `if`, etc.
/// The `#` and arguments after are not returned.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -481,23 +539,56 @@ pub fn emit_c_decl_map(
...
+                let line_start_offset = byte_offset_of(line_loc(n));
+                let line_end_offset = byte_offset_of(line_loc(n + 1));
+
+                &file_content[line_start_offset..line_end_offset]
+            };
+
+            // Start at a position where we know the decl has begun and look backwards for a
+            // preprocessor directive.
+            let mut preproc_line = mid.line - 1;
+            while preproc_line > begin.line {
```
kkysen, line 578: I think using the names like `range.strict_begin` and `range.earliest_begin` would be easier to understand here.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -481,23 +539,56 @@ pub fn emit_c_decl_map(
...
+        let mut begin_offset = byte_offset_of(begin);
+        let mid_offset = byte_offset_of(mid);
+        let mut end_offset = byte_offset_of(end);
+        assert!(begin_offset <= mid_offset);
+        assert!(mid_offset <= end_offset);
+
+        // Remove any preceding lines prior to a final #else/#elif/#endif, as long as
+        // we don't see a different preprocessor directive first. This avoids us
+        // attaching an entire preceding `#ifdef`'d-out declaration to this one.
+        if let Some(preproc_offset) = || -> Option<usize> {
```
kkysen, line 561: I think it'd be a little more clear if this was a named closure instead of an IIFE.

kkysen: Like `previous_closing_directive_line` and then return just `preproc_line` and then do the `byte_offset_of(line_loc(preproc_line + 1))` in the `if let Some`.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -336,7 +336,8 @@ impl TypedAstContext {
         // Sort decls by source location so we can reason about the possibly comment-containing gaps
         // between them.
         let mut decls_sorted = self.c_decls_top.clone();
-        decls_sorted.sort_by_key(|decl| self.c_decls[decl].begin_loc());
+        decls_sorted
+            .sort_by_key(|decl| (self.c_decls[decl].begin_loc(), self.c_decls[decl].end_loc()));
```
kkysen, line 344: > in the case of top-level multiple declarations like `int a, b;`, these declarations share the same begin loc. in this case we want to process a before b, so break ties with their end loc

Can you put this commit message as a comment here?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -596,10 +596,11 @@ pub fn emit_c_decl_map(
 
         assert!(begin_offset <= end_offset);
 
-        // Extend to include a single trailing semicolon if this decl is not a block
+        // Extend to include a single trailing semicolon or comma if this decl is not a block
         // (e.g., a variable declaration).
         if file_content.get(end_offset - 1) != Some(&b'}')
-            && file_content.get(end_offset) == Some(&b';')
+            && (file_content.get(end_offset) == Some(&b';')
+                || file_content.get(end_offset) == Some(&b','))
```
kkysen: Could you use `matches!`?  `matches!(file_content.get(end_offset), Some(&b';' | &b','))`.  I'm not sure what the exact syntax is though.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -402,7 +402,7 @@ impl TypedAstContext {
             {
                 let entry = CDeclSrcRange {
                     earliest_begin: earliest_begin_loc,
-                    strict_begin: begin_loc,
+                    strict_begin: begin_loc.max(earliest_begin_loc),
```
kkysen: > this is not necessarily the case for multiple declarations like `int a, b;`, where `earliest_begin_loc` takes on the end of the first declaration. it may be the case that we should instead enforce this invariant by moving `earliest_begin_loc` backwards instead of moving `begin_loc` forwards, depending on what best suits c2rust-postprocess

Could you also include this commit message as a comment, too?

kkysen (changes requested):
Generally LGTM.  I left some comments about clarity, and I think a few commit messages would be very helpful as inline comments, too.  Sorry it took me so long to get around to review this.

c2rust-transpile/src/c_ast/mod.rs
fw-immunant: Yes, but I don't think it's guaranteed to be the function name in all cases.

c2rust-transpile/src/translator/mod.rs
fw-immunant: Doing so would require lifting `line_loc` outside of the closure, while this keeps more helpers localized. I don't think this is a huge difference in terms of comprehensibility.

c2rust-transpile/src/translator/mod.rs
fw-immunant: I originally used longer names here but the shorter names really help keeping derived names like `..._offset` concise and for maintaining parallel structure. As we mutate `begin_offset` and `end_offset`, they lose their correspondence to the meanings of the fields of a `CDeclSrcRange`.

c2rust-transpile/src/translator/mod.rs
kkysen: https://en.wikipedia.org/wiki/Material_conditional

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -384,13 +395,24 @@ impl TypedAstContext {
...
+                // For multiple decls, e.g. `int a, b;`, `begin_loc` is shared, in which case it is
+                // earlier than `earliest_begin_loc` for decls after the first; to maintain their
+                // relative order we must either move `earliest_begin_loc` earlier or move
+                // `begin_loc` later.
+                // For now, we move `begin_loc` later, so that the range used by each variable from
+                // a multiple decl does not overlap the others. If other tooling would benefit more
+                // from maximal but overlapping ranges, we could go the other way.
+                let entry = CDeclSrcRange {
+                    earliest_begin: earliest_begin_loc,
+                    strict_begin: begin_loc.max(earliest_begin_loc),
```
kkysen:
```suggestion
                // from maximal but overlapping ranges, we could go the other way.
                let begin_loc = begin_loc.max(earliest_begin_loc);
                let entry = CDeclSrcRange {
                    earliest_begin: earliest_begin_loc,
                    strict_begin: begin_loc,
```

With the way the comment is phrased (``For now, we move `begin_loc` later``), this makes more sense.  Or you could adjust the comment.

c2rust-transpile/src/translator/mod.rs
kkysen: But `begin`/`begin.line` is different from `begin_offset`, right?  `begin` isn't changed at all.

c2rust-transpile/src/translator/mod.rs
kkysen: Can you do `let fileid = t.ast_context.file_map[t.main_file];` and use that `fileid` so it doesn't depend on `begin`?  `t.main_file` is where we get `file_content` from and `fileid` isn't checked.

kkysen: Actually, can you just do `fileid: 0, // unused`?  It's never used anywhere.

c2rust-transpile/src/translator/mod.rs
fw-immunant: The closest translation of material implication to Rust is `if x { y } else { true }`, but I think that's less idiomatic than `!x || y`.

c2rust-transpile/src/translator/mod.rs
fw-immunant: It's not that we can't move `line_loc` to a larger scope, it's that minimizing the scope of helpers makes it more obvious why they exist.

The value of `fileid` isn't important in this usage of the `SrcLoc`, but it seems cleaner to make a fully semantically valid `SrcLoc` than one only valid for certain uses and then have to explain why that's justified.

c2rust-transpile/src/translator/mod.rs
fw-immunant: `begin_offset` is mutated by the end of the function to no longer equal `byte_offset_of(begin)`, yes. It and `end_offset` are used to build the final slice that gets returned, while `mid_offset` is only used for mutual validation of the input parameters.

Is there a blocking concern here? What should this code look like instead? I feel like this PR has been back and forth through a lot of minor bikeshedding.

fw-immunant:
Rebased onto `master` so #1622 can be rebased too.

c2rust-transpile/src/translator/mod.rs
kkysen: IMO the IIFE is more confusing than a wider helper scope.  But this isn't a very big concern.

PR #1574, opened 2026-01-27
git: set `update = none` for all submodules so that cargo git dependencies don't clone them

kkysen:
* `c2rust`-side fix for the git dependency version of https://github.com/GaloisInc/Tractor-Crisp/issues/46.

`cargo` always recursively clones git dependencies, even though c2rust git dependencies don't need any of the big submodules here used just for testing.  Setting `update = none` is workaround (see <https://github.com/rust-lang/cargo/pull/10717>).  However, other users need to explicitly update submodules now.

With this change, it takes ~6 s to resolve a `c2rust-bitfields` git dependency, vs. many minutes before with all of the large submodules.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,9 @@
 # The `update = none`s are to avoid cargo cloning
 # all submodules for git dependencies on c2rust.
-# See <https://github.com/rust-lang/cargo/pull/10717>.
+# It's unclear if cargo should be doing this
+# (see <https://github.com/rust-lang/cargo/issues/4247>),
+# but it is right now, so `update = none` is the only workaround
+# (see <https://github.com/rust-lang/cargo/pull/10717>).
 # This means that other users of submodules need to update them manually:
 # `git submodule update --init --checkout`.
 


ahomescu:
Do we really need this? I feel like it could create other problems, only to optimize a rare download.

kkysen (commented):
> Do we really need this? I feel like it could create other problems, only to optimize a rare download.

I'm not sure if we *really* need this, but git dependencies aren't that rare.  A bunch of other crates have done this workaround, too.  So IMO it seems useful to me on its own, and didn't use to have this behavior before integrating `c2rust-testsuite` into this repo, so it's new breakage as well.

randomPoison:
Unfortunately this is fallout from merging the test suite into the main repo. We did that to improve the development workflow, but it does have the consequence that when using the repo as a git dependency cargo is going to try pull down all the submodules. The [cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?utm_source=chatgpt.com#git-submodules) specifically call out `update = none` as the way to prevent this. Considering that using git dependencies isn't all that uncommon, especially for in-development projects where the latest code may not be on crates.io yet, it seems reasonable that we setup the repo to not download submodules automatically.

One suggestion to make things a bit nice to work with: Could we update `test.py` to automatically run `git submodule update --init --checkout` if the `repo` dir is empty? That'd save us a bit of repetition in the CI script and would make the workflow for running integration tests in a fresh checkout less error-prone.

kkysen:
> One suggestion to make things a bit nice to work with: Could we update `test.py` to automatically run `git submodule update --init --checkout` if the `repo` dir is empty? That'd save us a bit of repetition in the CI script and would make the workflow for running integration tests in a fresh checkout less error-prone.

Yeah, that could probably work.

Also, I realized that `cargo install --git https://github.com/immunant/c2rust/`, which I feel is actually pretty common, now takes forever.

tests/integration/tests/__init__.py
```diff
@@ -175,7 +175,7 @@ def ensure_submodule_checkout(self):
         elif is_dir_empty(repo_dir):
             repo_dir = os.path.relpath(repo_dir, os.path.curdir)
             msg = f"submodule not checked out: {repo_dir}\n"
-            msg += f"(try running `git submodule update --depth 50 --init {repo_dir}`)"
+            msg += f"(try running `git submodule update --depth 50 --init --checkout {repo_dir}`)"
             die(msg)
```
kkysen, line 179: @randomPoison, do you mean here, instead of erroring with a message saying what to do, to just do it?  For CI, though, I think it makes sense to do it explicitly as a separate step, just for clarity and so any error messages are separate (occasionally fetching fails).

randomPoison: Having to explicitly clone as a separate step in CI is what I'd like to avoid. My biggest problem with this change is that it adds additional steps that we have to remember to do when running the test suite (now we have to use `--checkout` when pulling in submodules) and when adding new repos to the test suite (remember to set `update = none`, remember to add it to the list of of repos when we run `test.py`, remember to add it to the list of checkout steps in CI). Some of that is unavoidable, but we should be automating those steps as much as possible.

If your concern is being able to see error messages if submodule checkout fails, can't we have the script print those errors on failure? Is there some issue that makes that hard to do?

.github/workflows/internal-testsuite.yml
```diff
@@ -37,7 +37,20 @@ jobs:
...
       with:
         repository: immunant/c2rust
-        submodules: true
+        # `update = none` is set, which makes this not work, so we need to update them manually.
+        # See `.gitmodules` for more info.
+        submodules: false
+    
+    - name: Checkout submodules
+      run: |
+        git submodule update --init --checkout tests/integration/tests/curl/repo
```
ahomescu, line 46: Does this PR need to update anything at those checked out paths? I know git puts some placeholders there.

ahomescu (changes requested):
Could we add a comment somewhere that we're doing this because cargo doesn't let us control whether it updates the submodules?

tests/integration/tests/__init__.py
ahomescu: > Having to explicitly clone as a separate step in CI is what I'd like to avoid.

IMHO that also holds for local development.

.github/workflows/internal-testsuite.yml
kkysen: Wdym placeholders?

ahomescu: Never mind, those don't show up until you do `git submodule init` or `git submodule update --init`.

tests/integration/tests/__init__.py
kkysen: We were already explicitly cloning submodules in a separate step in CI previously, though.  This is maintaining the status quo.

We can change that, but it's also more separate from the `update = none` change here.

ahomescu:
For reference: I think cargo did the wrong thing here, and should have provided a way to download a dependency without submodule.

kkysen:
> Could we add a comment somewhere that we're doing this because cargo doesn't let us control whether it updates the submodules?

Yeah, sure.  You want me to add more here: https://github.com/immunant/c2rust/blob/a101fbe53ee2902531bfa6df756e79c792703ea7/.gitmodules?

> For reference: I think cargo did the wrong thing here, and should have provided a way to download a dependency without submodule.

I agree.  Harder to change that, though.  Especially since we're stuck on an older nightly.

ahomescu:
The cargo issue discussing submodules: https://github.com/rust-lang/cargo/issues/4247 (still open).

.gitmodules
```diff
@@ -1,74 +1,100 @@
+# The `update = none`s are to avoid cargo cloning
```
ahomescu, line 1: This comment looks good, maybe also mention https://github.com/rust-lang/cargo/issues/4247? We might be able to revert this PR when that issue is resolved.

tests/integration/tests/__init__.py
randomPoison: > We were already explicitly cloning submodules in a separate step in CI previously, though. This is maintaining the status quo.

Previously we didn't have to manually list all the submodules, we got them all with `submodules true`. I'd prefer to avoid having multiple lists of the repos needed for the tests. Since this change breaks our ability to use `submodules true`, I think it makes sense to also update `test.py` to pull down submodules as part of this PR.

kkysen: > Previously we didn't have to manually list all the submodules, we got them all with `submodules true`.

That was actually significantly over cloning the submodules, as there were a lot we aren't using.  We can also do `git submodule update --init --checkout` without a specific repo and it will do all of them, but I didn't do that here since that clones a lot of extra submodules.  Locally, you can do that, too, if that's easier.  The only difference now is you need the `--checkout`.

> I'd prefer to avoid having multiple lists of the repos needed for the tests.

I can also just store this in a CI variable/matrix, too, which would be simpler.

.gitmodules
kkysen: Fixed in 8d43e4e.

tests/integration/tests/__init__.py
randomPoison: > That was actually significantly over cloning the submodules, as there were a lot we aren't using

True, and I do prefer a solution that allows us to only clone the required repositories in order to keep CI fast.

> I can also just store this in a CI variable/matrix, too, which would be simpler.

That would help avoid repetition in the CI script, but doesn't help us during local development. Adding submodule checkout to the test runner script helps avoid the extra step when developing locally as well.

kkysen: You only have to do it locally once ever (per c2rust clone).  Also, if you have local changes in a submodule, this will overwrite them, which you probably don't want.

randomPoison: I figured we'd only run the checkout step if the `repo` dir is empty, so I don't think we'd have issues with overwriting local changes.

ahomescu:
LGTM (other than wishing cargo let us do it the other way). Handing it over to @randomPoison since there's some more feedback from her (I wish github had gerrit's +1 votes).

tests/integration/tests/__init__.py
kkysen: We discussed this offline and agreed that doing this should be fine; nothing should be in danger of being overwritten.  However, we can improve that later in a follow-up if it's enough of a priority.

PR #1578, opened 2026-01-29
Restore cross checks, part 2: fix the Rust runtime and unit tests

ahomescu:


cross-checks/rust-checks/runtime/src/hash/djb2.rs
```diff
@@ -37,14 +36,15 @@ impl Hasher for Djb2Hasher {
...
         self.0 = u32_chunks.fold(self.0, |h, cb| {
-            let cvec = u32x4::new(cb[0] as u32, cb[1] as u32, cb[2] as u32, cb[3] as u32);
+            let cvec = u32x4::from_array([cb[0] as u32, cb[1] as u32, cb[2] as u32, cb[3] as u32]);
             // The djb2 factors: powers of 33 from 33^3 to 33^0
-            const DJB2_FACTORS: u32x4 = u32x4::new(35937, 1089, 33, 1);
+            const DJB2_FACTORS: u32x4 = u32x4::from_array([35937, 1089, 33, 1]);
             let cmul = cvec * DJB2_FACTORS;
-            let ch1 = Ssse3U32x4::hadd(cmul, cmul);
-            let ch2 = Ssse3U32x4::hadd(ch1, ch1);
+            let ch1 = unsafe { core::arch::x86_64::_mm_hadd_epi32(cmul.into(), cmul.into()) };
```
kkysen, line 43: If you use `#[cfg(target_feature = "ssse3")]` on the function, these should be safe.  This was added in a newish version of Rust, though.

cross-checks/rust-checks/runtime/src/hash/djb2.rs
```diff
@@ -37,14 +36,15 @@ impl Hasher for Djb2Hasher {
...
+            let ch2 = unsafe { core::arch::x86_64::_mm_hadd_epi32(ch1, ch1) };
+            let cval = unsafe { core::arch::x86_64::_mm_extract_epi32(ch2, 0) };
```
fw-immunant, line 45: Do we need to worry about aarch64 portability here, or is this gated to x86[_64] only at a higher level?

cross-checks/rust-checks/runtime/src/hash/mod.rs
```diff
@@ -214,17 +214,19 @@ fn try_pointer<'a, T: ?Sized>(p: *const T) -> Option<&'a T> {
...
-              2: incb $1
-              3:"
-             : "=r" (_pv), "+r" (invalid)
-             : "r" (p)
-             : "cc"
-             : "volatile");
+        core::arch::asm!(
+            "jmp 1f",
+            ".word 2f - 1f",
+            ".ascii \"C2RUST_INVPTR\\0\"",
```
fw-immunant: `.asciz` might be easier to read than two levels of escaping to produce NUL termination.

fw-immunant (approved):
LGTM. `asm!` changes seem semantics-preserving--`volatile` and clobbering flags is the default in rust unless opted out.

cross-checks/rust-checks/runtime/src/hash/djb2.rs
kkysen: Also, if we do this, it should statically ensure https://github.com/immunant/c2rust/pull/1578/changes#r2742685242 is handled.

cross-checks/rust-checks/runtime/src/xcheck.rs
```diff
@@ -47,7 +47,6 @@ unsafe fn rb_xcheck(tag: u8, val: u64) {
 // This is the only approach that requires that libclevrbuf.so is linked in
 #[cfg(not(any(feature = "xcheck-with-dlsym", feature = "xcheck-with-weak")))]
 extern "C" {
-    #[no_mangle]
```
kkysen, line 50: What about the `#[no_mangle]`s all of the other `fn rb_check`s.  I saw a bunch of them in different crates in the part 1 PR.

ahomescu: I was testing only this crate, with `cargo test -p c2rust-xcheck-runtime`. The others probably need to be fixed too, but I haven't gotten around to it.

cross-checks/rust-checks/runtime/src/hash/djb2.rs
ahomescu: This specific function is gated on the `djb2-ssse3` feature for SSSE3 which I assumed people wouldn't even try to enable on aarch64 (and we didn't even have that back then). I could add extra build time checks to error out on that corner case, but is it worth it?

cross-checks/rust-checks/runtime/src/hash/mod.rs
ahomescu: Good point, yeah that's better. We should also update the C assembly at some point, I'll leave that for later.

cross-checks/rust-checks/runtime/src/hash/djb2.rs
fw-immunant: I don't think it's critical, was just curious what scaffolding we had in place here.

cross-checks/rust-checks/runtime/src/hash/djb2.rs
ahomescu: My initial concern was that `target_feature` would be unconditional, but it looks like you can control it with `RUSTFLAGS` so you can build both kinds of binaries on the same machine.

Re aarch64: isn't that redundant now? I don't think `target_feature="ssse3"` is ever enabled there.

cross-checks/rust-checks/runtime/src/hash/djb2.rs
ahomescu: I switched to a `target_feature` check per @kkysen 's suggestion. Doesn't that make the aarch64 check redundant?

cross-checks/rust-checks/runtime/src/hash/djb2.rs
kkysen: > Re aarch64: isn't that redundant now? I don't think `target_feature="ssse3"` is ever enabled there.

Yup, shouldn't be possible.

cross-checks/rust-checks/runtime/src/xcheck.rs
kkysen: Oh sure, I wasn't sure how many parts there will be.

PR #1579, opened 2026-01-30
Implement `related_decls` tool

### LINKED ISSUE #1563: Compute related definitions
Given a decl:
- find all other decls mentioned by its definition or signature
- find all of its uses
- get its signature

See §Rust related-defs analysis [here](https://docs.google.com/document/d/1JLCracAx9SwOYbHA_3KuxXrU50Hla7RF3qKOoX-EdLI/edit?tab=t.0#heading=h.guc98d3r0bh4).

fw-immunant:
* Fixes #1563.

This builds and works, with the following caveats:
- ~Uses of an item are returned as source locations; these need to be resolved to their immediately containing item.~ (fixed)
- ~Output should be formatted as JSON.~ (done)
- Function signatures are not printed cleanly; we dump R-A's internal representation instead of generating Rust source for types.
  - Update: it isn't completely clear what the consumer of signatures is going to want (e.g., a function signature can not necessarily stand alone if it references generic parameters from an outer `impl`, etc.), but for now we'll emit both this WIP format and the signature exactly as written at the definition site, under the `"written_signature"` JSON key.
- ~We might want to select only some outputs (uses/used items/fn signature)~ (consumers can just filter out irrelevant output entries for now), and ~we might want to accept multiple paths at once as input~ (done).
- ~No automated tests as of yet.~


tools/related_decls/src/main.rs
```diff
@@ -1,335 +1,320 @@
...
+        Definition::Module(x) => Some(ModuleDef::Module(x)),
+        Definition::Function(x) => Some(ModuleDef::Function(x)),
+        Definition::Adt(x) => Some(ModuleDef::Adt(x)),
+        Definition::Variant(x) => Some(ModuleDef::Variant(x)),
+        Definition::Const(x) => Some(ModuleDef::Const(x)),
+        Definition::Static(x) => Some(ModuleDef::Static(x)),
+        Definition::Trait(x) => Some(ModuleDef::Trait(x)),
+        Definition::TypeAlias(x) => Some(ModuleDef::TypeAlias(x)),
+        Definition::BuiltinType(x) => Some(ModuleDef::BuiltinType(x)),
+        _ => None,
```
spernsteiner, line 222: Unless there's a huge number of cases to ignore, IMO it's better to match each one explicitly instead of using a wildcard case.  If an upgrade to rust-analyzer adds new variants, we want to get a compile error here so we don't forget to update this function.

tools/related_decls/src/main.rs
```diff
@@ -1,335 +1,320 @@
...
-        });
+/// Generate a string representation of an item's full path.
+fn absolute_item_path(db: &RootDatabase, item: ModuleDef, edition: Edition) -> String {
+    let mut out = String::new();
+    if let Some(name) = item.krate(db).display_name(db) {
+        let _ = write!(&mut out, "::{}::", name.as_str());
+    }
+    if let Some(path) = item.canonical_module_path(db) {
+        append_module_path_to_string(db, path, edition, &mut out);
+    }
```
spernsteiner, line 253: Maybe worth noting (or `debug_assert!`ing) that `append_module_path_to_string` always leaves `::` at the end of `out`, similar to the other `write!` invocations.  This makes it clear that the `pop()` below is correct regardless of which branches were taken previously.

tools/related_decls/src/main.rs
```diff
@@ -347,55 +332,117 @@ fn main() {
     )
     .unwrap();
 
+    log::info!("loaded crate");
+
     // Assume the first file in `vfs` is the crate root.
     let (first_file_id, _) = vfs.iter().next().unwrap();
```
spernsteiner: We should move setup logic like this into `rust_util`.  Currently it's duplicated between this tool and `split_ffi`.  We had to change this logic in `split_ffi` to handle crates that have both a lib and a bin target (https://github.com/GaloisInc/Tractor-Crisp/pull/36/changes/ccd9e15942a848024973e0a624ede0185acddaa2) and we'll probably need a similar change here—better to have it in one place so fixes like that can be shared.

tools/related_decls/src/main.rs
```diff
@@ -1,335 +1,320 @@
...
+            })
+        })
+        .collect();
+    references
 }
-*/
 
-// Helpers for dealing with nested meta items in attrs, like `#[unsafe(no_mangle)]`
+use std::fmt::Write;
+use std::ops::Index;
```
spernsteiner: Move these to the top with the other imports

tools/related_decls/src/main.rs
```diff
@@ -347,55 +332,117 @@ fn main() {
...
+                log::trace!(
+                    "traversal saw item {:?}",
+                    module_def.name(db).as_ref().map(Name::as_str)
+                );
+
+                // Mark this item as found if its path matches one we're looking for
+                let abs_path = absolute_item_path(db, module_def, file.edition(db));
+                let canonical_path = module_def.canonical_path(db, file.edition(db));
+                unfound_paths.retain(|path| {
+                    if canonical_path.as_ref() == Some(path) || abs_path == *path {
```
spernsteiner: Sounds like this is looking for an entry in `unfound_paths` that exactly matches either `abs_path` or `canonical_path`.  Could this use two calls to `unfound_paths.remove(...)` instead (and do the `insert` if one succeeds)?  Calling `retain()` is O(N).

tools/related_decls/src/main.rs
```diff
@@ -1,335 +1,320 @@
...
-        self.ident.to_tokens(tokens);
-        self.eq.to_tokens(tokens);
-        self.name.to_tokens(tokens);
+/// Returns the paths of all items used by the given item.
+fn items_used_by(sema: &Semantics<RootDatabase>, module_def: ModuleDef) -> HashSet<ModuleDef> {
+    let dwb = module_def.as_def_with_body().unwrap();
+    let func = match dwb {
+        DefWithBody::Function(f) => f,
+        _ => panic!("not function!"),
+    };
```
randomPoison: Do we want the tool to recover gracefully when pointed at a definition that's not a function? `main` already returns a `Result`, seems like this function could do the same instead of panicking.

tools/related_decls/src/main.rs
```diff
@@ -347,55 +332,117 @@ fn main() {
...
+                            pp_function_signature(&db, &*sig).into(),
+                        );
+                        path_info.insert(
+                            "written_signature".to_owned(),
+                            function_signature_as_written(&sema, func_id.into()).into(),
+                        );
+                    }
+                }
+            }
+        }
```
randomPoison: This loop + match seems odd to me. You're creating an array of query types and iterating over it, but there's no shared code between the three listed cases. It seems like you could get the same behavior without the loop or match, i.e. just trying the three query types in order. I noticed there's a commented out `query` field in `Args`, is the intent to make the type of query a user-specified option? Maybe slap a TODO on there if that's the intent.

fw-immunant: It wasn't initially clear which would be more efficient, but further investigation makes it seem like querying everything at once is better to avoid having to load the r-a analysis cache multiple times. So I'll remove this type.

tools/related_decls/src/main.rs
fw-immunant: Yeah, this needed more implementation which is just about finished now.

tools/related_decls/src/main.rs
fw-immunant: Do we want to do this here or in the CRISP repo?

tools/related_decls/src/main.rs
fw-immunant: Any new variants that we care about would come from Rust adding new top-level items, which we'd also have to handle elsewhere. But it's fairly likely that r-a might add new variants that we don't care about as their go-to-definition support grows more specialized functionality. So I think it makes sense to leave this as a wildcard.

tools/related_decls/src/main.rs
spernsteiner: Probably better to do the change after moving everything over to the CRISP repo

Issue #1564, opened 2026-01-23
Transpile standard C APIs into Rust equivalents


thedataking:
Examples
- C `assert` (from `assert.h`) to Rust `assert!`
- printf and related

Can we handle function-like macros during transpilation?

Refactoring tool already has support for printf cleanup.

Might be best to handle `asserts` in c2rust-postprocessor.

ahomescu:
Pointing out that c2rust-refactor has `convert_printfs` at https://github.com/immunant/c2rust/blob/master/c2rust-refactor/src/transform/format.rs#L199.

randomPoison:
The easiest places for us to start are places where we can port to the idiomatic Rust API using only local reasoning, i.e. where we don't have to also refactor any of the surrounding code. For example, math functions like `sin` can easily be replaced with the corresponding safe Rust function call. These aren't necessarily the highest impact liftings, but they're relatively trivial and allow us to replace unsafe function calls with safe Rust equivalents, and can be a big win for codebases that heavily use any of these APIs.

The ones that I'm aware of and seem like good starting points:

- Math functions: `sin`, `cos`, `tan`, `sqrt`, `pow`, `log`, `exp`, `fabs`
- Program control: `exit`, `abort`
- `assert`: This one is a bit trickier because it's a macro and conditional compilation comes into play, but should still be doable.

Some other ones to consider:

- Reading error information: `errno` -> `io::Error::last_os_error().raw_os_error()`, `perror`, `strerror`
  - I'm a bit less sure this one is worth it because imo it doesn't move the needle much towards the actual Rust idiom of `Result`, and a lot of the uses of `errno` involve writing to it which we can't trivially make safe. But we could at least make reading error information safe, which might be a worthwhile intermediate step towards later liftings.
- Character classification: `isdigit`, `isalpha`, `isspace`, `tolower`, `toupper`.
  - These have somewhat obvious Rust equivalents, but also have locale-dependent behavior that doesn't easily translate to Rust and some weird edge cases around `EOF`. We might be able to conservatively translate these to their ascii-specific Rust versions in cases where we can verify that locale doesn't come into play.
- Raw memory primitives: `memcpy` -> `ptr::copy_nonoverlapping`, `memmove` -> `ptr::copy` (allows overlap), `memset` -> `ptr::write_bytes`.
[...]

PR #1586, opened 2026-02-04
refactor: Add pass for converting libc math functions

randomPoison:
Converts `sin(x)` to `x.sin()` for all of the libc math functions that have direct Rust equivalents. Part of https://github.com/immunant/c2rust/issues/1564.

There's a tricky edge case for this transform: The conversion from a function call to a method call has implications for type inference. Going from `sin(1.0)` to `1.0.sin()` doesn't work because the latter doesn't compile without adding a type suffix to the literal, e.g. `1.0f32.sin()`. I considered adding logic to add suffixes to unsuffixed literals, but decided not to because that doesn't solve the more general cases such as `let x = 1.0; sin(x);` or `sin(x.into())`. From what I can tell, c2rust always generates suffixed literals for floats, which means in practice this won't be an issue when running this transform on c2rust's immediate output. However, this may be a problem when combined with other refactoring passes, e.g. one that removes unnecessary literal suffixes or type annotations.

My feeling is that this would be tricky to handle in the refactorer because it requires reasoning about type inference and potentially adding type annotations far from the call site, but it would be trivial for an LLM to add the necessary annotations based on the compiler errors. So I lean towards keeping the refactor pass simple with the intent of it potentially requiring a postprocessing step to work in all cases.

### DIFF
--- a/c2rust-refactor/tests/convert_math_funcs/old.rs
+++ b/c2rust-refactor/tests/convert_math_funcs/old.rs
@@ -69,7 +69,7 @@ fn main() {
         let z = sin(0.0f64) + sin(1.0f64);
 
         // Nested call
-        let w = sin(sin(0.5f64));
+        let w = sin(cos(0.5f64));
 
         // Test sinf (f32 variant)
         let angle_f32: f32 = 1.5;
--- a/c2rust-refactor/tests/convert_math_funcs/new.rs
+++ b/c2rust-refactor/tests/convert_math_funcs/new.rs
@@ -69,7 +69,7 @@ fn main() {
         let z = 0.0f64.sin() + 1.0f64.sin();
 
         // Nested call
-        let w = 0.5f64.sin().sin();
+        let w = 0.5f64.cos().sin();
 
         // Test sinf (f32 variant)
         let angle_f32: f32 = 1.5;


c2rust-refactor/src/transform/math.rs
```diff
@@ -0,0 +1,232 @@
...
+/// let result = sin(angle);
+/// let result_f32 = sinf(angle_f32);
+/// ```
+///
+/// gets converted to:
+///
+/// ```ignore
+/// let result = angle.sin();
+/// let result_f32 = angle_f32.sin();
+/// ```
```
kkysen, line 67: You could make these tests with hidden lines (e.g. defining `angle`, `sin`, `sinf`, etc.) instead of `ignore`.

c2rust-refactor/src/transform/math.rs
```diff
@@ -0,0 +1,232 @@
...
+                *e = mk()
+                    .span(e.span)
+                    .method_call_expr(receiver, method, method_args);
+            }
+        });
+    }
+}
+
+pub fn register_commands(reg: &mut Registry) {
+    use super::mk;
```
kkysen, line 173: Is this a different `fn mk`?

c2rust-refactor/src/transform/math.rs
```diff
@@ -0,0 +1,232 @@
...
+            } else if let Some(&method) = binary_defs.get(&def_id) {
+                if args.len() != 2 {
+                    return;
+                }
+                let receiver = args[0].clone();
+                let method_args = vec![args[1].clone()];
+                *e = mk()
+                    .span(e.span)
+                    .method_call_expr(receiver, method, method_args);
+            }
```
kkysen, line 167: You can match on slices/`args`, which might help.

c2rust-refactor/src/transform/math.rs
```diff
@@ -0,0 +1,232 @@
+use rustc_ast::ptr::P;
+use rustc_ast::*;
```
kkysen: Can we just import the items needed?

randomPoison:
> Can you call the method as a function? Like `f32::sin(1.0)`.

Yeah that's also an option, and would avoid the issue with type inference. However that's not the idiomatic usage of those methods, and would leave the code still feeling more like C than Rust. I think we should start by generating the more idiomatic code, and see how that works in practice. If we need to add more logic here to handle the type inference issue I'd rather it be in the form of adding type annotations, which seems like the more idiomatic solution to me.

c2rust-refactor/src/transform/math.rs
randomPoison: What would be the point of that? These aren't code snippets that test anything functional, they just show what the transformation looks like. I don't want them being run as part of `cargo test`.

c2rust-refactor/src/transform/math.rs
randomPoison: Yeah, there's two `fn mk` in play here 😭 There's `ast_builder::mk` which gets used in the transformation itself to build AST nodes, and then this is `transform::mk` which is used to setup/register a transform.

c2rust-refactor/src/transform/math.rs
randomPoison: Ah yeah I like that a bit better.

c2rust-refactor/src/transform/math.rs
kkysen: I think it's just generally best practice to have the code examples compilable to make sure you don't make mistakes in them, things get out of date, etc.  It's not that big of a deal.

c2rust-refactor/src/transform/math.rs
kkysen: Ahh, that's annoying.  Being more explicit with `transform::mk` is probably better than, although I don't know if other transforms already all do it this way.

c2rust-refactor/src/transform/math.rs
```diff
@@ -0,0 +1,229 @@
...
+                    binary_defs.insert(def_id, "copysign");
+                }
+                "fmin" | "fminf" | "fminl" => {
+                    binary_defs.insert(def_id, "min");
+                }
+                "fmax" | "fmaxf" | "fmaxl" => {
+                    binary_defs.insert(def_id, "max");
+                }
+                _ => {}
+            }
```
kkysen:
```suggestion
            let unary_defs = &mut unary_defs;
            let binary_defs = &mut binary_defs;
            let (defs, name) = match fi.ident.as_str() {
                "sin" | "sinf" | "sinl" => (unary_defs, "sin"),
                "cos" | "cosf" | "cosl" => (unary_defs, "cos"),
                "tan" | "tanf" | "tanl" => (unary_defs, "tan"),
                "asin" | "asinf" | "asinl" => (unary_defs, "asin"),
                "acos" | "acosf" | "acosl" => (unary_defs, "acos"),
                "atan" | "atanf" | "atanl" => (unary_defs, "atan"),
                "sinh" | "sinhf" | "sinhl" => (unary_defs, "sinh"),
                "cosh" | "coshf" | "coshl" => (unary_defs, "cosh"),
                "tanh" | "tanhf" | "tanhl" => (unary_defs, "tanh"),
                "asinh" | "asinhf" | "asinhl" => (unary_defs, "asinh"),
                "acosh" | "acoshf" | "acoshl" => (unary_defs, "acosh"),
                "atanh" | "atanhf" | "atanhl" => (unary_defs, "atanh"),
                "sqrt" | "sqrtf" | "sqrtl" => (unary_defs, "sqrt"),
                "cbrt" | "cbrtf" | "cbrtl" => (unary_defs, "cbrt"),
                "log" | "logf" | "logl" => (unary_defs, "ln"),
                "exp" | "expf" | "expl" => (unary_defs, "exp"),
                "fabs" | "fabsf" | "fabsl" => (unary_defs, "abs"),
                "abs" | "labs" | "llabs" => (unary_defs, "abs"),
                "floor" | "floorf" | "floorl" => (unary_defs, "floor"),
                "ceil" | "ceilf" | "ceill" => (unary_defs, "ceil"),
                "round" | "roundf" | "roundl" => (unary_defs, "round"),
                "trunc" | "truncf" | "truncl" => (unary_defs, "trunc"),
                "pow" | "powf" | "powl" => (binary_defs, "powf"),
                "atan2" | "atan2f" | "atan2l" => (binary_defs, "atan2"),
                "hypot" | "hypotf" | "hypotl" => (binary_defs, "hypot"),
                "copysign" | "copysignf" | "copysignl" => (binary_defs, "copysign"),
                "fmin" | "fminf" | "fminl" => (binary_defs, "min"),
                "fmax" | "fmaxf" | "fmaxl" => (binary_defs, "max"),
                _ => return,
            };
            defs.insert(def_id, name);
```

c2rust-refactor/src/transform/math.rs
randomPoison: Other transforms do `use super::mk`, so I'm just following that pattern here. I don't think it matters one way or the other, there's not actually much ambiguity since `transform::mk` is only used in the one place.

ahomescu:
> I considered adding logic to add suffixes to unsuffixed literals

We have the opposite of that in the `remove_literal_suffixes` transform. Speaking of, it has a pretty basic [implementation of unification-based type inference](https://github.com/immunant/c2rust/blob/master/c2rust-refactor/src/transform/literals.rs#L223), I wonder if it would be useful here.

c2rust-refactor/src/transform/math.rs
```diff
@@ -0,0 +1,176 @@
...
+                return;
+            };
+
+            let Some(def_id) = cx.try_resolve_expr(f) else {
+                return;
+            };
+
+            if let Some(&method) = unary_defs.get(&def_id) {
+                let [receiver] = args.as_slice() else {
+                    return;
```
ahomescu, line 155: It might be useful to log here (and in the binary case below) that we got here, so we can investigate.

c2rust-refactor/tests/convert_math_funcs/old.rs
```diff
@@ -0,0 +1,116 @@
...
+    unsafe {
+        let angle = 3.14159f64 / 2.0f64;
+        let result = sin(angle);
+        println!("sin({}) = {}", angle, result);
+
+        // Multiple calls in one expression
+        let z = sin(0.0f64) + sin(1.0f64);
+
+        // Nested call
+        let w = sin(sin(0.5f64));
```
ahomescu: A `sin(cos(0.5f64))` might also be useful, to verify that the nested calls get rewritten in the right order.

c2rust-refactor/src/transform/math.rs
randomPoison: Not sure what there'd be to investigate, if the number of args is wrong then it's not one of the libc math funcs and we shouldn't transform it.

ahomescu: Yeah that's a good point, I guess they would be safe to ignore.

PR #1587, opened 2026-02-04
refactor: Add pass for converting `exit` and `abort`

randomPoison:
Convert calls to libc `abort` and `exit` to `std::process::abort` and `std::process::exit`. Part of https://github.com/immunant/c2rust/issues/1564.


c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+            if let ItemKind::Fn(_) = item.kind {
+                if crate::util::contains_name(&item.attrs, sym::no_mangle) {
+                    local_no_mangle_names.insert(item.ident.name);
+                }
+            }
+        });
+
+        let mut exit_defs = HashMap::new();
+        visit_nodes(krate, |fi: &ForeignItem| {
+            if crate::util::contains_name(&fi.attrs, sym::no_mangle) {
```
ahomescu: Just a nit: you could simplify the indentation by returning here early if the attribute is missing.


```suggestion
            if !crate::util::contains_name(&fi.attrs, sym::no_mangle) {
                return;
            }
            ...
```

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+                            Vec::<P<Expr>>::new(),
+                        );
+                    }
+                    "exit" => {
+                        if args.len() != 1 {
+                            return;
+                        }
+                        let status = args[0].clone();
+                        *e = mk().span(e.span).call_expr(
+                            mk().path_expr(vec!["", "std", "process", "exit"]),
```
kkysen: Can you use `abs_path_expre`?

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+                if crate::util::contains_name(&item.attrs, sym::no_mangle) {
+                    local_no_mangle_names.insert(item.ident.name);
+                }
+            }
+        });
+
+        let mut exit_defs = HashMap::new();
+        visit_nodes(krate, |fi: &ForeignItem| {
+            if crate::util::contains_name(&fi.attrs, sym::no_mangle) {
+                if let ForeignItemKind::Fn(_) = fi.kind {
```
ahomescu: Do we support `let ... else` in our nightly? If so, you could do
```
let ForeignItemKind::Fn(_) = fi.kind else { return; }
```

We also have similar `match_or!` and `match_or_else!` macros in `c2rust-refactor/src/macros.rs`.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+                    let def_id = cx.node_def_id(fi.id);
+
+                    // Ignore functions that are defined locally, either directly or as
+                    // indirect `extern "C"` imports, since those have to be custom
+                    // functions. We only want to translate calls to the foreign libc
+                    // functions.
+                    if local_no_mangle_names.contains(&fi.ident.name) {
+                        return;
+                    }
+                    if def_id.is_local() {
```
ahomescu: Isn't this redundant if you're immediately calling `get_if_local`? The latter will return `None`.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+                        match cx.hir_map().get_if_local(def_id) {
+                            Some(Node::ForeignItem(_)) => {}
+                            _ => return,
+                        }
+                    }
+
+                    match &*fi.ident.as_str() {
+                        "abort" => {
+                            exit_defs.insert(def_id, "abort");
```
ahomescu: Nit: an `enum` would be slightly more idiomatic than strings.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+                        if args.len() != 1 {
+                            return;
+                        }
+                        let status = args[0].clone();
+                        *e = mk().span(e.span).call_expr(
+                            mk().path_expr(vec!["", "std", "process", "exit"]),
+                            vec![status],
+                        );
+                    }
+                    _ => {}
```
ahomescu: We should at least log this if not assert, but if you switch to an enum you wouldn't need to do either because the match would be exhaustive.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+
+            let Some(def_id) = cx.try_resolve_expr(f) else {
+                return;
+            };
+
+            if let Some(&func_name) = exit_defs.get(&def_id) {
+                match func_name {
+                    "abort" => {
+                        *e = mk().span(e.span).call_expr(
+                            mk().path_expr(vec!["", "std", "process", "abort"]),
```
ahomescu: Could we go further and call `panic!`? It's not exactly equivalent when panic unwinds, but we might want that anyway.

`panic!` will even take a message, and show a backtrace.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+use crate::ast_manip::{visit_nodes, MutVisitNodes};
+use crate::command::{CommandState, Registry};
+use crate::transform::Transform;
+use crate::RefactorCtxt;
+
+/// # `convert_exits` Command
+///
+/// Usage: `convert_exits`
+///
+/// Converts C-style abort and exit function calls to use the equivalent
```
ahomescu, line 15: Might make sense to split these up into separate transforms, especially if we're using `panic!` for aborts (see later comment).

c2rust-refactor/tests/snapshots/snapshots__refactor-convert_exits.rs.snap
```diff
@@ -0,0 +1,37 @@
...
+    unsafe {
+        ::std::process::exit(1);
+    }
+}
+
+fn test_exit_variable() {
+    let code = 42;
+    unsafe {
+        ::std::process::exit(code);
+    }
```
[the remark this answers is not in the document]
ahomescu: +1, I was going to comment this exact thing. Could we remove the unsafe?

Alternatively, I think we have a transform elsewhere for that, but I don't know if it still works. `cargo fix` might also handle it.

c2rust-refactor/tests/snapshots/snapshots__refactor-convert_exits.rs.snap
```diff
@@ -0,0 +1,37 @@
+#![allow(unused)]
+
+extern "C" {
+    #[no_mangle]
+    fn abort() -> !;
```
ahomescu, line 9: Would `cargo fix` remove these, once they're unused?

c2rust-refactor/src/transform/exits.rs
kkysen: I don't think it's new enough for that, but if you're not binding anything, you can just do `if !matches!(...)`.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,110 @@
...
+        // Collect names of locally-defined #[no_mangle] functions that might
+        // shadow libc functions.
+        let mut local_no_mangle_names = HashSet::new();
+        visit_nodes(krate, |item: &Item| {
+            if let ItemKind::Fn(_) = item.kind {
+                if crate::util::contains_name(&item.attrs, sym::no_mangle) {
+                    local_no_mangle_names.insert(item.ident.name);
+                }
+            }
+        });
```
kkysen: This isn't too complex, but we do the exact same thing in #1586, so could we pull it out into a separate function?

c2rust-refactor/src/transform/exits.rs
kkysen: This could be enabled with a CLI arg maybe?

c2rust-refactor/src/transform/exits.rs
kkysen: Actually, the same for most of the `visit_nodes` call, too.  Maybe just have a function like `visit_non_local_no_mangle_items`?

c2rust-refactor/src/transform/exits.rs
randomPoison: Well I do like early returns

c2rust-refactor/src/transform/exits.rs
randomPoison: I've been using `let else` in c2rust-refactor and it seems to work.

c2rust-refactor/src/transform/exits.rs
randomPoison: Hm, I like this idea, `panic!` seems like a more idiomatic error exit than `abort`. But maybe we should do that as a separate pass, maybe one that turns std `abort` into `panic!`? At least it feels to me like those should be separate transforms for a couple of reasons:

- The libc -> std translations is semantics-preserving and guaranteed to be valid. Going to `panic!` introduces unwinding that wasn't in the original (though I suspect that's fine, and is likely the direction later lifting would go in anyway).
- There might be some amount of judgement we want to apply in deciding which `abort`s should become `panic!`s. `abort` is still valid in Rust, and there are some cases where you might use that instead of a panic, so blindly transforming all aborts into panics might not be right in all cases (though off the top of my head I can't think of an example where we'd want to keep abort over panic).
- When we introduce `panic!` we may also want to make changes to surrounding code, e.g. if the code is printing out an error immediately before aborting we might want to merge that into the `panic!` invocation. It's unclear to me if that's something sensible to do in c2rust-refactor or if that needs to be an LLM-directed lifting.

c2rust-refactor/src/transform/exits.rs
randomPoison: Yeah that's reasonable. Let's land https://github.com/immunant/c2rust/pull/1586 first, then I'll rebase this on top and pull out the common logic.

c2rust-refactor/src/transform/exits.rs
randomPoison: Oh huh, that apparently exists in c2rust-transpile, but not in c2rust-refactor 🙃

I looked at what the other transforms are doing and it seems like it'd be fine to remove the leading `::` part of the path entirely, so the path is just `std::process::exit`. I think that's the more idiomatic way of expressing it, the leading `::` doesn't really get used in hand-written code.

c2rust-refactor/src/transform/exits.rs
randomPoison: Removed this by adding an enum, as suggested elsewhere.

c2rust-refactor/src/transform/exits.rs
randomPoison: Per my [other comment](https://github.com/immunant/c2rust/pull/1587#discussion_r2775332212), I think we should leave this transform as-is and explore a separate transform for converting aborts to panics.

c2rust-refactor/tests/snapshots/snapshots__refactor-convert_exits.rs.snap
randomPoison: It does not, we'll have to handle the removal ourselves. It looks like one of the existing transforms [handles this](https://github.com/immunant/c2rust/blob/da114dc646fa445c44db9a4f98196c57b6e4512d/c2rust-refactor/src/transform/linkage.rs#L98-L109), we should either pull out that functionality and reuse it here, or maybe we should pull that into its own refactor pass that can be run after any transform that can remove references to extern items.

c2rust-refactor/tests/snapshots/snapshots__refactor-convert_exits.rs.snap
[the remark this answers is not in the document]
randomPoison: `cargo fix` does not remove unused `unsafe` (much to my surprise), but we do have [another transform that does that](https://github.com/immunant/c2rust/blob/da114dc646fa445c44db9a4f98196c57b6e4512d/c2rust-refactor/src/transform/funcs.rs#L283). Maybe I can run that pass as part of the test as well, and we can confirm the two refactors work in combination.

c2rust-refactor/src/transform/exits.rs
ahomescu: It makes sense to separate them, I'm even wondering if the current version should be two separate commands.

c2rust-refactor/tests/snapshots/snapshots__refactor-convert_exits.rs.snap
ahomescu: Yeah let's keep the passes separate and simple.

c2rust-refactor/src/transform/exits.rs
randomPoison: Yeah that's redundant and fixed now that I factored the "visit foreign fns" logic out into a helper fn.

c2rust-refactor/src/ast_manip/visit_node.rs
```diff
@@ -161,3 +166,49 @@ where
 {
     N::visit_nodes(target, callback)
 }
+
+/// Visit non-local foreign functions.
+///
+/// "Non-local" here means that the function isn't defined elsewhere in the
+/// crate. This is primarily intended for transforming calls to extern
+/// functions, e.g. the libc functions. Functions must be defined in an `extern
+/// "C"` block and be marked `no_mangle` in order to ensure that we only visit
```
kkysen:
```suggestion
/// "C"` block and be marked `#[no_mangle]` in order to ensure that we only visit
```

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,78 @@
+use rustc_ast::ptr::P;
+use rustc_ast::*;
```
kkysen: Can we expand this glob import, too?

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,78 @@
...
+            let Some(def_id) = cx.try_resolve_expr(f) else {
+                return;
+            };
+
+            if let Some(func) = exit_defs.get(&def_id) {
+                match func {
+                    ExitFn::Abort => {
+                        *e = mk().span(e.span).call_expr(
+                            mk().path_expr(vec!["std", "process", "abort"]),
+                            Vec::<P<Expr>>::new(),
```
kkysen, line 57:
```suggestion
                            vec![],
```

Does this not work?  That's annoying.

c2rust-refactor/src/transform/exits.rs
```diff
@@ -0,0 +1,78 @@
...
+                exit_defs.insert(def_id, ExitFn::Abort);
+            }
+            "exit" => {
+                exit_defs.insert(def_id, ExitFn::Exit);
+            }
+            _ => {}
+        });
+
+        MutVisitNodes::visit(krate, |e: &mut P<Expr>| {
+            let ExprKind::Call(ref f, ref args) = e.kind else {
```
kkysen:
```suggestion
            let ExprKind::Call(f, args) = &e.kind else {
```

c2rust-refactor/src/transform/exits.rs
randomPoison: Yeah we need a type annotation because the vec is empty. I'd either have to annotate the vec or `call_expr`.

randomPoison:
Ah yeah, if the snapshot PR already landed I can rebase and update the tests.

c2rust-refactor/tests/snapshots.rs
```diff
@@ -65,13 +65,21 @@ fn test_convert_math_funcs() {
...
 
-/// TODO Broken
-/// `unsafe`s are not actually removed.
 #[test]
 fn test_convert_math_skip() {
     test_refactor_named("convert_math_funcs", "convert_math_skip.rs");
 }
 
+#[test]
+fn test_convert_exits() {
```
kkysen, line 74: Can you keep these alphabetically sorted?  e < m

randomPoison: Sure, let me throw up another PR for that.

PR #1588, opened 2026-02-05
Prepare for v0.22.0 release

kkysen:
This follows our [release steps](https://github.com/immunant/c2rust/wiki/Release-Process), those still applicable at least, to release v0.22.0.  `clippy` uncovered a few bugs, which was very useful.  I didn't bother to fix the `clippy` errors on `c2rust-refactor`, though, as there are thousands, and they aren't auto-fixable (even though the suggestions are mostly automated).

Once released, this should fix https://github.com/GaloisInc/Tractor-Crisp/issues/46.

### DIFF
--- a/c2rust-macros/Cargo.toml
+++ b/c2rust-macros/Cargo.toml
@@ -1,12 +1,15 @@
 [package]
 name = "c2rust-macros"
-version = "0.22.0"
-authors = ["Stephen Crane <sjc@immunant.com>", "The C2Rust Project Developers <c2rust@immunant.com>"]
-edition = "2021"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
 description = "Procedural macro support crate for C2Rust"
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -246,6 +246,8 @@ struct MacroExpansion {
     ty: CTypeId,
 }
 
+type ZeroInits = IndexMap<CDeclId, (WithStmts<Box<Expr>>, IndexSet<Import>)>;
```
randomPoison, line 249: Huh, did clippy suggest pulling out this type alias? Is it only used in the one place (`zero_inits`)? I would normally only pull out a type alias if the type is used in multiple places, seems weird to use one in this case.

kkysen: Yeah, it complained it was too long/complex.

PR #1595, opened 2026-02-13
Restore fix_unused_unsafe transform and fix AST printing bug

randomPoison:
Add a test case for the fix_unused_unsafe transform, and fix an issue with how we were converting AST nodes to strings that was causing the transform to panic.

The issue with converting AST nodes to strings came from a subtle issue with how the pretty printer in `pprust` is meant to be used. `pprust::to_string` constructs a `pprust::State`, gives you that `State` object to interact with, and then extracts the printed string from the `State` at the end. This is meant to be used in combination with some `print_foo` methods on `State` that write into the state's internal buffer. However, we were using the `foo_to_string` methods, which directly return the printed string without updating the `State`'s buffer (or at least resets the state after the string is generated). This caused `pprust::to_string` to return an empty string, which then caused a panic in the rewriting process. The `foo_to_string` methods are what we want to use, and we can bypass `pprust::to_string` and construct a `State` object directly in order to use these methods. I looked at all the places where we were using `pprust::to_string` and fixed the ones that were using the API wrong.

The test reflects the current functionality of the transform, which only removes the `unsafe` keyword but leaves the block in place. I would like the further extend the transform pass to fully remove the block where safe to do so, but I'd like to do that in a separate PR to keep this one simple. I've added https://github.com/immunant/c2rust/issues/1596 to track that pending work.

### DIFF
--- a/c2rust-refactor/src/rewrite/strategy/print.rs
+++ b/c2rust-refactor/src/rewrite/strategy/print.rs
@@ -666,9 +666,7 @@ fn rewrite_at_impl<T>(old_span: Span, new: &T, mut rcx: RewriteCtxtRef) -> bool
 where
     T: PrintParse + RecoverChildren + Splice + MaybeGetNodeId,
 {
-    let new_string = new.to_string();
-    eprintln!("rewrite_at_impl: string: {new_string:?}");
-    let printed = add_comments(new_string.clone(), new, &rcx);
+    let printed = add_comments(new.to_string(), new, &rcx);
     let reparsed = T::parse(rcx.session(), &printed);
     let reparsed = reparsed.ast_deref();
 


c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -102,9 +102,7 @@ impl PrintParse for Stmt {
         // nodes.
         match self.kind {
             StmtKind::Expr(ref expr) => pprust::expr_to_string(expr),
-            _ => pprust::to_string(|s| {
-                s.stmt_to_string(self);
-            }),
+            _ => pprust::State::new().stmt_to_string(self),
```
ahomescu, line 105: Reading the callee here, it seems to me that the difference is we're switching from calling `s.stmt_to_string()` to `s.print_stmt()` instead. Is that correct?

ahomescu:
So the tl;dr of this PR is that we're switching from the `foo_to_string` methods to `print_foo` which is what we wanted anyway?

ahomescu (approved):

randomPoison:
> So the tl;dr of this PR is that we're switching from the `foo_to_string` methods to `print_foo` which is what we wanted anyway?

No, we're still using `foo_to_string`, those are the functions we want. But we're no longer calling them from inside `pprust::to_string`, because that's meant to be used with the `print_foo` helpers. For some reason half of the `print_foo` methods are private, so the `foo_to_string` ones are what we have to use in most cases.

Issue #1453, opened 2025-11-10
Refactorer tests are not run in CI


fw-immunant:
`scripts/test_rust_refactor.py` runs these, but CI does not run it. It seems like we may want to drop this script entirely and replace it with something that runs during `cargo test`.

PR #1598, opened 2026-02-19
refactor: add initial snapshot test for `rename_unnamed`, `reorganize_definitions` replacing old tests

kkysen:
* Fixes part of #1453.

This switches the old `c2rust-refactor` testing machinery to snapshot tests, which are much simpler, and unlike a lot of the old tests, should still work.  For now, I've just added the test for `rename_unnamed` and `reorganize_definitions`, but others should be fairly easy to add.  I also left a couple TODOs that I can fix here or in a follow-up PR.

### DIFF
--- a/c2rust-refactor/src/lib.rs
+++ b/c2rust-refactor/src/lib.rs
@@ -389,6 +389,23 @@ fn rebuild() {
     ops::compile(&ws, &compile_opts).expect("Could not rebuild crate");
 }
 
+/// This can only be called once.
+///
+/// Things like [`env_logger::init`] can only be called once.
+///
+/// Other things like [`env::set_var`] and [`env::remove_var`]
+/// are unsound to call on non-Windows in any multithreaded contexts.
+///
+/// So when using multiple threads, such as when running `cargo test`,
+/// this should be wrapped in a [`Once`].
+/// This solves calling things like [`env_logger::init`],
+/// but does not make [`env::set_var`], etc. sound to use.
+/// In practice, this mostly works, but is unsound.
+/// This can be fixed by using `cargo test --test-threads 1`,
+/// but this runs all of the tests sequentially, which is very slow.
+/// A better alternative would be something like `cargo nextest`,
+/// which runs tests in parallel but in separate processes,
+/// which is exactly what we want.
 fn init() {
     env_logger::init();
 


c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions.snap
kkysen, line 46: @ahomescu, how is this supposed to be handled on macOS?  The snapshot is being different on macOS due to something with this.

c2rust-refactor/src/lib.rs
```diff
@@ -409,6 +406,14 @@ pub fn lib_main(opts: Options) -> interface::Result<()> {
...
     env::set_var("RUSTFLAGS", rustflags);
 
+    rustc_driver::install_ice_hook();
+}
+
+static INIT: Once = Once::new();
+
+#[cfg_attr(feature = "profile", flame)]
+pub fn lib_main(opts: Options) -> interface::Result<()> {
+    INIT.call_once(init);
```
kkysen, line 433: This is a somewhat practical but unsound workaround.  `cargo test` (without `--test-threads 1`, which is slow) runs tests in threads, and calling `env::set_var` or `env::remove_var` with multiple threads on non-Windows OSes is unsound: https://doc.rust-lang.org/std/env/fn.set_var.html#safety.  One promising option is to switch to [`cargo nextest`](https://nexte.st/) instead, which runs each test in its own process.  That would also help with `c2rust transpile`, which also has multithreaded complications inherited from `libclangTooling`.  `cargo nextest` also seems pretty nice, and runs tests faster.

c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions.snap
kkysen: If it's not supposed to be tested on macOS at all, I can also make it x86_64 only.

c2rust-refactor/src/lib.rs
ahomescu: You could write that in a comment and leave it here. In any case, we should be fine as long as we sequence the variable writes to run before any actual tests (like you're doing).

c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions.snap
ahomescu: If you could make it Linux-only, that would be easier.

ahomescu (approved):

c2rust-refactor/src/lib.rs
kkysen: It's still unsound, though.  Just mostly works in practice.  Could we switch to `nextest`?  It shouldn't be much work.

kkysen: > You could write that in a comment and leave it here.

👍

c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions.snap
kkysen: Okay, I can do that.

c2rust-refactor/src/lib.rs
kkysen: I added the comment in 0d19e97.

PR #1600, opened 2026-02-20
refactor: add all remaining simple snapshot tests

kkysen:
* Fixes part of #1453.

These are all of the remaining `c2rust-refactor` tests that conform to this most simple layout:
* a command with no arguments
* has `$command_name/old.rs`
* the test succeeds and doesn't crash the refactorer

Of these, all but 1 have minor modifications to the `new.rs`/snapshot output (there's also `reconstruct_while/new.rs`, which is small, so `git` didn't register it as a move).  That's `fix_unused_unsafe/new.rs`, which doesn't seem to remove any `unsafe`s anymore.  Perhaps that's related to #1595?  Either way, it's good enough for the snapshot and we can fix it in its own PR.

Next, I'll do the more complex tests.

### DIFF
--- a/c2rust-refactor/tests/snapshots.rs
+++ b/c2rust-refactor/tests/snapshots.rs
@@ -110,3 +110,8 @@ fn test_struct_assign_to_update() {
 fn test_struct_merge_updates() {
     test_refactor("struct_merge_updates");
 }
+
+#[test]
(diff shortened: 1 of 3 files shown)


ahomescu:
Can we also start enabling these in CI?

PR #1603, opened 2026-02-20
refactor: Remove `-Awarnings` from snapshot tests and fix `fix_unused_unsafe` test

randomPoison:
So turns out the reason `fix_unused_unsafe` wasn't working is because we were suppressing warnings, and the transform relies on the warning analysis to determine what unsafe is unused. I don't really like this, but it doesn't seem like removing `-Awarnings` breaks anything, and I don't think it's worth building our own unused-unsafe analysis to avoid this.

@kkysen what was the reason for having `-Awarnings` in there to begin with? Is removing that going to cause any problems I'm not seeing?

### DIFF
--- a/c2rust-refactor/src/transform/funcs.rs
+++ b/c2rust-refactor/src/transform/funcs.rs
@@ -280,7 +280,7 @@ impl Transform for ToMethod {
 /// Usage: `fix_unused_unsafe`
 ///
 /// Find unused `unsafe` blocks and turn them into ordinary blocks. This relies
-/// on the compiler's unused-unsafe warnings, and will not work with warnings
+/// on the compiler's `#[warn(unused_unsafe)]` warnings, and will not work with warnings
 /// suppressed.
 pub struct FixUnusedUnsafe;
 


kkysen (commented):
Ahh, that's what was happening.  I added `-Awarnings` because when `c2rust-refactor` runs `rustc`, a ton of warnings are printed and these were very noisy (`c2rust transpile` emits code with a ton of warnings).  `cargo test` shows all of these and doesn't capture the output for some reason like it normally does.  However, `cargo nextest run` does, so it's not that big of an issue if we're using that.  So I think removing `-Awarnings` is okay here.

Could you add to `fix_unused_unsafe`'s docs that it relies on the warnings?  That's unexpected, so it's useful to know and call that out.

randomPoison:
Ah yeah, if I run `cargo test` I see all of the warning output. I was using `cargo nextest` so I didn't see it. I agree that that's probably fine since we're adopting nextest. I've also added the note to fix_unused_unsafe.

c2rust-refactor/src/transform/funcs.rs
```diff
@@ -279,7 +279,9 @@ impl Transform for ToMethod {
 ///
 /// Usage: `fix_unused_unsafe`
 ///
-/// Find unused `unsafe` blocks and turn them into ordinary blocks.
+/// Find unused `unsafe` blocks and turn them into ordinary blocks. This relies
+/// on the compiler's unused-unsafe warnings, and will not work with warnings
```
kkysen:
```suggestion
/// on the compiler's `#[warn(unused_unsafe)]` warnings, and will not work with warnings
```

Easier to grep for.

kkysen (approved):

PR #1604, opened 2026-02-21
refactor: Fix ordering of snapshot tests

randomPoison:
I added a comment in there about what the expected ordering is. It would be nice if copilot would see that and chastise us if we put any tests in the wrong order, but idk if it'll actually do that. At the least it's a reminder to me for the next time I add refactor tests.


kkysen (approved):
Lol, git generated a terrible diff here.  It's so bad at moves.

PR #1589, opened 2026-02-05
refactor: Add `parse!` substitution and enable support for macros with it

ahomescu:
* Fixes #1580.

This adds support for rewrites that emit macros, e.g., `rewrite_expr 'printf($e)' 'parse!(println!($e))'`. Without the `parse!` substitution, c2rust-refactor would try to parse `println!` in bindings not as a macro but as a special matcher directive.

PR #1608, opened 2026-02-24
refactor: tests: add `command_args: &[&str]`

kkysen:
This can be used by #1589.

### DIFF
--- a/c2rust-refactor/tests/snapshots.rs
+++ b/c2rust-refactor/tests/snapshots.rs
@@ -202,11 +202,13 @@ fn test_refactor(
         .flatten()
         .join("-");
     let snapshot_name = sanitize_file_name(&snapshot_name);
-    let command_args = shlex::try_join(command_args.iter().copied()).unwrap();
-    let rustc_args = shlex::try_join(rustc_args).unwrap();
-    let debug_expr = format!(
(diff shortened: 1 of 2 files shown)


c2rust-refactor/tests/snapshots.rs
```diff
@@ -153,13 +171,24 @@ fn test_refactor(
...
+    let snapshot_parts_with_cmd;
+    let snapshot_name_parts = if Some(command) == path.strip_suffix(".rs") {
+        snapshot_parts_no_cmd = ["refactor", path];
+        &snapshot_parts_no_cmd[..]
     } else {
-        format!("refactor-{command}-{path}")
+        snapshot_parts_with_cmd = ["refactor", command, path];
+        &snapshot_parts_with_cmd[..]
     };
+    let snapshot_name = [snapshot_name_parts, command_args]
```
ahomescu, line 200: Isn't this used as the file name of the snapshots? There are going to be strange characters in there, like `!`.

kkysen: Hmm, yeah, that's true.  We'll need to clean that.

kkysen: How's this: ae0d048?

PR #1609, opened 2026-02-24
refactor: tests: compile and format old and new `.rs` snapshot tests

kkysen:
* Fixes part of #1453.

This also checks that the before (old) and after (new) `.rs` files for the refactor snapshot tests are well-formatted and compile, with options to disable each of them.

Some of these tests have compile errors, so we have the option to disable them.  For example, `rename_unnamed.rs` has src loc annotations that are `rustc` errors.  `test_reflect.rs` produces type annotations that aren't correct syntax, so that doesn't compile.  And for formatting, sometimes what we want to test, like `reorder_derives.rs`, is obviated if we format it first.

### DIFF
--- a/c2rust-refactor/tests/snapshots.rs
+++ b/c2rust-refactor/tests/snapshots.rs
@@ -226,7 +226,7 @@ fn test_remove_unused_labels() {
 
 #[test]
 fn test_rename_unnamed() {
-    refactor("rename_unnamed").expect_compile_error(true).test();
+    refactor("rename_unnamed").test();
 }
 
 #[test]
--- a/c2rust-refactor/tests/snapshots/rename_unnamed.rs
+++ b/c2rust-refactor/tests/snapshots/rename_unnamed.rs
@@ -1,14 +1,4 @@
-#![feature(extern_types)]
-#![register_tool(c2rust)]
-#![allow(non_upper_case_globals)]
-#![allow(non_camel_case_types)]
-#![allow(non_snake_case)]
-#![allow(dead_code)]
-#![allow(mutable_transmutes)]
-#![allow(unused_mut)]
-
 pub mod bar {
-    #[header_src = "/home/user/some/workspace/foobar/bar.h"]
     pub mod bar_h {
         #[derive(Copy, Clone)]
         #[repr(C)]
@@ -33,7 +23,6 @@ pub mod bar {
 }
 
 pub mod foo {
-    #[header_src = "/home/user/some/workspace/foobar/foo.h"]
     pub mod foo_h {
         #[derive(Copy, Clone)]
         #[repr(C)]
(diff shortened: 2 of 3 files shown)


c2rust-refactor/tests/snapshots.rs
```diff
@@ -3,14 +3,126 @@ use c2rust_refactor::lib_main;
...
 use c2rust_rust_tools::EDITION;
 use insta::assert_snapshot;
 use std::path::Path;
 
-fn test_refactor_named(command: &str, path: &str) {
+#[must_use]
+struct RefactorTest<'a> {
+    command: &'a str,
+    path: Option<&'a str>,
+    old_expect_format_error: bool,
```
ahomescu, line 16: Repeating from the other PR: do we want to check this, or just make sure `old.rs` always compiles and is formatted? For example, couldn't we fix `rename_unnamed.rs` and the others instead?

kkysen: I realized I can fix `rename_unnamed.rs` (see eb35127) because the `#[header_src = "..."]` and `#![register_tool(c2rust)]` that it's using aren't actually needed for `rename_unnamed`.  But they are needed for `reorganize_definitions`, right?

There are also others like `let_x_uninitialized.rs` that I'm not sure how to fix without breaking what it's testing for.

kkysen: Hmm, actually maybe `reorganize_definitions` is broken and that's why its new `.rs` is failing, since its old `.rs` compiles fine (I'm not sure why, since it's basically the same as the previous `rename_unnamed.rs` before I deleted the erroring stuff).

ahomescu: Yeah `rename_unnamed` should work fine, and doesn't need the c2rust attributes.

It makes sense to discover that `new.rs` from `reorganize_definitions` is broken, we never compiled it before.

kkysen: What should I do about `let_x_uninitialized.rs`?

kkysen: `let_x_uninitialized` is actually obsolete now, so I'll delete it in a follow-up PR.

c2rust-refactor/tests/snapshots/reorder_derives.rs
```diff
@@ -15,12 +15,16 @@ pub struct S0 {}
 #[repr(u8)]
 #[derive(Copy)]
 #[derive(Debug)]
-pub struct S1 {}
+pub enum E1 {
```
ahomescu, line 18: Why is this an `enum` now?

ahomescu (approved):

c2rust-refactor/tests/snapshots/reorder_derives.rs
kkysen: See c7d49cc.  `#[repr(uN)]` isn't allowed on `struct`s.

PR #1611, opened 2026-02-24
refactor: Remove unnecessary blocks when cleaning up unused unsafe

### LINKED ISSUE #1596: Remove unnecessary blocks when cleaning up unused unsafe
The fix_unused_unsafe transform fixes unused unsafe by removing the `unsafe` keyword from unsafe blocks, but it leaves the blocks in place. In many cases it would be more idiomatic to remove the block entirely, preserving the block's contents. So for code like `unsafe { foo(); }`, we currently transform that to `{ foo(); }`, but it would be better to generate just `foo();`.

There are some edge cases to keep in mind here:

- Blocks can have comments attached, and deleting the block would take the comment along with it. If the block has a comment we need to make sure we move the comment to the first statement in the block to make sure it's not lost. If the block is empty such that there's nowhere to move the comment to, we should probably leave the block there so as to preserve the comment (a human or an LLM may be needed to make the judgement of what to do with the comment).
- If the block is used as an expression, i.e. if it's returning a value, we probably want to keep the block in place and just remove the `unsafe` keyword (which is the current behavior).
- If the block has anything inside it that requires `Drop`, we should leave the block in place so as not to alter drop order (this is another place where human or LLM judgement is necessary to decide if the changed drop order matters). This is maybe not a big concern because c2rust-transpile won't generate any code that relies on `Drop`, but could be a problem if fix_unused_unsafe is used later in the lifting process after other changes have been made. If it's easy to detect this case we should support it, but if drop analysis would be difficult or complex we can probably disregard this case for now.
- Some blocks are labeled, and shouldn't be removed as the labels are needed for control flow. These blocks generally shouldn't be `unsafe`, but depending on how we implement this logic that might be something that needs to be kept in mind.

I also think we should start by making this extra logic a part of fix_unused_unsafe, rather than pulling this out into its own transform. We have a few other transforms generating blocks in different ways, but it's not yet clear that any of them need this extra "remove unnecessary blocks" step. For now let's keep this cleanup logic part of fix_unused_unsafe to keep the use case narrow until we have a more concrete idea of where else this functionality would be useful.

randomPoison:
* Fixes #1596.

The deleted `parse_extras` function had logic for mapping the file-local comment positions to their global positions, which I also needed to do here. I merged that logic into `collect_comments` since in both places where we're calling it we want to do the same local-to-global remapping.

### DIFF
--- a/c2rust-refactor/src/transform/funcs.rs
+++ b/c2rust-refactor/src/transform/funcs.rs
@@ -408,8 +408,9 @@ impl<'a, 'tcx> MutVisitor for FixUnusedUnsafeFolder<'a, 'tcx> {
                 // because this is a block statement, which means that the tail expr isn't
                 // being used as part of an expression and so can (and must be) turned into
                 // a statement.
-                if let Some(last) = stmts.last_mut() && let StmtKind::Expr(expr) = &last.kind {
-                    last.kind = StmtKind::Semi(expr.clone());
+                if let Some(last) = stmts.last_mut() && let StmtKind::Expr(expr) = &mut last.kind {
+                    let expr = std::mem::replace(expr, mk().tuple_expr(Vec::<P<Expr>>::new()));
+                    last.kind = StmtKind::Semi(expr);
                 }
 
                 return SmallVec::from_vec(stmts);
@@ -430,7 +431,7 @@ impl<'a, 'tcx> MutVisitor for FixUnusedUnsafeFolder<'a, 'tcx> {
             }
 
             // We only remove the block if it consists of a single tail expr.
-            let [stmt] = &block.stmts[..] else {
+            let [stmt] = &mut block.stmts[..] else {
                 break 'noop;
             };
 
@@ -444,11 +445,11 @@ impl<'a, 'tcx> MutVisitor for FixUnusedUnsafeFolder<'a, 'tcx> {
                 break 'noop;
             }
 
-            let StmtKind::Expr(inner) = &stmt.kind else {
+            let StmtKind::Expr(inner) = &mut stmt.kind else {
                 break 'noop;
             };
 
-            *expr = inner.clone();
+            *expr = std::mem::replace(inner, mk().tuple_expr(Vec::<P<Expr>>::new()));
         }
 
         mut_visit::noop_visit_expr(expr, self)


c2rust-refactor/src/transform/funcs.rs
```diff
@@ -279,32 +283,118 @@ impl Transform for ToMethod {
...
-                let unused = result
+        // HACK: Check if there have been compilation errors before running the
+        // transformation. When there are errors, rustc's unsafe checker can produce
+        // incorrect results, incorrectly marking blocks with actual unsafe
+        // operations as "unused". This would cause us to remove the `unsafe`
+        // keyword from blocks that still need it, so we bail out in that case
+        // instead of corrupting the code.
+        if cx.session().diagnostic().has_errors().is_some() {
+            warn!("Skipping fix_unused_unsafe transform due to compilation errors");
+            return;
```
ahomescu, line 325: This will print the warning and then continue refactoring. This matters when we chain transforms (after we fix `commit`). Should we completely stop instead?

c2rust-refactor/src/transform/funcs.rs
```diff
@@ -279,32 +283,118 @@ impl Transform for ToMethod {
...
+
+        impl<'a, 'tcx> MutVisitor for FixUnusedUnsafeFolder<'a, 'tcx> {
+            fn flat_map_stmt(&mut self, mut stmt: Stmt) -> SmallVec<[Stmt; 1]> {
+                'noop: {
+                    let (StmtKind::Expr(expr) | StmtKind::Semi(expr)) = &mut stmt.kind else {
+                        break 'noop;
+                    };
+
+                    let expr_id = expr.id;
+                    let ExprKind::Block(block, label) = &mut expr.kind else {
```
ahomescu: Small nit:
```suggestion
                    let ExprKind::Block(block, None) = &mut expr.kind else {
```
so you exit early if there is a label.

c2rust-refactor/src/transform/funcs.rs
```diff
@@ -279,32 +283,118 @@ impl Transform for ToMethod {
...
+                    };
+
+                    if !self.is_unused_unsafe_block(block) {
+                        break 'noop;
+                    }
+
+                    let block_id = block.id;
+                    let hir_id = self.cx.hir_map().node_to_hir_id(block.id);
+                    let parent = self.cx.hir_map().get_parent_item(hir_id);
+                    let param_env = self.cx.ty_ctxt().param_env(parent);
```
ahomescu: There are some interesting subtleties here: if all type parameters are `T: Copy` (or fall into a few other niche cases), the compiler returns `ty.needs_drop(...) == false` below, otherwise it will return `true` for most types. It might be useful to 1) add a comment here, and 2) write a test or two that at least cover the `Copy` vs non-`Copy` cases.

c2rust-refactor/tests/snapshots.rs
```diff
@@ -78,6 +87,11 @@ fn test_fix_unused_unsafe() {
     test_refactor("fix_unused_unsafe");
 }
 
+#[test]
+fn test_fix_unused_unsafe_compile_error() {
+    test_refactor_named("fix_unused_unsafe", "fix_unused_unsafe_compile_error.rs");
```
kkysen: Can you rebase this on #1609?  You'll need to mark which files are expected to not compile.

c2rust-refactor/src/transform/funcs.rs
randomPoison: How would we completely stop the transform? We're returning here which should stop the refactoring right?

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
```diff
@@ -7,10 +7,43 @@ fn main() {
...
     unsafe {
+        // Block with no comment and no destructors, block should be removed but
+        // contents kept.
         let x = 42;
     }
 
     unsafe {}
 
+    // Has a comment, shouldn't be removed.
+    unsafe {}
```
kkysen, line 21: This comment is attached to the block in the AST?

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
```diff
@@ -7,10 +7,43 @@ fn main() {
...
+
+    unsafe {
+        // Contains a destructor, keep the block to avoid changing drop order.
+        let _string = String::from("string");
+    }
+
+    // Block expression should stay.
     let _ = unsafe { 5 + 5 };
```
kkysen: Why do we want to keep the block in this case?

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
```diff
@@ -7,10 +7,43 @@ fn main() {
...
+
+    unsafe {
+        // Contains a comment but no body. In theory we'd like to preserve this,
+        // but it's hard to detect the comment since it doesn't have an AST node
+        // it's attached to. For now we allow this to be deleted.
+    }
```
kkysen, line 47: Deleting comments doesn't seem good.

c2rust-refactor/src/transform/funcs.rs
randomPoison: Oh do you mean like stop the entire refactorer run? The `return` stops this one transform but not the rest of the refactorer run. Yeah I think it might be smart to not run the refactorer at all if there are compiler errors, since we don't know if there are other places like this that break when there are errors.

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
randomPoison: My thinking was that if a block is being used as an expression, it likely has other statements in it and so we don't wanna just remove the block. We could be smarter about this case, but I think in general if the block is being used an expression we should probably just leave it as-is and allow a higher level refactoring pass determine if the block is idiomatic to keep or not.

kkysen: I do this pretty often when I use `unsafe`, though, and in those cases I would remove the block if the `unsafe` were removed.  Can we remove the block for a single expression but keep it if there are any statements?

c2rust-refactor/src/transform/funcs.rs
randomPoison: Are you specifically referring to type parameters in generic functions? Or are you saying there's a different behavior here for `Copy` vs non-`Copy` types even in a non-generic context?

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
randomPoison: Yeah I believe so, at least we seem to be able to detect it in the comment map.

c2rust-refactor/src/transform/funcs.rs
```diff
@@ -293,6 +293,7 @@ impl Transform for ToMethod {
 /// - The block is unlabeled.
 /// - There are no comments on the block.
 /// - There are no values in the block that require drop.
+/// - The block doesn't have a tail expression (for statement-position blocks).
```
kkysen: I think you mean the block has no statements, a.k.a., it is a single tail expression.

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
randomPoison: Ah yeah, and we also want to be able to sink unsafe blocks to more tightly wrap unsafe ops, which is likely going to result in a lot of small unsafe block exprs, so handling this makes sense. The transform will now remove the block if it's a block expr consisting only of a tail expr.

c2rust-refactor/src/transform/funcs.rs
randomPoison: I added a test case to cover and document copy vs non-copy type params in a generic context, let me know if you also want a comment documenting that somewhere else as well.

c2rust-refactor/src/transform/funcs.rs
randomPoison: I split out an issue for handling this (https://github.com/immunant/c2rust/issues/1635), I don't think it makes sense to do in this PR, but it's probably worth disallowing running the refactorer when there are errors if we find more cases where problems crop up.

c2rust-refactor/tests/snapshots/fix_unused_unsafe.rs
randomPoison: Agreed, but I there are some deeper issues in how the refactorer handles comments that make me hesitant to try to address it in this PR.

The easiest way to handle comments is to track them based on the AST nodes that they're attached to, and with a trailing comment in a block like this there's no AST node for it to attach to. When we have a trailing comment like this, any rewriting of the block causes us to lose the comment, e.g. just removing the `unsafe` keyword but otherwise leaving the block the same will also delete any trailing comment. There's also a bug in how we associate comments with AST nodes which causes us to associate a trailing comment with a node that comes AFTER the block, which causes other bad interactions with the logic in this PR.

There's also a broader problem with this approach of associating comments with AST nodes, in that comments that appear in weird positions don't always have a corresponding node. For example, `unsafe /* comment */ {}`, where the comment appears between the `unsafe` keyword and the block, can't easily be handled this way. Cases like those are generally not idiomatic anyway, so I don't think it's worth putting effort into handing this better, but it highlights that we can't really handle comments in all cases (at least not without deeper fixes in the refactorer).

I'm writing up a ticket about comment handling in general, but for this PR I'd prefer to keep the behavior as-is, and leave it documented in the test that we have this unfortunate edge case.

kkysen: Are we able to get the span of the source and just manually search/re-lex for a comment?

randomPoison: Yeah, we can check the spans to at least detect the presence of a comment that's not attached to anything. I slapped together a prototype of it on another branch: https://github.com/immunant/c2rust/compare/legare/remove-blocks...legare/remove-blocks-but-handle-comments

But that feels very hacky to me, and this isn't the only place where this problem comes up. For example, the convert math/exits transforms can also eat comments in a similar way:

<details><summary>Examples of convert_math_funcs eating comments</summary>

```rust
let _ = sin(/* comment */ 1.0);
```

Becomes

```rust
let _ = 1.0.sin();
```

And

```rust
let _ = sin(
    // comment
    1.0
    // trailing comment
);
```

becomes

```rust
let _ = // comment
    1.0.sin();
```

</details>

It's not ideal, but for now I think the practical thing to do is accept that we can't always preserve comments in all cases. In general any attempt to rewrite a chunk of the AST is going to potentially eat comments. I think the best way to handle this (at least for now) is to leave it to a post-transform review pass (by a human or an LLM) to identify lost comments and restore them to an appropriate place.

randomPoison: Issue for tracking weird comments like this: https://github.com/immunant/c2rust/issues/1641

c2rust-refactor/tests/snapshots.rs
randomPoison: Yup, done.

c2rust-refactor/src/transform/funcs.rs
```diff
@@ -292,31 +295,38 @@ impl Transform for ToMethod {
...
-                    b.rules = BlockCheckMode::Default;
-                }
-            }
+        // HACK: Check if there have been compilation errors before running the
+        // transformation. When there are errors, rustc's unsafe checker can produce
+        // incorrect results, incorrectly marking blocks with actual unsafe
+        // operations as "unused". This would cause us to remove the `unsafe`
+        // keyword from blocks that still need it, so we bail out in that case
+        // instead of corrupting the code.
+        if cx.session().diagnostic().has_errors().is_some() {
```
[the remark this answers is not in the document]
randomPoison: Wrong for the version of the compiler we're using.

c2rust-refactor/src/transform/funcs.rs
```diff
@@ -325,6 +335,130 @@ impl Transform for FixUnusedUnsafe {
...
+            // Remove the block if there's nothing preventing us from doing so.
+            if !has_comments && !has_drop {
+                let mut stmts = mem::take(&mut block.stmts);
+
+                // If the block has a tail expr, turn it into a statement. This is valid
+                // because this is a block statement, which means that the tail expr isn't
+                // being used as part of an expression and so can (and must be) turned into
+                // a statement.
+                if let Some(last) = stmts.last_mut() && let StmtKind::Expr(expr) = &last.kind {
+                    last.kind = StmtKind::Semi(expr.clone());
```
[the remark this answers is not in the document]
randomPoison: I initially tried using `mem::take` to avoid the clone, but `Expr` doesn't implement `Default`, so I went with `clone` for simplicity. But we can manually make a default expr to swap with, so I've updated the code to avoid the clone.

PR #1614, opened 2026-02-25
Roll toolchain to nightly-2022-11-03

randomPoison:
~~Copilot choked and died on this, but Codex seems to have succeeded. I let it spin for like an hour and it got code compiling and tests running, so let's let CI run and see if anything is broken that I didn't catch locally.~~

Roll the nightly toolchain 6 weeks forward to nightly-2022-11-03. I've reviewed and cleaned up the changes and they all seem reasonable to me now, so this is ready for review.


c2rust-analyze/tests/filecheck/aggregate1.rs
```diff
@@ -8,7 +8,7 @@ struct S<'a> {
 pub fn aggregate() {
     // CHECK-DAG: pseudo-assigning fields Label { origin: Some(Origin([[R_U32_ORIGIN:[0-9]+]])){{.*}}}#&'a u32{{.*}}= Label { origin: Some(Origin({{.*}})){{.*}}&u32
     // CHECK-DAG: pseudo-assigning fields Label { origin: Some(Origin([[R_U32_ORIGIN:[0-9]+]])){{.*}}}#&'a u32{{.*}}= Label { origin: Some(Origin({{.*}})){{.*}}&u32
-    // CHECK-DAG: Aggregate literal label: Label{{.*}}origin_params: [('a, Origin([[R_U32_ORIGIN]])), ('h0, Origin({{.*}})]{{.*}}#S[]
+    // CHECK-DAG: Aggregate literal label: Label{{.*}}#S<'_>[]
```
ahomescu: I'm not sure how much we care about these.

c2rust-refactor/gen/ast.txt
```diff
@@ -292,7 +293,7 @@ enum ExprKind {
     Array(#[mac_table_seq] elems),
     ConstBlock(anon_const),
     Call(#[prec=POSTFIX] #[prec_special=Callee] func, #[mac_table_seq] args),
-    MethodCall(path_seg, #[prec_first=POSTFIX] #[mac_table_seq] args, span),
+    MethodCall(path_seg, #[prec=POSTFIX] receiver, #[mac_table_seq] args, span),
```
ahomescu, line 296: This is our internal attribute, I wonder why it decided this change was needed.

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -2267,22 +2272,32 @@ impl Builder {
...
             span: self.span,
         }
     }
 
     pub fn meta_item_attr(mut self, style: AttrStyle, meta_item: MetaItem) -> Self {
-        let mut attr = mk_attr_inner(meta_item);
-        attr.style = style;
-        self.attrs.push(attr);
+        self.attrs.push(Attribute {
+            id: AttrId::from_u32(0),
```
ahomescu, line 2299: We have a constant for this, `DUMMY_NODE_ID` I think.

c2rust-refactor/src/ast_manip/util.rs
```diff
@@ -26,8 +26,7 @@ impl PatternSymbol for Ident {
 impl PatternSymbol for Lit {
     fn pattern_symbol(&self) -> Option<Symbol> {
         match self.kind {
-            // FIXME: can this conflict with regular Err literals???
-            LitKind::Err(ref sym) => Some(*sym),
+            LitKind::Err => Some(self.token_lit.symbol),
```
ahomescu, line 33: Should we keep the comment?

c2rust-refactor/src/collapse/deleted.rs
```diff
@@ -77,12 +77,30 @@ impl<'a, 'ast> CollectDeletedNodes<'a, 'ast> {
...
-            ExprKind::Array(elements)
-            | ExprKind::Call(_, elements)
-            | ExprKind::MethodCall(_, elements, _)
-            | ExprKind::Tup(elements) => {
+            ExprKind::Array(elements) | ExprKind::Call(_, elements) | ExprKind::Tup(elements) => {
                 self.handle_seq(x.id, elements);
             }
+            ExprKind::MethodCall(_, receiver, elements, _) => {
+                let mut history = Vec::with_capacity(elements.len() + 1);
+                for n in std::iter::once(receiver).chain(elements.iter()) {
```
ahomescu: No idea if this is correct, we'll have to take a close look at it later.

c2rust-refactor/src/driver.rs
```diff
@@ -314,7 +310,7 @@ where
     // Force disable incremental compilation.  It causes panics with multiple typechecking.
     config.opts.incremental = None;
 
-    run_in_thread_pool_with_globals(Edition::Edition2021, 1, move || {
+    rustc_span::create_session_globals_then(Edition::Edition2021, move || {
```
ahomescu, line 205: This seems extremely questionably, `create_session_globals_then` was around in the old nightly, and the other function did a lot more than just call it. We'll need to check what happened upstream.

c2rust-refactor/src/driver.rs
```diff
@@ -456,21 +451,6 @@ fn build_session(
...
-    let source_map = Rc::new(SourceMap::with_file_loader_and_hash_kind(
-        file_loader,
-        sopts.file_path_mapping(),
-        hash_kind,
-    ));
-    // Put a dummy file at the beginning of the source_map, so that no real `Span` will accidentally
-    // collide with `DUMMY_SP` (which is `0 .. 0`).
-    source_map.new_source_file(FileName::Custom("<dummy>".to_string()), " ".to_string());
-
     let codegen_backend = get_codegen_backend(
```
ahomescu: Really weird that it decided to delete a bunch of code here, we need to check upstream.

Rua:
Why was that particular nightly version chosen and not, say, a much newer one?

kkysen (commented):
> Why was that particular nightly version chosen and not, say, a much newer one?

We were mostly just testing out codex on this.  We decided to first start with a smaller jump so that it could handle it better.  We may try a longer jump next, but there's a larger risk things go wrong, and in case we introduce some regression, it'll be harder to debug/rollback, and a much larger set of changes will be much more difficult to review as well.

c2rust-refactor/gen/ast.txt
randomPoison: This is necessary because the method call receiver was pulled out of the args list into its own field. The old `prec_first` attribute was specifically there to handle this case in the method call syntax, where we needed to give the first element of the args list different precedence. Now that the receiver is its own field, the `prec` attribute is the one to use.

Also, `prec_first` was only used in this one place, so we can probably remove it entirely after rolling the toolchain (unless you think we should keep it around in case that case comes up again in a future toolchain).

c2rust-refactor/src/ast_builder/builder.rs
randomPoison: `DUMMY_NODE_ID` is a `NodeId`, this as an `AttrId`. Other code within this file is doing `AttrId::from_u32(0)`, hence why it did the same here.

c2rust-refactor/src/ast_manip/util.rs
randomPoison: I think this comment is now irrelevant, though what we're doing here is a little weird so it's a bit hard to tell. This functionality is used in the matcher syntax, so we can do matches like `$x:Lit`. This ends up getting parsed in the AST as a `LitKind::Err`, since it's not a valid literal, and we then extract the `$x` part of the symbol from the error. This is a bit weird because we're overloading `LitKind::Err` for a case that's not actually an error, but it seems to work for our purposes.

We didn't have any tests around matching literals, so I've added a couple snapshot tests to cover that, along with adding support for substituting the matched literal into the rewritten code, which didn't previously work. I think we can remove the FIXME comment given that we can now be somewhat confident that it works as intended.

c2rust-refactor/src/ast_manip/util.rs
```diff
@@ -77,9 +76,7 @@ impl PatternSymbol for Stmt {
 impl PatternSymbol for Pat {
     fn pattern_symbol(&self) -> Option<Symbol> {
         match self.kind {
-            PatKind::Ident(BindingMode::ByValue(Mutability::Not), ref i, None) => {
-                i.pattern_symbol()
-            }
+            PatKind::Ident(BindingAnnotation(ByRef::No, _), ref i, None) => i.pattern_symbol(),
```
randomPoison, line 79: Oh weird, this one actually fixed a bug in the rewriter where we weren't handling mut patterns correctly. I added tests to exercise the relevant code paths and confirmed that the new behavior is correct.

c2rust-refactor/src/collapse/deleted.rs
randomPoison: Ah, it just duplicated the logic from `handle_seq` because `handle_seq` expects a slice and now the new shape for `MethodCall` splits the receiver from the args array. I've updated `handle_seq` to accept an iterator, and removed the duplicated logic.

c2rust-refactor/src/driver.rs
randomPoison: `run_in_thread_pool_with_globals` was made private (well, `pub(crate)`) so we can't call it anymore. `create_session_globals` is used internally in `run_in_thread_pool_with_globals`, so it makes sense that that's the fix here. It doesn't seem like we need to thread pool to run the refactorer here, so maybe `create_session_globals` does what we need? At least it doesn't seem to be interfering with our ability to run the refactorer.

randomPoison: Oh, we were specifically running it with only 1 thread before, so I think we can do without the thread pool.

c2rust-refactor/src/driver.rs
randomPoison: Yeah this one is very weird. I dug into it and can't actually find anything that changed upstream that necessitated any changes here. I've put things back the way they were, but it's worth noting that this function `build_session` is dead code, we don't call it from anywhere afaict. Do you know what purpose this was meant to serve? If we're not using it, we may just want to delete it entirely.

randomPoison: I dug into the the history to determine when this stuff became dead, looks like the only call to it was removed in [65d1abae9 ](https://github.com/immunant/c2rust/commit/65d1abae9ef2d3bb291b635cbc2e4a37dc0ed21c#diff-835fcd65bdd7b79f53678278ec0c7f38a94173326c74106845f4f251fd6ca7feL126-L133) back in 2019. I've opened https://github.com/immunant/c2rust/pull/1664 to delete these functions along with the commented-out code in this file, if there's some reason to keep these around we can discuss it over there.

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -206,6 +206,13 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                     None => {} // not a ptr cast (no dataflow constraints needed); let rustc typeck this
                 };
             }
+            CastKind::DynStar
+            | CastKind::IntToInt
+            | CastKind::FloatToInt
+            | CastKind::FloatToFloat
+            | CastKind::IntToFloat => {
+                // TODO: Do we need to do anything for these casts?
+            }
```
kkysen:
```suggestion
            CastKind::IntToInt
            | CastKind::FloatToInt
            | CastKind::FloatToFloat
            | CastKind::IntToFloat => {
                // Not ptr casts, and we don't allow ptr-to-int casts.
            }
            CastKind::DynStar => todo!("dyn* casts are too unstable and rare to bother supporting"),
```

c2rust-analyze/src/last_use.rs
```diff
@@ -200,11 +200,16 @@ impl ActionsBuilder {
             StatementKind::Retag(..) => panic!("unexpected StatementKind::Retag"),
             StatementKind::AscribeUserType(..) => {}
             StatementKind::Coverage(..) => {}
-            StatementKind::CopyNonOverlapping(ref cno) => {
-                self.push_operand(&cno.src, loc, WhichPlace::Operand(0));
-                self.push_operand(&cno.dst, loc, WhichPlace::Operand(0));
-                self.push_operand(&cno.count, loc, WhichPlace::Operand(0));
-            }
+            StatementKind::Intrinsic(ref intrinsic) => match &**intrinsic {
```
kkysen, line 203: Do you need `ref` if you're double dereferencing with `**`?

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -516,36 +513,48 @@ impl Builder {
         let func: Path = vec![func].make(&self);
 
         let args = MacArgs::Delimited(DelimSpan::dummy(), MacDelimiter::Parenthesis, {
-            let mut builder = TokenStreamBuilder::new();
+            let mut tokens = Vec::new();
```
kkysen, line 516:
```suggestion
            let mut tokens = Default::default();
```

That should work with future changes better (IIRC, there's another one coming).

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -2267,22 +2282,32 @@ impl Builder {
...
-                None,
-            ),
+                tokens: None,
+            })),
             span: self.span,
         }
     }
 
     pub fn meta_item_attr(mut self, style: AttrStyle, meta_item: MetaItem) -> Self {
-        let mut attr = mk_attr_inner(meta_item);
```
kkysen, line 2283: What about `fn mk_attr`?

c2rust-refactor/src/matcher/bindings.rs
```diff
@@ -340,7 +340,7 @@ fn maybe_get_type(c: &mut Cursor) -> Type {
 
 /// Rewrite tokens like `$foo:ty` into `$foo` and extract the types
 fn rewrite_token_stream(ts: TokenStream, bt: &mut BindingTypes) -> TokenStream {
-    let mut tsb = TokenStreamBuilder::new();
+    let mut tokens = Vec::new();
```
kkysen, line 343:
```suggestion
    let mut tokens = Default::default();
```

c2rust-refactor/src/matcher/subst.rs
```diff
@@ -68,7 +69,7 @@ impl<'a, 'tcx> SubstFolder<'a, 'tcx> {
     }
 
     fn subst_token_stream_bindings(&mut self, ts: TokenStream) -> TokenStream {
-        let mut tsb = TokenStreamBuilder::new();
+        let mut trees = Vec::new();
```
kkysen, line 72:
```suggestion
        let mut trees = Default::default();
```

c2rust-refactor/src/driver.rs
```diff
@@ -472,7 +468,6 @@ fn build_session(
     // Put a dummy file at the beginning of the source_map, so that no real `Span` will accidentally
     // collide with `DUMMY_SP` (which is `0 .. 0`).
     source_map.new_source_file(FileName::Custom("<dummy>".to_string()), " ".to_string());
-
```
kkysen: Not that important, but where did this come from?

c2rust-refactor/src/driver.rs
```diff
@@ -634,21 +628,15 @@ pub fn parse_block(sess: &Session, src: &str) -> P<Block> {
...
     // `parse_arg` is private, so we make do with `parse_attribute`,
     // `parse_pat`, & `parse_ty`.
-    const INNER_ATTR_FORBIDDEN: InnerAttrPolicy<'_> = InnerAttrPolicy::Forbidden {
-        reason: "inner attributes not allowed in function arguments",
-        saw_doc_comment: false,
-        prev_outer_attr_sp: None,
-    };
-
-    let mut attrs: Vec<ast::Attribute> = Vec::new();
+    let mut attrs = ast::AttrVec::new();
```
kkysen, line 407:
```suggestion
    let mut attrs = Default::default();
```

c2rust-refactor/tests/snapshots.rs
```diff
@@ -302,6 +302,30 @@ fn test_matcher_def() {
...
+        .test();
+}
+
+#[test]
+fn test_matcher_pat_mut() {
+    refactor("rewrite_stmts")
+        .command_args(&["let mut __p = __e;", "let __p = __e + 1;"])
+        .named("matcher_pat_mut.rs")
+        .test();
+}
```
kkysen, line 349: What are the new tests for?

pdg/src/snapshots/c2rust_pdg__tests__analysis_tests_misc_pdg_snapshot_debug.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: pdg/src/main.rs
-assertion_line: 376
+assertion_line: 386
```
kkysen, line 3: Was this updated by Codex or by `insta`?  Because I thought `insta` got rid of this line.  Although maybe `c2rust-pdg` is using an old `insta` version?  I ask because I noticed Codex added this in one of the transpiler snapshots when it shouldn't have when I was using it to update the macos and aarch64 snapshots.

pdg/src/snapshots/c2rust_pdg__tests__analysis_tests_misc_pdg_snapshot_debug.snap
```diff
@@ -211,9 +211,9 @@ g is_null=false {
 	n[16]: project[0]@0 n[15] => _   @ bb13[2]:  fn exercise_allocator;  _32 = ((*_33).0: i32);
 	n[17]: addr.load    n[15] => _   @ bb13[2]:  fn exercise_allocator;  _32 = ((*_33).0: i32);
 	n[18]: copy         n[1]  => _43 @ bb21[6]:  fn exercise_allocator;  _43 = _1;
-	n[19]: copy         n[18] => _42 @ bb21[7]:  fn exercise_allocator;  _42 = move _43 as *mut libc::c_void (Misc);
+	n[19]: copy         n[18] => _42 @ bb21[7]:  fn exercise_allocator;  _42 = move _43 as *mut libc::c_void (PtrToPtr);
 	n[20]: copy         n[1]  => _4  @ bb0[1]:   fn reallocarray;        _4 = _1;
-	n[21]: copy         n[20] => _1  @ bb1[3]:   fn reallocarray;        _0 = const pointers::REALLOC(move _4, move _5);
+	n[21]: copy         n[20] => _1  @ bb1[3]:   fn reallocarray;        _0 = const _(move _4, move _5);
```
kkysen, line 216: Do we know what's happening here?

.gitignore
```diff
@@ -58,3 +58,6 @@ c2rust-transpile/tests/snapshots/**/*.rs
 c2rust-transpile/tests/snapshots/**/*.c_decls.json
 c2rust-transpile/tests/c_decls_snapshots/**/*.rs
 c2rust-transpile/tests/c_decls_snapshots/**/*.c_decls.json
+
+# c2rust-analyze test outputs
+c2rust-analyze/c2rust_analysis_tests_minimal-*
```
kkysen, line 63: Is this that random binary that's sometimes generated?

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-08-08"
+channel = "nightly-2022-11-03"
```
kkysen, line 2: There are a bunch of other `nightly-2022-08-08`s.  Can you update all of them?

c2rust-refactor/src/collapse/deleted.rs
```diff
@@ -77,12 +79,12 @@ impl<'a, 'ast> CollectDeletedNodes<'a, 'ast> {
...
         match &x.kind {
-            ExprKind::Array(elements)
-            | ExprKind::Call(_, elements)
-            | ExprKind::MethodCall(_, elements, _)
-            | ExprKind::Tup(elements) => {
+            ExprKind::Array(elements) | ExprKind::Call(_, elements) | ExprKind::Tup(elements) => {
                 self.handle_seq(x.id, elements);
             }
+            ExprKind::MethodCall(_, receiver, elements, _) => {
+                self.handle_seq(x.id, std::iter::once(receiver).chain(elements.iter()));
```
kkysen:
```suggestion
            ExprKind::MethodCall(_, recv, args, _) => {
                self.handle_seq(x.id, std::iter::once(recv).chain(args));
```

This is what we call them everywhere else.

Also, the `.iter()` on `args` isn't necessary.

c2rust-refactor/src/collapse/deleted.rs
```diff
@@ -254,12 +256,21 @@ impl<'a, 'ast> MutVisitor for RestoreDeletedNodes<'a, 'ast> {
...
+                let mut all = Vec::with_capacity(elements.len() + 1);
+                all.push(std::mem::replace(receiver, mk().err_expr()));
+                all.extend(std::mem::take(elements).into_iter());
+
+                self.restore_seq(id, &mut all);
+
+                // Pull the new receiver and elements out of the merged list.
+                *receiver = all.remove(0);
+                *elements = all;
+            }
```
kkysen: This is a really weird implementation.

```suggestion
            ExprKind::MethodCall(_, recv, args, _) => {
                let mut elements = std::iter::once(mem::replace(recv, mk().err_expr()))
                    .chain(mem::take(args))
                    .collect();
                self.restore_seq(id, &mut elements);
                *recv = elements.remove(0);
                *args = elements;
            }
            _ => {}
```

c2rust-refactor/src/collapse/deleted.rs
```diff
@@ -44,11 +45,12 @@ struct CollectDeletedNodes<'a, 'ast> {
...
 impl<'a, 'ast> CollectDeletedNodes<'a, 'ast> {
-    fn handle_seq<T>(&mut self, parent: NodeId, nodes: &'ast [T])
+    fn handle_seq<T, I>(&mut self, parent: NodeId, nodes: I)
     where
-        T: GetNodeId + ListNodeIds + AsMacNodeRef,
+        I: IntoIterator<Item = &'ast T>,
+        T: GetNodeId + ListNodeIds + AsMacNodeRef + 'ast,
     {
-        let mut history = Vec::with_capacity(nodes.len());
+        let mut history = Vec::new();
```
kkysen:
```suggestion
        let nodes = nodes.into_iter();
        let mut history = Vec::with_capacity(nodes.size_hint().0);
```

This would keep the previous behavior.  However, we do call `history.clone()`, but don't call `.shrink_to_fit()`, so we are wasting memory there with the previous implementation, although there are much fewer reallocations.

c2rust-refactor/src/ast_manip/util.rs
kkysen: Could we move this fix to its own PR?  This PR is already quite large, so fixing the bug separately would be helpful.

c2rust-refactor/src/illtyped.rs
```diff
@@ -155,10 +155,13 @@ impl<'a, 'tcx, F: IlltypedFolder<'tcx>> MutVisitor for FoldIlltyped<'a, 'tcx, F>
...
-            ExprKind::MethodCall(_seg, args, _span) => {
+            ExprKind::MethodCall(_seg, recv, args, _span) => {
                 if let Some(fn_sig) = opt_fn_sig {
+                    if let Some(&ty) = fn_sig.inputs().get(0) {
+                        illtyped |= self.ensure(recv, ty);
+                    }
                     for (i, arg) in args.iter_mut().enumerate() {
-                        if let Some(&ty) = fn_sig.inputs().get(i) {
+                        if let Some(&ty) = fn_sig.inputs().get(i + 1) {
                             illtyped |= self.ensure(arg, ty);
```
kkysen, line 162: We have a bunch of different ways of handling the `recv`/`args` split.  Could we try to collapse that to just one?  That is, have a `fn combine_recv_and_args(recv: P<Expr>, args: Vec<P<Expr>>) -> impl Iterator<Item = P<Expr>>` that we use everywhere.  We can iterate over it, use `.zip` (e.g., here with `fn_sig.inputs()`), `.nth` (would work here, too), `.enumerate`, which I think should cover almost all of the use cases and keeps the usage much more standardized.

kkysen (changes requested):

c2rust-refactor/src/transform/literals.rs
```diff
@@ -600,15 +600,19 @@ impl<'a, 'kt, 'tcx> UnifyVisitor<'a, 'kt, 'tcx> {
...
                 }
 
-                self.visit_path_segment(ex.span, segment);
+                self.visit_path_segment(segment);
                 if let Some(&[ref input_key_trees @ .., output_key_tree]) =
                     callee_key_tree.get().children()
                 {
-                    for (arg_expr, arg_key_tree) in args.iter().zip(input_key_trees.iter()) {
+                    for (arg_expr, arg_key_tree) in std::iter::once(recv)
+                        .chain(args.iter())
```
kkysen:
```suggestion
                    for (arg_expr, arg_key_tree) in std::iter::once(recv)
                        .chain(args)
```

c2rust-analyze/src/last_use.rs
randomPoison: Yeah, we need to double-dereference it, and we need `ref` otherwise we'd have to change the surrounding `match`. `intrinsic` is a `&Box<NonDivergingIntrinsic>`, so we need to dereference it twice to get a `&NonDivergingIntrinsic` that we can actually match on. I considered removing the `ref`, but that would require changing the rest of the match arms, which I didn't want to do in order to keep the change minimal.

c2rust-refactor/src/ast_builder/builder.rs
randomPoison: This doesn't work, type inference fails when using `Default::default()`. Though I'm not really sure why, I would think that passing `tokens` to `TokenStream::new` would give it the necessary type information because `new` directly takes a `Vec<TokenTree>`, but ¯\\\_(ツ)\_/¯

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -206,11 +206,15 @@ impl<'tcx> TypeChecker<'tcx, '_> {
...
-            | CastKind::IntToInt
+            CastKind::IntToInt
             | CastKind::FloatToInt
             | CastKind::FloatToFloat
-            | CastKind::IntToFloat => {}
+            | CastKind::IntToFloat => {
+                // Not ptr casts, and we don't allow ptr-to-int casts.
+            }
+            CastKind::DynStar => {
+                unimplemented!("dyn* casts are too unstable and rare to bother supporting")
```
ahomescu, line 216: Weird, why did it add this?

c2rust-analyze/src/dataflow/type_check.rs
```diff
@@ -735,11 +739,12 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                                 continue;
                             }
                             Rvalue::Cast(
-                                CastKind::PtrToPtr | CastKind::FnPtrToPtr,
+                                CastKind::PtrToPtr | CastKind::FnPtrToPtr | CastKind::IntToInt,
```
ahomescu, line 742: Is this correct? `IntToInt` might truncate or extend.

c2rust-analyze/src/last_use.rs
```diff
@@ -200,7 +200,16 @@ impl ActionsBuilder {
...
             StatementKind::AscribeUserType(..) => {}
             StatementKind::Coverage(..) => {}
-            StatementKind::Intrinsic(..) => {}
+            StatementKind::Intrinsic(ref intrinsic) => match &**intrinsic {
+                NonDivergingIntrinsic::Assume(op) => {
+                    self.push_operand(op, loc, WhichPlace::Operand(0));
+                }
+                NonDivergingIntrinsic::CopyNonOverlapping(cno) => {
+                    self.push_operand(&cno.src, loc, WhichPlace::Operand(0));
+                    self.push_operand(&cno.dst, loc, WhichPlace::Operand(0));
```
ahomescu: Should these be 1 and 2?

c2rust-refactor/gen/ast.txt
ahomescu: Got it, yeah it makes sense now.

c2rust-refactor/src/ast_manip/util.rs
ahomescu: In the AST this pattern is for "Placeholder for a literal that wasn't well-formed in some way". Weird that we're using the symbol from that case, I'm not sure what's going on in this function but let's keep it.

c2rust-refactor/src/driver.rs
ahomescu: rustc still has `run_in_thread_with_globals` which still spawn a separate thread in that case, but that function is also private. I guess we can live without for now.

c2rust-refactor/src/ast_builder/builder.rs
kkysen: Ahh, weird, but yeah, if it doesn't work, then ignore my suggestion.

c2rust-analyze/src/dataflow/type_check.rs
kkysen: I added this.

c2rust-refactor/src/ast_builder/builder.rs
randomPoison: `mk_attr` and `mk_attr_inner` were changed to now take an `AttrIdGenerator` as an argument. The generator exists on the `Session` object, elsewhere we do `&sess.parse_sess.attr_id_generator`, but we don't have access to that in the AST builder. So this function was changed to follow the approach used in the `attribute` function above.

If we want to cut down on duplication a bit, we could rework this function to use the `attribute` function above, since that one is a more general function for creating attributes. However, it's worth noting that both of these functions are currently dead code, so I don't think it makes sense to worry about keeping them clean for now.

c2rust-refactor/src/driver.rs
randomPoison: Codex originally rewrote a chunk of this code, but I reverted that change here and opened https://github.com/immunant/c2rust/pull/1664 to delete the code entirely. I just accidently left this deleted line when I reverted the change.

c2rust-refactor/src/ast_builder/builder.rs
kkysen: Ahh, I think it's probably fine then.  I just wanted to make sure something wasn't missed.

c2rust-analyze/src/last_use.rs
randomPoison: Yeah that would make sense, though I don't know enough about c2rust-analyze to know for sure. I'll make that tweak here, but I think without test coverage or more investigation I don't have a way to confirm if that's right.

c2rust-analyze/src/dataflow/type_check.rs
randomPoison: The comment below this specifically mentions covering casts like `usize` to `size_t`, so this seems reasonable, but without spending more time digging I can't be sure.

c2rust-refactor/src/illtyped.rs
randomPoison: Hm, I don't think we need a helper function here, I think the common pattern of `iter::once(recv).chain(args)` is good enough. I did a pass over the places where we're handling `MethodCall` and unified a couple to use that approach.

.gitignore
randomPoison: Yup. I didn't wanna mess around with c2rust-analyze too much so I didn't bother digging into the tests to see why this was being generated. Ignoring the file is probably fine for our purposes here.

pdg/src/snapshots/c2rust_pdg__tests__analysis_tests_misc_pdg_snapshot_debug.snap
randomPoison: I think Codex did this. If I delete the file and then re-run insta it doesn't generate this line.

c2rust-refactor/tests/snapshots.rs
randomPoison: These were attempting to cover some of the parts of c2rust-refactor that were touched by the update. The first two test the `Lit` matcher in `rewrite_expr`, the last one tests that capturing a pattern with the `__p` matcher preserves the `mut` keyword in the rewritten code.

I think for future toolchain rolls I want to add tests like this in a separate PR ahead of the toolchain roll, but for expedience in this case I'd like to keep these in this PR.

c2rust-refactor/src/ast_manip/util.rs
randomPoison: For future toolchain rolls I agree test fixes like this should be split up into separate PRs, but for this one I'd like to keep it part of this PR because I don't think it's worth the extra time it'd take to split out a separate PR.

pdg/src/snapshots/c2rust_pdg__tests__analysis_tests_misc_pdg_snapshot_debug.snap
randomPoison: I think all of the changes in the snapshot file come from differences in how rustc internal stuff gets printed, there are no changes in the pdg code itself.

c2rust-analyze/src/borrowck/mod.rs
```diff
@@ -260,7 +260,7 @@ fn run_polonius<'tcx>(
     //pretty::write_mir_fn(tcx, mir, &mut |_, _| Ok(()), &mut std::io::stdout()).unwrap();
```
ahomescu, line 260: It would save us a lot of review time (and the agent itself a lot of cycles) if it updated every c2rust component separately next time. Then we could prioritize `c2rust-refactor` higher, and leave the others for later (I don't know how much we care about the ALLSTAR stuff right now).

c2rust-bitfields-derive/Cargo.toml
```diff
@@ -12,7 +12,7 @@ keywords.workspace = true
 categories.workspace = true
 
 [dependencies]
-# nightly-2022-08-08 (1.65) and nightly-2023-04-15 (1.70) require these exact versions.
+# nightly-2022-11-03 (1.65) and nightly-2023-04-15 (1.70) require these exact versions.
```
ahomescu, line 15: Is the new nightly still 1.65?

c2rust-refactor/src/analysis/type_eq.rs
```diff
@@ -859,12 +860,12 @@ impl<'lty, 'a, 'hir> Visitor<'hir> for UnifyVisitor<'lty, 'hir> {
...
             PatKind::Path(..) => {} // TODO
 
-            PatKind::Tuple(ref ps, None) => {
+            PatKind::Tuple(ref ps, dotdot) if dotdot.as_opt_usize().is_none() => {
                 for (expected, p) in rty.args.iter().zip(ps.iter()) {
                     self.ltt.unify(expected, self.pat_lty(p));
                 }
             }
-            PatKind::Tuple(ref _ps, Some(_dotdot_idx)) => {} // TODO
+            PatKind::Tuple(ref _ps, _dotdot) => {} // TODO
```
ahomescu, line 868: Maybe add a `todo!` here? I'd prefer if it crashed instead of silently failing.

c2rust-refactor/src/collapse/deleted.rs
```diff
@@ -44,11 +45,13 @@ struct CollectDeletedNodes<'a, 'ast> {
 }
 
 impl<'a, 'ast> CollectDeletedNodes<'a, 'ast> {
-    fn handle_seq<T>(&mut self, parent: NodeId, nodes: &'ast [T])
+    fn handle_seq<T, I>(&mut self, parent: NodeId, nodes: I)
```
ahomescu, line 48: This is a very interesting change (no complaints, it looks legit).

c2rust-refactor/src/context.rs
```diff
@@ -90,6 +90,8 @@ fn hir_id_to_span(id: HirId, hir_map: hir_map::Map) -> Option<NodeSpan> {
         // We do not have a SpanNodeKind for certain nodes
         Some(Node::TypeBinding(_)) => None,
         Some(Node::TraitRef(_)) => None,
+        Some(Node::ExprField(_)) => None,
```
ahomescu, line 93: These aren't technically `FieldDef`s. `ExprField` and `PatField` contain an `Expr` and `Pat`, respectively, so I'm wondering if we do need a new `NodeSpanKind` here or two. Maybe one shared between the two of them:
```
        Some(Node::ExprField(expr)) => Some(NodeSpan::new(expr.span, Field)),
        Some(Node::PatField(pat)) => Some(NodeSpan::new(pat.span, Field)),
```
since `ExprField` and `PatField` can't collide with each other, but they may collide with their inner values.

Otoh since `ExprField` is going to be something like `ident: value`, maybe there isn't the chance of a collision. Do we have a test for these?

c2rust-analyze/src/borrowck/mod.rs
randomPoison: Yeah I agree. I think before attempting larger toolchain rolls we need to trim down the code and disable stuff we don't actively want to support. There are a lot of changes in code that we're effectively treating as dead, and that adds a lot of noise to the PR. We also need to revive more of the c2rust-refactor test suite, since there are a bunch of transforms that are currently dead, making changes to that code hard to verify and thus harder to review.

c2rust-bitfields-derive/Cargo.toml
randomPoison: Ah, no it's 1.67 now. I've updated some instances of 1.65 to 1.67 where it seemed appropriate.

c2rust-refactor/src/collapse/deleted.rs
randomPoison: iirc this was my change. Codex had duplicated the logic in here somewhere else because the receiver for method calls was split out from the args list (i.e. previously the receiver was part of the args, now we get the receiver and the rest of the args separately). I changed this to take an iterator so that  I could do `iter::once(recv).chain(args)` and pass the combined list into here.

c2rust-refactor/src/context.rs
randomPoison: Hm, I don't think we need new node types, we already have `ExprField` and `NodeField` that we were using in a couple of places. I've updated this function to return span info for those.

What exactly do you mean by "collision" here? Are you referring to overlapping spans? My understanding is that `ExprField` and `PatField` can't overlap with each other because they show up in different positions (expressions vs patterns), but maybe you're referring to something else.

As far as I know none of our existing transforms hit this code path, so we don't have any tests to cover it. I can try adding one, but I'm hesitant to put together a contrived test for this functionality. I'd rather have tests for real transforms that naturally hit this code path, and I don't see a good option for that currently.

ahomescu: >  Are you referring to overlapping spans?

That's right, we were getting multiple nodes with identical spans because sometimes an AST node will include an inner node with the same span. For example, a `PathExpr` and the inner `Path` node overlap.

>  My understanding is that `ExprField` and `PatField` can't overlap with each other

That's right, but I was worried that `ExprField` and the inner `Expr` might collide. On second thought, that's probably not the case.

ahomescu (approved):

c2rust-refactor/src/ast_manip/util.rs
kkysen: Sure I guess.

kkysen (changes requested):
There's one more `nightly-2022-08-08` left in `c2rust-rust-tools/src/lib.rs`.  Otherwise, I think it LGTM now.

c2rust-refactor/src/context.rs
fw-immunant: Would field shorthand e.g. in `let field = ...; FooStruct { field }` cause the spans to collide? The shorthand is usable in both expressions and patterns.

randomPoison:
@kkysen Ah right good catch. I updated the comment to not specifically reference the exact nightly, I'd prefer that we don't have to keep updating references to the exact nightly every time we do a toolchain roll.

kkysen:
> I'd prefer that we don't have to keep updating references to the exact nightly every time we do a toolchain roll.

I usually do because it makes it very easy to find.  For example, once that nightly toolchain reaches edition 2024, that comment should be removed/changed, but it's now harder to find it.

randomPoison:
Ah sure, I see how that'd be useful. Changed it to keep the comment and just update the nightly name.

kkysen (approved):

PR #1618, opened 2026-02-25
refactor: tests: port tests that panic with `#[should_panic]`

kkysen:
I still added the snapshots that should be expected based on the `new.rs` files, so if we do fix the panic, we still have the snapshot test.


c2rust-refactor/tests/snapshots.rs
```diff
@@ -289,6 +289,18 @@ fn test_matcher_typed() {
         .test();
 }
 
+/// TODO Broken.
+/// This panics with the error:
+///
+/// ```shell
+///     thread 'rustc' panicked at 'Could not find an HIR id for NodeId NodeId(61); span=Some(NodeSpan { span: Span { lo: BytePos(237), hi: BytePos(325), ctxt: #0 }, kind: Stmt }), context=Some(NodeContextKey { stmt_index: Some(6), child_path: [], symbol: None, owner: Some(NodeId(8)) })', c2rust-refactor/src/context.rs:927:13
```
ahomescu, line 296: I don't think this is intended, looks like a c2rust-refactor bug.

kkysen: Yeah I agree. That's why I said this transform is broken.

ahomescu: In that case, I'm debating between 1) not adding it at all, 2) adding it commented out, or 3) adding it with the attribute. Although I guess it would be useful to know when it's fixed.

ahomescu (approved):

c2rust-refactor/tests/snapshots.rs
kkysen: I was considering that, too.  My thinking was that for 1, we may want it in the future at some point, so good to keep it around.  And then 3 seems strictly better than 2 since it won't bitrot as much and if it does get fixed, we'd know.

Issue #1605, opened 2026-02-21
c2rust-refactor commit command is broken


ahomescu:
When running multiple transforms with `commit` between them (even `noop ; commit ; noop`), the refactorer crashes with an array index error. This is likely caused by us not resetting some session globals between rustc invocations (my guess is `HygieneData`).

PR #1619, opened 2026-02-25
refactor: tests: add remaining tests (multi-command ones) besides ones with `select`

### LINKED ISSUE #1453: Refactorer tests are not run in CI

kkysen:
* Fixes #1453.

This adds all multi-command refactor tests besides tests with `select`.

This runs multiple refactor commands with just multiple `refactor(...).test()`s.  This generates a separate snapshot for each command, but this should be fine, and gives us more data to inspect if anything goes wrong.

The old tests used `;` for this, which avoids running the refactorer again.  Now we run the refactorer from scratch for each command.  This avoids the need to `commit` in between certain commands in order to re-evaluate the AST, as `commit` is broken (see #1605).


kkysen (commented):
Does this look like it should work?  I'll add the rest of them if so.

c2rust-refactor/tests/snapshots.rs
```diff
@@ -216,6 +216,24 @@ fn test_refactor(
 
 // NOTE: Tests should be listed in alphabetical order.
 
+#[test]
+fn test_autoretype_array() {
+    refactor("rewrite_expr")
+        .command_args(&["1 + 1", "2"])
+        .named("autoretype_array.rs")
+        .test();
+    refactor("autoretype").named("autoretype_array.new").test();
```
ahomescu, line 245: If we had to add a third command, would it also use `.named("autoretype_array.new")`?

kkysen: Yeah. Kind of confusing, but it works.

c2rust-refactor/tests/snapshots.rs
```diff
@@ -216,6 +215,45 @@ fn test_refactor(
...
+        .test();
+    // no commit
+    refactor("abstract")
+        .command_args(&[
+            "sub<T: Sub<T, Result=T>>(x: T, y: T) -> T",
+            "typed!(x, T) - y",
+            "x - y",
+        ])
+        .named("abstract.new")
+        .expect_compile_error(true)
```
kkysen, line 235: @ahomescu, see, it turned out to be useful to keep for a second command.

PR #1620, opened 2026-02-25
transpile: Translate enums with newtype structs

Rua:
- Depends on #1889
- Depends on #1891
- Depends on #1921
- Depends on #1956
- Depends on #1958

### DIFF
--- a/c2rust-transpile/src/translator/enums.rs
+++ b/c2rust-transpile/src/translator/enums.rs
@@ -7,8 +7,8 @@ use crate::{
     diagnostics::TranslationResult,
     translator::{signed_int_expr, ConvertedDecl, ExprContext, Translation},
     with_stmts::WithStmts,
-    CDeclKind, CEnumConstantId, CEnumId, CExprId, CExprKind, CLiteral, CQualTypeId, CTypeId,
-    CTypeKind, ConstIntExpr,
+    CDeclKind, CEnumConstantId, CEnumId, CExprId, CExprKind, CLiteral, CQualTypeId, CTypeKind,
+    ConstIntExpr,
 };
 
 impl<'c> Translation<'c> {
@@ -59,8 +59,8 @@ impl<'c> Translation<'c> {
         ))
     }
 
-    pub fn convert_enum_zero_initializer(&self, type_id: CTypeId) -> WithStmts<Box<Expr>> {
-        WithStmts::new_val(self.enum_for_i64(type_id, 0))
+    pub fn convert_enum_zero_initializer(&self, enum_id: CEnumId) -> WithStmts<Box<Expr>> {
+        WithStmts::new_val(self.enum_for_i64(enum_id, 0))
     }
 
     /// Translates a `DeclRef` for an `EnumConstant`.
@@ -97,7 +97,6 @@ impl<'c> Translation<'c> {
     pub fn convert_cast_to_enum(
         &self,
         ctx: ExprContext,
-        enum_type_id: CQualTypeId,
         enum_id: CEnumId,
         expr: Option<CExprId>,
         mut val: Box<Expr>,
@@ -122,7 +121,7 @@ impl<'c> Translation<'c> {
                 }
 
                 CExprKind::Literal(_, CLiteral::Integer(i, _)) => {
-                    val = self.enum_for_i64(enum_type_id.ctype, i as i64);
+                    val = self.enum_for_i64(enum_id, i as i64);
                     return Ok(WithStmts::new_val(val));
                 }
 
@@ -130,7 +129,7 @@ impl<'c> Translation<'c> {
                     if let &CExprKind::Literal(_, CLiteral::Integer(i, _)) =
                         &self.ast_context.index_unwrap_parens(subexpr_id).kind
                     {
-                        val = self.enum_for_i64(enum_type_id.ctype, -(i as i64));
+                        val = self.enum_for_i64(enum_id, -(i as i64));
                         return Ok(WithStmts::new_val(val));
                     }
                 }
@@ -139,20 +138,14 @@ impl<'c> Translation<'c> {
             }
         }
 
-        let target_ty = self.convert_type(enum_type_id.ctype)?;
-        val = mk().cast_expr(val, target_ty);
+        val = self.enum_constructor_expr(enum_id, val);
 
         Ok(WithStmts::new_val(val))
     }
 
     /// Given an integer value this attempts to either generate the corresponding enum
     /// variant directly, otherwise it converts a number to the enum type.
-    fn enum_for_i64(&self, enum_type_id: CTypeId, value: i64) -> Box<Expr> {
-        let enum_id = match self.ast_context.resolve_type(enum_type_id).kind {
-            CTypeKind::Enum(enum_id) => enum_id,
-            _ => panic!("{:?} does not point to an `enum` type", enum_type_id),
-        };
-
+    fn enum_for_i64(&self, enum_id: CEnumId, value: i64) -> Box<Expr> {
         if let Some(enum_constant_id) = self.enum_variant_for_i64(enum_id, value) {
             return self.enum_constant_expr(enum_constant_id);
         }
@@ -164,8 +157,7 @@ impl<'c> Translation<'c> {
             _ => signed_int_expr(value),
         };
 
-        let target_ty = self.convert_type(enum_type_id).unwrap();
-        mk().cast_expr(value, target_ty)
+        self.enum_constructor_expr(enum_id, value)
     }
 
     /// Returns the id of the variant of `enum_id` whose value matches `value`, if any.
@@ -190,6 +182,17 @@ impl<'c> Translation<'c> {
         let name = self.renamer.borrow().get(&enum_constant_id).unwrap();
         self.add_import(enum_constant_id, &name);
         mk().ident_expr(name)
+    }
+
+    fn enum_constructor_expr(&self, enum_id: CEnumId, value: Box<Expr>) -> Box<Expr> {
+        let enum_name = self
+            .type_converter
+            .borrow()
+            .resolve_decl_name(enum_id)
+            .unwrap();
+        self.add_import(enum_id, &enum_name);
+
+        mk().cast_expr(value, mk().ident_ty(enum_name))
     }
 
     fn is_variant_of_enum(&self, enum_id: CEnumId, enum_constant_id: CEnumConstantId) -> bool {
--- a/c2rust-transpile/src/translator/enums.rs
+++ b/c2rust-transpile/src/translator/enums.rs
@@ -23,10 +23,16 @@ impl<'c> Translation<'c> {
             .borrow()
             .resolve_decl_name(enum_id)
             .expect("Enums should already be renamed");
-        let ty = self.convert_type(integral_type.ctype)?;
-        Ok(ConvertedDecl::Item(
-            mk().span(span).pub_().type_item(enum_name, ty),
-        ))
+        let integral_type_rs = self.convert_type(integral_type.ctype)?;
+        let field = mk().pub_().enum_field(integral_type_rs);
+        let item = mk()
+            .span(span)
+            .call_attr("derive", vec!["Clone", "Copy"])
+            .call_attr("repr", vec!["transparent"])
+            .pub_()
+            .struct_item(enum_name, vec![field], true);
+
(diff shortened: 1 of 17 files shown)


Rua:
It seems that the transpiler is generating working code, but then the refactorer is breaking the imports. I don't know if that's simply a bug in the refactorer, or something else.

Edit: wondering if the refactorer is not handling `impl` blocks properly, to support associated constants?

kkysen (commented):
Btw, #1623 should help with the error messages here, which I saw were atrocious (#1621).

Rua:
The errors from refactor don't seem to have gone away. This PR hasn't changed anything about refactor, so is that a bug in the refactorer?

kkysen (commented):
> The errors from refactor don't seem to have gone away. This PR hasn't changed anything about refactor, so is that a bug in the refactorer?

Oh sorry, I meant #1623 should help make the error messages legible, but not fix them.

This is likely either a bug in the refactorer, or the refactorer is assuming something about what code the transpiler generates for `enum`s and now can't handle a different translation of `enum`s.  If it's the latter, we should probably fix it here.

Rua:
I've not looked at the refactorer code at all yet, so I wouldn't really know where to look for a possible cause. The CI error messages don't give much of a hint.

kkysen:
> I've not looked at the refactorer code at all yet, so I wouldn't really know where to look for a possible cause. The CI error messages don't give much of a hint.

@ahomescu, do you have any idea maybe?

ahomescu:
Could you share the refactorer errors you're getting?

Rua:
The errors are happening in CI. I can't run the tests on NixOS myself.

Rua:
I've removed the second commit from the PR, the one that uses associated constants instead of global ones. That seems to be the one that the refactorer has trouble with.

EDIT: I guess not? Hmm.

ahomescu:
If you run the transpiler with `--reorganize-definitions` (we're renaming that btw, but I don't think it has landed), are you maybe missing the `c2rust::src_loc` attributes on the newtypes?

Rua:
@ahomescu  Doesn't seem so. Enabling it on `tests/unit/enums/src/top_enum.c` produces

```rust
#[derive(Clone, Copy)]
#[repr(transparent)]
#[c2rust::src_loc = "2:1"]
pub struct E(pub ::core::ffi::c_uint);
#[c2rust::src_loc = "4:5"]
pub const B: E = E(1);
#[c2rust::src_loc = "3:5"]
pub const A: E = E(0);
```

Rua:
Finally managing to get all the CI to pass, with a bunch of dependencies to fix the blocking issues.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2625,11 +2625,16 @@ class TranslateASTVisitor final
 
 void TypeEncoder::VisitEnumType(const EnumType *T) {
     auto ed = T->getDecl()->getDefinition();
+    
+    if (!ed) {
```
ahomescu, line 2629: Could you use stacked PRs to make this easier to review? I already looked at these lines in the other PR. Thanks!

ahomescu: On second thought, do we support stacked PRs in our repo? It seems the feature is not public yet.

Rua: I haven't seen that feature anywhere yet.

kkysen (commented):
All of the dependency PRs have been merged now, so could you rebase?

Rua:
I made some additional changes here, that allowed eliminating some unnecessary casts that were generated in previous versions of this PR.

ahomescu:
This is a pretty significant change from the old transpiled code. bindgen makes this configurable https://docs.rs/bindgen/latest/bindgen/enum.EnumVariation.html, I'm wondering if we should do the same. How difficult would that be for this PR?

Rua:
I made some changes to try to support both, at least internally to the transpiler. But because of how snapshot tests are set up, there's no way to ensure both methods work.

randomPoison:
We talked about this change a little bit internally and I wanted to leave my thoughts here for visibility.

I don't think making the enum into a newtype is buying us much in terms of correctness or type safety. In normal Rust code a newtype is useful because it allows us to avoid confusing values of the underlying type when those values mean different things. We can also do encapsulation to enforce invariants about the inner value, exposing only valid operations on the newtype. Neither of those really apply to the transpiled code generated by c2rust, since the generated code is always going to be operating directly on the inner value.

Where I do see value in this change is that it might set us up better for later liftings, i.e. turning the newtype into a proper enum. The big potential advantage is that we could make the variant values into associated constants, which would then give us a similar structure to proper enums:

```rust
struct MyEnum(i32);

impl MyEnum {
    const Foo: Self = Self(0);
    const Bar: Self = Self(1);
}

let enum_val = MyEnum::Bar;
match val {
    MyEnum::Foo => println!("Got a Foo"),
    MyEnum::Bar => println!("Got a Bar"),
    _ => println!("Got something else!"),
}
```

I think that would be a good middle ground, where we preserve the C semantics (the enum is really just a number with some associated constants) but structure things more closely to how the idiomatic Rust equivalent would look.

Looking at the comment history on the PR this might have also been what the PR was originally doing? If the issue there was related to the refactorer not handling impl blocks (which is entirely possible, we don't generate impl blocks anywhere else afaik) then we should fix that.

Also, with regards to making this configurable, I'm not sure what that buys us. If we go down the associated constants route, then I think this is a strictly better translation than the current setup. Do we have any use cases where using a newtype would cause problems (maybe a question for @ahomescu)? If we do add a flag for this, then I think we should also extend the snapshot tests so that we can test both options. That shouldn't be too hard to add, and would be generally useful for testing places where we make the output configurable.

Rua:
> Looking at the comment history on the PR this might have also been what the PR was originally doing?

Yes, but the associated constants part was split out into #1659. Both phases of the change needed separate fixes to the refactorer, and it was easier to see what broke what this way. The two PRs can be merged back, but since they both exist now there's probably not much point?

randomPoison:
Ah cool, I didn't see there was a second PR. I'm fine with doing this as two separate PRs so long as we end up with associated constants 👍

Rua:
So, should `EnumMode` be removed, or?

randomPoison:
Yeah good question. @ahomescu how do you feel about it? I'm inclined to say that we don't need to support both approaches, but maybe there's a use case for the non-newtype approach that makes it worth preserving

ahomescu:
I didn't realize this was waiting for my response. I think we should keep `EnumMode`. The Rust enum newtypes are sometimes clunky to use for some common C patterns like bitflags. For example if you have a `x = Foo::A | Foo::B;` in C where `A` and `B` are enumeration constants, then you need to either implement operations like `BitOr` for the newtype, use an existing crate like `bitflags`, or do something clunky like `Foo(Foo::A.0 | Foo::B.0)`.

ahomescu:
Continuing my previous comment: bindgen has the same issue https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.bitfield_enum and interestingly enough one of the modes they support is "bitfield enum" which is "newtype with bitwise operations implemented".

Rua:
> The Rust enum newtypes are sometimes clunky to use for some common C patterns like bitflags.

@ahomescu Bitflags are a case that is better translated with an integer than with a newtype. But then there will also be uses where a newtype fits better. If there is a command-line option to choose the output, then both options are possible, but they affect *all* enum uses, even the ones where the other representation would be preferred. No matter what you choose, there will be cases where the chosen representation is less ideal.

However, the newtype representation has one very obvious advantage over the integer one: it preserves information about the enum type, like being a real distinct type, and the possibility for associated constants, that is lost with a simple type alias. Specific instances of newtypes can be refactored by the user into integer type aliases in the case of bitflags and the like, but the reverse is not possible. So a user who wants newtypes for some instances and integers for others will opt for the newtype. The integer representation is more lossy.

ahomescu:
Could we implement "newtype with bitwise operations" then? It's a bit hacky but it seems like it covers the most common use cases.

Rua:
@ahomescu that would be best done with a separate support crate, similar to the one for bitfields. Do you agree?

ahomescu:
> a separate support crate, similar to the one for bitfields

With a macro that the transpiler would invoke? Sure, that could work.

c2rust-transpile/src/translator/enums.rs
```diff
@@ -23,10 +23,24 @@ impl<'c> Translation<'c> {
             .borrow()
             .resolve_decl_name(enum_id)
             .expect("Enums should already be renamed");
-        let ty = self.convert_type(integral_type.ctype)?;
-        Ok(ConvertedDecl::Item(
-            mk().span(span).pub_().type_item(enum_name, ty),
-        ))
+        let integral_type_rs = self.convert_type(integral_type.ctype)?;
+        let item = match self.tcfg.enum_mode {
+            EnumMode::NewType => {
```
ahomescu: Oh I thought you were removing `EnumMode` now. I'd be fine with either keeping it, or replacing with "newtype with operations".

tests/unit/enums/src/test_enums.rs
```diff
@@ -35,8 +35,8 @@ const BUFFER_SIZE6: usize = 1;
 
 #[test]
 pub fn test_variants() {
```
ahomescu, line 37: Do we not have any test (unit or snapshot) with a C bitmask enum?

Rua:
@ahomescu I've been looking into how to make the bitwise operators work here, but it doesn't seem so trivial. Bitwise operations on enums are emitted in the AST with casts from and back to the enum type, as demonstrated by the currently produced code:
```c
typedef enum { Foo0, Foo1 } Foo;

void test_enums(void) {
    Foo foo = Foo0;
    Foo bar = Foo1;
    foo |= bar;
}
```
```rust
pub type Foo = ::core::ffi::c_uint;
pub const Foo1: Foo = 1;
pub const Foo0: Foo = 0;
#[no_mangle]
pub unsafe extern "C" fn test_enums() {
    let mut foo: Foo = Foo0;
    let mut bar: Foo = Foo1;
    foo = (foo as ::core::ffi::c_uint | bar as ::core::ffi::c_uint) as Foo;
}
```

The code produced by this PR actually seems like an *improvement*:
```rust
#[derive(Clone, Copy)]
#[repr(transparent)]
pub struct Foo(pub ::core::ffi::c_uint);
pub const Foo1: Foo = Foo(1);
pub const Foo0: Foo = Foo(0);
#[no_mangle]
pub unsafe extern "C" fn rust_test_enums() {
    let mut foo: Foo = Foo0;
    let mut bar: Foo = Foo1;
    foo = Foo(foo.0 | bar.0);
}
```

So it's really not the newtypes that are causing the code to become less clean here, it's the intermediate casts. And those are part of the AST, so c2rust is really just being faithful here. Making use of bitwise operators here would involve bypassing all these casts, but that applies just as much to the old code as the new, and would add a lot of transpiler complexity.

ahomescu:
Okay I guess we can land it as is for now, but I'd really like to eventually hide the internal value (not have it `pub`).

Are you still removing the `Consts` mode?

ahomescu (changes requested):

Rua:
This now depends on #1921 to make tests pass.

There also seems to be an import issue with libgit2 in the testsuite now. I don't know what the cause of that is yet.

ahomescu:
This looks good now, if you resolve the conflicts and it passes the tests then we can land it.

ahomescu (approved):

c2rust-transpile/src/translator/enums.rs
```diff
@@ -152,8 +186,7 @@ impl<'c> Translation<'c> {
             _ => signed_int_expr(value),
```
ahomescu, line 186: Do we get a compilation error here if the value is negative but the underlying type is one of the other unsigned types? E.g. `(enum E)-1` where `E` is a `uint8_t`.

ahomescu (changes requested):

c2rust-transpile/src/translator/enums.rs
Rua: Oh you're right, that needs a cast. But that seems to be a problem already in `master`, going back at least to the beginning of 2026. I'll make a fix in a separate PR.

Rua: #1965

ahomescu (approved):

PR #1622, opened 2026-02-25
Add testing for C decl source range extraction

fw-immunant:
Includes some other minor cleanups made on the way.

### DIFF
--- a/c2rust-ast-exporter/src/clang_ast.rs
+++ b/c2rust-ast-exporter/src/clang_ast.rs
@@ -215,8 +215,8 @@ pub fn process(items: Value) -> error::Result<AstContext> {
         .into_iter()
         .map(|(path, loc)| {
             let path = match path.as_str() {
-                "" => None,
-                "?" => None,
+                "" => None,  // Clang gives an empty path for builtins, for example.
+                "?" => None, // The AST exporter gives "?" when `tryGetRealPathName` fails.
                 path => Some(Path::new(path).to_path_buf()),
             };
             SrcFile {
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -221,6 +221,16 @@ impl TypedAstContext {
             }
         }
 
+        // Ensure that the main file is actually present. This is not the case if it is empty,
+        // in which case the AST exporter's visitor would have never observed it.
+        if !files.iter().any(|f| f.path.as_ref() == Some(&main_file)) {
+            file_map.push(files.len());
+            files.push(SrcFile {
+                path: Some(main_file.clone()),
+                include_loc: None,
+            });
+        }
+
         TypedAstContext {
             main_file,
             files,
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -1484,7 +1484,9 @@ impl<'c> Translation<'c> {
             type_converter.translate_valist = true
         }
 
-        let main_file = ast_context.find_file_id(main_file).unwrap_or(0);
+        let main_file = ast_context
+            .find_file_id(main_file)
+            .expect("could not find FileId for main file");
         let items = indexmap! {main_file => ItemStore::new()};
 
         Translation {


c2rust-ast-exporter/src/lib.rs
```diff
@@ -69,12 +73,14 @@ fn get_ast_cbors(
...
+        let lock = CLANG_MUTEX.lock().unwrap();
         let ptr = ast_exporter(
             args_ptrs.len() as c_int,
             args_ptrs.as_ptr(),
             debug.into(),
             &mut res,
         );
+        drop(lock);
         hashmap = marshal_result(ptr);
         drop_export_result(ptr);
```
kkysen, line 85: Is `drop_export_result` thread-safe?

c2rust-transpile/tests/snapshots.rs
```diff
@@ -138,7 +139,7 @@ fn transpile(platform: Option<&str>, c_path: &Path) {
 
 #[test]
 fn transpile_all() {
```
kkysen:
```suggestion
fn transpile_all_snapshots() {
```

c2rust-ast-exporter/src/lib.rs
fw-immunant: Yeah, it just does C++ deallocation operations (`delete` etc.), which are thread-safe, on fields of the `ExportResult`.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -168,14 +169,52 @@ fn transpile_all() {
...
+        ])
+        .arg(c_path)
+        .status();
+    assert!(status.unwrap().success());
+
+    let (_temp_dir, temp_path) =
+        c2rust_transpile::create_temp_compile_commands(&[c_path.to_owned()]);
+    c2rust_transpile::transpile(
+        TranspilerConfig {
+            emit_c_decl_map: true,
```
kkysen: Can we just enable this for everything in `fn config`?  And then refactor the non-snapshot part of `fn transpile`/`fn transpile_snapshot` into a shared function, because they're basically both the same except for the end that does the snapshot testing.

c2rust-transpile/tests/snapshots.rs
kkysen: Also, could you rebase?  This stuff is gone now.

c2rust-ast-exporter/src/lib.rs
```diff
@@ -48,6 +49,9 @@ pub fn get_untyped_ast(
     clang_ast::process(items).map_err(|e| Error::new(ErrorKind::InvalidData, format!("{}", e)))
 }
 
+/// libClangTooling is not thread-safe, so we must not allow concurrent calls to `ast_exporter`.
+static CLANG_MUTEX: Mutex<()> = Mutex::new(());
```
kkysen, line 53: Could you add this in a `# Safety` comment to `unsafe fn ast_exporter`, too?  Wherever it's declared.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -168,14 +169,52 @@ fn transpile_all() {
...
     });
 }
+
+#[test]
+fn check_c_decl_map() {
+    insta::glob!("c_decls_snapshots/*.c", transpile_with_c_decl_map_snapshot);
+}
+
+fn transpile_with_c_decl_map_snapshot(c_path: &Path) {
+    let status = Command::new("clang")
```
ahomescu: Would the `cc` crate be usable here? It's a bit more idiomatic, but I'm not sure it can be used outside build scripts.

kkysen: Perhaps, but if we want to change that, it should probably go in its own PR.  This is existing copied functionality that we probably don't need to copy (see https://github.com/immunant/c2rust/pull/1622#discussion_r2855207971).

c2rust-transpile/tests/snapshots.rs
fw-immunant: I've factored out the shared code, but setting `emit_c_decls_json` for all snapshots generates a bunch of irrelevant output files, so I think it's better to only set it when we actually want that output.

c2rust-ast-exporter/src/lib.rs
```diff
@@ -89,6 +95,7 @@ mod ffi {
 }
 
 extern "C" {
+    // Safety: Not thread-safe; must not be called multiple times concurrently.
```
kkysen:
```suggestion
    /// # Safety
    ///
    /// Not thread-safe; must not be called multiple times concurrently.
```

c2rust-transpile/tests/snapshots.rs
kkysen: What's wrong with some extra `.gitignore`d output files?  This adds some more testing at what is IMO negligible cost (some extra ignored files).

c2rust-transpile/tests/c_decls_snapshots/nh.c
```diff
@@ -0,0 +1,34 @@
...
+{
+#define FOO 4000+50
+    return FOO;
+}
+// comment before multiple decl
+int a1, a2, a3;
+void forwarddeclbeforeotherdef(int x);
+void funcafterdecl(int n) {}
+/*real defn*/
+void forwarddeclbeforeotherdef(int x) { }
```
kkysen:
```suggestion
another_func(void)
{
    int var = 0;
    return 4 + var;
}
int
func_holding_define(void)
{
#define FOO 4000+50
    return FOO;
}
// comment before multiple decl
int a1, a2, a3;
void forward_decl_before_other_def(int x);
void func_after_decl(int n) {}
/*real defn*/
void forward_decl_before_other_def(int x) { }
```

Was there a reason these names didn't have separators?  It's pretty hard to read without them.

c2rust-transpile/tests/snapshots.rs
fw-immunant: Done.

fw-immunant:
> The tests LGTM now, but I think the tests are failing on the C decl map generation for some of the other files now, so it's good that we caught those.

It looks like it fails for the empty `dummy.c` file because in that case `t.ast_context.get_file_path(t.main_file)` fails. Investigating more presently.

PR #1626, opened 2026-02-27
transpile: add `--postprocess` to run `c2rust-postprocess`

kkysen:
`c2rust-postprocess` is run after transpiling and potentially refactoring (if `--reorganize-definitions` and not `--disable-refactoring`).  Since `c2rust-postprocess` is not a Rust binary, it's not always a sibling to the `c2rust`/`c2rust-transpile` binary.  It's a bash script that runs a Python script with `uv`.  Thus, to find it, we look in `$PATH` first and if it's not there, we assume we're in the `c2rust` repo and `cargo build` has been run and find the relative path of `c2rust-postprocess` (i.e., `target/$profile/../../c2rust-postprocess/c2rust-postprocess`).

Similarly to `--reorganize-definitions`/`c2rust-refactor`, this requires `--emit-build-files`, as we need the `crate_file` (e.g., `src/lib.rs`) to run `c2rust-postprocess`.  `c2rust-postprocess` can be easily set up to run on a single Rust file, though, so we should be able to drop this requirement later.

### DIFF
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -445,8 +445,7 @@ pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]
                 let crate_file = crate_file.as_deref();
                 reorganize_definitions(&tcfg, &build_dir, crate_file)
                     .unwrap_or_else(|e| warn!("Reorganizing definitions failed: {}", e));
-                run_postprocess(&tcfg, &build_dir, crate_file)
-                    .unwrap_or_else(|e| warn!("postprocess failed: {e}"));
+                run_postprocess(&tcfg, &build_dir, crate_file).unwrap_or_else(|e| warn!("{e}"));
                 workspace_members.push(lcmd_name);
             }
         }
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/lib.rs
```diff
@@ -558,6 +568,79 @@ fn reorganize_definitions(
...
+///
+/// This assumes the subcommand executable is either in `$PATH`
+/// or in the same relative directory as it is in the repo
+/// and the current executable is in `target/$profile/`.
+fn invoke_postprocess(crate_file: &Path, build_dir: &Path) -> Result<(), Error> {
+    let subcommand = "c2rust-postprocess";
+    let subcommand_path_buf;
+    let subcommand_path = match which(subcommand) {
+        Ok(_) => Path::new(subcommand),
+        Err(_) => {
```
fw-immunant, line 579: It seems like we could avoid the `which` dep by checking for the relative file first, and otherwise using the command name rather than a path to allow `exec`'s built-in `PATH` lookup to kick in.

c2rust-postprocess/c2rust-postprocess
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-uv run postprocess "$@"
+uv run --project "$(dirname "$0")" postprocess "$@"
```
fw-immunant, line 2: What does this change achieve? I'm not an expert on `uv`.

kkysen: It wouldn't work if ran from different directories before.  This tells `uv` to look up stuff like `pyproject.toml` in that directory.

c2rust-transpile/src/lib.rs
kkysen: I tried that at first, but it makes juggling the args and error handling trickier, which I didn't think was worth it since TOCTOU errors seem very unlikely here (installing/building `c2rust` at the same time as running it).  I could switch back, it just didn't really seem worth the complexity.  Also note that we're still running `exec` with just `c2rust-postprocess`, not the looked up command, as this makes any error messages clearer.

PR #1627, opened 2026-02-28
transpile: Update keywords

cydparser:
+ Add `await` and `gen`
+ Remove symbols that are not keywords
+ Reorder to match the Rust Reference

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,6 +493,7 @@ dependencies = [
  "dtoa",
  "failure",
  "fern",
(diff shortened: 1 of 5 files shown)


kkysen (commented):
This LGTM, but I'd like to add tests.  I can take a look at that first.

cydparser:
Thanks for adding a test. I've determined the issue with `await` and will open a new PR.

cydparser:
See #1636

PR #1628, opened 2026-03-01
transpile: Wrap transmuted string literals in `const` blocks when address is taken

### LINKED ISSUE #1571: C2Rust generates invalid Rust by taking address of a temporary when translating wide string literals (E0745)
C2Rust generates **invalid Rust code** when translating C wide string literals (L"...").
The generated code takes the address of a **temporary value** produced by `core::mem::transmute`, which fails to compile with:

```
error[E0745]: cannot take address of a temporary
```

In C, wide string literals have **static storage duration**, but the generated Rust code does not preserve this property.
**Original C Program**

```c
#include <stdio.h>
#include <wchar.h>

int main() {
    const wchar_t* wstr = L"Wide String";
    printf("Wide string, First char code=%d\n", (int)wstr[0]);
    return 0;
}
```

**C2Rust-Generated Rust**

```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
#![feature(raw_ref_op)]
extern "C" {
}
pub type wchar_t = ::libc::wchar_t;
    let mut wstr: *const wchar_t = &raw const ::core::mem::transmute::<
        [u8; 48],
        [::core::ffi::c_int; 12],
    >(
        *b"W\0\0\0i\0\0\0d\0\0\0e\0\0\0 \0\0\0S\0\0\0t\0\0\0r\0\0\0i\0\0\0n\0\0\0g\0\0\0\0\0\0\0",
    ) as *const wchar_t;
    printf(
        b"Wide string, First char code=%d\n\0" as *const u8
            as *const ::core::ffi::c_char,
        *wstr.offset(0 as ::core::ffi::c_int as isize) as ::core::ffi::c_int,
    );
}
pub fn main() {
}
```

**Observed Behavior**
Compilation fails with:

```
error[E0745]: cannot take address of a temporary
```

because `&raw const transmute(...)` takes the address of a **temporary value**, whose lifetime ends immediately.
**Environment**
- c2rust version: v0.21.0
- rustc version: 1.91.1

Rua:
- Fixes #1571 

This introduces another unstable feature, but it seems like the most straightforward way to solve the issue. Borrowing a `const` value promotes it to static, so that avoids the dangling pointers to temporaries. It could be solved by using a separate `const` variable to avoid the unstable inline `const`, but that results in even messier output code I think.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@wide_strings.c.2021.linux.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@wide_strings.c.2021.linux.snap
@@ -10,7 +10,7 @@ expression: cat tests/snapshots/os-specific/wide_strings.2021.linux.rs
     unused_assignments,
     unused_mut
 )]
-#![feature(raw_ref_op)]
+#![feature(inline_const, raw_ref_op)]
 extern "C" {
     fn wcslen(__s: *const wchar_t) -> ::core::ffi::c_ulong;
 }
@@ -26,6 +26,11 @@ pub static mut static_array_longer: [wchar_t; 3] =
 pub static mut static_array_shorter: [wchar_t; 1] =
     unsafe { ::core::mem::transmute::<[u8; 4], [wchar_t; 1]>(*b"x\0\0\0") };
 #[no_mangle]
+pub static mut static_ptr: *const wchar_t = unsafe {
+    const { ::core::mem::transmute::<[u8; 8], [::core::ffi::c_int; 2]>(*b"x\0\0\0\0\0\0\0") }
+        .as_ptr() as *const wchar_t
+};
+#[no_mangle]
 pub unsafe extern "C" fn func() {
     let mut array: [wchar_t; 2] =
         ::core::mem::transmute::<[u8; 8], [wchar_t; 2]>(*b"x\0\0\0\0\0\0\0");
@@ -33,5 +38,8 @@ pub unsafe extern "C" fn func() {
         ::core::mem::transmute::<[u8; 12], [wchar_t; 3]>(*b"x\0\0\0\0\0\0\0\0\0\0\0");
     let mut array_shorter: [wchar_t; 1] =
         ::core::mem::transmute::<[u8; 4], [wchar_t; 1]>(*b"x\0\0\0");
+    let mut ptr: *const wchar_t =
+        const { ::core::mem::transmute::<[u8; 8], [::core::ffi::c_int; 2]>(*b"x\0\0\0\0\0\0\0") }
+            .as_ptr() as *const wchar_t;
     let mut len: size_t = wcslen(&raw mut array as *mut wchar_t) as size_t;
 }
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -119,16 +119,29 @@ impl<'c> Translation<'c> {
 
                 if ctx.needs_address && element_size == 1 {
                     // Unlike in C, Rust string literals are already references by default.
-                    // So if the address needs to be taken, just make a bare literal.
+                    // So if the address needs to be taken, just make a bare literal and let
+                    // `convert_address_of_common` cast it to the appropriate type.
+                    // Strings with element_size > 1 cannot be cast from a byte literal for
+                    // alignment reasons, and need a transmute.
                     Ok(WithStmts::new_val(val))
                 } else {
                     // std::mem::transmute::<[u8; size], ctype>(*b"xxxx")
                     let array_ty = mk().array_ty(mk().ident_ty("u8"), mk().lit_expr(len as u128));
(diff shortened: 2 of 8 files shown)


Rua:
Converting this to draft because this code has issues with alignment.

kkysen (commented):
> Converting this to draft because this code has issues with alignment.

What was happening?  Do we have tests for wide strings?

Rua:
Casting a pointer from `u8` to a larger type will result in an underaligned pointer. The only correct way to solve this is by converting the value itself, i.e. the current transmute, so that the compiler knows how to align it. But the transmute is problematic because it's a temporary.

Rua:
Figured something out. Wrapping the transmute in a `const { }` block will result in the correct `'static` lifetime when it's borrowed:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=83fb6894dba0cd83af3e13b9f79078d1

Inline `const` is available in the nightly currently used by c2rust, but still unstable. Alternatively, the same effect might be achievable, in a cleaner way, by creating another custom proc macro crate for c2rust that does this internally.

Rua:
I've reimplemented this from scratch, to something that hopefully actually does work. It now depends on #1631, #1632 and #1633.

kkysen (commented):
> Inline `const` is available in the nightly currently used by c2rust, but still unstable. Alternatively, the same effect might be achievable, in a cleaner way, by creating another custom proc macro crate for c2rust that does this internally.

Unstable features that have since been stabilized should be fine to use.  We're trying to update our nightly version, and also upgrade the transpiler to emit edition 2024 code.

kkysen (commented):
Could you rebase now?  All of the prereq PRs should be merged now.  Then I can review.

kkysen:
Lol you did one second later.

c2rust-transpile/tests/snapshots/snapshots__transpile@wide_strings.c.2021.linux.snap
```diff
@@ -27,12 +27,20 @@ pub static mut static_array_longer: [wchar_t; 3] =
 pub static mut static_array_shorter: [wchar_t; 1] =
     unsafe { ::core::mem::transmute::<[u8; 4], [wchar_t; 1]>(*b"x\0\0\0") };
 #[no_mangle]
+pub static mut static_ptr: *const wchar_t = unsafe {
+    const { ::core::mem::transmute::<[u8; 8], [::core::ffi::c_int; 2]>(*b"x\0\0\0\0\0\0\0") }
```
kkysen, line 30: Hmm, it seems like we're not using the portable `override_ty` here (`c_int` instead of `wchar_t`).

Rua: Is that a problem with this PR or more a general observation/note for the future?

kkysen: That's a problem here, but I'm not sure why it's happening.

`c_int` is not always the same as `wchar_t` (https://github.com/search?q=repo%3Arust-lang%2Flibc+%22type+wchar_t+%3D%22&type=code), so on some platforms, this won't compile.

Rua: Looking at the AST, the variable itself has type `const wchar_t*` but its initializer expression is `const int*`? And Clang seems to think the latter can be directly assigned to the former without an implicit cast.

Rua: Ah, it's the array-to-pointer decay. Because it's treated as a cast, it's not propagating the `wchar_t` type but instead casting to it.

Rua:
Going by the problematic code posted in https://github.com/immunant/c2rust/issues/1571#issue-3860921964, it looks like the type propagation issue already existed before this PR. So that would make it a separate issue that needs fixing on its own?

c2rust-transpile/tests/snapshots/snapshots__transpile@wide_strings.c.2021.linux.snap
kkysen: We haven't fully figured this out yet, right?

Rua: #1645 is meant to address that as far as I know. But is it necessary to have a fix before this can be merged? As far as I can tell, it wasn't broken by this PR.

kkysen: Yeah, I think it's fine to fix later/separately.

kkysen (approved):

PR #1629, opened 2026-03-02
transpile: Cast `alloca` expression to void first

### LINKED ISSUE #1572: `__builtin_alloca` translation returns `*mut u8` causing type mismatch for `void*`
When translating `__builtin_alloca`, C2Rust generates code that returns a `*mut u8` via `Vec<u8>::as_mut_ptr()`.
 However, in C, `__builtin_alloca` is commonly used in contexts expecting a `void*`.
 As a result, the generated Rust code fails to compile due to a type mismatch (`*mut u8` vs `*mut c_void`).
**Original C Program**
```
#include <stdio.h>

int main() {
    size_t n3 = 0;
    void* ptr3 = (void*)__builtin_alloca(n3);
    printf("Size=%zu, PtrNonNull=%d\n", 
            n3, ptr3 != NULL);
}
```
**C2Rust-Generated Rust**
```
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
extern "C" {
}
pub type size_t = usize;
pub const NULL: *mut ::core::ffi::c_void = ::core::ptr::null_mut::<
    ::core::ffi::c_void,
>();
    let mut alloca_allocations: Vec<Vec<u8>> = Vec::new();
    let mut n3: size_t = 0 as size_t;
    alloca_allocations.push(::std::vec::from_elem(0, n3 as usize));
    let mut ptr3: *mut ::core::ffi::c_void = alloca_allocations
        .last_mut()
        .unwrap()
        .as_mut_ptr();
    printf(
        b"Size=%zu, PtrNonNull=%d\n\0" as *const u8 as *const ::core::ffi::c_char,
        n3,
        (ptr3 != NULL) as ::core::ffi::c_int,
    );
    return 0;
}
pub fn main() {
}
```
**Observed Behavior**
Compilation fails with:
```
expected `*mut c_void`, found `*mut u8`
```
**Environment**
- c2rust version: v0.21.0
- rustc version: 1.91.1

Rua:
- Fixes #1572.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -388,16 +388,15 @@ impl<'c> Translation<'c> {
                         vec![init_expr],
                     ));
 
-                    // c2rust_alloca_allocations.last_mut().unwrap().as_mut_ptr()
-                    let last_mut_expr = mk().method_call_expr(
-                        mk().ident_expr(alloca_allocations_name),
-                        "last_mut",
-                        vec![],
-                    );
-                    let unwrap_expr = mk().method_call_expr(last_mut_expr, "unwrap", vec![]);
-                    let as_mut_ptr_expr = mk().method_call_expr(unwrap_expr, "as_mut_ptr", vec![]);
+                    // c2rust_alloca_allocations.last_mut().unwrap().as_mut_ptr() as *mut ::core::ffi::c_void
+                    let expr = mk().ident_expr(alloca_allocations_name);
+                    let expr = mk().method_call_expr(expr, "last_mut", vec![]);
+                    let expr = mk().method_call_expr(expr, "unwrap", vec![]);
+                    let expr = mk().method_call_expr(expr, "as_mut_ptr", vec![]);
+                    let pointee_ty = mk().abs_path_ty(vec!["core", "ffi", "c_void"]);
+                    let expr = mk().cast_expr(expr, mk().mutbl().ptr_ty(pointee_ty));
 
-                    Ok(WithStmts::new(vec![push_stmt], as_mut_ptr_expr))
+                    Ok(WithStmts::new(vec![push_stmt], expr))
                 })
             }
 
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/builtins.rs
```diff
@@ -388,16 +388,15 @@ impl<'c> Translation<'c> {
...
-                    let as_mut_ptr_expr = mk().method_call_expr(unwrap_expr, "as_mut_ptr", vec![]);
-
-                    Ok(WithStmts::new(vec![push_stmt], as_mut_ptr_expr))
+                    // c2rust_alloca_allocations.last_mut().unwrap().as_mut_ptr() as *mut ::core::ffi::c_void
+                    let mut expr = mk().ident_expr(alloca_allocations_name);
+                    expr = mk().method_call_expr(expr, "last_mut", vec![]);
+                    expr = mk().method_call_expr(expr, "unwrap", vec![]);
+                    expr = mk().method_call_expr(expr, "as_mut_ptr", vec![]);
+                    let pointee_ty = mk().abs_path_ty(vec!["core", "ffi", "c_void"]);
+                    expr = mk().cast_expr(expr, mk().mutbl().ptr_ty(pointee_ty));
```
kkysen:
```suggestion
                    let expr = mk().ident_expr(alloca_allocations_name);
                    let expr = mk().method_call_expr(expr, "last_mut", vec![]);
                    let expr = mk().method_call_expr(expr, "unwrap", vec![]);
                    let expr = mk().method_call_expr(expr, "as_mut_ptr", vec![]);
                    let pointee_ty = mk().abs_path_ty(vec!["core", "ffi", "c_void"]);
                    let expr = mk().cast_expr(expr, mk().mutbl().ptr_ty(pointee_ty));
```

Rua:
Do you prefer multiple variables over a single mutable one? Why's that? (Just curious)

kkysen:
> Do you prefer multiple variables over a single mutable one? Why's that? (Just curious)

In Rust, generally, immutable variables are preferable.  If it doesn't need to be mutable, then it shouldn't.  And it doesn't here, it just needs a new binding, which Rust thankfully makes simple by allowing shadowing.  When it's immutable, when reading the code, you don't have to consider mutations, which makes it easier to understand.

Issue #1582, opened 2026-02-03
File-scope compound literal translated incorrectly, causing segfault on write


rikki322:
When translating a **C99 file-scope compound literal** used to initialize a global pointer, c2rust generates Rust code where the pointer ends up referencing temporary and/or read-only data (via an array literal and `.as_ptr()` cast to `*mut`). Writing through this pointer causes a runtime crash.
**Original C Program**
```
#include <stdio.h>

/* file-scope compound literal: static storage duration (C99) */
static int *p = (int[4]){10, 20, 30, 40};

int main(void) {
    p[1] = 99;
    printf("%d\n", p[1]);
    return 0;
}
```
**C2Rust-Generated Rust**
```
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
extern "C" {
    fn printf(__format: *const ::core::ffi::c_char, ...) -> ::core::ffi::c_int;
}
static mut p: *mut ::core::ffi::c_int = [10, 20, 30, 40].as_ptr()
    as *mut ::core::ffi::c_int;
unsafe fn main_0() -> ::core::ffi::c_int {
    *p.offset(1 as ::core::ffi::c_int as isize) = 99 as ::core::ffi::c_int;
    printf(
        b"%d\n\0" as *const u8 as *const ::core::ffi::c_char,
        *p.offset(1 as ::core::ffi::c_int as isize),
    );
    return 0 as ::core::ffi::c_int;
}
pub fn main() {
    unsafe { ::std::process::exit(main_0() as i32) }
}
```
**Observed result:**
`Segmentation fault (core dumped)`
**Environment**
- c2rust version: v0.21.0
- rustc version: `nightly-2022-08-08`

Rua:
This seems to be a holdover from #1407, which fixed the issue for local variables but not statics. I initially thought that the issue wouldn't apply for statics, but it seems that it does.

The problem is that c2rust does not currently support the solution that was used there, because generating additional statements in a static leads to a "Expected no side-effects in static initializer" panic. So that panic needs to be resolved first before this can be fixed.

PR #1630, opened 2026-03-02
transpile: Redefine and document meaning of `is_static` and `is_const`

Rua:
`is_static` and `is_const` were used somewhat inconsistently, often in situations where code was simply testing for a const context, regardless of what kind. This PR changes the meanings a bit:
- `is_const` is now for any const context, regardless of source.
- `is_static` is now for static variables, even if their initialisers are not going to run in a const context but deferred.
- Testing for macros specifically is done with `expanding_macro`.

`is_static` is currently unused, but looks like it will be necessary to properly fix issue #1582. It may also be needed in the future to test specifically for static const contexts, which appear to allow some things not allowed in other const contexts.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -119,15 +119,24 @@ pub enum ReplaceMode {
 #[derive(Copy, Clone, Debug)]
 pub struct ExprContext {
     used: bool,
-    is_static: bool,
+
+    /// In a Rust const context, for example in a static initializer or constant-like macro
+    /// translation.
     is_const: bool,
+
+    /// Evaluating a C global/static variable.
+    /// This is usually in a const context, but doesn't have to be, for example with initializers
+    /// that are executed by the `c2rust_run_static_initializers` function.
+    #[allow(dead_code)]
+    is_static: bool,
+
     decay_ref: DecayRef,
     is_bitfield_write: bool,
 
-    // We will be referring to the expression by address. In this context we
-    // can't index arrays because they may legally go out of bounds. We also
-    // need to explicitly cast function references to fn() so we get their
-    // address in function pointer literals.
+    /// We will be referring to the expression by address. In this context we
+    /// can't index arrays because they may legally go out of bounds. We also
+    /// need to explicitly cast function references to fn() so we get their
+    /// address in function pointer literals.
     needs_address: bool,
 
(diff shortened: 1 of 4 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -4803,9 +4808,9 @@ impl<'c> Translation<'c> {
     /// Produce zero-initializers for structs/unions/enums, looking them up when possible.
     fn zero_initializer(
         &self,
+        ctx: ExprContext,
```
fw-immunant, line 4825: I thought that this change might mean we need to include the context (or relevant portions) as part of the key for the zero initializer cache, `Translation::zero_inits`, but it turns out that that cache is keyed by the `CDeclId`, and if each `CDeclId` is only translated in one context, this shouldn't be necessary.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -119,15 +119,23 @@ pub enum ReplaceMode {
 #[derive(Copy, Clone, Debug)]
 pub struct ExprContext {
     used: bool,
-    is_static: bool,
+
+    /// In a Rust const context, for example in a static initializer or macro translation.
```
fw-immunant: Specifically a constant-like macro translation; translations of function-like macros (when enabled) aren't const contexts.

docs/source_walkthrough.md
```diff
@@ -208,13 +208,15 @@ is going to be emitted or discarded.
...
-is being used in the initializer for a static variable.
-Rust has many extra restrictions on the expressions
-that can be used for a static initializer.
-In some cases, we can still generate valid code
-at the cost of readability.
-This fallback is enabled by this attribute.
+The `is_const` attribute indicates that an expression will be inside a Rust
+`const` context. Rust has many extra restrictions on the expressions that can
+be used inside a `const` context. In some cases, we can still generate valid
+code at the cost of readability. This fallback is enabled by this attribute.
```
kkysen: The previous paragraph used semantic line breaks (e.g., only using line breaks after punctuation and other semantic breaks) so that modifying parts of a phrase won't trigger a cascade of reformatting line breaks and create diffs unrelated to the actual change.

Rua: How is this formatting now?

kkysen: LGTM!

PR #1633, opened 2026-03-03
transpile: Move some translation from `convert_address_of_common` to `convert_literal`

Rua:
Rather than `convert_address_of_common` completely discarding the previously translated `val` for string literals, it now does part of the correct translation in `convert_literal` so that no discarding is needed. This will probably be needed to get #1628 working, since `add_feature` (for const blocks) will have an effect even if code is discarded.

Also, `as_ptr` is now used, to be consistent with other kinds of array.

### DIFF
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -131,7 +131,7 @@ impl<'c> Translation<'c> {
         ) = (arg_expr_kind, arg_is_macro)
         {
             if is_array_decay {
-                ref_cast_pointee_ty = Some(mk().ident_ty("u8"));
+                val = val.map(|val| mk().method_call_expr(val, "as_ptr", vec![]));
             } else {
                 let size = self.ast_context.array_len(literal_cty.ctype) * element_size as usize;
                 ref_cast_pointee_ty =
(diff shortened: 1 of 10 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -611,6 +611,14 @@ impl TypedAstContext {
         }
     }
 
+    /// Returns the length of an array type, or panics.
+    pub fn array_len(&self, typ: CTypeId) -> usize {
+        match self.resolve_type(typ).kind {
+            CTypeKind::ConstantArray(_, len) => len,
+            _ => panic!("CTypeId {typ:?} does not point to a ConstantArray"),
```
kkysen:
```suggestion
            kind => panic!("CTypeId {typ:?} is {kind:?}, not a ConstantArray"),
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4083,18 +4083,20 @@ impl<'c> Translation<'c> {
             Paren(_, val) => self.convert_expr(ctx, val, override_ty),
 
             CompoundLiteral(qty, val) => {
-                let val = self.convert_expr(ctx, val, override_ty)?;
```
kkysen, line 4086: Ah, without this makes more sense now.

PR #1636, opened 2026-03-04
transpile: Dedup keywords

cydparser:
This change extracts keywords into `c2rust-transpile::RUST_KEYWORDS` and reduces the amount of renaming in a few places. If I have understood the code correctly, `RESERVED_NAMES` should have only been used when renaming types, but it was also being used when renaming field names and module names.

### DIFF
--- a/c2rust-transpile/src/convert_type.rs
+++ b/c2rust-transpile/src/convert_type.rs
@@ -29,118 +29,12 @@ pub struct TypeConverter {
     extern_crates: CrateSet,
 }
 
-pub const RESERVED_NAMES: [&str; 100] = [
-    // Keywords currently in use
-    "as",
-    "async",
-    "await",
-    "break",
-    "const",
-    "continue",
-    "crate",
-    "dyn",
-    "else",
-    "enum",
-    "extern",
-    "false",
-    "fn",
-    "for",
-    "if",
-    "impl",
-    "in",
-    "let",
-    "loop",
-    "match",
-    "mod",
-    "move",
-    "mut",
-    "pub",
-    "ref",
-    "return",
-    "self",
-    "Self",
-    "static",
-    "struct",
-    "super",
-    "trait",
-    "true",
-    "type",
-    "unsafe",
-    "use",
-    "where",
-    "while",
-    // Keywords reserved for future use
-    "abstract",
-    "become",
-    "box",
-    "do",
-    "final",
-    "gen",
-    "macro",
-    "override",
-    "priv",
-    "try",
-    "typeof",
-    "unsized",
-    "virtual",
-    "yield",
-    // Types exported in prelude
-    "Copy",
-    "Send",
-    "Sized",
-    "Sync",
-    "Drop",
-    "Fn",
-    "FnMut",
-    "FnOnce",
-    "Box",
-    "ToOwned",
-    "Clone",
-    "PartialEq",
-    "PartialOrd",
-    "Eq",
-    "Ord",
-    "AsRef",
-    "AsMut",
-    "Into",
-    "From",
-    "Default",
-    "Iterator",
-    "Extend",
-    "IntoIterator",
-    "DoubleEndedIterator",
-    "ExactSizeIterator",
-    "Option",
-    "Result",
-    "SliceConcatExt",
-    "String",
-    "ToString",
-    "Vec",
-    "bool",
-    "char",
-    "f32",
-    "f64",
-    "i8",
-    "i16",
-    "i32",
-    "i64",
-    "i128",
-    "isize",
-    "u8",
-    "u16",
-    "u32",
-    "u64",
-    "u128",
-    "usize",
-    "str",
-];
-
 impl TypeConverter {
     pub fn new(tcfg: &TranspilerConfig) -> TypeConverter {
         TypeConverter {
             edition: tcfg.edition,
             translate_valist: tcfg.translate_valist,
-            renamer: Renamer::new(&RESERVED_NAMES),
+            renamer: Renamer::type_namespace(),
             fields: HashMap::new(),
             suffix_names: HashMap::new(),
             features: HashSet::new(),
@@ -197,7 +91,7 @@ impl TypeConverter {
 
         self.fields
             .entry(record_id)
-            .or_insert_with(|| Renamer::new(&RESERVED_NAMES))
+            .or_insert_with(|| Renamer::keywords())
             .insert(FieldKey::Field(field_id), name)
             .expect("Field already declared")
     }
@@ -206,7 +100,7 @@ impl TypeConverter {
         let field = self
             .fields
             .entry(record_id)
-            .or_insert_with(|| Renamer::new(&RESERVED_NAMES));
+            .or_insert_with(|| Renamer::keywords());
         let key = FieldKey::Padding(padding_idx);
         if let Some(name) = field.get(&key) {
             name
--- a/c2rust-transpile/src/lib.rs
+++ b/c2rust-transpile/src/lib.rs
@@ -20,6 +20,7 @@ use std::process::Command;
 use std::{env, io};
 
 use crate::compile_cmds::CompileCmd;
+use crate::renamer::RUST_KEYWORDS;
 use c2rust_rust_tools::{rustfmt, RustEdition};
 use failure::{format_err, Error};
 use itertools::Itertools;
@@ -36,7 +37,6 @@ use c2rust_ast_exporter as ast_exporter;
 
 use crate::build_files::{emit_build_files, get_build_dir, CrateConfig};
 use crate::compile_cmds::get_compile_commands;
-use crate::convert_type::RESERVED_NAMES;
 pub use crate::translator::ReplaceMode;
 use std::prelude::v1::Vec;
 
@@ -248,7 +248,7 @@ fn str_to_ident_checked(s: &str, check_reserved: bool) -> String {
     let s = str_to_ident(s);
 
     // make sure the name does not clash with keywords
-    if check_reserved && RESERVED_NAMES.contains(&s.as_str()) {
+    if check_reserved && RUST_KEYWORDS.contains(&s.as_str()) {
         format!("r#{}", s)
     } else {
         s
--- a/c2rust-transpile/src/renamer.rs
+++ b/c2rust-transpile/src/renamer.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::hash::Hash;
-use std::iter::FromIterator;
 
 struct Scope<T> {
     name_map: HashMap<T, String>,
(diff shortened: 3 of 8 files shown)


c2rust-transpile/src/lib.rs
```diff
@@ -236,14 +235,73 @@ fn str_to_ident(s: &str) -> String {
     s.chars().map(char_to_ident).collect()
 }
 
+// Keywords obtained from https://doc.rust-lang.org/reference/keywords.html
+#[rustfmt::skip] // Preserve one keyword per line.
+pub const RUST_KEYWORDS: [&str; 52] = [
```
kkysen: If you use `&[&str]`, you don't have to specify the length.

c2rust-transpile/src/lib.rs
```diff
@@ -236,14 +235,73 @@ fn str_to_ident(s: &str) -> String {
     s.chars().map(char_to_ident).collect()
 }
 
+// Keywords obtained from https://doc.rust-lang.org/reference/keywords.html
```
kkysen: Could we have a list of strict, reserved, and weak keywords matching the reference?  And could you mark the ones that have been added in later editions with comments (e.g., `"gen", // added in edition 2024`)?

kkysen: The weak keywords are also contextual, so they may need to be added in certain contexts.

kkysen: Actually, I think they're valid in all places where identifiers are valid.

c2rust-transpile/src/convert_type.rs
```diff
@@ -25,62 +25,7 @@ pub struct TypeConverter {
...
-    "macro",
-    "override",
-    "priv",
-    "try",
-    "typeof",
-    "unsized",
-    "virtual",
-    "yield",
-    // Types exported in prelude
+const RENAME_PRELUDE_TYPES: [&str; 48] = [
```
kkysen: Not sure we need to call this `RENAME_`.  Also, there are prelude types missing from this too.  It also includes `trait`s, so maybe `PRELUDE_TYPE_NAMESPACE` is more accurate?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -1509,19 +1509,22 @@ impl<'c> Translation<'c> {
...
+                    [
+                        // Prevent use for other reasons
+                        "main",
+                        // Prelude names
+                        "drop",
+                        "Some",
+                        "None",
+                        "Ok",
+                        "Err",
+                    ]
```
kkysen: Besides `main`, should this go into a `PRELUDE_VALUE_NAMESPACE`?

c2rust-transpile/src/convert_type.rs
```diff
@@ -1,7 +1,7 @@
 use crate::c_ast::CDeclId;
-use crate::c_ast::*;
 use crate::diagnostics::TranslationResult;
 use crate::renamer::*;
+use crate::{c_ast::*, RUST_KEYWORDS};
```
kkysen: Could you expand the glob?  It was there already, but they're annoying.

c2rust-transpile/src/convert_type.rs
```diff
@@ -25,62 +25,7 @@ pub struct TypeConverter {
     extern_crates: CrateSet,
 }
 
-pub const RESERVED_NAMES: [&str; 100] = [
```
kkysen, line 32: Maybe this and all of the prelude names can go into `renamer.rs` instead?

c2rust-transpile/src/renamer.rs
```diff
@@ -42,16 +42,29 @@ pub struct Renamer<T> {
...
     /// reserved names will exclude those names from being chosen as the mangled names from
     /// the insert method.
     pub fn new(reserved_names: &[&str]) -> Self {
-        let set: HashSet<String> = HashSet::from_iter(reserved_names.iter().map(|&x| x.to_owned()));
-        Renamer {
-            scopes: vec![Scope::new_with_reserved(set)],
-            next_fresh: 0,
-        }
+        Renamer::from_iter(reserved_names.into_iter().map(|s| *s))
     }
```
kkysen:
```suggestion
    /// Creates a new renaming environment with a single, empty scope.
    /// The given sets of reserved names will exclude those names
    /// from being chosen as the mangled names from the insert method.
    pub fn new(reserved_names: &[&[&str]]) -> Self {
        let set = reserved_names
            .iter()
            .flat_map(|&names| names)
            .map(|&s| s.to_owned())
            .collect::<HashSet<_>>();
        Self {
            scopes: vec![Scope::new_with_reserved(set)],
            next_fresh: 0,
        }
    }
```

I think this should be sufficient.  For all of the call sites, we give it 1 or multiple sets of reserved names.

kkysen: Then we could have, in `impl Renamer`, `fn keywords() -> Self`, `fn type_namespace() -> Self`, and `fn value_namespace() -> Self`.  This makes it clearer at the call sites what we intend to use.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -190,12 +189,9 @@ fn generate_keywords_test() {
         "break", "const", "continue", "else", "enum", "extern", "for", "if", "return", "static",
         "struct", "while", "do", "typeof", "char",
     ];
-    // These don't work yet.  We need to fix these.
-    let broken_rust_keywords = ["await"];
-    let mut c_code = RESERVED_NAMES
+    let mut c_code = RUST_KEYWORDS
```
kkysen, line 283: This was updated in the second commit even though it was renamed in the first commit?  Could you just make sure each commit compiles and passes the tests?  Thanks!  It's fine, and good, to split separate steps/changes into separate commits, but each should compile and pass tests on its own.

c2rust-transpile/src/convert_type.rs
kkysen: Also, could you leave the edition comments for these, too?  We're trying to support edition 2024, and may have to support edition 2021 at the same time.  It might be fine to always rename edition 2024 names even when targeting older editions (if not, we'd want separate arrays for each edition), but it's still good to have them documented.

c2rust-transpile/src/lib.rs
cydparser: I've added comments for the editions.

> Could we have a list of strict, reserved, and weak keywords matching the reference?

What would be the benefit of that? Do reserved keywords ever need to be treated differently than strict ones?

c2rust-transpile/src/convert_type.rs
cydparser: I like `PRELUDE_TYPE_NAMESPACE`.

I'll leave adding comments about editions to whoever does the future work for supporting multiple editions.

c2rust-transpile/src/translator/mod.rs
cydparser: `PRELUDE_VALUE_NAMESPACE` only contains types and traits, whereas this list contains functions and data constructors.

c2rust-transpile/src/convert_type.rs
cydparser: Sure, if the line still needs to be modified after I make your suggested changes.

c2rust-transpile/src/convert_type.rs
cydparser: I like this idea combined with your suggestion to add Renamer constructors.

c2rust-transpile/src/renamer.rs
cydparser: Done

c2rust-transpile/tests/snapshots.rs
cydparser: I don't have a good reason for keeping the test changes in a separate commit.  I'll squash the commits into a single one.

c2rust-transpile/src/translator/mod.rs
kkysen: Types and traits go in the type namespace.  Functions, variables, consts, statics, enum variants, etc. (like all of these) go in the value namespace.

cydparser: Ah, I didn't process the _VALUE_ bit. Placing the values in an const is fine.

c2rust-transpile/src/lib.rs
```diff
@@ -247,7 +246,7 @@ fn str_to_ident_checked(s: &str, check_reserved: bool) -> String {
     let s = str_to_ident(s);
 
     // make sure the name does not clash with keywords
-    if check_reserved && RESERVED_NAMES.contains(&s.as_str()) {
+    if check_reserved && renamer::RUST_KEYWORDS.contains(&s.as_str()) {
```
kkysen: You can `use crate::renamer::RUST_KEYWORDS;`.

c2rust-transpile/src/renamer.rs
```diff
@@ -37,6 +37,128 @@ impl<T: Clone + Eq + Hash> Scope<T> {
...
+    "struct",
+    "super",
+    "trait",
+    "true",
+    "type",
+    "unsafe",
+    "use",
+    "where",
+    "while",
+    // Resserved keywords
```
kkysen:
```suggestion
    // Reserved keywords
```

c2rust-transpile/src/renamer.rs
```diff
@@ -46,14 +168,31 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
...
+    }
+
+    pub fn type_namespace() -> Self {
+        Renamer::new(&[RUST_KEYWORDS, PRELUDE_TYPE_NAMESPACE])
+    }
+
+    pub fn value_namespace() -> Self {
+        Renamer::new(&[RUST_KEYWORDS, PRELUDE_VALUE_NAMESPACE])
+    }
+
```
kkysen, line 190:
```suggestion

    pub fn global_value_namespace() -> Self {
        Renamer::new(&[RUST_KEYWORDS, PRELUDE_VALUE_NAMESPACE, &["main"]])
    }
```

c2rust-transpile/src/renamer.rs
```diff
@@ -37,6 +37,128 @@ impl<T: Clone + Eq + Hash> Scope<T> {
...
+    "u64",
+    "u128",
+    "usize",
+    "str",
+];
+
+#[rustfmt::skip] // Preserve one symbol per line.
+const PRELUDE_VALUE_NAMESPACE: &[&str] = &[
+    // Prevent use for other reasons
+    "main",
```
kkysen: `main` should be fine in other contexts.  Can you do this?  https://github.com/immunant/c2rust/pull/1636/changes#r2936319565

c2rust-transpile/src/lib.rs
kkysen: > What would be the benefit of that? Do reserved keywords ever need to be treated differently than strict ones?

Hmm, I guess you're right, there's not really a benefit.  You can leave it as is.

cydparser:
@kkysen, thanks for the reviews. I think all requested changes have been made.

c2rust-transpile/src/lib.rs
```diff
@@ -248,7 +247,7 @@ fn str_to_ident_checked(s: &str, check_reserved: bool) -> String {
     let s = str_to_ident(s);
 
     // make sure the name does not clash with keywords
-    if check_reserved && RESERVED_NAMES.contains(&s.as_str()) {
+    if check_reserved && crate::renamer::RUST_KEYWORDS.contains(&s.as_str()) {
```
kkysen:
```suggestion
    if check_reserved && RUST_KEYWORDS.contains(&s.as_str()) {
```

Can you add `use crate::renamer::RUST_KEYWORDS;`?

c2rust-transpile/src/renamer.rs
```diff
@@ -46,14 +165,35 @@ impl<T: Clone + Eq + Hash> Renamer<T> {
...
     /// the insert method.
-    pub fn new(reserved_names: &[&str]) -> Self {
-        let set: HashSet<String> = HashSet::from_iter(reserved_names.iter().map(|&x| x.to_owned()));
+    pub fn new(reserved_names: &[&[&str]]) -> Self {
+        let set: HashSet<String> = HashSet::from_iter(
+            reserved_names
+                .iter()
+                .flat_map(|&names| names)
+                .map(|&s| s.to_owned()),
+        );
```
kkysen:
```suggestion
        let set = reserved_names
            .iter()
            .flat_map(|&names| names)
            .map(|&s| s.to_owned())
            .collect::<HashSet<_>>();
```

kkysen (approved):
I fixed the couple of little remaining issues myself.  Merging now.

PR #1637, opened 2026-03-04
postprocess: add `--enable-llm`/`$C2RUST_POSTPROCESS_ENABLE_LLM` to make LLM usage explicitly opt-in

kkysen:
We had discussed adding a `$C2RUST_POSTPROCESS_LLM_MODEL` environment variable to make LLM usage explicitly opt-in, but having it be the model argument meant that we couldn't set a default for it, which is convenient, so I added a boolean `--enable-llm`/`$C2RUST_POSTPROCESS_ENABLE_LLM` arg instead.  I also renamed `--model-id` to `--llm-model` to be more explicit about LLM usage there as well.  If specifying the specific model is important to make explicit, though, I could switch to that version instead, though.

### DIFF
--- a/c2rust-postprocess/postprocess/__init__.py
+++ b/c2rust-postprocess/postprocess/__init__.py
@@ -4,7 +4,8 @@ c2rust-postprocess: Transfer comments from C functions to Rust functions using L
 
 import argparse
 import logging
-from argparse import BooleanOptionalAction
+import os
+from argparse import ArgumentTypeError, BooleanOptionalAction
 from collections.abc import Sequence
 from pathlib import Path
 
@@ -20,6 +21,20 @@ from postprocess.transforms import (
     CommentTransfer,
 )
 from postprocess.utils import existing_file
+
+
+def parse_booly(arg: str) -> bool:
+    """
+    Parse a "booly"/bool-like value, meant for CLI args and environment variables.
+    """
+
+    match arg.lower():
+        case "true" | "1" | "yes" | "y":
+            return True
+        case "false" | "0" | "no" | "n" | "":
+            return False
+        case _:
+            raise ArgumentTypeError(f"booly value expected, got '{arg}'")
 
 
 def build_arg_parser() -> argparse.ArgumentParser:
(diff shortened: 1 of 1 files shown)


fw-immunant:
I would vote against using an environment variable unless absolutely required.

kkysen (commented):
> I would vote against using an environment variable unless absolutely required.

Wait, how come?  I thought we agreed on doing that.

fw-immunant:
> > I would vote against using an environment variable unless absolutely required.
> 
> Wait, how come? I thought we agreed on doing that.

It was mentioned in the meeting discussion that we could have an environment variable, but it wasn't clear to me whether it was actually required by anything. If it isn't, I'd rather not use the environment, because it tends to get messy--it's easy to forget modifications to it in a shell, it can be inherited across process spawning without consideration, and generally acts more like an imperative pile of state that must be re-parsed by every interested program rather than matching the simplicity of passing precise inputs to a function. Dropping the environment variable also lets us avoid the vagueness of "what counts as yes" that `parse_booly` has to clarify.

thedataking:
This is not what I had in mind (but PR is easy to fix): 

Users that run `postprocess` directly should know what it does. Why? 
- It will barf if `GEMINI_API_KEY` is not set OR the user has explicitly asked for a specific remote model on the CLI.
- The postprocess README is super explicit about what this tool does: **LLM-based postprocessing of c2rust transpiler output**. Caveat emptor.

Users that run `c2rust transpile` could be surprised that their (potentially confidential/proprietary) is sent to a remote API for LLM processing. We want to guard against that surprise.

So please don't change `postprocess` by adding new flags or required env variables as this PR does.

Please, instead focus on making sure that when running `c2rust transpile` with post-processing enabled, the user accepts that we send their code to an LLM hosted remotely. AFAIK, we agreed that postprocessing is opt-in. 

Our options are therefore:

- we could require that the user explicitly specify the model to use, if they pass "gemini-3.0-pro-preview" or "gpt-5-2025-08-07", it is reasonable to assume they know what they're doing. However, I think it is beneficial to default the model to something we know works well and is cost-efficient for the tasks we run, so I don't love that option.

- we could require an extra flag `--allow-code-to-be-sent-to-remote-llm-for-postprocessing`, `--enable-remote-llm`, etc.

- we could allow an env variable too. Since Frances has misgivings and it can be ambiguous what values gives consent, I am fine not using env variables. My thinking was that since the API key must be in the environment anyway, but Frances thinks differently so let's follow his intuition.

- we could require that the user is explicit about what API key to use (e.g. `OPENAI_API_KEY`, `GEMINI_API_KEY`) but we should of course not require the user to specify the key value on the command line lest it inadvertently gets stored in the command history. 

Optional: we could mention or quickly describe the LLM-based postprocessor in the main README.md out of an over-abundance of caution.

thedataking:
From discussion at meeting today:

Per: could take inspiration from codex, gemini, etc. - decided we don't need separate flag 👎 
Frances votes for: `--remote-llm-postprocess` - decided that this is a fine solution! 👍

kkysen:
* Superseded by #1666.

PR #1640, opened 2026-03-05
transpile: replace `--reorganize-definitions`/`--disable-refactoring` with `--emit-refactor-annotations`/`--refactor`

kkysen:
Previously, `--reorganize-definitions` would emit annotations and run the refactor, with `--disable-refactoring` needed to only emit annotations. This simplifies things to additive flags.
`--emit-refactor-annotations` only emits the annotations. `--refactor` implies `--emit-refactor-annotations` and runs the default refactor transforms.


c2rust-transpile/src/build_files/mod.rs
```diff
@@ -264,7 +264,7 @@ fn emit_lib_rs(
     let rs_xcheck_backend = tcfg.cross_check_backend.replace('-', "_");
     let json = json!({
         "lib_rs_file": file_name,
-        "reorganize_definitions": tcfg.reorganize_definitions,
+        "reorganize_definitions": tcfg.emit_refactor_annotations,
```
kkysen: @ahomescu, is this the correct change here?  I'm not familiar with how cross-checks are supposed to work.

fw-immunant (changes requested):
`--reorganize-definitions` does not just enable annotations, it also changes the structure of the emitted code to include modules and cross-module imports. So while we should probably clarify the name, the new name shouldn't suggest that the flag is merely additive.

c2rust-transpile/src/build_files/mod.rs
ahomescu: Actually I think this might not be needed anymore. We used to have the following in `lib.rs.hbs`:
```
-{{#if reorganize_definitions~}}
-    #![feature(custom_attribute)]
-{{/if}}
```
but that's gone now so you could probably remove this line.

ahomescu (approved):

c2rust-transpile/src/build_files/mod.rs
kkysen: Removed in #1665, along with `translate_valist`, which is unused, too, now.

PR #1642, opened 2026-03-06
transpile: switch `insta::glob!`s to separate tests

kkysen:
This replaces `insta::glob!` usage in `c2rust-transpile/tests/snapshots.rs` with separate, dedicated `#[test] fn`s.

This is what we do for `c2rust-refactor/tests/snapshots.rs`, too.  Now we can easily run an individual test, run all of the tests in parallel much faster, and not stop at the first error if one fails.  This is really useful as I upgrade to edition 2024 (PR soon).

The `*-specific` directories are kept, although we don't need them anymore.  We also delete the `dummy` files, as there's no more issue with `insta::glob!` using a different prefix when there's only one matching file in a directory.

Also, the `c_decls_snapshots` were using absolute paths in their debug expressions, so the snapshot would be different on different machines.  As I switched to `#[test] fn`s, this uses relative paths by default and thus is automatically fixed as well.

### DIFF
--- a/c2rust-transpile/tests/snapshots.rs
+++ b/c2rust-transpile/tests/snapshots.rs
@@ -419,11 +419,6 @@ fn test_varargs() {
         .run();
 }
 
-#[test]
-fn check_c_decl_map() {
-    insta::glob!("c_decls_snapshots/*.c", transpile_with_c_decl_map_snapshot);
-}
-
 fn transpile_with_c_decl_map_snapshot(c_path: &Path) {
     compile_and_transpile_file(c_path, config());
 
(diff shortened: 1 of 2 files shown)


c2rust-transpile/tests/snapshots.rs
```diff
@@ -434,3 +429,9 @@ fn transpile_with_c_decl_map_snapshot(c_path: &Path) {
         serde_json::from_str(&fs::read_to_string(&c_decls_path).unwrap()).unwrap();
     insta::assert_json_snapshot!(snapshot_name, json, &debug_expr);
 }
+
+#[test]
+fn test_nh() {
```
fw-immunant: It would be clearer for this test name to mention that it's testing c-decls extraction.

fw-immunant (approved):

c2rust-transpile/tests/snapshots.rs
kkysen: Renamed to `fn test_c_decls_nh`.

PR #1643, opened 2026-03-06
transpile: upgrade most things for edition 2024

### LINKED ISSUE #1298: `#![feature(stdsimd)]` was split up and removed in 1.78, so stop emitting it when we upgrade past 1.65
We still emit `#![feature(stdsimd)]` in a lot of places because we're still pinned to an old nightly on 1.65, but in 1.78 (https://github.com/rust-lang/rust/pull/117372), it was split into multiple features.  So when we upgrade to stable/a newer nightly for the transpiler output, we need to handle this.

kkysen:
This fixes the following differences with edition 2024:
* [x] `#[unsafe(no_mangle)]`
* [x] `#[unsafe(export_name = "")]`
* [x] `#[unsafe(link_section = "")]`
* [x] `unsafe extern "C" {`
* [ ] (#1663) `unsafe fn`s and `safe fn`s in `unsafe extern "C" {}`s
  * not required, default to `unsafe fn`s
* [x] `#![feature(stdsimd)]` removed and split up
* [x] stabilized `#![feature(raw_ref_op)]` removed
* [x] stabilized `#![feature(label_break_value)]` removed
* [x] stabilized `#![feature(asm)]` removed
* [x] `[profile.release] strip = "debuginfo"` being the default
* [ ] (#1658) updated atomic intrinsics (`atomics.c`)
* [ ] (#1636) `gen` reserved keyword (`keywords.c`)
* [ ] (#1657) updated `VaListImpl` (`varargs.c`)
* [ ] (#1662) `#[warn(unsafe_op_in_unsafe_fn)]` enabled by default
  * In `tests/unit/`, we add `#![allow(unsafe_op_in_unsafe_fn)]` for now.
  * Proper fix of an outer `unsafe` block in `unsafe fn`s (or finer-grained ones) we can do later.
* [ ] (#1656) `pref_align_of` removed (https://github.com/rust-lang/rust/pull/141803)
  * used to translate `__alignof`
  * no alternative, so may have to drop support

This also runs the correct `rustfmt`, as we were previously just picking up whichever one `rustup` resolved.

* Fixes #1298.

This updates all of the transpiler snapshot tests and `tests/unit/` tests that work on edition 2024 with the above fixes.  The remaining fixes I'll fix in separate PRs, as they might be trickier.

### DIFF
--- a/c2rust-transpile/tests/snapshots.rs
+++ b/c2rust-transpile/tests/snapshots.rs
@@ -332,6 +332,11 @@ fn test_factorial() {
 }
 
 #[test]
+fn test_fn_attrs() {
+    transpile("fn_attrs.c").run();
+}
+
+#[test]
 fn test_gotos() {
     transpile("gotos.c").run();
 }
--- a/c2rust-transpile/tests/snapshots/fn_attrs.c
+++ b/c2rust-transpile/tests/snapshots/fn_attrs.c
@@ -1,0 +1,47 @@
+static __attribute__((always_inline)) void always_inline_static(void) {}
+static __attribute__((__noinline__)) void noinline_static(void) {}
+static void inline inline_static(void) {}
+// __inline__ can be used in place of inline for ISO C89 compatibility
+static void __inline__ alt_kw_inline_static(void) {}
+static void inline __attribute__((__gnu_inline__)) gnu_inline_static(void) {}
+static void __attribute__((used, __cold__)) cold_used_attrs(void) {}
+void __attribute__((__always_inline__)) always_inline_nonstatic(void) {}
+void inline __attribute__((gnu_inline)) gnu_inline_nonstatic(void) {}
+void __attribute__((noinline)) noinline_nonstatic(void) {}
+void inline inline_nonstatic(void) {}
+void __inline__ alt_kw_inline_nonstatic(void) {}
+extern void gnu_inline_non_canonical_definition_extern(void);
+extern void inline inline_extern(void) {}
+extern void __inline__ alt_kw_inline_extern(void) {}
+extern void inline __attribute__((always_inline)) always_inline_extern(void) {}
+extern void inline __attribute__((__gnu_inline__)) gnu_inline_extern(void) {}
+extern void inline __attribute__((gnu_inline, always_inline)) always_inline_gnu_inline_extern(void) {}
+extern void inline __attribute__((gnu_inline)) gnu_inline_non_canonical_definition_extern(void) {}
+
+// Skipped to keep this a platform-independent test.
+// #ifndef __APPLE__
+// // aliases are not allowed on darwin
+// void __attribute__((alias("inline_extern"))) aliased_fn(void);
+// #endif // __APPLE__
+
+void ensure_use(void) {
+    always_inline_static();
+    always_inline_nonstatic();
+    gnu_inline_nonstatic();
+    inline_static();
+    alt_kw_inline_static();
+    noinline_static();
+    inline_nonstatic();
+    alt_kw_inline_nonstatic();
+    gnu_inline_static();
+    inline_extern();
+    alt_kw_inline_extern();
+    always_inline_extern();
+    gnu_inline_extern();
+    always_inline_gnu_inline_extern();
+    gnu_inline_non_canonical_definition_extern();
+
+// #ifndef __APPLE__
+//     aliased_fn();
+// #endif // __APPLE__
+}
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@fn_attrs.c.2021.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@fn_attrs.c.2021.snap
@@ -1,0 +1,71 @@
+---
+source: c2rust-transpile/tests/snapshots.rs
+expression: cat tests/snapshots/fn_attrs.2021.rs
+---
+#![allow(
+    dead_code,
+    non_camel_case_types,
+    non_snake_case,
+    non_upper_case_globals,
+    unused_assignments,
+    unused_mut
+)]
+#![feature(linkage)]
+#[inline(always)]
+unsafe extern "C" fn always_inline_static() {}
+#[inline(never)]
+unsafe extern "C" fn noinline_static() {}
+#[inline]
+unsafe extern "C" fn inline_static() {}
+#[inline]
+unsafe extern "C" fn alt_kw_inline_static() {}
+#[inline]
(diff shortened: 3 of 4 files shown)


Rua:
In the 2024 edition, the [`unsafe-op-in-unsafe-fn`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn) lint now also warns by default. A simple fix could be to just add an extra `unsafe { }` block to each function, but ultimately marking only the actual unsafe operations as unsafe could be preferred.

c2rust-ast-builder/src/builder.rs
```diff
@@ -513,15 +514,15 @@ impl Builder {
         K: Make<Path>,
         V: Make<Lit>,
     {
-        let meta = mk().meta_namevalue(key, value);
+        let meta = self.clone().meta_namevalue(key, value);
```
ahomescu, line 517: So this now copies the unsafety of the item, into the attribute? Is that always safe to duplicate?

c2rust-ast-builder/src/builder.rs
```diff
@@ -1974,34 +1982,35 @@ impl Builder {
     {
         let path = path.make(&self);
         let args = args.make(&self);
-        Meta::List(MetaList {
+        let span = self.span;
+        self.meta(Meta::List(MetaList {
```
ahomescu, line 1986: Will this emit things like `#[unsafe(derive(Clone, Copy))]`? That's what it reads like to me.

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -289,7 +290,15 @@ fn emit_lib_rs(
 /// on a nightly toolchain until the `c_variadics` feature is stable.
 fn emit_rust_toolchain(tcfg: &TranspilerConfig, build_dir: &Path) {
     let output_path = build_dir.join("rust-toolchain.toml");
-    let output = include_str!("generated-rust-toolchain.toml").to_string();
+    let toolchain = tcfg.edition.toolchain().strip_prefix("+").unwrap();
+    let output = format!(
+        "
```
ahomescu: If you used a raw string here `r#"`, would you still need the escapes below?

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -414,13 +416,16 @@ impl<'c> Translation<'c> {
...
-                // TODO See #1298.
-                // In Rust 1.7, `#![feature(stdsimd)]` was removed and split into (at least):
-                // `#![feature("stdarch_arm_hints")]` and
-                // `#![cfg_attr(target_arch = "arm", feature(stdarch_arm_neon_intrinsics))]`.
-                // self.use_feature("stdarch_arm_hints");
-                // self.use_feature("stdarch_arm_neon_intrinsics"); // TODO need to add `cfg_attr` support.
+                if self.tcfg.edition < Edition2024 {
+                    self.use_feature("stdsimd");
+                } else {
+                    // Edition 2024 was release in Rust 1.85.
```
ahomescu:
```suggestion
                    // Edition 2024 was released in Rust 1.85.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -370,18 +372,43 @@ pub fn stmts_block(mut stmts: Vec<Stmt>) -> Block {
...
+    new_name: &str,
+    old_name: &str,
+    edition: RustEdition,
+) -> Builder {
     if new_name == old_name {
         if in_extern_block {
             mk() // There is no mangling by default in extern blocks anymore
         } else {
-            mk().single_attr("no_mangle") // Don't touch my name Rust!
+            mk().unsafety(attr_unsafety(edition))
```
ahomescu, line 404: I think (not 100% sure) that if you do this, it will also make the item itself `unsafe`. Is that what you intended?

c2rust-ast-builder/src/builder.rs
ahomescu: For example, `#[inline]` is a safe function attribute even in 2024.

c2rust-ast-builder/src/builder.rs
kkysen: Yes.  We want `self`'s `unsafety` to propagate to inner `meta_namevalue`s it creates.

c2rust-ast-builder/src/builder.rs
kkysen: No, not unless you call `.unsafe_()` before creating it, which we don't.  We only call `.unsafety` on attributes that are actually `unsafe`, which are updated in this PR.

c2rust-transpile/src/build_files/mod.rs
kkysen: Let me check.

c2rust-transpile/src/translator/mod.rs
kkysen: It doesn't, only the attribute.  You can look at the snapshots to see what this does, and only the attributes are made `unsafe`.

c2rust-transpile/src/build_files/mod.rs
kkysen: Fixed in 220a2dd.

c2rust-ast-builder/src/builder.rs
kkysen: Oh, I do see `#[unsafe(inline)]` is being created.  That shouldn't be happening.  `unsafety` must be getting propagated somewhere else where it shouldn't be.

kkysen: That's now fixed in 2712287 and tested in 79cae37.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -370,18 +372,45 @@ pub fn stmts_block(mut stmts: Vec<Stmt>) -> Block {
...
+fn extern_block_unsafety(edition: RustEdition) -> Unsafety {
+    if edition >= Edition2024 {
+        Unsafety::Unsafe
+    } else {
+        Unsafety::Normal
+    }
+}
+
+/// Whether attributes can be `unsafe` in this edition.
+fn attr_unsafety(edition: RustEdition) -> Unsafety {
```
ahomescu, line 385: Could be named slightly better, e.g., `attr_needs_unsafe`.

kkysen: Hmm, I feel like `attr_needs_unsafe` implies it'll return a `bool`, whereas this returns `Unsafety`.  I'm going to merge this now as-is, but I can come back and change this later.

PR #1644, opened 2026-03-06
transpile: Properly handle address-of static compound literals

### LINKED ISSUE #1582: File-scope compound literal translated incorrectly, causing segfault on write

### LINKED ISSUE #1610: Allow `WithStmts` in static initialisers to contain statements
Currently, returning a `WithStmts` that contains statements, inside a static initialiser, triggers a panic: "Expected no side-effects in static initializer". But that requirement is too tight: it's perfectly valid for static initialisers to contain statements, at least if they are declaration statements.

This issue currently blocks #1582. The solution used in #1407 works for local variables, but can't be extended to statics because of the panic.

Rua:
- Fixes #1582.
- Fixes #1610.

This allows limited use of statements in translations of statics, only if the statements are `Stmt::Item`. This is then used to solve the original bug, by declaring a fresh static variable to take the address of.

This code doesn't currently account for situations in which the compound literal expression needs to be sectioned off for late initialisation, as that would have complicated the code a fair bit further.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2132,7 +2132,7 @@ impl<'c> Translation<'c> {
                     if self.static_initializer_is_unsafe(initializer, typ) {
                         init.set_unsafe()
                     }
-                    let init = init.to_unsafe_pure_expr().unwrap();
+                    let init = init.wrap_unsafe().to_pure_expr().unwrap();
                     let item = static_def.span(span).static_item(new_name, ty, init);
                     items.push(item);
 
@@ -2480,7 +2480,8 @@ impl<'c> Translation<'c> {
                 // (since we translate all `fn`s as `unsafe`). Therefore, in `const` contexts,
                 // expose any underlying unsafety in the initializer with an `unsafe` block.
                 let init = if ctx.is_const {
-                    init.to_unsafe_pure_expr()
+                    init.wrap_unsafe()
+                        .to_pure_expr()
                         .expect("init should not have any statements")
                 } else {
                     init.into_value()
@@ -2488,7 +2489,7 @@ impl<'c> Translation<'c> {
 
                 let zeroed = self.implicit_default_expr(ctx, typ.ctype)?;
                 let zeroed = if ctx.is_const {
-                    zeroed.to_unsafe_pure_expr()
+                    zeroed.wrap_unsafe().to_pure_expr()
                 } else {
                     zeroed.to_pure_expr()
                 }
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3108,53 +3108,57 @@ impl<'c> Translation<'c> {
                 }
 
                 let mut val = mk().path_expr(vec![rustname]);
-
-                // If the variable is actually an `EnumConstant`, we need to add a cast to the
-                // expected integral type.
-                if let &CDeclKind::EnumConstant { .. } = decl {
-                    val = self.convert_cast_from_enum(qual_ty.ctype, val)?;
-                }
-
-                // If we are referring to a function and need its address, we
-                // need to cast it to fn() to ensure that it has a real address.
                 let mut set_unsafe = false;
-                if ctx.needs_address() {
-                    if let CDeclKind::Function { parameters, .. } = decl {
-                        let ty = self.convert_type(qual_ty.ctype)?;
-                        let actual_ty = self
-                            .type_converter
-                            .borrow_mut()
-                            .knr_function_type_with_parameters(
-                                &self.ast_context,
-                                qual_ty.ctype,
-                                parameters,
-                            )?;
-                        if let Some(actual_ty) = actual_ty {
-                            if actual_ty != ty {
-                                // If we're casting a concrete function to
-                                // a K&R function pointer type, use transmute
-                                self.import_type(qual_ty.ctype);
 
-                                val = transmute_expr(actual_ty, ty, val);
-                                set_unsafe = true;
-                            }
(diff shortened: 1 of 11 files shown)


c2rust-transpile/src/with_stmts.rs
```diff
@@ -90,6 +90,18 @@ impl<T> WithStmts<T> {
         &mut self.stmts
     }
 
+    /// If all statements in self.stmts are Item statements, returns the contained Items.
+    /// Otherwise, returns None.
```
kkysen:
```suggestion
    /// If all statements in `self.stmts` are [`Item`]s, returns the contained [`Item`]s.
    /// Otherwise, returns [`None`].
```

c2rust-transpile/src/translator/mod.rs
kkysen, line 1: Can we wait to merge this until after #1643 merges, which has gotten pretty long?  #1643 changes things like `mk_linkage`, `"link_section"`, etc., so it's going to have conflicts here, and I'd like to review the changes here after it's rebased on that.

c2rust-transpile/src/with_stmts.rs
```diff
@@ -17,6 +17,13 @@ impl<T> WithStmts<T> {
             is_unsafe: false,
         }
     }
+    pub fn new_unsafe(stmts: Vec<Stmt>, val: T) -> Self {
+        WithStmts {
+            stmts,
+            val,
+            is_unsafe: true,
+        }
+    }
```
kkysen:
```suggestion

    pub fn new_unsafe(stmts: Vec<Stmt>, val: T) -> Self {
        WithStmts {
            stmts,
            val,
            is_unsafe: true,
        }
    }

```

I'm not sure why there aren't newlines between the functions here, but let's fix it for new functions.

kkysen: Also, I feel like these 4 functions would be better off as builder-like methods, like this:

```rust
    pub fn new_val(val: T) -> Self {
        WithStmts {
            stmts: vec![],
            val,
            is_unsafe: false,
        }
    }

    pub fn stmts(self, stmts: Vec<Stmt>) -> Self {
        Self { stmts, ..self }
    }

    pub fn is_unsafe(self, is_unsafe: bool) -> Self {
        Self { is_unsafe, ..self }
    }
```

c2rust-transpile/tests/snapshots/snapshots__transpile@compound_literals.c.2021.snap
```diff
@@ -14,18 +14,22 @@ input_file: c2rust-transpile/tests/snapshots/compound_literals.c
 #![feature(raw_ref_op)]
 #[no_mangle]
 pub static mut static_single_int: ::core::ffi::c_int = 42;
+static mut c2rust_fresh0: ::core::ffi::c_int = 42;
 #[no_mangle]
 pub static mut static_single_int_ptr: *mut ::core::ffi::c_int =
-    &42 as *const ::core::ffi::c_int as *mut ::core::ffi::c_int;
+    unsafe { &raw const c2rust_fresh0 as *mut ::core::ffi::c_int };
```
kkysen, line 20: Why do these need to be `unsafe`?

Rua: Hmm, I guess taking a raw pointer is not unsafe, only taking a regular reference is.

c2rust-transpile/src/with_stmts.rs
Rua: `WithStmts` doesn't use the builder pattern anywhere, so converting it to one seems like a separate refactor that belongs in its own PR?

c2rust-transpile/tests/snapshots/snapshots__transpile@compound_literals.c.2021.snap
Rua: It seems that `convert_address_of_common` doesn't mark its expression as unsafe, and never did even before my refactorings? Even when it's taking the address of a static? How did it work before then?

Rua:
I realised that in the current nightly, taking the raw address of a static is `unsafe`, but it's not unsafe in the final stabilised version of the `raw_ref_op` feature. So I gated that to only apply to the 2021 edition. I also did the same for `DeclRef`s of variables, where it seemed to have been missing all this time.

ahomescu (approved):

PR #1645, opened 2026-03-06
[DNM] Fix wide-string `override_ty`

fw-immunant:
This isn't suitable for merging, but shows the constraints we're working in--namely, we need to synthesize a new `CTypeId` (for the appropriately-sized array type of wide-chars, to replace the same array of `c_int`s) in a place where the `TypedAstContext` is not mutable.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3778,10 +3778,41 @@ impl<'c> Translation<'c> {
                             return self.convert_expr(ctx, expr, override_ty);
                         }
                     }
+                    if kind == CastKind::ArrayToPointerDecay && override_ty.is_some() {
+                        let target_elem = match &self
+                            .ast_context
+                            .resolve_type(override_ty.unwrap().ctype)
+                            .kind
+                        {
+                            CTypeKind::Pointer(qtype) => qtype.ctype,
+                            k => panic!("ArrayToPointerDecay producing non-pointer type {k:?}"),
+                        };
+                        let new_array_override_ty_kind =
+                            match &self.ast_context.resolve_type(source_ty.ctype).kind {
+                                CTypeKind::ConstantArray(_elem, len) => {
+                                    CTypeKind::ConstantArray(target_elem, *len)
+                                }
+                                CTypeKind::VariableArray(_elem, len) => {
+                                    CTypeKind::VariableArray(target_elem, *len)
+                                }
+                                k => panic!("ArrayToPointerDecay on non-array type {k:?}"),
+                            };
+                        //
+                        //           && Some(source_ty.ctype) != override_ty.map(|x| x.ctype)
+                        let new_array_override_ty = self
+                            .ast_context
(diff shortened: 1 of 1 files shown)


Rua:
I was thinking about this but had a different approach in mind. Instead of trying to synthesize a new type, just pass an override for the pointee/element type to `convert_array_to_pointer_decay` (and only there). For array decay it's always the same pattern between source and target types, where `[foo; N]` turns into `*const/mut foo`. It's not even technically a cast, anyway.

fw-immunant:
> I was thinking about this but had a different approach in mind. Instead of trying to synthesize a new type, just pass an override for the pointee/element type to `convert_array_to_pointer_decay` (and only there). For array decay it's always the same pattern between source and target types, where `[foo; N]` turns into `*const/mut foo`. It's not even technically a cast, anyway.

I believe the problem with that approach is that the `val` argument to `convert_array_to_pointer_decay` has already been translated (its type is `WithStmts<Box<Expr>>`) with the wrong `override_ty`.

Of course, we could address that via inversion of control, changing `convert_cast` to be responsible for the translation of the value to cast.

fw-immunant:
> This isn't suitable for merging, but shows the constraints we're working in--namely, we need to synthesize a new `CTypeId` (for the appropriately-sized array type of wide-chars, to replace the same array of `c_int`s) in a place where the `TypedAstContext` is not mutable.

One note - I used `dashmap` here in the hopes that it would let us insert new `CTypeId`s to the `c_types` map without disturbing the required `Index` impl which must return `&CType`, but since it doesn't actually do that (we have to leak a boxed `CType` to do so because `DashMap` returns its own `Ref<V>` type rather than `&V`), dashmap is not really helping compared to a `RefCell<HashMap>`, and what would actually solve the `Index` problem is an append-only, address-stable hashmap. That should fix the hackiness side of things here.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3778,10 +3778,41 @@ impl<'c> Translation<'c> {
...
+                                    CTypeKind::ConstantArray(target_elem, *len)
+                                }
+                                CTypeKind::VariableArray(_elem, len) => {
+                                    CTypeKind::VariableArray(target_elem, *len)
+                                }
+                                k => panic!("ArrayToPointerDecay on non-array type {k:?}"),
+                            };
+                        //
+                        //           && Some(source_ty.ctype) != override_ty.map(|x| x.ctype)
+                        let new_array_override_ty = self
```
Rua, line 3802: Rather than trying to synthesize a completely new array type here, could `convert_array_to_pointer_decay` not be modified to take an element type instead of source and target types, and then called directly?

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -716,9 +716,26 @@ impl TypedAstContext {
     }
 
     pub fn type_for_kind(&self, kind: &CTypeKind) -> Option<CTypeId> {
```
Rua, line 718: This function doesn't actually return `None` anymore, that I can see?

PR #1647, opened 2026-03-07
transpile: Split off new `macros` module

Rua:
No functional changes, just moving more stuff to dedicated modules.

### DIFF
--- a/c2rust-transpile/src/translator/macros.rs
+++ b/c2rust-transpile/src/translator/macros.rs
@@ -1,0 +1,251 @@
+use c2rust_ast_builder::mk;
+use failure::format_err;
+use log::{info, trace};
+use proc_macro2::{Span, TokenStream};
+use syn::{Expr, MacroDelimiter};
+
+use crate::c_ast::{CDeclId, CExprId, CQualTypeId, CTypeId, CTypeKind};
+use crate::diagnostics::{TranslationError, TranslationResult};
+use crate::translator::{ConvertedDecl, ExprContext, MacroExpansion, Translation};
+use crate::with_stmts::WithStmts;
+use crate::TranslateMacros;
+
+impl<'c> Translation<'c> {
+    pub fn convert_macro(
+        &self,
+        ctx: ExprContext,
+        decl_id: CDeclId,
+        span: Span,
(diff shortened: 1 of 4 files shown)


c2rust-transpile/src/translator/macros.rs
```diff
@@ -0,0 +1,252 @@
...
+use log::{info, trace};
+use proc_macro2::{Span, TokenStream};
+use syn::{Expr, MacroDelimiter};
+
+use crate::{
+    diagnostics::{TranslationError, TranslationResult},
+    translator::{ConvertedDecl, ExprContext, MacroExpansion, Translation},
+    with_stmts::WithStmts,
+    CDeclId, CExprId, CQualTypeId, CTypeId, CTypeKind, TranslateMacros,
+};
```
kkysen: nit: I think we generally tend to prefer (not a hard rule at all) [`imports_granularity = "Module"`](https://rust-lang.github.io/rustfmt/?version=v1.9.0&search=import#imports_granularity), so one grouped import per module instead of per crate.  Unfortunately the `rustfmt` setting for this is unstable.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -4225,78 +4225,6 @@ impl<'c> Translation<'c> {
...
-
-                // Ignoring Complex casts for now
-                _ => {
-                    warn!(
-                        "Unknown CastKind for {:?} to {:?} cast. Defaulting to BitCast",
-                        source_ty_kind, target_ty_kind,
-                    );
-
-                    CastKind::BitCast
-                }
```
kkysen, line 4454: Why not leave this inside `fn CastKind::from_types`?

Rua: I wasn't sure if it would be appropriate to have a `warn!` inside a relatively simple conversion function like that. That seemed like a higher-level thing better suited to the surrounding context. I can move it in if you prefer.

Rua: I removed that part of the PR because it seemed kinda off-topic anyway.

kkysen (approved):

PR #1652, opened 2026-03-08
transpile: Split off various `c_ast` submodules

Rua:


c2rust-transpile/src/c_ast/c_type.rs
```diff
@@ -0,0 +1,826 @@
+use super::{
+    Attribute, CDeclId, CDeclKind, CEnumId, CExprId, CRecordId, CTypedefId, Located,
+    TypedAstContext,
+};
```
kkysen: Can you just make these normal imports?

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -16,12 +16,12 @@ pub use self::conversion::*;
...
 pub use c2rust_ast_exporter::clang_ast::{BuiltinVaListKind, SrcFile, SrcLoc, SrcSpan};
 
+mod c_type;
 mod conversion;
 pub mod iterators;
 mod print;
 
-#[derive(Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Copy, Clone)]
-pub struct CTypeId(pub u64);
+pub use c_type::*;
```
kkysen: Can you just change the uses to include `c_type`?  There's no public API that we need to preserve.

c2rust-transpile/src/c_ast/mod.rs
kkysen: It might be useful to keep all of these together.  I'm not sure, though.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -919,111 +660,6 @@ impl TypedAstContext {
...
-    /// In Clang 15 and below, the Clang AST resolves typedefs in the expression type of unary and
-    /// binary expressions. For example, a BinaryExpr node adding two `size_t` expressions will be
-    /// given an `unsigned long` type rather than the `size_t` typedef type. This behavior changed
-    /// in Clang 16. This method adjusts AST node types to match those produced by Clang 16 and
-    /// newer; on these later Clang versions, it should have no effect.
-    ///
-    /// This pass is necessary because we reify some typedef types (such as `size_t`) into their own
-    /// distinct Rust types. As such, we need to make sure we know the exact type to generate when
-    /// we translate an expr, not just its resolved type (looking through typedefs).
-    pub fn bubble_expr_types(&mut self) {
```
kkysen, line 1134: Not sure if this should go in `c_expr` or `c_type`.

c2rust-transpile/src/c_ast/c_expr.rs
```diff
@@ -0,0 +1,925 @@
...
+}
+
+impl Display for UnTypeOp {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+impl UnOp {
+    /// Check is the operator is rendered before or after is operand.
```
kkysen:
```suggestion
    /// Check if the operator is rendered before or after the operand.
```

Found a typo.  Fix it in a follow-up commit in this PR?

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -337,939 +310,6 @@ impl TypedAstContext {
...
-            CTypeKind::Function(_, _, _, no_return, _) => no_return,
-            _ => false,
-        }
-    }
-
-    /// Pessimistically try to check if an expression is `const`.
-    /// If it's not, or we can't tell if it is, return `false`.
-    ///
-    /// This should be a top-down, pessimistic/conservative analysis.
-    pub fn is_const_expr(&self, expr: CExprId) -> bool {
```
kkysen, line 882: Previously, related things like `fn is_const_expr` and `fn is_const_stmt` were grouped together, but this reorganization breaks that.

c2rust-transpile/src/c_ast/mod.rs
Rua: I don't have a particular stance on it.

c2rust-transpile/src/c_ast/mod.rs
Rua: You mean the `include!` macro?

c2rust-transpile/src/c_ast/mod.rs
Rua: Hmm, what do you suggest doing?

c2rust-transpile/src/c_ast/mod.rs
Rua: Ah no, you mean just putting `use crate::c_ast::c_type::` and such in other modules.

c2rust-transpile/src/c_ast/mod.rs
Rua: I could move all the `impl TypedAstContext` back into the main module?

c2rust-transpile/src/c_ast/mod.rs
Rua: Types just happen to be a property of expressions, but it's modifying `CExprKind`'s rather than anything type-related. So `c_expr` seems more appropriate?

kkysen (changes requested):
As I mentioned in the other PR, we don't have the capacity to review more pure-refactoring changes.  Sometimes they're fairly easy to review, like that `C*` prefix one, but here it's much more tricky, since we have to consider where each item goes and which items were supposed to be together, etc.  So you can leave this PR open; maybe we'll eventually get to it, since cleaning up the huge modules with often unrelated items is still good, but we won't have time to review it now.  So I think you can also stop rebasing it; I assume that takes some work as well.

Of these commits, c8f19cf is IMO nice and easy to review, and 5019320 is just a docs/typo fix, so we can merge those if you want to pull those out separately.

PR #1655, opened 2026-03-11
transpile: Add AST node parent information

Rua:
This collects information about the parent nodes of each AST node (except types). It's not currently used for anything, but could potentially be used to solve #1321, by allowing conversion functions across the transpiler to easily find the types of their parent nodes.

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -388,17 +388,23 @@ impl ConversionContext {
 
     /// Add a `CStmt` node into the `TypedAstContext`
     fn add_stmt(&mut self, id: ImporterId, stmt: CStmt) {
-        self.typed_context.c_stmts.insert(CStmtId(id), stmt);
+        let id = CStmtId(id);
+        self.typed_context.add_stmt_parents(id, &stmt.kind);
+        self.typed_context.c_stmts.insert(id, stmt);
     }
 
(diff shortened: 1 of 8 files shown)


ahomescu:
I don't think we should merge a massive change that isn't used for anything. Edit: unless we also want #1679, in which case this is a prerequisite and not an independent change.

Rua:
There are also other things that this makes possible, not all of which have been considered yet necessarily. One thing that comes to mind is being able to completely replace `ExprContext`, so that all context information is available globally and no longer relies on passing things down the call stack. That may be a requirement for making a fix for #1321, since macro translation is done outside the call stack and thus doesn't currently know what context each of its expansions exists in.

EDIT: To illustrate, the `ctx` being passed here is almost certainly wrong, because it's the context of the current expansion site, and not of the `MacroObject` itself, nor of all expansion sites:
https://github.com/immunant/c2rust/blob/c11f9aa29bc7512a0b814aad214e2177bc36e42e/c2rust-transpile/src/translator/mod.rs#L4179

Rua:
I've rebased this onto master, so it no longer requires #1652.

Rua:
#1679 no longer depends on this PR, so this is now a standalone addition that is not required for anything anymore. If you decide it's not useful you can close it.

ahomescu:
Still debating whether to land this, I'll leave it open in case it might be useful later.

PR #1656, opened 2026-03-12
transpile: replace `pref_align_of` with `align_of` in edition 2024

kkysen:
Since `core::intrinsics::pref_align_of` was removed, we can't fully support `__alignof`/`__alignof__` anymore, but using normal `core::mem::align_of` should still be a valid value.

This also updates `tests/unit/misc` to edition 2024, as its usage of `__alignof__` was the only thing holding it back.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3379,6 +3379,7 @@ impl<'c> Translation<'c> {
         &self,
         mut type_id: CTypeId,
         preferred: bool,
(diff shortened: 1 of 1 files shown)


thedataking (commented):
The translation proposed here seems valid. 

However, the translation of GNUs `__alignof__` is no longer bit-for-bit faithful (e.g.  `__alignof__(double) == 8` but `alignof(double) == 4` on i386) and the source code might rely on a particular layout via static asserts or allocation decisions. Would it make sense to emit a warning?

thedataking (approved):

PR #1658, opened 2026-03-14
transpile: fix atomic intrinsics in edition 2024 

kkysen:
Most of these commits are refactoring things to be in a shape where adding edition 2024 support is simple.  Then the last commit, 07c2635, actually adds edition 2024 support.

### DIFF
--- a/c2rust-transpile/src/translator/atomics.rs
+++ b/c2rust-transpile/src/translator/atomics.rs
@@ -49,9 +49,15 @@ impl<'c> Translation<'c> {
     fn atomic_intrinsic_expr_edition_2024(
         &self,
         base_name: &str,
-        num_ty_params: usize,
         orders: &[Ordering],
     ) -> Box<Expr> {
+        let num_ty_params = match base_name {
+            "fence" => 0,
+            "load" | "store" | "xchg" | "cxchg" | "cxchgweak" => 1,
+            "xadd" | "xsub" | "or" | "xor" | "nand" | "and" => 2,
+            _ => unimplemented!("unknown atomic intrinsic: {base_name}"),
+        };
+
         let args = repeat_n("_".to_owned(), num_ty_params)
             .chain(orders.iter().map(|&order| {
                 let order = order_ty_name(order);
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/atomics.rs
```diff
@@ -338,16 +391,16 @@ impl<'c> Translation<'c> {
...
+        order_succ: Ordering,
+        order_fail: Ordering,
         dst: Box<Expr>,
         old_val: Box<Expr>,
         src_val: Box<Expr>,
         returns_val: bool,
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
-        self.use_feature("core_intrinsics");
-
         // Emit `atomic_cxchg(a0, a1, a2).idx`
```
kkysen, line 407: Some of these comments I left as is.  I'm not sure how to phrase them with edition 2024, and also even with just edition 2021, this leaves out `cxchgweak`.

c2rust-transpile/src/translator/atomics.rs
```diff
@@ -3,7 +3,94 @@ use crate::format_translation_err;
...
+            std::any::type_name::<Ordering>()
+        ),
+    }
+}
+
 impl<'c> Translation<'c> {
+    fn atomic_intrinsic_expr_edition_2021(
+        &self,
+        base_name: &str,
+        orders: &[Ordering],
```
ahomescu, line 41: There are at most 2 of these, right? We could add an assertion.

c2rust-transpile/src/translator/atomics.rs
ahomescu: If you mean rewrite the comment, I think it's fine as is.

kkysen: Yeah, I was talking about the comment.  I'll leave it as is then.

c2rust-transpile/src/translator/atomics.rs
kkysen: It'll show up as an error in the snapshot tests, but sure.

kkysen: Fixed in acc840d.

PR #1659, opened 2026-03-14
transpile: Translate `EnumConstant` with associated constants

### LINKED ISSUE #1498: Translate enums as newtype structs with associated constants?

Rua:
- Depends on #1746
- Depends on #1620
- Fixes #1498

### DIFF
--- a/c2rust-transpile/src/translator/enums.rs
+++ b/c2rust-transpile/src/translator/enums.rs
@@ -2,6 +2,7 @@ use c2rust_ast_builder::mk;
 use proc_macro2::Span;
 use syn::Expr;
 
+use crate::c_ast::iterators::SomeId;
 use crate::{
     diagnostics::TranslationResult,
     translator::{signed_int_expr, ConvertedDecl, ExprContext, Translation},
@@ -36,8 +37,6 @@ impl<'c> Translation<'c> {
     pub fn convert_enum_constant(
         &self,
         enum_constant_id: CEnumConstantId,
-        span: Span,
-        value: ConstIntExpr,
     ) -> TranslationResult<ConvertedDecl> {
         let name = self
             .renamer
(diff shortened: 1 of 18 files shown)


c2rust-transpile/src/translator/enums.rs
```diff
@@ -23,44 +25,57 @@ impl<'c> Translation<'c> {
...
+                    name,
+                );
+                let (span, init) = self.make_enum_constant_init(enum_constant_id);
+                mk().span(span)
+                    .pub_()
+                    .const_impl_item(name_rs, enum_type.clone(), init)
+            })
+            .collect();
+
+        let impl_block = mk().impl_item(mk().ident_ty(enum_name), constants);
```
ahomescu: Could this get a span?

ahomescu (changes requested):

c2rust-transpile/src/translator/enums.rs
Rua: I've given it the same span as the newtype struct. I don't know if that's correct, I'm not sure what the purpose of the spans is.

ahomescu: One use is comment preservation, they get inserted into the final Rust text using the spans.

ahomescu (approved):

PR #1662, opened 2026-03-16
transpile: add `unsafe {}` blocks in `unsafe fn`s and add `#![deny(unsafe_op_in_unsafe_fn)]`

kkysen:
Now that `#![warn(unsafe_op_in_unsafe_fn)]` is on by default in edition 2024, we want to avoid such warnings (without just doing `#![allow(unsafe_op_in_unsafe_fn)]`, as we were before for `tests/unit`).  We do this by wrapping `unsafe fn`'s bodies in an `unsafe {}` if they are non-empty.  Later, we could potentially narrow these `unsafe {}`s to their specific `unsafe` operations.  This is only on by default in edition 2024, but it works in edition 2021, too, so we just unconditionally do it (simpler and better practice), and emit `#![deny(unsafe_op_in_unsafe_fn)]` to enforce it as well.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -1606,17 +1606,20 @@ impl<'c> Translation<'c> {
         let mut features = vec![];
         features.extend(self.features.borrow().iter());
         features.extend(self.type_converter.borrow().features_used());
-        let mut pragmas: PragmaVec = vec![(
-            "allow",
-            vec![
-                "non_upper_case_globals",
-                "non_camel_case_types",
-                "non_snake_case",
-                "dead_code",
-                "unused_mut",
-                "unused_assignments",
-            ],
-        )];
+        let mut pragmas: PragmaVec = vec![
+            (
+                "allow",
+                vec![
+                    "non_upper_case_globals",
+                    "non_camel_case_types",
+                    "non_snake_case",
+                    "dead_code",
+                    "unused_mut",
+                    "unused_assignments",
+                ],
+            ),
+            ("deny", vec!["unsafe_op_in_unsafe_fn"]),
+        ];
         if self.tcfg.cross_checks {
             features.append(&mut vec!["plugin"]);
             pragmas.push(("cross_check", vec!["yes"]));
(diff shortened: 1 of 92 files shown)


Rua:
I suppose it could be useful to have an easy way to disable these automatic blocks later, to test if the correct `unsafe` is added in the right places? Perhaps via a temporary command line argument that instructs the transpiler to attempt to emit only the `unsafe` blocks it thinks it needs, rather than the single function-wide one?

c2rust-ast-builder/src/builder.rs
```diff
@@ -1407,11 +1407,24 @@ impl Builder {
...
+            // so we need to wrap any `unsafe` operation in an `unsafe` block.
+            // For now, just wrap the whole function body in an `unsafe` block,
+            // which we can later narrow down to individual `unsafe` operations.
+            // In previous editions, this does not warn by default,
+            // but it can be turned on, and it's generally good practice anyways,
+            // so we do this unconditionally (i.e., no edition check needed).
+            let unsafe_expr = mk().unsafe_block_expr(block);
+            block = mk().block(vec![mk().expr_stmt(unsafe_expr)]);
+        }
+
```
fw-immunant, line 1427: I'm not sure we should do this in `c2rust_ast_builder::fn_item`. We only call this function a handful of times, so I guess it's not a huge deal either way. But I'd think that whether the function signature is unsafe or not is no longer generally relevant to whether the function body contains unsafe--only the caller knows that. Existing callers will be following the previous requirements of the 2021 edition, but there are only 3 of those (1 in `translator/main_function.rs` and 2 in `translator/mod.rs`), so we could update callers about as easily without pushing any complexity into the AST builder.

Not a blocking concern.

kkysen: There are only 3 callers right now, but we could easily add more, so we'd have to make sure they all include (or purposefully exclude) this.  I thought about doing it that way at first, but it seemed more complex and brittle to me.

I conditioned it on the function's unsafety because the safe `fn`s we emit, like `main`, are fully safe or already have `unsafe {}` blocks around their `unsafe` operations.

I think if we do decide to change how this works later, we should be able to easily change this then.

kkysen:
> I suppose it could be useful to have an easy way to disable these automatic blocks later, to test if the correct `unsafe` is added in the right places? Perhaps via a temporary command line argument that instructs the transpiler to attempt to emit only the `unsafe` blocks it thinks it needs, rather than the single function-wide one?

When we switch to `unsafe {}` blocks around operations, I think we can easily remove/disable this.  I'm not sure I see the need to add a CLI argument for this now.  We can just delete/comment out this code at that point.  Unless I'm misunderstanding your point.

PR #1664, opened 2026-03-16
refactor: Delete dead code in `driver.rs`

randomPoison:
`build_session` and `build_session_from_args` became dead code in [65d1abae9 ](https://github.com/immunant/c2rust/commit/65d1abae9ef2d3bb291b635cbc2e4a37dc0ed21c#diff-835fcd65bdd7b79f53678278ec0c7f38a94173326c74106845f4f251fd6ca7feL126-L133) back in 2019. Likewise, the commented out code has been sitting there since the same commit, so I think we're fine to delete all of this.


randomPoison:
> Also, could you squash the formatting commit so that they all remain green and bisectsble?

I always squash and merge, so when it lands on master it'll only be a single commit.

c2rust-refactor/src/driver.rs
```diff
@@ -92,115 +90,6 @@ impl<'a, 'tcx: 'a> RefactorCtxt<'a, 'tcx> {
     }
 }
 
-// /// Various driver bits that we have lying around at the end of `phase_1_parse_input`.  This is
```
ahomescu, line 95: Oh you're deleting this code. I'm not sure why it was here in the first place, or later commented out. Did you check the git history?

ahomescu (approved):

c2rust-refactor/src/driver.rs
randomPoison: It got commented out in the same toolchain roll that made the other stuff unused, I think it was replaced by rustc internals but I didn't dig into it too deeply since the code has been dead for years.

PR #1671, opened 2026-03-17
Fix compat with Clang 22

fw-immunant:


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2224,8 +2236,6 @@ class TranslateASTVisitor final
         auto recordAlignment = 0;
         auto byteSize = 0;
 
-        auto t = D->getTypeForDecl();
-
```
fw-immunant, line 2271: These are unused and the `getTypeForDecl` method no longer exists in Clang 22.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -813,6 +813,21 @@ impl ConversionContext {
                     self.processed_nodes.insert(new_id, expected_ty);
                 }
 
+                TypeTag::TagPredefinedSugarType => {
+                    let kind = from_value(ty_node.extras[0].clone())
+                        .expect("Predefined sugar type kind not found");
+
+                    // See `clang::PredefinedSugarKind`.
+                    let predef_sugar_ty = match kind {
```
kkysen, line 821: Does this affect the portable type code?

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -656,6 +656,7 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
         VisitQualType(t);
     }
 
+    #if CLANG_VERSION_MAJOR < 22
```
ahomescu: Do we indent preprocessor directives? In most projects, they start at column 1.

c2rust-transpile/src/c_ast/conversion.rs
kkysen: `PredefinedSugarKind` seems to be similar to what we're doing with the portable types, but just for `size_t`, `ssize_t`, and `ptrdiff_t`, so I'm wondering if we need to handle this specially.

fw-immunant:
I think there's an additional problem that comes up when actually testing against Clang 22: we hit an infinite loop with increasing memory consumption in `prune_unwanted_decls`. It seems show up while traversing libc types within the definition of `FILE`. I'm not sure exactly what's going on, but wonder if it has to do with the removal of `ElaboratedType`.

Crocodoctopus:
The infinite loops appear to have been caused by cycles in `DFNodes` and `NodeVisitor`. Adding guards gets all tests passing again w/ clang-22 except for two, which generate slightly different Rust:
```
// test_varargs
-    fn wcslen(__s: *const wchar_t) -> ::core::ffi::c_ulong;
+    fn wcslen(__s: *const wchar_t) -> usize;
```
```
// test_wide_strings
-    vprintf(fmt, ap.as_va_list());
+    vprintf(fmt as *const ::core::ffi::c_char, ap.as_va_list());
```
This is likely due to changes with clang-22 handles AST sugar but I'm unsure. Is it important that we keep tests outputting the _exact_ same Rust code?

ahomescu:
> Is it important that we keep tests outputting the exact same Rust code?

I think so, otherwise people with different clang versions will see different test results. We could say we prefer the shorter/more idiomatic output, but that would still cause the other clang version to fail the tests.

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: Was `t` an unused value? How is it safe to just delete?

c2rust-transpile/src/c_ast/iterators.rs
```diff
@@ -376,13 +376,15 @@ impl<'context> Iterator for DFExpr<'context> {
 pub struct DFNodes<'context> {
     context: &'context TypedAstContext,
     stack: Vec<SomeId>,
+    visited: HashSet<SomeId>,
```
ahomescu, line 380: It might be useful to have a comment why we now need to track visited children (and didn't need before clang 22).

thedataking:
From offline discussion:
> This is likely due to changes with clang-22 handles AST sugar but I'm unsure. 

likely culprit https://github.com/llvm/llvm-project/pull/143653

> Is it important that we keep tests outputting the _exact_ same Rust code?

no, tests should allow the new output from clang-22 and later since it is better

ahomescu:
In CI we seem to have clang 15 and 17 running, and the new outputs are failing on those. Since we want c2rust+clang22+ to output the new improved types, I think we'll need to improve the snapshots to include the clang version and duplicate the failing tests. It would also be a good idea to have a clang22 run in CI in addition to the current ones.

Crocodoctopus:
> In CI we seem to have clang 15 and 17 running, and the new outputs are failing on those. Since we want c2rust+clang22+ to output the new improved types, I think we'll need to improve the snapshots to include the clang version and duplicate the failing tests. It would also be a good idea to have a clang22 run in CI in addition to the current ones.

I had a bug that was causing a few tests w/ clang-18 to fail. Not sure how I missed that the first time. The other tests likely failed for the same reason. Should be resolved now.

How would I go about updating CI to include testing against clang22?

Edit: looks like no test runners provide clang22 by default

c2rust-ast-exporter/src/AstExporter.cpp
fw-immunant: Yes, `t` was unused here and in `VisitEnumDecl`.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -105,6 +106,18 @@ fn transpile_snapshot(
...
     imported_crates: &[&str],
 ) {
+    let c_file_name = c_path.file_name().unwrap().to_str().unwrap();
+    let c_file_name = sanitize_file_name(&c_file_name);
+
+    // Some versions of clang can produce results different from their snapshots,
+    // those tests append the clang version to the snapshot of the failing test.
+    let clang_ver = match (get_clang_major_version(), c_file_name.as_str()) {
+        (None, _) => panic!("Could not get clang version!"),
+        (Some(22), "varargs.c") => "clang22",
```
ahomescu: Don't we need "22 or higher" here? The snapshots can also be run locally.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -105,6 +106,18 @@ fn transpile_snapshot(
...
+        (Some(22), "wide_strings.c") => "clang22",
+        _ => <_>::default(),
```
ahomescu: Since going forward the new output is the default, it might be better to output in reverse: `.clang21` for the old files, empty for the new.

Crocodoctopus: I can set the old ones to be the oldest support versions that still pass tests, so clang7 I think? It makes sense to change the names on the clang version that require them to be updated.

Does it still make sense to keep these clang22? Its easier to keep track of I think. It tell you exactly what version of clang the snapshot should pass at.

c2rust-transpile/tests/snapshots.rs
Crocodoctopus: Yep.

.github/workflows/ci.yml
```diff
@@ -10,6 +10,10 @@ jobs:
     strategy:
       matrix:
         include:
+          - runner: ubuntu-latest
+            os: Linux
+            arch: x86_64
+            clang-version: 22
           - runner: ubuntu-latest
```
ahomescu, line 13: Maybe rename this one? `ubuntu-latest-clang18`?

.github/workflows/ci.yml
```diff
@@ -10,6 +10,10 @@ jobs:
     strategy:
       matrix:
         include:
+          - runner: ubuntu-latest
```
ahomescu: I wonder if at some point we'll start having problems when `ubuntu-latest` upgrades their LLVM.

Crocodoctopus: Possibly. Switched it off to 24.04, which is what ubunut-latest currently is.

.github/workflows/ci.yml
Crocodoctopus: This one is also 24.04 now.

c2rust-transpile/src/c_ast/iterators.rs
Crocodoctopus: Afaik, cycles could exist pre-clang-22 as well, we just never hit one.

.github/workflows/ci.yml
```diff
@@ -10,7 +10,13 @@ jobs:
     strategy:
       matrix:
         include:
-          - runner: ubuntu-latest
+          # TODO: Eventually github will probably have a test runner that
```
ahomescu: I think @thedataking set up our CI, adding him to take a look.

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: Why did you close the comment? It's hard to tell from the UI, is this preprocessor directive indented? It shouldn't be.

.github/workflows/ci.yml
thedataking: We don't have CI coverage for all our clang versions. It is easier to test clang-22 on mac, brew can install it. 

Also, I don't want a big CI change mixed into a PR that's about clang/LLVM22. If we want to update CI, that can be done in a follow up so we can consider that separately. For now, please rip out all the CI changes here.

Concrete issues:
- we're pinning the runner; is that a good idea?
- the way clang 22 is provisioned is totally different from the other provisioning steps (see `Install packages (Ubuntu)`)
- we're adding an OS and version check to a bunch of runner steps. This seems messy in and of itself.

If you think it is super important to test clang-22, try rolling the mac clang version to 22. If that doesn't work, let's discuss alternatives. ~My base stance is: we're not testing every clang version, what makes 22 special.~

Edit: forgot that clang-22 generates different code, sorry my bad.

Looking at https://github.com/actions/runner-images, it does not seem that Ubuntu 26.04 is imminent which is why I'm speculating that macOS might be the easiest way to install newer clang versions.

ahomescu:
`c2rust-build-paths` also has a list of supported clang/LLVM versions, you'll probably need to update that too.

.github/workflows/ci.yml
ahomescu: Right, but we still need to fix it eventually? `ubuntu-latest` will catch up.

thedataking (changes requested):
Please see if you can easily test clang-22 on the mac runners since Ubuntu doesn't package it. I would like to avoid the special casing for this package.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -672,6 +674,16 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 
         VisitQualType(t);
     }
+
+    #if CLANG_VERSION_MAJOR >= 22
```
ahomescu: Same here: preprocessor directives should start on column 1.

ahomescu (changes requested):

.github/workflows/ci.yml
ahomescu: > My base stance is: we're not testing every clang version, what makes 22 special.

Isn't that obvious from this PR? They made a breaking change to the AST.

.github/workflows/internal-testsuite.yml
```diff
@@ -81,6 +81,16 @@ jobs:
...
 
+    - name: Install Clang 22 (Ubuntu)
+      if: runner.os == 'Linux' && matrix.clang-version == 22
+      run: |
+        # Remove all other clang installs
+        sudo apt-get -qq update
+        sudo apt-get -qq install -y curl gnupg lsb-release software-properties-common
+        curl -fsSL -o llvm.sh https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 22
```
ahomescu: Could we install this with `brew` instead?

.github/workflows/ci.yml
ahomescu: From offline discussion: we agreed that the CI changes should go into a separate PR. That means that we won't have CI testing for clang 22 in between, but that should be fine for a short while.

ahomescu (changes requested):

.github/workflows/ci.yml
thedataking: > Isn't that obvious from this PR? They made a breaking change to the AST.

Yup. Spoke too soon. Should have refreshed my memory the non-CI bits of this change. My bad.

c2rust-ast-exporter/src/AstExporter.cpp
Crocodoctopus: Thought I fixed that a long time ago... had to put a few temporary hacks in AstExporter to get it working on Gentoo, guess it got reverted at some point.

.github/workflows/ci.yml
Crocodoctopus: Added support for testing clang22 @ macos. Looks like its passing okay now.

Also I can clean up the linux clang22 step instead if that's okay too. I can probably remove the special case if statements in the other test cases, and keep it entirely contained within the special install step. For ex, something like this:

```
      # TODO: Temporary hack until github has ubuntu-26.04 test runners
      - name: Install Clang 22 (Ubuntu)
        if: runner.os == 'Linux' && matrix.clang-version == 22
        run: |
          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
          brew install -q llvm@22
          export LLVM_PREFIX=$(brew --prefix llvm@22)
          echo "LLVM_CONFIG_PATH=$LLVM_PREFIX/bin/llvm-config" >> $GITHUB_ENV
          echo "LLVM_LIB_DIR=$($LLVM_PREFIX/bin/llvm-config --libdir)" >> $GITHUB_ENV          
```

Edit: Actually, it might auto detect llvm-config now, installing it might be all we need to do.

thedataking: There is an existing provisioning step called  `Provision Debian Packages` in `internal-testsuite.yml` but you are adding a separate step in front of it with some duplication (e.g. `sudo apt-get -qq update`).

It seems simplest to me to just test clang-22 on macOS. 

Otherwise, you should cleanly update the existing provision Debian packages step. As I said elsewhere, I think substantial CI changes should be done on a separate PR.

c2rust-ast-exporter/src/AstExporter.cpp
Crocodoctopus: It was fixed, and then at some point unfixed. Its fixed again.

.github/workflows/ci.yml
Crocodoctopus: All Ubuntu/clang22 is removed from CI. Added Macos/clang22 to test matrix.

.github/workflows/internal-testsuite.yml
Crocodoctopus: Yes, but this code no longer exists.

ahomescu (approved):

ahomescu:
LGTM, if it looks good to @thedataking we can land it.

thedataking (approved):
LGTM 🚀

ahomescu:
@Crocodoctopus It needs a rebase, then feel free to merge this.

ahomescu:
@Crocodoctopus can you do a rebase instead of merge?

ahomescu (changes requested):
See previous comment about rebase.

Crocodoctopus:
Should be good to go now.

ahomescu (approved):

PR #1679, opened 2026-03-20
transpile: Compute and store `override_ty`s ahead of time

### LINKED ISSUE #1321: transpile: const macros aren't translated with portable types

Rua:
- Depends on #1757
- Depends on #1756
- Fixes #1321

Before proceeding to remove `override_ty`, I added a temporary assert to compare the value passed to `convert_expr` with the value being generated by the new method. The assert eventually passed in the full CI suite, which likely means they match perfectly.

There was a comment left here concerning this issue:
https://github.com/immunant/c2rust/blob/4887dcef9a5dcaf0842269d2b82d00bd12c0d171/c2rust-transpile/src/translator/mod.rs#L4198

I don't think the cast can be removed, though. In the snapshot test, the `LITERAL_FLOAT` and `NESTED_FLOAT` macros switched type from `double` to `float`, which removed some casts to `float`, but caused new casts to `double` to be added elsewhere.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1383,6 +1383,18 @@ impl TypedAstContext {
             _ => false,
         }
     }
+
+    /// Same as the `Index` trait, but doesn't bypass `Paren`.
+    pub fn index_expr_raw(&self, index: CExprId) -> &CExpr {
+        static BADEXPR: CExpr = Located {
+            loc: None,
+            kind: CExprKind::BadExpr,
+        };
+        match self.c_exprs.get(&index) {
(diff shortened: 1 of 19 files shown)


Rua:
This can now be reviewed. It's a big change, certainly. But since it fixes a major pain point with macro expansion, maybe worth it?

ahomescu (changes requested):
Seems to need a rebase.

ahomescu (changes requested):
Needs a rebase.

PR #1681, opened 2026-03-21
transpile: Add `--no-split-library` option

### LINKED ISSUE #1675: option to produce single-target binary crates
Currently, when c2rust-transpile is run with `--binary main_cu`, it generates a Cargo crate with two targets, a lib target and a bin target.  It puts the main compilation unit `main_cu` in the bin target and puts all other compilation units in the lib target.  This means there are two separate targets, both of which contain non-trivial code.

It would be convenient for CRISP if there was only one target to worry about.  One option would be to put all compilation units into the bin target, and remove the lib target entirely.  Another option would be to put all compilation units into the lib target, and have the bin target consist only of a trivial stub that calls `lib::main()`.

ahomescu:
Implement the new `--no-split-library` option that merges the emitted static library into every binary, writing a top-level `c2rust-bin-{name}.rs` file for each of them. Each top-level file includes the entire submodule hierarchy that previously lived in the erased library.

* Fixes #1675.

### DIFF
--- a/c2rust-transpile/src/build_files/mod.rs
+++ b/c2rust-transpile/src/build_files/mod.rs
@@ -69,7 +69,8 @@ pub struct CrateConfig<'lcmd> {
 /// Emit `Cargo.toml` and `lib.rs` for a library or `main.rs` for a binary.
 /// Returns a single-element vector with the path to `lib.rs` or `main.rs`
 /// (or `[]` if the output file existed already). This may return multiple
-/// elements if there are multiple binaries in the current session.
+/// elements if there are multiple binaries in the current session, e.g.,
+/// if the `--no-split-library` command line flag is present.
 pub fn emit_build_files<'lcmd>(
     tcfg: &TranspilerConfig,
     build_dir: &Path,
@@ -102,16 +103,35 @@ pub fn emit_build_files<'lcmd>(
     };
 
     emit_build_rs(tcfg, &reg, build_dir, ccfg.link_cmd);
-    emit_lib_rs(
-        tcfg,
-        &reg,
-        build_dir,
-        ccfg.modules,
-        &ccfg.pragmas,
-        &ccfg.crates,
-    )
-    .into_iter()
-    .collect()
+    if tcfg.no_split_library {
+        tcfg.binaries
+            .iter()
+            .flat_map(|binary| {
+                emit_lib_rs(
+                    tcfg,
+                    &reg,
+                    build_dir,
+                    ccfg.modules.clone(),
+                    &ccfg.pragmas,
+                    &ccfg.crates,
+                    Some(binary),
+                )
+                .into_iter()
+            })
+            .collect()
+    } else {
+        emit_lib_rs(
+            tcfg,
+            &reg,
+            build_dir,
+            ccfg.modules.clone(),
+            &ccfg.pragmas,
+            &ccfg.crates,
+            None,
+        )
+        .into_iter()
+        .collect()
+    }
 }
 
 #[derive(Serialize)]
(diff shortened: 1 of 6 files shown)


spernsteiner:
I tested this on P01 and it worked—with a few minor hacks to Cargo.toml, it produces both a binary and a library that pass their respective test cases.  I've only looked at the output, though—someone more familiar with c2rust-transpile internals should review the implementation.

c2rust/src/bin/c2rust-transpile.rs
```diff
@@ -191,6 +191,11 @@ struct Args {
 
     #[clap(long, value_enum, default_value_t)]
     cross_check_backend: CrossCheckBackend,
+
+    /// Whether to emit a separate library from the binaries (the default),
+    /// or to fold the library into each binary module.
```
kkysen: Does folding mean duplicating?

c2rust-transpile/src/build_files/mod.rs
kkysen: Can you update these docs?

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -102,8 +102,10 @@ pub fn emit_build_files<'lcmd>(
             ccfg.modules,
             ccfg.pragmas,
             &ccfg.crates,
-        )
-    })
+        ).into_iter().collect()
+    } else {
```
kkysen: An early return might be easier to read here.

c2rust-transpile/src/lib.rs
```diff
@@ -460,12 +459,11 @@ pub fn transpile(tcfg: TranspilerConfig, cc_db: &Path, extra_clang_args: &[&str]
     }
 
     if tcfg.emit_build_files {
-        let crate_file =
+        let crate_files =
             emit_build_files(&tcfg, &build_dir, top_level_ccfg, Some(workspace_members));
-        let crate_file = crate_file.as_deref();
```
kkysen, line 463: You could just replace this line with `let crate_files = crate_files.as_slice()`, more similar to what was there before.

c2rust-transpile/src/lib.rs
```diff
@@ -605,20 +603,24 @@ fn invoke_postprocess(crate_file: &Path, build_dir: &Path) -> Result<(), Error>
...
-        .arg(crate_file)
-        .current_dir(build_dir);
-    let status =  cmd
-        .status()
-        .map_err(|e| {
-            let path = subcommand_path.display();
-            failure::format_err!("unable to run {path}: {e}\nNote that {subcommand} must be installed separately from c2rust and c2rust-transpile.\
-            It must be either installed in $PATH or c2rust/c2rust-transpile must be in `target/$profile/` from the c2rust repo.")
-        })?;
+    // TODO: is it safe to run postprocess over multiple binaries
```
kkysen, line 612: I think for current postprocess commands like comment insertion, it'll just skip inserting comments if they've already been inserted.

c2rust-transpile/src/lib.rs
kkysen, line 560: How is the refactorer supposed to work with multiple binaries and no library?

ahomescu: Badly, I think. I haven't tried it yet, but if the multiple binaries have overlapping modules then the refactorer might get thrown off.

c2rust/src/bin/c2rust-transpile.rs
ahomescu: Yes, every binary now imports every module in the library. This could be a concern for our other tools.

spernsteiner:
I mistakenly ran `c2rust-transpile --no-split-library` with no `--binary` option, and instead of reporting an error, it produced a Cargo.toml that doesn't define any build targets (no `[lib]` or `[[bin]]` sections).  It also didn't produce a top-level `lib.rs` for the project.

ahomescu:
We landed `--thin-binaries` instead, closing this for now.

PR #1686, opened 2026-03-24
transpile: Swap cast and negation for integer literals

### LINKED ISSUE #623: Feature request: Don't wrap the numeric part of negative literals in parentheses
The transpiler currently translates negative numeric literals into two separate concepts - the operator and the positive number, which causes redundant parentheses:

```c
-1
```

to

```rust
-(1 as libc::c_int)
```

This is a bit confusing in the code (it makes it more difficult to refactor with search/replace).

Since unary operators and `as` have higher priority than other operators, I think the translation:

```rust
-1 as libc:c_int
```

shouldn't break the code (I wonder though, which are the edge cases where this doesn't apply).

Interestingly, this indirectly breaks Clippy! When the `clippy::unnecessary_cast` is applied to `-(1)`, Clippy erroneously converts the expressions to `-(1)_u16` and similar (just a curiosity, not C2rust's fault :smile:).

Rua:
- Fixes #623.

The issue was closed long ago because it was deemed to be not worth the effort. I decided to make the effort anyway because it looks nicer. :)

### DIFF
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -14,15 +14,21 @@ impl<'c> Translation<'c> {
         ty: CQualTypeId,
         val: u64,
         base: IntBase,
+        negative: bool,
     ) -> TranslationResult<Box<Expr>> {
         let lit = match base {
             IntBase::Dec => mk().int_unsuffixed_lit(val),
             IntBase::Hex => mk().float_unsuffixed_lit(&format!("0x{:x}", val)),
             IntBase::Oct => mk().float_unsuffixed_lit(&format!("0o{:o}", val)),
         };
+        let mut expr = mk().lit_expr(lit);
+
+        if negative {
+            expr = mk().unary_expr(UnOp::Neg(Default::default()), expr);
+        }
 
         let target_ty = self.convert_type(ty.ctype)?;
-        Ok(mk().cast_expr(mk().lit_expr(lit), target_ty))
+        Ok(mk().cast_expr(expr, target_ty))
     }
 
     /// Return whether the literal can be directly translated as this type.
@@ -49,7 +55,9 @@ impl<'c> Translation<'c> {
         lit: &CLiteral,
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         match *lit {
-            CLiteral::Integer(val, base) => Ok(WithStmts::new_val(self.mk_int_lit(ty, val, base)?)),
+            CLiteral::Integer(val, base) => {
+                Ok(WithStmts::new_val(self.mk_int_lit(ty, val, base, false)?))
+            }
 
             CLiteral::Character(val) => {
                 let val = val as u32;
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -24,7 +24,7 @@ impl<'c> Translation<'c> {
         let mut expr = mk().lit_expr(lit);
 
         if negative {
-            expr = mk().unary_expr(UnOp::Neg(Default::default()), expr);
+            expr = neg_expr(expr);
         }
 
         let target_ty = self.convert_type(ty.ctype)?;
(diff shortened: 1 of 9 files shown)


c2rust-transpile/src/translator/operators.rs
```diff
@@ -934,4 +915,38 @@ impl<'c> Translation<'c> {
...
+            .ast_context
+            .resolve_type(cqual_type.ctype)
+            .kind
+            .is_unsigned_integral_type();
+
+        if let (&CExprKind::Literal(_, CLiteral::Integer(val, base)), false) =
+            (&self.ast_context[arg].kind, is_unsigned_integral_type)
+        {
+            // If we are negating a literal, generate a negated literal directly.
+            // This will create an expression like `-1 as ty` without parentheses,
```
ahomescu: These are not equivalent: e.g. `-(128 as i8)` is a compile error, while `-128 as i8` is perfectly valid. I'm not sure which one we want yet, but this PR changes semantics.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -2,14 +2,6 @@
 
 use super::*;
 
-fn neg_expr(arg: Box<Expr>) -> Box<Expr> {
-    mk().unary_expr(UnOp::Neg(Default::default()), arg)
-}
-
-fn wrapping_neg_expr(arg: Box<Expr>) -> Box<Expr> {
-    mk().method_call_expr(arg, "wrapping_neg", vec![])
-}
```
kkysen, line 11: I think keeping these separate is still useful.  They're simple and clear.  The other changes in this commit are good, though.  These could also be moved to `Builder` methods perhaps.

c2rust-transpile/src/translator/literals.rs
```diff
@@ -14,15 +14,21 @@ impl<'c> Translation<'c> {
...
     ) -> TranslationResult<Box<Expr>> {
         let lit = match base {
             IntBase::Dec => mk().int_unsuffixed_lit(val),
             IntBase::Hex => mk().float_unsuffixed_lit(&format!("0x{:x}", val)),
             IntBase::Oct => mk().float_unsuffixed_lit(&format!("0o{:o}", val)),
         };
+        let mut expr = mk().lit_expr(lit);
+
+        if negative {
+            expr = mk().unary_expr(UnOp::Neg(Default::default()), expr);
```
kkysen: This could use `fn neg_expr`, for example.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -927,15 +927,26 @@ impl<'c> Translation<'c> {
...
-            if is_unsigned_integral_type {
-                mk().method_call_expr(val, "wrapping_neg", vec![])
-            } else {
-                mk().unary_expr(UnOp::Neg(Default::default()), val)
-            }
-        });
 
-        Ok(val)
+        if let (&CExprKind::Literal(_, CLiteral::Integer(val, base)), false) =
+            (&self.ast_context[arg_id].kind, is_unsigned_integral_type)
```
kkysen, line 928: Can you use `if let &CExprKind::Literal(_, CLiteral::Integer(val, base)) = &self.ast_context[arg_id].kind && is_unsigned_integral_type` here yet?  Or is it not stabilized yet?  It'd be a lot clearer.

kkysen: Also, `let false = is_unsigned_integral_type` just means it's not an unsigned type, right?  Don't we also need to check the value of the literal to see if it's negative?

Rua: > Or is it not stabilized yet? It'd be a lot clearer.

`if let` chains are only stabilised in the 2024 edition sadly.

> Also, `let false = is_unsigned_integral_type` just means it's not an unsigned type, right? Don't we also need to check the value of the literal to see if it's negative?

`val` is unsigned, so the negativeness is guaranteed by the fact that we're in a negation operator already.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -927,17 +919,31 @@ impl<'c> Translation<'c> {
...
-        if resolved_ctype.kind.is_unsigned_integral_type() {
-            Ok(val.map(wrapping_neg_expr))
-        } else {
-            Ok(val.map(neg_expr))
-        }
+        Ok(val)
     }
 }
+
+fn neg_expr(arg: Box<Expr>) -> Box<Expr> {
```
kkysen: Is there a reason to move these down here?

Rua: My usual practice is to put helper functions below the function that they "help". Just makes more sense to me that way, to have the important things first. Do you want me to move it back?

kkysen: Ah, I usually do the opposite.  I don't have a strong preference, but if it's one way to begin with, I don't think it's worth changing, so could you move them back?  Thanks!  They might also make sense in `c2rust-ast-builder` if you prefer that, though, although I don't have a strong preference for that either.

c2rust-transpile/src/translator/literals.rs
kkysen: Can we do this?

kkysen (approved):

PR #1687, opened 2026-03-24
transpile: Use `f128` for `*l` float builtins

### LINKED ISSUE #1684: C2Rust translates `__builtin_huge_vall()` to `f64::INFINITY` instead of `f128` infinity
C2Rust incorrectly transpiles `__builtin_huge_vall()`.
In C, `__builtin_huge_vall()` produces a `long double` infinity value.
However, C2Rust translates it to `::core::f64::INFINITY` even when the destination type is `::f128::f128`, which causes a Rust type mismatch.
**Reproducer**
```c
int main(void) {
       long double hugel = __builtin_huge_vall();
       return 0;
}
```
**Transpiled Rust**
```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
    let mut hugel: ::f128::f128 = ::core::f64::INFINITY;
}
pub fn main() {
}
```
**Observed Behavior**
The generated Rust code fails to compile due to a type mismatch:
```
  expected struct `f128`, found `f64`
```
**Environment**
- c2rust version: v0.21.0
- platform: Ubuntu 24.04
- Rust: nightly-2022-08-08
- Clang version: 17.0.6

Rua:
- Fixes #1684.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -75,15 +75,29 @@ impl<'c> Translation<'c> {
             "__builtin_huge_valf" => Ok(WithStmts::new_val(
                 mk().abs_path_expr(vec!["core", "f32", "INFINITY"]),
             )),
-            "__builtin_huge_val" | "__builtin_huge_vall" => Ok(WithStmts::new_val(
+            "__builtin_huge_val" => Ok(WithStmts::new_val(
                 mk().abs_path_expr(vec!["core", "f64", "INFINITY"]),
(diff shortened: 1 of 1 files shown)


ahomescu (approved):

kkysen (commented):
One little nit: I was confused what `l` meant; I think `*l` is clearer for this, so I changed the PR title, if you want to update the commit message, too.

kkysen (approved):

PR #1689, opened 2026-03-25
transpile: Pass correct types when translating prefix inc/dec operators

### LINKED ISSUE #1237: Incorrect Use of `.offset_from(1)` When Transpiling Pointer Decrement on Struct Field
## Description

C2Rust incorrectly translates valid C code containing a pointer decrement on a volatile struct field, such as` --(global.p);`, into Rust using `.offset_from(1)`, which is invalid and causes a type mismatch at compile time.

## C code:

```c
int arr[10];
struct S { int * volatile p; } global;
int func(int *restrict r) {
    --(global.p);
    return *r;
}

int main() {
     return 0;
}
```

## Translated Rust Code

```rust
#[derive(Copy, Clone)]
#[repr(C)]
pub struct S {
    pub p: *mut libc::c_int,
}
#[no_mangle]
pub static mut arr: [libc::c_int; 10] = [0; 10];
#[no_mangle]
pub static mut global: S = S {
    p: 0 as *const libc::c_int as *mut libc::c_int,
};
#[no_mangle]
pub unsafe extern "C" fn func(mut r: *mut libc::c_int) -> libc::c_int {
    ::core::ptr::write_volatile(
        &mut global.p as *mut *mut libc::c_int,
        (::core::ptr::read_volatile::<
            *mut libc::c_int,
        >(&global.p as *const *mut libc::c_int))
            .offset_from(1) as *mut libc::c_int,
    );
    ::core::ptr::read_volatile::<*mut libc::c_int>(&global.p as *const *mut libc::c_int);
    return *r;
}
```

## Cargo build error:

```text
  --> src/runner.rs:29:26
   |
29 |             .offset_from(1) as *mut libc::c_int,
   |              ----------- ^ expected `*const i32`, found `usize`
   |              |
   |              arguments to this method are incorrect
   |
   = note: expected raw pointer `*const i32`
                     found type `usize`
note: method defined here
  --> /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ptr/mut_ptr.rs:856:25
```

c2rust version: `C2Rust 0.20.0`

Rua:
- Fixes #1237.

Firstly, the types being passed were wrong in general, as if the arguments had been switched around. But also, for pointer increment/decrement, the implied RHS is not itself a pointer. So this PR solves that by using `int` as the implied type there instead.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -727,6 +727,12 @@ impl<'c> Translation<'c> {
         } else {
             c_ast::BinOp::AssignSubtract
         };
+
+        let arg_type = self.ast_context[arg]
+            .kind
+            .get_qual_type()
+            .ok_or_else(|| format_err!("bad arg type"))?;
+
         let one = match self.ast_context.resolve_type(ty.ctype).kind {
             // TODO: If rust gets f16 support:
             // CTypeKind::Half |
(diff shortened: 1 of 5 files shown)


kkysen (commented):
Can we also add a test for this, too?  Thanks!

Rua:
I added a test, but it's failing because of something unrelated. Possibly a bug in the snapshot test suite?

c2rust-transpile/tests/snapshots/snapshots__transpile@exprs.c.2024.snap
```diff
@@ -74,3 +79,19 @@ pub unsafe extern "C" fn statement_expr() {
...
 }
+#[unsafe(no_mangle)]
+pub static mut volatile_global_struct: S = S {
+    p: ::core::ptr::null::<::core::ffi::c_int>() as *mut ::core::ffi::c_int,
+};
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn decrement_volatile() {
+    unsafe {
+        ::core::ptr::write_volatile(
+            &mut volatile_global_struct.p as *mut *mut ::core::ffi::c_int,
```
kkysen: We need to be using `&raw mut` now that it's edition 2024.  We should open an issue for this.  In the meantime, you can mark this test as a compile error for edition 2024 only.

kkysen: Also, could you put the snapshot test before the other commit and then have it updated in that commit?  Thanks!

kkysen (approved):

Issue #288, opened 2020-09-03
`false` is not translated well


jrmuizel:
```c
#include <stdbool.h>
bool make_false() {
    return false;
}
```
translates to:
```rust
pub unsafe extern "C" fn make_false() -> bool {
    return 0 as libc::c_int != 0;
}
```

PR #1693, opened 2026-03-25
transpile: Translate some literals in conditionals as `bool` directly

### LINKED ISSUE #190: `0i32 != 0` vs `false`
I'm looking through the c2rust translation of a bit of C code and I'm seeing statements like 
```c
bool nonempty = false;
```
translated into
```rust
let mut nonempty: bool = 0i32 != 0;
```
which seems like a pretty roundabout way to simply say `false`.

Rua:
- Depends on #1647.
- Fixes #190.

Makes things look a little nicer. I've tried to be conservative here and only consider literal `0` and `1` this way. No other literals, and no macros either.

Unfortunately that includes the `false` and `true` macros found in the `stdbool.h` header. There would need to be some way to detect those and treat them differently from other macros. See also #288.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2634,6 +2634,16 @@ impl<'c> Translation<'c> {
                 null_pointer_case(ptr, !target)
             }
 
+            CExprKind::Literal(_, ref literal @ CLiteral::Integer(0 | 1, _))
+                if !self.expr_is_expanded_macro(ctx, cond_id, None) =>
+            {
+                // If there is a literal `0` or `1` here, translate them directly rather than
+                // with a comparison. But not if they're inside a macro; we want to keep that.
+                // TODO: What about the `false` and `true` macros in stdbool.h?
+                let val = mk().lit_expr(mk().bool_lit(target == literal.get_bool()));
+                Ok(WithStmts::new_val(val))
+            }
+
             CExprKind::Unary(_, c_ast::UnOp::Not, subexpr_id, _) => {
                 self.convert_condition(ctx, !target, subexpr_id)
             }
--- a/c2rust-transpile/tests/snapshots/bool.c
+++ b/c2rust-transpile/tests/snapshots/bool.c
@@ -1,0 +1,33 @@
+#include <stdbool.h>
+
+#define MACRO_INT0 0
+#define MACRO_IDENT0 false
+
+#define MACRO_INT1 1
+#define MACRO_IDENT1 true
+
+void test_bool() {
+    bool int0 = 0;    
+    bool ident0 = false;
+    bool macro_int0 = MACRO_INT0;
+    bool macro_ident0 = MACRO_IDENT0;
+
+    bool int1 = 1;
+    bool ident1 = true;
+    bool macro_int1 = MACRO_INT1;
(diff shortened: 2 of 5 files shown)


kkysen (commented):
I merged #1647, so you can rebase now.  Also, can you add a snapshot test for this (add it before the changes, and then the commit changing the behavior updates the snapshot so I can see the diff)?  Thanks!

c2rust-transpile/src/translator/macros.rs
```diff
@@ -203,16 +204,22 @@ impl<'c> Translation<'c> {
...
+            let source_cty = CQualTypeId::new(macro_ty);
+            let target_cty = expr_ty;
+
+            let source_ty_kind = &self.ast_context.resolve_type(source_cty.ctype).kind;
+            let target_ty_kind = &self.ast_context.resolve_type(target_cty.ctype).kind;
+            let kind = CastKind::from_types(source_ty_kind, target_ty_kind).unwrap_or_else(|| {
+                warn!(
+                    "Unknown CastKind for {:?} to {:?} cast. Defaulting to BitCast",
+                    source_ty_kind, target_ty_kind,
+                );
```
kkysen:
```suggestion
                warn!("Unknown CastKind for {source_ty_kind:?} to {target_ty_kind:?} cast. Defaulting to BitCast");
```

For a follow-up commit.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2634,6 +2634,16 @@ impl<'c> Translation<'c> {
                 null_pointer_case(ptr, !target)
             }
 
+            CExprKind::Literal(_, ref literal @ CLiteral::Integer(0 | 1, _))
+                if !self.expr_is_expanded_macro(ctx, cond_id, None) =>
+            {
+                // If there is a literal `0` or `1` here, translate them directly rather than
+                // with a comparison. But not if they're inside a macro, we want to keep that.
```
```suggestion
                // with a comparison. But not if they're inside a macro; we want to keep that.
```

c2rust-transpile/src/translator/mod.rs
```diff
@@ -2634,6 +2634,16 @@ impl<'c> Translation<'c> {
...
+                // TODO: What about the `false` and `true` macros in stdbool.h?
```
kkysen, line 2642: The `true` and `false` macros turn into `true_0` and `false_0` `const`s, although not with the `bool` type.  So the translation is decent enough already I think.  Not sure exactly what to do here.

Rua: The macros could be given special treatment elsewhere, so that they get translated directly as `bool` rather than as `int`. But that's more involved.

c2rust-transpile/tests/snapshots/bool.c
```diff
@@ -0,0 +1,33 @@
+#include <stdbool.h>
+
+#define MACRO_INT0 0
+#define MACRO_IDENT0 false
+
+#define MACRO_INT1 1
+#define MACRO_IDENT1 1
```
kkysen: Is this meant to be `true`?

Rua: Oops, yes!

kkysen (approved):

PR #1695, opened 2026-03-25
transpile: Import `num_traits::Float` for method calls on `f128`

### LINKED ISSUE #1690: C2Rust generates calls to `f128` float methods without importing required traits
C2Rust generates Rust code that calls floating-point methods on `f128` (e.g., `is_infinite`, `is_sign_positive`) without importing the traits that provide these methods.
As a result, the generated Rust code fails to compile unless the user manually adds the required trait import.
**Reproducer**
```c
#include <stdio.h>
#include <math.h>

int main(void) {
    long double huge = __builtin_huge_val();
    printf("huge_val: Input=(none), Result=%s\n",
           isinf(huge) ? "INF" : "NOT_INF");
    return 0;
}
```
**Transpiled Rust**
```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
extern "C" {
}
    let mut huge: ::f128::f128 = ::f128::f128::new(::core::f64::INFINITY);
    printf(
        b"huge_val: Input=(none), Result=%s\n\0" as *const u8
            as *const ::core::ffi::c_char,
        if if huge.is_infinite() {
            if huge.is_sign_positive() { 1 } else { -1 }
        } else {
            0
        } != 0
        {
            b"INF\0" as *const u8 as *const ::core::ffi::c_char
        } else {
            b"NOT_INF\0" as *const u8 as *const ::core::ffi::c_char
        },
    );
}
pub fn main() {
}
```
**Observed Behavior**
The generated Rust code fails to compile:
```
error[E0599]: no method named `is_infinite` found for struct `f128`
error[E0599]: no method named `is_sign_positive` found for struct `f128`
```
The compiler indicates that these methods are provided by traits such as:
```
num_traits::float::Float
```
but the generated code does not import them.
**Expected Behavior**
C2Rust should generate the necessary trait imports when emitting calls to trait-provided floating-point methods on `f128`.
**Environment**
- c2rust version: v0.21.0
- platform: Ubuntu 24.04
- Rust: nightly-2022-08-08
- Clang version: 17.0.6

Rua:
- Fixes #1690.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -112,15 +112,9 @@ impl<'c> Translation<'c> {
                 ))
             }
             "__builtin_signbit" | "__builtin_signbitf" | "__builtin_signbitl" => {
-                // Long doubles require the Float trait from num_traits to call this method
-                if builtin_name == "__builtin_signbitl" {
-                    self.with_cur_file_item_store(|item_store| {
-                        item_store.add_use(true, vec!["num_traits".into()], "Float");
-                    });
-                }
+                self.import_num_traits(args[0])?;
 
                 let val = self.convert_expr(ctx.used(), args[0], None)?;
-
                 Ok(val.map(|v| {
                     let val = mk().method_call_expr(v, "is_sign_negative", vec![]);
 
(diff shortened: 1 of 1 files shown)


c2rust-transpile/src/translator/builtins.rs
```diff
@@ -723,6 +723,26 @@ impl<'c> Translation<'c> {
...
+    fn import_num_traits(&self, arg_id: CExprId) -> TranslationResult<()> {
+        let arg_type_id = self.ast_context[arg_id]
+            .kind
+            .get_qual_type()
+            .ok_or_else(|| format_err!("bad arg type"))?;
+        let arg_type_kind = &self.ast_context.resolve_type(arg_type_id.ctype).kind;
+
+        match arg_type_kind {
+            CTypeKind::LongDouble | CTypeKind::Float128 => {
+                self.use_crate(ExternCrate::NumTraits);
```
ahomescu, line 749: Was this not needed before because there's a cast after? Or how did this work before?

ahomescu (approved):

c2rust-transpile/src/translator/builtins.rs
Rua: I'm not sure it ever worked. These methods are not inherent methods of `f128`, they are provided by the `Float` trait. So without importing that, you just get a compile error.

kkysen: I don't think we ever tested `f128` stuff much before, so that was probably the case.

kkysen (approved):

PR #1696, opened 2026-03-25
transpile: Use `f128::f128::ZERO` constructor in conditions

### LINKED ISSUE #1692: Incorrect translation of `long double` → `bool` causes Rust type mismatch
When translating a cast from `long double` to `bool`, the transpiler generates a comparison against an untyped floating-point literal (`0.`).
 However, the `long double` value is translated to `::f128::f128`, which results in a **type mismatch (`f128` vs default float literal)** and causes the generated Rust code to fail compilation.
**Reproducer**
```c
#include <stdbool.h>

int main(void) {
    long double ld_zero = 0.0L;
    long double ld_non_zero = 0.1L;

    bool ld_zero_as_bool = (bool)ld_zero;
    bool ld_non_zero_as_bool = (bool)ld_non_zero;
}
```
**Generated Rust**
```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
    let mut ld_zero: ::f128::f128 = ::f128::f128::new(0.0);
    let mut ld_non_zero: ::f128::f128 = ::f128::f128::new(0.1);
    let mut ld_zero_as_bool: bool = ld_zero != 0.;
    let mut ld_non_zero_as_bool: bool = ld_non_zero != 0.;
    return 0;
}
pub fn main() {
}
```
**Observed Behavior**
The generated Rust code fails to compile:
```
  --> src/main.rs:12:48
   |
12 |     let mut ld_zero_as_bool: bool = ld_zero != 0.;
   |                                                ^^ expected struct `f128`, found floating-point number

  --> src/main.rs:13:56
   |
13 |     let mut ld_non_zero_as_bool: bool = ld_non_zero != 0.;
   |                                                        ^^ expected struct `f128`, found floating-point number
```
**Environment**
- c2rust version: v0.21.0
- platform: Ubuntu 24.04
- Rust: nightly-2022-08-08
- Clang version: 17.0.6

Rua:
- Fixes #1692.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4596,7 +4596,10 @@ impl<'c> Translation<'c> {
             };
 
             // The backup is to just compare against zero
-            let zero = if ty.is_floating_type() {
+            let zero = if let CTypeKind::LongDouble | CTypeKind::Float128 = ty {
+                self.use_crate(ExternCrate::F128);
+                mk().abs_path_expr(vec!["f128", "f128", "ZERO"])
+            } else if ty.is_floating_type() {
                 mk().lit_expr(mk().float_unsuffixed_lit("0."))
             } else {
                 mk().lit_expr(mk().int_unsuffixed_lit(0))


ahomescu (approved):

kkysen (approved):

kkysen:
Could we also add a snapshot test for this?

Rua:
I've now made a separate PR to add tests for all the PRs affecting `f128`.

PR #1697, opened 2026-03-25
transpile: Handle `InitList` correctly for all scalar types

### LINKED ISSUE #1691: C2Rust fails to translate valid scalar initializer list for `double` (`Init list not implemented for Double`)
C2Rust fails to transpile a valid C program that initializes a `double` using a braced scalar initializer.
In C, scalar initialization with braces is valid, and the following program should be accepted.
**Reproducer**
```c
int main(void) {
    double x = { 0.0 };
    return 0;
}
```
**Observed Behavior**
C2Rust fails during translation with:
```
error: Failed to translate main: Init list not implemented for Double
```
It also still emits an incomplete Rust file:
```rust
pub fn main() {
}
```
where `main_0` is missing.
**Environment**
- c2rust version: v0.21.0
- platform: Ubuntu 24.04
- Rust: nightly-2022-08-08
- Clang version: 17.0.6

Rua:
- Fixes #1691.

### DIFF
--- a/c2rust-transpile/tests/snapshots/scalar_init.c
+++ b/c2rust-transpile/tests/snapshots/scalar_init.c
@@ -1,0 +1,19 @@
+#include <stdbool.h>
+
+enum E {
+    A,
+};
+
+void scalar_init() {
+    bool b = { true };
+    char c = { 'c' };
+    int i = { 42 };
+    long l = { 42l };
+    float f = { 42.0f };
+    double d = { 42.0 };
+    long double ld = { 42.0l };
+    // Doesn't work on MacOS, so avoid it for now so this test doesn't become OS-specific.
+    // __float128 q = { 42.0q };
+    enum E e = { A };
+    int *p = { &i };
+}
--- a/c2rust-transpile/tests/snapshots/scalar_init.c
+++ b/c2rust-transpile/tests/snapshots/scalar_init.c
@@ -1,4 +1,5 @@
 #include <stdbool.h>
+#include <stdlib.h>
 
 enum E {
     A,
@@ -16,4 +17,29 @@ void scalar_init() {
     // __float128 q = { 42.0q };
     enum E e = { A };
     int *p = { &i };
+
+// Empty initializer lists for scalars was only added in C23 and clang 17.
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L) || (defined(__clang__) && (__clang_major__ >= 17))
+    bool eb = {};
+    char ec = {};
+    int ei = {};
+    long el = {};
+    float ef = {};
+    double ed = {};
+    long double eld = {};
+    // __float128 eq = {};
+    // enums don't allow empty init lists in C
+    int *ep = {};
+#else
+    bool eb;
+    char ec;
+    int ei;
+    long el;
+    float ef;
+    double ed;
+    long double eld;
+    // __float128 eq;
+    // enums don't allow empty init lists in C
+    int *ep;
+#endif
 }
--- a/c2rust-rust-tools/src/lib.rs
+++ b/c2rust-rust-tools/src/lib.rs
@@ -1,5 +1,6 @@
 use itertools::Itertools;
 use log::warn;
+use std::collections::HashSet;
 use std::fmt;
 use std::fmt::Display;
 use std::fmt::Formatter;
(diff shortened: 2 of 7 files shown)


ahomescu (approved):

kkysen (approved):
LGTM, too, but could we also add a snapshot test here?

Rua:
I want to include `f128` in the init list tests, but that will not work until #1688 has merged. So I'll wait.

kkysen (commented):
> I want to include `f128` in the init list tests, but that will not work until #1688 has merged. So I'll wait.

#1688 has been merged now.  And can you do the same before and after snapshot test thing?

Rua:
I don't think that will work here. Init lists for many of these types were simply not implemented before this PR, so it would just not transpile.

kkysen:
> I don't think that will work here. Init lists for many of these types were simply not implemented before this PR, so it would just not transpile.

Ah, okay, then never mind then.  Simple snapshot test after will be good.

c2rust-transpile/tests/snapshots/scalar_init.c
kkysen, line 1: I think you forgot to add a `#[test]` in `c2rust-transpile/tests/snapshots.rs` to actually run this.  How were you generating the snapshots then?

kkysen: I just forced push this (adding the `#[test]`) and it seems like there's an error:

```sh
    thread 'test_scalar_init' panicked at 'called `Option::unwrap()` on a `None` value', c2rust-transpile/src/translator/literals.rs:260:38
```

kkysen (changes requested):

Rua:
It looks like this is now running into the same problem as #1716, so I'll wait for #1719 to merge. Also, `__float128` apparently just isn't available on MacOS? How come `f128.c` seems to be passing all these tests?

Edit: Ah, that's why... see #1727.

c2rust-transpile/tests/snapshots/scalar_init.c
```diff
@@ -0,0 +1,19 @@
...
+    char c = { 'c' };
+    int i = { 42 };
+    long l = { 42l };
+    float f = { 42.0f };
+    double d = { 42.0 };
+    long double ld = { 42.0l };
+    // Doesn't work on MacOS, so avoid it for now so this test doesn't become OS-specific.
+    // __float128 q = { 42.0q };
+    enum E e = { A };
+    int *p = { &i };
```
kkysen, line 19: You were previously testing a bunch of empty inits

```c
    char ec = {};
    int ei = {};
    long el = {};
    float ef = {};
    double ed = {};
    long double eld = {};
    __float128 eq = {};
    enum E ee = {};
    int *ep = {};
```

Do those work?  That's what the change from `ids.first().unwrap()` to `if let Some(&first) = ids.first()` is supposed to handle, right?

If so, could you add these back in a separate commit where you also move the change from `ids.first().unwrap()` to `if let Some(&first) = ids.first()`?  Thanks!

kkysen (changes requested):

c2rust-transpile/tests/snapshots/scalar_init.c
Rua: I thought they worked, but apparently C disallows this for scalars. So it was giving compile errors. It only default initialises for more complex types. I left out the `unwrap` though because the panic was still happening even with the compile error happening on the C side, and obscuring the real reason for the error.

kkysen: It seems to only disallow it for `enum`s from what I can tell: https://godbolt.org/z/TPcG9d6xv

Rua: Ah, I'll add those tests back then.

kkysen: Could you also do that split commit that I suggested?  Also, I think a default-initialized `bool` test is missing, causing the current snapshot failure.

Rua: Huh, it's still giving errors with the empty initialisers, but only for Clang 15?

kkysen (commented):
It seems that empty initializer list initialization of scalars was only added in C23, which Clang 15 doesn't support.  Let me think of a solution.

Rua:
I think you had the right idea? I changed it a bit and now it works.

kkysen:
@Rua, this one got closed, too.  Could you reopen it?  We can merge both of them now.

PR #1703, opened 2026-03-28
transpile: Use `convert_cast` in `operators.rs`

### LINKED ISSUE #340: `|=` operator on `bool` results in invalid Rust being generated
I tried running c2rust on [BRRtools](https://github.com/Optiroc/BRRtools) as an example of a small but nontrivial C project. However the generated Rust fails to compile. The input C in question is:

https://github.com/Optiroc/BRRtools/blob/9f6c4708150120284102f9899632d7f5f0b0638a/src/brr_encoder.c#L596-L598

I minified this example to the following C code (`x` can be either `bool` or `int`):

```c
#include <stdbool.h>

// both bool and int reproduce the bug
extern bool x;

int main() {
	bool initial_block = false;
	initial_block |= x;
}
```

Running the latest c2rust 030df665741bb9d1b29fe41cdbfd7223a67c5995 produces the following Rust code:

```rust
#![allow(dead_code, mutable_transmutes, non_camel_case_types, non_snake_case,
         non_upper_case_globals, unused_assignments, unused_mut)]
#![register_tool(c2rust)]
#![feature(main, register_tool)]
use ::c2rust_out::*;
extern "C" {
    // both bool and int reproduce the bug
    #[no_mangle]
    static mut x: bool;
}
    let mut initial_block: bool = 0 as libc::c_int != 0;
    initial_block = (initial_block as libc::c_int | x as libc::c_int) as bool;
    return 0;
}
#[main]
pub fn main() { unsafe { ::std::process::exit(main_0() as i32) } }
```

which produces an error when compiling on both `cargo +nightly-2019-12-05` and stable:

```
error[E0054]: cannot cast as `bool`
  --> src/brr_encoder.rs:13:21
   |
13 |     initial_block = (initial_block as libc::c_int | x as libc::c_int) as bool;
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: compare with zero instead: `(initial_block as libc::c_int | x as libc::c_int) != 0`
```

I've uploaded at repo at https://github.com/nyanpasu64/c2rust-bug. Steps to reproduce:

```
c2rust transpile --binary brr_encoder --overwrite-existing compile_commands.json
cargo build
```

### LINKED ISSUE #1701: transpile: Compound assignment with unsigned RHS sometimes casts twice
C code:
```C
enum E { A };

void test(void) {
    int i = 0;
    unsigned int u = 0;
    enum E e = A;
    e += u;
    i += u;
}
```

Transpiled result:
```rust
pub type E = ::core::ffi::c_uint;
pub const A: E = 0;
#[no_mangle]
pub unsafe extern "C" fn test() {
    unsafe {
        let mut i: ::core::ffi::c_int = 0 as ::core::ffi::c_int;
        let mut u: ::core::ffi::c_uint = 0 as ::core::ffi::c_uint;
        let mut e: E = A;
        e = (e as ::core::ffi::c_uint).wrapping_add(u) as E as E;
        i = (i as ::core::ffi::c_uint).wrapping_add(u) as ::core::ffi::c_int
            as ::core::ffi::c_int;
    }
}
```

The cast to `E` and `c_int` is added twice. This seems to occur if both conditions are met:
- The RHS is unsigned.
- The addition triggers automatic promotion of the LHS to the RHS type, requiring a cast back afterwards.

Rua:
- Fixes #1701
- Fixes #340

The various edge cases in these bits of code were essentially duplicating what `convert_cast` was doing, but worse and easier to miss. So this just reuses the more robust logic that `convert_cast` already has for this.

### DIFF
--- a/c2rust-transpile/tests/snapshots/bool.c
+++ b/c2rust-transpile/tests/snapshots/bool.c
@@ -27,7 +27,6 @@ void test_bool() {
     MACRO_INT1 ? 2 : 3;
     MACRO_IDENT1 ? 2 : 3;
 
-    // Doesn't compile currently.
-    // See https://github.com/immunant/c2rust/issues/340.
+    // Tests https://github.com/immunant/c2rust/issues/340
     int0 |= int1;
 }
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -4149,7 +4149,7 @@ impl<'c> Translation<'c> {
         }
     }
 
-    fn convert_cast(
+    pub fn convert_cast(
         &self,
         ctx: ExprContext,
         source_cty: CQualTypeId,
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -206,6 +206,7 @@ impl<'c> Translation<'c> {
         read: Box<Expr>,
         write: Box<Expr>,
         rhs: Box<Expr>,
+        initial_lhs_type_id: CQualTypeId,
         compute_lhs_type_id: CQualTypeId,
         compute_res_type_id: CQualTypeId,
         lhs_type_id: CQualTypeId,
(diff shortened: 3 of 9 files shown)


kkysen (commented):
Can you also add before/after snapshot tests for this, too?  Note that if the before test fails to compile, we can now handle that with a `.expect_compile_error(true)`.

Rua:
I added a test to `bool.c`, which was added in #1693. So this PR now depends on that one.

c2rust-transpile/tests/snapshots/bool.c
```diff
@@ -27,7 +27,6 @@ void test_bool() {
     MACRO_INT1 ? 2 : 3;
     MACRO_IDENT1 ? 2 : 3;
 
-    // Doesn't compile currently.
-    // See https://github.com/immunant/c2rust/issues/340.
+    // https://github.com/immunant/c2rust/issues/340
```
kkysen: Do we still need this?

Rua: I've seen comments like that in a few places throughout the tests, to point out why a particular thing is being tested for.

kkysen: Hmm, maybe write it as `Tests https://github.com/immunant/c2rust/issues/340`?

kkysen (approved):

PR #1705, opened 2026-03-28
transpile: Treat `LValueToRValue` casts as potential volatile reads

### LINKED ISSUE #292: Reads to volatiles are not mapped to read_volatile
The RIOT-OS code base I'm translating as part of the efforts around https://github.com/rust-lang/rust-bindgen/issues/1344 contains a function like

```c
 static inline kernel_pid_t thread_getpid(void)
 {
     extern volatile kernel_pid_t sched_active_pid;
 
     return sched_active_pid;
 }
```

My previous manual translation of this was `unsafe { ::core::ptr::read_volatile(&bindgen_generated_symbols::sched_active_pid) }`. When translating with `c2rust transpile comile_commands.json  --preserve-unused-functions --emit-modules --emit-no-std` (using #291), I get

```rust
#[inline]
pub unsafe extern "C" fn thread_getpid() -> kernel_pid_t {
    extern "C" {
        #[link_name = "sched_active_pid"]
        static mut sched_active_pid_0: kernel_pid_t;
    }
    return sched_active_pid;
}
```

which, in my understanding, allows the compiler to mush together multiple calls in a single function. That's not particularly bad for get_pid (which won't change in code that the Rust compiler will see as contiguous), but other functions will depend on the read to be volatile.

(I'm not fuly sure whether that's a bug in C2Rust, me expecting things it can't know or just the lack of a flag, but after looking through the docs and existing issues, I'm hoping for the former).

Rua:
- Depends on #1717.
- Fixes #292.

Previously, volatile reads were handled in `convert_expr` for variable `DeclRef` directly. But that isn't the right place, since all variable `DeclRef` are lvalues and you can't tell at that point whether it's being read or written. Likewise for dereference expressions, which are still lvalues as a whole.

So I've moved it to where `LValueToRValue` casts are handled, which seems to be more correct. Now `volatile_read` is indeed generated.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3032,20 +3032,15 @@ impl<'c> Translation<'c> {
         lhs_type: CQualTypeId,
         write: bool,
     ) -> TranslationResult<Box<Expr>> {
-        let mutbl = if write {
-            Mutability::Mutable
-        } else {
-            Mutability::Immutable
-        };
         let addr_lhs = match *lhs {
             Expr::Unary(ExprUnary {
                 op: UnOp::Deref(_),
                 expr: e,
                 ..
             }) => {
-                if write == lhs_type.qualifiers.is_const {
+                if write && lhs_type.qualifiers.is_const {
                     let lhs_type = self.convert_type(lhs_type.ctype)?;
-                    let ty = mk().set_mutbl(mutbl).ptr_ty(lhs_type);
+                    let ty = mk().set_mutbl(Mutability::Mutable).ptr_ty(lhs_type);
 
                     mk().cast_expr(e, ty)
                 } else {
@@ -3053,6 +3048,12 @@ impl<'c> Translation<'c> {
                 }
             }
             _ => {
+                let mutbl = if write {
+                    Mutability::Mutable
+                } else {
+                    Mutability::Immutable
+                };
+
                 self.use_feature("raw_ref_op");
                 mk().set_mutbl(mutbl).raw_borrow_expr(lhs)
             }
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3456,12 +3456,6 @@ impl<'c> Translation<'c> {
 
                 let mut val = mk().path_expr(vec![rustname]);
 
-                // If the variable is volatile and used as something that isn't an LValue, this
-                // constitutes a volatile read.
-                if lrvalue.is_rvalue() && qual_ty.qualifiers.is_volatile {
-                    val = self.volatile_read(val, qual_ty)?;
-                }
-
                 // If the variable is actually an `EnumConstant`, we need to add a cast to the
                 // expected integral type.
                 if let &CDeclKind::EnumConstant { .. } = decl {
@@ -3518,7 +3512,7 @@ impl<'c> Translation<'c> {
 
                 // if the context wants a different type, add a cast
                 if let Some(expected_ty) = override_ty {
-                    if expected_ty != qual_ty {
+                    if lrvalue.is_rvalue() && expected_ty != qual_ty {
                         val = mk().cast_expr(val, self.convert_type(expected_ty.ctype)?);
                     }
                 }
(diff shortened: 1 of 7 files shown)


kkysen (commented):
Could you add a before/after snapshot test here, too?  Thanks!

Rua:
Tests added, but this now depends on #1689 for the `volatile.c` test.

EDIT: Depends on #1717 instead because it worked out easier that way.

Rua:
It looks like this is failing because an extra cast is being inserted, before `read_volatile` has a chance to get to it. Casting lvalues is probably a bad idea, so that seems like a bug?

Edit: The casts are now only done when the value is an rvalue. For lvalues, the cast is performed after the `LValueToRValue` cast.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3659,16 +3653,8 @@ impl<'c> Translation<'c> {
                             return self.convert_expr(ctx, expr, override_ty);
                         }
                     }
-                    // LValueToRValue casts don't actually change the type, so it still makes sense
```
Rua, line 3662: This code was removed because it doesn't actually seem to do anything anymore. None of the possible inner lvalue expression types use `override_ty`.

c2rust-transpile/tests/snapshots/snapshots__transpile@volatile.c.2024.snap
```diff
@@ -24,10 +24,11 @@ pub static mut volatile_global_struct: S = S {
 pub unsafe extern "C" fn test_volatile() {
     unsafe {
         let mut vi: ::core::ffi::c_int = 0 as ::core::ffi::c_int;
-        let mut i: ::core::ffi::c_int = vi;
+        let mut i: ::core::ffi::c_int =
+            ::core::ptr::read_volatile::<::core::ffi::c_int>(&raw const vi);
         ::core::ptr::write_volatile(&raw mut vi, i);
         let mut pvi: *mut ::core::ffi::c_int = &raw mut vi;
-        i = *pvi;
+        i = ::core::ptr::read_volatile::<::core::ffi::c_int>(pvi as *const ::core::ffi::c_int);
```
kkysen: Is the `::<::core::ffi::c_int>` needed here?  Since it's not used for `write_volatile`.

c2rust-transpile/src/translator/mod.rs
kkysen: So this removal is somewhat independent of the other changes?  If so, could you put this removal here in its own commit with your comment here as the commit message?  Unless I'm misunderstanding what's happening here.

c2rust-transpile/tests/snapshots/snapshots__transpile@volatile.c.2024.snap
Rua: There was a comment in the code that says why it's there:

```rust
        // We explicitly annotate the type of pointer we're reading from
        // in order to avoid omitted bit-casts to const from causing the
        // wrong type to be inferred via the result of the pointer.
```

c2rust-transpile/src/translator/mod.rs
Rua: It's not independent, it's also needed for the rest to make sense. I added casting to the handling of `LValueToRValue` casts, but that would not make sense if the inner expression was already being translated as the right type. Then you'd just be casting it to itself. In this case, it turns out to be a no-op, but still.

c2rust-transpile/tests/snapshots/snapshots__transpile@volatile.c.2024.snap
kkysen: Ah, oops, I didn't see that.  Is the bit cast to const a mistake here then, though, as the comment says it's supposed to be omitted?

Rua: I believe the cast is being added by this code:
https://github.com/immunant/c2rust/blob/9169379dc0c359e22c43c7d41161a54dfc6c48f7/c2rust-transpile/src/translator/mod.rs#L3046-L3053

Since `pvi` is a mutable pointer, and `write` is `false` here, the comparison succeeds and a cast is added. Not sure if it really makes sense to do that even for `mut` > `const` though, since that can just be autocoerced.

c2rust-transpile/src/translator/mod.rs
kkysen: Okay, that makes sense.

kkysen (approved):

PR #1711, opened 2026-03-31
transpile: Prefix `c_ast::{BinOp, UnOp, UnTypeOp}` with `C` to disambiguate better

Rua:
There's both a Rust and a C `BinOp`, and they keep causing name clashes. Since the naming scheme elsewhere already prefixes C types with `C` (`CExpr`, `CLiteral` etc), this makes these follow that scheme so there's less hassle in the future.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -17,14 +17,13 @@ use log::{error, trace, warn};
 use proc_macro2::{Punct, Spacing::*, Span, TokenStream, TokenTree};
 use syn::spanned::Spanned as _;
 use syn::{
-    AttrStyle, BareVariadic, Block, Expr, ExprBinary, ExprBlock, ExprBreak, ExprCast, ExprField,
-    ExprIndex, ExprParen, ExprReturn, ExprUnary, FnArg, ForeignItem, ForeignItemFn,
+    AttrStyle, BareVariadic, BinOp, Block, Expr, ExprBinary, ExprBlock, ExprBreak, ExprCast,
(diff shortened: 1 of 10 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -24,7 +24,7 @@ use syn::{
     ItemTrait, ItemTraitAlias, ItemType, ItemUnion, ItemUse, Lit, MacroDelimiter, PathSegment,
     ReturnType, Stmt, Type, TypeTuple, UseTree, Visibility,
 };
-use syn::{BinOp, UnOp}; // To override `c_ast::{BinOp,UnOp}` from glob import.
+use syn::{BinOp, UnOp};
```
kkysen: Can you fold this into the `use syn::{}` above?

kkysen (approved):
This LGTM, and it does fix the annoying need to prefix things with `c_ast::` before.  That said, I wanted to let you know that we have limited review bandwidth, so we will prioritize reviewing PRs (when we have time) that fix things (as you have done many times ❤️) vs. the more refactoring/reorganizing PRs, which are nice to have, but generally not as important.

Rua:
Yeah that's totally fair! I kinda figured that was the case, so I've tried to avoid basing fixes on top of refactors and keeping the refactors smaller and more limited in scope.

kkysen:
> Yeah that's totally fair! I kinda figured that was the case, so I've tried to avoid basing fixes on top of refactors and keeping the refactors smaller and more limited in scope.

Yup, I've noticed!  It's been very appreciated.

Issue #398, opened 2022-05-10
Generated Rust for inline assembly violates borrow-checking rules


oinoom:
For the following example:

```c
int six(void) {
    int out = 0;
    int six = 6;

    asm("add %0, %1\n\t"
        : "=r"(out)
        : "r"(six), "0"(out));

    return out;
}
```

the following rust code is generated

```rust
#[no_mangle]
pub unsafe extern "C" fn rust_six() -> libc::c_int {
    let mut out: libc::c_int = 0 as libc::c_int;
    let mut six: libc::c_int = 6 as libc::c_int;
    let fresh9 = &mut out;
    let fresh10;
    let fresh11 = out;
    asm!(
        "add {0}, {1}\n\t", inlateout(reg) c2rust_asm_casts::AsmCast::cast_in(fresh9,
        fresh11) => fresh10, inlateout(reg) six => _, options(preserves_flags, pure,
        readonly)
    );
    c2rust_asm_casts::AsmCast::cast_out(fresh9, fresh11, fresh10);
    return out;
}
```

which results in the following violations

```
error[E0503]: cannot use `out` because it was mutably borrowed
  --> src/asm.rs:71:19
     |
69 |     let fresh9 = &mut out;
     |                  -------- borrow of `out` occurs here
70 |     let fresh10;
71 |     let fresh11 = out;
     |                   ^^^ use of borrowed `out`
72 |     asm!(
73 |         "add {0}, {1}\n\t", inlateout(reg) c2rust_asm_casts::AsmCast::cast_in(fresh9,
     |                                                                               ------ borrow later used here
```

PR #1712, opened 2026-03-31
asm-casts: Take raw pointer for `out` argument

Rua:
This is needed to make the transpiler generate raw borrows for assembly, which in turn is needed to deal with #309 and #398.

This is a potentially breaking API change for the asm-casts crate, because it makes the trait methods `unsafe`.

### DIFF
--- a/c2rust-asm-casts/src/lib.rs
+++ b/c2rust-asm-casts/src/lib.rs
@@ -14,8 +14,8 @@ pub struct AsmCast<Out, In>(PhantomData<(Out, In)>);
 pub trait AsmCastTrait<Out, In> {
     type Type;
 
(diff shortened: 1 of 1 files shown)


kkysen (approved):
LGTM.  As for the breaking changes, the only thing we need to worry about is when `c2rust transpile` generates code assuming the breaking change in `c2rust-asm-casts` and then hardcodes a version in `Cargo.toml` instead of a relative dependency.  This shouldn't be a concern in this PR, but may be one when the transpiler is updated to take advantage of this.

kkysen (commented):
Wait, sorry, I just merged this, but I realized it has no safety comments.  Could you add those in a follow-up PR?  Both for the method/trait definition and the use sites.

PR #1714, opened 2026-03-31
asm-casts: Add safety documentation

Rua:
Not sure if `out` actually needs to be dereferenceable for `cast_in`, but if it's part of a safety contract it can be relied on for later changes?

### DIFF
--- a/c2rust-asm-casts/src/lib.rs
+++ b/c2rust-asm-casts/src/lib.rs
@@ -14,7 +14,14 @@ pub struct AsmCast<Out, In>(PhantomData<(Out, In)>);
 pub trait AsmCastTrait<Out, In> {
     type Type;
 
+    /// # Safety
+    ///
+    /// `out` must be dereferenceable.
     unsafe fn cast_in(_: *mut Out, x: In) -> Self::Type;
+
+    /// # Safety
+    ///
+    /// `out` must be dereferenceable.
     unsafe fn cast_out(out: *mut Out, _: In, x: Self::Type);
 }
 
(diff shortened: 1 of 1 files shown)


c2rust-asm-casts/src/lib.rs
```diff
@@ -23,10 +23,14 @@ macro_rules! impl_triple {
 		impl<$($param),*> AsmCastTrait<$out, $in> for AsmCast<$out, $in> {
 			type Type = $inner;
 
+            /// # Safety
+            /// `out` must be dereferenceable.
```
kkysen:
```suggestion
            /// # Safety
            ///
            /// `out` must be dereferenceable.
```

Also, could you put these on the trait definition, not the implementation?

c2rust-asm-casts/src/lib.rs
```diff
@@ -126,7 +130,9 @@ mod tests {
 
                     let x = 42usize as $ty1;
                     let mut y: $ty2 = 0 as $ty2;
+                    // Safety: decayed reference is always dereferenceable.
                     let z = unsafe { AsmCast::cast_in(&mut y, x) } + 1;
+                    // Safety: decayed reference is always dereferenceable.
```
kkysen:
```suggestion
                    // SAFETY: decayed reference is always dereferenceable.
let z = unsafe { AsmCast::cast_in(&mut y, x) } + 1;
                    // SAFETY: decayed reference is always dereferenceable.
```

kkysen (approved):
> Not sure if `out` actually needs to be dereferenceable for `cast_in`, but if it's part of a safety contract it can be relied on for later changes?

Yeah.  The previous contract was that it was a `&mut` and so this isn't breaking that, and we can weaken the requirements later without breaking anything.

PR #1716, opened 2026-04-01
transpile: Use raw borrows in assembly translations

Rua:
- Fixes #398 
- Fixes #309

Since this depends on the changes made to asm-casts in #1712, I don't know if this will work as-is or if other changes are needed. It doesn't seem like the asm-casts crate is available for snapshot tests in any case?

PR #1719, opened 2026-04-01
transpile: tests: add `.expect_unresolved_import` instead of hardcoding `libc`

kkysen:
Also, we check that the error messages are as expected, too (meaning we actually run `rustc`).  This also makes the `libc` tests more explicit.

This should also handle #1716.

### DIFF
--- a/c2rust-rust-tools/src/lib.rs
+++ b/c2rust-rust-tools/src/lib.rs
@@ -245,7 +245,7 @@ fn run_rustc(
     // So we instead just check that these errors are as expected
     // based on the expected crates to be imported (`imported_crates`).
     let stderr = str::from_utf8(&output.stderr).unwrap();
-    let error_lines = stderr
+    let mut error_lines = stderr
         .split('\n')
         // .split(|&b| b == b'\n')
         .filter(|line| line.starts_with("error[E"))
@@ -272,13 +272,16 @@ fn run_rustc(
         dbg!(absolute_use_path);
         dbg!(relative_path);
 
-        assert!(
-            error_lines.contains(absolute_path)
-                || error_lines.contains(absolute_use_path)
-                || error_lines.contains(relative_path)
-        );
+        // Pre-compute to avoid `||` short-circuiting.
+        let absolute_path = error_lines.remove(absolute_path);
+        let absolute_use_path = error_lines.remove(absolute_use_path);
+        let relative_path = error_lines.remove(relative_path);
+
+        assert!(absolute_path || absolute_use_path || relative_path);
     }
     if !imported_crates.is_empty() {
+        dbg!(&error_lines);
+        assert!(error_lines.is_empty());
         return;
     }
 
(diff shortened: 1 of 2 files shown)


c2rust-rust-tools/src/lib.rs
```diff
@@ -206,10 +231,59 @@ fn run_rustc(rs_path: &Path, edition: RustEdition, crate_name: &str, expect_erro
...
+    let status = output.status;
     if status.success() {
         fs_err::remove_file(&rlib_path).unwrap();
     }
+    io::stdout().write_all(&output.stdout).unwrap();
+    io::stderr().write_all(&output.stderr).unwrap();
+
+    // Using rustc itself to build snapshots that reference crates (like libc) is difficult because we don't know
+    // the appropriate --extern libc=/path/to/liblibc-XXXXXXXXXXXXXXXX.rlib to pass.
+    // Skip for now, as we've already compared the literal text,
```
fw-immunant, line 244: This comment ("skip for now") doesn't make sense anymore as compilation is no longer being skipped.

c2rust-rust-tools/src/lib.rs
```diff
@@ -206,10 +231,59 @@ fn run_rustc(rs_path: &Path, edition: RustEdition, crate_name: &str, expect_erro
...
+        // .split(|&b| b == b'\n')
+        .filter(|line| line.starts_with("error[E"))
+        .collect::<HashSet<_>>();
+    dbg!(&error_lines);
+    for imported_crate in imported_crates {
+        // For `::{imported_crate}::*`.
+        let absolute_path = match edition {
+            Edition2021 => format!("error[E0433]: failed to resolve: could not find `{imported_crate}` in the list of imported crates"),
+            Edition2024 => format!("error[E0433]: cannot find `{imported_crate}` in the crate root"),
+        };
```
fw-immunant, line 258: It seems to me that we should parse the error code from the message and filter on that plus the mentioned library name. Hard-coding the whole error message seems very brittle.

fw-immunant (dismissed):
The general approach seems okay (not wonderful but Cargo is very resistant to helping us find the requisite information here), but I wasn't happy with some specifics. After addressing inline comments should be good to go.

c2rust-rust-tools/src/lib.rs
kkysen: How do I parse the library name, though, without hardcoding the error message?  I agree it could be brittle, but it'd have a simple error and then we update the error message needed, so it's a simple fix if it's ever changed, too.

fw-immunant: For the library name we do probably have to choose between something brittle (like hard-coding the error message) and something loose (like a simple string-contains check), but for the error code we should be able to parse in a somewhat robust way. E.g., in JSON output format it's given in its own "code" field. Even without JSON output, the "error[ENNNN]" prefix should be relatively stable.

c2rust-rust-tools/src/lib.rs
kkysen: Updated now.

c2rust-rust-tools/src/lib.rs
kkysen: We discussed this offline, and agreed that it's not worth it to change this here.  Parsing would mean more work, and it's easy enough to just update things when they break instead of trying to make it more future-proof to start with.

fw-immunant (approved):

PR #1721, opened 2026-04-02
transpile: Use wrapping functions for atomic unsigned arithmetic

Rua:
- Fixes #1708

### DIFF
--- a/c2rust-transpile/src/translator/atomics.rs
+++ b/c2rust-transpile/src/translator/atomics.rs
@@ -350,8 +350,14 @@ impl<'c> Translation<'c> {
                     let order = static_order(order);
                     let val =
                         val1.expect("__atomic arithmetic operations must have a val argument");
+                    let val_type_id = self.ast_context[val1_id.unwrap()]
+                        .kind
+                        .get_qual_type()
+                        .ok_or_else(|| format_err!("bad val1 type"))?;
                     ptr.and_then(|ptr| {
-                        val.and_then(|val| self.convert_atomic_op(ctx, atomic_op, order, ptr, val))
+                        val.and_then(|val| {
+                            self.convert_atomic_op(ctx, atomic_op, order, ptr, val, val_type_id)
+                        })
                     })
                 } else {
                     unimplemented!("atomic not implemented: {}", name)
@@ -390,6 +396,7 @@ impl<'c> Translation<'c> {
         order: Ordering,
         dst: Box<Expr>,
         src: Box<Expr>,
+        src_type_id: CQualTypeId,
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         // Emit `atomic_func(a0, a1) (op a1)?`
         let atomic_func =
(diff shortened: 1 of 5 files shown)


c2rust-transpile/src/translator/atomics.rs
```diff
@@ -345,51 +345,18 @@ impl<'c> Translation<'c> {
...
-                let order = static_order(order);
-
-                let fetch_first =
-                    name.starts_with("__atomic_fetch") || name.starts_with("__c11_atomic_fetch");
-                let val = val1.expect("__atomic arithmetic operations must have a val argument");
-                ptr.and_then(|ptr| {
-                    val.and_then(|val| {
-                        self.convert_atomic_op(ctx, intrinsic_name, order, ptr, val, fetch_first)
+            _ => {
+                if let Some(atomic_op) = CAtomicOp::from_atomic_fn(name) {
```
kkysen: Putting this in the match arm is still unstable, right?

Rua: Yeah, I think that only got stabilised a few releases ago?

c2rust-transpile/src/translator/atomics.rs
```diff
@@ -482,3 +439,159 @@ impl<'c> Translation<'c> {
...
+    }
+
+    pub fn rust_bin_op(self) -> BinOp {
+        match self {
+            Self::FetchAdd | Self::AddFetch => BinOp::Add(Default::default()),
+            Self::FetchSub | Self::SubFetch => BinOp::Sub(Default::default()),
+            Self::FetchOr | Self::OrFetch => BinOp::BitOr(Default::default()),
+            Self::FetchAnd | Self::AndFetch => BinOp::BitAnd(Default::default()),
+            Self::FetchXor | Self::XorFetch => BinOp::BitXor(Default::default()),
+            // Gets translated to `!(atomic_nand(arg0, arg1) & arg1)`
```
kkysen:
```suggestion
            // Gets translated to `!(atomic_nand(arg0, arg1) & arg1)`.
            // See uses of `Self::is_nand`.
```

c2rust-transpile/src/translator/atomics.rs
```diff
@@ -482,3 +439,159 @@ impl<'c> Translation<'c> {
         }
     }
 }
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CAtomicOp {
```
kkysen: There are other atomic ops that this doesn't cover.  Maybe this should be `CAtomicBinOp`?  `fn convert_atomic_op` should also probably be `fn convert_atomic_bin_op`, too.

Rua: It could in principle be extended to cover the other atomic operations too. But that wasn't the original intent of this PR.

c2rust-transpile/src/translator/atomics.rs
```diff
@@ -424,13 +431,24 @@ impl<'c> Translation<'c> {
...
-                // For nand, return `!(atomic_nand(arg0, arg1) & arg1)`
-                mk().unary_expr(UnOp::Not(Default::default()), val)
+
+            let src_type_kind = &self.ast_context.resolve_type(src_type_id.ctype).kind;
+            let arg1 = mk().ident_expr(arg1_name);
+
+            let mut val = if let (true, Some(wrapping_arith_fun)) = (
+                src_type_kind.is_unsigned_integral_type(),
+                atomic_op.rust_wrapping_arith_fn(),
+            ) {
```
kkysen:
```suggestion
            let mut val = if let Some(wrapping_arith_func) = atomic_op.rust_wrapping_arith_fn().filter(|_| src_type_kind.is_unsigned_integral_type()) {
```

Not sure how to format this, but I think this is cleaner than the `true` matching with a tuple.

Rua: This would be a prime candidate for a `let` chain, heh.

kkysen: Yup, definitely.

kkysen (approved):

PR #1727, opened 2026-04-03
transpile: Fix `f128.c` snapshot test

Rua:
It seems that the way snapshot tests work has changed, and so this test was never actually being run? That would explain why there were no errors from it. Now that it's actually working, it needed to be split because MacOS apparently doesn't support `__float128` at all.


Rua:
It seems that Clang 15 doesn't support `__float128` either? How does c2rust deal with that situation in tests?

kkysen (commented):
This isn't really splitting `f128.c` as it is actually testing it, right?  A `#[test]` for `f128.c` was missing before, right?

Rua:
It's splitting it by OS, but yes.

Rua:
This now depends on #1730, which allows the Clang 15 test to pass.

c2rust-transpile/tests/snapshots.rs
```diff
@@ -436,6 +436,14 @@ fn test_call_only_once() {
     transpile("call_only_once.c").os_specific(true).run();
 }
 
+#[test]
+fn test_f128() {
+    transpile("f128.c")
+        .expect_compile_error(true)
```
kkysen: Can this work with #1719 now?

kkysen (approved):

PR #1728, opened 2026-04-03
transpile: Qualify `::c2rust_bitfields::BitfieldStruct` in derives

### LINKED ISSUE #1726: C2Rust-generated bitfield code fails to compile due to missing `use c2rust_bitfields::BitfieldStruct`
When translating C structs that use bitfields, C2Rust generates Rust code using the `BitfieldStruct` derive macro and `#[bitfield(...)]` attributes from the `c2rust_bitfields` crate. However, the generated code does not include the required import:
```rust
use c2rust_bitfields::BitfieldStruct;
```
As a result, the generated Rust code fails to compile.
**Reproducer**
```c
#include <stdio.h>
#include <stdint.h>

struct PacketHeader {
    uint8_t version : 3;
    uint8_t type : 2;
    uint8_t flags : 3;
    uint16_t sequence : 10;
    uint16_t length : 6;
};

int main(void) {
    struct PacketHeader h = {0};
    h.version = 5;
    h.sequence = 513;
    printf("version=%u sequence=%u\n", h.version, h.sequence);
    return 0;
}
```
**Generated Rust (excerpt)**
```rust
#[derive(Copy, Clone, BitfieldStruct)]
#[repr(C)]
pub struct PacketHeader {
    #[bitfield(padding)]
    pub c2rust_padding: [u8; 1],
    #[bitfield(name = "version", ty = "uint8_t", bits = "0..=2")]
    #[bitfield(name = "type_0", ty = "uint8_t", bits = "3..=4")]
    #[bitfield(name = "flags", ty = "uint8_t", bits = "5..=7")]
    #[bitfield(name = "sequence", ty = "uint16_t", bits = "16..=25")]
    #[bitfield(name = "length", ty = "uint16_t", bits = "26..=31")]
    pub version_type_0_flags_sequence_length: [u8; 3],
}
```
Compilation Error:
```
error: cannot find derive macro `BitfieldStruct` in this scope
   --> src/main.rs:16:23
    |
16 | #[derive(Copy, Clone, BitfieldStruct)]
   |                       ^^^^^^^^^^^^^^

error: cannot find attribute `bitfield` in this scope
   --> src/main.rs:19:7
    |
19 |     #[bitfield(padding)]
```
**Expected Behavior**
When generating Rust code that uses `BitfieldStruct` and `#[bitfield]`, C2Rust should automatically emit the required import:
```rust
use c2rust_bitfields::BitfieldStruct;
```
so that the generated code compiles without manual modification.
**Environment**
- c2rust version: v0.21.0
- platform: Ubuntu 24.04
- Rust: nightly-2022-08-08
- Clang version: 17.0.6

Rua:
- Fixes #1726

### DIFF
--- a/c2rust-ast-builder/src/builder.rs
+++ b/c2rust-ast-builder/src/builder.rs
@@ -2008,6 +2008,14 @@ impl Builder {
         self.meta(Meta::Path(path))
     }
 
+    pub fn abs_meta_path<Pa>(self, path: Pa) -> Meta
+    where
+        Pa: Make<Path>,
+    {
+        let path = mk().abs_path(path);
+        self.meta_path(path)
+    }
+
     /// makes a meta item with the given path and some arguments
     /// # Examples
     ///
--- a/c2rust-transpile/src/translator/structs_unions.rs
+++ b/c2rust-transpile/src/translator/structs_unions.rs
@@ -71,8 +71,8 @@ impl<'a> Translation<'a> {
 
         let mut derives = vec![];
         if !contains_va_list {
-            derives.push("Copy");
-            derives.push("Clone");
+            derives.push(mk().meta_path("Copy"));
+            derives.push(mk().meta_path("Clone"));
         };
         let has_bitfields =
             fields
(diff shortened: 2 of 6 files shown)


kkysen (commented):
LGTM, but could we also add a snapshot test for this?  You can base it on #1719, too.

Rua:
Looks like your additions to the snapshot test suite are not enough to clear up the compile errors in this case. Because the derive macro can't be imported, it's not generating methods and now _those_ are also missing.

kkysen (approved):
I'll open a PR after this to fix the error handling.

PR #1569, opened 2026-01-26
tests/integration: add `c2rust-postprocess` for `json-c`

kkysen:

PR #1729, opened 2026-04-04
tests/integration: add c2rust-postprocess for json-c 

thedataking:
Landing Khyber's work from PR #1569. Simplified by omitting `--fail-fast` option.


.github/workflows/internal-testsuite.yml
```diff
@@ -114,16 +114,24 @@ jobs:
         export LLVM_CONFIG_PATH=/usr/bin/llvm-config-${{ matrix.clang-version }}
         cargo build --release
 
+    - name: Build tools
+      run: |
+        (cd tools/split_rust && cargo build --release)
```
ahomescu: I think you can do `cargo build --release --manifest-path tools/split_rust/Cargo.toml` instead. Alternatively, should they be in the top-level workspace?

ahomescu (approved):

.github/workflows/internal-testsuite.yml
kkysen: Since it needs to run under `tools/` or a subdirectory to pick up the `rust-toolchain.toml`, it seems simpler to use the old way of `cd`ing.

PR #1731, opened 2026-04-05
transpile: tests: Run `test_translator.py` from `cargo nextest` with the currently built transpiler path

chiragdhawan24:
- Fixes Issue #1398

When running the tests, `test_translator.py` could end up using a stale `c2rust`/`c2rust-transpile` binary unless `cargo build --release` had been run manually beforehand.

This change updates the test flow so repo tests use the latest version of `c2rust-transpile` binary built by `cargo nextest run`, by passing the current transpiler path into `test_translator.py` and overriding the configured transpiler path for that test run.

This removes the manual pre-build requirement and makes local test execution less error-prone.

Testing:
- ran `cargo nextest run` locally
- verified that `test_translator.py` was invoked, and it received the expected `c2rust-transpile` path
- confirmed the tests ran successfully without manually running `cargo build --release`


c2rust/tests/test_translator_uses_current_transpiler.rs
```diff
@@ -0,0 +1,20 @@
+use std::process::Command;
+
+#[test]
+fn test_translator_uses_current_transpiler() {
+    let transpile_path = env!("CARGO_BIN_EXE_c2rust-transpile");
+
+    let status = Command::new("python")
```
ahomescu: Do we need to use `uv` here? Does this work without?

c2rust/tests/test_translator_uses_current_transpiler.rs
```diff
@@ -0,0 +1,20 @@
...
+        .arg("../scripts/test_translator.py")
```
ahomescu: Does this require the test to always be run from the c2rust directory?

scripts/test_translator.py
```diff
@@ -578,7 +578,7 @@ def run(self) -> List[TestOutcome]:
                 args += ["--target", self.target]
 
             retcode, stdout, stderr = cargo[args].run(retcode=None)
-
```
ahomescu, line 581: No white space changes please.

scripts/test_translator.py
```diff
@@ -689,11 +697,13 @@ def main() -> None:
 
     logging.debug("args: %s", " ".join(sys.argv))
 
+
```
ahomescu: Is this needed?

c2rust/tests/test_translator_uses_current_transpiler.rs
kkysen: `test_translator.py` has a `#!` running `uv`, so you should just run the script directly.  No `python` or `uv` needed here.

c2rust/tests/test_translator_uses_current_transpiler.rs
kkysen: `cargo test`/`nextest` always runs with a consistent CWD, so this should be fine.

c2rust/tests/test_translator_uses_current_transpiler.rs
```diff
@@ -0,0 +1,20 @@
...
+        .arg("../tests/unit")
+        .arg("--transpiler")
+        .arg(transpile_path)
+        .status()
+        .expect("failed to run python script");
+
+    assert!(
+        status.success(),
+        "test_translator.py failed with status: {}",
+        status
```
kkysen:
```suggestion
        "test_translator.py failed with status: {status}"
```

c2rust/tests/test_translator_uses_current_transpiler.rs
```diff
@@ -0,0 +1,20 @@
+use std::process::Command;
+
+#[test]
+fn test_translator_uses_current_transpiler() {
```
kkysen:
```suggestion
fn test_translator() {
```

This just runs the whole `test_translator.py` script, so I think this is more appropriate.  Same for the file name.

c2rust/tests/test_translator_uses_current_transpiler.rs
kkysen: Could you also print the whole command that failed?  Like this

```rust
let mut cmd = Command::new("../scripts/test_translator.py");
cmd.args(["../tests/unit", "--transpiler"]);
cmd.arg(transpile_path);
let status = cmd.status().unwrap_or_else(|e| panic!("{cmd:?} failed: {e}");
assert!(status.success(), "{cmd:?} failed with {status}");
```

kkysen (changes requested):
Thanks for this!  A few other things:

We use `cargo nextest run` instead of `cargo test`, so run that locally instead.

Can you update the README and anywhere else that references `test_translator.py` to say you can also just run `cargo nextest run`?

c2rust/tests/test_translator_uses_current_transpiler.rs
chiragdhawan24: I have addressed this in the latest commit.

c2rust/tests/test_translator_uses_current_transpiler.rs
chiragdhawan24: I have addressed this in the latest commit.

scripts/test_translator.py
chiragdhawan24: I have addressed this in the latest commit.

scripts/test_translator.py
chiragdhawan24: I have addressed this in the latest commit.

c2rust/tests/test_translator.rs
```diff
@@ -0,0 +1,16 @@
+use std::process::Command;
+
+#[test]
+fn test_translator() {
+    let transpile_path = env!("CARGO_BIN_EXE_c2rust-transpile");
+
+    let mut cmd = Command::new("../scripts/test_translator.py");
+    cmd.args(["../tests/unit", "--transpiler"]);
+    cmd.arg(transpile_path);
```
kkysen:
```suggestion
    cmd.args(["../tests/unit", "--transpiler", transpile_path]);
```

I didn't realize `transpile_path` is a `&str` (from `env!`) instead of a `&Path`, so I think you can do this.

c2rust/tests/test_translator.rs
```diff
@@ -0,0 +1,16 @@
...
+
```
kkysen, line 12:
```suggestion
    // `test_translator.py` needs to be able to re-select the toolchain for different editions.
    cmd.env_remove("RUSTUP_TOOLCHAIN");

```

chiragdhawan24: I have addressed these changes in the latest commit.

c2rust/tests/test_translator.rs
chiragdhawan24: I have addressed this change in `cmd`'s args in the latest commit.

chiragdhawan24:
> Thanks for this! A few other things:
> 
> We use `cargo nextest run` instead of `cargo test`, so run that locally instead.
> 
> Can you update the README and anywhere else that references `test_translator.py` to say you can also just run `cargo nextest run`?

I have updated README and other files where `test_translator.py` was mentioned. Kindly review the changes and let me know if they're alright.

docs/README-developers.md
```diff
@@ -89,6 +89,8 @@ More details about tests can be found in the [tests folder](../tests/).
 *Note*: These run integration tests that invoke `c2rust transpile`.
 `cargo test` only runs unit tests and doc tests as of now.
 
+*Note*: The tests executed by `test_translator.py` can also be invoked simply by running `cargo nextest run` in the main c2rust workspace. `cargo nextest run` even ensures that the repo tests use the latest version of `c2rust-transpile` binary.
```
kkysen:
```suggestion
Most tests are found in the [`tests`](../tests/) folder.

Most tests can also be run with

```sh
cargo nextest run
```

which includes running the [`tests/unit`](../tests/unit/) tests.
The `tests/unit` tests can also be run manually using

```sh
./scripts/test_translator.py tests/unit
```

which also allows you to run with other options.

The [`tests/integration`](../tests/integration/) tests
aren't run automatically by `cargo nextest run`.
To run them, run

```sh
# Needs to be run from `tests/integration/` (or further inside)
# to correctly load the `pyproject.toml`.
cargo build --release
(export PATH=$PWD/target/release:$PATH && export C2RUST_DIR=$PWD && cd tests/integration && ./test.py curl json-c lua nginx zstd li
bxml2 python2 libmcs)
```

chiragdhawan24: I have changed the `docs/README-developers.md` as per the said changes in the latest commit. 

Kindly re-review the same. Would really appreciate if this could now be merged into the master branch.

ahomescu (commented):
I have one main objection to this: the Python scripts should be wrappers around `cargo test` or `nextest`, not the other way around. I don't think it's a good idea to call out from `cargo`. Could you just add a check instead?

PR #1732, opened 2026-04-07
refactor: Fix a bug in `reorganize_definitions` for uses of foreign types

### LINKED ISSUE #1733: reorganize_definitions produces broken code on libexpat
@spernsteiner reported that running the transpiler followed by `reorganize_definitions` on libexpat produces code that doesn't compile:
```
error[E0412]: cannot find type `XML_ParserStruct` in this scope
   --> lib.rs:528:33
    |
528 |     pub type XML_Parser =  *mut XML_ParserStruct;
    |                                 ^^^^^^^^^^^^^^^^ not found in this scope
    |
help: consider importing this struct
    |
528 +     use crate::src::tests::runtests::XML_ParserStruct;
    |

For more information about this error, try `rustc --explain E0412`.
```

The problem happens in the `expat_h` header module that is missing a `pub use ...::XML_ParserStruct;` that should be copied from somewhere.

ahomescu:
* Fixes #1733.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -1585,15 +1585,9 @@ impl<'a, 'tcx> HeaderDeclarations<'a, 'tcx> {
                         if let ItemKind::Use(..) = item.kind {
                             // If the import refers to an existing foreign item, do
                             // not replace it.
-                            let path = self.cx.resolve_use_id(item.id);
-                            if let Some(did) = path.res.opt_def_id() {
-                                if let Some(Node::ForeignItem(_)) =
-                                    self.cx.hir_map().get_if_local(did)
-                                {
-                                    return ContainsDecl::Definition(existing_decl);
-                                }
+                            if is_use_of_foreign(&item, &self.cx) {
+                                return ContainsDecl::Definition(existing_decl);
                             }
-
                             return ContainsDecl::Equivalent(existing_decl);
                         }
                         if foreign_equiv(&existing_foreign, &item) {
@@ -1629,15 +1623,9 @@ impl<'a, 'tcx> HeaderDeclarations<'a, 'tcx> {
                             return ContainsDecl::Equivalent(existing_decl);
                         } else if let ItemKind::Use(_) = existing_item.kind {
                             // A use takes precedence over a foreign declaration
-                            // unless the use refers to the foreign declaration are
-                            // attempting to insert.
-                            let path = self.cx.resolve_use_id(existing_item.id);
-                            if let Some(did) = path.res.opt_def_id() {
-                                if let Some(Node::ForeignItem(_)) =
-                                    self.cx.hir_map().get_if_local(did)
-                                {
-                                    return ContainsDecl::Definition(existing_decl);
-                                }
+                            // unless the use also refers to a foreign.
+                            if is_use_of_foreign(&existing_item, &self.cx) {
+                                return ContainsDecl::Definition(existing_decl);
                             }
                             return ContainsDecl::Equivalent(existing_decl);
                         }
@@ -1716,6 +1704,19 @@ fn foreign_equiv(foreign: &ForeignItem, item: &Item) -> bool {
 
         _ => false,
     }
+}
+
+/// Check if the [`Item`] is a [`Use`] of another [`ForeignItem`].
+fn is_use_of_foreign(item: &Item, cx: &RefactorCtxt) -> bool {
+    if !matches!(item.kind, ItemKind::Use(..)) {
+        return false;
+    }
+
+    let path = cx.resolve_use_id(item.id);
+    let Some(did) = path.res.opt_def_id() else {
+        return false;
+    };
+    matches!(cx.hir_map().get_if_local(did), Some(Node::ForeignItem(_)))
 }
 
 /// Check if the `Item` has the `#[header_src = "/some/path"]` attribute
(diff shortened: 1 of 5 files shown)


c2rust-refactor/tests/snapshots/reorganize_definitions.rs
```diff
@@ -217,6 +217,38 @@ pub mod foo {
...
+        pub type FooPtr = *mut Foo;
+    }
+}
+pub mod foreign2 {
+    #[c2rust::header_src = "foo.h:1"]
+    pub mod foo_h {
+        extern "C" {
+            #[c2rust::src_loc = "1:1"]
+            pub type Foo;
+        }
```
randomPoison: Do we generate extern types? Those are unstable right? The test enables the feature for it, but didn't previously  have any instances of extern types.

randomPoison (commented):
Can you give a bit of context for what this change is accomplishing and why it's needed, e.g. what bug you're trying to fix? I'm not intimately familiar with reorganize definitions, nor with the CRISP changes that I assume are the motivation for this, and I'm having trouble understanding from the test change what the expected change in behavior is.

In general, it'd be helpful to provide this context in the PR description. Reviewing a change without context is hard and time consuming because I have to do a bunch of digging to figure out the context for myself.

ahomescu:
> Can you give a bit of context for what this change is accomplishing and why it's needed, e.g. what bug you're trying to fix?

We don't seem to have a bug yet, I can go open one. Stuart reported a crash in `reorganize_definitions` when transpiling libexpat.

c2rust-refactor/tests/snapshots/reorganize_definitions.rs
ahomescu: This new test is a minimization of something actually happening for libexpat. I'm not sure what C code the transpiler is generating them for, I can investigate.

ahomescu:
```c
struct XML_ParserStruct;
typedef struct XML_ParserStruct *XML_Parser;
```
This is the original C type.

ahomescu:
> I'm not intimately familiar with reorganize definitions

I'm not sure who else is, other than me.

c2rust-refactor/tests/snapshots/reorganize_definitions.rs
kkysen: > Do we generate extern types?

Yes.

> Those are unstable right?

Yes.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1718,6 +1710,19 @@ fn foreign_equiv(foreign: &ForeignItem, item: &Item) -> bool {
     }
 }
 
+/// Check if the `Item` is a `Use` of another `ForeignItem`.
```
kkysen: nit:

```suggestion
/// Check if the [`Item`] is a [`Use`] of another [`ForeignItem`].
```

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1562,8 +1562,12 @@ impl<'a, 'tcx> HeaderDeclarations<'a, 'tcx> {
             for existing_decl in existing_decls {
                 match &existing_decl.kind {
                     DeclKind::Item(existing_item) => match (&existing_item.kind, &item.kind) {
-                        // Replace a use with a real definition
+                        // Replace a use with a real definition, but a use of
+                        // an internal definition takes precedence over a foreign.
```
kkysen:
```suggestion
                        // an internal definition takes precedence over a foreign one.
```

randomPoison:
> We don't seem to have a bug yet, I can go open one.

That's fine, I don't necessarily need a ticket for it, but it'd be helpful if you can explain here what behavior you're trying to change.

> > I'm not intimately familiar with reorganize definitions
> 
> I'm not sure who else is, other than me.

That's also fine, I'm still able to review this, and reviewing changes like this is part of building an understanding of the refactorer and this transform. But since I (and others) don't have the context already, it's helpful if you can provide that context.

kkysen (changes requested):
Could you put the new test code in a new snapshot test?  The current one doesn't compile after `reorganize_definitions`, so it's missing an important part of testing (which would catch this bug).  So could you put it in its own snapshot file, initially set `.new_expect_compile_error(true)`, and then remove that in the 3rd commit that fixes the bug?

Otherwise, the changes to the refactorer and to the snapshot LGTM.

ahomescu:
> So could you put it in its own snapshot file

That was my original approach, but it seemed like we already put all the tests for a transform in one file? Or should I use `named`?

c2rust-refactor/src/transform/reorganize_definitions.rs
ahomescu: That line got a bit long, but that's fine.

c2rust-refactor/tests/snapshots.rs
```diff
@@ -463,3 +463,11 @@ fn test_test_reflect() {
...
     refactor("uninit_to_default").test();
 }
+
+#[test]
+fn test_reorganize_foreign_types() {
+    refactor("reorganize_definitions")
+        .named("reorganize_foreign_types.rs")
+        .new_expect_compile_error(true)
+        .test();
+}
```
kkysen: The tests are supposed to be sorted alphabetically.  Could you put it right after `fn test_reorganize_definitions`?

kkysen (approved):
LGTM now besides sorting the tests.

PR #1739, opened 2026-04-15
transpile: `prune_unwanted_decls` only after setting `prenamed_decls`

### LINKED ISSUE #1738: transpile: `typedef` of anonymous types are lost if name is not used
```C
typedef enum { A } E;

int foo(void) {
    return A;
}
```

```rust
#![allow(
    dead_code,
    non_camel_case_types,
    non_snake_case,
    non_upper_case_globals,
    unused_assignments,
    unused_mut
)]
pub type C2Rust_Unnamed = ::core::ffi::c_uint;
pub const A: C2Rust_Unnamed = 0;
#[no_mangle]
pub unsafe extern "C" fn foo() -> ::core::ffi::c_int {
    return A as ::core::ffi::c_int;
}
```

Rather than keeping the name `E`, the typedef's name is being lost here, and the enum is instead being translated as a regular anonymous enum. Seems likely related to the code that prunes unwanted declarations? It is probably eliminating the typedef because it's not used anywhere.

Rua:
- Depends on #1743
- Fixes #1738

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1171,19 +1171,18 @@ impl TypedAstContext {
             }
         }
 
-        // Unset c_main if we are not retaining its declaration
-        if let Some(main_id) = self.c_main {
-            if !wanted.contains(&main_id) {
-                self.c_main = None;
-            }
-        }
-
         // Prune any declaration that isn't considered live
         self.c_decls
             .retain(|&decl_id, _decl| wanted.contains(&decl_id));
 
-        // Prune top declarations that are not considered live
-        self.c_decls_top.retain(|x| wanted.contains(x));
+        // Remove references to removed decls that are held elsewhere.
+        self.c_decls_top.retain(|x| self.c_decls.contains_key(x));
+
+        if let Some(main_id) = self.c_main {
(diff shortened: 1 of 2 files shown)


fw-immunant:
Panic in CI:
```
[ FAILED ] translate enum_fwd_decl.c
thread 'main' panicked at 'Could not find CDeclId(155) in TypedAstContext', c2rust-transpile/src/c_ast/mod.rs:1009:44
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: c2rust_transpile::c_ast::TypedAstContext::set_prenamed_decls
   3: c2rust_transpile::translator::translate
   4: c2rust_transpile::transpile_single
   5: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   6: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   7: c2rust_transpile::transpile
   8: c2rust_transpile::main
```

Rua:
Yeah, it's panicking on the problem reported in #1740. The AST contains a `CDeclId` that doesn't refer to anything.

ahomescu (approved):

kkysen (approved):

PR #1743, opened 2026-04-15
ast-exporter: handle foreign/extern enums properly

### LINKED ISSUE #1740: transpile: Panic on forward declared enum
The C standard officially forbids forward declaring an enum, which is demonstrated by `clang -Wpedantic`. But c2rust itself has a test that relies on it, so presumably it wants to support the feature: https://github.com/immunant/c2rust/blob/3558805d0c94ce42b3efbce59e8cef784dcc78f4/tests/unit/enums/src/enum_fwd_decl.c

However, this produces an invalid AST, since the `CTypeKind::Enum` has an invalid decl ID. In that test, the enum isn't actually used, so it just gets pruned as dead code and never translated. But if the dummy function is modified to use the struct, the transpiler crashes when it tries to look up the decl:

```c
typedef enum _Evas_Filter_Support Evas_Filter_Support;

struct _Evas_Func
{
    Evas_Filter_Support (*gfx_filter_supports) (void *engine, void *cmd);
};

struct _Evas_Func foo(struct _Evas_Func i) { return i;}
```

> thread 'main' panicked at 'Could not find CDeclId(50) in TypedAstContext', c2rust-transpile/src/c_ast/iterators.rs:333:59
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Note that this code _does_ just compile on clang, albeit with a warning if you have `-Wpedantic`. The invalid decl ID seems to arise from the original clang AST, which gives a clang AST node id of zero in this case. c2rust doesn't handle the zero case and just treats it like it's a valid ID, but nothing ever gets defined with that ID, leading to the panic.

Rua:
- Fixes #1740

`getDefinition` returns `NULL` if there is no definition, only a forward declaration. That would then be exported as node number 0, which would not have any valid decl associated with it. This PR adds a check for the `NULL` case and uses the canonical decl instead.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2625,11 +2625,20 @@ class TranslateASTVisitor final
 
 void TypeEncoder::VisitEnumType(const EnumType *T) {
     auto ed = T->getDecl()->getDefinition();
+
+    // getDefinition can return NULL if there is only a forward declaration like `enum Foo;`
+    // Forward declarations of enums are not actually allowed by the C standard,
+    // but are supported by LLVM and are used in real code, see
+    // https://github.com/immunant/c2rust/pull/1743#discussion_r3120875668
+    if (!ed) {
+        ed = T->getDecl()->getCanonicalDecl();
+    }
+
     encodeType(T, TagEnumType, [T, ed](CborEncoder *local) {
         cbor_encode_uint(local, uintptr_t(ed));
     });
 
-    if (ed != nullptr) astEncoder->TraverseDecl(ed);
+    astEncoder->TraverseDecl(ed);
 }
 
 void TypeEncoder::VisitRecordType(const RecordType *T) {
--- a/tests/unit/enums/src/enum_fwd_decl.c
+++ b/tests/unit/enums/src/enum_fwd_decl.c
@@ -10,4 +10,4 @@ struct _Evas_Func
 };
 
 // dummy item imported by `test_enums.rs`
-int foo(int i) { return i;}
+struct _Evas_Func foo(struct _Evas_Func i) { return i;}


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2625,11 +2625,16 @@ class TranslateASTVisitor final
 
 void TypeEncoder::VisitEnumType(const EnumType *T) {
     auto ed = T->getDecl()->getDefinition();
+    
+    if (!ed) {
+        ed = T->getDecl()->getCanonicalDecl();
```
ahomescu, line 2630: A comment would be useful explaining why we're falling back on this.

tests/unit/enums/src/enum_fwd_decl.c
```diff
@@ -10,4 +10,4 @@ struct _Evas_Func
 };
 
 // dummy item imported by `test_enums.rs`
-int foo(int i) { return i;}
\ No newline at end of file
+struct _Evas_Func foo(struct _Evas_Func i) { return i;}
```
ahomescu, line 13: Why was this change needed?

tests/unit/enums/src/enum_fwd_decl.c
ahomescu, line 5: Does this cover your transpiler change? Or do we need an extra test for it?

ahomescu (changes requested):

c2rust-ast-exporter/src/AstExporter.cpp
Rua: Other than `ed` being NULL?

tests/unit/enums/src/enum_fwd_decl.c
Rua: Without it, the `struct` goes unused, and so c2rust prunes it, which defeats the point of the test.

tests/unit/enums/src/enum_fwd_decl.c
Rua: What transpiler change?

ahomescu: I meant the AST exporter change (which I'm guessing impacts the transpiler).

ahomescu:
Since forward declarations of `enum`s aren't allowed in C (per https://en.cppreference.com/c/language/enum and also the issue you opened), I'm starting to think we should fix our test instead.

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: `ed` is NULL if there is no definition, which is fine as the real reason.

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2625,11 +2625,18 @@ class TranslateASTVisitor final
 
 void TypeEncoder::VisitEnumType(const EnumType *T) {
     auto ed = T->getDecl()->getDefinition();
+
+    // getDefinition can return NULL if there is only a forward declaration like
+    // `enum Foo;`
```
ahomescu: It might be worth elaborating a bit on what's going on here: we're seeing a forward declaration of an `enum` which the C standard theoretically forbids, but both our test code and its upstream source at https://github.com/Enlightenment/efl/blob/master/src/lib/evas/include/evas_private.h#L77 do it anyway.

tests/unit/enums/src/enum_fwd_decl.c
Rua: It covers the change. Without the fix and only the changes to this file, the transpiler panics when transpiling it.

Rua:
> I'm starting to think we should fix our test instead.

What kind of fix?

ahomescu (approved):

kkysen (approved):

Issue #1736, opened 2026-04-15
Emit fine-grained unsafe operations


ahomescu:
We've gone back and forth (https://github.com/immunant/c2rust/pull/1720) on `deny_unsafe_op_in_unsafe_fn` and wrapping the entire body of every function in `unsafe`, but I think we have a third option: emit very narrow `unsafe` blocks only around the operations that need them, i.e., those at https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html.

Normally I'd argue to add a refactoring pass that does this, but maybe it wouldn't be too complicated to add to the transpiler? It wouldn't be difficult either way, other than tracking whether callees are unsafe (or just assuming they all are).

One downside I can think of is the transpiled code might get even more verbose than it already is.

ahomescu:
Another option would be to update the existing `sink_unsafe` transform to sink the blocks all the way down to the smallest expressions.

Rua:
Emitting blocks around any transpiled unsafe ops to begin with is probably a lot easier than trying to transform it later.

PR #1745, opened 2026-04-16
transpile: Remove `static_initializer_is_unsafe`

Rua:
- Depends on #1644.

This was already slated for removal in a TODO comment, and it doesn't seem like it's actually needed. `WithStmts` is handling everything already. Removing this function helps a little towards #1736 as well.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -427,7 +427,7 @@ impl<'c> Translation<'c> {
                     warn!("{builtin_name} has no Rust equivalent; emitting null pointer");
                 }
                 let level = self.convert_expr(ctx.unused(), args[0], None)?;
-                Ok(level.and_then(|_| -> TranslationResult<_> {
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/with_stmts.rs
```diff
@@ -90,6 +99,18 @@ impl<T> WithStmts<T> {
         &mut self.stmts
     }
 
+    /// If all statements in self.stmts are [`Item`] statements, returns the contained [`Item`]s.
+    /// Otherwise, returns [`None`].
+    pub fn stmts_to_items(&mut self) -> Option<Vec<Box<Item>>> {
+        std::mem::take(&mut self.stmts)
```
ahomescu, line 105: If there are non-`Item` elements in `stmts`, will this function leave us with an empty `WithStmts`? That doesn't seem right, we'd be losing statements.

ahomescu (changes requested):

ahomescu (approved):

PR #1746, opened 2026-04-20
refactor: Support associated constants in `reorganize_definitions`

### LINKED ISSUE #1660: Refactorer breaks imports of associated constants
Currently visible in the CI output of #1659, which introduces associated constants. The transpiler produces the right code, but the refactorer mishandles it afterwards.

Rua:
- Fixes #1660

This was quite tricky to implement, but it's now working and #1659 passes its tests with these additions. Only `const` associated items are supported. `type` items could be added later, but `fn` is probably never going to work because they are too complex to test for equivalence.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -289,10 +289,87 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
             keep_items
         }
 
+        // First, remove and store `impl` items, indexing them by the type they belong to.
+        let mut impls: HashMap<DefId, MovedDeclImpl> = HashMap::new();
+        FlatMapNodes::visit(krate, |mut item: P<Item>| {
+            if let Some((path, _)) = parse_source_header(&item.attrs) {
+                let header_ident = item.ident;
+                if let ItemKind::Mod(_, ModKind::Loaded(ref mut mod_items, _, _)) = &mut item.kind {
+                    mod_items.retain(|item| {
+                        if let ItemKind::Impl(r#impl) = &item.kind {
+                            // Only keep `impl` items with simple path types, and only if they
+                            // contain nothing but `const` items.
+                            fn impl_parent_decl<'hir>(
+                                cx: &'hir RefactorCtxt,
+                                r#impl: &Impl,
+                            ) -> Option<(DefId, &'hir hir::Path<'hir>)>
+                            {
+                                // Only inherent `impl`s that contain no items other than `const`.
+                                if r#impl.of_trait.is_some()
+                                    || !r#impl
+                                        .items
+                                        .iter()
+                                        .all(|item| matches!(item.kind, AssocItemKind::Const(..)))
+                                {
+                                    return None;
+                                };
+                                let ty = match cx.hir_map().find(r#impl.self_ty.id)? {
+                                    hir::Node::Ty(ty) => ty,
+                                    _ => return None,
+                                };
+                                let ty_path = match &ty.kind {
+                                    hir::TyKind::Path(hir::QPath::Resolved(None, path)) => path,
+                                    _ => return None,
+                                };
+                                match ty_path.res {
+                                    Res::Def(_, def_id) => Some((def_id, ty_path)),
+                                    _ => None,
+                                }
+                            }
+
+                            if let Some((decl_def_id, impl_type_path)) =
+                                impl_parent_decl(&self.cx, &r#impl)
+                            {
+                                match impls.entry(decl_def_id) {
+                                    Entry::Occupied(_) => warn!(
+                                        "decl {decl_def_id:?} ({impl_type_path:?}) \
+                                        has more than one `impl` block, dropping the others"
+                                    ),
+                                    Entry::Vacant(entry) => {
+                                        let parent_header =
+                                            HeaderInfo::new(header_ident, path.clone());
+                                        entry.insert(MovedDeclImpl {
+                                            item: item.clone(),
+                                            parent_header,
+                                        });
+                                    }
+                                }
+                            }
+
+                            false
+                        } else {
+                            true
+                        }
+                    });
+
+                    if mod_items.is_empty() {
+                        smallvec![]
+                    } else {
+                        smallvec![item]
+                    }
+                } else {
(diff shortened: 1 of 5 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -287,6 +287,67 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
             keep_items
         }
 
+        // First, remove and store `impl` items, indexing them by the type they belong to.
+        let mut impls: HashMap<DefId, MovedDeclImpl> = HashMap::new();
+        FlatMapNodes::visit(krate, |mut item: P<Item>| {
+            if let Some((path, _)) = parse_source_header(&item.attrs) {
+                let header_item = item.clone();
+                if let ItemKind::Mod(_, ModKind::Loaded(ref mut mod_items, _, _)) = &mut item.kind {
```
ahomescu, line 295: Could this be simplified by filtering directly for `ItemKind::Impl` here instead of whole modules? The code below already prunes the modules, no need to do it twice.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1085,10 +1158,23 @@ struct MovedDecl {
     namespace: Namespace,
     loc: Option<SrcLoc>,
     parent_header: HeaderInfo,
+    impl_: Option<MovedDeclImpl>,
```
ahomescu, line 1161:
```suggestion
    r#impl: Option<MovedDeclImpl>,
```

I don't particularly like the old hack of appending an underscore.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1085,10 +1158,23 @@ struct MovedDecl {
...
+}
+
+#[derive(Debug, Clone)]
+struct MovedDeclImpl {
```
ahomescu, line 1165: This could use some comments. It wasn't clear to me at first why the `impl_: Option<MovedDeclImpl>` below was needed, then figured it out: the `MovedDecl` is the actual type definition, and the `impl` is for the `const`s attached to it.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1085,10 +1158,23 @@ struct MovedDecl {
...
 }
 
 impl MovedDecl {
-    fn new<T>(decl: T, def_id: DefId, namespace: Namespace, parent_header: HeaderInfo) -> Self
+    fn new<T>(
+        decl: T,
+        def_id: DefId,
+        namespace: Namespace,
+        parent_header: HeaderInfo,
+        impl_: Option<MovedDeclImpl>,
```
ahomescu, line 1176: At least a short comment would be useful here, something like "the `impl` attached to this type".

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -287,6 +287,67 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                                    _ => return None,
+                                };
+                                match ty_path.res {
+                                    Res::Def(_, def_id) => Some(def_id),
+                                    _ => None,
+                                }
+                            }
+
+                            if let Some(impl_type_def_id) = impl_type_def_id(&self.cx, &impl_) {
+                                impls.entry(impl_type_def_id).or_insert_with(|| {
```
ahomescu, line 325: I'd like to know if the entry already exists, and at least log that if not assert or panic.

ahomescu (changes requested):

ahomescu:
Thanks for figuring this out, the fix looks good generally (modulo a couple of minor comments).

c2rust-refactor/src/transform/reorganize_definitions.rs
Rua: It needs the `path` to pass to `HeaderInfo`, right?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -287,6 +287,79 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
             keep_items
         }
 
+        // First, remove and store `impl` items, indexing them by the type they belong to.
+        let mut impls: HashMap<DefId, MovedDeclImpl> = HashMap::new();
+        FlatMapNodes::visit(krate, |mut item: P<Item>| {
+            if let Some((path, _)) = parse_source_header(&item.attrs) {
+                let header_item = item.clone();
```
ahomescu: We only need this for `header_item.ident`, could we borrow or clone just that instead? This has also been bothering me in the old code, but I never got around to fixing it.

c2rust-refactor/src/transform/reorganize_definitions.rs
ahomescu: Good point, and it also needs `header_item.ident`. I think you're right, we need to keep the current structure.

ahomescu (approved):

c2rust-refactor/src/transform/reorganize_definitions.rs
Rua: Yep, that works!

c2rust-refactor/tests/snapshots/reorganize_definitions.rs
kkysen, line 1: The `reorganize_definitions.rs` test doesn't compile after refactoring, so could we move this addition to its own test?  Like how there's `reorganize_foreign_types.rs`.

Rua: Huh? What's making it not compile?

kkysen: I forget, but there are some other errors.

Rua: There are no errors in CI though, and that's all I have to go by.

kkysen: It's disabled: https://github.com/immunant/c2rust/blob/80bb8c354ab373b80c8095c42976e32fe8abef31/c2rust-refactor/tests/snapshots.rs#L444

kkysen (approved):

PR #1751, opened 2026-04-23
transpile: Fix snapshot tests

Rua:
#1727 still had errors when it was merged.


kkysen (approved):
Oops, sorry I missed that.  There was some other `related_decls` CI error, too, so I got confused.

PR #1753, opened 2026-04-23
Disable related_decls in CI

randomPoison:
Currently related_decls is failing in CI because of changes in the latest Rust toolchain. There's a PR for a fix in the CRISP repo (https://github.com/GaloisInc/Tractor-Crisp/pull/106), but in the meantime this is broken locally for us. I think if the development of the tool is mainly happening in the CRISP repo and we're just vendoring a copy here then we don't need to be testing this in CI (similar to how we wouldn't test a crates.io crate in our CI).

I've left the other tools enabled for now, but if the others are similarly being developed in the CRISP repo then we might want to remove those from our CI as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,7 @@ jobs:
       matrix:
         crate:
           - merge_rust
+          # - related_decls
           - rust_util
           - split_ffi_entry_points
           - split_rust
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,7 +187,6 @@ jobs:
       matrix:
         crate:
           - merge_rust
-          - related_decls
           - rust_util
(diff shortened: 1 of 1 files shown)


kkysen (commented):
I'd just comment it out for now, so we know what we'd like to re-enable eventually.  Otherwise, I think this makes sense to me.

randomPoison:
I don't think this is temporary, I'm thinking that we don't want to be doing CI for the CRISP tools here. We should treat them like 3rd-party crates that we're just vendoring into this repo, and not something that we want to be doing any development on here.

That said, it also wouldn't be a big deal to leave that commented out. I've put that comment back.

randomPoison:
@kkysen can you hit the approve button when you have a sec? I need an actual approval before I can merge.

kkysen (approved):

PR #1756, opened 2026-04-24
transpile: Minor improvements of cast handling

Rua:
Two unrelated changes that modify the same code, so they are bundled. Double-translating is generally bad, because translation can have side-effects on the translator, like adding a feature or importing a type.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3628,11 +3628,9 @@ impl<'c> Translation<'c> {
                 // But for literals, if we don't absolutely have to cast, we would rather the
                 // literal is translated according to the type we're expecting, and then we can
                 // skip the cast entirely.
-                if let (Some(ty), CExprKind::Literal(_ty, lit)) =
-                    (override_ty, &self.ast_context[expr].kind)
-                {
-                    if !is_explicit && self.literal_matches_ty(lit, ty) {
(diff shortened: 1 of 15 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -3654,24 +3664,21 @@ impl<'c> Translation<'c> {
                     // But for literals, if we don't absolutely have to cast, we would rather the
                     // literal is translated according to the type we're expecting, and then we can
                     // skip the cast entirely.
-                    if let (Some(ty), CExprKind::Literal(_ty, lit)) =
-                        (override_ty, &self.ast_context[expr].kind)
-                    {
-                        if self.literal_matches_ty(lit, ty) {
-                            return self.convert_expr(ctx, expr, override_ty);
+                    if let CExprKind::Literal(_ty, lit) = &self.ast_context[expr].kind {
+                        if self.literal_matches_ty(lit, target_ty) {
```
ahomescu: Are truncations correctly covered here? For example `1000 as c_char as c_int` is different from `1000 as c_int`.

Rua: It shouldn't be possible for code to have a literal that's too big to fit into its type. If I try using `0xfffffffffffffffffL` in some C code, LLVM errors out.

ahomescu (approved):

PR #1759, opened 2026-04-28
transpile: Do not swap negation and literal if literal is in a macro

Rua:
A small addition to #1686 so that the swapping does not happen if the literal occurs inside a macro. Because otherwise that bypasses macro expansion, which is not desirable.


fw-immunant (approved):
Good catch, thanks for the test coverage.

PR #1761, opened 2026-04-28
transpile: Use `null_mut` in statics/const contexts too

Rua:
Apparently this just compiles, so the code that made an exception for this case was not (no longer?) needed.


kkysen (approved):
Hmm, I had thought we needed this and tested it, but I guess not.

PR #1763, opened 2026-04-30
transpile: In `name_reference`, also check expression for side effects

Rua:
- Depends on #1764 

Strange that this simple oversight was not caught sooner, despite already being clearly visible in the snapshot tests...


ahomescu:
I merged the dependency, does this need a rebase?

ahomescu (approved):

ahomescu (approved):

PR #1764, opened 2026-04-30
transpile: Use raw pointers for lvalues with side effects

Rua:
Named references needed to be removed from binary conditionals because it prevented this from working. Named references are intended for lvalues, but binary conditionals are rvalues. That means they can't be raw-borrowed, but can be stored by value.


ahomescu:
Is there no snapshot test that covers this? If not, could you add one?

ahomescu (changes requested):
See previous comment on snapshot test.

Rua:
There is, but because of #1763 it doesn't trigger. I can add an additional test that does trigger this code path currently.

c2rust-transpile/tests/snapshots/conditions.c
```diff
@@ -83,3 +83,22 @@ int nested_early_returns(int x) {
     }
     return x;
 }
+
+void conditional_operator(const unsigned sz, int buf[const]) {
+    int *x = &buf[0], *y = &buf[1];
+    *(0 ? &y : &x) = (int *) 10;
```
ahomescu: Do `x` and `y` need to be `int*`? Wouldn't the following be just as good?

```suggestion
    int x = 0, y = 1;
    *(0 ? &y : &x) = 10;
```

c2rust-transpile/tests/snapshots/conditions.c
```diff
@@ -83,3 +83,22 @@ int nested_early_returns(int x) {
     }
     return x;
 }
+
+void conditional_operator(const unsigned sz, int buf[const]) {
```
ahomescu, line 87: You could leave the old `if_else.c` alone and add just your functions to the new `conditions.c` file.

ahomescu (changes requested):

c2rust-transpile/tests/snapshots/conditions.c
ahomescu: Ah I see, the old `if_else.c` is not in `snapshots.rs`, is that right? Okay then it makes sense.

c2rust-transpile/tests/snapshots/conditions.c
Rua: Then no `ref mut` is generated anymore. Admittedly I can't think of any more straightforward way to trigger `ref mut` generation, this is just what I found in the unit tests.

Rua: Thought of some tests, added them to `exprs.c`.

c2rust-transpile/src/translator/named_references.rs
```diff
@@ -124,11 +124,12 @@ impl<'c> Translation<'c> {
 
                 let ptr_name = self.renamer.borrow_mut().fresh();
 
-                // let ref mut p = lhs;
+                // let p = &raw mut lhs;
```
ahomescu, line 127: I wouldn't be against switching for `let ref mut` to `let p = &mut lhs`, but why raw pointers? That adds extra unsafety that wasn't even there in the C code.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -187,7 +187,7 @@ impl<'c> Translation<'c> {
             )))
         } else {
             let lhs = self.convert_cast(
-                ctx,
+                ctx.used(),
```
ahomescu, line 215: Shouldn't the assignment operator already have this? If not, maybe setting it in the caller would be better.

c2rust-transpile/src/translator/operators.rs
ahomescu: Couldn't this be `ctx.used()`? Then you wouldn't need the other in the callee, like my other comment mentions.

ahomescu (changes requested):

c2rust-transpile/src/translator/named_references.rs
Rua: It can lead to borrow checker errors in some cases. #1763 had CI errors without this change, which is what prompted making it in the first place. `ref mut` was borrowing an array element, but that ended up borrowing the whole array, which meant that later code that accessed a different array element was causing a conflict. In principle that possibility has always existed, but fixing the side-effect detection exposed the underlying issue.

c2rust-transpile/src/translator/operators.rs
Rua: The assignment operator's `used` refers to the result of the assignment expression itself. Like in `foo = (bar += baz)`. Whereas the `used` that were added here refer to the individual subexpressions that are being generated as part of translating the (compound) assignment.

c2rust-transpile/src/translator/operators.rs
Rua: It could, but putting it on the `convert_cast` call directly matches the other code and generally makes it more explicit that it's there.

c2rust-transpile/src/translator/named_references.rs
Rua: Also, separately, full borrows can be bad for UB reasons, because they can trigger pointer invalidation. There are issues with that more generally that can't be easily addressed (see #301), but this improves the situation a little at least.

c2rust-transpile/src/translator/named_references.rs
```diff
@@ -122,13 +122,14 @@ impl<'c> Translation<'c> {
             } else {
                 // This is the case where we explicitly need to factor out possible side-effects.
 
-                let ptr_name = self.renamer.borrow_mut().fresh();
+                let ptr_name = self.renamer.borrow_mut().pick_name("c2rust_side_effects");
```
ahomescu:
```suggestion
                let ptr_name = self.renamer.borrow_mut().pick_name("c2rust_lvalue");
```

c2rust-transpile/src/translator/named_references.rs
ahomescu: In that case we can keep the pointers, but a comment explaining why would be really useful.

ahomescu (changes requested):

c2rust-transpile/src/translator/named_references.rs
Rua: It's not technically the lvalue, but a pointer to it. Would `c2rust_lvalue_ptr` make more sense?

Asking mainly because `c2rust_lvalue` is already used for storing a compound literal as an lvalue, in cases where its address is taken. I don't know if it's desirable here to use the same name for a different meaning.

ahomescu: `c2rust_lvalue_ptr` sounds good.

ahomescu (approved):

PR #1766, opened 2026-04-30
transpile: Refactor `WithStmts` and add new helper methods

Rua:
`WithStmts` has methods to map to `Expr`, `Result<Expr>`, and `Result<WithStmts>`, but a method to map to `WithStmts` alone was missing. I found myself wanting such a method quite a few times. So this adds one. In order to make the naming more consistent, methods that return `Result` now include `try` in the name, fitting how things are named in the Rust standard library. That then frees up the name `and_then` to be used for the missing method.

`WithStmts::zip` helps keep code cleaner, as it reduces the amount of nested closures that are otherwise needed.

### DIFF
--- a/c2rust-transpile/src/translator/atomics.rs
+++ b/c2rust-transpile/src/translator/atomics.rs
@@ -163,14 +163,14 @@ impl<'c> Translation<'c> {
         }
 
         match name {
-            "__atomic_load" | "__atomic_load_n" | "__c11_atomic_load" => ptr.and_then(|ptr| {
+            "__atomic_load" | "__atomic_load_n" | "__c11_atomic_load" => ptr.and_then_try(|ptr| {
                 let order = static_order(order);
 
                 let atomic_load = self.atomic_intrinsic_expr("load", &[order]);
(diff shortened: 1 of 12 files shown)


c2rust-transpile/src/translator/atomics.rs
```diff
@@ -354,8 +341,8 @@ impl<'c> Translation<'c> {
                         .kind
                         .get_qual_type()
                         .ok_or_else(|| format_err!("bad val1 type"))?;
-                    ptr.and_then(|ptr| {
-                        val.and_then(|val| {
+                    ptr.and_then_try(|ptr| {
```
ahomescu: Could this also be a `zip`?

ahomescu (changes requested):

c2rust-transpile/src/translator/atomics.rs
Rua: Yes, added!

ahomescu (approved):

ahomescu:
It needs a rebase.

Rua:
Done.

PR #1770, opened 2026-05-03
transpile: Split off `can_propagate_cast` function, elide some more casts

### LINKED ISSUE #128: Enum comparisons receive superflous and verbose casts
```c
typedef enum {
    FIND, ENTER
} ACTION;

int main(void) {
    ACTION action = FIND;
    if (action == FIND) {
        return 2;
    }

    return 1;
}
```

becomes

```rust
pub type ACTION = libc::c_uint;
pub const ENTER: ACTION = 1;
pub const FIND: ACTION = 0;
    let mut action: ACTION = FIND;
    if action as libc::c_uint == FIND as libc::c_int as libc::c_uint {
        return 2i32
    }
    return 1i32;
}
```

This does seem to be true to the AST (but is unnecessarily verbose):
```rust
  `-CompoundStmt 0x5293368 <col:16, line:12:1>
    |-DeclStmt 0x52931d0 <line:6:5, col:25>
    | `-VarDecl 0x5293130 <col:5, col:21> col:12 used action 'ACTION':'ACTION' cinit
    |   `-ImplicitCastExpr 0x52931b8 <col:21> 'ACTION':'ACTION' <IntegralCast>
    |     `-DeclRefExpr 0x5293190 <col:21> 'int' EnumConstant 0x5292dd0 'FIND' 'int'
    |-IfStmt 0x52932f8 <line:7:5, line:9:5>
    | |-<<<NULL>>>
    | |-<<<NULL>>>
    | |-BinaryOperator 0x5293280 <line:7:9, col:19> 'int' '=='
    | | |-ImplicitCastExpr 0x5293250 <col:9> 'unsigned int' <IntegralCast>
    | | | `-ImplicitCastExpr 0x5293238 <col:9> 'ACTION':'ACTION' <LValueToRValue>
    | | |   `-DeclRefExpr 0x52931e8 <col:9> 'ACTION':'ACTION' lvalue Var 0x5293130 'action' 'ACTION':'ACTION'
    | | `-ImplicitCastExpr 0x5293268 <col:19> 'unsigned int' <IntegralCast>
    | |   `-DeclRefExpr 0x5293210 <col:19> 'int' EnumConstant 0x5292dd0 'FIND' 'int'
```

Rua:
- Depends on #1620
- Fixes #128

### DIFF
--- a/c2rust-transpile/src/translator/enums.rs
+++ b/c2rust-transpile/src/translator/enums.rs
@@ -6,7 +6,7 @@ use crate::{
     diagnostics::TranslationResult,
     translator::{signed_int_expr, ConvertedDecl, ExprContext, Translation},
     with_stmts::WithStmts,
-    CDeclKind, CEnumConstantId, CEnumId, CExprId, CExprKind, CQualTypeId, CTypeKind, ConstIntExpr,
+    CDeclKind, CEnumConstantId, CEnumId, CQualTypeId, CTypeId, CTypeKind, ConstIntExpr,
 };
 
 impl<'c> Translation<'c> {
@@ -76,6 +76,10 @@ impl<'c> Translation<'c> {
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         let val = self.enum_constant_expr(enum_constant_id);
 
+        if self.enum_constant_matches_type(expr_type_id.ctype, enum_constant_id) {
+            return Ok(WithStmts::new_val(val));
+        }
+
         // Add a cast to the expected integral type.
         let enum_id = self.ast_context.parents[&enum_constant_id];
         self.convert_cast_from_enum(ctx, enum_id, expr_type_id, val)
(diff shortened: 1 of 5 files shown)


ahomescu (changes requested):
Needs a rebase.

ahomescu (approved):

c2rust-transpile/src/translator/enums.rs
```diff
@@ -208,13 +189,17 @@ impl<'c> Translation<'c> {
         mk().call_expr(mk().ident_expr(enum_name), vec![value])
     }
 
-    fn is_variant_of_enum(&self, enum_id: CEnumId, enum_constant_id: CEnumConstantId) -> bool {
-        let variants = match self.ast_context[enum_id].kind {
-            CDeclKind::Enum { ref variants, .. } => variants,
-            _ => panic!("{:?} does not point to an `enum` declaration", enum_id),
+    pub fn enum_constant_matches_type(
```
ahomescu: Does this need to be `pub`?

c2rust-transpile/src/translator/enums.rs
```diff
@@ -74,11 +74,16 @@ impl<'c> Translation<'c> {
         expr_type_id: CQualTypeId,
         enum_constant_id: CEnumConstantId,
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
-        let val = self.enum_constant_expr(enum_constant_id);
+        let mut val = WithStmts::new_val(self.enum_constant_expr(enum_constant_id));
 
-        // Add a cast to the expected integral type.
-        let enum_id = self.ast_context.parents[&enum_constant_id];
-        self.convert_cast_from_enum(ctx, enum_id, expr_type_id, val)
+        if !self.enum_constant_matches_type(expr_type_id.ctype, enum_constant_id) {
```
ahomescu: This could be an inverted early return, and then the rest of the function could stay like it was.

ahomescu (changes requested):
There is a regression in this PR: for the following C code:
```C
enum Bar { BarN1 = -1, Bar0, Bar1 };
typedef enum Bar BarT;

void f(void) {
    BarT b = Bar1;
    b = Bar0;
}
```
the old code emits
```rust
let mut b: BarT = Bar1;
b = Bar0;
```
but this PR produces
```rust
let mut b: BarT = Bar(Bar1.0);
b = Bar(Bar0.0);
```

Is this intentional?

c2rust-transpile/src/translator/enums.rs
Rua: I made it `pub(crate)` now. I think in general most `pub` items in the transpiler could be `pub(crate)`?

c2rust-transpile/src/translator/enums.rs
Rua: Done.

Rua:
No, that was not intentional. Fixed it now. Not sure why I made the change that caused that in the first place.

ahomescu (approved):

PR #1774, opened 2026-05-03
transpile: Cast character literals to provided type instead of `i32`

Rua:
There doesn't seem to be any reason to cast these to `i32` when they are generally `int` in C. It results in a type mismatch with the rest of the code if `c_int` is ever not defined as `i32`.


ahomescu:
> There doesn't seem to be any reason to cast these to `i32` when they are generally `int` in C.

Taking a guess here: the Rust type of character literals in Rust is `char`, and that's not always equivalent to `u32` or `i32`. For example, `u16::TryFrom<u32>` [was added in Rust 1.34](https://doc.rust-lang.org/std/primitive.u16.html#impl-TryFrom%3Cu32%3E-for-u16), but `u16::TryFrom<char>` [didn't happen until 1.74](https://doc.rust-lang.org/std/primitive.u16.html#impl-TryFrom%3Cchar%3E-for-u16).

ahomescu (approved):

Issue #1775, opened 2026-05-03
`case` statements do not preserve enum constants


Rua:
When matching on an enum, the `match` arms are translated with integer literals, rather than with enum constants. Named constants are important for the readability of the code, so preserving those would really help.

As far as I can tell, there are multiple reasons that all contribute to this happening:
- In C/Clang, enums get cast to their integral type in a `switch`. Thus, both the scrutinee and the case expressions are all `c_uint` generally.
- C2rust always translates `ConstantExpr` with the evaluated constant value, never with the original expression.
- Translations of arbitrary expressions can contain elements that are not compatible with `match` arms, such as casts.

PR #1776, opened 2026-05-04
transpile: Improve `case` translation

Rua:
- Depends on #1770
- Depends on #1774

Translating `case` statements is currently a very ad-hoc affair, so this is an attempt to make it a bit more generic and flexible. `expr_to_pat` now converts expressions to equivalent patterns, preserving the semantics and returning `None` if an expression with no equivalent is encountered.

Translation functions can now use `is_pattern` to detect if they are translating an expression that's meant to appear inside a Rust pattern, and adjust accordingly. That will help towards getting #1775 to work. Currently, all expressions that cannot (conservatively) be translated in pattern context return a translation error. That will then trigger fallback to directly using the `ConstIntExpr` value that's provided inside the `Case` value.

Additionally, it seems that it fixes a missing macro expansion in `macrocase.c`, so that's a bonus!

### DIFF
--- a/c2rust-transpile/src/cfg/mod.rs
+++ b/c2rust-transpile/src/cfg/mod.rs
@@ -1942,7 +1942,7 @@ impl CfgBuilder {
                 let pat = result
                     .and_then(|_| {
                         translator
-                            .convert_expr(ctx.used(), resolved, None)
+                            .convert_expr(ctx.const_().pattern().used(), resolved, None)
                             .map_err(|err| ("convert_expr", err.to_string()))
                     })
                     .and_then(|val| {
--- a/c2rust-transpile/src/cfg/mod.rs
+++ b/c2rust-transpile/src/cfg/mod.rs
@@ -1935,30 +1935,38 @@ impl CfgBuilder {
 
                 // Case
                 let resolved = translator.ast_context.unwrap_cast_expr(case_expr);
-                let branch = match translator.ast_context.index_unwrap_parens(resolved).kind {
-                    CExprKind::Literal(..) | CExprKind::ConstantExpr(_, _, Some(_)) => {
-                        match translator
-                            .convert_expr(ctx.used(), resolved, None)?
-                            .to_pure_expr()
-                        {
-                            Some(expr) => match *expr {
-                                Expr::Lit(lit) => Some(mk().lit_pat(lit.lit)),
-                                Expr::Path(path) => Some(mk().path_pat(path.path, path.qself)),
-                                _ => None,
-                            },
-                            _ => None,
-                        }
-                    }
-                    _ => None,
+                let result = match translator.ast_context.index_unwrap_parens(resolved).kind {
+                    CExprKind::Literal(..) | CExprKind::ConstantExpr(_, _, Some(_)) => Ok(()),
+                    _ => Err(("match", "wrong CExprKind".to_string())),
                 };
+                let pat = result
+                    .and_then(|_| {
+                        translator
+                            .convert_expr(ctx.used(), resolved, None)
+                            .map_err(|err| ("convert_expr", err.to_string()))
+                    })
+                    .and_then(|val| {
+                        val.to_pure_expr()
+                            .ok_or_else(|| ("to_pure_expr", "".to_string()))
+                    })
+                    .and_then(|expr| match *expr {
+                        Expr::Lit(lit) => Ok(mk().lit_pat(lit.lit)),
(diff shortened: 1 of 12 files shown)


c2rust-transpile/src/cfg/mod.rs
```diff
@@ -2415,3 +2411,173 @@ impl Cfg<Label, StmtOrDecl> {
...
+            Ok(Pat::TupleStruct(PatTupleStruct {
+                attrs,
+                qself,
+                path,
+                paren_token,
+                elems,
+            }))
+        }
+
+        Expr::Lit(ExprLit { attrs, lit }) => Ok(Pat::Lit(PatLit { attrs, lit })),
```
ahomescu: According to https://github.com/dtolnay/syn/blob/master/src/pat.rs#L12 some of these are just aliases, e.g., `ExprLit as PatLit`. You could simplify some of these arms.

c2rust-transpile/src/cfg/mod.rs
```diff
@@ -2415,3 +2411,173 @@ impl Cfg<Label, StmtOrDecl> {
         Ok(())
     }
 }
+
+fn expr_to_pat(expr: Expr) -> Result<Pat, String> {
```
ahomescu: These aren't exactly CFG-specific, maybe they should go in `rust_ast` or `c2rust-ast-builder`?

c2rust-transpile/src/cfg/mod.rs
```diff
@@ -2415,3 +2411,173 @@ impl Cfg<Label, StmtOrDecl> {
...
+    use syn::{
+        ExprArray, ExprCall, ExprLit, ExprParen, ExprPath, ExprReference, ExprStruct, ExprTuple,
+        ExprUnary, FieldPat, FieldValue, LitInt, PatLit, PatParen, PatPath, PatReference, PatSlice,
+        PatStruct, PatTuple, PatTupleStruct, UnOp,
+    };
+
+    match expr {
```
ahomescu: There is some speculative generality here which should be fine for this patch, but generally is better avoided.

c2rust-transpile/src/cfg/mod.rs
```diff
@@ -2415,3 +2411,173 @@ impl Cfg<Label, StmtOrDecl> {
...
+            expr,
+        }) => {
+            let Expr::Lit(ExprLit {
+                attrs: _,
+                lit: Lit::Int(lit_int),
+            }) = *expr else {
+                return Err("`ExprUnary::expr` is not an `ExprLit` with `lit: Lit::Int`".into());
+            };
+
+            let repr = format!("-{}{}", lit_int.base10_digits(), lit_int.suffix());
```
ahomescu: If we ever pass a negative `lit_int` here, that will produce a double negative e.g. `--1` which is invalid Rust. It might be worth adding a comment or something.

ahomescu (approved):

c2rust-transpile/src/cfg/mod.rs
Rua: I've added a check so that it catches that.

c2rust-transpile/src/cfg/mod.rs
Rua: I've moved it to `rust_ast`.

c2rust-transpile/src/cfg/mod.rs
Rua: I've done that now, and added comments about the type aliases.

c2rust-transpile/src/translator/macros.rs
```diff
@@ -214,14 +214,24 @@ impl<'c> Translation<'c> {
...
+            match self.make_cast(ctx, CQualTypeId::new(macro_ty), expr_ty, val) {
+                Ok(new_val) => val = new_val,
+                Err(err) => {
+                    info!(
+                        "Could not convert cast of macro {} for {:?}: {}",
+                        self.renamer.borrow_mut().get(macro_id).unwrap(),
+                        expr_id,
+                        err
+                    );
+                    return Ok(None);
```
ahomescu, line 226: Should this be gated on `is_pattern`? It is a minor functional change.

ahomescu (approved):

c2rust-transpile/src/translator/macros.rs
Rua: I think it's a good change to have in general. It means that code that currently errors out for whatever reason will now be able to build, albeit with reduced macro conversion.

PR #1782, opened 2026-05-08
Add support for __auto_type

### LINKED ISSUE #1779: Support `__auto_type`
Linux uses `__auto_type` on pre-C23 compilers which is a gcc extension.

ahomescu:
Fixes #1779

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -674,6 +674,22 @@ private:
 
         VisitQualType(t);
     }
+
+    void VisitAutoType(const AutoType *T) {
+        if (!T->isGNUAutoType()) {
+            printDiag(Context, DiagnosticsEngine::Warning,
+                      "Encountered unexpected auto type",
+                      src_loc, src_range);
+            return;
+        }
+
+        auto t = T->desugar();
+        auto qt = encodeQualType(t);
+        encodeType(T, TagAutoType,
+                   [qt](CborEncoder *local) { cbor_encode_uint(local, qt); });
+
+        VisitQualType(t);
+    }
 };
 
 class TranslateASTVisitor final
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -760,6 +760,7 @@ impl TypedAstContext {
                 CDeclKind::Typedef { typ: ty, .. } => ty.ctype,
                 _ => panic!("Typedef decl did not point to a typedef"),
             },
+            Auto(ty) => ty,
             _ => return typ,
         };
         self.resolve_type_id(ty)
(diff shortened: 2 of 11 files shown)


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -674,6 +672,22 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
 
         VisitQualType(t);
     }
+
+    void VisitAutoType(const AutoType *T) {
+        if (!T->isGNUAutoType()) {
+            printDiag(Context, DiagnosticsEngine::Error,
+                      "auto types other than __auto_type are not supported. Aborting.",
+                      src_loc, src_range);
+            abort();
```
thedataking: nit: could also warn and return `TagTypeUnknown`. Isn't that more in line with how we otherwise handle unssupported features?

thedataking (approved):

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: Okay I'll change it tomorrow.

c2rust-transpile/tests/snapshots/auto_type.c
```diff
@@ -0,0 +1,4 @@
+int foo() {
+  __auto_type x = 42;
```
thedataking, line 6: should we also cover pointer to int, structs, typedefs? Fine to leave out if we'd only end up testing the type inference machinery in clang and not the transpiler.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -2674,6 +2679,10 @@ impl CTypeKind {
             Vector(_, _) => false,
             UnhandledSveType => false,
             Atomic(_) => false,
+            // TODO: the integer might be in range for the deduced type.
```
thedataking: I think there is only one callsite and we're already passing in a resolved type so how can we even get `Auto` here? Am I missing something? Caller is:

```rust
let ty_kind = &self.ast_context.resolve_type(ty.ctype).kind;
        match *lit {
            CLiteral::Integer(value, _) if ty_kind.is_integral_type() && !ty_kind.is_bool() => {
                ty_kind.guaranteed_integer_in_range(value)
            }
            // `convert_literal` always casts these to i32.
            CLiteral::Character(_value) => matches!(ty_kind, CTypeKind::Int32),
            CLiteral::Floating(value, _) if ty_kind.is_floating_type() => {
                ty_kind.guaranteed_float_in_range(value)
            }
            _ => false,
        }
```

ahomescu: That gets called from the converter for `ImplicitCastExpr` and `ExplicitCastExpr`. Considering the current snapshot looks like `let mut x: ::core::ffi::c_int = 42 as ::core::ffi::c_int;`, with an extra redundant cast, I think we do need to cover the `Auto` case somewhere.

thedataking: This is dead code as far as I can tell. I inserted a `panic!` to verify and ran
`cargo nextest run -p c2rust-transpile --test snapshots`. No failures, no panics.

ahomescu: I don't want to take this line out because then the compiler will complain about the missing case. At best it should be a panic, do you want that instead?

PR #1788, opened 2026-05-09
Support C11 generics

### LINKED ISSUE #1786: Support for C11 generics
The Linux kernel uses `_Generic` for a bunch of core primitives like `smp_load_acquire`, `READ_ONCE`, and more. Generics were added in C11 and c2rust technically only supports C99, but it would be useful to add these.

ahomescu:
Add support for C11 `_Generic` generics where the result can be determined by clang at compile time so we transpile the resolved expression directly. Fixes #1786

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -1501,7 +1501,13 @@ class TranslateASTVisitor final
     }
 
     bool VisitGenericSelectionExpr(GenericSelectionExpr *E) {
-        printDiag(Context, DiagnosticsEngine::Warning, "Encountered unsupported generic selection expression", E);
+        if (E->isResultDependent()) {
+            printDiag(Context, DiagnosticsEngine::Warning, "Encountered unsupported generic selection expression", E);
+            return true;
+        }
+
+        std::vector<void *> childIds{E->getResultExpr()};
+        encode_entry(E, TagGenericExpr, childIds);
         return true;
     }
 
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -2001,6 +2001,19 @@ impl ConversionContext {
                     self.expr_possibly_as_stmt(expected_ty, new_id, node, e)
                 }
 
+                ASTEntryTag::TagGenericExpr if expected_ty & (EXPR | STMT) != 0 => {
+                    let wrapped = node.children[0].expect("Expected generic expression");
+                    let ty_old = node.type_id.expect("Expected expression to have type");
+                    let ty = self.visit_qualified_type(ty_old);
+
+                    // Use the existing `Paren` kind instead of adding a new one
+                    // that would just be a no-op wrapper around the inner
+                    // expression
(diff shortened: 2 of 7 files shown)


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -1480,7 +1480,13 @@ class TranslateASTVisitor final
     }
 
     bool VisitGenericSelectionExpr(GenericSelectionExpr *E) {
-        printDiag(Context, DiagnosticsEngine::Warning, "Encountered unsupported generic selection expression", E);
+        if (E->isResultDependent()) {
```
thedataking, line 1504: note to self: this cannot return true in C; we'd have to be parsing C++ where the type of the controlling expression depends on a not-yet substituted template parameter. We support the `__auto_type` keyword in C now but the type of such variables are eagerly resolved.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -1986,6 +1986,19 @@ impl ConversionContext {
                     self.expr_possibly_as_stmt(expected_ty, new_id, node, e)
                 }
 
+                ASTEntryTag::TagGenericExpr if expected_ty & (EXPR | STMT) != 0 => {
```
ahomescu, line 2004: Technically we could have done even without this tag, and have the C++ code emit `TagParenExpr` directly. This is more future-proof though.

c2rust-transpile/tests/snapshots/generics.c
```diff
@@ -0,0 +1,10 @@
+#define inc(x) (_Generic((x), \
+  float: (x) + 1.0f,          \
+  double: (x) + 1.0,          \
+  default: (x) + 1))
+
+void foo() {
```
thedataking, line 15: Could we also test controlling expression with side effects (must not be evaluated per C11 §6.5.1.1)? 

```C
int n = 0;
_Generic((n++), int: n, default: n);  // n must end up 0, not 1
```

It would also be good to cover some of the Linux idioms:
- no default 
- pointer arms with const discrimination
```C
  struct s { int x; };
  struct t { int y; };
  #define cast(p) (_Generic((p), \
      const struct s *: (const struct t *)(p), \
      struct s *:       (struct t *)(p)))
  void f(struct s *p, const struct s *cp) {
      struct t *t1 = cast(p);
      const struct t *t2 = cast(cp);
  }
```
- generics inside of `typeof` (used for READ_ONCE, WRITE_ONCE):
```C
  // include/linux/compiler_types.h:
  #define __unqual_scalar_typeof(x) typeof(_Generic((x), \
      char: (char)0, \
      __scalar_type_to_expr_cases(char), \
      __scalar_type_to_expr_cases(short), \
      __scalar_type_to_expr_cases(int),   \
      __scalar_type_to_expr_cases(long),  \
      __scalar_type_to_expr_cases(long long), \
      default: (x)))
```

we could just test:
```C
 #define unqual_typeof(x) typeof(_Generic((x), \
      int: (int)0, const int: (int)0, default: (x)))
  void f(const int x) { unqual_typeof(x) y = x; }
```

Also consider file-scoped use `static int k = _Generic(0, int: 1, default: 0);`.

thedataking (approved):

PR #1792, opened 2026-05-10
ast-exporter: Export some macro arg expansions

Rua:
- Depends on #1795

The "chain" of expansions is a little different for macros with args (that is, functional macros). This PR does two things:
- Handles arg expansions differently, so that const macros that call functional macros are now expanded.
- Exports the arg info to the Rust side to be used to hopefully make functional macros possible later.

Not all macro args are currently handled by this because of technical limitations. This code is only able to "see" macro args in the macro where they were expanded into tokens. They are not visible in macros that take an arg and then pass it on to another macro call in an arg. Maybe a solution will eventually be found for that. For now at least, the AST on the Rust side contains expressions that are macro arg expansions!

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -1451,17 +1451,21 @@ class TranslateASTVisitor final
         }
 
         curMacroExpansionSource =
-            Lexer::getSourceText(ExpansionStack[0], Mgr, Context->getLangOpts());
+            Lexer::getSourceText(ExpansionStack[0].range, Mgr, Context->getLangOpts());
 
-        for (auto &ExpansionRange : ExpansionStack) {
+        for (auto &elem : ExpansionStack) {
+            if (elem.isParameter) {
+                continue;
+            }
+
             StringRef name;
-            MacroInfo *mac = getMacroInfo(ExpansionRange.getBegin(), name);
+            MacroInfo *mac = getMacroInfo(elem.range.getBegin(), name);
 
             if (!mac || mac->getNumTokens() == 0) {
                 return true;
             }
 
-            if (VisitMacro(name, ExpansionRange.getBegin(), mac, E)) {
+            if (VisitMacro(name, elem.range.getBegin(), mac, E)) {
                 curMacroExpansionStack.push_back(mac);
             }
         }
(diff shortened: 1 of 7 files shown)


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -723,13 +723,20 @@ class TranslateASTVisitor final
...
+        StringRef parameter;
+    };
 
     // This stores a raw encoding of the macro call site SourceLocation, since
     // SourceLocation isn't hashable.
-    std::unordered_set<unsigned> macroCallSites;
-    SmallVector<MacroInfo*, 1> curMacroExpansionStack;
-    StringRef curMacroExpansionSource;
+    // TODO: Can this be made to use `unordered_set` for speed?
+    std::set<std::pair<unsigned, const StringRef>> macroCallSites;
```
ahomescu, line 744: What do these `StringRef`s point into? Is there a potential lifetime issue here?

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -769,7 +776,7 @@ class TranslateASTVisitor final
     // Template required because Decl and Stmt don't share a common base class
     void encode_entry_raw(void *ast, ASTEntryTag tag, SourceRange loc,
                           const QualType ty, bool rvalue,
-                          bool isVaList, bool encodeMacroExpansions,
+                          bool isVaList, bool encodeMacroInvocations,
```
ahomescu, line 786: Could you explain why the rename from "expansions" to "invocations"? I'm guessing there's a subtle difference but I'm not sure what it is.

Rua: In my understanding, an invocation is the place where the macro name is written, the "call". And the expansion is the act or result of replacing the macro name with its output tokens. Most of the logic for macros works "backwards": the fully expanded macro, in the form of a C expression, is the start point, and c2rust is working to find the original macro invocation(s)/call(s) that produced it. So I felt that naming that "expansion" was confusing and not very descriptive.

Maybe there are better things to name these though?

c2rust-ast-exporter/src/AstExporter.cpp
Rua: They point to the Clang AST in the same way that `macros` does, which also holds a `StringRef`. I assume that both live as long as the AST is kept alive?

ahomescu (approved):

c2rust-ast-exporter/src/AstExporter.cpp
ahomescu: Just as a small nit: refactoring and renames like these make later rebases/merges can produce conflicts in later rebases/merges. It's not worth changing this PR though, it's fine as is.

PR #1798, opened 2026-05-12
Handle Clang counted_by types

### LINKED ISSUE #1382: c2rust panic when encounter `__counted_by__` attribute
reduced testcase

```c
struct {
  int length;
  unsigned char data[] __attribute__ ((__counted_by__ (length)));
} eed_4;
```

https://c.godbolt.org/z/1cMd7PveW

`__counted_by__` introduced in clang 18.

thedataking:
Encountered while transpiling code for the Linux kernel but this should be broadly useful.

Part of `-fbounds-safety`. RFC: https://discourse.llvm.org/t/rfc-enforcing-bounds-safety-in-c-fbounds-safety/70854 

> Kernel headers say __counted_by__ was introduced in clang 18, and counted_by_ptr in clang 22.

Implementation notes: 
- The AST types we need to handle this feature are not available in clang 18 (confirmed by CI failures),
- This PR does not handle the later `counted_by_ptr` attribute; at least not fully.

Fixes #1382.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -314,17 +314,7 @@ private:
     }
 
 #if CLANG_VERSION_MAJOR >= 19
-    void VisitCountAttributedType(const CountAttributedType *T) {
-        auto t = T->desugar();
-        auto qt = encodeQualType(t);
-
-        encodeType(T, TagAttributedType, [qt](CborEncoder *local) {
-            cbor_encode_uint(local, qt);
-            cbor_encode_null(local);
-        });
-
-        VisitQualType(t);
-    }
+    void VisitCountAttributedType(const CountAttributedType *T);
 #endif
 
     void VisitParenType(const ParenType *T) {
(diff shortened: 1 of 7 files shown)


fw-immunant:
Does this fix #1382?

thedataking:
> Does this fix #1382?

I think so. Updated the PR. Thanks!

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -313,6 +313,20 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
...
     }
 
+#if CLANG_VERSION_MAJOR >= 20
+    void VisitCountAttributedType(const CountAttributedType *T) {
+        auto t = T->desugar();
+        auto qt = encodeQualType(t);
+
+        encodeType(T, TagAttributedType, [qt](CborEncoder *local) {
+            cbor_encode_uint(local, qt);
+            cbor_encode_null(local);
```
ahomescu: Shouldn't this also encode the attribute, followed by the `getCountExpr`?

c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -313,6 +313,20 @@ class TypeEncoder final : public TypeVisitor<TypeEncoder> {
         VisitQualType(t);
     }
 
+#if CLANG_VERSION_MAJOR >= 20
```
ahomescu: Kernel headers say `__counted_by__` was introduced in clang 18, and `counted_by_ptr` in clang 22.

thedataking: `CountAttributedType` AST class was added in https://github.com/llvm/llvm-project/pull/78000 which landed March 20, 2024.

LLVM 18.1.0 (first major stable 18 release) was released on March 6, so the above PR must have missed the 18 release train. 

[PR 78000](https://github.com/llvm/llvm-project/pull/78000) says:
> This patch also extends the parsing logic to parse the counted_by argument 

So it sounds like clang 18 can parse `counted_by` but only 19+ has the sugar types we want to use.

Left a note re. `counted_by_ptr`.

c2rust-ast-exporter/src/AstExporter.cpp
thedataking: Done in a separate commit since it adds a `TagCountAttributedType` tag to store the additional information.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -887,6 +887,31 @@ impl ConversionContext {
                     self.processed_nodes.insert(new_id, TYPE);
                 }
 
+                TypeTag::TagCountAttributedType => {
+                    let ty_id = from_value(ty_node.extras[0].clone())
+                        .expect("Count attributed type child not found");
+                    let ty = self.visit_qualified_type(ty_id);
+
+                    let kind = match expect_opt_str(&ty_node.extras[1])
+                        .expect("Count attributed type kind not found")
```
ahomescu: Instead of the double `expect` with the duplicated error, how about `.ok().flatten().expect(...)`?

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -887,6 +887,31 @@ impl ConversionContext {
...
+                        "counted_by_or_null" => CountAttributedKind::CountedByOrNull,
+                        "sized_by_or_null" => CountAttributedKind::SizedByOrNull,
+                        other => panic!("Unknown CountAttributedType kind: {}", other),
+                    };
+
+                    let count_expr = expect_opt_u64(&ty_node.extras[2])
+                        .expect("Count attributed type count expr field missing")
+                        .map(|id| self.visit_expr(id as ClangId));
+
+                    let ty = CTypeKind::CountAttributed(ty, kind, count_expr);
```
ahomescu: Couldn't this be `CTypeKind::Attributed` with a new `Attribute::CountedBy` variant?

ahomescu (changes requested):

c2rust-transpile/src/c_ast/conversion.rs
thedataking: Yup. `.flatten().expect()` seems to do the trick since `expect_opt_str` doesn't return `Result`.

ahomescu (approved):

PR #1806, opened 2026-05-14
transpile: Allow macro conversion to succeed only for some expansions

Rua:
If there is ever a failure to convert one of the expansion sites of a macro, or it fails to fold into a single common type, then the entire macro is discarded and left in its expanded form. This PR adds a fallback mechanism: if the conversion succeeds for some of the expansion sites, then those will be converted to use the macro, and only the ones that failed will be left expanded.

The `expr_results` value stores all the errors of the conversion attempts for each expansion site. Currently, the `Ok` variant is just empty, but this is planned to hold information about functional macro arguments in the future.


ahomescu (changes requested):
Is there a test or snapshot you could add here?

PR #1810, opened 2026-05-16
transpile: Use strict provenance APIs for pointer casts

### LINKED ISSUE #1136: better handle strict provenance when it's stabilized
`#![feature(strict_provenance)]` should be stabilized soon (https://github.com/rust-lang/rust/pull/130350), and there are places where we can handle things better than provenance exposing `as` casts.  One such place is https://github.com/immunant/c2rust/pull/1134#pullrequestreview-2320843839, where a `bool` to ptr cast is always an invalid ptr, and so we know it's non-provenance preserving.

Rua:
- Depends on #1811
- Fixes #1136 

The strict provenance APIs changed since the currently used nightly 1.67, so I've selected the API based on the edition, and left comments for when things were changed. Casts involving function pointers are not covered by this, as they use transmutes, and it doesn't seem immediately obvious how to make that work.

### DIFF
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -1,5 +1,6 @@
 use c2rust_ast_builder::{mk, properties::Mutability};
 use c2rust_ast_exporter::clang_ast::LRValue;
+use c2rust_rust_tools::RustEdition;
 use failure::{err_msg, format_err};
 use syn::{BinOp, Expr, Type, UnOp};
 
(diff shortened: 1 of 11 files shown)


c2rust-transpile/src/translator/pointers.rs
```diff
@@ -483,17 +484,68 @@ impl<'c> Translation<'c> {
...
+                    self.use_feature("strict_provenance");
+
+                    // Rust 1.79: method name changed to `with_exposed_provenance[_mut]`
+                    match mutability {
+                        Mutability::Immutable => "from_exposed_addr",
+                        Mutability::Mutable => "from_exposed_addr_mut",
+                    }
+                }
+                RustEdition::Edition2024 => match mutability {
+                    Mutability::Immutable => "with_exposed_provenance",
```
ahomescu, line 535: Doesn't this require a previous call to `expose_provenance`?

ahomescu (approved):

c2rust-transpile/src/translator/pointers.rs
Rua: Yes, but the same is technically true of casts, too.

PR #1811, opened 2026-05-16
refactor: Don't remove raw borrows in `remove_unnecessary_refs`

Rua:
Small bugfix for a problem that showed itself in CI with #1810.


c2rust-refactor/src/transform/canonicalize_refs.rs
```diff
@@ -85,7 +85,7 @@ impl Transform for RemoveUnnecessaryRefs {
 
 fn remove_ref(expr: &mut P<Expr>) {
     match &expr.kind {
-        ExprKind::AddrOf(_, _, inner) => *expr = inner.clone(),
+        ExprKind::AddrOf(BorrowKind::Ref, _, inner) => *expr = inner.clone(),
```
ahomescu, line 88: This also covers `&pin` borrows, but I guess that's fine? The transpiler wouldn't emit those anyway.

ahomescu (approved):

PR #1815, opened 2026-05-20
transpile: support `__attribute__((cleanup(func)))`

### LINKED ISSUE #1814: Transpile `__cleanup__(func)` compiler attribute
C2Rust silently drops the cleanup attribute which ties a cleanup function that the compiler calls when a variable goes out of scope.

thedataking:
Translate `__attribute__((cleanup(func)))` using a `CleanupGuard<T>` whose `Drop` calls the user's cleanup function on scope exit.

Closes #1814.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -2403,6 +2403,7 @@ impl<'c> Translation<'c> {
                 ref ident,
                 initializer,
                 typ,
+                ref attrs,
                 ..
             } => {
                 assert!(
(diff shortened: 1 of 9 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -2460,6 +2461,31 @@ impl<'c> Translation<'c> {
...
+                            .renamer
+                            .borrow()
+                            .get(&fn_id)
+                            .expect("cleanup function not registered with renamer");
+                        let cleanup_ident = mk().ident(&cleanup_name);
+                        let var_ident = mk().ident(&rust_name);
+                        let guard_ident = mk().ident(&format!("_cleanup_{}", rust_name));
+                        syn::parse_quote! {
+                            let #guard_ident = CleanupGuard(
+                                &mut #var_ident as *mut _ as *mut _,
```
fw-immunant: Shouldn't we be using `&raw` or `addr_of!` here?

fw-immunant (approved):
Looks good overall, except for concerns around constructing reference to obtain raw pointer. I would have liked more context on several changes (e.g. the first which appears to be a longstanding bug for which a reproducer would be good to see) but they do check out.

c2rust-transpile/src/translator/mod.rs
thedataking: yes, fixed, thanks!

thedataking:
> I would have liked more context on [...] the first which appears to be a longstanding bug 

Ack on your overall point. For the first commit, it turns out that `def` can never be null (it is initialized to `VD`, which is never null in this function), so I changed it to simply remove the dead code. The null guard was most likely carried over from the `VisitFunctionDecl` handler, where `def` comes from `getDefinition()` and the null case is real.

Rua:
Since this got merged, I'm getting a ton of "unrecognised attribute" spam in the output. Was that intended?

PR #1817, opened 2026-05-21
transpile: Translate enums in `switch` as their own type instead of integers

### LINKED ISSUE #1775: `case` statements do not preserve enum constants

Rua:
- Depends on #1620
- Depends on #1770
- Depends on #1776
- Fixes #1775

Enum constant names are now preserved in match arms where present, which significantly improves code fidelity.

### DIFF
--- a/c2rust-transpile/src/cfg/mod.rs
+++ b/c2rust-transpile/src/cfg/mod.rs
@@ -448,7 +448,6 @@ impl GenTerminator<StructureLabel<StmtOrDecl>> {
 pub struct SwitchCases {
     cases: Vec<(Pat, Label)>,
     default: Option<Label>,
-    #[allow(unused)]
     expected_type_id: Option<CQualTypeId>,
 }
 
@@ -1936,8 +1935,19 @@ impl CfgBuilder {
                 self.add_wip_block(wip, Jump(this_label.clone()));
 
                 // Case
+                let switch_case = self.switch_expr_cases.last_mut().ok_or_else(|| {
+                    format_err!(
+                        "Cannot find the 'switch' wrapping this ({:?}) 'case' statement",
+                        stmt_id,
+                    )
+                })?;
+
                 let pat = translator
-                    .convert_expr(ctx.const_().pattern().used(), case_expr, None)
+                    .convert_expr_with_optional_cast(
+                        ctx.const_().pattern().used(),
+                        switch_case.expected_type_id,
+                        case_expr,
+                    )
                     .map_err(|err| ("convert_expr", err.to_string()))
                     .and_then(|val| {
                         val.to_pure_expr()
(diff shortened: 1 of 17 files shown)


ahomescu (changes requested):
This needs a rebase, but CI should pass now.

ahomescu:
I think there might be one codegen bug here: for this C code
```
enum Bar { BarN1 = -1, Bar0, Bar1 };   // integral type: int
void f(int x) { enum Bar b = x; }
```
the transpiled Rust code fails to compile. Let me know if you reproduce it.

Rua:
Yeah I could reproduce it. Fixed now, I hope.

ahomescu:
This needs another rebase.

ahomescu (changes requested):
Needs rebase.

c2rust-transpile/src/cfg/mod.rs
```diff
@@ -1934,6 +1936,13 @@ impl CfgBuilder {
                 self.add_wip_block(wip, Jump(this_label.clone()));
 
                 // Case
+                let switch_case = self.switch_expr_cases.last_mut().ok_or_else(|| {
+                    format_err!(
+                        "Cannot find the 'switch' wrapping this ({:?}) 'case' statement",
+                        stmt_id,
+                    )
+                })?;
+
```
fw-immunant, line 1944: I don't understand why this is moved earlier in the file than it was originally performed, with a variable name being assigned now. If this is necessary of preferable for some reason, it's a subtle one and deserves a comment. If not, please revert this hunk.

fw-immunant (commented):
A few small nits.

The first is commented inline on ``transpile: Add expected type for enums in `SwitchCases`­``, and the other is that I'm not a big fan of ``transpile: Make `target_type_id` parameter of `convert_expr_with_cast` optional``: it adds a branch to the function that makes the function not do what it says on the tin, which isn't what any of the existing callers want. Where ``transpile: Translate enums in `switch` as their own type instead of integers`` changes calls from `convert_expr` to the new `convert_expr_with_cast`, I'd prefer if it changed them to a new function (e.g. "`convert_expr_possibly_casting`" or such) that does that dispatch to either `convert_expr` or the old `convert_expr_with_cast`.

Otherwise, this LGTM and is a really clean idiomaticity win. Thanks for taking the numerous steps to make this improved translation happen!

This also seems to want a rebase now.

c2rust-transpile/src/cfg/mod.rs
Rua: It needs to be earlier because `switch_case` is now used below when calculating `pat`.

fw-immunant: I only see two references to `switch_case` here: the definition above, and the usage at the original definition site, after the definition of `pat`. Am I mistaken?

fw-immunant: Oh, the usage is added later, in commit ``transpile: Translate enums in `switch` as their own type instead of integers``. The retrocausality is confusing here but the change does end up being necessary.

Rua: Sorry, I see where the confusion came from, because the code was moved in a different commit than the one that added the extra uses. I've changed that now.

Rua:
I've now added `convert_expr_with_optional_cast` instead.

c2rust-transpile/src/cfg/mod.rs
fw-immunant: Thanks!

fw-immunant (approved):
Looks great now, thanks!

ahomescu (approved):

PR #1820, opened 2026-05-22
refactor: In `alter_ty_path`, return on failed unpack

### LINKED ISSUE #1818: Refactorer crashes if the processed code has `derive(PartialEq, Eq)`
Seems to be triggered by CI in #1817:

> thread 'main' panicked at 'expected TyKind::Path(qself, path), got Rptr(None, MutTy { ty: Ty { id: NodeId(100617), kind: Path(None, Path { span: Span { lo: BytePos(129373), hi: BytePos(129382), ctxt: #57 }, segments: [PathSegment { ident: C2Rust_Unnamed#0, id: NodeId(100618), args: None }], tokens: None }), span: Span { lo: BytePos(129373), hi: BytePos(129382), ctxt: #57 }, tokens: None }, mutbl: Not })', c2rust-refactor/src/path_edit.rs:127:17

Rua:
- Fixes #1818

This is *a* fix, but I don't know if it's actually the *correct* fix.

### DIFF
--- a/c2rust-refactor/src/path_edit.rs
+++ b/c2rust-refactor/src/path_edit.rs
@@ -124,7 +124,7 @@ where
                     _ => {}
                 }
 
-                unpack!([&mut t.kind] TyKind::Path(qself, path));
+                unpack!([&mut t.kind] TyKind::Path(qself, path); return);
                 let (new_qself, new_path) =
                     self.handle_qpath(id, qself.clone(), path.clone(), qpath);
                 *qself = new_qself;
--- a/c2rust-refactor/src/macros.rs
+++ b/c2rust-refactor/src/macros.rs
@@ -13,7 +13,14 @@ macro_rules! match_or_else {
     ([$e:expr] $($arm_pat:pat => $arm_body:expr),*; $or_else:expr) => {
         match $e {
             $( $arm_pat => $arm_body, )*
-            ref x @ _ => $or_else(x),
+            _ => $or_else,
+        }
+    };
+
+    ([$e:expr] $($arm_pat:pat => $arm_body:expr),*; $or_else_var:ident => $or_else:expr) => {
+        match $e {
+            $( $arm_pat => $arm_body, )*
+            ref $or_else_var @ _ => $or_else,
         }
     };
 }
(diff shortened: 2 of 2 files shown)


c2rust-refactor/src/path_edit.rs
```diff
@@ -124,7 +124,7 @@ where
                     _ => {}
                 }
 
-                unpack!([&mut t.kind] TyKind::Path(qself, path));
+                unpack!([&mut t.kind] TyKind::Path(qself, path); return);
```
ahomescu, line 127: Silencing errors isn't really the way to fix issues, let's investigate this.

Any idea why a type would appear as `hir::TyKind::Path` in the HIR, but as an `Rptr` in the AST? Maybe an alias? Could you print out which type this is?

Rua: I printed out `qpath` when `t.kind` isn't `TyKind::Path`:

```rust
if !matches!(t.kind, TyKind::Path(..)) {
    panic!("{qpath:?}");
}
```
and got:

```rust
Resolved(
    None,
    Path {
        span: Span {
            lo: BytePos(129233),
            hi: BytePos(129242),
            ctxt: #57
        },
        res: PrimTy(
            Bool
        ),
        segments: [
            PathSegment {
                ident: bool#57,
                hir_id: HirId {
                    owner: OwnerId {
                        def_id: DefId {
                            krate: crate0,
                            index: DefIndex(
                                50254
                            )
                        }
                    },
                    local_id: 25
                },
                res: PrimTy(
                    Bool
                ),
                args: None,
                infer_args: false
            }
        ]
    }
)
```

So it's... `bool`? Not sure how to make sense of that one.

Rua: Random thought: if the refactorer is making modifications to the AST, what happens to the HIR? Is it modified in parallel? If not, what keeps the two in sync? Could that be the issue here?

ahomescu: The HIR is not preserved. The refactorer is expected to make independent modifications, or reload the crate if it makes changes that impact the HIR (and needs it to be correct).

Rua: So there's a bug with that elsewhere in the refactorer, that's making the two go out of sync?

ahomescu: Is this before or after any transform? If before, then yes the AST and HIR should match.

ahomescu: I looked into this a bit myself. The most likely cause is a span value mismatch between the `span_to_hir_map` and `span_to_node_id_map` for the expansion of `#[derive(PartialEq)]`.

ahomescu: Added a repro test and root cause fix at https://github.com/immunant/c2rust/pull/1967.

Rua: Thank you!

PR #1822, opened 2026-05-23
Use raw identifier for reserved keywords (continued)

cydparser:
I've rebased and updated an old PR: https://github.com/immunant/c2rust/pull/295

### DIFF
--- a/c2rust-transpile/src/translator/structs_unions.rs
+++ b/c2rust-transpile/src/translator/structs_unions.rs
@@ -283,6 +283,7 @@ impl<'a> Translation<'a> {
                     field_name,
                     bytes,
                     attrs,
+                    single_bitfield: _,
                 } => {
                     let ty = mk().array_ty(
                         mk().ident_ty("u8"),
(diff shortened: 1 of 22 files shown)


c2rust-transpile/src/translator/structs_unions.rs
```diff
@@ -283,6 +283,7 @@ impl<'a> Translation<'a> {
                     field_name,
                     bytes,
                     attrs,
+                    ..
```
ahomescu: What else is here? If it's only one or two extra fields, I'd add them explicitly.

c2rust-transpile/src/translator/structs_unions.rs
```diff
@@ -962,7 +968,19 @@ impl<'a> Translation<'a> {
         }
 
         // Find leftover bitfield group at end: it's all set
-        if let Some(field_group) = last_bitfield_group.take() {
+        if let Some(mut field_group) = last_bitfield_group.take() {
+            if let FieldType::BitfieldGroup {
+                ref mut field_name,
+                single_bitfield,
```
ahomescu: If you only care about the `!single_bitfield` case here, could you do something like `single_bitfield: false,` on this line?

ahomescu (changes requested):

c2rust-transpile/src/translator/structs_unions.rs
cydparser: Good point. Done.

cydparser:
I do not know why a CI check failed. 

```
[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
```

ahomescu (approved):

PR #1823, opened 2026-05-23
scripts: Add `TRACE` logging option to translator test

Rua:
Found myself wanting this often. The Python library doesn't support `TRACE` as a level, so it's converted to `DEBUG` there.


thedataking:
It is not clear from the PR body what problem this solves. Feel free to re-open with stronger justification.

As a meta-comment, we're receiving many more PRs that we currently have have bandwidth to review (unfortunately) and land so we need to prioritize the ones with high leverage and clear benefits.

PR #1825, opened 2026-05-23
transpile: Make unwrapping `Paren` explicit

### LINKED ISSUE #1824: Macro replacement fails if the macro name is in parentheses
```c
#define FOO 42

void foo(void) {
    int i = (FOO);
}
```

result:
```rust
pub const FOO: ::core::ffi::c_int = 42 as ::core::ffi::c_int;
#[no_mangle]
pub unsafe extern "C" fn foo() {
    let mut i: ::core::ffi::c_int = 42 as ::core::ffi::c_int;
}
```

The macro itself is emitted, but then does not get used. The reason is that indexing `TypedAstContext` with an expression ID automatically skips over parentheses. So, when `convert_expr` is called with the ID of a parenthesis expression, it checks if that expression is a macro invocation (which it's not), then proceeds to convert it as if it were the *inner* expression (the `42` literal). `convert_expr` is never called with the expression ID of the `42` literal expression itself, so it's never replaced with a macro.

In general, automatically bypassing parentheses is a footgun precisely because of things like this, so I wonder if it would be better removed.

Rua:
- Fixes #1824 

The indexing operation for `CExprId` now no longer silently unwraps parentheses. I've replaced all indexing operations with `index_unwrap_parens` so that it's immediately obvious what's going on. Certain uses of it can later be refactored to use plain indexing if it's determined to be safe.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -359,7 +359,7 @@ impl TypedAstContext {
         use SomeId::*;
         match id {
             Stmt(id) => self.index(id).loc,
-            Expr(id) => self.index(id).loc,
+            Expr(id) => self.index_unwrap_parens(id).loc,
             Decl(id) => self.index(id).loc,
             Type(id) => self.index(id).loc,
         }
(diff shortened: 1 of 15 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1502,21 +1518,15 @@ impl Index<CTypeId> for TypedAstContext {
 
 impl Index<CExprId> for TypedAstContext {
     type Output = CExpr;
+
     fn index(&self, index: CExprId) -> &CExpr {
         static BADEXPR: CExpr = Located {
             loc: None,
             kind: CExprKind::BadExpr,
         };
         match self.c_exprs.get(&index) {
```
ahomescu:
```suggestion
        self.c_exprs.get(&index).unwrap_or(&BADEXPR)
```

ahomescu (changes requested):

ahomescu (approved):

PR #1826, opened 2026-05-25
ast-exporter: Do not process macros for the syntactic form of `InitListExpr`

Rua:
In the exporter, the `macroCallSites` variable keeps track of which macro call sites have already been visited, so that only the outermost expression gets processed for that call site. But that depends on Clang's visiting order, and in the case of init lists, the syntactic form appears to be visited first, which then prevents macros from being associated with the semantic form. This PR skips macro processing for the syntactic form (which c2rust ignores anyway), fixing expansion for init lists in the transpiler.


fw-immunant (approved):
LGTM, thanks for the fix!

ahomescu (approved):

PR #1831, opened 2026-05-29
postprocess: support multiple transforms (part 1)

thedataking:
This is the first two commits from PR #1543.


c2rust-postprocess/postprocess/__init__.py
```diff
@@ -93,7 +93,20 @@ def build_arg_parser() -> argparse.ArgumentParser:
         help="Update the Rust in-place",
     )
 
-    # TODO: add option to select what transforms to apply
+    parser.add_argument(
+        "--transform",
+        type=str,
+        required=False,
+        action="append",
+        default=["comments"],
```
Crocodoctopus, line 101: Do we want comments by default? I'm currently testing postprocess integration via a transformer that appends a dummy comment instead of invoking an llm, so there is a usecase for excluding it.

Crocodoctopus (approved):
Works on my integration branch, lgtm.

c2rust-postprocess/postprocess/__init__.py
thedataking: > I'm currently testing postprocess integration via a transformer that appends a dummy comment instead of invoking an llm, so there is a usecase for excluding it.

I think we should pick defaults that we think are helpful to users that are less familiar with this tool than us. As more an more transforms come online, they should be default to on. If there are changes that make testing easier, I'm happy to discuss those separately.

PR #1833, opened 2026-05-31
transpile: Add `CExprKind::ImaginaryLiteral`

### LINKED ISSUE #1802: C2Rust panics when translating C99 complex number using `I` from `<complex.h>`
C2Rust crashes when translating a valid C program that uses C99 complex numbers and the imaginary unit macro `I` from `<complex.h>`.
The translator first reports an invalid exported Clang AST involving a `TagUnaryOperator` node from the macro expansion `I`, then panics with `no entry found for key`.
Even if C2Rust does not fully support complex floating-point types or the `I` macro, it should emit a graceful unsupported-feature diagnostic instead of panicking.
**To Reproduce**
```c
#include <complex.h>

int main(void) {
    double complex z = 1.0 + 2.0 * I;
    return 0;
}
```
Run C2Rust translation:
```
c2rust transpile test.c
```
**Observed Behavior**
```
Transpiling test.c
warning: Missing child 96774104899376 of node AstNode { tag: TagUnaryOperator, children: [Some(96774104899376)], loc: SrcSpan { fileid: 5, begin_line: 4, begin_column: 36, end_line: 4, end_column: 36 }, type_id: Some(96774104846208), rvalue: RValue, macro_expansions: [], macro_expansion_text: Some("I"), extras: [Text("__extension__"), Bool(false)] }
Exported Clang AST was invalid. Check warnings above for unimplemented features.
--> test.c:4:36
 [-Wclang-ast]
thread 'main' panicked at 'no entry found for key', c2rust-transpile/src/c_ast/iterators.rs:332:69
```
**Expected Behavior**
C2Rust should not panic. It should either:
1. translate the C99 complex expression correctly, or
2. report that complex floating-point expressions or the I macro are unsupported, without crashing during AST traversal.
**Environment**
- c2rust version: v0.21.0
- platform: Ubuntu 24.04
- Rust: nightly-2022-08-08
- Clang version: 17.0.6

Rua:
- Fixes #1802

This doesn't actually enable support for complex numbers. It just lays down some groundwork so that the AST stuff is there, ready for a future implementation.

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2588,7 +2588,8 @@ class TranslateASTVisitor final
 #endif // CLANG_VERSION_MAJOR
 
     bool VisitImaginaryLiteral(ImaginaryLiteral *L) {
-        printDiag(Context, DiagnosticsEngine::Warning, "Encountered unsupported imaginary literal", L);
+        std::vector<void *> childIds{L->getSubExpr()};
+        encode_entry(L, TagImaginaryLiteral, childIds);
         return true;
     }
 
--- a/c2rust-ast-exporter/src/ast_tags.hpp
+++ b/c2rust-ast-exporter/src/ast_tags.hpp
@@ -95,6 +95,7 @@ enum ASTEntryTag {
     TagStringLiteral,
     TagCharacterLiteral,
     TagFloatingLiteral,
+    TagImaginaryLiteral,
 };
 
 enum TypeTag {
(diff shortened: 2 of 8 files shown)


ahomescu (changes requested):
Generally looks good, is there a snapshot test you could add this to or make a new one?

Rua:
@ahomescu Not really, the transpiler still doesn't actually support complex types so it would just give a transpiler error. This PR only makes it do that, rather than panicking from an invalid AST as in the linked issue.

ahomescu (approved):

PR #1835, opened 2026-06-02
transpile: Refactor binary operand handling and handle complex numbers

Rua:
Both a refactor and a functional improvement. Complex numbers aren't supported yet, so I haven't included any tests for that, but this is needed to support them later.

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -454,6 +454,12 @@ impl ConversionContext {
             self.add_type(new_id, not_located(rust_type_kind));
             self.processed_nodes
                 .insert(new_id, self::node_types::OTHER_TYPE);
+
+            let complex_kind = CTypeKind::Complex(CTypeId(new_id));
+            let new_id = self.id_mapper.fresh_id();
+            self.add_type(new_id, not_located(complex_kind));
+            self.processed_nodes
+                .insert(new_id, self::node_types::OTHER_TYPE);
         }
 
         // Continue popping Clang nodes off of the stack of nodes we have promised to visit
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1361,7 +1361,7 @@ impl TypedAstContext {
                             None
                         }
                     }
-                    CExprKind::Binary(_ty, op, lhs, rhs, _, _) => {
+                    CExprKind::Binary(result_type_id, op, lhs, rhs, _, _) => {
                         let rhs_type_id =
                             self.ast_context.c_exprs[&rhs].kind.get_qual_type().unwrap();
                         let lhs_kind = &self.ast_context.c_exprs[&lhs].kind;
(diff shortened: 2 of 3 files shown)


ahomescu (changes requested):
Could you add a test or two for this (unit and/or snapshot)?

Rua:
Not really sure what to test for, that isn't already tested?

ahomescu:
Okay we can land this as is since it's a pretty small change, but it would be nice in the future if changes were split between 1) refactorings (no tests needed), and 2) actual functional changes, that would need a test or two to handle those specific corner cases.

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1361,32 +1361,48 @@ impl TypedAstContext {
...
+                                        lhs_type_id
+                                    } else {
+                                        rhs_type_id
+                                    };
+
+                                // For complex arithmetic, complex and real values can be mixed.
+                                // See if a `Complex` version of the new result type exists.
+                                if matches!(result_type_kind, CTypeKind::Complex(..)) {
+                                    let new_result_type_kind =
+                                        CTypeKind::Complex(result_type_id.ctype);
```
ahomescu, line 1564: What if `result_type_id` is already a `Complex`? Would something like `double f(double _Complex a, double _Complex b) { ... a + b ... }` panic here? That could be a new test.

c2rust-transpile/src/c_ast/conversion.rs
ahomescu, line 452: All the pullback types are integers, do any of them make sense for complex?

ahomescu (changes requested):

c2rust-transpile/src/c_ast/conversion.rs
Rua: They'd probably not be used much, but since C supports them, they probably ought to be here too.

c2rust-transpile/src/c_ast/mod.rs
Rua: If it's already a `Complex` then it would either stay the same, or be replaced with the complex version of one of the `PULLBACK_KINDS`, same as happens with other numeric types right now. The code here only applies to binary operators, so it won't modify a function call.

ahomescu: My point was: is it incorrect to create a type like `CTypeKind::Complex(Complex(...))`?

Rua: In C, yes.

ahomescu: Can you check if the following C functions cause transpiler panics?
```C
double _Complex f(double _Complex a, double _Complex b) {
    return a * b;
}

double _Complex g(double a, double _Complex b) {
    return a * b;
}

typedef double real;

double _Complex h(double _Complex a, real b) {
    return a * b;
}
```

Rua: Hmm, it does, and I see the problem now. I managed to fix the first two cases, but the third one still panics because of the `typedef`. No `real _Complex` type exists. I need to think of a way around that...

Rua: Ok, I've completely rewritten that part of the code. It should be good now?

PR #1838, opened 2026-06-02
Update postprocess for integration into GaloisInc/Tractor-Crisp

Crocodoctopus:
Adds a few assert relaxations to keep postprocess, as well as a change to update_rust_definition invocation that should probably be its own PR.


Crocodoctopus:
Small PR to get postprocess working with gpt on json-c. Mostly it just relaxed some asserts, which probably necessarily be fatal. Perhaps these should be controlled by flags?

PR #1840, opened 2026-06-06
Amend postprocess for json-c lifting support.

Crocodoctopus:
This PR does two things, neither of which I'm full convinced of yet:

I update the prompt because json-c has very poor comment style hygiene, and it was tripping up the llm. It may be better to normalize json-c manually, rather than trying to play prompt-fu every time a codebase has weird comments, but the prompt may also not be very good either.

I also drop the assert on `c_comments` / `rust_comments`; the llm was doing weird things with whitespaces that were technically code changes, but probably permissible?

### DIFF
--- a/c2rust-postprocess/postprocess/transforms/comments.py
+++ b/c2rust-postprocess/postprocess/transforms/comments.py
@@ -136,7 +136,13 @@ class CommentsTransform(AbstractTransform):
         rust_comments = get_rust_comments(rust_fn)
         logging.debug(f"{rust_comments=}")
 
-        assert c_comments == rust_comments
+        if c_comments != rust_comments:
+            print(
+                "Comment mismatch for "
+                f"{identifier}:\n"
+                f"C comments:\n{c_comments}\n\n"
+                f"Rust comments:\n{rust_comments}\n"
+            )
 
         print(get_highlighted_rust(rust_fn))
 
--- a/c2rust-postprocess/postprocess/transforms/comments.py
+++ b/c2rust-postprocess/postprocess/transforms/comments.py
@@ -78,10 +78,12 @@ class CommentsTransform(AbstractTransform):
 
         # TODO: make this function take a model and get prompt from model
         prompt_text = """
-        Transfer the comments from the following C function to the corresponding Rust function.
-        Do not add any comments that are not present in the C function.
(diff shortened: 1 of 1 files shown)


thedataking:
I tested like this:
```shell
cd integration/tests/json-c
GEMINI_API_KEY="$GEMINI_API_KEY" c2rust-postprocess --update-rust --log-level DEBUG --exclude-file postprocess-exclude.yml repo/lib.rs --ident-filter json_object_iter_begin
```
on this C function 
```C
/**
 * ****************************************************************************
 */
struct json_object_iterator
json_object_iter_begin(struct json_object* obj)
{
    struct json_object_iterator iter;
    struct lh_table* pTable;

    /// @note json_object_get_object will return NULL if passed NULL
    ///       or a non-json_type_object instance
    pTable = json_object_get_object(obj);
    JASSERT(NULL != pTable);

    /// @note For a pair-less Object, head is NULL, which matches our
    ///       definition of the "end" iterator
    iter.opaque_ = pTable->head;
    return iter;
}
```

and got

```Rust
#[no_mangle]
pub unsafe extern "C" fn json_object_iter_begin(mut obj: *mut json_object) -> crate::src::json_object_iterator::json_object_iterator {
    let mut iter: crate::src::json_object_iterator::json_object_iterator = crate::src::json_object_iterator::json_object_iterator {
    opaque_:  ::core::ptr::null::<::core::ffi::c_void>(),
};
    let mut pTable: *mut lh_table = ::core::ptr::null_mut::<lh_table>();
    // @note json_object_get_object will return NULL if passed NULL
    //       or a non-json_type_object instance
    pTable =  json_object_get_object(obj as *const json_object) as
    *mut lh_table as *mut lh_table;
    // @note For a pair-less Object, head is NULL, which matches our
    //       definition of the "end" iterator
    iter.opaque_ = (*pTable).head as *const ::core::ffi::c_void;
    return iter;
}
```

Without this PR. So I think the current prompt and the current check works fine for some models such as the default gemini-3.5-flash but perhaps not the one you used when developing this PR.

Update: `gpt-5.5-2026-04-23` also works correctly in my testing.

thedataking:
Closing since code seems to work with latest models from OpenAI and Google. I'm sure there are issues with older models but we don't have a strong reason to support them. Feel free to reopen if issue resurfaces on recent model.

PR #1841, opened 2026-06-06
merge_rust: Sync with upstream implementation

Crocodoctopus:
Syncs c2rust's `merge_rust` with [Crisp's](https://github.com/GaloisInc/Tractor-Crisp/tree/main/tools/merge_rust).

I also add --update-only by default when invoked via `update_rust_definition`, as that seems to be what the old behavior was doing.


thedataking (approved):
Looks good overall.

I think it would have been cleaner to put `__init__.py` in its own commit after updating `merge_rust`. That would avoid mixing changes made upstream and locally. 

Also, I think the commits updating the upstream tools should include the SHA of Tractor-Crisp so we don't have to guess what version the upstream was in when this was copied.

PR #1842, opened 2026-06-08
ast-exporter, transpile: Unify handling of attributes

Rua:
All attributes now pass through `encodeAttribute` on the C++ side, and `parse_attribute[s]` on the Rust side. The last commit fixes the issue mentioned [here](https://github.com/immunant/c2rust/pull/1815#issuecomment-4583411264).

### DIFF
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -1121,10 +1121,15 @@ class TranslateASTVisitor final
     bool VisitAttributedStmt(AttributedStmt *S) {
         std::vector<void*> childIds { S->getSubStmt() };
         encode_entry(S, TagAttributedStmt, childIds,
-                     [S](CborEncoder *array){
+                     [S](CborEncoder *local){
+                         CborEncoder array;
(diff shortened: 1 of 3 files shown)


c2rust-ast-exporter/src/AstExporter.cpp
```diff
@@ -2557,6 +2531,81 @@ class TranslateASTVisitor final
         return true;
     }
 
+    void encodeAttribute(CborEncoder *array, const Attr *attr) {
+        CborEncoder attr_info;
+
+        switch (attr->getKind()) {
+            case attr::Alias:
+                cbor_encoder_create_array(array, &attr_info, 2);
```
ahomescu: The duplicated `create_array`-`encode_text_stringz`-`close_container` calls could be split into a helper function that takes a callback for the extra values, or maybe something with even stronger typing.

ahomescu (changes requested):

c2rust-ast-exporter/src/AstExporter.cpp
Rua: Done.

ahomescu (approved):

PR #1843, opened 2026-06-08
transpile: Override pointer difference expression type to `PtrDiff`

### LINKED ISSUE #1837: transpile: Pointer difference results in `c_long` rather than `ptrdiff_t`
Since C99, the result of a pointer difference operation is defined to be `ptrdiff_t`. c2rust overrides the type of certain expressions with `size_t`, but not in this case. Consequently, when the result is cast to something else, it gets cast first to `c_long`:

```c
#include <stddef.h>

void ptr_diff(void) {
    int *p1 = NULL;
    int *p2 = NULL;
    int diff = p1 - p2;
}
```
Result:
```rust
pub const NULL: *mut ::core::ffi::c_void = ::core::ptr::null_mut::<
    ::core::ffi::c_void,
>();
#[no_mangle]
pub unsafe extern "C" fn ptr_diff() {
    let mut p1: *mut ::core::ffi::c_int = ::core::ptr::null_mut::<::core::ffi::c_int>();
    let mut p2: *mut ::core::ffi::c_int = ::core::ptr::null_mut::<::core::ffi::c_int>();
    let mut diff: ::core::ffi::c_int = p1.offset_from(p2) as ::core::ffi::c_long
        as ::core::ffi::c_int;
}
```

Rua:
- Depends on #1850
- Fixes #1837

### DIFF
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -395,6 +395,26 @@ impl<'c> Translation<'c> {
         WithStmts::new_val(res).set_unsafe()
     }
 
+    /// Creates a pointer difference expression. Returns an expression of type `isize`.
+    pub(crate) fn make_pointer_difference(
+        &self,
+        lhs_rs: Box<Expr>,
+        rhs_rs: Box<Expr>,
+        pointee_type_id: CTypeId,
+    ) -> WithStmts<Box<Expr>> {
+        let mut expr_rs = mk().method_call_expr(lhs_rs, "offset_from", vec![rhs_rs]);
+
+        // If the pointee is a variable array type, the actual pointee type used by `offset_from`
+        // will be its element type rather than the whole array. So we need to divide by the
+        // variable holding the length of the array.
+        if let Some(sz) = self.compute_size_of_expr(pointee_type_id) {
+            let div_rs = cast_int(sz, "isize", false);
+            expr_rs = mk().binary_expr(BinOp::Div(Default::default()), expr_rs, div_rs);
+        }
+
+        WithStmts::new_val(expr_rs).set_unsafe()
+    }
+
     /// Construct an expression for a NULL at any type, including forward declarations,
     /// function pointers, and normal pointers.
     pub fn null_ptr(&self, type_id: CTypeId) -> TranslationResult<Box<Expr>> {
--- a/c2rust-transpile/tests/snapshots/exprs.c
+++ b/c2rust-transpile/tests/snapshots/exprs.c
@@ -1,3 +1,5 @@
+#include <stddef.h>
+
 int puts(const char *str);
 
 static int side_effect(){
@@ -96,4 +98,13 @@ void statement_expr() {
     });
 
     puts("should be unreachable!");
+}
+
+void pointer_arithmetic(void) {
+    int i1[] = { 0, 1 };
+    int i2[] = { 10, 11 };
+    int *p1 = i1;
+    int *p2 = i2;
+    ptrdiff_t diff = p1 - p2;
+    int diff_int = p1 - p2;
 }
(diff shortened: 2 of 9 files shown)


c2rust-transpile/tests/snapshots/snapshots__transpile@exprs.c.2021.clang15.snap
```diff
@@ -82,3 +82,11 @@ pub unsafe extern "C" fn statement_expr() {
     };
     puts(b"should be unreachable!\0".as_ptr() as *const ::core::ffi::c_char);
 }
+#[no_mangle]
+pub unsafe extern "C" fn pointer_arithmetic() {
+    let mut i1: [::core::ffi::c_int; 2] = [0 as ::core::ffi::c_int, 1 as ::core::ffi::c_int];
+    let mut i2: [::core::ffi::c_int; 2] = [10 as ::core::ffi::c_int, 11 as ::core::ffi::c_int];
+    let mut p1: *mut ::core::ffi::c_int = &raw mut i1 as *mut ::core::ffi::c_int;
+    let mut p2: *mut ::core::ffi::c_int = &raw mut i2 as *mut ::core::ffi::c_int;
+    let mut diff: ::core::ffi::c_int = p1.offset_from(p2) as ::core::ffi::c_int;
```
ahomescu: Should there be a `ptrdiff_t` here somewhere?

ahomescu (changes requested):

c2rust-transpile/tests/snapshots/snapshots__transpile@exprs.c.2021.clang15.snap
Rua: Added that too now.

Rua:
This now depends on #1850 so that the definition of `ptrdiff_t` in the snapshot test doesn't differ between platforms.

fw-immunant (approved):
LGTM!

c2rust-transpile/src/translator/pointers.rs
```diff
@@ -385,6 +385,23 @@ impl<'c> Translation<'c> {
...
+    /// Creates a pointer difference expression. Returns an expression of type `isize`.
+    pub fn make_pointer_difference(
+        &self,
+        lhs_rs: Box<Expr>,
+        rhs_rs: Box<Expr>,
+        pointee_type_id: CTypeId,
+    ) -> WithStmts<Box<Expr>> {
+        let mut expr_rs = mk().method_call_expr(lhs_rs, "offset_from", vec![rhs_rs]);
+
+        if let Some(sz) = self.compute_size_of_expr(pointee_type_id) {
```
ahomescu, line 410: Is this needed? If `lhs_rs` is already of type `*mut $pointee_type` then `offset_from` returns the size in units of `$pointee_type`. There is `byte_offset_from` for the byte distance. Am I missing something here?

Rua: I'm not sure. I left it unchanged because it was already in the original code.

ahomescu: Ah I think I figured it out: variable-length arrays like `T a[n];` get transpiled to `*mut T` but the actual size of the pointee is `sizeof(T) * n`. This is an extremely niche corner case, but I guess it could happen in practice.

A comment could be useful here, for posterity.

ahomescu (changes requested):

c2rust-transpile/tests/snapshots/exprs.c
```diff
@@ -66,3 +68,12 @@ void statement_expr() {
...
     puts("should be unreachable!");
 }
+
+void pointer_arithmetic(void) {
+    int i1[] = { 0, 1 };
+    int i2[] = { 10, 11 };
+    int *p1 = i1;
+    int *p2 = i2;
+    ptrdiff_t diff = p1 - p2;
+    int diff_int = p1 - p2;
```
ahomescu, line 109: An extra test with some VLAs could be useful here.

Rua: I've added a test now.

ahomescu (approved):

PR #1845, opened 2026-06-09
tests: integration: json-c: Fix configure step

thedataking:
Don't treat calloc-transposed-args and implicit-function-declaration lints as an errors. Issue encountered on Ubuntu 26.04/GCC 15.


Crocodoctopus (approved):
Looks fine to me.

PR #1846, opened 2026-06-09
postprocess: Preprocess the C source code passed to transforms

thedataking:
The `CommentsTransform` can fail if presented with a C and Rust function pair where some comments are inside directives that are compiled out during preprocessing and thus should not be transferred to the Rust.

We can simply run the C compiler on the C snippet and it will pick up compile commands (if any) and give us the preprocessed version back.


c2rust-transpile/src/translator/mod.rs
```diff
@@ -679,12 +707,117 @@ pub fn translate_failure(tcfg: &TranspilerConfig, msg: &str) {
...
+/// output line. Other `#` lines (e.g. `#define` passed through in
+/// directives-only mode) are not markers and yield `None`.
+fn parse_line_marker(line: &str) -> Option<(u64, &str)> {
+    let rest = line.strip_prefix('#')?;
+    let rest = rest.trim_start_matches([' ', '\t']);
+    let digits_len = rest.bytes().take_while(|b| b.is_ascii_digit()).count();
+    let line_no: u64 = rest.get(..digits_len)?.parse().ok()?;
+    let rest = rest[digits_len..].trim_start_matches([' ', '\t']);
+    let path = rest.strip_prefix('"')?;
+    let path = &path[..path.find('"')?];
```
fw-immunant, line 824: Nit: what about filenames containing quotes--are they escaped somehow, or should we prefer `rfind` over `find`? Will we ever need to unescape paths somehow before comparison to what the filesystem holds?

This code probably doesn't have to be absolutely bulletproof, but a comment here would be a good start.

c2rust-transpile/tests/snapshots/snapshots__c_decls@nh.c.snap
```diff
@@ -3,22 +3,78 @@ source: c2rust-transpile/tests/snapshots.rs
...
+      "definition": "int d;",
+      "preprocessed_definition": null
+    },
+    "e": {
+      "definition": "int e;",
+      "preprocessed_definition": null
+    },
+    "f": {
+      "definition": "void f(void){}",
+      "preprocessed_definition": "void f(void){}void g(void){}/*comment for h*/void h(void){}int d;int e;;;;;;;;int another;"
```
fw-immunant: `preprocessed_definition` here contains other functions that follow on the same line. Does this pattern trip up the postprocessor?

c2rust-transpile/tests/snapshots.rs
```diff
@@ -596,3 +596,9 @@ fn test_c_decls_nh() {
     let c_path = Path::new("tests/c_decls_snapshots/nh.c");
     transpile_with_c_decl_map_snapshot(c_path);
 }
+
+#[test]
+fn test_c_decls_directives() {
+    let c_path = Path::new("tests/c_decls_snapshots/directives.c");
+    transpile_with_c_decl_map_snapshot(c_path);
+}
```
fw-immunant, line 652: Should be sorted before `test_c_decls_nh`; each section in this file is alphabetically sorted.

fw-immunant (approved):
Generally looks good; inline comments about a couple edge-cases and formatting in tests.

One larger question I have is how this interacts with configurability--is this an improvement for only the case where we fully remove all preprocessor logic (as opposed to lowering it to `#[cfg]`s with Hayroll?

c2rust-transpile/src/translator/mod.rs
thedataking: Any quotes in filenames should be escaped. Same for newlines and backslashes, so this *should* work.

Totally agree this deserves a comment!

c2rust-transpile/tests/snapshots/snapshots__c_decls@nh.c.snap
thedataking: Good point. Let's skip and warn on such pathological cases.

thedataking:
> One larger question I have is how this interacts with configurability--is this an improvement for only the case where we fully remove all preprocessor logic (as opposed to lowering it to #[cfg]s with Hayroll?

Yes, I think it is the right for comment transfer today since it lets us validate transferred comments in more cases.

I haven't thought much about how this would work when lowering `#[cfg]`s with Hayroll. Couldn't the postprocessor just run before Hayroll? It would transfer comments for the active config, and Hayroll would handle comments in the branches it reintroduces.

PR #1850, opened 2026-06-11
transpile: Bypass platform-specific internal typedefs for portable types

### LINKED ISSUE #1307: Skip platform-dependent intermediate typedefs
See https://github.com/immunant/c2rust/pull/1266#discussion_r2234828316.

As a follow-up to #1266, it'd be better to also emit platform-independent intermediate typedefs.  #1266 makes the final typedefs, e.x. `uint32_t` point to the same underlying `u32`, but they go through different intermediate typedefs on different platforms.  This likely needs some hardcoding for things like `__uint32_t` on Linux or `__darwin_size_t` on macOS.

Rua:
- Fixes #1307

This eliminates a bunch of differences between platforms, so that they produce the same output. It does seem like the order of the typedefs can still differ, so they sometimes still need separate snapshot tests. Maybe a solution can be found for that in the future too.


fw-immunant (approved):
Thanks for the PR! It looks good, but it would be nice to restore some documentation for the top-level `fn bypass_typedefs`. Do you want to add docs or should I do so after merging?

Rua:
@fw-immunant I've added docs for it now.

PR #1851, opened 2026-06-11
postprocess: Add `--on-error` flag for non-fatal transform failures

Crocodoctopus:
Adds configurable transform error handling via `--on-error`.

Comment transform failures now raise `TransformError` instead of relying on a hard assert. The CLI can either abort on the first failure, keep going while returning exit 1, or warn and return exit 0.

As part of this, `cache.update(...)` now happens only after the comment parity check succeeds.

### DIFF
--- a/c2rust-postprocess/postprocess/__init__.py
+++ b/c2rust-postprocess/postprocess/__init__.py
@@ -17,6 +17,7 @@ from postprocess.models import api_key_from_env, get_model_by_id
 from postprocess.models.gpt import GPTModel
 from postprocess.models.mock import MockGenerativeModel
 from postprocess.transforms import get_transform_by_id
+from postprocess.transforms.base import TransformError
 from postprocess.transforms.comments import (
     SYSTEM_INSTRUCTION,
     AbstractGenerativeModel,
@@ -98,6 +99,16 @@ def build_arg_parser() -> argparse.ArgumentParser:
     )
 
     parser.add_argument(
+        "--on-error",
+        type=str,
+        required=False,
+        default="keep-going",
+        choices=["abort", "keep-going", "warn"],
+        help="Handle per-function transform failures: abort at first failure,"
+        " keep going with exit 1, or warn and exit 0 (default: keep-going)",
+    )
+
+    parser.add_argument(
         "--transform",
         type=str,
         required=False,
@@ -174,19 +185,39 @@ def main(argv: Sequence[str] | None = None):
             if transform_id.strip()
         )
         transforms = [
-            get_transform_by_id(transform_id, cache=cache, model=model)
+            get_transform_by_id(
+                transform_id,
+                cache=cache,
+                model=model,
+            )
             for transform_id in transform_ids
         ]
 
+        failures = 0
+        failure_log_level = (
+            logging.WARNING if args.on_error == "warn" else logging.ERROR
+        )
         for transform in transforms:
-            transform.apply_dir(
+            failures += transform.apply_dir(
                 root_rust_source_file=args.root_rust_source_file,
                 exclude_list=IdentifierExcludeList(src_path=args.exclude_file),
                 ident_filter=args.ident_filter,
                 update_rust=args.update_rust,
+                keep_going=args.on_error != "abort",
+                failure_log_level=failure_log_level,
             )
 
+        if failures:
+            logging.log(
+                failure_log_level, f"Failed to transform {failures} function(s)"
+            )
+            if args.on_error != "warn":
+                return 1
+
         return 0
+    except TransformError as error:
+        logging.exception(f"Aborting at first transform failure: {error}")
+        return 1
     except KeyboardInterrupt:
         logging.warning("Interrupted by user, terminating...")
         return 130  # 128 + SIGINT(2)
--- a/c2rust-postprocess/postprocess/transforms/comments.py
+++ b/c2rust-postprocess/postprocess/transforms/comments.py
@@ -9,7 +9,7 @@ from postprocess.definitions import (
     update_rust_definition,
 )
 from postprocess.models import AbstractGenerativeModel, api_key_from_env
-from postprocess.transforms.base import AbstractTransform
+from postprocess.transforms.base import AbstractTransform, TransformError
 from postprocess.utils import get_highlighted_rust, remove_backticks
 
 # TODO: get from model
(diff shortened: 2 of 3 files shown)


thedataking (changes requested):
Right now, the postprocessor only transfers comments but we have PRs to process assertions, and other useful cleanup. I don't think we should have a comment-specific flag. 

Please consider a more generic design where we accept a flag (or flag pair `--fail-fast`/`--no-fail-fast` with the former being the default) that any Transform that derives from `AbstractTransform` can use to decide whether to keep going or not.

Crocodoctopus:
The flag ended up being --continue-on-error, which is now passed to AbstractTransform for future transforms.

thedataking:
> The flag ended up being --continue-on-error, which is now passed to AbstractTransform for future transforms.

`cargo test` supports `--keep-going` which is what you initially proposed.
I mentioned `--fail-fast`, `--no-fail-fast` because `cargo nextest` exposes that. 
Why did you chose `--continue-on-error`? Considering all the options, I think your `--keep-going` suggestion is best since (I assume) more people use and know about the `cargo test` CLI than the `cargo nextest` one.

c2rust-postprocess/postprocess/__init__.py
```diff
@@ -95,6 +95,17 @@ def build_arg_parser() -> argparse.ArgumentParser:
         help="Update the Rust in-place",
     )
 
+    parser.add_argument(
+        "--continue-on-error",
+        required=False,
+        action="store_true",
+        default=False,
+        help=(
+            "Skip non-fatal transform errors, continue processing, and still "
```
thedataking: make concise : `Do not abort postprocessing as soon as there is an error`.

c2rust-postprocess/postprocess/transforms/comments.py
```diff
@@ -136,7 +132,26 @@ def apply_ident(
...
+        if c_comments != rust_comments:
+            msg = "\n".join(
+                [
+                    f"Comment mismatch for {identifier}",
+                    f"Before (C comments): {c_comments}",
+                    f"After (Rust comments): {rust_comments}",
+                ]
+            )
+            if self.continue_on_error:
+                logging.warning(msg)
```
thedataking: before this commit, this was a hard error (or assert, but same difference), everything about your naming indicates this is still an error but we may choose to keep going past it. Shouldn't we log this as an error?

thedataking:
from offline discussion: `--keep-going` or `--no-fail-fast` isn't exactly what we need here since we also want to option to downgrade errors to warnings so postprocessing can run in best effort mode. Simplest solution that covers CI, development, and CRISP use requirements:

Tri-state `on-error {abort, keep-going, warn}` flag. With the following semantics:

  - `abort`: stop at first failure, nonzero exit (development/debugging)
  - `keep-going`: process everything, report all failures, nonzero exit (CI)
  - `warn`: process everything, log failures as warnings, exit 0 (embedding in a larger tool)

`keep-going` is the safe default for users of c2rust.

We can use exceptions to simplify error handling:

```
  class TransformError(Exception): ...

  if c_comments != rust_comments:
      raise TransformError(f"comment mismatch for {identifier}: {c_comments=} {rust_comments=}")
```

 Then the policy logic goes here:

  1. The per-function loop in apply_file (base.py:108) wraps the apply_ident call in try/except TransformError; on abort it re-raises, otherwise it records the failure and
  continues. 
  2. main() looks at the collected failure count at the end: nonzero exit unless mode is warn (where it logs a summary like "postprocessed N functions, skipped M due to errors" and returns 0).

For PR #1849 specifically (not blocking for this PR) :
One complication is that we might want to always keep going if `TrimTransform` fails but the `on-error` flag is flexible enough to allow more states if we absolutely need them.

c2rust-postprocess/postprocess/__init__.py
Crocodoctopus: New prompt is now concise.

c2rust-postprocess/postprocess/__init__.py
```diff
@@ -154,19 +165,39 @@ def main(argv: Sequence[str] | None = None):
...
+        if failures:
+            logging.log(
+                failure_log_level, f"Failed to transform {failures} function(s)"
             )
+            if args.on_error != "warn":
+                return 1
 
         return 0
+    except TransformError as error:
+        logging.error(f"Aborting at first transform failure: {error}")
```
thedataking: When the `assert`, we'd see where the problem occured.

I think you can use use logging.exception(...) (or logging.error(..., exc_info=True)) so the traceback to the raise site is printed.

c2rust-postprocess/postprocess/transforms/__init__.py
```diff
@@ -5,13 +5,20 @@
 
 
 def get_transform_by_id(
-    id: str, cache: AbstractCache, model: AbstractGenerativeModel
+    id: str,
+    cache: AbstractCache,
+    model: AbstractGenerativeModel,
+    continue_on_error: bool = False,
```
thedataking: Where is this used? Can we get rid of it?

thedataking (changes requested):
PR body is stale. Cache.update was moved to after the comment-verbatim check. Worth doing but you should document the change when you update the PR body.

`continue_on_error` appears to be a vestige from previous iteration; remove it?

c2rust-postprocess/postprocess/transforms/__init__.py
Crocodoctopus: It is removed.

c2rust-postprocess/postprocess/__init__.py
Crocodoctopus: 199 is now `logging.exception`

Crocodoctopus:
I was unaware the top-most comment should be edited to reflect the current state of the PR, I'll do that from now on.

Instances of the old `continue_on_error` were all removed.

This should be g2g?

thedataking (approved):
LGTM with one nitpick: 

Please undo formatting changes in transforms/__init__.py and comments.py. 
- `transforms/__init__.py` changes can just be removed from this commit entirely 
- comments.py: just need to revert `CommentsTransform.__init__` 

(`logging.exception` works well here, nice find. note that it prints the error at the end of the traceback as well as in the f-string  you pass as an argument - should be harmless.)

PR #1852, opened 2026-06-12
postprocess: Tighten model input handling

Crocodoctopus:
Adds guards for model selection, as well as cleans up an env read that happened deeper than it needed to; env reading now happens in __init__ and is passed down.

I think it wouldn't be a bad idea to combine `GEMINI_API_KEY` with `OPENAI_API_KEY` here too, but it would depend on what else uses this API already.

### DIFF
--- a/c2rust-postprocess/postprocess/__init__.py
+++ b/c2rust-postprocess/postprocess/__init__.py
@@ -120,11 +120,12 @@ def get_model(model_id: str) -> AbstractGenerativeModel:
             f"API key for model {model_id} not found in env; "
             "using cached responses only."
         )
-        return MockGenerativeModel()
+        return MockGenerativeModel(id=model_id)
 
     # TODO: remove google specific API bits
     return get_model_by_id(
         model_id,
+        api_key=api_key,
         generation_config={
             "system_instruction": types.Content(
                 role="system", parts=[types.Part.from_text(text=SYSTEM_INSTRUCTION)]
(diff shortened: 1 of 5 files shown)


c2rust-postprocess/postprocess/__init__.py
```diff
@@ -170,3 +171,6 @@ def main(argv: Sequence[str] | None = None):
     except KeyboardInterrupt:
         logging.warning("Interrupted by user, terminating...")
         return 130  # 128 + SIGINT(2)
+    except ValueError as error:
```
thedataking, line 174: this seems unrelated to key handling; please remove

c2rust-postprocess/postprocess/models/__init__.py
```diff
@@ -11,14 +11,19 @@ def api_key_from_env(model_id: str) -> str | None:
         return os.getenv("GEMINI_API_KEY")
     elif model_id.startswith("gpt"):
         return os.getenv("OPENAI_API_KEY")
-    return None
 
+    raise ValueError(f"Unsupported model identifier: {model_id}")
```
thedataking, line 15: why do you need this?

thedataking:
> I think it wouldn't be a bad idea to combine GEMINI_API_KEY with OPENAI_API_KEY here too, but it would depend on what else uses this API already.

I don't know what that would buy us.

> Adds guards for model selection, as well as cleans up an env read that happened deeper than it needed to; env reading now happens in init and is passed down.

What specifically was wrong with the current approach?

thedataking:
I don't think we want this change. Please try https://github.com/immunant/c2rust/pull/1853 and see if that unblocks  https://github.com/GaloisInc/Tractor-Crisp/pull/131.

PR #1853, opened 2026-06-12
postprocess: Support CRISP_API_MODEL/_KEY/_BASE env var overrides

thedataking:
When CRISP_API_MODEL is set, route through the OpenAI-compatible client using the Tractor-Crisp env vars, overriding CLI model selection.

### DIFF
--- a/c2rust-postprocess/postprocess/__init__.py
+++ b/c2rust-postprocess/postprocess/__init__.py
@@ -4,6 +4,7 @@ c2rust-postprocess: Transfer comments from C functions to Rust functions using L
 
 import argparse
 import logging
+import os
 from argparse import BooleanOptionalAction
 from collections.abc import Sequence
 from pathlib import Path
@@ -13,6 +14,7 @@ from google.genai import types
 from postprocess.cache import DirectoryCache, FrozenCache
 from postprocess.exclude_list import IdentifierExcludeList
 from postprocess.models import api_key_from_env, get_model_by_id
+from postprocess.models.gpt import GPTModel
 from postprocess.models.mock import MockGenerativeModel
 from postprocess.transforms import get_transform_by_id
 from postprocess.transforms.comments import (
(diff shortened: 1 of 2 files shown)


c2rust-postprocess/postprocess/__init__.py
```diff
@@ -114,6 +116,19 @@ def build_arg_parser() -> argparse.ArgumentParser:
 
 
 def get_model(model_id: str) -> AbstractGenerativeModel:
+    # CRISP_API_MODEL/_KEY/_BASE (see github.com/GaloisInc/Tractor-Crisp)
+    # override CLI model selection; CRISP endpoints are OpenAI-compatible.
+    if crisp_model_id := os.getenv("CRISP_API_MODEL"):
```
Crocodoctopus, line 121: I think we want to fail if one `CRISP_*` env is set, but the rest aren't? Otherwise, with current fallback behavior, its not clear whether postprocess does or does not support mixing some of its inputs with some of Crisp's inputs (ex. with just `CRISP_API_KEY` set, postprocess ignores it).

I thought about policy regarding allowing mixing them, and it feels very foot-gun-y. For example, if we have CRISP set for gpt, then run `GEMINI_KEY=.. postprocess --llm-model gemini-something` in that environment, the expected behavior might be postprocess runs with gemini, but implicitly `CRISP_API_BASE` gets forwarded under our nose, so I think we should be very clear we do not support this behavior.

We may also want this behavior for when postprocess inputs are passed when Crisp envs are detected?

Crocodoctopus:
I can take over this PR if you'd like.

Edit: forgot to submit review, this comment was probably confusing.

c2rust-postprocess/postprocess/__init__.py
thedataking: > I think we want to fail if one CRISP_* env is set, but the rest aren't? Otherwise, with current fallback behavior, its not clear whether postprocess does or does not support mixing some of its inputs with some of Crisp's inputs (ex. with just CRISP_API_KEY set, postprocess ignores it).

It already fails unless `OPENAI_API_KEY` happens to be set to a valid key.
I've added guards so that it either uses `CRISP_API_KEY` or terminates with an error.

> I thought about policy regarding allowing mixing them, and it feels very foot-gun-y. For example, if we have CRISP set for gpt, then run GEMINI_KEY=.. postprocess --llm-model gemini-something in that environment, the expected behavior might be postprocess runs with gemini, but implicitly CRISP_API_BASE gets forwarded under our nose, so I think we should be very clear we do not support this behavior.

That wouldn't work. If `CRISP_API_MODEL` is set, then `postprocess` would use the openAI API which would refuse to use any non-openAI model. Most users won't set the CRISP variables and those who do needs to use CRISP properly.

Crocodoctopus (approved):
lg2m

PR #1857, opened 2026-06-16
postprocess: Add cronjob-based testing of c2rust-postprocess

Crocodoctopus:
Test PR c2rust-postprocess testing in CI.

The basic idea is to run a daily cronjob that runs c2rust-postprocess on json-c with a real API key and `--on-error keep-going`, generating new cache entries. PR CI runners will run their postprocess tests with cache-only mode and `--on-error warn`. This has the side effect that breaking changes to c2rust-postprocess that still work under `--on-error warn` will only become apparent during the daily cronjob. 

In addition, the daily cronjob will prune its inner cache entries after 90 days. Note that the consumer job is restore-only, and cannot update these values; ie only the cronjob may decide when a cache entry has been recently-used.

### DIFF
--- a/.github/workflows/ci-c2rust-postprocess.yml
+++ b/.github/workflows/ci-c2rust-postprocess.yml
@@ -1,0 +1,139 @@
+name: ci-c2rust-postprocess
+
+on:
+  schedule:
+    - cron: "0 4 * * *"
+
+jobs:
+  run-testsuite:
+    strategy:
+      matrix:
+        include:
+          - runner: ubuntu-latest
+            os: Linux
+            arch: x86_64
+            clang-version: 18
+          - runner: ubuntu-22.04
+            os: Linux
+            arch: x86_64
+            clang-version: 15
+      fail-fast: false
+    name: "postprocess-test-and-cache (${{ matrix.runner }}: ${{ matrix.os }} ${{ matrix.arch}}, Clang ${{ matrix.clang-version }})"
+    # The type of runner that the job will run on
+    runs-on: ${{ matrix.runner }}
+
+    env:
+      # TODO: Remove this when we know the cronjob works
+      C2RUST_POSTPROCESS_ARGS: --ident-filter '^(json_object_set_userdata|json_object_to_file)$'
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      # Working dir is /home/runner/work/c2rust/c2rust
+      - name: Checkout c2rust
+        uses: actions/checkout@v4
+        with:
+          repository: immunant/c2rust
+          # `update = none` is set, which makes this not work, so we need to update them manually.
+          # See `.gitmodules` for more info.
+          submodules: false
+
+      - name: Checkout submodules
+        run: |
+          git submodule update --init --checkout tests/integration/tests/json-c/repo
+
+      - uses: astral-sh/setup-uv@v6
+
+      - run: uv python install
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-workspace-crates: true
+
+      - name: Install Rust toolchains
+        run: |
+          rustup toolchain install nightly-2022-11-03 \
+            --profile minimal --component rustfmt,rustc-dev
+          rustup toolchain install nightly-2023-04-15 \
+            --profile minimal --component rustfmt
+
+      - name: Provision Debian Packages
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get -qq install    \
+              build-essential         \
+              libbrotli-dev           \
+              libbz2-dev              \
+              libclang-${{ matrix.clang-version }}-dev \
+              libdb-dev               \
+              libgcrypt20             \
+              libgdbm-dev             \
+              libreadline-dev         \
+              libidn2-dev             \
+              libldap2-dev            \
+              liblzma-dev             \
+              libnghttp2-dev          \
+              libpcre3-dev            \
+              libpsl-dev              \
+              librtmp-dev             \
+              libtool                 \
+              libzstd-dev             \
+              rcs                     \
+              tcl-dev                 \
+              tk-dev                  \
+              zlib1g-dev
+
+      # Runs a single command using the runners shell
+      # Working dir is /home/runner/work/c2rust/c2rust
+      - name: Build c2rust
+        env:
+          LLVM_CONFIG_PATH: /usr/bin/llvm-config-${{ matrix.clang-version }}
+        run: cargo build --release
+
+      - name: Build tools
+        run: |
+          (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
+          (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
+
+      # Use separate restore/save steps so we still save the cache even if the
+      # testsuite fails.
+      - name: Restore postprocess cache
+        uses: actions/cache/restore@v4
+        with:
+          path: c2rust-postprocess/tests/llm-cache
+          # Restore the newest cache matching this runner/compiler prefix.
+          key: postprocess-cache-${{ matrix.runner }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-
+
+      - name: Run c2rust testsuite
+        env:
+          GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+        run: |
+          export PATH=$PWD/target/release:$PWD/c2rust-postprocess:$HOME/.local/bin:$PATH
+          echo "PATH=$PATH"
+          export C2RUST_DIR=$PWD
+          # Needs to be run from `tests/integration/` (or further inside)
+          # to correctly load the `pyproject.toml`.
+          (cd tests/integration && ./test.py json-c)
+
+      # Prune by `metadata.toml`'s atime so hot entries survive and stale ones
+      # disappear; GitHub cache eviction is what keeps the daily snapshot
+      # history bounded.
+      - name: Prune stale postprocess cache entries
+        run: |
+          cutoff=$(date -d '90 days ago' +%s)
+          while IFS= read -r -d '' file; do
+            atime=$(awk -F ' = ' '$1 == "atime" { print $2; exit }' "$file")
+            if [[ -n "$atime" && $atime -lt $cutoff ]]; then
+              echo "evicting ${file%/metadata.toml}"
+              rm -rf "${file%/metadata.toml}"
+            fi
+          done < <(find c2rust-postprocess/tests/llm-cache -mindepth 4 -maxdepth 4 -name metadata.toml -print0)
+
+      - name: Save postprocess cache
+        if: always()
+        uses: actions/cache/save@v4
+        with:
+          path: c2rust-postprocess/tests/llm-cache
+          # Do not use hashFiles() here; atime changes on every cron run, so a
+          # content hash would just give us a per-run key anyway.
+          key: postprocess-cache-${{ matrix.runner }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-${{ github.run_id }}
--- a/.github/workflows/internal-testsuite.yml
+++ b/.github/workflows/internal-testsuite.yml
@@ -14,6 +14,8 @@ jobs:
   # This workflow contains a single job called "build"
   run-testsuite:
     strategy:
+      # ci-c2rust-postprocess.yml generates postprocess caches for this matrix.
+      # Keep both matrices in sync so this workflow can restore those caches.
       matrix:
         include:
           - runner: ubuntu-latest
@@ -28,6 +30,11 @@ jobs:
     name: "run-testsuite (${{ matrix.runner }}: ${{ matrix.os }} ${{ matrix.arch}}, Clang ${{ matrix.clang-version }})"
     # The type of runner that the job will run on
     runs-on: ${{ matrix.runner }}
+
+    env:
+      # Uses cache-only responses; skips uncached work and exercises cached
+      # postprocess output usage.
+      C2RUST_POSTPROCESS_ARGS: --on-error warn
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
@@ -105,6 +112,13 @@ jobs:
         # `cd tools` to pick up `tools/rust-toolchain.toml` (stable, for `edition = "2024"`)
         (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
         (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
+
+    - name: Restore postprocess cache
+      uses: actions/cache/restore@v4
+      with:
+        path: c2rust-postprocess/tests/llm-cache
+        # Restore the newest available cache entry that matches this prefix.
+        key: postprocess-cache-${{ matrix.runner }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-
 
     - name: Run c2rust testsuite
       run: |
--- a/c2rust-postprocess/postprocess/cache.py
+++ b/c2rust-postprocess/postprocess/cache.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import time
 from abc import ABC, abstractmethod
 from hashlib import sha256
 from pathlib import Path
@@ -187,6 +188,8 @@ class DirectoryCache(AbstractCache):
             return None
         logging.debug(f"Cache hit: {cache_file}:\n{toml}")
         data = tomli.loads(toml)
+        data["atime"] = int(time.time())
+        cache_file.write_text(to_multiline_toml(data))
 
         return data["response"]
 
@@ -205,6 +208,7 @@ class DirectoryCache(AbstractCache):
             "model": model,
             "messages": messages,
             "response": response,
+            "atime": int(time.time()),
         }
         toml = to_multiline_toml(data)
 
--- a/tests/integration/tests/templates.py
+++ b/tests/integration/tests/templates.py
@@ -199,17 +199,21 @@ def autogen_postprocess(
     if not (ag and isinstance(ag, bool)):
         return
 
-    params = {"args": ""}
+    args: list[str] = []
 
     exclude_file = conf.with_name("postprocess-exclude.yml")
     if exclude_file.exists():
         # args are relative to script dir
-        params["args"] = shlex.join(
-            ["--exclude-file", str(exclude_file.relative_to(conf.parent))]
-        )
+        args.extend(["--exclude-file", str(exclude_file.relative_to(conf.parent))])
+
+    extra_args = os.getenv("C2RUST_POSTPROCESS_ARGS")
+    if extra_args:
+        args.extend(shlex.split(extra_args))
 
     if verbose:
-        params["args"] = f"--log-level DEBUG {params['args']}".rstrip()
+        args[0:0] = ["--log-level", "DEBUG"]
+
+    params = {"args": shlex.join(args)}
 
     out_path = conf.with_name("postprocess.gen.sh")
     render_script(POSTPROCESS_SH, out_path, params)


.github/workflows/internal-testsuite.yml
```diff
@@ -31,6 +31,9 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+    - name: Set postprocess cache policy
+      run: echo "C2RUST_POSTPROCESS_ARGS=--on-error warn" >> "$GITHUB_ENV"
```
thedataking: Nit: description does not match what step actually does does.
Nit: why are you not using the `env` key like you do on line 151?


you should add a short comment noting why we don't want postprocessor errors to be fatal except outside the cronjob.

.github/workflows/internal-testsuite.yml
```diff
@@ -139,3 +142,79 @@ jobs:
           ${{ github.workspace }}/tests/integration/**/compile_commands.json
           ${{ github.workspace }}/tests/integration/tests/refactor-diffs/**
       if: always()
+
+  postprocess-test:
+    runs-on: ubuntu-latest
+
+    env:
+      GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+      C2RUST_POSTPROCESS_ARGS: --on-error warn
```
thedataking: I assume this is temporary? Once key is set, we want `keep-going` (the default, so no need to set it)?

thedataking (commented):
Missing: using the Github cache action to restore and save c2rust-postprocessor entries on the cronjob or PR runs. If that's why the PR is marked draft, no worries.

.github/workflows/internal-testsuite.yml
```diff
@@ -139,3 +142,79 @@ jobs:
           ${{ github.workspace }}/tests/integration/**/compile_commands.json
           ${{ github.workspace }}/tests/integration/tests/refactor-diffs/**
       if: always()
+
+  postprocess-test:
```
thedataking: Lines 5-10 control when this workflow runs. Should this go in a separate file (`ci-postprocess.yml`?)

You'll need to investigate how this affects caching ... not just for the postprocessor cache entries but also how it works with `Swatinem/rust-cache@v2` and `awalsh128/cache-apt-pkgs-action@latest`.

Crocodoctopus: I'm using `internal-testsuite.yml` for testing, since github requires a workflow to be on master before I can launch any jobs. Currently, `postprocess-test` will run on push, which is fine for this.

Keeping this open for when I make that change.

Crocodoctopus:
Most of this is pretty hacked together, I'll address feedback as it stabilizes.

.github/workflows/internal-testsuite.yml
Crocodoctopus: Yes, the cronjob is now `keep-going` (more specifically, not set).

.github/workflows/internal-testsuite.yml
Crocodoctopus: This is stale. (Except the comment, which I added)

Crocodoctopus:
Currently, the cronjob launches the same way as the c2rust-testsuite workflow to ensure the latter can access the former's cache (for testing). When the cronjob hits master, on-pr and on-push workflows should have access to that cache as well(?). Tests show c2rust-testsuite/run-testsuite is using cached data from the cronjob, which runs postprocess on `json_object_set_userdata` and `json_object_to_file`. 

Most real changes are in `.github/workflows/daily-cronjobs.yml`, almost every other change is temporary; I tried to keep TODOs for each change that shouldn't be reviewed. I'm leaving this draft until all temporaries are gone.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,118 @@
+name: daily-cronjobs
+
+# Daily cronjobs workflow. Individual jobs can be added here over time.
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 4 * * *"
+  # TODO: remove once testing concludes and this no longer needs pull_request.
+  pull_request:
```
thedataking: Why make this workflow run on all PRs. That uses tokens for no good reasons and risk blocking the work of others.

I sent you an example where runs were scoped to a particular branch name pattern:

```
push:
    branches:
      # Triggers on any branch containing 'postprocess-ci' in its name
      - '*postprocess-ci*'
```

This will allow you to run this workflow every time you push the branch.

.github/workflows/daily-cronjobs.yml
```diff
@@ -0,0 +1,118 @@
...
+jobs:
+  postprocess-test-and-cache:
+    runs-on: ubuntu-latest
+
+    env:
+      # TODO: remove once postprocess can pick its own model defaults again.
+      C2RUST_POSTPROCESS_MODEL: gemini-3.1-flash-lite
+      GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+      C2RUST_POSTPROCESS_CACHE_PREFIX: postprocess-cache-
+      # Generates new cache and tests fresh model responses.
```
thedataking: stray comment?

c2rust-postprocess/postprocess/transforms/base.py
```diff
@@ -102,9 +102,11 @@ def apply_file(
                 continue
 
             if ident_regex and not ident_regex.search(identifier):
+                # TODO: remove this temporary ident-filter log once the
+                # narrow cronjob test surface is gone.
                 logging.info(
-                    f"Skipping Rust fn {identifier} in {rust_source_file}"
-                    f"due to ident filter {ident_filter}"
+                    f"Skipping Rust fn {identifier} in {rust_source_file} "
+                    "due to ident filter"
```
thedataking: this seems like a trivial change; I don't see why this PR would touch this file at all. Just more stuff for me to review and for you to undo.

.github/workflows/internal-testsuite.yml
```diff
@@ -29,6 +29,14 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ${{ matrix.runner }}
 
+    env:
+      # TODO: remove once postprocess can pick its own model defaults again.
+      C2RUST_POSTPROCESS_MODEL: gemini-3.1-flash-lite
+      # Uses cache-only responses; skips uncached work and exercises cached
+      # postprocess output usage.
+      C2RUST_POSTPROCESS_ARGS: --on-error warn
+      C2RUST_POSTPROCESS_CACHE_PREFIX: postprocess-cache-
```
thedataking: why do you need this?

.github/workflows/daily-cronjobs.yml
```diff
@@ -0,0 +1,118 @@
...
+
+jobs:
+  postprocess-test-and-cache:
+    runs-on: ubuntu-latest
+
+    env:
+      # TODO: remove once postprocess can pick its own model defaults again.
+      C2RUST_POSTPROCESS_MODEL: gemini-3.1-flash-lite
+      GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+      C2RUST_POSTPROCESS_CACHE_PREFIX: postprocess-cache-
```
thedataking: I don't think you need this. If this is to avoid hardcoding the cache prefix, you just move the problem one level down by hardcoding the variable name that contains the cache prefix. 

just use the same string when you populate the cache here and when you restore it in testsuite runs. See also comment w.r.t. cache key.

.github/workflows/daily-cronjobs.yml
```diff
@@ -0,0 +1,118 @@
...
+      # Generates new cache and tests fresh model responses.
+
+    # Steps represent a sequence of tasks that will be executed as part of the job.
+    steps:
+      - name: Require model key
```
thedataking: I don't think we should have this step. I'm not sure it is correct or how it will interact with forked repos.

.github/workflows/internal-testsuite.yml
```diff
@@ -29,6 +29,14 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ${{ matrix.runner }}
 
+    env:
+      # TODO: remove once postprocess can pick its own model defaults again.
+      C2RUST_POSTPROCESS_MODEL: gemini-3.1-flash-lite
```
thedataking: Why do you set this? Outside of the cronjob run we want the postprocessor to do nothing on cache misses.

.github/workflows/internal-testsuite.yml
```diff
@@ -40,7 +48,7 @@ jobs:
         # `update = none` is set, which makes this not work, so we need to update them manually.
         # See `.gitmodules` for more info.
         submodules: false
-
```
thedataking, line 43: Spurious whitespace change.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,118 @@
...
+      # Generates new cache and tests fresh model responses.
+
+    # Steps represent a sequence of tasks that will be executed as part of the job.
+    steps:
+      - name: Require model key
+        if: ${{ env.GEMINI_API_KEY == '' }}
+        run: exit 1
+
+      - name: Checkout c2rust
+        uses: actions/checkout@v4
```
thedataking, line 34: Don't strip the helpful comment from `internal-testsuite.yml`; I didn't remember why we pass submodules `none` here:

```
    - name: Checkout c2rust
      uses: actions/checkout@v4
      with:
        repository: immunant/c2rust
        # `update = none` is set, which makes this not work, so we need to update them manually.
        # See `.gitmodules` for more info.
        submodules: false
```

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,118 @@
...
+            llvm
+            llvm-dev
+            libzstd-dev
+            rcs
+            tcl-dev
+            tk-dev
+            zlib1g-dev
+
+      # Explicit apt install is required so dpkg sees build-essential for the
+      # testsuite requirement check, even though it is also listed in the
```
thedataking: why do we not have to do this in `internal-testsuite.yml` then?

.github/workflows/daily-cronjobs.yml
```diff
@@ -0,0 +1,118 @@
...
+          # no longer needed.
+          export C2RUST_POSTPROCESS_ARGS="--ident-filter '^(json_object_set_userdata|json_object_to_file)$'"
+          (cd tests/integration && ./test.py json-c)
+
+      - name: Save postprocess cache
+        if: ${{ success() }}
+        uses: actions/cache/save@v4
+        with:
+          path: c2rust-postprocess/tests/llm-cache
+          key: ${{ env.C2RUST_POSTPROCESS_CACHE_PREFIX }}${{ github.job }}-${{ github.run_id }}-${{ github.run_attempt }}
```
thedataking: I don't follow the reasoning behind this key. Won't this create a bunch of unique caches and increase our storage use very day (storage is metered on github).

Why not use the same key everywhere? The cronjob should pick up the cache from the previous day and add to it; hopefully most days it will mostly be cache hits so the API use will be minimal.

If you look at `internal-testsuite.yml`, it uses a matrix strategy:
```
 strategy:
      matrix:
        include:
          - runner: ubuntu-latest
            os: Linux
            arch: x86_64
            clang-version: 18
          - runner: ubuntu-22.04
            os: Linux
            arch: x86_64
            clang-version: 15
      fail-fast: false
    name: "run-testsuite (${{ matrix.runner }}: ${{ matrix.os }} ${{ matrix.arch}}, Clang ${{ matrix.clang-version }})"
    # The type of runner that the job will run on
    runs-on: ${{ matrix.runner }}
```

My best *guess* is that most of the time, we will generate the same cache entries no matter whether we're on ubuntu-22.04/clang-15 or ubuntu-latest/clang-18. However, if my guess is wrong, it _could_ make sense to use separate caches for the two different OS/compiler pairings. We should keep things as simple as possible here so that suggests starting with a single cache until we have evidence that it does not work well.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,118 @@
...
+        env:
+          LLVM_CONFIG_PATH: /usr/bin/llvm-config-18
+        run: cargo build --release
+
+      - name: Build tools
+        run: |
+          (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
+          (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
+
+      - name: Restore postprocess cache
```
thedataking: Why do you need to have separate restore and save steps? 

The [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#input-parameters-for-the-cache-action) show you the simplest possible use:

```
- name: Cache Gradle packages
  uses: actions/cache@v4
  with:
    path: |
      ~/.gradle/caches
      ~/.gradle/wrapper
```

This will automatically restore and save the cache. Only use two steps if you have to.

You can also find a commented out use in `internal-testsuite.yml`:

```
# - name: Cache testsuite compile_commands
    #   uses: actions/cache@v4
    #   with:
    #     path: |
    #       ${{ github.workspace }}/tests/integration/tests/**/compile_commands.json
    #     key: ${{ runner.os }}-ccdb-${{ steps.get-image-ver.outputs.version }}
```

(You won't find any live uses of the cache action since Khyber switched to more specialized caches for apt packages and rust, etc.)

c2rust-postprocess/postprocess/transforms/comments.py
```diff
@@ -105,12 +105,17 @@ def apply_ident(
...
+            # TODO: remove this temporary cache-hit logging once the cronjob
+            # postprocess path is no longer being debugged in CI.
+            logging.info(f"Cache hit for {identifier}; using cached response")
             return
 
         response = self.model.generate_with_tools(messages)
 
         if response is None:
             if api_key_from_env(model) is None:
+                # TODO: remove this temporary cache-miss warning once the
```
thedataking: this is not a temporary warning; don't remove it.

.github/workflows/daily-cronjobs.yml
```diff
@@ -0,0 +1,118 @@
...
+          (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
+          (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
+
+      - name: Restore postprocess cache
+        uses: actions/cache/restore@v4
+        with:
+          path: c2rust-postprocess/tests/llm-cache
+          key: ${{ env.C2RUST_POSTPROCESS_CACHE_PREFIX }}
+
+      - name: Run json-c testsuite
```
thedataking: I think you should have started from the corresponding snippet in `internal-testsuite.yml`:

```
    - name: Run c2rust testsuite
      run: |
        find tests/integration -type f -name compile_commands.json -delete
        export PATH=$PWD/target/release:$PWD/c2rust-postprocess:$HOME/.local/bin:$PATH
        echo "PATH=$PATH"
        export C2RUST_DIR=$PWD
        # Needs to be run from `tests/integration/` (or further inside)
        # to correctly load the `pyproject.toml`.
        (cd tests/integration && ./test.py curl json-c lua nginx zstd libxml2 python2 libmcs)
```

reasons:
- should use same name; don't want json-c in the step name since we'll be adding more projects
- drops useful comments about why we cd into `tests/integration`
- easier to understand when steps are similar
- I think you're right to drop the `compile_commands.json` deletion step since it is probably a vestige from when we cached compile commands in `internal-testsuite.yml`. If you agree, open separate PR to remove the compile_command.json deletion and commented-out cache step in `internal-testsuite.yml`.

finally, a note on process (no action required, just sharing for future reference):

I'm wondering if you passed passed `--verbose` like this:
`(cd tests/integration && ./test.py  --verbose json-c)`

... would you still have to make the logging tweaks to the python files? Fine either way since you marked the stuff you need to undo (I would have put that in a separate commit since its easy to drop and easy to mark that commit "DO NOT MERGE")

thedataking (changes requested):
## general

- I don't think the caching is set up the way we want it. We want a single file that we reuse every day unless you determine and explain precisely why that is a bad idea.
- several spurious changes needs cleanup
- the new .yml file should be called `ci-c2rust-postprocess.yml`, same for workflow name. it is far more important to highlight *what* is being tested than *when* something is being tested.

## cache trimming

did you consider how to prevent the cache from growing unreasonably large over time?

I think the postprocessor touches cache entries on hit to mark when they were last used. If the way github archives and restores cache files preserves last modified time, you can simply delete cache entries that have not been used for the last N weeks or months. Let's start with something like 3 months or 12 weeks (whatever is easier), we can always expire entries faster if the cache grows too large.

## avoiding duplication

several steps are (or ought to be) identical between the new yml file and the existing ones, e.g. provisioning. 

is there an *easy* way to deduplicate such steps? if deduplication is not super easy; just drop it. the duplication is mild enough that we do want to do something even moderately complex and in any case, this is something we should discuss first (to agree on approach, if any) and handle on a separate PR anyway.

.github/workflows/daily-cronjobs.yml
Crocodoctopus: It's gone now.

c2rust-postprocess/postprocess/transforms/base.py
Crocodoctopus: It was for testing, the `--ident-filter` was making it needlessly annoying to search logs. Its in the DNM now.

.github/workflows/internal-testsuite.yml
Crocodoctopus: Cache prefix is now inlined.

.github/workflows/daily-cronjobs.yml
Crocodoctopus: Inlined.

.github/workflows/daily-cronjobs.yml
Crocodoctopus: Removed.

.github/workflows/internal-testsuite.yml
Crocodoctopus: For testing. The inner cache is keyed by model name, so even though we don't provide it an API key, it still needed to match the cronjob's model. This is in the DNM now.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: Comment is back.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: Edit: Switched over to `cache@v4` on the cronjob's side.

.github/workflows/internal-testsuite.yml
Crocodoctopus: Whitespace is back.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: I had to use `pull_request` because the cache generated by the cronjob was not visible to the `c2rust-testsuite` otherwise. This has since been moved to the DNM commit.

.github/workflows/daily-cronjobs.yml
Crocodoctopus: Done wrt keeping keep the steps similar. 

Here is the PR for the `compile_commands.json` part #1871.

.github/workflows/daily-cronjobs.yml
Crocodoctopus: Added a similar matrix for OS/clang pairings, as c2rust output can be appreciably different enough to cause a false cache hit.

Edit: Cache pruning is now done via a date-based pruning step, in the same workflow.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: This was because I was using `awalsh128/cache-apt-pkgs-action@latest` in the cronjob. I've since switched over to apt-get. Changing both of them to use cached packages should be another PR.

Crocodoctopus:
Reworked the cronjob cache to key on `postprocess-cache-{matrix}-{runner id}`, and the "consumer" job to key on `postprocess-cache-{matrix}-`, I also introduced a pruning job on the same workflow that cleans caches prefixed with `postprocess-cache-` that are older than 90 days. It looks like there was a lot of code duplication that can't really be helped.  Also renamed the `daily-cronjob.yml` to `ci-c2rust-postprocess.yml`.

Crocodoctopus:
Currently, we run postprocess twice on two separate github caches, but since internally it keys on the exact c2rust output, I think we can probably run these in series on use a unified cache?

Regardless, I think this should suffice.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,177 @@
+name: ci-c2rust-postprocess
+
+# Daily cronjobs workflow. Individual jobs can be added here over time.
```
thedataking: nit: this comment isn't saying much; I think we can do without it.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,177 @@
...
+      - name: Run c2rust testsuite
+        run: |
+          export PATH=$PWD/target/release:$PWD/c2rust-postprocess:$HOME/.local/bin:$PATH
+          echo "PATH=$PATH"
+          export C2RUST_DIR=$PWD
+          # Needs to be run from `tests/integration/` (or further inside)
+          # to correctly load the `pyproject.toml`.
+          (cd tests/integration && ./test.py json-c)
+
+  prune-postprocess-caches:
```
thedataking: It looks like this step is deleting the entire cache produced by the cronjob. 

I'm not sure it will ever go stale since you're re-generating it daily and also expiring it daily.

What I had in mind was expiring individual cache entries that have not been accessed for some time. The flow would be something like:

- workflow unpacks cache archive file (done automatically by cache action)
- c2rust postprocess runs, potentially adding new cache entries, updating entries that were hits
- pruning step: here you delete cache entries that have not been hit for a long time
- workflow packs cache archive file (also done automatically by cache action)

So we keep the same cache file around (precisely one per entry in the target matrix) but we do not keep old entries inside it around.


Your strategy would delete all cache entries whether they were recently used by the postprocessor or not and thus force re-generation almost daily. We should just keep the useful entries around.

Crocodoctopus: Ah, when you said prune cache, I assumed you meant the github cache, as I had to generate new entries due to cache being immutable once created. We can avoid that entirely by deleting and saving a statically named cache, but we wont be able to use the `cache@v4` action. (Also, the 1 day thing is part of the DNM commit, its 90 otherwise)

Basically, we want:
  1. Restore old cache.
  2. Run tests.
  3. Prune local entries.
  4. Delete the old GitHub cache archive.
  5. Save the new cache under the same key.

... I believe this can done in the same job?

Crocodoctopus:
Removed the prune step, as github removes unused caches after 7 days automatically. Introduced an inner cache prune based on `atime` instead, evicting individual cache entries after 90 days. Each cronjob creates a unique cache keyed on its matrix and unique run id.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,128 @@
...
+          (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
+          (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
+
+      - name: Cache postprocess cache
+        uses: actions/cache@v4
+        with:
+          path: c2rust-postprocess/tests/llm-cache
+          key: postprocess-cache-${{ matrix.runner }}-${{ matrix.os }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-${{ github.run_id }}
+          restore-keys: |
+            postprocess-cache-${{ matrix.runner }}-${{ matrix.os }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-
```
thedataking: I still think this key is too complicated, we shouldn't include both runner and os. I'd drop the latter.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      # If a cached leaf's metadata.toml's atime is beyond the cutoff, remove the leaf dir. 
+      - name: Prune stale postprocess cache entries
```
thedataking: the next step might update a cache entry you'd prune here, why aren't you running the prune step after c2rust-postprocessor runs?

.github/workflows/internal-testsuite.yml
```diff
@@ -106,6 +111,12 @@ jobs:
         (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
         (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
 
+    - name: Restore postprocess cache
+      uses: actions/cache/restore@v4
+      with:
+        path: c2rust-postprocess/tests/llm-cache
+        key: postprocess-cache-${{ matrix.runner }}-${{ matrix.os }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-
```
thedataking: please add comment explaining that this picks newest available cache with that prefix

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: Done.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: No reason; moved prune after postprocess run.

.github/workflows/internal-testsuite.yml
Crocodoctopus: Added a comment here, as well as a few others in spots that also rely on implicit github cache behaviors.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: We spoke about this offline: We decided to generate a new cache per day, and prune internally only. We'll make use of github's build-in 7 day stale cache eviction to keep the cache timeline bounded.

.github/workflows/ci-c2rust-postprocess.yml
thedataking: `pull_request` should be removed before this PR can land. 

If you are correct that you need `pull_request` here for the cache to be visible to `c2rust-testsuite`; don't we have a more fundamental problem on our hands? how is this gonna run just once a day then?

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,134 @@
...
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 4 * * *"
+  pull_request:
+
+jobs:
+  run-testsuite:
+    strategy:
```
thedataking, line 9: since this was copied from `internal-testsuite.yml` we should probably put a comment on the strategy over there saying that any updates to the matrix must be reflected in `ci-c2rust-postprocess.yml`.

.github/workflows/ci-c2rust-postprocess.yml
```diff
@@ -0,0 +1,134 @@
...
+          (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
+
+      - name: Cache postprocess cache
+        uses: actions/cache@v4
+        with:
+          path: c2rust-postprocess/tests/llm-cache
+          # `atime` is part of the cache's state, so keying by the hash (or a
+          # variant of it) would produce a hard-to-spot bug. `atime` is
+          # updated on each postprocess run, so we need a new cache each run
+          # even if the hash does not change.
```
thedataking: This comment is a good start. However, I think the whole story is:

- github caches are immutable so run must produce an unique key
- unique cache keys are usually generated using `hashfiles()`
- hashfiles would update the `atime` of cache files which we rely on to prune the postprocessor cache
- therefore, use $github.run_id to produce a unique cache key instead

The way to think about this is: If you resume work on this in 5 years or someone else started working on this, did we leave enough breadcrumbs to avoid having to discover this solution once more?

.github/workflows/internal-testsuite.yml
```diff
@@ -106,6 +111,13 @@ jobs:
         (cd tools && cargo build --release --manifest-path split_rust/Cargo.toml)
         (cd tools && cargo build --release --manifest-path merge_rust/Cargo.toml)
 
+    - name: Restore postprocess cache
+      uses: actions/cache/restore@v4
+      with:
+        path: c2rust-postprocess/tests/llm-cache
+        # Restore the newest available cache entry that matches this prefix.
+        key: postprocess-cache-${{ matrix.runner }}-${{ matrix.arch }}-clang-${{ matrix.clang-version }}-
```
thedataking, line 121: https://github.com/actions/cache/blob/main/restore/README.md#inputs

- `key` - An explicit key for a cache entry. See [creating a cache key](https://github.com/actions/cache/blob/main/README.md#creating-a-cache-key).
...
- `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for `key`.

Don't we need `restore-keys` here?

Edit: using `key` seems to work correctly but I still wonder if we ought to use `restore-keys` since we expect to restore based on prefix-match, not exact match. In other words, what is least surprising?

Crocodoctopus: As per [`cache`](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-key-matching): "The cache action first searches for cache hits for key and the cache version in the branch containing the workflow run. If there is no hit, it searches for prefix-matches for key, and if there is still no hit, it searches for restore-keys and the version."

As per [`restore`](https://github.com/actions/cache/blob/main/restore/README.md): "The restore action restores a cache. It works similarly to the cache action except that it doesn't have a post step to save the cache. This action provides granular ability to restore a cache without having to save it. It accepts the same set of inputs as the cache action."

I confirmed this against the logs: [run-testsuite](https://github.com/immunant/c2rust/actions/runs/28476015855/job/84400679956#logs) finds the exact cache produced by [postprocess-test-and-cache](https://github.com/immunant/c2rust/actions/runs/28476015727/job/84400679507#logs). Up to you whether to use `restore-keys` or not, but this is explicitly documented behavior.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: Done.

.github/workflows/ci-c2rust-postprocess.yml
Crocodoctopus: Updated the comment. There is a slight bit more context that was needed: even if `hashFiles` didn't touch `atime`, it is still an incorrect solution. It would need to both not touch `atime`, and include `atime` in its hash, as it is a necessary part of the cache's state. Since postprocess always bumps `atime` (a miss AND a hit does this), `hashFiles` effectively becomes a more complicated way of of producing a new per-run cache.

.github/workflows/internal-testsuite.yml
```diff
@@ -29,6 +31,11 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ${{ matrix.runner }}
 
+    env:
+      # Uses cache-only responses; skips uncached work and exercises cached
+      # postprocess output usage.
+      C2RUST_POSTPROCESS_ARGS: --on-error warn
```
thedataking, line 37: I don't think `C2RUST_POSTPROCESS_ARGS` is read anywhere.

I'm guessing this was read by your DNM commit. I think you should update `POSTPROCESS_SH` in templates.py to pass `C2RUST_POSTPROCESS_EXTRA_ARGS` on the line where we run `c2rust-postprocess` (note the `EXTRA` in the variable name). 

IIRC, your temporary commit modified `autogen_postprocess` instead; I don't think that's the way to go.

thedataking (changes requested):
- `C2RUST_POSTPROCESS_ARGS` isn't used AFAICT. See detailed instructions in comment. 
- have you verified that atime is preserved by the tarballs on restore? otherwise, the comment on lines 98-106 rests on a wrong premise AND nothing ever gets pruned.
- I asked you if you could use cache without explicit save/restore steps. I think we need to go back to your earlier design because we want to run the save step unconditionally (`if: always()`). I _think_ you had the right approach in an earlier version; my bad. Otherwise, a failing run throws away all the successful generations.

Did you remove the `ident-filter` stuff? We shouldn't run on all functions before this code has had some time to soak.

.github/workflows/internal-testsuite.yml
Crocodoctopus: You are in fact correct, it was never being read. Fixed.

Crocodoctopus:
> Did you remove the ident-filter stuff? We shouldn't run on all functions before this code has had some time to soak.

Good call, ~though It'll need a bit more work to be done. Postprocess in the CI takes `--exclude-file` which takes precedence over `--ident-filter`. I think both tags should work together anyways, so its worthwhile to fix.~
Edit: this was wrong, they already do this.

> have you verified that atime is preserved by the tarballs on restore? otherwise, the comment on lines 98-106 rests on a wrong premise AND nothing ever gets pruned.

I have, and it is definitely not.

Crocodoctopus:
Fixed `C2RUST_POSTPROCESS_ARGS` not being read, and restored the `--ident-filter` and confirmed it is working.

I also moved `atime` into `metadata.toml`, which maybe should be called `last_cache_hit` instead? Regardless, we now prune on that instead. Also cleaned up the comment before the save cache action. I also tested inner cache eviction [here](https://github.com/immunant/c2rust/actions/runs/28635098805/job/84919694326?pr=1857) by temporarily setting the prune time to 10m and moving the prune before the testsuite run step. Looks good to me.

thedataking:
Landing this work on PR 1883 instead.

PR #1858, opened 2026-06-16
transpile: handle module name collisions

randomPoison:
In testing against libgit2, we ran into a case where module names could collide, leading to a generated Rust file not being included in the build. The situation is that we have both a `hash.c` file and some files in a `hash` directory, e.g. `hash/sha1.c`. C2rust then wants to generate both

```rust
mod src {
    mod hash;
}
```

and

```rust
mod src {
    mod hash {
        mod sha1;
    }
}
```

In this case, the latter is what gets generated, effectively dropping `src/hash.rs` from the Rust build.

This PR has c2rust detect these collisions and rename the colliding module, so we'd generate something like

```rust
mod src {
    #[path = "hash.rs"]
    mod hash_;
    mod hash {
        mod sha1;
    }
}
```

### DIFF
--- a/c2rust-transpile/src/build_files/mod.rs
+++ b/c2rust-transpile/src/build_files/mod.rs
@@ -1,6 +1,7 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fs::{self, File};
 use std::io::Write;
+use std::ops::Bound::{Excluded, Unbounded};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
@@ -107,7 +108,7 @@ pub fn emit_build_files<'lcmd>(
     })
 }
 
-#[derive(Serialize)]
+#[derive(Clone, Debug, Serialize)]
 struct Module {
     path: Option<String>,
     name: String,
@@ -116,19 +117,23 @@ struct Module {
 }
 
 #[derive(Debug, Default)]
-struct ModuleTree(BTreeMap<String, ModuleTree>);
+struct ModuleTree {
+    children: BTreeMap<String, ModuleTree>,
+    path_modules: Vec<Module>,
+}
 
 impl ModuleTree {
     /// Convert the tree representation into a linear vector
     /// and push it into `res`
     fn linearize(&self, res: &mut Vec<Module>) {
-        for (name, child) in self.0.iter() {
+        res.extend(self.path_modules.iter().cloned());
+        for (name, child) in self.children.iter() {
             child.linearize_internal(name, res);
         }
     }
 
     fn linearize_internal(&self, name: &str, res: &mut Vec<Module>) {
-        if self.0.is_empty() {
+        if self.path_modules.is_empty() && self.children.is_empty() {
             res.push(Module {
                 name: name.to_string(),
                 path: None,
@@ -150,6 +155,55 @@ impl ModuleTree {
                 close: true,
             });
         }
+    }
+
+    fn insert_nested(&mut self, path: Vec<String>) {
+        let mut cur = self;
+        for name in path {
+            cur = cur.children.entry(name).or_default();
+        }
+    }
+
+    fn insert_path_module(
+        &mut self,
+        parent_path: &[String],
+        file_name: String,
+        module_name: String,
+    ) {
+        let mut cur = self;
+        for name in parent_path {
+            cur = cur.children.entry(name.clone()).or_default();
+        }
+        let name = cur.unique_child_name(&module_name);
+        cur.path_modules.push(Module {
+            path: Some(file_name),
+            name,
+            open: false,
+            close: false,
+        });
+    }
+
+    fn unique_child_name(&self, module_name: &str) -> String {
+        if !self.child_name_exists(module_name) {
+            return module_name.to_owned();
+        }
+
+        for i in 1.. {
+            let candidate = format!("{module_name}_{i}");
+            if !self.child_name_exists(&candidate) {
+                return candidate;
+            }
+        }
+
+        unreachable!("We tried all the numbers and couldn't find one that didn't collide")
+    }
+
+    fn child_name_exists(&self, module_name: &str) -> bool {
+        self.children.contains_key(module_name)
+            || self
+                .path_modules
+                .iter()
+                .any(|module| module.name == module_name)
     }
 }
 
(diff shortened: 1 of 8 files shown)


c2rust-transpile/src/build_files/mod.rs
```diff
@@ -151,6 +155,55 @@ impl ModuleTree {
...
+        });
+    }
+
+    fn unique_child_name(&self, module_name: &str) -> String {
+        if !self.child_name_exists(module_name) {
+            return module_name.to_owned();
+        }
+
+        for i in 1.. {
+            let candidate = format!("{module_name}{i}");
```
ahomescu: The transpiler renamer uses the scheme `{name}_{i}`, it would make sense to keep that for consistency.

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -182,17 +235,41 @@ fn convert_module_list(
...
+            let relpath = m.strip_prefix(build_dir).ok()?;
+            let path = module_path(relpath);
+            Some((m, path))
+        })
+        .collect::<Vec<_>>();
+    let collision_modules = internal_modules
+        .iter()
+        .filter_map(|(module, path)| {
+            internal_modules
+                .iter()
```
ahomescu: This is quadratic in the number of modules, it might be worth optimizing this a little. Maybe sort them first, or use a trie?

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -207,10 +284,25 @@ fn convert_module_list(
             }
         }
     }
+    for relpath in collision_relpaths {
+        let mut path = module_path(&relpath);
+        let module_name = format!("{}_", path.pop().unwrap());
```
ahomescu: `"{}_0"` for consistency with the transpiler? Or anything other than a trailing `_`?

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -207,10 +284,25 @@ fn convert_module_list(
...
+        let file_name = relpath.file_name().unwrap().to_str().unwrap().to_string();
+        module_tree.insert_path_module(&path, file_name, module_name);
```
ahomescu, line 305: Doesn't `insert_path_module` already checks for uniqueness? Why do you need the extra trailing underscore?

c2rust-transpile/src/build_files/mod.rs
ahomescu: Don't we also need to check the new module for collisions?

ahomescu (changes requested):

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -151,6 +155,55 @@ impl ModuleTree {
...
+        }
+
+        for i in 1.. {
+            let candidate = format!("{module_name}{i}");
+            if !self.child_name_exists(&candidate) {
+                return candidate;
+            }
+        }
+
+        unreachable!("We tried all the numbers and couln't find one that didn't collide")
```
thedataking: nit: couldn't

c2rust-transpile/src/build_files/mod.rs
```diff
@@ -182,17 +235,41 @@ fn convert_module_list(
     });
 
     let mut res = vec![];
-    let mut module_tree = ModuleTree(BTreeMap::new());
+    let mut module_tree = ModuleTree::default();
+    // A C file can have the same stem as a sibling directory, such as
```
thedataking, line 240: Edit: resolving since I don't want to be a distraction and this is a minor win at best.

Iterating over the internal modules and binaries separately simplifies 2nd loop (and avoids the repeated calls to `module_path` but that's a minor win):

```rust
      // A C file can have the same stem as a sibling directory, such as
      // `hash.c` and `hash/sha1.c`. The current nested module emitter cannot
      // represent both at `mod hash`, so emit the file module with an explicit
      // path and keep the directory in the nested tree.
      //
      // Compute each internal module's path (the sanitized identifier for every
      // path component) once here, then reuse it below instead of re-deriving it.
      let internal_modules = modules
          .iter()
          .filter_map(|m| {
              if tcfg.is_binary(m) {
                  return None;
              }
              let relpath = m.strip_prefix(build_dir).ok()?;
              let path = module_path(relpath);
              Some((m, path))
          })
          .collect::<Vec<_>>();

      // A module collides when another module's path is strictly longer and has
      // this module's path as a prefix (a file whose stem matches a sibling
      // directory). Record the result by index so we can consume the paths below.
      let is_collision = internal_modules
          .iter()
          .map(|(_, path)| {
              internal_modules
                  .iter()
                  .any(|(_, other)| other.len() > path.len() && other.starts_with(path))
          })
          .collect::<Vec<_>>();

      for ((m, path), collides) in internal_modules.into_iter().zip(is_collision) {
          if collides {
              // The file's stem shadows a sibling directory, so emit it as an
              // explicit-path module inside its parent instead of nesting it.
              let mut parent = path;
              let stem = parent.pop().unwrap();
              let file_name = m.file_name().unwrap().to_str().unwrap().to_string();
              module_tree.insert_path_module(&parent, file_name, format!("{stem}_"));
          } else {
              // The module is inside the build directory, use nested modules.
              module_tree.insert_nested(path);
          }
      }

      // Binaries and modules outside the build directory get an explicit `#[path]`.
      for m in &modules {
          // Internal (non-binary, in-build-dir) modules were handled above.
          if m.strip_prefix(build_dir).is_ok() && !tcfg.is_binary(m) {
              continue;
          }
          let relpath = diff_paths(m, build_dir).unwrap();
          let path = Some(relpath.to_str().unwrap().to_string());
          let name = get_module_name(m, true, false, false).unwrap();
          res.push(Module {
              path,
              name,
              open: false,
              close: false,
          });
      }
```

Claude wrote this. Feel free to close without action; just wanted to share an interesting alternative.

If we want to solve the `O(n * n)` complexity, that can be done inside `is_collision`.

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -39,11 +39,24 @@ impl<'a> MutVisitor for LoadModules<'a> {
             }
 
             ModKind::Unloaded => {
-                // Look for dir_path/foo.rs, then try dir_path/foo/mod.rs
-                let mut mod_file_path = self.dir_path.join(ident.as_str()).with_extension("rs");
-                if !self.source_map.file_exists(&mod_file_path) {
-                    mod_file_path = self.dir_path.join(ident.as_str()).join("mod.rs");
-                }
+                // An explicit `#[path = "..."]` overrides the usual file lookup;
```
ahomescu: This is not a transpile change, split it?

c2rust-refactor/src/ast_manip/load_modules.rs
ahomescu, line 62: I don't think this is correct for the `#[path = ...]` case, it should be the full path given instead.

c2rust-refactor/src/ast_manip/load_modules.rs
thedataking: Moved to #1894

c2rust-refactor/src/ast_manip/load_modules.rs
thedataking: Ack. Moving to #1894

ahomescu (changes requested):

ahomescu (approved):

PR #1859, opened 2026-06-17
refactor: Account for idents in function signatures in `reorganize_definitions`

randomPoison:
libgit2 surfaced a bug in reorganize-definitions where if a type was referenced only in a functions's signature (but not its body), reorganize-definitions would fail to emit an import for that type. The fix seems simple, just account for types used by the function signature.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -247,30 +247,23 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
         fn keep_items(items: &[P<Item>]) -> HashSet<NodeId> {
             let mut keep_items = HashSet::new();
             let mut used_idents = HashSet::new();
+            let mut collect_used_idents = |item: &Item| {
+                visit_nodes(item, |path: &Path| {
+                    if let [segment] = &path.segments[..] {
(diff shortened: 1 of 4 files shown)


ahomescu (approved):

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -247,30 +245,23 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
-                        ..
-                    }) => {
+                    ItemKind::Fn(box Fn { body: Some(_), .. }) => {
                         keep_items.insert(item.id);
-                        visit_nodes(&**body, |path: &Path| {
-                            if let [segment] = &path.segments[..] {
-                                used_idents.insert(segment.ident);
-                            }
-                        });
+                        collect_used_idents(item);
```
ahomescu, line 261: This is an extra refactoring that wasn't really necessary, I think the fix could have been one line. I don't hate the new code though, so let's land it as is.

thedataking (approved):
> I don't hate the new code though, so let's land it as is.

I agree. This looks reasonable to me too. Removing draft status and updating PR body accordingly.

ahomescu:
Is this ready to land?

PR #1860, opened 2026-06-17
transpile: Various fixes to type handling for builtins

Rua:
- Depends on #1865 (to ensure that `CTypeKind::Bool` is always present)
- Fixes #1707

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1306,17 +1306,18 @@ impl TypedAstContext {
         }
     }
 
-    /// Bubble types of unary and binary operators up from their args into the expression type.
-    ///
-    /// In Clang 15 and below, the Clang AST resolves typedefs in the expression type of unary and
-    /// binary expressions. For example, a BinaryExpr node adding two `size_t` expressions will be
-    /// given an `unsigned long` type rather than the `size_t` typedef type. This behavior changed
-    /// in Clang 16. This method adjusts AST node types to match those produced by Clang 16 and
-    /// newer; on these later Clang versions, it should have no effect.
+    /// Overrides the result types of certain expressions to one of the `PULLBACK_KINDS` where
+    /// appropriate. Then bubbles those types up to the parent expressions.
     ///
     /// This pass is necessary because we reify some typedef types (such as `size_t`) into their own
     /// distinct Rust types. As such, we need to make sure we know the exact type to generate when
     /// we translate an expr, not just its resolved type (looking through typedefs).
+    ///
+    /// Additionally, in Clang 15 and below, the Clang AST resolves typedefs in the expression type
+    /// of unary and binary expressions. For example, a BinaryExpr node adding two `size_t`
+    /// expressions will be given an `unsigned long` type rather than the `size_t` typedef type.
+    /// This behavior changed in Clang 16. This method adjusts AST node types to match those
+    /// produced by Clang 16 and newer.
     pub fn bubble_expr_types(&mut self) {
         struct BubbleExprTypes<'a> {
             ast_context: &'a mut TypedAstContext,
@@ -1379,6 +1380,30 @@ impl TypedAstContext {
                         self.ast_context,
                         self.ast_context.c_exprs[&e].kind.get_qual_type().unwrap(),
                     ),
+                    CExprKind::Call(result_type_id, callee_id, _) => {
+                        let CExprKind::ImplicitCast(_, callee_id, CastKind::BuiltinFnToFnPtr, _, _) =
+                            self.ast_context.index_unwrap_parens(callee_id).kind
+                        else {
+                            return;
+                        };
+                        let CExprKind::DeclRef(_, decl_id, _) =
+                            self.ast_context.index_unwrap_parens(callee_id).kind
+                        else {
+                            return;
+                        };
+                        let CDeclKind::Function { ref name, .. } = self.ast_context[decl_id].kind
+                        else {
+                            return;
+                        };
+
+                        match name.as_str() {
+                            "__builtin_object_size" => {
+                                let type_id = self.ast_context.type_for_kind(&CTypeKind::Size);
+                                Some(result_type_id.with_ctype(type_id))
+                            }
+                            _ => None,
+                        }
+                    }
                     CExprKind::Paren(_ty, e) => self.ast_context.c_exprs[&e].kind.get_qual_type(),
                     CExprKind::UnaryType(_, op, _, _) => {
                         // All of these `CUnTypeOp`s should return `size_t`.
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -114,30 +114,37 @@ impl<'c> Translation<'c> {
                 ))
             }
             "__builtin_signbit" | "__builtin_signbitf" | "__builtin_signbitl" => {
+                let val = self.convert_expr(ctx.used(), args[0], None)?;
                 self.import_num_traits(args[0])?;
 
-                let val = self.convert_expr(ctx.used(), args[0], None)?;
                 Ok(val.map(|v| {
                     let val = mk().method_call_expr(v, "is_sign_negative", vec![]);
-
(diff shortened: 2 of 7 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -1379,6 +1379,30 @@ impl TypedAstContext {
...
+                        };
+
+                        match name.as_str() {
+                            "__builtin_object_size" => {
+                                let type_id = self.ast_context.type_for_kind(&CTypeKind::Size);
+                                Some(result_type_id.with_ctype(type_id))
+                            }
+                            _ => None,
+                        }
+                    }
```
fw-immunant, line 1590: If we're adding more logic to `bubble_expr_types` that is not simply bubbling up types of operations to smooth over pre/post Clang-16 differences in the AST, then we should at the very least include a comment inline noting that this is happening.

fw-immunant (approved):
LGTM overall, but would appreciate a comment where mentioned inline.

c2rust-transpile/src/c_ast/mod.rs
Rua: I've made some changes to the documentation now.

fw-immunant (changes requested):
NAK on the comment change - now the name of `bubble_expr_types` and its comment disagree on its primary purpose. It exists to homogeneize types between Clang versions by bubbling up types from child AST nodes to parent ones (and these types are not necessarily ones in PULLBACK_KINDS, e.g. in the case of bit shifts), and the other things it does are unrelated to this function. They should really be done separately, so we can remove the function when we no longer support Clang <16. Translating arbitrary C operations producing one type to Rust ones that producing a different type (based on the operation's identity rather than its children) seems best done in expr translation rather than by altering types here.

We don't necessarily need to change where this logic happens in this PR, but I meant to ask for a comment that documents the tech debt here rather than making it sound like the current confusing arrangement is by design.

Rua:
> They should really be done separately, so we can remove the function when we no longer support Clang <16.

Iterating through the whole AST is somewhat expensive, so I figure it's more efficient to fit it all together like this.

> Translating arbitrary C operations producing one type to Rust ones that producing a different type (based on the operation's identity rather than its children) seems best done in expr translation rather than by altering types here.

Perhaps, but the difficulty here is that the "bubbling" is done bottom-up while translation operates top-down. For example, if you have an expression like `sizeof(Foo) + 1` then you want the override on the `sizeof` type to propagate to the `+` expression too. The current code does that, but doing it with a top-down `convert_expr` call tree is much harder and messier. Though perhaps #1679 could provide a solution there?

Rua:
I've now renamed `bubble_expr_types` to better reflect what it actually does.

c2rust-transpile/tests/snapshots/builtin_object_size.c
```diff
@@ -0,0 +1,4 @@
+void test_builtin_object_size(void) {
+    int x = 0;
+    unsigned long n = __builtin_object_size(&x, 1);
```
ahomescu: Could use a few more tests for the other builtins.

c2rust-transpile/tests/snapshots/snapshots__transpile@builtin_object_size.c.2021.clang15.snap
```diff
@@ -0,0 +1,24 @@
...
+    unused_assignments,
+    unused_mut
+)]
+#![feature(raw_ref_op)]
+#[no_mangle]
+pub unsafe extern "C" fn test_builtin_object_size() {
+    let mut x: ::core::ffi::c_int = 0 as ::core::ffi::c_int;
+    &raw mut x;
+    let mut n: ::core::ffi::c_ulong = (if 1 as ::core::ffi::c_int & 2 == 0 {
+        -1isize as usize
```
ahomescu: Now we have a literal suffix and 2 casts? That seems like overkill.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -290,107 +281,140 @@ impl<'c> Translation<'c> {
...
+                    vec![
+                        (mk().path_segment("last_mut"), vec![]),
+                        (mk().path_segment("unwrap"), vec![]),
+                        (mk().path_segment("as_mut_ptr"), vec![]),
+                    ],
+                );
+
+                source_type_id = target_type_id;
+                let target_type_kind = &self.ast_context.resolve_type(target_type_id.ctype).kind;
+                let needs_cast = match target_type_kind {
```
ahomescu, line 403: Wouldn't `make_cast` handle this automatically? Why do you need a separate implementation here?

c2rust-transpile/tests/snapshots/builtin_object_size.c
Rua: I don't really know how the builtins are used or what their use cases are, so I wouldn't know what would be a good test for them.

c2rust-transpile/src/translator/builtins.rs
Rua: The issue is that there may not be a `CTypeId` corresponding to `*mut u8`, so there is no type to pass to `make_cast`.

c2rust-transpile/tests/snapshots/snapshots__transpile@builtin_object_size.c.2021.clang15.snap
Rua: The `isize` suffix is needed because -1 is negative and won't fit in a `usize`. The final cast is the one that casts the result of the builtin to `ulong`, so it mimics the C AST.

c2rust-transpile/tests/snapshots/builtin_object_size.c
```diff
@@ -0,0 +1,4 @@
...
+}
```
ahomescu: You could add a second line here to test an explicit cast:
```
unsigned long n2 = (unsigned long)__builtin_object_size(&x, 1);
```

c2rust-transpile/tests/snapshots/builtin_object_size.c
ahomescu: See my other comment below for a single line change here.

For other builtins, here's something you could add that should confirm that these builtins return an `int` that's the same as `int32_t`:
```
#include <stdint.h>

void test_builtin_expected_type(double d, int y) {
    int32_t ffs = __builtin_ffs(y);
    int32_t clz = __builtin_clz((unsigned)y);
    int32_t isnan = __builtin_isnan(d);
    int32_t isinf_sign = __builtin_isinf_sign(d);
    int32_t constant_p = __builtin_constant_p(y);
    int32_t cond = y ? __builtin_ffs(y) : 0;
}
```

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -174,85 +164,86 @@ impl<'c> Translation<'c> {
...
-                Ok(val.map(|x| {
-                    let call = mk().method_call_expr(x, seg, vec![]);
-                    mk().cast_expr(call, mk().path_ty(vec!["i32"]))
-                }))
+                source_type_id.ctype = self.ast_context.type_for_kind(&CTypeKind::Bool);
+
+                val.map(|x| mk().method_call_expr(x, seg, vec![]))
             }
             "__builtin_isinf_sign" => {
+                let zero = self.mk_int_lit(ctx.used(), result_type_id, 0, IntBase::Dec, false)?;
```
ahomescu: Two of these use `result_type_id`, the third `target_type_id`. Shouldn't they all be the same?

ahomescu (changes requested):

c2rust-transpile/src/translator/builtins.rs
Rua: That was a mistake, fixed now.

c2rust-transpile/tests/snapshots/builtin_object_size.c
Rua: Done.

c2rust-transpile/tests/snapshots/builtin_object_size.c
Rua: Added.

PR #1861, opened 2026-06-18
tests: integration: add `libgit2`

thedataking:
This code base surfaced transpiler and refactorer issues fixed in PRs #1858 and #1859. Adding `libgit2` helps us test the correctness of the fixes and catch regressions.

PR #1862, opened 2026-06-18
transpile: Support `__atomic_thread_fence`/`__atomic_signal_fence` builtins

thedataking:
Needed to land #1861.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -4,11 +4,12 @@
 use super::*;
 
 use crate::format_translation_err;
-use crate::translator::atomics::CAtomicBinOp;
+use crate::translator::atomics::{order_ty_name, CAtomicBinOp};
 use crate::translator::simd::simd_fn_from_builtin_fn;
 use c2rust_rust_tools::RustEdition::Edition2024;
 use log::warn;
 use std::sync::atomic::Ordering::Acquire;
+use std::sync::atomic::Ordering::Relaxed;
 use std::sync::atomic::Ordering::Release;
 use std::sync::atomic::Ordering::SeqCst;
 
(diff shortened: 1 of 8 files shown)


c2rust-transpile/src/translator/builtins.rs
```diff
@@ -525,6 +526,36 @@ impl<'c> Translation<'c> {
...
+                        builtin_name
+                    )
+                })?;
+                let call_expr = if order == Relaxed {
+                    mk().tuple_expr(vec![])
+                } else {
+                    let base_name = if builtin_name == "__atomic_thread_fence" {
+                        "fence"
+                    } else {
+                        "singlethreadfence"
```
ahomescu: The docs mention https://doc.rust-lang.org/stable/std/sync/atomic/fn.compiler_fence.html as a stable alternative, could we use that instead?

ahomescu (changes requested):

tests/unit/builtins/src/atomics.c
```diff
@@ -66,3 +66,25 @@ void new_atomics(const unsigned buffer_size, int buffer[const])
     __atomic_store_n(&x, 0, __ATOMIC_RELAXED);
     buffer[i++] = x;
 }
+
+void fences(const unsigned buffer_size, int buffer[const])
```
ahomescu, line 70: A snapshot might be useful too.

c2rust-transpile/src/translator/builtins.rs
thedataking: > compiler_fence corresponds to atomic_signal_fence in C and C++.

yup, that seems like the better choice for the else arm. it is both in std and core since 1.21.

ahomescu (approved):

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -642,11 +642,41 @@ impl<'c> Translation<'c> {
         method_name: &str,
         args: &[CExprId],
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
+        // The two operands can be distinct same-width types (e.g. `size_t`/`usize`
+        // vs `unsigned long`/`u64`), so coerce both to the output (pointee) type,
+        // which is also where `overflowing_*`'s result is stored.
```
fw-immunant: This can only occur for the `_add_`/`_sub_`/etc. builtins, not the `_addl_`/`_subl_` or other type-specific ones. Noting that in this comment would clarify how this situation relates to the types visible at the link a few lines above.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -642,11 +642,41 @@ impl<'c> Translation<'c> {
...
+        };
+        let result_ty = self.convert_type(result_ty_id)?;
+
         let args = self.convert_exprs(ctx.used(), args, None)?;
         args.and_then_try(|args| {
             let [a, b, c]: [_; 3] = args
                 .try_into()
                 .map_err(|_| "`convert_overflow_arith` must have exactly 3 arguments")?;
+            let a = mk().cast_expr(a, result_ty.clone());
+            let b = mk().cast_expr(b, result_ty);
```
fw-immunant: If we really wanted we could only add each cast in the case where the given input's type differs from the result type.

fw-immunant (approved):
A couple inline comments, no blocking concerns.

c2rust-transpile/src/translator/builtins.rs
thedataking: This is in code that should land in PR 1863, not here. Addressing in separate PR.

PR #1863, opened 2026-06-18
transpile: Cast overflow builtin operands to the output type

thedataking:
The type-generic `__builtin_{add,sub,mul}_overflow` accept operand and result types that all differ, which the previous translation did not handle. When all three types translate to one Rust type, keep mapping to the native `overflowing_*` methods; otherwise compute in `i128` and derive the overflow flag from a narrowing roundtrip check. Mixed-type uses involving 128-bit integers are rejected rather than translated wrongly (#1878); a new snapshot-test knob pins which decls are rejected.

Covers the overflow-builtin uses in libgit2 (#1861), which are all ≤64-bit integer types.

~~Possibly superseded by the broader builtin type-handling rework in #1860 / #1707.~~
Tagging @Rua for coordination.

### DIFF
--- a/c2rust-transpile/src/translator/builtins.rs
+++ b/c2rust-transpile/src/translator/builtins.rs
@@ -680,32 +680,93 @@ impl<'c> Translation<'c> {
         method_name: &str,
         args: &[CExprId],
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
+        let &[a_arg, b_arg, out_arg] = args else {
+            return Err(TranslationError::generic(
+                "`convert_overflow_arith` must have exactly 3 arguments",
+            ));
+        };
+
+        let arg_type = |arg: CExprId| {
+            self.ast_context
+                .index_unwrap_parens(arg)
+                .kind
+                .get_type()
+                .ok_or_else(|| TranslationError::generic("overflow builtin argument has no type"))
+        };
+        let result_ty_id = self
+            .ast_context
+            .get_pointee_qual_type(arg_type(out_arg)?)
+            .ok_or_else(|| TranslationError::generic("overflow builtin output is not a pointer"))?
+            .ctype;
+        let type_kind = |ty| &self.ast_context.resolve_type(ty).kind;
+        let a_kind = type_kind(arg_type(a_arg)?);
+        let b_kind = type_kind(arg_type(b_arg)?);
+        let result_kind = type_kind(result_ty_id);
+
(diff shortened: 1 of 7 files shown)


Rua:
@thedataking I don't think there is any overlap between the two PRs, since this one deals with the parameter types and not the actual result type of the function.

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -642,11 +642,41 @@ impl<'c> Translation<'c> {
...
+            }
+        };
+        let result_ty = self.convert_type(result_ty_id)?;
+
         let args = self.convert_exprs(ctx.used(), args, None)?;
         args.and_then_try(|args| {
             let [a, b, c]: [_; 3] = args
                 .try_into()
                 .map_err(|_| "`convert_overflow_arith` must have exactly 3 arguments")?;
+            let a = mk().cast_expr(a, result_ty.clone());
```
ahomescu: gcc is being very loose with the type system here:
> These built-in functions promote the first two operands into infinite precision signed type and perform addition on those promoted operands. The result is then cast to the type the third pointer argument points to and stored there. [...] The first built-in function allows arbitrary integral types for operands and the result type must be pointer to some integral type other than enumerated or boolean type

I'm not sure casting to the result type is correct here.

thedataking: I added additional tests where the first two types are wider than the result type. The current implementation failed those tests. Should be fixed now; good catch!

c2rust-transpile/src/translator/builtins.rs
```diff
@@ -680,32 +680,102 @@ impl<'c> Translation<'c> {
...
+                    mk().ident_pat(&over_name),
                 ]),
                 None,
                 Some(overflowing),
+            )))];
+            let mut overflowed = mk().ident_expr(&over_name);
+            let mut out_name = result_name.clone();
+
+            if let Some(result_ty) = result_ty {
+                // `c2rust_result as result_ty` truncates silently. Casting back
```
ahomescu, line 751: One alternative here is to to `$result_ty::try_from($result_name)`. That should return `Ok` if the value fits, or `Err(TryFromIntError)` otherwise.

ahomescu (approved):

c2rust-transpile/src/translator/builtins.rs
thedataking: Addressed separately on #1879 so this PR can land and unblock `libgit2` support.

PR #1865, opened 2026-06-18
transpile: Ensure that there are always types for primitive type kinds

Rua:
This allows code elsewhere to rely on these kinds always existing, for calling `type_for_kind` and the like. This can be expanded to include more types in the future if needed.

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -503,6 +503,28 @@ impl ConversionContext {
             self.visit_node(untyped_context, node_id, new_id, expected_ty)
         }
 
+        // Check what primitive kinds were emitted by the compiler.
+        let mut found_kinds: HashMap<_, _> = CTypeKind::PRIMITIVE_KINDS
+            .into_iter()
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -503,6 +503,28 @@ impl ConversionContext {
             self.visit_node(untyped_context, node_id, new_id, expected_ty)
         }
 
+        // Check what primitive kinds were emitted by the compiler.
```
ahomescu, line 506: Nit: this could be moved up just before `PULLBACK_KINDS`, for consistency with the order from `mod.rs` (and IMHO handling the primitives first then the pullbacks makes sense).

ahomescu (approved):

c2rust-transpile/src/c_ast/conversion.rs
Rua: It can't be, because it relies on `c_types` being filled by the main node processing code right above it first. If it's moved, then `c_types` will still be empty. The idea here was to add these types only if the compiler didn't already provide them.

PR #1868, opened 2026-06-23
tests: make cleanup_attr trace statics thread-local to fix race

thedataking:
The six tests share `trace_buf`/`trace_n` but libtest runs them in parallel, so they intermittently race. Marking the statics `_Thread_local` (faithfully transpiled to `#[thread_local]`) gives each test thread its own trace.

Failed run: https://github.com/immunant/c2rust/actions/runs/27427103335/job/81067618174


fw-immunant (approved):
LGTM, unsure whether other tests technically need the same fix.

PR #1874, opened 2026-06-27
Fix a bug in array indexing and update the snapshots

ahomescu:


thedataking (approved):

thedataking:
@Rua we're merging this fix to turn `master` green. Please take a look and object if you see an issue with the fix. We can always revert it and land an a better version.

PR #1876, opened 2026-06-29
postprocess: Add basic code parity check to comment transform

### LINKED ISSUE #1869: postprocess: Validate full function structure before applying model output
`c2rust-postprocess` currently verifies only that comments were transferred correctly. It does not verify that the model preserved the rest of the function structure. 

This lets malformed rewrites slip through. For example, if a model output drops a closing } from a function body, that invalid Rust can still be written into the tree. The result is two failures:
1)  **The updated source no longer builds**.
2)  A later merge_rust invocation can fail because it parses the root Rust file up front and expects valid Rust syntax.

Failure has been observed by running `c2rust-postprocess` on `json-c-transpile/src/linkhash.rs` with `gemini-3.1-flash-lite`. Smarter models don't appear to exhibit this behavior, probably still worth fixing.

Crocodoctopus:
Adds a small token based check to ensure the llm doesn't touch the code. This is not foolproof, as adding whitespaces can slip through and still prevent building, but it is enough to prevent a fairly common class of failure (see #1869) on lower-end llms.

### DIFF
--- a/c2rust-postprocess/postprocess/definitions/__init__.py
+++ b/c2rust-postprocess/postprocess/definitions/__init__.py
@@ -10,6 +10,7 @@ import tree_sitter_rust as tsrust
 from pygments import lex
 from pygments.lexer import RegexLexer
 from pygments.lexers.c_cpp import CLexer
+from pygments.lexers.rust import RustLexer
 from pygments.token import Comment
 from tree_sitter import Language, Node, Parser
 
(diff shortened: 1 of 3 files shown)


thedataking:
I don't think this is quite the right approach. I think the postprocessor should support validation hooks which it can call to check the generated output. Could be a script to run `cargo build` or run tests. If a validation hook fails, we can roll back the project to a known good state.

PR #1879, opened 2026-07-02
transpile: Use try_from to detect overflow builtin result truncation

thedataking:
Follow-up to #1863 addressing @ahomescu's review feedback: emit `result_ty::try_from(result).is_err()` instead of casting the narrowed result back up to `i128` and comparing. The checks are equivalent; `try_from` reads better in the generated code:

```rust
let (c2rust_result, c2rust_overflowed) = (a as i128).overflowing_add(b as i128);
let c2rust_result_narrow = c2rust_result as ::core::ffi::c_ulonglong;
*out = c2rust_result_narrow;
return (c2rust_overflowed || ::core::ffi::c_ulonglong::try_from(c2rust_result).is_err())
    as ::core::ffi::c_int;
```

The `try_from` path is derived from the already-converted result type, so crate-local aliases (`size_t`) and absolute paths (`::core::ffi::c_int`) both come out right. A new `overflow_mixed.c` snapshot pins the generated idiom.

Based on #1863; will retarget to master when that merges.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@overflow_mixed.c.2021.clang15.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@overflow_mixed.c.2021.clang15.snap
@@ -1,0 +1,37 @@
+---
+source: c2rust-transpile/tests/snapshots.rs
+expression: cat tests/snapshots/overflow_mixed.2021.clang15.rs
+---
+#![allow(
+    clippy::missing_safety_doc,
+    dead_code,
+    non_camel_case_types,
+    non_snake_case,
+    non_upper_case_globals,
+    unused_assignments,
+    unused_mut
+)]
+#[no_mangle]
+pub unsafe extern "C" fn mixed_types(
+    mut a: ::core::ffi::c_longlong,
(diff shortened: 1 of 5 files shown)


c2rust-transpile/tests/snapshots/snapshots__transpile@overflow_mixed.c.2021.clang15.snap
```diff
@@ -0,0 +1,37 @@
...
+        as ::core::ffi::c_int;
+}
+#[no_mangle]
+pub unsafe extern "C" fn narrow_result(
+    mut a: ::core::ffi::c_longlong,
+    mut b: ::core::ffi::c_longlong,
+    mut out: *mut ::core::ffi::c_int,
+) -> ::core::ffi::c_int {
+    let (c2rust_result, c2rust_overflowed) = (a as i128).overflowing_mul(b as i128);
+    let c2rust_result_narrow = c2rust_result as ::core::ffi::c_int;
```
ahomescu, line 33: If you're using `try_from` below, you don't need this cast here because `Ok(x)` is what you want.

PR #1882, opened 2026-07-03

thedataking:
The refactor tool's module loader ignored `#[path = "..."]` and always derived the file from the module name. The transpiler emits `#[path]` for modules whose name was deconflicted from a sibling (e.g. `hash.c` and `hash/` becoming `hash_` and `hash`), so honor it and fall back to the name-based lookup otherwise.


thedataking:
Folded into #1858 — the honor-`#[path]` change is the refactor-side companion to that PR's transpile-side emit of `#[path]`, so they belong in one PR.

PR #1889, opened 2026-07-07
transpile: Add `convert_decl_ref` and `convert_enum_constant_decl_ref`

Rua:
Some of the refactorings needed for #1620, split off so that they can be merged earlier.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3600,13 +3600,11 @@ impl<'c> Translation<'c> {
             .ok_or_else(|| format_err!("name not declared: '{}'", varname))?;
 
         // Import the referenced global decl into our submodule
-        if self.tcfg.reorganize_definitions {
-            self.add_import(decl_id, &rustname);
-            // match decl {
-            //     CDeclKind::Variable { is_defn: false, .. } => {}
-            //     _ => self.add_import(decl_id, &rustname),
-            // }
-        }
+        self.add_import(decl_id, &rustname);
+        // match decl {
+        //     CDeclKind::Variable { is_defn: false, .. } => {}
+        //     _ => self.add_import(decl_id, &rustname),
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -3695,6 +3570,145 @@ impl<'c> Translation<'c> {
...
+            } = decl
+            {
+                return Err(format_translation_err!(
+                    None,
+                    "Cannot refer to static duration variable in a const expression",
+                ));
+            }
+        }
+
+        if let CDeclKind::EnumConstant { .. } = decl {
```
ahomescu, line 3595: You're moving this before the `reorganize_definitions` check below, is this correct?

Rua: In principle yes, because `convert_enum_constant_decl_ref` calls `enum_constant_expr`, which does the importing itself. What puzzles me more is why the import is gated behind `reorganize_definitions` in the first place? Should `enum_constant_expr` be doing the same? I don't understand the original logic behind it.

ahomescu: In `reorganize_definitions` mode we emit the code coming in from `#include`d headers in separate sub-modules. In that case, each use of a header declaration needs to import it from its sub-module, so the transpiler calls `add_import`. There is one interesting thing going on there: that method calls `add_use_item` which checks if the imported item lives in the same module as the use, and it skips the import in that case. I think that makes the `if self.tcfg.reorganize_definitions` check redundant before `self.add_import`.

It's weird to me because both checks were introduced in the same commit https://github.com/immunant/c2rust/commit/7154048c138, but they still feel redundant.

Rua: I've added a new commit to remove the check.

ahomescu (approved):

PR #1890, opened 2026-07-07
transpile: Use correct types for enums with in/decrement

Rua:
Some of the refactorings needed for #1620, split off so that they can be merged earlier.


ahomescu:
I merged #1892 so now we have some extra enums tests. Could you add this case to the enums snapshot and/or unit tests?

Rua:
@ahomescu Done.

ahomescu (approved):

PR #1894, opened 2026-07-08

thedataking:


c2rust-refactor/src/ast_manip/load_modules.rs
ahomescu: Copying comment from #1858 

> I don't think this is correct for the `#[path = ...]` case, it should be the full path given instead.

ahomescu (changes requested):

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -39,11 +50,24 @@ impl<'a> MutVisitor for LoadModules<'a> {
...
 
             ModKind::Unloaded => {
-                // Look for dir_path/foo.rs, then try dir_path/foo/mod.rs
-                let mut mod_file_path = self.dir_path.join(ident.as_str()).with_extension("rs");
-                if !self.source_map.file_exists(&mod_file_path) {
-                    mod_file_path = self.dir_path.join(ident.as_str()).join("mod.rs");
-                }
+                // An explicit `#[path = "..."]` overrides the usual file lookup;
+                // the transpiler emits this for modules whose name was deconflicted
+                // from a sibling (e.g. `hash.c` and `hash/` becoming `hash_` and `hash`).
```
ahomescu: Wouldn't `hash.c` be `hash_1`?

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -39,11 +50,24 @@ impl<'a> MutVisitor for LoadModules<'a> {
...
+                let path_attr = attrs
+                    .iter()
+                    .find(|attr| attr.has_name(sym::path))
+                    .and_then(|attr| attr.value_str());
+
+                let mod_file_path = if let Some(path) = path_attr {
+                    self.dir_path.join(path.as_str())
```
ahomescu, line 62: Do we need to strip the extension here? If there's a `#[path = "foo.rs"]` that has submodules, do those start at `foo/`?

c2rust-refactor/src/ast_manip/load_modules.rs
```diff
@@ -52,16 +76,27 @@ impl<'a> MutVisitor for LoadModules<'a> {
...
                     .parse_mod(&token::Eof)
-                    .expect("failed to parse {mod_file_path:?}");
+                    .expect(&format!("failed to parse {mod_file_path:?}"));
 
                 attrs.append(&mut inner_attrs);
                 *mod_kind = ModKind::Loaded(items, Inline::No, inner_span);
+
+                if path_attr.is_some() {
+                    // If the module was loaded from a `#[path="..."]` attribute, use
+                    // the parent directory of the specified file for any child modules.
```
ahomescu, line 86: See comment above: should it be just the parent directory, or `{parent_dir}/{mod_name}`?

ahomescu: Answering my own question:
> For path attributes on modules not inside inline module blocks, the file path is relative to the directory the source file is located.

ahomescu (approved):

PR #1900, opened 2026-07-12
transpile: recognize `ssize_t` in stdio.h and unistd.h

thedataking:
glibc may declare `ssize_t` in `stdio.h` or `unistd.h`, depending on header inclusion order. Recognize those declaration sites so all translation units map it consistently to `isize`.

Includes a regression test for the `stdio.h` case.


fw-immunant (approved):
What an ugly implementation detail. LGTM.

PR #1904, opened 2026-07-14
Roll toolchain to nightly-2023-04-15

thedataking:
This rolls the compiler-coupled workspace from `nightly-2022-11-03` to
`nightly-2023-04-15`, updates CI/provisioning/MSRV metadata, and migrates the
rustc-private APIs used by PDG, instrumentation, refactor, and analyzer.
Generated Edition 2021/2024 pins and the standalone stable tools remain
unchanged.

The most important review areas, highest risk first:

- Refactor AST, macro, token, and transformation handling (`e45e9f117`,
  `df0617272`, `e4aeeaf12`). This reconciles the generated AST model with rustc,
  adds traversal for interpolated macro arguments, and reparses item headers
  when rustc does not retain tokens. Follow-up review hardened structured
  `format_args!` equivalence, inherited-visibility insertion (including an item
  at source byte zero), unresolved-use handling, HIR reverse lookup, and
  AutoRetype's treatment of pre-existing compiler errors. Focused regressions
  cover each case. `3c1d3ed29` also fixes a pre-existing reversed edge direction
  in nonterminal ID matching that the macro rewrites depend on.
- Analyzer MIR-to-source rewrite mapping (`257998b74`, `c313757aa`). The target
  compiler splits some expression loads across MIR statements, folds user
  variables into `LocalKind::Temp`, and emits `PlaceMention`. The migrated
  ordering/origin logic fixes the current fixtures, and MIR unlowering now uses
  `LocalInfo::User` so named variables are not mistaken for compiler-generated
  temporaries. Skipping closure body owners still deserves scrutiny in case
  downstream analysis expects closures to be handled independently.
- Dynamic instrumentation (`af584c03a`). Query, place-projection, pointer-type,
  terminator, and unwind APIs changed. End-to-end instrumentation passes
  locally, but drop/unwind behavior on code outside the existing fixtures
  remains a meaningful risk.
- Snapshot changes (`a1c3aef68`, `10aec3989`). PDG changes are limited to
  allocation identifiers and shifted MIR statement locations; graph topology,
  permissions, and counts are unchanged. The refactor snapshots reflect newly
  working macro traversal/precedence and omission of the default `Vec` allocator
  parameter rather than bulk compiler-output churn.

Follow-up migration review also made every workspace member inherit the
workspace `rust-version = "1.70"`, so published crates retain the documented
MSRV instead of silently accepting the newer compiler used for this roll.

Bisection note: the source cannot support both rustc-private API surfaces at once, so the commits from the pin bump (`643224221`) through the last migration commit (`c313757aa`) intentionally do not build individually. When bisecting across this range, skip it: `git bisect skip 643224221..c313757aa`.


ahomescu:
Are we still keeping the instrumentation code updated? We had a discussion at some point to let that fall behind.

c2rust-refactor/gen/ast.txt
```diff
@@ -423,14 +448,13 @@ enum PatKind {
...
-struct Lit { token_lit, kind, span }
-
 struct StrLit { style, symbol, suffix, span, symbol_unescaped }
 
+struct MetaItemLit { symbol, suffix, kind, span }
+
 enum LitKind {
     Str(sym, style),
-    ByteStr(bytes),
+    ByteStr(#[rewrite_ignore] bytes, style),
```
ahomescu: Interesting, I wonder why it decided to add the attribute here.

thedataking:
> Are we still keeping the instrumentation code updated? We had a discussion at some point to let that fall behind.

From the meeting discussion: we'll update it for this roll. 

For next roll, we could first create a backup branch and then disable unused components on `master`.

c2rust-refactor/gen/ast.txt
thedataking: Good catch, seems it wasn't actually needed. Dropped since the test suite passes either way.

(It's still on `ExprKind::IncludedBytes`, where it is indeed needed)

c2rust-refactor/src/analysis/ownership/mod.rs
```diff
@@ -43,6 +43,18 @@ mod mono;
 /*
 mod mono_filter;
 */
+
+/// Render a definition path in the stable format used by ownership annotations
+/// and analysis fixtures. This intentionally includes every rustc
+/// disambiguator, including zero-valued ones.
+pub fn def_path_string_no_crate(tcx: TyCtxt<'_>, def_id: DefId) -> String {
```
ahomescu, line 54: Why is this needed now?

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -203,9 +203,141 @@ impl Make<FloatTy> for ty::FloatTy {
...
+                tokens: TokenStream::new(tokens),
+            })
+        }
+    }
+}
+
 impl Make<Lit> for rustc_hir::Lit {
     fn make(self, _mk: &Builder) -> Lit {
-        Lit::from_lit_kind(self.node, self.span)
+        token_lit_from_kind(&self.node)
```
ahomescu, line 348: Can we keep the span here somehow?

c2rust-refactor/src/ast_manip/util.rs
```diff
@@ -187,9 +186,9 @@ impl UseInfo {
     /// Retrieve the list of Idents and their NodeIds defined by the given UseTree
     pub fn from_use_tree(tree: &UseTree, id0: NodeId) -> Vec<Self> {
         match &tree.kind {
-            UseTreeKind::Simple(_, id1, id2) => vec![Self {
+            UseTreeKind::Simple(_) => vec![Self {
                 ident: tree.ident(),
-                ids: vec![id0, *id1, *id2],
+                ids: vec![id0],
```
ahomescu, line 191: I think this is wrong now, we needed those extra node IDs. I'm not sure where to get them now.

ahomescu: Ah never mind, they changed the AST upstream so the separate imports from the different name spaces show up as differend `UseTreeKind::Simple` nodes.

c2rust-refactor/src/rewrite/strategy/print.rs
```diff
@@ -853,7 +851,8 @@ impl RewriteAt for Item {
                 describe_rewrite(inner_span, reparsed_span, &rcx);
                 let mut rw =
                     TextRewrite::adjusted(inner_span, reparsed_span, self.get_adjustment(&rcx));
-                RecoverChildren::recover_children(&reparsed, &m_items, rcx.enter(&mut rw));
+                let new_items = m_items.iter().cloned().collect::<Vec<_>>();
```
ahomescu: Is this needed? It's immediately borrowed, so it seems like an unnecessary copy.

c2rust-refactor/src/context.rs
```diff
@@ -837,7 +870,7 @@ impl<'a, 'tcx> RefactorCtxt<'a, 'tcx> {
                     None
                 } else {
                     let path = self.try_resolve_use_id(item.id)?;
-                    namespace(&path.res)
+                    path.res.iter().find_map(namespace)
```
ahomescu: `find_map` returns the first non-`None`, but `try_resolve_use_id` now returns up to 3 different items. We probably need to change this into `item_namespaces` that returns a `Vec<Namespace>` and update its callers (they're all in `reorganize_definitions.rs`).

c2rust-refactor/src/resolve.rs
```diff
@@ -80,7 +80,11 @@ pub fn module_children(tcx: TyCtxt, did: DefId) -> Vec<(Symbol, Res<!>)> {
                 module_children(tcx, krate.as_def_id())
             }
 
-            Use(ref path, _kind) => module_children(tcx, path.res.def_id()),
+            Use(ref path, _kind) => path
+                .res
+                .first()
```
ahomescu: I'm not sure we should grab only the first one, but all the resolved nodes instead.

ahomescu (changes requested):

tests/integration/tests/__init__.py
ahomescu, line 263: This could be its own separate PR from the roll.

c2rust-refactor/src/ast_builder/builder.rs
ahomescu: Or it might not be needed anymore, can you double check?

c2rust-refactor/src/analysis/ownership/mod.rs
thedataking: Split out into #1906 since it seems to fix a bug we had before the roll.

tests/integration/tests/__init__.py
thedataking: I think this was because this PR was not set to merge into `master`. Fixed.

c2rust-refactor/src/rewrite/strategy/print.rs
thedataking: No,  fixed by calling the existing slice impl explicitly.

c2rust-refactor/src/context.rs
thedataking: Done  following your suggestion: item_namespace is now item_namespaces and returns a `SmallVec<[Namespace; 3]>` preserving rustc's type/value/macro order, so no resolution is dropped.

c2rust-refactor/src/resolve.rs
thedataking: Done. Latest version now extracts every `Res::Def(DefKind::Mod, _)` entry (deduplicated) and concatenates children only from those; an import with no module resolution yields no children, matching the non-fatal unresolved-use behavior

c2rust-refactor/src/ast_builder/builder.rs
thedataking: The target `token::Lit` has no span field, so the token itself can't carry it. However, the span is still representable one level up, and we now keep it there.

ahomescu (approved):

c2rust-refactor/src/context.rs
```diff
@@ -829,28 +863,44 @@ impl<'a, 'tcx> RefactorCtxt<'a, 'tcx> {
         self.sess.crate_types().contains(&CrateType::Executable)
     }
 
-    pub fn item_namespace(&self, item: &Item) -> Option<Namespace> {
+    /// Return every namespace the given item occupies.
```
ahomescu, line 913: This is a refactoring that should be a separate PR.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -98,6 +97,13 @@ struct ModuleInfo {
 
     /// Set of item idents defined in this module, per namespace
     items: PerNS<HashSet<Ident>>,
+
+    /// Targets of the simple imports in this module, per namespace. Used to
+    /// recognize when an ident in `items` is only occupied by an import of a
+    /// declaration that is being moved into this module: such an import
+    /// collapses into a self-import and is removed when paths are updated,
+    /// so it does not block the move.
+    import_targets: PerNS<HashMap<Ident, DefId>>,
```
ahomescu, line 104: This and many other changes in this file are part of a major refactoring and fix that is not part of the toolchain roll. It should be a separate PR.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1616,12 +1782,11 @@ impl<'a, 'tcx> HeaderDeclarations<'a, 'tcx> {
...
 
+        // Each declaration is stored exactly once regardless of how many
+        // namespaces it occupies, so this emits every declaration once.
         let mut all_items = unnamed_items
-            .type_ns
             .into_iter()
-            .chain(unnamed_items.value_ns.into_iter())
-            .chain(idents.type_ns.into_values().flatten())
-            .chain(idents.value_ns.into_values().flatten())
+            .chain(idents.into_values().flatten())
```
ahomescu, line 1795: This seems like a non-trivial change that we should split.

c2rust-refactor/src/context.rs
thedataking: Done in #1911

c2rust-refactor/src/transform/reorganize_definitions.rs
thedataking: Done in #1911

c2rust-refactor/src/transform/reorganize_definitions.rs
thedataking: Done in #1912.

PR #1910, opened 2026-07-18
tests: integration: curl: add bear config file

ahomescu:
bear 4.0 switched to a brand new Rust rewrite which handles duplicates differently.
Add a new `bear.yml` configuration file for the curl test that fixes issues with that test.


thedataking (approved):
LGTM based on the understanding that for now, this is only needed by `curl`. If this was needed for all tests, we should adopt a more general solution.

PR #1911, opened 2026-07-21
refactor: handle multi-namespace reorganizer imports

thedataking:
A simple import can resolve simultaneously in Rust's type, value, and macro namespaces, but the reorganizer currently treats each declaration as occupying only one namespace.

Preserve rustc's `import_res_map` in `RefactorCtxt` and expose import resolutions as deterministic `(Namespace, Res)` pairs. Update reorganizer bookkeeping so declarations are stored once while being tracked, matched, moved, and deduplicated independently in every namespace they occupy.

This fixes two cases:

- A simple import resolving to both a type and a value now blocks an unrelated declaration in the appropriate namespace.
- Separate type/value imports that rewrite to the same path collapse into one valid multi-namespace import.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -98,6 +98,11 @@ struct ModuleInfo {
 
     /// Set of item idents defined in this module, per namespace
     items: PerNS<HashSet<Ident>>,
+
+    /// Targets of simple imports in this module, per namespace. This lets
+    /// destination selection distinguish a real name conflict from an import
+    /// of the declaration being moved into the module.
+    import_targets: PerNS<HashMap<Ident, DefId>>,
 }
 
 impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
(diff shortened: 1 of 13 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -618,8 +623,15 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                        } else if let ItemKind::Use(tree) = &item.kind {
+                            if matches!(tree.kind, UseTreeKind::Simple(..)) {
+                                let ident = tree.ident();
+                                for (namespace, _) in self.cx.resolved_imports_for_item(item) {
+                                    info.items[namespace].insert(ident);
+                                }
+                            }
                         } else {
-                            if let Some(namespace) = self.cx.item_namespace(item) {
+                            for namespace in self.cx.item_namespaces(item) {
```
ahomescu: This should return 0 or 1 items here since it can't be a `Use`, right? The code is probably fine as-is, so maybe a comment? Or be extra paranoid and `assert!`?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1000,13 +1044,35 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                                        let new_node_id = self.st.next_node_id();
+                                        let inserted = remapped_paths
+                                            .insert(new_node_id, (*parent, def_id))
+                                            .is_none();
+                                        assert!(inserted);
+                                        generated_import_targets
+                                            .insert(new_node_id, smallvec![(namespace, *parent)]);
+                                        items.push(
+                                            mk().id(new_node_id)
+                                                .use_simple_item(path, None::<String>),
```
ahomescu, line 1126: Something to look into in a follow up: might we be importing conflicting identifiers here, either across the 3 namespaces or within each individual one?

ahomescu (approved):

c2rust-refactor/src/transform/reorganize_definitions.rs
thedataking: Added assert.

c2rust-refactor/src/transform/reorganize_definitions.rs
thedataking: Within a namespace, a conflicting generated import hits the `panic!("Conflicting imports of ...")` in the dedup pass below, so we fail loudly rather than emitting bad code.

PR #1917, opened 2026-07-22
refactor: remove unused flamer instrumentation

thedataking:
Remove the broken `flamer` attribute instrumentation and dependency while retaining manual `flame` profiling.


c2rust-refactor/Cargo.toml
```diff
@@ -29,13 +28,12 @@ slotmap = {version = "1.0.7", features = ["unstable"]}
 derive_more = "=0.99.17"
 c2rust-macros = { path = "../c2rust-macros", version = "0.22.1" }
 flame = { version = "0.2.2", optional = true }
```
ahomescu: What's the difference between `flame` and `flamer`? Should we get rid of both?

thedataking: Done in separate commit.

ahomescu (approved):

PR #1918, opened 2026-07-22
refactor: Add support for bitfields in reorganize_definitions

ahomescu:
This was slop coded with Claude but it looks decent to me. It even added a snapshot test.

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -7,6 +7,8 @@ use std::mem;
 
 use crate::transform::Transform;
 use rustc_ast::ptr::P;
+use rustc_ast::token::{self, TokenKind};
+use rustc_ast::tokenstream::{TokenStream, TokenTree};
 use rustc_ast::util::comments::{Comment, CommentStyle};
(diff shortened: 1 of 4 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -2129,6 +2261,88 @@ fn is_nested(tree: &UseTree) -> bool {
...
+    for tree in delim_args.tokens.trees() {
+        let mut tree = tree.clone();
+        match &mut tree {
+            TokenTree::Token(tok, _) => match &mut tok.kind {
+                TokenKind::Ident(name, _) => {
+                    saw_ty = *name == ty_sym;
+                    saw_eq = false;
+                }
+                TokenKind::Eq => saw_eq = saw_ty,
+                TokenKind::Literal(lit) if saw_ty && saw_eq && lit.kind == token::LitKind::Str => {
```
ahomescu, line 2309: I don't think there's ever a case here where `!saw_ty && saw_eq` because of the line just before, so the `saw_ty` check here could be removed but I don't want to go down that rabbit hole.

randomPoison (approved):

PR #1920, opened 2026-07-23
transpile: Emit integer literal suffixes where possible

Rua:
A minor style improvement to the produced code.

### DIFF
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -2,10 +2,10 @@
 //! This code is used to generate literal expressions of various kinds.
 //! These include integer, floating, array, struct, union, enum literals.
 
-use failure::format_err;
-
 use super::*;
+use failure::format_err;
 use std::iter;
+use syn::{Path, TypePath};
 
 impl<'c> Translation<'c> {
     /// Generate an integer literal corresponding to the given type, value, and base.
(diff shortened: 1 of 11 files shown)


c2rust-transpile/src/translator/literals.rs
```diff
@@ -16,19 +16,34 @@ impl<'c> Translation<'c> {
...
-            IntBase::Dec => mk().int_unsuffixed_lit(val),
-            IntBase::Hex => mk().float_unsuffixed_lit(&format!("0x{:x}", val)),
-            IntBase::Oct => mk().float_unsuffixed_lit(&format!("0o{:o}", val)),
+        let target_ty = self.convert_type(ty.ctype)?;
+        let suffix = numeric_literal_suffix(&target_ty);
+        let is_suffix = suffix.is_some();
+
+        let lit_str = if let Some(suffix) = suffix {
+            match base {
+                IntBase::Dec => format!("{val}_{suffix}"),
```
ahomescu: @randomPoison @fw-immunant @thedataking Do we have a preference here between with and without the underscore? I'm slightly against it.

c2rust-transpile/src/translator/literals.rs
```diff
@@ -16,19 +16,34 @@ impl<'c> Translation<'c> {
...
+                IntBase::Hex => format!("0x{val:x}_{suffix}"),
+                IntBase::Oct => format!("0o{val:o}_{suffix}"),
+            }
+        } else {
+            match base {
+                IntBase::Dec => format!("{val}"),
```
ahomescu: These `format!` invocations might be expensive w.r.t. code size. You could refactor this so it computes the base number first, then appends the suffix with `String::push_str` (the result of `format` is a `String` anyway).

c2rust-transpile/src/translator/literals.rs
Rua: I find the underscores a little clearer, but it doesn't matter to me that much.

randomPoison: I'm slightly in favor of underscore, but seems like a pretty inconsequential detail.

fw-immunant: An anti-underscore vote here.

Rua: Oh no XD

ahomescu (approved):

PR #1921, opened 2026-07-24
transpile: Deduplicate in/decrement operator handling

Rua:
The code for post-increment basically reimplements part of the pre-increment code as well as parts of the compound assignment code. This PR deduplicates that by introducing a new common function, and also makes pre-increment emit a Rust compound assignment when it can, resulting in cleaner output.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@enums.c.2021.clang15.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@enums.c.2021.clang15.snap
@@ -41,10 +41,10 @@ pub unsafe extern "C" fn test_enums() {
     foo = (foo as ::core::ffi::c_uint).wrapping_add(1) as Foo;
     bar = (bar as ::core::ffi::c_int + 1) as Bar;
     let c2rust_fresh0 = foo;
-    foo = foo.wrapping_add(1);
+    foo = (foo as ::core::ffi::c_uint).wrapping_add(1) as Foo;
     bar = c2rust_fresh0 as Bar;
     let c2rust_fresh1 = bar;
-    bar = bar + 1;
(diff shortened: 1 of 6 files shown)


c2rust-transpile/tests/snapshots/snapshots__transpile@enums.c.2021.clang15.snap
```diff
@@ -41,10 +41,10 @@ pub unsafe extern "C" fn test_enums() {
     foo = (foo as ::core::ffi::c_uint).wrapping_add(1) as Foo;
     bar = (bar as ::core::ffi::c_int + 1) as Bar;
     let c2rust_fresh0 = foo;
-    foo = foo.wrapping_add(1);
+    foo = (foo as ::core::ffi::c_uint).wrapping_add(1) as Foo;
```
ahomescu, line 44: Where did these extra casts come from?

ahomescu (changes requested):

c2rust-transpile/tests/snapshots/snapshots__transpile@enums.c.2021.clang15.snap
Rua: They are produced the same way as they are for regular compound assignment `foo += 1`. Specifically, the `compute_lhs_type_id` and `compute_res_type_id`, which specify when a compound assignment has its arithmetic/logic operator performed using a different type from its LHS operand.

In this case, the casts are superfluous, but that is because `convert_cast_from_enum` and `convert_cast_to_enum` unconditionally cast. https://github.com/immunant/c2rust/pull/1620/commits/a589bee55f166952866704d5b7c431d35ebbc141 (as part of #1620) is meant to fix that.

ahomescu (approved):

PR #1924, opened 2026-07-25
refactor: reorganize_definitions: use the fallback/C ABI instead of Rust

ahomescu:
Both sites now use `Abi::FALLBACK` (= `Abi::C { unwind: false }`).

`Regression test: `test_reorganize_implicit_extern`
(`tests/snapshots/reorganize_implicit_extern.rs`); on pre-fix code its
implicit-`extern` declaration fails to merge with the `extern "C"` duplicate,
producing an `extern "Rust"` block and a spurious `compute_1` rename.

`reorganize_definitions.rs:843-847` and `:1779-1782` both do:

```rust
let abi = m.abi.and_then(|abi| abi::lookup(&abi.symbol.as_str())).unwrap_or(Abi::Rust);
```

Rust's default ABI for `extern {}` blocks is `"C"` (rustc's own lowering uses
`Abi::FALLBACK`, which is `Abi::C`). Consequences:

- An implicit `extern { fn foo(); }` block never dedups against an
  `extern "C" { fn foo(); }` declaration — the ABI-mismatch `continue` at
  `:2144` in `find_foreign_item` skips it.
- Worse, its items are re-emitted by `into_items` via
  `mk().extern_(Abi::Rust)` (`:1990`), silently changing the declared ABI of
  foreign functions.

Latent only because transpiled code usually writes `extern "C"` explicitly.
Fix: fall back to `Abi::C { unwind: false }` (or `Abi::FALLBACK`).

### DIFF
--- a/c2rust-refactor/tests/snapshots/reorganize_implicit_extern.rs
+++ b/c2rust-refactor/tests/snapshots/reorganize_implicit_extern.rs
@@ -1,0 +1,40 @@
+#![feature(register_tool)]
+#![register_tool(c2rust)]
+#![allow(non_camel_case_types)]
+#![allow(dead_code)]
+#![allow(unused_imports)]
+
+pub mod first {
+    #[c2rust::header_src = "/home/user/some/workspace/first.h:1"]
+    pub mod first_h {
+        // An `extern` block without an explicit ABI string defaults to "C",
+        // so this declaration must merge with the explicit `extern "C"`
+        // declaration of the same function in second.h below.
+        extern {
+            #[c2rust::src_loc = "3:0"]
+            pub fn compute(x: i32) -> i32;
+        }
+    }
+    use self::first_h::compute;
+
+    pub fn call_first(x: i32) -> i32 {
+        unsafe { compute(x) }
+    }
+}
+
+pub mod second {
+    #[c2rust::header_src = "/home/user/some/workspace/second.h:1"]
+    pub mod second_h {
+        extern "C" {
+            #[c2rust::src_loc = "2:0"]
+            pub fn compute(x: i32) -> i32;
+        }
+    }
+    use self::second_h::compute;
+
+    pub fn call_second(x: i32) -> i32 {
+        unsafe { compute(x) }
+    }
+}
+
+fn main() {}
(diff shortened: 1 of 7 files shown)


ahomescu:
According to https://doc.rust-lang.org/reference/items/external-blocks.html#r-items.extern.abi.default, `"C"` is indeed the default.

c2rust-refactor/tests/snapshots/reorganize_implicit_extern.rs
```diff
@@ -0,0 +1,40 @@
+#![feature(register_tool)]
+#![register_tool(c2rust)]
+#![allow(non_camel_case_types)]
+#![allow(dead_code)]
+#![allow(unused_imports)]
+
+pub mod first {
+    #[c2rust::header_src = "/home/user/some/workspace/first.h:1"]
+    pub mod first_h {
+        // An `extern` block without an explicit ABI string defaults to "C",
```
thedataking, line 10: The transpiler always emits an explicit `.extern_("C")` (c2rust-transpile/src/translator/mod.rs:1231, :1459, :2009, :2236, :2781), so an implicit extern block can't come from transpiled code.

ahomescu: Right, which is why we probably didn't catch this. Do we want to be defensive here, or punt this PR?

thedataking (changes requested):
From offline discussion: we don't have a C pattern that is not UB and would trigger this pattern. We assume that the transpiler operates on valid C and that the refactorer operates on whatever Rust the transpiler outputs.

ahomescu:
> From offline discussion: we don't have a C pattern that is not UB and would trigger this pattern. We assume that the transpiler operates on valid C and that the refactorer operates on whatever Rust the transpiler outputs.

Did you mean to reply to this PR or #1935 ? Edit: if you meant 1935, could you approve 1924 so we can land it, since it's a legitimate bug fix?

thedataking:
> > From offline discussion: we don't have a C pattern that is not UB and would trigger this pattern. We assume that the transpiler operates on valid C and that the refactorer operates on whatever Rust the transpiler outputs.
> 
> Did you mean to reply to this PR or #1935 ? Edit: if you meant 1935, could you approve 1924 so we can land it, since it's a legitimate bug fix?

My bad. Reviewing this one.

thedataking (changes requested):
From offline discussion:
- we shouldn't add tests here since it gives the impression that this bug can be triggered
- would be better to `expect(...)` here so this fails loudly. 
- alternatively, just fix the Abi to FALLBACK, don't add comments or tests.

thedataking (approved):
The fallback correction is sound. I’m withdrawing the expect and test-removal requests. Please describe this as defensive handling of unexpected input; we haven’t identified a triggering case in normal transpiler output.

PR #1925, opened 2026-07-25
refactor: reorganize_definitions: fix multi-namespace imports splitting

ahomescu:

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -1136,24 +1136,27 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
                         let mut items = smallvec![];
                         if let ItemKind::Use(_) = &item.kind {
                             if let Some((path, def_ids)) = multi_namespace_uses.get(&item.id) {
-                                // The retained import uses the first (type/value/
-                                // macro ordered) resolution. Determine where that
-                                // resolution lives even when it was not remapped.
-                                let other_mod_id = remapped_paths
-                                    .get(&item.id)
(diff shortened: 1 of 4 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1136,24 +1136,27 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                                        self.cx.def_qpath(first_def).1
+                                    }
+                                    None => path.clone(),
+                                };
                                 for &(namespace, def_id) in &def_ids[1..] {
                                     if let Some(Replacement { path, parent, .. }) =
                                         self.path_mapping.get(&def_id)
                                     {
-                                        if other_mod_id != *parent {
+                                        if !path.ast_equiv(&retained_path) {
```
thedataking, line 1161: [P2] Preserve aliases when emitting the second namespace import

This broader split condition now reaches `use_simple_item(path, None::<String>)` when the two targets are in the same module but have different names. That generated import ignores the original `as` alias and can introduce a name collision.

I reproduced this at stack head using `reorganize_split_renamed_import.rs`: change the import in `user` to `use crate::dest::dest_h::tick as alias;`, change its type/value uses to `alias`, and add `fn tick() {}` to `user`. The input compiles. The output adds `use crate::dest::tick;` alongside that function, producing E0255 (`tick` defined multiple times). The earlier transform produces compiling output for this case.

The synthesized import should preserve the original binding name, so this case emits `use crate::dest::tick as alias;`.

thedataking (commented):
Review of stack 1953: the newly generated split import needs to preserve the original binding name.

PR #1927, opened 2026-07-25
refactor: reorganize_definitions: keep glob imports untouched

ahomescu:


ahomescu:
If we're skipping stuff that's not emitted by c2rust, this one might be a no too because the transpiler doesn't emit globs.

PR #1928, opened 2026-07-25
refactor: canonicalize moved paths that refer to item left behind

ahomescu:
Follow-up limitation found while writing the test — FIXED: items moved out
of a header could not rely on bindings that stay behind in the header (e.g.
the kept glob). A bare path like `thing` that resolved through the glob was
not canonicalized on the way out (`is_relative_path` treats only
`self::`/`super::` as relative), so the moved item failed to compile with
E0412. `move_items` now runs `canonicalize_moved_decl_paths` over every
moved declaration (and its `impl` block): a single-segment path to a local,
non-moving item def is rewritten to its absolute path via `def_qpath`.
Paths are left alone when the destination module still binds the ident to
the same def — through a pre-existing import, an import moving along with
the declaration (moved `use` targets are now recorded in
`ModuleInfo::import_targets`, mirroring `update_module_info_items`), or
because the target is defined in the destination module itself — which
keeps the existing snapshots byte-identical. Paths to moved defs are still
rewritten later by `update_paths` via `path_mapping`. The
`reorganize_glob_import` test now uses the bare `thing` field type and
verifies it becomes `crate::defs::thing`.

---

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -27,7 +27,7 @@ use crate::ast_builder::mk;
 use crate::ast_manip::util::{
     is_c2rust_attr, is_exported, is_relative_path, join_visibility, namespace, split_uses,
 };
-use crate::ast_manip::{visit_nodes, AstEquiv, FlatMapNodes, MutVisitNodes};
+use crate::ast_manip::{visit_nodes, AstEquiv, FlatMapNodes, MutVisit, MutVisitNodes};
 use crate::command::{CommandState, Registry};
 use crate::driver::Phase;
 use crate::path_edit::fold_resolved_paths_with_id;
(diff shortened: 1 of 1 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1014,6 +1055,65 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                            dest_info.import_targets[ns]
+                                .get(&ident)
+                                .map_or(false, |&target| canonical(target) == canonical(def_id))
+                        });
+                        let defined_in_dest = def_id.as_local().map_or(false, |ldid| {
+                            let mod_hir_id = cx.ty_ctxt().parent_module_from_def_id(ldid);
+                            cx.hir_map().local_def_id_to_node_id(mod_hir_id) == dest_info.id
+                        });
+                        if !import_covers && !defined_in_dest {
+                            return cx.def_qpath(def_id);
```
thedataking, line 1108: [P2] Preserve generic arguments when qualifying a path

`def_qpath(def_id)` reconstructs the definition's name without the original path's generic arguments. This turns a field type `Value<i32>` into `crate::defs::Value`, causing E0107. The destination's glob import makes this a regression: the output before this PR already compiled because `Value<i32>` remained resolvable there. Please carry the original segment's generic arguments onto the qualified path.

This case uses a generic Rust type; it applies to code that has already been edited after transpilation.

<details>
<summary>Reproducer</summary>

The input compiles. Run `c2rust-refactor reorganize_definitions --rewrite-mode alongside -- repro.rs --edition=2021` and compile the resulting `repro.new` to see E0107.

```rust
#![feature(register_tool)]
#![register_tool(c2rust)]
#![allow(dead_code, non_camel_case_types, unused_imports)]

pub mod defs {
    pub struct Value<T> { pub val: T }
}
pub mod user {
    use crate::defs::*;
    #[c2rust::header_src = "/tmp/user.h:1"]
    pub mod user_h {
        use crate::defs::*;
        pub struct config { pub thing: Value<i32> }
    }
    pub fn go(_: user_h::config) {}
}
fn main() {}
```
</details>

thedataking (commented):
Review of stack 1953: qualifying a moved path drops its generic arguments.

PR #1929, opened 2026-07-25
refactor: handle impls correctly by moving instead of dropping them

ahomescu:

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -70,6 +70,15 @@ pub struct Reorganizer<'a, 'tcx: 'a> {
     // the strings using this map.
     bitfield_ty_defs: HashMap<NodeId, HashMap<String, DefId>>,
 
+    // `impl` blocks collected out of header modules by `remove_header_items`
+    // that were not carried alongside a moved declaration via
+    // `MovedDecl::r#impl` (trait impls, impls with non-`const` items, extra
+    // impls beyond the one carried per declaration, and impls whose self
+    // type was never itself individually processed as a moved declaration).
+    // `move_items` reattaches each one directly to its self type's
+    // destination module once that is known.
+    pending_impls: Vec<(DefId, MovedDeclImpl)>,
+
     // Counter used by `unique_ident`
     ident_counter: HashMap<Ident, usize>,
 
@@ -130,6 +139,7 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
             modules: IndexMap::new(),
             path_mapping: HashMap::new(),
             bitfield_ty_defs: HashMap::new(),
+            pending_impls: Vec::new(),
             stdlib_id: DUMMY_NODE_ID,
             ident_counter: HashMap::new(),
             widened_defs: HashSet::new(),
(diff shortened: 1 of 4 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -888,22 +930,64 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                    );
+                    continue;
+                }
+            };
+            let mut item = moved_impl.item;
+            self.canonicalize_moved_decl_paths(&mut item, dest_module_id, &matching_defs);
+            pending_impl_items
+                .entry(dest_module_id)
+                .or_default()
+                .push(item);
```
thedataking, line 988: [P2] Account for deferred impls when merging their self types

Pending impls bypass the declaration deduplication logic and are all appended after `path_mapping` has unified their self types. If two equivalent header types each have two const-only impl blocks, the first impl is carried and deduplicated with the type, but both copies of the second impl are appended to the same resulting type. The output then fails with E0592 (duplicate associated constant `B`).

The input below compiles, and the earlier transform produced compiling output. Please include the deferred impls in type-equivalence decisions and deduplicate equivalent blocks when their self types merge. If the deferred impls differ, those differences also need to affect whether the self types can be unified.

<details>
<summary>Reproducer</summary>

Run `c2rust-refactor reorganize_definitions --rewrite-mode alongside -- repro.rs --edition=2021` and compile the resulting `repro.new`.

```rust
#![feature(register_tool)]
#![register_tool(c2rust)]
#![allow(dead_code, non_camel_case_types)]

pub mod a {
    #[c2rust::header_src = "/tmp/types.h:1"]
    pub mod types_h {
        pub struct Thing { pub x: i32 }
        impl Thing { pub const A: i32 = 1; }
        impl Thing { pub const B: i32 = 2; }
    }
    pub fn go() -> i32 { types_h::Thing::A }
}
pub mod b {
    #[c2rust::header_src = "/tmp/types.h:1"]
    pub mod types_h {
        pub struct Thing { pub x: i32 }
        impl Thing { pub const A: i32 = 1; }
        impl Thing { pub const B: i32 = 2; }
    }
    pub fn go() -> i32 { types_h::Thing::A }
}
fn main() {}
```
</details>

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -888,22 +930,64 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
...
+                None => {
+                    warn!(
+                        "could not find a destination module for an `impl` \
+                         block on {self_def_id:?}; dropping it"
+                    );
+                    continue;
+                }
+            };
+            let mut item = moved_impl.item;
+            self.canonicalize_moved_decl_paths(&mut item, dest_module_id, &matching_defs);
```
thedataking, line 984: [P2] Rewrite the trait reference before moving a trait impl

`canonicalize_moved_decl_paths` uses `fold_resolved_paths_with_id`, which does not visit `TraitRef` paths. When an impl moves to its self type's module, its trait name can therefore remain relative to the old header scope.

In the input below, the trait moves to `crate::user::MyTrait` and the impl moves to `defs`, but the emitted impl still says `impl MyTrait for crate::defs::Thing`. The input compiles; the output fails with E0405 because `MyTrait` is not in scope in `defs`. Please rewrite the impl's trait reference, including any trait paths in bounds, as part of relocation.

This exercises the trait-impl support introduced here with edited Rust code.

<details>
<summary>Reproducer</summary>

Run `c2rust-refactor reorganize_definitions --rewrite-mode alongside -- repro.rs --edition=2021` and compile the resulting `repro.new`.

```rust
#![feature(register_tool)]
#![register_tool(c2rust)]
#![allow(dead_code, non_camel_case_types)]

pub mod defs {
    pub struct Thing { pub x: i32 }
}
pub mod user {
    #[c2rust::header_src = "/tmp/user.h:1"]
    pub mod user_h {
        pub trait MyTrait { fn val(&self) -> i32; }
        impl MyTrait for crate::defs::Thing {
            fn val(&self) -> i32 { self.x }
        }
    }
    pub fn go() {}
}
fn main() {}
```
</details>

thedataking (commented):
Review of stack 1953: two compile regressions in deferred impl handling, reproduced with compiling inputs and compared against the earlier transform.

PR #1930, opened 2026-07-25
refactor: fix nondeterministic ordering for foreign items

ahomescu:

### DIFF
--- a/c2rust-refactor/tests/snapshots/reorganize_extern_block_order.rs
+++ b/c2rust-refactor/tests/snapshots/reorganize_extern_block_order.rs
@@ -1,0 +1,69 @@
+#![feature(register_tool)]
+#![register_tool(c2rust)]
+#![allow(non_camel_case_types)]
+#![allow(dead_code)]
+
+// The destination all the standard library headers below are merged into.
+// It has to already exist for the `BEGIN`/`END` header comments to make it
+// into the output.
+pub mod stdlib {
+    pub const STDLIB_MARKER: i32 = 0;
+}
+
+pub mod api {
+    // Both headers are under `/usr/include`, so they are treated as standard
+    // library headers and all of their declarations land in the same `stdlib`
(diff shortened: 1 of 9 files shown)


c2rust-refactor/tests/snapshots/reorganize_extern_block_order.rs
```diff
@@ -0,0 +1,69 @@
...
+        pub struct aa_ty {
+            pub x: i32,
+        }
+
+        extern "C" {
+            #[c2rust::src_loc = "3:0"]
+            pub fn aa_c_fn() -> i32;
+        }
+
+        extern "sysv64" {
```
thedataking: [P1] Make the extern-block ordering test portable to macOS ARM

The new `test_reorganize_extern_block_order` runs unconditionally, but its fixture uses `extern "sysv64"`, which is unsupported on `aarch64-apple-darwin`. The harness compiles the input before running the transform, so this fails before testing ordering.

The [macOS ARM CI job](https://github.com/immunant/c2rust/actions/runs/30964293212/job/92174776208) fails this test with `error[E0570]: "sysv64" is not a supported ABI for the current target` at this line. I also reproduced that diagnostic locally by compiling a minimal `no_core` fixture for `aarch64-apple-darwin` with the pinned toolchain.

Please use two ABIs supported across the tested targets, or gate this architecture-specific fixture/test and keep portable ordering coverage for the other targets.

thedataking (commented):
One additional finding from checking CI before recommending approvals for stack 1953: the new extern-block ordering fixture is not portable to the supported macOS ARM target.

c2rust-refactor/tests/snapshots/reorganize_extern_block_order.rs
ahomescu: I replaced `sysv64` with `system` which should be supported everywhere.

PR #1936, opened 2026-07-25
refactor: compare foreign function declarations by resolved type

ahomescu:

Some relevant output from Claude:
```
    Regression test: `test_reorganize_foreign_fn_type_identity`
    (`tests/snapshots/reorganize_foreign_fn_type_identity.rs`).
    Each translation unit has its own `buf`, and the two differ in field visibility,
    so the transform correctly keeps them apart and renames the second to `buf_1`.
    Both units then declare `fill(b: *mut buf)` — spelled identically, but naming a
    different type in each unit.
    
    `find_foreign_item` compares the two with `compatible_fn_prototypes`.
    A foreign item has no body and so no typeck results to resolve against,
    and the comparison falls through to a purely syntactic check that sees `*mut buf`
    against `*mut buf` and reports a match. The second declaration is dropped
    and both call sites are pointed at the first, so the caller written against
    `buf_1` now passes it to a function taking `*mut buf`:
    
    Comparing foreign functions by resolved type routed the comparison through
    `structural_eq_tys_impl`, whose array arm lets a zero-length array match an
    array of any length. That leniency is there for an `extern` array
    declaration, which C lets omit the length that the definition gives, and
    which we translate to a zero length; it does not hold inside a function
    signature. `*mut [c_int; 0]` and `*mut [c_int; 4]` are distinct Rust types
    with no coercion between them, so two declarations that differ only there
    are not interchangeable at a call site.
    
    `test_reorganize_foreign_fn_rename` regressed on exactly that: its two
    `compute` declarations were collapsed into one and `b_call` was rewritten
    to call a function it cannot pass its argument to.
    
    `TypeCompare` now carries an `exact_array_lens` flag, off by default so
    that the `extern` array declaration it exists for keeps matching its
    definition, and on for `compatible_fn_sigs` and `compatible_fn_prototypes`.
    An array length reached through a signature, including one nested in a
    pointed-to struct, has to match exactly.

    `find_foreign_item` decided whether two foreign functions were the same by
    handing their `FnDecl`s to `compatible_fn_prototypes`. That resolves each
    parameter through `opt_node_type`, but a foreign item has no body and so no
    typeck results. Two declarations from different modules whose parameters were
    merely spelled alike were reported identical, one was dropped, and its
    callers were rewritten to a function taking a different type.
    
    `compatible_foreign_fns` now compares function signatures with `compatible_fn_sigs`.
    `match_exports` already compared foreign functions against external ones
    this way; this brings the crate-local comparison in line with it. The
    syntactic path is kept as a fallback for a signature that will not resolve,
    which is also why the arity check added earlier still matters.
    
    Structurally identical types from different modules still compare equal, so
    this only separates declarations that genuinely differ.
```

### DIFF
--- a/c2rust-refactor/src/context.rs
+++ b/c2rust-refactor/src/context.rs
@@ -1147,12 +1147,17 @@ pub struct CalleeInfo<'tcx> {
 
 type DefMapping = HashMap<DefId, DefId>;
 
+#[derive(Clone, Copy)]
 pub struct TypeCompare<'a, 'tcx: 'a, 'b> {
     cx: &'a RefactorCtxt<'a, 'tcx>,
 
     /// Mapping from old DefId to new DefId for defs that have been replaced
     /// after types were resolved.
     def_mapping: Option<&'b DefMapping>,
+
+    /// Require array lengths to match exactly, instead of letting a
+    /// zero-length array match an array of any length.
+    exact_array_lens: bool,
 }
 
 impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
@@ -1160,6 +1165,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
         Self {
             cx,
             def_mapping: None,
+            exact_array_lens: false,
         }
     }
 
@@ -1167,6 +1173,21 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
         Self {
             cx,
             def_mapping: Some(def_mapping),
+            exact_array_lens: false,
+        }
+    }
+
+    /// Return a copy of this comparison that requires array lengths to match
+    /// exactly.
+    ///
+    /// The zero-length leniency exists for an `extern` array declaration,
+    /// which C lets omit the length that its definition gives. Nothing else
+    /// makes a zero-length array interchangeable with a longer one: they are
+    /// distinct Rust types, and Rust will not coerce between them.
+    fn with_exact_array_lens(self) -> Self {
+        Self {
+            exact_array_lens: true,
+            ..self
         }
     }
 
@@ -1378,6 +1399,11 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
     /// Compare two function declarations for equivalent argument and return types,
     /// ignoring argument names.
     pub fn compatible_fn_prototypes(&self, decl1: &FnDecl, decl2: &FnDecl) -> bool {
+        // A parameter or return type is passed by value at every call site, so
+        // the two signatures have to agree on it exactly; the zero-length
+        // array leniency only holds for an `extern` array declaration.
+        let strict_cmp = self.with_exact_array_lens();
+
         // `zip` below stops at the shorter parameter list, so the lengths have
         // to be compared separately. Otherwise a declaration is compatible
         // with any other one that merely extends it, which is exactly the
@@ -1389,7 +1415,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
         }
 
         let mut args = decl1.inputs.iter().zip(decl2.inputs.iter());
-        if !args.all(|(arg1, arg2)| self.structural_eq_ast_tys(&arg1.ty, &arg2.ty, true)) {
+        if !args.all(|(arg1, arg2)| strict_cmp.structural_eq_ast_tys(&arg1.ty, &arg2.ty, true)) {
             return false;
         }
 
@@ -1405,12 +1431,16 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
             FnRetTy::Ty(ty) => &ty,
         };
 
-        self.structural_eq_ast_tys(ty1, ty2, true)
+        strict_cmp.structural_eq_ast_tys(ty1, ty2, true)
     }
 
     /// Compare two ty function signatures for equivalent argument and return
     /// types, ignoring argument names.
     pub fn compatible_fn_sigs(&self, sig1: &FnSig<'tcx>, sig2: &FnSig<'tcx>) -> bool {
+        // See `compatible_fn_prototypes` for why a signature compares array
+        // lengths exactly.
+        let strict_cmp = self.with_exact_array_lens();
+
         if sig1.inputs().len() != sig2.inputs().len() {
             return false;
         }
@@ -1420,14 +1450,14 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
         }
 
         for (&arg_ty1, &arg_ty2) in sig1.inputs().iter().zip(sig2.inputs().iter()) {
-            if !self.structural_eq_tys_with_vis(arg_ty1, arg_ty2) {
+            if !strict_cmp.structural_eq_tys_with_vis(arg_ty1, arg_ty2) {
                 return false;
             }
         }
 
         let out_ty1 = sig1.output();
         let out_ty2 = sig2.output();
-        self.structural_eq_tys_with_vis(out_ty1, out_ty2)
+        strict_cmp.structural_eq_tys_with_vis(out_ty1, out_ty2)
     }
 
     /// Compare two AST types for structural equivalence, ignoring names.
@@ -1559,7 +1589,8 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
                 // array types with global array definitions, but it should be
                 // apply in practice as we translate empty extern array lengths
                 // into 0 length extern arrays.
-                if len1 != len2 && len1 != Some(0) && len2 != Some(0) {
+                let lenient = !self.exact_array_lens && (len1 == Some(0) || len2 == Some(0));
+                if len1 != len2 && !lenient {
                     trace!("Array lengths don't match: {:?} and {:?}", n1, n2);
                     return false;
                 }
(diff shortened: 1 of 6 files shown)


c2rust-refactor/src/context.rs
```diff
@@ -1389,7 +1415,7 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
         }
 
         let mut args = decl1.inputs.iter().zip(decl2.inputs.iter());
-        if !args.all(|(arg1, arg2)| self.structural_eq_ast_tys(&arg1.ty, &arg2.ty, true)) {
+        if !args.all(|(arg1, arg2)| this.structural_eq_ast_tys(&arg1.ty, &arg2.ty, true)) {
```
fw-immunant: `this` vs `self` is highly confusing naming. Never use `this` as a variable name in Rust unless referring to some problem domain where "this" is an existing term.

c2rust-refactor/src/context.rs
```diff
@@ -1147,26 +1147,47 @@ pub struct CalleeInfo<'tcx> {
 
 type DefMapping = HashMap<DefId, DefId>;
 
+#[derive(Clone, Copy)]
 pub struct TypeCompare<'a, 'tcx: 'a, 'b> {
```
fw-immunant, line 1151: A doc comment explaining the purpose of this type would be good to have, as the name `TypeCompare` is not entirely clear about how the data stored here relates to comparing types--is it a pass over source code which compares types, guidance for which types to compare, a specification of what to compare about types, the contextual information required to compare types, etc.

In practice it seems like it's the latter, and now being augmented with the penultimate.

c2rust-refactor/src/context.rs
```diff
@@ -1405,12 +1431,16 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
...
+        this.structural_eq_ast_tys(ty1, ty2, true)
     }
 
     /// Compare two ty function signatures for equivalent argument and return
     /// types, ignoring argument names.
     pub fn compatible_fn_sigs(&self, sig1: &FnSig<'tcx>, sig2: &FnSig<'tcx>) -> bool {
+        // See `compatible_fn_prototypes` for why a signature compares array
+        // lengths exactly.
+        let this = self.with_exact_array_lens();
+
```
fw-immunant, line 1443: Again, `this` vs `self` is going to confuse and upset anyone who reads this code.

c2rust-refactor/src/context.rs
```diff
@@ -1559,7 +1589,8 @@ impl<'a, 'tcx, 'b> TypeCompare<'a, 'tcx, 'b> {
                 // array types with global array definitions, but it should be
                 // apply in practice as we translate empty extern array lengths
                 // into 0 length extern arrays.
-                if len1 != len2 && len1 != Some(0) && len2 != Some(0) {
+                let lenient = !self.exact_array_lens && (len1 == Some(0) || len2 == Some(0));
+                if len1 != len2 && !lenient {
```
fw-immunant, line 1593: Is there any way we could *avoid* translating incomplete array types to zero-length arrays, e.g. by translating to a pointer to a DST or such? Our test coverage of what we currently do here seems pretty minimal (e.g. our `c2rust-transpile/tests/snapshots/incomplete_arrays.c` snapshot generates no code), but in practice I don't see how Rust code using elements from a zero-length array (even if its actual definition is foreign) can be sound.

fw-immunant:
The commit message for `compare foreign function declarations by resolved type` is exhaustingly long, but doesn't answer questions like:

- when will the signature fail to resolve, resulting in syntactic comparison?
- what are the specific `E0308`/`E0603` errors referenced?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -2210,6 +2216,36 @@ enum ContainsDecl<'a> {
...
+/// function, and so may be collapsed into one.
+///
+/// The two declarations come from different modules, so comparing their
+/// written-out signatures is not enough: a parameter spelled `*mut stat` in
+/// each can name a different `stat` in each module, and merging on that basis
+/// silently repoints one module's calls at the other module's type. A foreign
+/// item has no body, so its signature carries no node types for
+/// `compatible_fn_prototypes` to compare, and it falls back to comparing the
+/// syntax. Compare the resolved signatures instead, and only fall back to the
+/// syntactic comparison when a signature cannot be resolved.
```
fw-immunant: This comment seems more appropriate for an internal comment than a doc comment.

fw-immunant (changes requested):
I think this is functionally reasonable, modulo various comments. That said, the newly-added test does appear to be failing. I'd also prefer if commit messages generated by an LLM were demarcated as quotation (if included at all) rather than appearing like a message from the committer.

c2rust-refactor/src/context.rs
ahomescu: That seems independent of the current change, `TypeCompare` is a really old type. I'll make a separate PR for that.

c2rust-refactor/src/context.rs
ahomescu: Renamed it to `strict_cmp` which does seem like a better alternative.

c2rust-refactor/src/context.rs
ahomescu: Re this line: I'm not sure why we needed the new derive, so I deleted it.

Update: I found out why we needed it, it's because `with_exact_array_lens` takes a reference but produces a new object.

c2rust-refactor/src/context.rs
ahomescu: We might need to do different things for each of the following cases:
* `extern` array, e.g. `extern int foo[];`. I'm not sure we can use a DST here, since a pointer or reference to it becomes a fat pointer.
* Function argument: `void foo(int arg[]);`. This is an array-to-pointer decay, so replacing it with `int *arg` might be fine.
* Flexible array member at the end of a structure. I think this is the only one that could be a DST, but we would have to detect it.

I'm not sure the transpiler handles these cases differently, I'll have to go check.

ahomescu:
I need to investigate what happened here with all the rebasing, because the test and the fix don't seem to match anymore.

ahomescu:
> when will the signature fail to resolve, resulting in syntactic comparison?

There are now more details in the comment that cover this case.

> what are the specific E0308/E0603 errors referenced?

Is that relevant? Anyway, E0308 is `Expected type did not match the received type.` and E0603 is `A private item was used outside its scope.` but I think the latter is gone now.

c2rust-refactor/src/context.rs
```diff
@@ -1147,26 +1147,47 @@ pub struct CalleeInfo<'tcx> {
...
 
+#[derive(Clone, Copy)]
 pub struct TypeCompare<'a, 'tcx: 'a, 'b> {
     cx: &'a RefactorCtxt<'a, 'tcx>,
 
     /// Mapping from old DefId to new DefId for defs that have been replaced
     /// after types were resolved.
     def_mapping: Option<&'b DefMapping>,
+
+    /// Require array lengths to match exactly, instead of letting a
```
ahomescu, line 1158: This is a fix for a pre-existing array issue that is now exposed by the more accurate signature comparison. I'm starting to think it should be its own separate PR, but in either case it needs to lands first. Thoughts?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -2210,6 +2216,43 @@ enum ContainsDecl<'a> {
...
+    decl2: &FnDecl,
+) -> bool {
+    let tcx = cx.ty_ctxt();
+    let sig_of = |foreign: &ForeignItem| {
+        tcx.fn_sig(cx.node_def_id(foreign.id))
+            .subst_identity()
+            .no_bound_vars()
+    };
+    match (sig_of(foreign1), sig_of(foreign2)) {
+        (Some(sig1), Some(sig2)) => cx.compatible_fn_sigs(&sig1, &sig2),
```
thedataking, line 2253: [P2] Account for actual type deduplication when matching signatures

`compatible_fn_sigs` compares ADTs structurally. Two types with the same fields therefore compare equal even when they are never merged by this transform. For example, the input below compiles, but reorganization at stack head drops the second `fill` declaration and redirects its caller to the first. The output fails with E0308: expected `*mut A`, found `&mut B`. The transform before this signature-comparison change produces compiling output.

Please require the argument/return types to be identical after accounting for the definitions actually being unified, or retain separate foreign declarations when the nominal types remain distinct.

<details>
<summary>Reproducer</summary>

Save as `repro.rs`, run `c2rust-refactor reorganize_definitions --rewrite-mode alongside -- repro.rs --edition=2021`, then compile `repro.new` with `rustc --edition=2021 --crate-name repro --emit=metadata repro.new`.

```rust
#![feature(register_tool)]
#![register_tool(c2rust)]
#![allow(dead_code, non_camel_case_types, unused_imports)]

pub mod defs {
    #[repr(C)]
    pub struct A { pub x: i32 }
    #[repr(C)]
    pub struct B { pub x: i32 }
}
pub mod a {
    #[c2rust::header_src = "/tmp/io.h:1"]
    pub mod io_h {
        use crate::defs::A;
        extern "C" { pub fn fill(a: *mut A); }
    }
    pub unsafe fn run() { io_h::fill(&mut crate::defs::A { x: 0 }); }
}
pub mod b {
    #[c2rust::header_src = "/tmp/io.h:1"]
    pub mod io_h {
        use crate::defs::B;
        extern "C" { pub fn fill(b: *mut B); }
    }
    pub unsafe fn run() { io_h::fill(&mut crate::defs::B { x: 0 }); }
}
fn main() {}
```
</details>

thedataking (commented):
Review of stack 1953: the resolved-signature comparison can merge foreign declarations whose nominal argument types remain distinct.

PR #1945, opened 2026-07-25
refactor: keep the original symbol when renaming an extern declaration


ahomescu:
Fresh copy of #1935 since I closed that one.

---

### DIFF
--- a/c2rust-refactor/src/transform/reorganize_definitions.rs
+++ b/c2rust-refactor/src/transform/reorganize_definitions.rs
@@ -18,7 +18,7 @@ use rustc_hir::{self as hir, Node};
 use rustc_middle::metadata::ModChild;
 use rustc_middle::ty::{self, ParamEnv};
 use rustc_span::symbol::{kw, Ident};
-use rustc_span::{BytePos, Symbol, DUMMY_SP};
+use rustc_span::{sym, BytePos, Symbol, DUMMY_SP};
 use rustc_target::spec::abi::{self, Abi};
 use smallvec::smallvec;
 use thin_vec::ThinVec;
@@ -752,7 +752,7 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
                     // multiple items with the same name in a namespace.
                     let old_name = ident.name.as_str();
                     let new_name = format!("{old_name}_{idx}");
-                    warn!("Renaming identifier {old_name} to {new_name} due to collision");
+                    item.preserve_link_name(ident);
                     item.ident_mut().name = Symbol::intern(&new_name);
                 }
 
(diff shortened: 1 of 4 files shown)


c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1594,6 +1595,30 @@ impl MovedDecl {
         }
     }
 
+    /// Pin down the symbol an `extern` declaration links against, before its
+    /// ident is changed to resolve a collision. A foreign function or static
+    /// links against its own name unless a `#[link_name]` says otherwise, so
+    /// renaming one silently repoints it at a symbol that does not exist.
```
thedataking: I'd drop this whole comment. The function name conveys the same information as the first sentence. 
The second sentence explains the Rust language which is not what we normally use comments for here. 

The most recent changes to this file has added these overly verbose comments. Claude spews them unless you ask it to only comment on non-obvious/surprising aspects of the code. Same for commit messages, and PR bodies. We should rein it in.

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1594,6 +1595,30 @@ impl MovedDecl {
...
+        // An `extern type` has no symbol of its own to preserve.
+        if !matches!(
+            item.kind,
+            ForeignItemKind::Fn(..) | ForeignItemKind::Static(..)
+        ) {
+            return;
+        }
+        if item.attrs.iter().any(|attr| attr.has_name(sym::link_name)) {
+            return;
+        }
```
thedataking, line 1619: Should we warn when we're unable to put a `link_name` attribute on the item?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -2261,6 +2286,26 @@ fn is_nested(tree: &UseTree) -> bool {
     }
 }
 
+/// Build a `#[link_name = "<symbol>"]` attribute.
```
thedataking: This duplicates the string-attribute construction already implemented by `Builder::str_attr`. Could we reuse that API here, or expose a singular `str_attribute` constructor from the AST builder, rather than adding a transform-local AST constructor?

thedataking (changes requested):

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -753,6 +753,7 @@ impl<'a, 'tcx> Reorganizer<'a, 'tcx> {
                     let old_name = ident.name.as_str();
                     let new_name = format!("{old_name}_{idx}");
                     warn!("Renaming identifier {old_name} to {new_name} due to collision");
```
thedataking: nit: should we move this warning into `preserve_link_name` so you either emit

` warn!("Renaming identifier {old_name} to {new_name} due to collision");`
or
```
 warn!(
                "Item {} already has a `#[link_name]` attribute, keeping the existing one",
                orig,
            );
```

but not both? If you do this, we could line up the warnings like this:

- "`#[link_name]` attribute on item {} added to preserve link name"
- "`#[link_name]` attribute on item {} already exists; cannot preserve link name"

thedataking (approved):

PR #1947, opened 2026-07-25
refactor: reorganize_definitions: keep glob imports untouched


ahomescu:
Fresh copy of #1927

---

### DIFF
--- a/c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions-reorganize_implicit_extern.rs.snap
+++ b/c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions-reorganize_implicit_extern.rs.snap
@@ -11,7 +11,7 @@ expression: c2rust-refactor reorganize_definitions --rewrite-mode alongside -- t
 pub mod first {
 
     // =============== BEGIN first_h ================
-    extern "C" {
+    extern "Rust" {
 
         pub fn compute(x: i32) -> i32;
     }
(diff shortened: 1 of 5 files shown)


c2rust-refactor/tests/snapshots/snapshots__refactor-reorganize_definitions-reorganize_implicit_extern.rs.snap
```diff
@@ -11,7 +11,7 @@ expression: c2rust-refactor reorganize_definitions --rewrite-mode alongside -- t
 pub mod first {
 
     // =============== BEGIN first_h ================
-    extern "C" {
+    extern "Rust" {
```
thedataking, line 14: [P1] Keep the corrected implicit-extern snapshot

Commit `25efe877c` restores the pre-fix expectation here, even though #1924 already changed implicit extern blocks to use the C ABI. At stack head, the transform emits one `extern "C"` declaration in `first` and redirects the second caller to it; this snapshot instead expects `extern "Rust"` plus a renamed `compute_1`.

I ran `cargo test --offline -p c2rust-refactor --test snapshots -- --test-threads=1`: 54 passed and only `test_reorganize_implicit_extern` failed on this mismatch. Please drop the snapshot-reverting commit or restore the expectation from #1924.

thedataking (commented):
Review of stack 1953: the implicit-extern snapshot is reverted by this PR and fails with the corrected ABI behavior.

PR #1956, opened 2026-07-28
transpile: Use a single global `Renamer` with namespacing

Rua:
- Depends on #1957 

Some names in Rust exist in multiple namespaces at once. Tuple structs for example create a name in both the types and the values namespace. Thus, a tuple struct named `Foo` will conflict with a type alias `Foo` but also with a function named `Foo`, which is a problem for #1620 because it uses tuple structs.

The current `Renamer` setup, with entirely different instances for the types and values namespaces, doesn't handle that situation. This PR adds a `Namespaces` mask to `Scope`, and then makes everything else specify the correct mask when querying/inserting names into `Renamer`. Finally, a single `Renamer` instance is now shared between `Translator` and `TypeConverter`, so that names inserted into one will conflict with those in the other, if they share a namespace.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -1671,8 +1671,7 @@ impl<'c> Translation<'c> {
             type_converter: RefCell::new(type_converter),
             ast_context,
             tcfg,
-            // TODO: Use Renamer::value_namespace() for most renamings.
-            renamer: RefCell::new(Renamer::global_value_namespace()),
+            renamer: RefCell::new(Renamer::keywords_and_prelude()),
             zero_inits: RefCell::new(IndexMap::new()),
             function_context: RefCell::new(FuncContext::new()),
             potential_flexible_array_members: RefCell::new(IndexSet::new()),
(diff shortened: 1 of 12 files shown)


c2rust-transpile/tests/snapshots/renamer.c
```diff
@@ -28,3 +28,8 @@ void impl(type in) {
     mut.where = 0;
     pub(mut.where);
 }
+
+void local_type(void) {
```
ahomescu, line 1: This needs a rebase.

c2rust-transpile/src/translator/mod.rs
```diff
@@ -885,33 +888,27 @@ pub fn translate(
...
-
         // Used for testing; so that we don't overlap with C function names
         if let Some(ref prefix) = t.tcfg.prefix_function_names {
             prefix_names(&mut t, prefix);
         }
 
+        fn ns_for_decl(decl_kind: &CDeclKind) -> Namespaces {
+            use CDeclKind::*;
+            match decl_kind {
+                Struct { .. } | Union { .. } | Enum { .. } | Typedef { .. } => Namespaces::types(),
```
ahomescu, line 899: Should this match #1958 and use `types() | values()` for the constructors? Although I'd prefer keeping the current PR free of semantic changes, and extending the namespaces here in a follow-up.

ahomescu (changes requested):

c2rust-transpile/src/translator/mod.rs
Rua: The code here refers to C types, and c2rust does not translate any of those with tuple structs right now. `Enum` will be changed here in the tuple struct PR, but the rest should stay.

ahomescu (approved):

PR #1957, opened 2026-07-28
transpile: Fix renaming bug with locally declared types

Rua:
There was a bug in the handling of local decls, where type decls would be inserted into the renamer at local scope, even though they are emitted at global scope by c2rust. This would then trigger renaming of subsequent variables with the same name.

### DIFF
--- a/c2rust-transpile/tests/snapshots/renamer.c
+++ b/c2rust-transpile/tests/snapshots/renamer.c
@@ -1,0 +1,4 @@
+void local_type(void) {
+    struct foo {};
+    int foo;
+}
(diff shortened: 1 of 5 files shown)


c2rust-transpile/tests/snapshots/renamer.c
```diff
@@ -28,3 +28,8 @@ void impl(type in) {
     mut.where = 0;
     pub(mut.where);
 }
+
+void local_type(void) {
```
ahomescu, line 1: The rest of this file does test raw keywords, like the file name says. Maybe add a new `renamed.c` file and leave the old `raw_keywords.c` alone?

ahomescu (changes requested):

c2rust-transpile/tests/snapshots/renamer.c
Rua: I put them together because there's already a lot of snapshot tests, and it's getting a bit messy. If you prefer to have more separate tests I can do that.

ahomescu (approved):

PR #1959, opened 2026-07-29
Update nix flake

Rua:
Following #1913.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,7 +11,7 @@
   };
 
   outputs =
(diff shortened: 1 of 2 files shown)


flake.nix
thedataking, line 64: could this be removed too?

thedataking (approved):
Thanks!

flake.nix
Rua: Nothing is highlighted in the snippet, what do you want removed?

PR #1963, opened 2026-07-31
ast-exporter: pass compilation database directory to Clang

### LINKED ISSUE #1809: json-compilation-database: Error while opening JSON database: Not a directory
When I pass a compilation database to c2rust (running LLVM 22.1.5 on Linux), the following message is repeated for every input file:
```
Error while trying to load a compilation database:
Could not auto-detect compilation database from directory "/projectdir/compile_commands.json"
No compilation database found in /projectdir/compile_commands.json or any parent directory
json-compilation-database: Error while opening JSON database: Not a directory
fixed-compilation-database: Error while opening fixed database: Not a directory
Running without flags.
```
And indeed, none of the flags in the compilation database are used, causing errors.

Debugging a bit, it seems the way c2rust passes the build file to libtooling (`-p <path-to-json-file>`) is incorrect, at least for LLVM 22. It expects the build path passed in with -p to be a directory containing compile_commands.json.

`CompilationDatabase::autoDetectFromDirectory` seems to be the relevant function here.

ShiroKSH:
Fixes #1809.

`CommonOptionsParser` expects `-p` to name the build directory, but the AST exporter passed `compile_commands.json` itself. Pass its parent directory instead, using `.` for a bare relative path, and cover both forms.

This is a general correctness fix; #1809 exposed the existing mismatch on Clang 22.

### DIFF
--- a/c2rust-ast-exporter/src/lib.rs
+++ b/c2rust-ast-exporter/src/lib.rs
@@ -58,6 +58,13 @@ pub fn get_untyped_ast(
 /// libClangTooling is not thread-safe, so we must not allow concurrent calls to `ast_exporter`.
 static CLANG_MUTEX: Mutex<()> = Mutex::new(());
 
+fn clang_build_path(cc_db: &Path) -> &Path {
+    cc_db
+        .parent()
+        .filter(|path| !path.as_os_str().is_empty())
+        .unwrap_or_else(|| Path::new("."))
+}
+
 fn get_ast_cbors(
     file_path: &Path,
     cc_db: &Path,
@@ -68,7 +75,7 @@ fn get_ast_cbors(
     let mut args_owned = vec![CString::new("ast_exporter").unwrap()];
     args_owned.push(CString::new(file_path.to_str().unwrap()).unwrap());
     args_owned.push(CString::new("-p").unwrap());
-    args_owned.push(CString::new(cc_db.to_str().unwrap()).unwrap());
+    args_owned.push(CString::new(clang_build_path(cc_db).to_str().unwrap()).unwrap());
 
     for &arg in extra_args {
         args_owned.push(CString::new(["-extra-arg=", arg].join("")).unwrap())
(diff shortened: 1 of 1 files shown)


thedataking (commented):
> CommonOptionsParser treats -p as a build directory and searches that directory for the compilation database. The AST exporter passed the JSON file itself, so newer Clang versions attempted to treat the file as a directory and continued without the database flags.

https://github.com/llvm/llvm-project/commits/main/clang/lib/Tooling/CommonOptionsParser.cpp has not changed its functionality since 2022. I also don't see any recent changes to https://github.com/llvm/llvm-project/commits/main/clang/lib/Tooling/CompilationDatabase.cpp that would make this a clang 22 specific issue.

I do agree that we ought to pass the parent directory of the compilation database, so that part looks good to me. However, I think we ought to clarify whether this is a general correctness fix or what versions of clang need this.

**nitpick**:  C2Rust validation means passing CI so we do not need a validation section. Generally, PRs bodies are shorter and not split into sections. I'd encourage you to take a look at other PRs and make yours follow project conventions.

ShiroKSH:
Thanks — I shortened the PR body to match the project convention and clarified that this is a general correctness fix; #1809 only exposed the existing mismatch on Clang 22. I also rebased the branch onto the current `master`.

c2rust-ast-exporter/src/lib.rs
```diff
@@ -151,3 +158,24 @@ unsafe fn marshal_result(
     }
     output
 }
+
+#[cfg(test)]
```
thedataking: I don't think we should be adding these tests. We mainly rely on the integration tests to detect cases where `compile_commands.json` is not being picked up correctly. It should break most (all) the builds.

IIRC, the macOS builds use clang-22 and linux builds use 15 an 18 so we cover a broad range of releases.

ShiroKSH: Removed the two local unit tests in `102b4f5cf`, leaving this behavior to the integration matrix as requested. `cargo fmt --all -- --check` passes.

thedataking (approved):
Thanks!

PR #1967, opened 2026-08-05
Fix span collisions from PartialEq derive

### LINKED ISSUE #1818: Refactorer crashes if the processed code has `derive(PartialEq, Eq)`

ahomescu:
Fixes #1818.

### DIFF
--- a/c2rust-refactor/tests/snapshots.rs
+++ b/c2rust-refactor/tests/snapshots.rs
@@ -429,6 +429,21 @@ fn test_rename_unnamed() {
     refactor("rename_unnamed").test();
 }
 
+/// Currently panics (#1818): the `PartialEq` derive expansion stamps the span of the
+/// `PartialEq` token on every node it generates, so the NodeId -> HirId lookup pairs the
+/// `&C2Rust_Unnamed` parameter type of the derived `fn eq` with the HIR node for `bool`:
+///
+/// ```shell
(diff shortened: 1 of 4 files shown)


c2rust-refactor/tests/snapshots.rs
```diff
@@ -429,6 +429,17 @@ fn test_rename_unnamed() {
     refactor("rename_unnamed").test();
 }
 
+/// Regression test for #1818: the `PartialEq` derive expansion stamps the span of the
+/// `PartialEq` token on every node it generates, which used to make the NodeId -> HirId
+/// lookup pair the `&C2Rust_Unnamed` parameter type of the derived `fn eq` with the HIR
+/// node for `bool` and panic in `alter_ty_path`.
```
fw-immunant:
```suggestion
```
This file only has caveats/TODOs for individual test cases where needed and this comment is also present inside the snapshot input file itself.

fw-immunant (approved):
The fix makes sense to me. I'd prefer if we removed the extra copy of the explanatory comment being added to `snapshots.rs`.

c2rust-refactor/tests/snapshots.rs
ahomescu: If I had to pick between leaving the comments in `snapshots.rs` vs the actual files, I'd pick the former because all test in the test cases gets at least triplicated in all the snapshots. I agree though that we shouldn't show up in two source places, and we don't seem to have any others in `snapshots.rs` so I'll delete in.

PR #1970, opened 2026-08-08
transpile: Add `MacroInvocationInfo`

Rua:
Following up from #1792, this adds the necessary infrastructure to the AST to handle functional macro arguments. The `MacroInvocationInfo::arguments` value is currently unused.

### DIFF
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -51,6 +51,7 @@ mod node_types {
 
 type ClangId = u64;
 type ImporterId = u64;
+type MacroInvocationId = u32;
 
 /// Correspondence between old/new IDs.
 ///
@@ -244,6 +245,12 @@ pub struct ConversionContext {
     /// Typed context we are building up during the conversion
     typed_context: TypedAstContext,
 
+    /// Maps expressions to the stack of macro invocations that expanded to them.
+    macro_invocations: IndexMap<CExprId, Vec<MacroInvocationId>>,
+
+    /// Maps macro invocation IDs to information about the invocation (macro ID, arguments, etc).
+    macro_infos: HashMap<MacroInvocationId, MacroInvocationInfo>,
+
     pub invalid_clang_ast: bool,
 }
 
@@ -329,6 +336,8 @@ impl ConversionContext {
             processed_nodes: HashMap::new(),
             visit_as,
             typed_context: TypedAstContext::new(input_path, &untyped_context.files),
+            macro_invocations: IndexMap::new(),
+            macro_infos: HashMap::new(),
             invalid_clang_ast,
         };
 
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -1047,13 +1068,38 @@ impl ConversionContext {
...
+                        });
+                    assert_eq!(macro_id, info.macro_id);
+
+                    // Fill in the values that were provided in the `MacroInvocationInfoRaw`.
+                    if let Some(parameter) = parameter {
+                        // This expression is an argument for a functional macro.
+                        info.arguments.insert(parameter.clone(), expr_id);
+                    } else {
+                        // This expression is the result of a macro expansion.
+                        info.expr_id = expr_id;
```
ahomescu, line 1104: Could this already be a different value that's not `u64::MAX` or `expr_id`? We could check.

ahomescu: `expr_id = CExprId(new_id)` so is it guaranteed to always be unique? In that case we don't need to check.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -554,31 +563,43 @@ impl ConversionContext {
...
+                    self.typed_context.unwrap_implicit_cast_expr(expr_id)
+                };
+
+                // For each invocation ID, retrieve its corresponding `MacroInvocationInfo`
+                // and wrap it in an `Rc`.
+                let infos = invocation_ids
+                    .into_iter()
+                    .map(|key| {
+                        let info = macro_infos
+                            .remove(&key)
```
ahomescu, line 584: Can't a key be needed multiple times here?

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -554,31 +563,43 @@ impl ConversionContext {
...
+                            .expect("No macro_infos entry for key {key}");
```
ahomescu: `expect` takes a fixed string not a format, so `{key}` here is printed as is.

c2rust-transpile/src/translator/macros.rs
```diff
@@ -67,20 +64,22 @@ impl<'c> Translation<'c> {
...
+        macro_id: CDeclId,
     ) -> TranslationResult<(Box<Expr>, CTypeId)> {
-        let (val, ty) = expansions
+        let ctx = ctx.const_().set_expanding_macro(macro_id);
+        let (val, ty) = self.ast_context.macro_expansions[&macro_id]
             .iter()
-            .try_fold::<Option<(WithStmts<Box<Expr>>, CTypeId)>, _, _>(None, |canonical, &id| {
-                self.can_convert_const_macro_expansion(id)?;
+            .try_fold::<Option<(WithStmts<Box<Expr>>, CTypeId)>, _, _>(None, |canonical, info| {
+                let &MacroInvocationInfo { expr_id, .. } = info.as_ref();
```
ahomescu, line 73: `let expr_id = info.expr_id;` would be simpler (and you wouldn't need the new import).

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -1047,13 +1068,38 @@ impl ConversionContext {
             };
 
             if expected_ty & EXPR != 0 {
```
ahomescu, line 1070: What about the callers of `expr_possibly_as_stmt`? I.e. expressions used as statements?

ahomescu (changes requested):

c2rust-transpile/src/c_ast/conversion.rs
Rua: A key refers to a single macro invocation, and each invocation should end up in `macro_invocations` exactly once.

c2rust-transpile/src/translator/macros.rs
Rua: I expect to add more fields here in a future PR.

c2rust-transpile/src/c_ast/conversion.rs
Rua: Fixed.

c2rust-transpile/src/c_ast/conversion.rs
Rua: Those get wrapped in a `CStmtKind::Expr`, in which the statement part and the expression part are handled separately.

c2rust-transpile/src/c_ast/conversion.rs
Rua: The only way it could be not unique is if the macro invocation key is reused for multiple invocations. I've added an `assert` for that now.

c2rust-transpile/src/c_ast/conversion.rs
```diff
@@ -554,31 +563,43 @@ impl ConversionContext {
...
+
+                // For each invocation ID, retrieve its corresponding `MacroInvocationInfo`
+                // and wrap it in an `Rc`.
+                let infos = invocation_ids
+                    .into_iter()
+                    .map(|key| {
+                        let info = macro_infos
+                            .remove(&key)
+                            .unwrap_or_else(|| panic!("No macro_infos entry for key {key}"));
+                        let info = Rc::new(MacroInvocationInfo { expr_id, ..info });
```
ahomescu, line 586: Do we want to also normalize `info.arguments` here (through `unwrap_implicit_cast_expr` or something else)?

ahomescu (approved):

c2rust-transpile/src/c_ast/conversion.rs
Rua: Possibly, but I'm not sure. It kinda works "in reverse" when it comes to macro arguments, because they are placed within the macro definition. So if casts are left as-is, then they are considered to form part of the argument passed at the call site. If they are unwrapped, then they are considered part of the context surrounding the argument within the macro definition.

PR #1971, opened 2026-08-08
transpile: Refactor code related to `MacroExpansion`

Rua:
This prepares the code for adding a hashmap with macro arguments to both `ConvertedMacro` and `ConvertedMacroExpr`, and also generally looks a little cleaner?

### DIFF
--- a/c2rust-transpile/src/translator/macros.rs
+++ b/c2rust-transpile/src/translator/macros.rs
@@ -30,14 +30,13 @@ impl<'c> Translation<'c> {
         );
 
         match maybe_replacement {
-            Ok((replacement, ty)) => {
+            Ok((replacement, converted)) => {
                 trace!("  to {:?}", replacement);
 
-                let converted = ConvertedMacro { ty };
+                let ty = self.convert_type(converted.ty)?;
                 self.converted_macros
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/macros.rs
```diff
@@ -30,14 +31,13 @@ impl<'c> Translation<'c> {
         );
 
         match maybe_replacement {
-            Ok((replacement, ty)) => {
+            Ok((replacement, converted)) => {
                 trace!("  to {:?}", replacement);
 
-                let expansion = MacroExpansion { ty };
-                self.macro_expansions
+                let ty = self.convert_type(converted.ty)?;
```
ahomescu: If `convert_ty` returns an error, we could intercept it and handle it the same way as the error case on line 48 below. Or call `convert_ty` from an `.and_then` on `maybe_replacement` above.

ahomescu (changes requested):

c2rust-transpile/src/translator/macros.rs
Rua: Done.

ahomescu (approved):

PR #1973, opened 2026-08-12
transpile: Unset `needs_address` in various places

Rua:
`needs_address` is being set, but it is then passed down to all subexpressions, even those whose address is never actually taken.

### DIFF
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -201,7 +201,7 @@ impl<'c> Translation<'c> {
             return self.convert_expr(ctx.used(), arg, None);
         }
 
-        self.convert_expr(ctx.used(), arg, None)?
+        self.convert_expr(ctx.used().set_needs_address(false), arg, None)?
             .try_map(|val: Box<Expr>| {
                 if let CTypeKind::Function(..) =
                     self.ast_context.resolve_type(cqual_type.ctype).kind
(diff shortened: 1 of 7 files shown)


ahomescu:
I think `is_pattern` from 1776 has the same problem. We might need a larger refactoring to make all these context flags handle recursion correctly. It's weird though, all the mutators on `ExprContext` return `Self` so they should make a copy of the context and not modify the original, so what's going on here?

Rua:
> I think `is_pattern` from 1776 has the same problem. We might need a larger refactoring to make all these context flags handle recursion correctly.

I have already been looking into the overall issue. Some context flags should be "inherited", others should reset back to a default whenever they are passed to a subexpression.

> It's weird though, all the mutators on `ExprContext` return `Self` so they should make a copy of the context and not modify the original, so what's going on here?

The problem isn't the copying, it's that the modified context of a parent expression gets passed to all the subexpressions too. So in this case, if `needs_address` is set when calling `convert_expr` or such, if the expression consists of further subexpressions, then `needs_address` will still be set when translating those subexpressions as well. You see this in the tests I've added: if the address is taken of an array-subscript expression, and either the array or the subscript are themselves expression types that are sensitive to the `needs_address` flag, then those expressions are also translated as if their address is taken, when they shouldn't be.

c2rust-transpile/src/translator/pointers.rs
ahomescu, line 201: Do we want `set_needs_address(false)` for this one?

There might be a regression here for this test case:
```
int *deref_addrof_compound(void) {
    return *&(int[]){1, 2, 3};
}
```
Can you reproduce it?

ahomescu (changes requested):

ahomescu:
> The problem isn't the copying, it's that the modified context of a parent expression gets passed to all the subexpressions too. 

Ah I understand now, it's in the opposite direction of what I thought.

c2rust-transpile/src/translator/pointers.rs
Rua: I added a regression test for it and fixed the issue.

ahomescu (approved):

PR #1974, opened 2026-08-13
Fix per-field alignment attribute being dropped during transpilation

user3polarjet:
[...] my intent was to try to transpile [x264 encoder](https://github.com/x264project/x264) to rust. It extensively uses alignment attributes for msvc, clang, gcc. And i hit this bug. It kinda pissed me off, so i tried to fix it [with an LLM, generating this change].

LLM-generated PR description follows.

## Summary
- c2rust preserved `__attribute__((aligned(N)))` on whole structs but silently dropped it when applied to an individual field, since the alignment was never captured by the AST exporter, threaded through `CDeclKind::Field`, or used when laying out the generated struct.
- The AST exporter now encodes a field's manual alignment (mirroring what it already does for whole records).
- The transpiler forces the generated struct's `repr` to `align(M)` (M = max manual alignment among its fields) and inserts the padding needed to reach the field's real byte offset, matching clang's layout without inflating the field's own size (an initial `#[repr(align(N))]`-wrapper-type approach was tried and rejected after direct testing showed it inflates the field's size to a multiple of N, which clang's per-field attribute does not).
- `__alignof__`/`_Alignof` applied directly to such a field is also fixed to report the manual alignment instead of the field's natural type alignment.

## Test plan
- [x] Added `FieldAlignStruct` / `test_field_alignment` to `tests/unit/structs`, covering alignment, offset, size, and read/write through the field.
- [x] Verified generated layout (alignment/offset/size) against real `clang`-compiled output directly (not just the Rust side) for a `__attribute__((aligned(16)))` field.
- [x] Confirmed the rest of `tests/unit/structs/src/structs.c` (existing struct/bitfield/packing/alignment coverage) still compiles unaffected.

### DIFF
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3225,6 +3225,31 @@ impl<'c> Translation<'c> {
             ConvertVector(..) => Err(TranslationError::generic("convert vector not supported")),
 
             UnaryType(result_type_id, kind, opt_expr, arg_ty) => {
+                // `__alignof__`/`_Alignof` applied directly to a struct
+                // field (e.g. `__alignof__(s->f)`) should reflect that
+                // field's own manual `__attribute__((aligned(N)))`, if any,
+                // rather than its type's natural alignment.
+                if matches!(kind, CUnTypeOp::AlignOf | CUnTypeOp::PreferredAlignOf) {
+                    if let Some(field_expr) = opt_expr {
+                        if let CExprKind::Member(_, _, decl_id, _, _) =
+                            self.ast_context.index_unwrap_parens(field_expr).kind
+                        {
+                            if let CDeclKind::Field {
+                                manual_alignment: Some(alignment),
+                                ..
+                            } = self.ast_context[decl_id].kind
+                            {
+                                let ty = self.convert_type(result_type_id.ctype)?;
+                                let val = mk().cast_expr(
+                                    mk().lit_expr(mk().int_unsuffixed_lit(alignment)),
+                                    ty,
+                                );
+                                return Ok(WithStmts::new_val(val));
+                            }
+                        }
+                    }
+                }
+
                 let result = match kind {
                     CUnTypeOp::SizeOf => match opt_expr {
                         None => self.compute_size_of_type(
--- a/c2rust-ast-exporter/src/AstExporter.cpp
+++ b/c2rust-ast-exporter/src/AstExporter.cpp
@@ -2569,6 +2569,15 @@ class TranslateASTVisitor final
                          // 4. Encode the type's full bit width (even if a
                          // bitfield)
                          cbor_encode_uint(array, bitWidth);
+
+                         // 5. Encode manually specified alignment
+                         // (e.g. `__attribute__((aligned(N)))` on the field)
+                         auto align = D->getMaxAlignment();
+                         if (align == 0) {
+                             cbor_encode_null(array);
+                         } else {
+                             cbor_encode_uint(array, align / 8);
+                         }
                      });
 
         // This might be the only occurrence of this type in the translation unit
--- a/c2rust-transpile/src/c_ast/conversion.rs
+++ b/c2rust-transpile/src/c_ast/conversion.rs
@@ -2457,12 +2457,15 @@ impl ConversionContext {
                         from_value(node.extras[2].clone()).expect("Did not find field bit offset");
                     let platform_type_bitwidth =
                         from_value(node.extras[3].clone()).expect("Did not find field bitwidth");
+                    let manual_alignment =
+                        expect_opt_u64(&node.extras[4]).expect("Expected field alignment");
                     let field = CDeclKind::Field {
                         name,
                         typ,
                         bitfield_width,
                         platform_bit_offset,
                         platform_type_bitwidth,
+                        manual_alignment,
                     };
                     self.add_decl(new_id, located(node, field));
                     self.processed_nodes.insert(new_id, FIELD_DECL);
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -1708,6 +1708,9 @@ pub enum CDeclKind {
         bitfield_width: Option<u64>,
         platform_bit_offset: u64,
         platform_type_bitwidth: u64,
+        /// Manually specified alignment in bytes, e.g. from
+        /// `__attribute__((aligned(N)))` on the field itself.
+        manual_alignment: Option<u64>,
     },
 
     MacroObject {
--- a/c2rust-transpile/src/translator/structs_unions.rs
+++ b/c2rust-transpile/src/translator/structs_unions.rs
@@ -101,6 +101,28 @@ impl<'a> Translation<'a> {
             _ => {}
         }
 
+        // A field with its own `__attribute__((aligned(N)))` forces clang to
+        // raise the *whole record's* alignment to (at least) N, since every
+        // instance of the struct must be aligned enough for that field.
+        // Rust's automatic `repr(C)` layout only derives struct alignment
+        // from each field's natural type alignment, so make the requirement
+        // explicit; combined with `repr(C)`, `align(N)` sets a *minimum*
+        // alignment without disturbing fields that already need more.
+        // (The field's own byte offset is corrected separately, via explicit
+        // padding inserted in `get_field_types`.)
+        let field_manual_alignment = fields
+            .iter()
+            .filter_map(|field_id| match self.ast_context.index(*field_id).kind {
+                CDeclKind::Field {
+                    manual_alignment, ..
+                } => manual_alignment,
+                _ => None,
+            })
+            .max();
+        if let Some(alignment) = field_manual_alignment {
+            reprs.push(mk().meta_list("align", vec![alignment]));
+        }
+
         if let Some(alignment) = manual_alignment {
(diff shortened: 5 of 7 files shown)


c2rust-transpile/src/translator/mod.rs
```diff
@@ -3226,6 +3226,31 @@ impl<'c> Translation<'c> {
             ConvertVector(..) => Err(TranslationError::generic("convert vector not supported")),
 
             UnaryType(result_type_id, kind, opt_expr, arg_ty) => {
+                // `__alignof__`/`_Alignof` applied directly to a struct
```
Rua, line 3228: Should this new code be added to the `compute_align_of_type` function instead?

user3polarjet: Not directly, because `compute_align_of_type` only receives the operand's `CTypeId`. The field-specific alignment belongs to the field declaration, not its type. For example, the type is
  still `int` with its natural alignment, while this particular field may require alignment 16.

  At this point in `UnaryType`, we still have `opt_expr` and can resolve the `Member` expression to its `CDeclKind::Field`.

  I could refactor this into a separate helper such as `compute_align_of_expr`, or pass the field alignment into `compute_align_of_type`, if you would prefer the alignment handling to be
  grouped there.

Rua: Yeah, passing `opt_expr` to `compute_align_of_type` seems like the cleanest option.

user3polarjet: Done in d2bdafd29 — passed opt_expr through to compute_align_of_type and moved the field-alignment check there. Verified output is unchanged.

fw-immunant: NAK on this change--after d2bdafd, `compute_align_of_type` does not compute the align of a \*type\* when a `Some` `opt_expr` is passed; in that case, it computes the alignment of a struct field, completely independently from the passed `type_id`. While this is the information some contexts need, it makes much more sense for those contexts to do something not called `compute_align_of_type` in those cases. The correspondence between name and actual behavior is important to preserve.

fw-immunant:
The commits here claim to be authored by "Your Name" from an account with activity set to private, and the interactions from the PR author here read like unedited LLM output without a spark of human involvement. While we do want this bug fixed, this is fairly disrespectful of the time spent by human collaborators here and it doesn't make me happy to see. I encourage the human responsible for the @user3polarjet account to drop by and say hello in their own words.

c2rust-transpile/src/translator/mod.rs
user3polarjet: First of i am sorry. Your comments on ai is fair and change of account name too. Wont happen again. Should i create a new pull request with real username and email set?

> The correspondence between name and actual behavior is important to preserve.

So you want me to revert to the first solution?

fw-immunant: Thanks for the comment; I appreciate the apology and am glad to welcome you into the community of C2Rust contributors. If possible I would prefer keeping the changes in this PR just to minimize noise, but please change the commits to bear a name (legal name not required) and e-mail associated with you, and explain your intent (even something as simple as "I am transpiling program X, which uses aligned struct fields, and decided to use LLM Y to try to implement support for these in C2Rust, generating this change" would do) in the PR description.

I would prefer reverting/removing the second commit and leaving only the first, yes. It would also be fine to replace the second commit with one introducing a separate `compute_align_of_field` function, but I don't think that's necessary unless @Rua has a strong opinion in favor of it.

Rua: No, removing it seems fine, now that you mentioned that `compute_align_of_type` is doing something different.

user3polarjet: Ok, i reverted the commit, then set my nickname which prefer and email neometastar@gmail.com. Also rebased commits to change nickname and mail.

Well, my intent was to try to transpile [x264 encoder](https://github.com/x264project/x264) to rust. It extensively uses alignment attributes for msvc, clang, gcc. And i hit this bug. It kinda pissed me off, so i tried to fix it. Next i fixed it, and it successfully compiled. But unfortunately it is slower than C. See logs below ( 80 fps C vs 64 fps rust). Both versions are compiled without manually-written assembly functions that come with the x264. x86 64 laptop i have.

But this is not because `rust is slower than c ha-ha`. But due to the fact how rust treats raw mutable pointers, it has to assume the worst case scenario where both are aliased, so .i guess, rustc does not generate vector instructions and we lose productivity.

So i looked into how c2rust manages converting raw mutable pointer to references both non-mutable and mutable. And then i discovered that c2rust actually has some analysis tool for that with ownership analysis. And there i saw that it uses bfs and brute-force search to predict correct semantics ( this judgement i did at a glance, so i can be wrong). This is where i stopped for now. Maybe later i could create a pull request to propose solutions for better analysis if have spare time. 

So yeah that is the story.

```
user@dist:~/work/x264_work/x264$ ./bench_c 
x264 [error]: invalid tune 'none'
x264 [info]: using cpu capabilities: none!
x264 [info]: profile High, level 4.0, 4:2:0, 8-bit
Encoded 300 frames in 3.718 seconds (80.68 FPS)
x264 [info]: frame I:2     Avg QP:16.50  size:  2132
x264 [info]: frame P:76    Avg QP:18.69  size:   124
x264 [info]: frame B:222   Avg QP:27.98  size:    81
x264 [info]: consecutive B-frames:  1.3%  0.0%  0.0% 98.7%
x264 [info]: mb I  I16..4:  0.0% 99.4%  0.6%
x264 [info]: mb P  I16..4:  0.0%  0.0%  0.0%  P16..4:  0.0%  0.0%  0.0%  0.0%  0.0%    skip
:100
.0%
x264 [info]: mb B  I16..4:  0.0%  0.0%  0.0%  B16..8:  0.1%  0.0%  0.0%  direct: 0.0%  skip
:99.
9%  L0:54.4% L1:45.6% BI: 0.0%
x264 [info]: 8x8 transform intra:99.2% inter:0.0%
x264 [info]: coded y,uvDC,uvAC intra: 0.4% 0.0% 0.0% inter: 0.0% 0.0% 0.0%
x264 [info]: i16 v,h,dc,p:  0%  0% 100%  0%
x264 [info]: i8 v,h,dc,ddl,ddr,vr,hd,vl,hu: 48%  0% 52%  0%  0%  0%  0%  0%  0%
x264 [info]: i4 v,h,dc,ddl,ddr,vr,hd,vl,hu: 30% 15% 55%  0%  0%  0%  0%  0%  0%
x264 [info]: i8c dc,h,v,p: 100%  0%  0%  0%
x264 [info]: Weighted P-Frames: Y:0.0% UV:0.0%
x264 [info]: ref P L0: 95.5%  0.0%  4.5%
x264 [info]: ref B L0: 70.2% 27.9%  2.0%
x264 [info]: ref B L1: 99.6%  0.4%
x264 [info]: kb/s:25.38
user@dist:~/work/x264_work/x264$ ./bench_rust
x264 [error]: invalid tune 'none'
x264 [info]: using cpu capabilities: none!
x264 [info]: profile High, level 4.0, 4:2:0, 8-bit
Encoded 300 frames in 4.665 seconds (64.31 FPS)
x264 [info]: frame I:2     Avg QP:16.50  size:  2132
x264 [info]: frame P:76    Avg QP:18.69  size:   124
x264 [info]: frame B:222   Avg QP:27.98  size:    81
x264 [info]: consecutive B-frames:  1.3%  0.0%  0.0% 98.7%
x264 [info]: mb I  I16..4:  0.0% 99.4%  0.6%
x264 [info]: mb P  I16..4:  0.0%  0.0%  0.0%  P16..4:  0.0%  0.0%  0.0%  0.0%  0.0%    skip
:100
.0%
x264 [info]: mb B  I16..4:  0.0%  0.0%  0.0%  B16..8:  0.1%  0.0%  0.0%  direct: 0.0%  skip
:99.
9%  L0:54.4% L1:45.6% BI: 0.0%
x264 [info]: 8x8 transform intra:99.2% inter:0.0%
x264 [info]: coded y,uvDC,uvAC intra: 0.4% 0.0% 0.0% inter: 0.0% 0.0% 0.0%
x264 [info]: i16 v,h,dc,p:  0%  0% 100%  0%
x264 [info]: i8 v,h,dc,ddl,ddr,vr,hd,vl,hu: 48%  0% 52%  0%  0%  0%  0%  0%  0%
x264 [info]: i4 v,h,dc,ddl,ddr,vr,hd,vl,hu: 30% 15% 55%  0%  0%  0%  0%  0%  0%
x264 [info]: i8c dc,h,v,p: 100%  0%  0%  0%
x264 [info]: Weighted P-Frames: Y:0.0% UV:0.0%
x264 [info]: ref P L0: 95.5%  0.0%  4.5%
x264 [info]: ref B L0: 70.2% 27.9%  2.0%
x264 [info]: ref B L1: 99.6%  0.4%
x264 [info]: kb/s:25.38
```

fw-immunant: Thanks! I'll move this info into the PR description.

fw-immunant (approved):
The code changes from the LLM look reasonable, including comments which seem informative and appropriate. This does seem to address the relevant issue, which is a deficiency we've wanted to fix for some time.

fw-immunant:
CI is failing because the added test imports `std::mem::offset_of`, which was stabilized in rustc 1.77, after our target nightly (nightly-2023-04-15). I think in this nightly, `offset_of!` is in the prelude but not present in `std::mem`.

PR #1975, opened 2026-08-16
transpile: Refactor, document and correct uses of `used()` flag

Rua:
- Depends on #1976 
- Depends on #1978

I've documented how `is_used` is supposed to be set, and added/removed `used()` accordingly.

If you're wondering what's up with some of the code in the `exprs.c` test suddenly disappearing: The transpiler normally eliminates unused code, which includes unused results of binary expressions:
- `1 + 1;` gets eliminated altogether
- `1 + func();` gets reduced to `func();`

But for some reason unary expressions were not handled the same way, and were emitted even if they didn't do anything. Now they get removed too.


fw-immunant (commented):
Code changes all look good, but before merging I want to go verify that this doesn't interfere with the way Hayroll inserts no-op expressions into c2rust inputs to link C preprocessor macro expansions with their c2rust translations.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -33,8 +33,8 @@ impl<'c> Translation<'c> {
             }
 
             op if op.is_logical() => {
-                let lhs = self.convert_condition(ctx, true, lhs)?;
-                let rhs = self.convert_condition(ctx, true, rhs)?;
+                let lhs = self.convert_condition(ctx.used(), true, lhs)?;
```
ahomescu, line 30: Shouldn't this inherit from the operator? E.g. in `a == b` the two operands are only used if the entire comparison is.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -589,46 +591,22 @@ impl<'c> Translation<'c> {
...
-                let arg = self.convert_expr(ctx, arg, expected_type_id)?;
-                Ok(arg)
-            }
+            CUnOp::Extension => self.convert_expr(ctx, arg, expected_type_id),
             CUnOp::Real | CUnOp::Imag | CUnOp::Coawait => {
                 panic!("Unsupported extension operator")
             }
-        }?;
-
-        // Some unused unary operators (`-foo()`) may have side effects, so we need
```
ahomescu, line 606: Could you add a test for volatile reads and writes (something like ` volatile int vi; -vi;`) and make sure we're not losing them? And `__extension__` although I think we already have a test for that.

ahomescu (changes requested):

c2rust-transpile/src/translator/operators.rs
Rua: Logical operators are `||` `&&`, that's a comparison operator.

c2rust-transpile/src/translator/operators.rs
Rua: I noticed that this code already loses the `vi;` statement in the current master (without the negation):
```c
volatile int vi;
vi;
```

Is that a problem? I'm not sure what the semantics of C are meant to be here. But if this is supposed to emit a volatile read, then it would need to be fixed separately from this PR.

Rua: I've made a separate PR for that issue now, and made this one depend on it.

ahomescu (changes requested):
Needs a rebase, it has conflicts.

ahomescu (approved):

ahomescu:
> before merging I want to go verify that this doesn't interfere

@fw-immunant It looks ready to merge to me, but I'll defer to you. Feel free to push the button whenever you think.

c2rust-transpile/src/translator/operators.rs
```diff
@@ -589,46 +592,22 @@ impl<'c> Translation<'c> {
             }
 
             CUnOp::Deref => self.convert_deref(ctx, expr_type_id, arg),
-            CUnOp::Plus => self.convert_expr(ctx.used(), arg, expected_type_id), // promotion is explicit in the clang AST
+            CUnOp::Plus => self.convert_expr(ctx, arg, expected_type_id), // promotion is explicit in the clang AST
```
ahomescu, line 589: We have a few places that needed this, like builtins and struct literals. Can you test all of the following and make sure they still work?
* `-__builtin_bswap32(f())`
* `!__builtin_expect(f(), 0)`
* `&(struct S){f(), 2};`
* `-(struct S){f(), 2}.x;`

c2rust-transpile/src/translator/literals.rs
```diff
@@ -199,7 +199,7 @@ impl<'c> Translation<'c> {
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         // C compound literals are lvalues, but equivalent Rust expressions generally are not.
         // So if an address is needed, store it in an intermediate variable first.
-        if !ctx.needs_address() || ctx.expanding_macro.is_some() {
+        if !ctx.needs_address() || !ctx.is_used || ctx.expanding_macro.is_some() {
```
ahomescu: What if it has side-effects? See my other comment: `-(struct S){f(), 2}.x;`

c2rust-transpile/src/translator/literals.rs
```diff
@@ -199,7 +199,7 @@ impl<'c> Translation<'c> {
...
             return self.convert_expr(ctx, val, override_ty);
```
ahomescu: The first condition checks for `!ctx.needs_address()` but the others do not. Should this have a `ctx.set_needs_address(false);`?

c2rust-transpile/src/translator/mod.rs
```diff
@@ -3917,7 +3931,7 @@ impl<'c> Translation<'c> {
         expr: WithStmts<Box<Expr>>,
         panic_msg: &str,
     ) -> WithStmts<Box<Expr>> {
-        if ctx.is_unused() {
+        if !ctx.is_used {
             // Recall that if `used` is false, the `stmts` field of the output must contain
```
ahomescu: Update the comment here?

ahomescu (changes requested):

c2rust-transpile/src/translator/literals.rs
Rua: I'm trying to think of a situation where it matters. If the expression is not used, then only the side effects are emitted, and `needs_address` would have to somehow have an effect on the side effect statements.

c2rust-transpile/src/translator/operators.rs
Rua: I've added tests for them now.

c2rust-transpile/src/translator/literals.rs
Rua: I've added it now.

ahomescu (approved):

PR #1976, opened 2026-08-16
transpile: Use `convert_side_effects_expr` in more places

Rua:


fw-immunant (approved):
LGTM, thanks for the cleanup!

PR #1978, opened 2026-08-20
transpile: Consider volatile reads a side effect

Rua:
Fixes the issue noted in https://github.com/immunant/c2rust/pull/1975#discussion_r3820204206, that volatile reads with unused results are omitted when they shouldn't be.

### DIFF
--- a/c2rust-transpile/tests/snapshots/snapshots__transpile@auto_type.c.2021.clang15.snap
+++ b/c2rust-transpile/tests/snapshots/snapshots__transpile@auto_type.c.2021.clang15.snap
@@ -23,8 +23,6 @@ pub unsafe extern "C" fn foo() -> ::core::ffi::c_int {
     let mut px: *mut ::core::ffi::c_int = &raw mut x;
     let mut sx: ::core::ffi::c_ulong =
         ::core::mem::size_of::<::core::ffi::c_int>() as ::core::ffi::c_ulong;
-    let mut y: bar = bar {
-        x: x as ::core::ffi::c_int,
-    } as bar;
+    let mut y: bar = bar { x: x };
     return y.x;
 }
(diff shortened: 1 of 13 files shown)


c2rust-transpile/tests/snapshots/snapshots__transpile@auto_type.c.2021.clang15.snap
```diff
@@ -23,8 +23,6 @@ pub unsafe extern "C" fn foo() -> ::core::ffi::c_int {
     let mut px: *mut ::core::ffi::c_int = &raw mut x;
     let mut sx: ::core::ffi::c_ulong =
         ::core::mem::size_of::<::core::ffi::c_int>() as ::core::ffi::c_ulong;
-    let mut y: bar = bar {
-        x: x as ::core::ffi::c_int,
-    } as bar;
+    let mut y: bar = bar { x: x };
```
ahomescu, line 26: Are these side effects of the new code, or something that was there before?

c2rust-transpile/tests/snapshots/volatile.c
```diff
@@ -11,4 +11,10 @@ void test_volatile(void) {
 
     // https://github.com/immunant/c2rust/issues/1237
     --(volatile_global_struct.p);
+
+    // Unused reads, should be included as side effects.
```
ahomescu, line 15: Add a case for volatile struct member reads?

ahomescu (approved):

c2rust-transpile/tests/snapshots/snapshots__transpile@auto_type.c.2021.clang15.snap
Rua: That's from using `make_cast` for `LValueToRValue` casts. Those casts were unnecessary, but were emitted because the `CTypeId`s were being compared rather than the `CTypeKind`s.

c2rust-transpile/tests/snapshots/volatile.c
Rua: Done.

PR #1979, opened 2026-08-21
transpile: Also prune unwanted expressions

### LINKED ISSUE #1977: Decl pruning can break const macro translation
This was discovered by Jordan Eikenberry (@madberries) [affecting `libusb`](https://github.com/libusb/libusb/blob/3601696fae591a4236a5030ea7b42c717e610aa5/libusb/libusb.h#L1195) and minimized by Claude.

I have confirmed it affects current HEAD (`1e8dfe71c5ee720e49189c39d9687b702e58be61`).

```c
/*  $ c2rust transpile --translate-const-macros conservative qqq.c
 *  $ rustc --edition 2021 --crate-type lib qqq.rs
 *
 * Emits `pub const CONTROL_SETUP_SIZE: ::core::ffi::c_ulong = ::core::mem::size_of::<control_setup>();`
 * which rustc rejects with `expected u64, found usize`.
 */

#include <stddef.h>

struct control_setup { unsigned char b[8]; };

#define CONTROL_SETUP_SIZE (sizeof(struct control_setup))

static size_t unused_use(void) { return CONTROL_SETUP_SIZE; }
size_t live_use(void) { return CONTROL_SETUP_SIZE; }
```

If the `unused_use` is removed entirely, or if `--preserve-unused-functions` is applied, the bug does not manifest.

The simplest fix is to pass the expansion's selected type as the `override_ty` which results in a cast to `c_ulong` applied to the initializer, but arguably a better fix would be to change the behavior of `bubble_expr_types` and/or `prune_unwanted_decls` to eliminate the divergence in type assignment (in favor of `usize`).

This seems related to #1320 but I don't think it's a duplicate of any existing issue I've seen. However, it's possible #1679 will fix this?

Rua:
- Fixes #1977.

### DIFF
--- a/c2rust-transpile/src/c_ast/iterators.rs
+++ b/c2rust-transpile/src/c_ast/iterators.rs
@@ -94,7 +94,7 @@ fn immediate_expr_children_all_types(kind: &CExprKind) -> Vec<SomeId> {
         }
         // We need to iterate the struct type if this offsetof is variable,
         // since it may not get instantiated
-        OffsetOf(_, OffsetOfKind::Variable(qty, _, _)) => intos![qty.ctype],
+        OffsetOf(_, OffsetOfKind::Variable(qty, _, index)) => intos![qty.ctype, index],
         OffsetOf(..) | Literal(..) | ImplicitValueInit(..) => vec![],
         DeclRef(..) => vec![], // don't follow references back!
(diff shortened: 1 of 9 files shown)


fw-immunant (approved):
Code LGTM. But it would be really nice to include a test to demonstrate that this fixes #1977.

Rua:
I've added a regression test.

fw-immunant:
> I've added a regression test.

Wonderful, thank you.

Issue #1981, opened 2026-08-22
`c2rust_run_static_initializers` fails with function-scoped statics


brk:
Small reproducer:
```c
struct foo { int* x; int nelts; };

#define ARR_SIZE(a) (sizeof(a)/sizeof(a[0]))

/* 
globals are supported
     const int global_arr[] = { 1, 2, 3 };
     const struct foo global_s[] = { global_arr, ARR_SIZE(global_arr) };
*/

int main() {
  static const int local_static_arr[] = { 1, 2, 3 };
  static const struct foo local_static_s[] = { local_static_arr, ARR_SIZE(local_static_arr) };
  return 0;
}
```

Relevant parts of translation:

```rs
static mut local_static_s: [foo; 1] = [foo { x: ::core::ptr::null_mut::<::core::ffi::c_int>(), nelts: 0, }; 1];

unsafe fn main_0() -> ::core::ffi::c_int {
    static mut local_static_arr: [::core::ffi::c_int; 3] = [ 1 as ::core::ffi::c_int, 2, 3 ];
    return 0 as ::core::ffi::c_int;
}

unsafe extern "C" fn c2rust_run_static_initializers() {
    local_static_s = [foo {
        x: &raw const local_static_arr as *const ::core::ffi::c_int as *mut ::core::ffi::c_int,
        nelts: ::core::mem::size_of::<[::core::ffi::c_int; 3]>()
            .wrapping_div(::core::mem::size_of::<::core::ffi::c_int>())
            as ::core::ffi::c_int,
    }];
}
```

which fails because `local_static_arr` is not in scope within `fn c2rust_run_static_initializers()`.

This bug affects https://github.com/mity/md4c

PR #1983, opened 2026-08-25
transpile: hoist function-local statics to the root of the current module

ahomescu:
Function-local statics with "uncompilable" initializers are referenced from `c2rust_run_static_initializers` but previously emitted inside the owner function. Hoist them to the top of the current module so `c2rust_run_static_initializers` can reference them by name. Example from issue #1981:
```rust
int main() {
  static const int local_static_arr[] = { 1, 2, 3 };
  static const struct foo local_static_s[] = { local_static_arr, ARR_SIZE(local_static_arr) };
  return 0;
}
```
Here `local_static_s` and `local_static_arr` (because the former references it) need to be moved out of `main`.

Fixes issue #1981.

### DIFF
--- a/c2rust-transpile/src/translator/functions.rs
+++ b/c2rust-transpile/src/translator/functions.rs
@@ -114,6 +114,10 @@ impl<'c> Translation<'c> {
     ) -> TranslationResult<ConvertedDecl> {
         self.function_context.borrow_mut().enter_new(name);
 
+        if let Some(body) = body {
+            self.collect_function_statics_to_hoist(body);
+        }
+
         self.with_scope(|| {
             let mut args: Vec<FnArg> = vec![];
 
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -292,6 +292,9 @@ pub struct Translation<'c> {
     // Accumulated outputs
     pub features: RefCell<IndexSet<&'static str>>,
     sectioned_static_initializers: RefCell<Vec<Stmt>>,
+    /// Function-local statics that need hoisting to module scope because a
+    /// sectioned static's initializer references them.
+    function_statics_to_hoist: RefCell<IndexSet<CDeclId>>,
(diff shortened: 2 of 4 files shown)


ahomescu:
Here is the new Rust output for the test at the top level of `sections.rs`:
```rust
static mut rust_local_static_arr: [::core::ffi::c_int; 3] = [
    1 as ::core::ffi::c_int,
    2 as ::core::ffi::c_int,
    3 as ::core::ffi::c_int,
];
static mut rust_local_static_s: [slice_ref; 1] = [slice_ref {
    p: ::core::ptr::null::<::core::ffi::c_int>(),
    len: 0,
}; 1];
static mut rust_chain_a: [::core::ffi::c_int; 3] = [
    5 as ::core::ffi::c_int,
    6 as ::core::ffi::c_int,
    7 as ::core::ffi::c_int,
];
static mut rust_chain_b: *mut ::core::ffi::c_int =
    unsafe { &raw const rust_chain_a as *mut ::core::ffi::c_int };
static mut rust_chain_c: *mut *mut ::core::ffi::c_int =
    ::core::ptr::null_mut::<*mut ::core::ffi::c_int>();
```

fw-immunant:
Could the test here also exercise the renamer by including multiple functions with distinct, identically-named `static`s?

ahomescu:
Done, the test now produces:
```
static mut rust_sectioned_ptr_0: uintptr_t = 0;
static mut rust_sectioned_ptr_1: uintptr_t = 0;
pub static mut rust_sectioned_ptr: uintptr_t = 0;
// ...
    rust_sectioned_ptr = NULL.expose_addr() as uintptr_t;
    rust_sectioned_ptr_0 = (&raw mut rust_section_me).expose_addr() as uintptr_t;
    rust_sectioned_ptr_1 = (&raw mut rust_section_me3).expose_addr() as uintptr_t;
```

fw-immunant (approved):
LGTM.

PR #1984, opened 2026-08-25
transpile: Deep comparison of types in `make_cast`

Rua:
- Depends on #1978

This allows for detecting cases where two types should be considered equal at a deeper level. For example, `*const usize` vs `*const size_t` or `[usize; N]` vs `[size_t; N]`.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -841,6 +841,160 @@ impl TypedAstContext {
         self.index(resolved_typ_id)
     }
 
+    /// Returns whether `qual_type_id1` and `qual_type_id2` describe the same type with the same
+    /// qualifiers. `resolver` is called on every `CTypeId` before recursing.
+    pub(crate) fn qual_types_eq(
+        &self,
+        qual_type_id1: CQualTypeId,
+        qual_type_id2: CQualTypeId,
+        resolver: &impl Fn(&Self, CTypeId) -> CTypeId,
+    ) -> bool {
+        qual_type_id1.qualifiers == qual_type_id2.qualifiers
+            && self.types_eq(qual_type_id1.ctype, qual_type_id2.ctype, resolver)
+    }
+
+    /// Returns whether `type_id1` and `type_id2` describe the same type.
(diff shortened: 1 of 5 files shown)


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -841,6 +841,155 @@ impl TypedAstContext {
         self.index(resolved_typ_id)
     }
 
+    /// Returns whether `type_id1` and `type_id2` describe the same type with the same qualifiers.
```
ahomescu:
```suggestion
    /// Returns whether `qual_type_id1` and `qual_type_id2` describe the same type with the same qualifiers.
```

c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -841,6 +841,155 @@ impl TypedAstContext {
...
+    /// `resolver` is called on every `CTypeId` before recursing.
+    pub(crate) fn types_eq(
+        &self,
+        mut type_id1: CTypeId,
+        mut type_id2: CTypeId,
+        resolver: &impl Fn(&Self, CTypeId) -> CTypeId,
+    ) -> bool {
+        type_id1 = resolver(self, type_id1);
+        type_id2 = resolver(self, type_id2);
+        self.type_kinds_eq(&self[type_id1].kind, &self[type_id2].kind, resolver)
```
ahomescu: Could short-circuit `if type_id1 == type_id2` and return early either here or in `type_kinds_eq` below.

ahomescu (commented):
This could use tests for deep comparisons of both arrays and function pointers.

c2rust-transpile/src/c_ast/mod.rs
Rua: Done.

c2rust-transpile/src/c_ast/mod.rs
Rua: Done.

Rua:
I've added tests now for casts for pointers to arrays, and for pointers to functions where either a parameter or the return is a typedef.

ahomescu (approved):
LGTM, needs a rebase.

PR #1988, opened 2026-08-31
transpile: Remove `ternary_needs_parens`

Rua:
This seems to be leftover cruft from before c2rust used `syn` to generate its output. `syn` handles this automatically, so it's not necessary anymore.

### DIFF
--- a/c2rust-transpile/src/translator/operators.rs
+++ b/c2rust-transpile/src/translator/operators.rs
@@ -16,12 +16,6 @@ impl<'c> Translation<'c> {
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         let expr_type_id = expected_type_id.unwrap_or(result_type_id);
 
-        // If we're not making an assignment, a binop will require parens
-        // applied to ternary conditionals
-        if !op.is_assignment() {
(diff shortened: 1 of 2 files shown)


c2rust-transpile/src/translator/operators.rs
```diff
@@ -16,12 +16,6 @@ impl<'c> Translation<'c> {
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
         let expr_type_id = expected_type_id.unwrap_or(result_type_id);
 
-        // If we're not making an assignment, a binop will require parens
-        // applied to ternary conditionals
-        if !op.is_assignment() {
```
ahomescu, line 21: This is an actual, visible output change: instead of `({ if foo { ... } else { ... })` now it comes out without the parentheses. I guess we can leave them out, but it's worth discussing first.

ahomescu: Or does `syn` automatically wrap the conditional in parens?

Rua: Yes, `syn` does that itself.

ahomescu (approved):

PR #1989, opened 2026-09-02
transpile: Conservatively disallow mutable lvalues inside macros

Rua:
- Depends on #1993 

Mutable lvalues are values that can either be assigned to, or have a mutable pointer created to them. If they are in a macro, that could pose problems. DeclRefs (including those to mutable variables) are already disallowed in macros, but another "out of nothing" source of mutable lvalues is compound literals. Currently, a macro containing a mutable compound literal is translated as-is, including something like this:

```c
#define MUTATE_ME (&(int){ 42 })

int *p1 = MUTATE_ME;
int *p2 = MUTATE_ME;
```

Each mutable compound literal expression is supposed to create its own distinct memory location (immutable ones are allowed to be coalesced). So these pointers *must* be different, and their pointees must be able to be modified independently.

I've been very conservative here and disallowed all mutable lvalues in macros. There are certainly instances that could be re-allowed on a case-by-case basis. Certain instances of lvalues that undergo an `LValueToRValue` cast for example.

### DIFF
--- a/c2rust-transpile/src/translator/literals.rs
+++ b/c2rust-transpile/src/translator/literals.rs
@@ -197,9 +197,13 @@ impl<'c> Translation<'c> {
         val: CExprId,
         override_ty: Option<CQualTypeId>,
     ) -> TranslationResult<WithStmts<Box<Expr>>> {
+        if !qty.qualifiers.is_const && ctx.expanding_macro.is_some() {
+            return Err("mutable lvalues are not supported inside macros".into());
+        }
+
         // C compound literals are lvalues, but equivalent Rust expressions generally are not.
         // So if an address is needed, store it in an intermediate variable first.
-        if !ctx.needs_address || ctx.expanding_macro.is_some() {
+        if !ctx.needs_address {
             return self.convert_expr(ctx, val, override_ty);
         }
 
--- a/c2rust-transpile/src/translator/mod.rs
+++ b/c2rust-transpile/src/translator/mod.rs
@@ -3544,8 +3544,8 @@ impl<'c> Translation<'c> {
                 matches!(expr_kind, CExprKind::ExplicitCast(..)),
             ),
 
-            Unary(result_type_id, op, arg, _lrvalue) => {
-                self.convert_unary_operator(ctx, override_ty, result_type_id, op, arg)
+            Unary(result_type_id, op, arg, lrvalue) => {
+                self.convert_unary_operator(ctx, override_ty, result_type_id, op, arg, lrvalue)
             }
 
             Conditional(ty, cond, lhs, rhs) => {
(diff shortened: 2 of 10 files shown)


c2rust-transpile/src/translator/literals.rs
```diff
@@ -197,9 +197,13 @@ impl<'c> Translation<'c> {
...
+        if !qty.qualifiers.is_const && ctx.expanding_macro.is_some() {
+            return Err("mutable lvalues are not supported inside macros".into());
+        }
+
         // C compound literals are lvalues, but equivalent Rust expressions generally are not.
         // So if an address is needed, store it in an intermediate variable first.
-        if !ctx.needs_address || ctx.expanding_macro.is_some() {
+        if !ctx.needs_address {
             return self.convert_expr(ctx, val, override_ty);
         }
```
ahomescu, line 209: Do we want to get here with `ctx.expanding_macro.is_some() == true`?

Rua: Yes, but not always. Removing the check means it happens if the address is/isn't taken inside the macro. If the address is taken, then that results in a secondary statement being added, which then fails `to_pure_expr` during macro conversion.

ahomescu: This changes the behavior of the transpiler for something like
```
#define CA ((const int[]){1, 2})
void f(void) { const int *p = CA; }
void g(void) { const int *p = CA; }
```
where master produces something like 
```
pub const CA: [::core::ffi::c_int; 2] = [1 as ::core::ffi::c_int, 2 as ::core::ffi::c_int];
#[no_mangle]
pub unsafe extern "C" fn f() {
    let mut p: *const ::core::ffi::c_int = CA.as_ptr();
}
#[no_mangle]
pub unsafe extern "C" fn g() {
    let mut p: *const ::core::ffi::c_int = CA.as_ptr();
}
```
while this PR outputs:
```
pub const CA: [::core::ffi::c_int; 2] = [1 as ::core::ffi::c_int, 2 as ::core::ffi::c_int];
#[no_mangle]
pub unsafe extern "C" fn f() {
    let c2rust_lvalue_1: [::core::ffi::c_int; 2] =
        [1 as ::core::ffi::c_int, 2 as ::core::ffi::c_int];
    let mut p: *const ::core::ffi::c_int = &raw const c2rust_lvalue_1 as *const ::core::ffi::c_int;
}
#[no_mangle]
pub unsafe extern "C" fn g() {
    let mut p: *const ::core::ffi::c_int = CA.as_ptr();
}
```

Are you sure this is working as intended?

Rua: I investigated that and found that it happens because `recreate_const_macro_from_expansions` inherits the full `ExprContext` from its parent expression, including the `needs_address` flag. #1990 fixes that issue, so I could make this PR depend on that one?

PR #1992, opened 2026-09-02
postprocess: Validate applied rewrites transactionally with cargo check

### LINKED ISSUE #1869: postprocess: Validate full function structure before applying model output

Hellboy28D:
Related to #1869

Currently, transformed Rust is written into the tree without verifying it
still compiles, so malformed model output (e.g. a dropped `}`) breaks the
build and is misattributed to the rewrite.

This adds transactional validation to `c2rust-postprocess`:

- Candidates are applied in batches and validated with `cargo check`;
  any batch that breaks the build is rolled back, and the failing batch
  is bisected to isolate the culprit rewrite.
- The baseline crate is validated **before** any rewrites are applied,
  so a pre-existing broken build is never blamed on the model (exits 1
  with a clear error).
- New repeatable `--validate-cmd CMD` flag runs extra user-supplied shell
  commands (e.g. `cargo build`) after `cargo check`; every command must
  exit 0 for a rewrite to be accepted.

Complementary to the syn-based structural parity check discussed in
#1869 / #1876: `cargo check` catches invalid Rust, while a parity check
would also catch valid-but-modified code. Does not close #1869.

Tests cover the pass, rollback, and baseline-failure paths.


thedataking:
The c2rust postprocessor is much less mature than the transpiler and refactorer. 
Are you using it and hitting a problem where you need the validation? I'm asking since that would be the first time we learn of a third party use of this component.

The stack containing https://github.com/immunant/c2rust/pull/1997 should land next. It is a more fundamental change to make the postprocessor run at an acceptable clip.

Once the above lands, feel free to rebase this PR and ping me for review.

Issue #11, opened 2018-07-03
Broken (or private) links in known-limitations.md


vi:
They go to https://github.com/GaloisInc/C2Rust/issues/, but the issues are not viewable.

glguy:
I've removed the broken links. It'd be nice to make all the internal issues external but we'll need to figure that out separately.

thedataking:
I've removed the links from `docs/known-limitations.md` too. Closing.

PR #1998, opened 2026-09-07
Roll rustc_private toolchain to nightly-2025-01-01

thedataking:
**Formatting split:** formatting-only changes are isolated in [the final commit](https://github.com/immunant/c2rust/commit/efbaf99339bef90ab43b2364499b916d9d17e39b). The preceding commits contain the compiler port and review fixes. The completed repository tree is unchanged.

**Unsafe extern correction:** `reorganize_definitions` retains each foreign block's safety, groups and deduplicates declarations by ABI and safety qualifier, and emits blocks in first-seen source order. Valid `unsafe extern` input remains compilable after moving declarations out of header modules, including mixed safety qualifiers and ABIs.

**Review follow-up:** all 15 findings in [Claude's review](https://github.com/immunant/c2rust/pull/1998#issuecomment-5568432411) have been evaluated. Eleven led to fixes in five focused commits; four were set aside with baseline evidence or scope reasons below. The final local release build, **275 workspace tests (same existing Miri skip)**, **29 shell cases**, and json-c/Lua translation/refactoring/build/runtime stages pass. [Main CI](https://github.com/immunant/c2rust/actions/runs/34096328490) and [nine-project integration](https://github.com/immunant/c2rust/actions/runs/34096328354) passed before these follow-ups; fresh runs are [main CI](https://github.com/immunant/c2rust/actions/runs/34109425362) and [integration](https://github.com/immunant/c2rust/actions/runs/34109425386). The final follow-up **Fable 5.1** assessment found no functional regression in the inspected diff; its full-test and documentation conditions are satisfied. The fresh GitHub runs are still in progress.

The workspace nightly is old enough that its `rustc_private` tools cannot process Rust 2024. This rolls the workspace to **nightly-2025-01-01**, verified as rustc **1.85.0-nightly (`d117b7f211835282b3b177dc64245fff0327c04c`)**, and ports refactoring, analysis, PDG, and instrumentation together. The required rustfmt, rustc-dev, rust-src, Miri, and rust-analyzer components are available.

The workspace remains **edition 2021**, and published-crate **MSRVs remain 1.70**. Those MSRVs apply to building the published crates; generated code has its own compiler requirements. Standalone tools retain their stable toolchain.

Existing nightly defaults for generated code are unchanged: `nightly-2023-04-15` for edition 2021 and `nightly-2026-03-03` for edition 2024. Some translations in either edition can compile on stable, but translations using unstable features still require a compatible nightly. For example, defining C variadic functions and consuming their arguments through `VaList` still require `c_variadic`; calling an externally defined variadic function is supported on stable. The defaults match the compiler APIs currently emitted by C2Rust. In particular, the edition-2024 generator uses a newer `VaList` API absent from the workspace nightly, so replacing that pin alone would break variadic output. This API selection is a C2Rust compatibility choice, not an edition requirement.

One generated-code adaptation is necessary for the real refactor workflow: edition 2021 integer/pointer conversions use equivalent exposed-provenance `as` casts instead of the removed `expose_addr`/`from_exposed_addr` APIs. The explicit `usize` intermediate, enum handling, mutability, and single evaluation are preserved. Both exact compiler implementations define the old/new APIs in terms of these casts. A behavioral regression compiles and executes the generated program on both relevant 2021 compilers and the separate 2024 pin.

The original migration has **nine focused commits** in the reading order below, followed by **five review-fix commits** and **one formatting-only commit**. Pointer compatibility is an independent preparation. The subsequent compiler-port commits are interdependent: intermediate commits do not build the full workspace, and the completed series is the buildable, validated landing unit. Tests and snapshot changes travel with the implementation they validate. Master's CI provisioning fix is included in the base. The follow-up commits preserve the original reading order and make the review changes independently inspectable.

| Commit | Review focus |
| --- | --- |
| [1. transpile: preserve generated pointer casts across compiler pins](https://github.com/immunant/c2rust/commit/1400575cd604361bcb6f91de6a9ec765c8d02d89) | Generated pointer compatibility and its CI compiler prerequisite, with compile/run coverage on all three compiler configurations. |
| [2. build: select the Rust 1.85-era workspace toolchain](https://github.com/immunant/c2rust/commit/0e23827577c47757d5e9815d10cd573110117dda) | Workspace pin, CI installs, and compiler-role documentation. |
| [3. refactor: port the AST schema and traversal infrastructure](https://github.com/immunant/c2rust/commit/a655ce198260cc84b6796b0c7c1de4ac8dc35f78) | AST schema/generated wrappers, copied walkers and attribution, traversal-preservation tests. |
| [4. refactor: preserve macro reconstruction and source printing](https://github.com/immunant/c2rust/commit/de69c792cdae45639d7aa039b986c6a02bbf4a99) | AST builders, macro collapse/matching, fresh-AST printing, literal and token-consumer regressions, and unsafe extern preservation during declaration regrouping. |
| [5. refactor: port compiler sessions, resolution, and diagnostics](https://github.com/immunant/c2rust/commit/5ecc70d014acaa5a7d1362847d8886f874565817) | Compiler ownership/reload, HIR identity, diagnostics, body analysis, and driver/parser/resolution tests. |
| [6. refactor: preserve type reflection and compiler-specific snapshots](https://github.com/immunant/c2rust/commit/fdddc75698d7ed9591de645f4d38f5f5c5bdf759) | Function-pointer and binder reflection, type queries, compiler-specific snapshot routing and expectations. |
| [7. refactor: port ownership analysis and transformation consumers](https://github.com/immunant/c2rust/commit/bd0fa7e82e73afdba5af8f8b8e7e0f9ca7b0d277) | Ownership MIR analysis and remaining transformation consumers, including abstract-expression fixtures. |
| [8. instrument: port MIR events and validate PDG preservation](https://github.com/immunant/c2rust/commit/04928bd9b7f96a81006c41d5cce616fc04696f93) | MIR instrumentation, unwind/failure regressions, PDG snapshot explanations and fixture freshness. |
| [9. analyze: adapt dataflow and rewrites to the new MIR lowering](https://github.com/immunant/c2rust/commit/3b82de0a686232c8b733e5a17752579242ae23b0) | Analyzer MIR lowering, dataflow, pointer permissions, Pin rewrites, and associated regressions. |

The follow-up commits are:

| Commit | Review changes |
| --- | --- |
| [10. Target-width constants](https://github.com/immunant/c2rust/commit/5c9f1acbc5a58ed413f380fd625480854fad9dd6) | Fix `usize`/`isize` literal truncation and host-width cast evaluation; test 16/32/64-bit values and real rewrites in both editions. |
| [11. Edition default and unsafe lookup](https://github.com/immunant/c2rust/commit/584cfb0911c7800a4f8d019b2a50bc3962382c89) | Restore omitted-edition 2021 default, use a span set, and add actual Cargo lint-flag and accepted-error coverage. |
| [12. Printer allocation and visitor duplication](https://github.com/immunant/c2rust/commit/0336cb81424233da64d00b82701de5497cb07f04) | Share one lazy ParseSess per print operation, cache attribute token materialization, and consolidate equivalent associated-item paths with baseline-checked regressions. |
| [13. Check-only MIR and explicit tail calls](https://github.com/immunant/c2rust/commit/d0346c37a5d53d7c1bf10a8980e2f09aeee8aec3) | Enable MIR checking for metadata-only builds and reject unsupported explicit tail calls before writing metadata. |
| [14. Explicit analyzer structures](https://github.com/immunant/c2rust/commit/0287077afdd9f34a6fb30ff1409dc914b26c3672) | Model typed-let bindings explicitly; narrow dead-code expectations, reuse sorting, and clarify fallback diagnostics. |
| [15. Formatting only](https://github.com/immunant/c2rust/commit/efbaf99339bef90ab43b2364499b916d9d17e39b) | Isolate rustfmt changes, including the two disabled transform modules, from functional changes. |

<details>
<summary>Feedback set aside and qualifications</summary>

- **#1, Cargo unused-unsafe regression:** the old compiler explicitly treated allowed warnings as used blocks. Exact old/new Cargo executions both retain redundant blocks under tool-added `-Awarnings`, and both remove only redundant blocks under `--force-warn unused_unsafe`. Added coverage; preserved suppression.
- **#4, diagnostic reset:** accepted preexisting errors are an established refactoring contract. Old/new executable sequences have matching exits and byte-identical output; malformed syntax and new autoretype errors still fail. The actual compiler flushes delayed bugs while the previous real-error count remains, matching old handler-drop behavior.
- **#5, repeated loop-analysis pass:** the structural concern is plausible, but no working hot path was established. The command's existing regression explicitly expects its broken HIR lookup to panic; other old/new probes no-op before the pass. A speculative batching rewrite was discarded.
- **#11, replacing type reflection with `_`:** Pat and UnsafeBinder have nameable Rust forms, unlike closures/coroutines. The refactor API handles Rust beyond direct C output; erasing these constructors would break the existing nested-binder, pattern and definition-mapping tests.
- **#2 severity qualification:** the 16-bit evaluator bug is real and fixed, but the asserted unsafe rewrite was not demonstrated. Replacement bounds reject `65536usize as u16` before equality comparison. The preexisting signed-minimum case remains conservatively unsimplified; its negative runtime value is tested.

</details>

No dependency roll or workspace edition conversion is included.

Validation used Linux x86_64, an LLVM 18 AST exporter, an isolated baseline archive at `e1e5bf257863107c54e9f42b345c2aeccd925458`, exact old/target compiler source checkouts, and separate build directories. Workspace Cargo build/nextest commands used `--locked` and denied warnings.

| Workflow | Baseline | Candidate |
| --- | --- | --- |
| Release workspace build | Pass | Pass |
| Release workspace nextest | 234 passed, 1 existing skip | **275 passed, same skip**; all 234 baseline and all 270 pre-review passing test identities retained |
| Python translator runner, `tests/unit` | 148 successes, 2 expected failures | Same; no unexpected failures or successes |
| Original refactor shell runner | 29 passed | 29 passed |
| json-c and Lua integration | All 16 selected stages passed | Same stages passed; 12 translation/refactor/compile/runtime stages rerun, 4 C build stages cached after earlier candidate execution |
| [Nine-project GitHub integration, Clang 15/18](https://github.com/immunant/c2rust/actions/runs/34078443738) | 76 successful configured stages per job | Same 76 stage identities and scripts per job, all successful; no missing or extra stages |
| Published transpiler crates, exact Rust 1.70 | Check passed | Check passed; review follow-ups do not change these crates |
| Runtime crates / bitfields `no_std`, Rust 1.70 | 2 tests and checks passed | Sources unchanged |
| Postprocess lint/type checks and pytest | **58 passed, 0 skipped** | **58 passed, 0 skipped**, including transpiler-dependent fixtures |
| Four standalone tools on stable | Format/build/nextest/doc passed; each has 0 tests | Sources/toolchain unchanged |

The [rebased main CI](https://github.com/immunant/c2rust/actions/runs/34096328490) passed every configured job, including Ubuntu Clang 15/18, macOS Clang 17/22, all four standalone tools, and postprocess. The [rebased nine-project integration workflow](https://github.com/immunant/c2rust/actions/runs/34096328354) also passed both Clang jobs. The latest five corrective commits have completed local validation and are rerunning both workflows.

Each broader integration job retained nine transpiles, eight configured refactors, and the associated build/runtime and cache-only postprocess stages. Postprocess CI exercised cached responses and uncached-work handling; each Clang job applied 50 cached json-c comment updates, while Lua/Python2 skipped uncached work. Downstream builds and runtime checks passed. No live-LLM execution is claimed.

The 41 new tests are included in the 275 count; five were added for the GitHub review. Refactor and analyzer regressions compile and execute edition 2021/2024 outputs; instrumentation covers normal and unwind paths across editions 2015/2018/2021/2024 at minimum MIR optimization, and verifies that invalid edition-2024 input fails without overwriting metadata. Pin adjustment coverage compiles/runs the emitted rewrite. The explicit-tail-call regression checks metadata only because this compiler cannot codegen that feature.

Snapshot review: each PDG profile retains all 80 graph write-permission sets. Each profile’s node count changes from 755 to 758 because the target retains three identity pointer casts in `pointers::main`, `test_arg`, and `test_arg_rec`; comparison contracts only those source-justified copy nodes. Other differences are compiler printing and MIR numbering. The PDG harness now rebuilds the primary fixture package/profile so changed instrumenter contents cannot reuse stale `RUSTC_WRAPPER` artifacts. Refactor snapshot changes are the compiler's new type-ascription spelling and two local overflow-lint allowances in an existing arithmetic fixture. Original transformation assertions remain, and all 46 refactor snapshot tests pass.

<details>
<summary>Core reproduction commands</summary>

Use Python 3.12+ with `scripts/requirements.txt`, separate baseline/candidate `CARGO_TARGET_DIR` values, and the corresponding checkout/compiler:

```sh
export LLVM_CONFIG_PATH=/usr/bin/llvm-config-18
export CLANG_PATH=/usr/bin/clang-18
export LIBCLANG_PATH=/usr/lib/llvm-18/lib
export CC=/usr/bin/clang-18
RUSTFLAGS=-Dwarnings RUSTDOCFLAGS=-Dwarnings cargo build --release --locked
RUSTFLAGS=-Dwarnings RUSTDOCFLAGS=-Dwarnings cargo nextest run --release --workspace --locked --no-fail-fast
python scripts/test_translator.py tests/unit
# The shell runner expects target/debug/c2rust-refactor; point it at the tested binary.
(cd c2rust-refactor/tests && ./run-all-tests.sh)
# Set C2RUST_DIR and put built c2rust/standalone tools and Python dependencies on PATH.
python tests/integration/test.py json-c lua --refactor-jobs 1 --stages \
  autogen configure make transpile cargo.transpile check.transpile \
  refactor cargo.refactor check.refactor
cargo +1.70.0 check --locked -p c2rust -p c2rust-transpile
cargo fmt --all --check
git diff --check
```

Postprocess: run `uv sync` and `uv run pytest -v -ra` in `c2rust-postprocess`, with built split/merge tools on PATH and the tested `c2rust` binary available at `target/release/c2rust` (the fixture’s expected path).

</details>

Previous migration discussions and follow-up fixes were reviewed, including [#1328](https://github.com/immunant/c2rust/pull/1328), [#1614](https://github.com/immunant/c2rust/pull/1614), [#1904](https://github.com/immunant/c2rust/pull/1904)/[#1913](https://github.com/immunant/c2rust/pull/1913), and the span-collision fix [#1967](https://github.com/immunant/c2rust/pull/1967). API adaptations were checked against [the exact target compiler](https://github.com/rust-lang/rust/tree/d117b7f211835282b3b177dc64245fff0327c04c), with the old implementation used for behavioral comparisons.

Adversarial review: Opus 5 subsystem reviews identified issues addressed with parser diagnostic/recovery fixes, Pin reborrow preservation, structural attribute-token equality, and additional behavioral checks. **Fable 5.1 completed the final semantic and reviewability assessment on the verified exact model and found no blocking defect.** Its remaining evidence questions were checked against the exact printer source and old/new executable diagnostics; accepted-error warnings, exit behavior, and output matched baseline, and resolver state is copied before HIR queries consume it.

Macro printing renders the updated AST while preserving baseline bare-fragment spelling. Exact old-binary comparisons and focused regressions compile and execute five discriminating token-consumer cases in both editions, including `stringify!`, `concat!`, arbitrary token-tree forwarding, and negative postfix expressions. Preexisting text-serialization limits remain explicit. All review findings were resolved through source checks or behavioral evidence; the final full validation above was repeated after the last code correction.

Limits: live-LLM postprocessing was not exercised. The existing PDG Miri test stays skipped. The already-excluded `related_decls` test still fails on a missing `OTHER_CONSTANT`; legacy C cross-checks fail importing removed runner helpers, and Rust cross-checks cannot build their currently resolved dependencies on either compiler (old Cargo edition parsing versus target MSRV). These are baseline limitations, not passing coverage. Existing unsupported refactor cases were not disabled or relabeled. The already-broken range-loop command's module traversal remains a deferred performance concern. Instrumentation enables per-pass MIR checking for both full builds and `cargo check`, overriding `-Zvalidate-mir=no`. Explicit tail calls are rejected with a diagnostic because their return transfers and entrypoint finalization are not yet supported.

### DIFF
--- a/c2rust-transpile/tests/toolchain_pointer_casts.rs
+++ b/c2rust-transpile/tests/toolchain_pointer_casts.rs
@@ -1,0 +1,100 @@
+//! Generated pointer casts must work on both generated-code and workspace pins.
+mod common;
+
+use c2rust_rust_tools::RustEdition;
+use std::fs;
+use std::path::Path;
+use std::process::{Command, Output};
+use tempfile::TempDir;
+
+fn successful(command: &mut Command) -> Output {
+    let output = command.output().unwrap();
+    assert!(
+        output.status.success(),
+        "{command:?}: {}\n{}",
+        output.status,
+        String::from_utf8_lossy(&output.stderr)
+    );
+    output
+}
+
+fn compile_and_run(source: &Path, edition: RustEdition, toolchain: Option<&str>) -> Vec<u8> {
+    let executable = source.with_extension(format!("{}.bin", toolchain.unwrap_or("workspace")));
+    successful(
+        Command::new("rustc")
+            .args(toolchain)
+            .args([
+                "--edition",
+                edition.as_str(),
+                "--crate-name",
+                "pointer_casts",
+            ])
+            .arg(source)
+            .arg("-o")
+            .arg(&executable),
+    );
+    successful(&mut Command::new(executable)).stdout
+}
+
+#[test]
+fn integer_pointer_casts_preserve_width_mutability_and_single_evaluation() {
+    let fixture = TempDir::new().unwrap();
+    let source = fixture.path().join("pointer_casts.c");
+    fs::write(
+        &source,
+        r#"
+typedef __UINTPTR_TYPE__ uintptr_t;
+extern int printf(const char *, ...);
+static int values[3] = {11, 22, 33};
+static int calls;
+static const int *constant_pointer = (const int *)16;
+enum small { ZERO = 0, ONE = 1 };
+static int *next_pointer(void) { calls++; return &values[1]; }
+static uintptr_t next_address(uintptr_t address) { calls++; return address; }
+int main(void) {
+    uintptr_t address = (uintptr_t)next_pointer();
+    if (calls != 1 || address != (uintptr_t)&values[1]) return 1;
+    int *mutable = (int *)next_address(address);
+    if (calls != 2 || mutable != &values[1]) return 2;
+    *mutable = 47;
+    const int *immutable = (const int *)next_address(address);
+    if (calls != 3 || *immutable != 47 || values[0] != 11 || values[2] != 33) return 3;
+    unsigned long long wider = (unsigned long long)immutable;
+    if ((uintptr_t)wider != address) return 4;
+    unsigned char narrow = (unsigned char)immutable;
+    if (narrow != (unsigned char)address) return 5;
+    void *negative = (void *)(long)-1;
+    if ((uintptr_t)negative != ~(uintptr_t)0) return 6;
+    void *from_enum = (void *)ONE;
+    if ((uintptr_t)from_enum != 1 || (enum small)(void *)0 != ZERO) return 7;
+    __uint128_t wide = ((__uint128_t)1 << (sizeof(void *) * 8)) | address;
+    int *from_wide = (int *)wide;
+    if (from_wide != mutable || *from_wide != 47) return 8;
+    uintptr_t zero = 0;
+    if ((const int *)zero != 0 || (uintptr_t)constant_pointer != 16) return 9;
+    printf("%d %d %d %d\n", values[0], *immutable, values[2], calls);
+    return 0;
+}
+"#,
+    )
+    .unwrap();
+    // Each check in the fixture exits unsuccessfully if pointer behavior differs.
+    let expected = b"11 47 33 3\n";
+
+    let (_commands_dir, commands) = c2rust_transpile::create_temp_compile_commands(&[source]);
+    for edition in RustEdition::ALL.iter().copied() {
+        let output_dir = fixture.path().join(edition.as_str());
+        let mut config = common::config(edition);
+        config.emit_c_decl_map = false;
+        config.output_dir = Some(output_dir.clone());
+        c2rust_transpile::transpile(config, &commands, &[]);
+        let generated = output_dir.join("src/pointer_casts.rs");
+        assert_eq!(
+            compile_and_run(&generated, edition, Some(edition.toolchain())),
+            expected
+        );
+        if edition == RustEdition::Edition2021 {
+            assert_eq!(compile_and_run(&generated, edition, None), expected);
+        }
+    }
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,8 @@ jobs:
       # Run after `rust-cache` so that this is cached.
       - name: Install Rust toolchains
         run: |
+          rustup toolchain install nightly-2025-01-01 \
+            --profile minimal --component rustfmt
           rustup toolchain install nightly-2023-04-15 \
             --profile minimal --component rustfmt,rustc-dev
           rustup toolchain install nightly-2026-03-03 \
--- a/c2rust-transpile/src/translator/pointers.rs
+++ b/c2rust-transpile/src/translator/pointers.rs
@@ -569,6 +569,14 @@ impl<'c> Translation<'c> {
                 val.map(|val| mk().cast_expr(val, size_type_rs))
             };
 
+            // The old from_exposed_addr[_mut] APIs were removed before the
+            // workspace compiler. An `as` cast has exactly their exposed-
+            // provenance semantics and works on both the generated edition
+            // 2021 toolchain and the workspace refactorer's toolchain.
+            if self.tcfg.edition == RustEdition::Edition2021 {
+                return Ok(val.map(|val| mk().cast_expr(val, target_ty)));
+            }
+
             // Then convert the `usize` into a pointer.
             let pointee_type_id = self
                 .ast_context
@@ -576,22 +584,9 @@ impl<'c> Translation<'c> {
                 .expect("target type must be a pointer");
             let mutability = pointee_type_id.mutability();
 
-            let fn_name = match self.tcfg.edition {
-                RustEdition::Edition2021 => {
-                    // Rust 1.76: feature name changed to `exposed_provenance[_mut]`
-                    // Rust 1.84: stabilized
-                    self.use_feature("strict_provenance");
-
-                    // Rust 1.79: method name changed to `with_exposed_provenance[_mut]`
-                    match mutability {
-                        Mutability::Immutable => "from_exposed_addr",
-                        Mutability::Mutable => "from_exposed_addr_mut",
-                    }
-                }
-                RustEdition::Edition2024 => match mutability {
-                    Mutability::Immutable => "with_exposed_provenance",
-                    Mutability::Mutable => "with_exposed_provenance_mut",
-                },
+            let fn_name = match mutability {
+                Mutability::Immutable => "with_exposed_provenance",
+                Mutability::Mutable => "with_exposed_provenance_mut",
             };
             let pointee_type_rs = self.convert_pointee_type(pointee_type_id.ctype)?;
             let type_args = mk().angle_bracketed_args(vec![pointee_type_rs]);
@@ -630,19 +625,12 @@ impl<'c> Translation<'c> {
             }))
         } else {
             // First convert the pointer to `usize`.
-            let method_name = match self.tcfg.edition {
-                RustEdition::Edition2021 => {
-                    // Rust 1.76: feature name changed to `exposed_provenance`
-                    // Rust 1.84: stabilized
-                    self.use_feature("strict_provenance");
-
-                    // Rust 1.79: method name changed to `expose_provenance`
-                    "expose_addr"
-                }
-                RustEdition::Edition2024 => "expose_provenance",
-            };
-
-            let val = val.map(|val| mk().method_call_expr(val, method_name, vec![]));
+            let val = val.map(|val| match self.tcfg.edition {
+                // expose_addr, now removed, is equivalent to this cast. Keep
+                // the usize intermediate before converting to the C target type.
+                RustEdition::Edition2021 => mk().cast_expr(val, mk().ident_ty("usize")),
+                RustEdition::Edition2024 => mk().method_call_expr(val, "expose_provenance", vec![]),
+            });
 
             // Then cast the `usize` to the target type.
             let size_type_id = self.ast_context.type_for_kind(&CTypeKind::Size);


thedataking:
## Code review: toolchain roll to `nightly-2025-01-01`

Reviewed `master...roll-toolchain` (171 files, ~7.5k insertions), excluding the regenerated `.snap` files (spot-checked; consistent with the new MIR spelling). New rustc APIs were cross-checked against the actual `nightly-2025-01-01` sources, and the copied `ast_manip/fold/walk.rs` was diffed function-by-function against upstream `rustc_ast/src/mut_visit.rs` — all 46 functions match modulo rustfmt, with `visit_tts` the only intentional (documented) adaptation. The precedence migration (`gen/rewrite.py` + `binop_left_prec` + `strategy/print.rs`) is internally consistent on the new contiguous `ExprPrecedence` scale.

### Correctness

**2. `eval_const` truncates `isize`/`usize` literals to 16 bits** — `c2rust-refactor/src/transform/casts.rs:474` and `:498`

```rust
LitKind::Int(i, LitIntType::Signed(IntTy::Isize)) => Some(ConstantValue::Int(i.get() as i16 as i128)),
LitKind::Int(i, LitIntType::Unsigned(UintTy::Usize)) => Some(ConstantValue::Uint(i.get() as u16 as u128)),
```

Pre-existing, but both lines are touched here and they gate redundant-cast removal. `70000usize` folds to `4464`, so a valid simplification is skipped.

**3. Refactoring without `--edition` now parses as Rust 2015** — `c2rust-refactor/src/driver.rs:198`

`run_compiler` dropped `config.opts.edition = Edition::Edition2021`, and `run_refactoring` no longer wraps the session in `create_session_globals_then(Edition2021, ..)`. For the documented `CmdLine` form (`c2rust-refactor ... -- src/main.rs`), `build_session_options` leaves the 2015 default, so paths/`dyn`/prelude resolve differently and typed (Phase 2/3) commands fail on edition-2021 output. Every new test passes `--edition` explicitly, so the default is untested.

### Efficiency

**6. A fresh `ParseSess` per interpolated token** — `c2rust-refactor/src/ast_manip/print.rs:40`

`printable_tokens` builds a `ParseSess` (own `SourceMap`, `DiagCtxt`, registries) inside the `.map()` closure, recursively. `tokens_to_string`/`nonterminal_to_string` are called from `rewrite/strategy/print.rs` for every reprinted node, so one pass creates thousands of throwaway sessions. Hoisting one session above the closure gives identical output.

**7. Linear span scan per block** — `c2rust-refactor/src/context.rs:73`

`is_unused_unsafe_block` does `Vec::contains` and is called (via `funcs.rs:364`) for every block/expr/stmt: O(blocks x spans). A `FxHashSet<Span>` built once in `with_unused_unsafe_spans` makes it O(blocks).

**8. `same_attr` reallocates token trees quadratically** — `c2rust-refactor/src/collapse/macros.rs:369`

`restore_attrs` runs `old.filter(|a| !new.any(|b| same_attr(a, b)))` and `same_attr` calls `token_trees()` on *both* sides each time — 2·K² `Vec<TokenTree>` allocations for K attributes. Hoist `a`'s trees out of the inner loop and precompute the new attrs' trees once.

**9. Whole-crate MIR validation forced on** — `dynamic_instrumentation/src/callbacks.rs:23`

`config.opts.unstable_opts.validate_mir = true` replaces a targeted `Validator` run over just the instrumented body with validation of every body after every MIR pass, and it silently overrides an explicit `-Zvalidate-mir=no` — including via the `--rustflags` path `tests/editions.rs` itself uses to control MIR options.

### Simplification / altitude

**10. Duplicated `flat_map_assoc_item` branches** — `c2rust-refactor/src/collapse/macros.rs:672` (and `:229`)

In `ReplaceTokens` the `AssocCtxt::Impl` and `AssocCtxt::Trait` arms are byte-for-byte identical apart from the local binding name; in `CollapseMacros` ~30 lines are duplicated, differing only in `mac_impl_item` vs `mac_trait_item`. `fn_edit.rs` already shows the right shape for this API merge (map `ctxt` to a small value, then one code path). The same boilerplate landed in `funcs.rs`, `items.rs`, `output_exprs.rs`, and `cfg_attr.rs`.

**12. Type annotation smuggled into an identifier field** — `c2rust-analyze/src/rewrite/expr/convert.rs:747`

`reborrow_pin` passes `"__c2rust_pin: ::core::pin::Pin<&mut _>"` as `Rewrite::Let1`'s *name*, relying on `apply.rs` emitting `let {name} = ` verbatim. Every other `Let1` use passes a real binding name; a future change that validates or reformats that string emits uncompilable code. A `Rewrite::LetTyped(name, ty, expr)` variant states the intent.

**13. Blanket `#[allow(dead_code)]` on three central enums** — `c2rust-analyze/src/util.rs:20`, `:104`, `:122`

Suppresses dead-code reporting for all present and future variants/fields of `RvalueDesc`, `UnknownDefCallee`, and `Callee`. The lint is only about fields read solely through `Debug`; a field-scoped `#[expect(dead_code)]` keeps the signal for the rest of the type.

**14. `DefId` sort key copy-pasted seven times** — `c2rust-analyze/src/analyze.rs:635`, `:1543`, `:1562`, `:1602`, `:1637`, plus `c2rust-refactor/src/transform/externs.rs` and `analysis/ownership/mod.rs`

`sorted_def_ids` already exists in `analyze.rs`. One `fn def_id_sort_key(did: DefId) -> (u32, u32)` (or reusing that helper) keeps the seven comparators from drifting.

**15. Inconsistent unsupported-type rendering** — `c2rust-analyze/src/context.rs:1626`

The new `Pat(..) | UnsafeBinder(..) | CoroutineClosure(..)` arm prints `"{:?}"` while the adjacent arm prints `"unknown:{:?}"`, so a fallback rendering is indistinguishable from a fully labeled one. The comment also claims these constructors "retain their recursive labels", but the arm never descends into `lty.args`.

c2rust-transpile/tests/toolchain_pointer_casts.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#[test]
+fn integer_pointer_casts_preserve_width_mutability_and_single_evaluation() {
+    let fixture = TempDir::new().unwrap();
+    let source = fixture.path().join("pointer_casts.c");
+    fs::write(
+        &source,
+        r#"
+typedef __UINTPTR_TYPE__ uintptr_t;
```
ahomescu, line 46: Couldn't this be either a regular translator test or a snap instead?

c2rust-macros/src/lib.rs
```diff
@@ -26,12 +26,15 @@ impl VisitorImpls {
         let folder_ident = Ident::new(&folder_name, Span::call_site());
 
         if !walk.stmts.is_empty() {
-            let noop_fn_name = format!("noop_{}", method_name);
+            let noop_fn_name = format!(
+                "walk_{}",
```
ahomescu, line 30: Are these equivalent?

c2rust-refactor/gen/util.py
```diff
@@ -24,7 +24,10 @@ def g(*args, **kwargs):
 @comma_sep
 def struct_fields(fields, suffix, bind_mode):
     for f in fields:
-        yield 'r#%s: %s%s%s' % (f.name, bind_mode, f.name, suffix)
+        if not suffix:
```
ahomescu: Nit: `if suffix:` could go first.

c2rust-refactor/src/ast_manip/ast_deref.rs
```diff
@@ -1,11 +1,16 @@
...
+use rustc_ast::tokenstream::DelimSpacing;
 use rustc_ast::tokenstream::{DelimSpan, LazyAttrTokenStream, Spacing, TokenStream, TokenTree};
 use rustc_ast::*;
-use rustc_span::hygiene::SyntaxContext;
-use rustc_span::source_map::{Span, Spanned};
-use rustc_span::symbol::{Ident, Symbol};
+use rustc_data_structures::packed::Pu128;
+use rustc_errors::ErrorGuaranteed;
+use rustc_span::source_map::Spanned;
+use rustc_span::Span;
```
ahomescu: Nit: these could all be folded into a single line.

c2rust-refactor/src/transform/casts.rs
```diff
@@ -472,7 +476,10 @@ fn eval_const<'tcx>(e: P<Expr>, cx: &RefactorCtxt) -> Option<ConstantValue> {
                 LitKind::Int(i, LitIntType::Unsuffixed) => Some(ConstantValue::Uint(i.get())),
 
                 LitKind::Int(i, LitIntType::Signed(IntTy::Isize)) => {
-                    Some(ConstantValue::Int(i.get() as i16 as i128))
+                    Some(ConstantValue::Uint(i.get()).cast(
+                        SimpleTy::Size(true),
+                        cx.ty_ctxt().data_layout.pointer_size.bits(),
```
ahomescu, line 481: `i16` was intentional, it's the largest guaranteed size for a Rust pointer. I don't think we should change this.

c2rust-refactor/src/ast_manip/fold/walk.rs
```diff
@@ -0,0 +1,530 @@
+//! Defaults made private in rustc's mutable AST visitor.
+//!
+//! Copied from rustc_ast/src/mut_visit.rs at Rust commit
```
ahomescu, line 3: Do we want this approach (copying code verbatim from rustc)?

c2rust-refactor/src/ast_manip/fold/walk.rs
```diff
@@ -0,0 +1,530 @@
+//! Defaults made private in rustc's mutable AST visitor.
```
ahomescu, line 1: Hmmm could we auto-generate some of this from `ast.txt`?

c2rust-refactor/src/ast_manip/lr_expr.rs
```diff
@@ -278,3 +296,9 @@ where
         fold_expr_with_context(e, Context::Rvalue, |e, ctx| callback(e, ctx))
     })
 }
+
+impl<T: LRExpr + ?Sized + Clone> LRExpr for Arc<T> {
+    lr_expr_fn!((self, next(T)) => {
+        next(Arc::make_mut(self));
```
ahomescu: If the `Arc` has multiple references, the `make_mut` will first clone it into a single-reference copy, then return a mutable reference here. I'm not sure that's always the right behavior here, we should investigate.

c2rust-refactor/src/ast_manip/number_nodes.rs
```diff
@@ -59,13 +59,14 @@ fn reset_nonterminal_node_ids(nt: &mut Nonterminal, visitor: &mut ResetNodeIds)
...
-        Nonterminal::NtIdent(..) | Nonterminal::NtLifetime(..) | Nonterminal::NtMeta(..) => {}
+        Nonterminal::NtMeta(..) => {}
     }
 }
 
 fn reset_interpolated_node_ids(tokens: TokenStream, visitor: &mut ResetNodeIds) -> TokenStream {
     tokens
-        .into_trees()
+        .iter()
+        .cloned()
```
ahomescu, line 69: Do we need the clones here?

c2rust-refactor/src/ast_manip/util.rs
```diff
@@ -30,7 +31,8 @@ impl PatternSymbol for Ident {
 
 impl PatternSymbol for Lit {
     fn pattern_symbol(&self) -> Option<Symbol> {
-        (self.kind == token::LitKind::Err).then_some(self.symbol)
+        (self.kind == token::LitKind::Str && self.suffix == Some(Symbol::intern("$c2rust_pattern")))
```
ahomescu, line 34: Why did it add the suffix here?

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -106,11 +106,11 @@ impl<'a> Make<Mutability> for &'a str {
...
 }
 
-impl<'a> Make<Unsafe> for &'a str {
-    fn make(self, _mk: &Builder) -> Unsafe {
+impl<'a> Make<Safety> for &'a str {
+    fn make(self, _mk: &Builder) -> Safety {
         match self {
-            "" | "safe" | "normal" => Unsafe::No,
-            "unsafe" => Unsafe::Yes(DUMMY_SP),
+            "" | "safe" | "normal" => Safety::Default,
```
ahomescu: Shouldn't `safe` map to `Safety::Safe`?

c2rust-refactor/src/ast_builder/builder.rs
```diff
@@ -1435,17 +1460,25 @@ impl Builder {
...
         })
     }
 
     pub fn err_expr(self) -> P<Expr> {
+        // This value is a temporary replacement while moving an expression
+        // out of the AST, not a recovered parser error with a diagnostic.
         P(Expr {
             id: self.id,
-            kind: ExprKind::Err,
+            kind: ExprKind::Dummy,
```
ahomescu, line 1482: What happened to `Err`?

c2rust-refactor/src/ast_manip/print.rs
```diff
@@ -0,0 +1,418 @@
+//! Print transformed macro fragments from their AST, rather than cached tokens.
```
ahomescu, line 1: Is this a copy of rustc code, or brand new?

c2rust-refactor/src/transform/reorganize_definitions.rs
```diff
@@ -1993,15 +1999,20 @@ impl<'a, 'tcx> HeaderDeclarations<'a, 'tcx> {
...
+                DeclKind::ForeignItem(fi, abi, safety) => {
                     if last_foreign_item_mod != Some(cur_mod_name) {
                         st.add_comment(
                             fi.id,
                             make_header_comment(last_foreign_item_mod, cur_mod_name),
                         );
                         last_foreign_item_mod = Some(cur_mod_name);
                     }
-                    foreign_items.entry(abi).or_default().push(fi);
+                    // Group by qualifier, not its source span, so separate unsafe
```
ahomescu, line 2035: Isn't this approximately the same as #1930 ?

thedataking: The changes overlap in the same regrouping code, but fix different bugs: #1930 makes extern-block ordering deterministic and preserves header comments; #1998 preserves unsafe. One of us gonna have to rebase.

ahomescu:
A few low-severity observations from a review pass on this PR (CI is green and the port looks correct overall):

1. **`reorganize_definitions`: extern block emission order is nondeterministic** (`c2rust-refactor/src/transform/reorganize_definitions.rs` ~L2013, L2037). The grouped foreign items live in a `std::collections::HashMap`, and the key now includes the block's safety discriminant. Any crate that mixes `extern "C" { … }` and `unsafe extern "C" { … }` (or already mixes ABIs, which was nondeterministic before this PR) gets its emitted blocks, and the module header comments attached to them, in an order that varies between runs. Sorting the entries before emission would make the output reproducible.

2. **`reorganize_definitions`: dedup ignores block safety** (`find_foreign_item`, ~L2213). Grouping now splits by safety, but the equivalence check still matches only on `(ident, namespace, abi)`. A declaration present in both a plain and an `unsafe` extern block collapses to whichever header was seen first, inheriting that block's qualifier. Low impact since mixed blocks are already an error in edition 2024, but the two keys should probably agree.

3. **`dynamic_instrumentation`: post-instrumentation MIR validation only runs on full builds** (`dynamic_instrumentation/src/callbacks.rs` ~L45). The comment is right that the pass manager validates unconditionally at the transition to `Runtime(Optimized)`, but that transition never happens under `cargo check`. Previously the `Validator` ran on every `mir_built`; now `c2rust-instrument -- check` can accept malformed instrumented MIR and still write metadata, and the failure surfaces later as an ICE in a real build. Since `Validator` is private, setting `-Zvalidate-mir` in the session options from the config callback would restore the old coverage.

4. **`c2rust-analyze`: `Adjust::ReborrowPin` handled on one side only** (`rewrite/expr/convert.rs` ~L716 vs `rewrite/expr/unlower.rs` ~L586). `apply_adjustment` gained a `ReborrowPin` arm, but `peel_adjustments` has no matching case and falls through to the "unsupported adjustment" warning, so the new arm is unreachable outside its unit test. Cosmetic, since pinned reborrows don't occur in C-derived code.

ahomescu:
A few more from a second pass, again all low severity:

5. **Hard-coded toolchain pin in a transpile test** (`c2rust-transpile/tests/toolchain_pointer_casts.rs` L100). The test passes the literal `"+nightly-2025-01-01"` to `compile_and_run`, duplicating `rust-toolchain.toml`. The next roll will not find it by grepping for the toolchain file, and CI then either fails with "toolchain not installed" or keeps validating generated code against the old compiler. Passing no `+toolchain` so bare `rustc` resolves to the workspace pin, as the c2rust-refactor tests do, avoids that. Separately, L82 shells out to a `clang` driver binary, which no other transpile test needs (they only use libclang).

6. **`dynamic_instrumentation`: `TerminatorKind::TailCall` is not instrumented** (`dynamic_instrumentation/src/instrument.rs` ~L791). `visit_terminator` handles `Call` and `Return` and lets `TailCall` fall into the catch-all, while c2rust-analyze now treats tail calls as calls. A `become f(p)` gets no arg or return pointer events, and a `main` ending in a tail call never gets its `finalize` call. Only reachable with the unstable `explicit_tail_calls` feature, so this is cosmetic for C-derived code.

7. **Duplicated typing-environment construction** (`c2rust-refactor/src/illtyped.rs` L25, `transform/retype.rs` L1300, `transform/funcs.rs` L404). `illtyped.rs` hand-builds the same value the PR's new `context::empty_typing_env()` helper returns and that every other call site already uses. The other two literally expand rustc's own `TypingEnv::non_body_analysis(tcx, def_id)` constructor, which exists on the pinned nightly.

8. **`reflect.rs` L120 round-trips safety through a string**: `mk().unsafety(fn_sig.safety.to_string().as_str())` formats `hir::Safety` with `Display` only so the builder can parse the word back. A direct match on the two variants avoids depending on rustc's formatting.

Two more surfaced that I did not verify and would not post without a closer look: the print rewrite strategy now deep-clones every node twice and runs two extra walks per print, which may be noticeable on large crates, and the per-command rebuild of the node-to-def map sorts the entries only to drop them into an unordered hash map.

c2rust-transpile/tests/toolchain_pointer_casts.rs
thedataking: This also needs to catch incompatibilities between the generated-code and workspace compilers. The regular translator runner selects only one compiler per edition, and a snapshot would only catch changes to the emitted text.

c2rust-macros/src/lib.rs
thedataking: Yes: `noop_visit_expr(node, visitor)` became `walk_expr(visitor, node)`, with similar renames for the flat-map defaults. They still walk the children without calling the override for the current node again.

c2rust-refactor/src/transform/casts.rs
thedataking: The 16-bit bound is still there when introducing a pointer-sized literal. Here we’re evaluating an existing literal: truncating `70000isize` to `i16` gives 4464 even on a 32- or 64-bit target. Using the target’s pointer width gives the actual value while retaining 16-bit behavior on a 16-bit target.

c2rust-refactor/src/ast_manip/fold/walk.rs
thedataking: These defaults became private, but `WalkAst` still needs to walk children without calling the current node’s override again. Keeping the upstream implementations preserves that behavior; public walkers are still used directly.

c2rust-refactor/src/ast_manip/fold/walk.rs
thedataking: Some of it, yes. But `ast.txt` doesn’t encode callback order, visit versus flat-map callbacks, or the `VISIT_TOKENS` gate. Generating these faithfully would need extra schema annotations; I’d keep the upstream defaults for now.

c2rust-refactor/src/ast_manip/lr_expr.rs
thedataking: This implementation turned out to be unused: the generated expression walker skips the shared byte payload. Removed it.

c2rust-refactor/src/ast_manip/number_nodes.rs
thedataking: Yes. `into_trees()` is gone and `iter()` only gives borrowed trees; the backing storage is private. We need owned trees to rebuild the stream, and a separate nonterminal to reset IDs without changing other owners’ fragments.

c2rust-refactor/src/ast_manip/util.rs
thedataking: `LitKind::Err` now requires an emitted diagnostic, so it can’t double as our `$name:lit` marker. The suffix identifies an internal matcher placeholder; its `$` prevents collisions with ordinary Rust literal suffixes.

c2rust-refactor/src/ast_builder/builder.rs
thedataking: Yes, fixed. The function-pointer reflection caller now handles default safety explicitly, since an ordinary safe function pointer is spelled `fn`, without a `safe` qualifier.

c2rust-refactor/src/ast_builder/builder.rs
thedataking: `Err` now requires an emitted diagnostic. This helper just leaves a temporary placeholder while moving an expression, so `Dummy` fits: it carries no diagnostic and must be replaced before lowering.

c2rust-refactor/src/ast_manip/print.rs
thedataking: This is new C2Rust code around rustc’s existing printer. It prepares copies of the transformed fragments so the printer uses their current AST instead of stale cached tokens.

thedataking:
Fixed the extern-block ordering and safety deduplication, MIR checking during `cargo check`, and the missing pinned-reborrow mapping.

For the second comment: the pointer test now follows the workspace pin and no longer invokes `clang`; the typing environments use the existing helpers. Explicit tail calls now fail with a clear unsupported-feature error, since partial tracing would still miss return transfers or finalization. The safety conversion was already fixed in the previous push.

Also removed the duplicate printer clone and skipped the extra binder walk when unnecessary. The map sort stays: rustc’s unordered-iteration API requires ordering before collecting into our hash map.

ahomescu:
A general nit: the existing code coalesces imports, but many of the new imports are one-per-line. Could we stay consistent?

c2rust-refactor/src/interact/main_thread.rs
```diff
@@ -279,9 +300,31 @@ pub fn interact_command(args: &[String], config: Config, registry: command::Regi
...
-        InteractState::new(state, buffers_available, to_worker, to_client).run_loop(main_recv);
-    });
+    let mut registry = registry;
+    let mut main_recv = main_recv;
+    let mut pending = None;
+    loop {
+        let buffers = buffers_available.clone();
+        let worker = to_worker.clone();
+        let client = to_client.clone();
+        let (next_registry, receiver, resume) = driver::run_refactoring(
```
ahomescu, line 310: Is this faster or slower than the old approach?

c2rust-refactor/src/interact/main_thread.rs
```diff
@@ -190,7 +212,6 @@ impl InteractState {
 
             RunCommand { name, args } => {
                 info!("running command {} with args {:?}", name, args);
-                self.state.load_crate();
```
ahomescu, line 193: This is a significant change, are we preserving the crate across runs now?

c2rust-refactor/src/command.rs
```diff
@@ -296,266 +302,154 @@ impl RefactorState {
...
-            cs.phase = phase;
+        let disk_state = disk_state.get_or_insert_with(|| {
+            let mut krate = rustc_interface::passes::parse(session);
+            load_modules(&mut krate, &session.psess, source_map);
+            remove_paren(&mut krate);
+            number_nodes(&mut krate);
+            DiskState::new(krate, source_map, session, node_map)
+        });
+
+        let mut need_load = true;
```
ahomescu, line 325: There was a comment here, why is it gone?

c2rust-refactor/src/command.rs
```diff
@@ -296,266 +302,154 @@ impl RefactorState {
...
+        let mut cs = CommandState::new(
+            krate.take().unwrap_or_else(|| {
+                need_load = false;
```
ahomescu, line 328: Same here: where did the comment go?

c2rust-refactor/src/command.rs
```diff
@@ -296,266 +302,154 @@ impl RefactorState {
...
-                    cs.krate.replace(expanded);
-                    max_crate_node_id = Some(next_node_id);
-                    remove_paren(cs.krate.get_mut());
-                }
-            }
-
-            cs.phase = phase;
+        let disk_state = disk_state.get_or_insert_with(|| {
+            let mut krate = rustc_interface::passes::parse(session);
+            load_modules(&mut krate, &session.psess, source_map);
```
ahomescu, line 319: Some comments missing here.

c2rust-refactor/src/command.rs
```diff
@@ -296,266 +302,154 @@ impl RefactorState {
...
+            krate.take().unwrap_or_else(|| {
+                need_load = false;
+                disk_state.orig_krate.clone()
+            }),
+            Phase::Phase1,
+            marks.clone(),
+            ParsedNodes::default(),
+            node_id_counter.clone(),
+        );
+        if need_load {
```
ahomescu, line 336: Comment?

c2rust-refactor/src/command.rs
```diff
@@ -296,266 +302,154 @@ impl RefactorState {
...
+            load_modules(&mut *cs.krate.borrow_mut(), &session.psess, source_map);
+        }
+        let unexpanded = cs.krate().clone();
+        if phase != Phase::Phase1 {
+            reset_node_ids(&mut *cs.krate.borrow_mut());
```
ahomescu, line 341: Comment?

c2rust-refactor/src/driver.rs
```diff
@@ -187,94 +194,59 @@ where
...
 {
-    // Force disable incremental compilation.  It causes panics with multiple typechecking.
-    config.opts.incremental = None;
-
-    rustc_span::create_session_globals_then(Edition::Edition2021, move || {
-        let state = RefactorState::new(config, cmd_reg, file_io, marks);
-        f(state)
+    let compiler_config = clone_config(&config);
+    let loader = Some(Box::new(ArcFileIO(file_io.clone())) as Box<dyn FileLoader + Send + Sync>);
+    run_compiler(compiler_config, loader, move |compiler| {
```
ahomescu, line 219: We were intentionally trying to avoid this because it uses `run_in_thread_pool_with_globals` and does a bunch of extra stuff we don't want. We really just want the code at https://github.com/rust-lang/rust/blob/main/compiler/rustc_interface/src/util.rs#L313.

ahomescu:
There are _a lot!_ of reformatting changes, maybe that should be split into a separate PR.

c2rust-analyze/src/borrowck/type_check.rs
```diff
@@ -467,7 +471,8 @@ impl<'tcx> TypeChecker<'tcx, '_> {
                 _ => panic!("unsupported rvalue AggregateKind {:?}", kind),
             },
 
-            Rvalue::Len(..) => {
+            Rvalue::UnaryOp(rustc_middle::mir::UnOp::PtrMetadata, ref op) => {
```
ahomescu, line 474: Why does `Len` => `UnaryOp` keep the same implementation?

c2rust-refactor/src/transform/casts/tests.rs
```diff
@@ -1,9 +1,44 @@
-use super::{check_double_cast, DoubleCastAction, SimpleTy};
+use super::{check_double_cast, ConstantValue, DoubleCastAction, SimpleTy};
 use quickcheck::{quickcheck, Arbitrary, Gen};
 use rand::Rng;
 use z3::ast::{Ast, BV};
 use z3::{Config, Context, SatResult, Solver};
 
+#[test]
```
ahomescu, line 7: Nit: test should probably go at the end of the file.

c2rust-refactor/src/transform/casts.rs
ahomescu: This replaces the `as u16 as u128` with a direct cast to the target pointer size. How is it the same?

c2rust-refactor/tests/toolchain_casts.rs
```diff
@@ -0,0 +1,35 @@
...
+"#
+    .replace("MINIMUM", &isize::MIN.to_string());
+    for edition in ["2021", "2024"] {
+        let fixture = Fixture::new();
+        let path = fixture.write("casts.rs", &source);
+        let before = compile_and_run(&path, edition);
+        assert_eq!(before, format!("70000 70000 -70000 0 {}\n", isize::MIN));
+        refactor(&path, edition, vec![command("remove_redundant_casts", &[])]);
+        let after = fs::read_to_string(&path).unwrap();
+        assert!(after.contains("let unsigned = 70000u64;"), "{after}");
```
ahomescu, line 28: This is absolutely not a correct rewrite from `70000usize as u64`, the old logic was intentionally what we wanted. On some platforms `usize` might be `u16`, and we need the refactoring to be correct there as well. We absolutely _should not_ be using the target pointer width here.

c2rust-refactor/src/transform/casts.rs
```diff
@@ -431,8 +431,14 @@ enum ConstantValue {
 }
 
 impl ConstantValue {
-    fn cast(self, ty: SimpleTy) -> Self {
+    fn cast(self, ty: SimpleTy, pointer_bits: u64) -> Self {
```
ahomescu, line 434: We _absolutely should not_ be passing this in, this whole commit is wrong.

dynamic_instrumentation/src/callbacks.rs
```diff
@@ -15,11 +15,9 @@ impl rustc_driver::Callbacks for MirTransformCallbacks {
         config.opts.incremental = None;
         config.override_queries = Some(override_queries);
 
-        // The compiler's MIR validator is now private. Preserve mandatory
-        // validation via the pass manager: this checks all bodies downstream
-        // after compiler passes, rather than just our body immediately here.
-        // It is broader and potentially slower, but must remain enabled even
-        // when the caller disables the compiler's optional validation.
+        // The compiler's MIR validator is private. Enable per-pass validation
```
ahomescu, line 18: Why did this comment change?

PR #2001, opened 2026-09-08
transpile: Fix translation failures exposed by SQLite

thedataking:
Fix five translation failures exposed by SQLite's individual source files:

- Register the main file before emitting the exporter's file table. Comment-only translation units otherwise introduce an unknown file ID and panic during include-path lookup.
- Construct integer-derived opaque pointers through `c_void` pointers. Exposed-provenance constructors require sized pointees, while opaque extern types are unsized.
- Specify the destination type when reading function pointers with `va_arg`. Enclosing casts can prevent Rust from inferring it.
- Lower atomic function pointer operations through raw pointers in both Clang atomic expressions and legacy `__sync_*` builtins. Rust's atomic intrinsics reject the `Option<fn>` representation of nullable C callbacks.
- Initialize `_Atomic` storage through its underlying value type. Default initialization otherwise fails for local atomic declarations, including explicitly initialized ones.

### DIFF
--- a/c2rust-transpile/src/translator/atomics.rs
+++ b/c2rust-transpile/src/translator/atomics.rs
@@ -34,7 +34,67 @@ pub(crate) fn order_ty_name(order: Ordering) -> &'static str {
     }
 }
 
+// Rust represents nullable C function pointers as Option<fn>, which atomic
+// intrinsics reject. Both AtomicExpr and legacy builtin calls must operate on
+// the raw pointer representation, restoring the function pointer on results.
+pub(crate) struct AtomicValue {
+    function_pointer: Option<Box<Type>>,
+}
+
+impl AtomicValue {
+    fn raw_pointer_type() -> Box<Type> {
+        mk().mutbl().ptr_ty(mk().tuple_ty(vec![]))
+    }
+
+    pub(crate) fn storage(&self, ptr: Box<Expr>) -> Box<Expr> {
+        if self.function_pointer.is_some() {
+            mk().cast_expr(ptr, mk().mutbl().ptr_ty(Self::raw_pointer_type()))
+        } else {
+            ptr
+        }
+    }
+
+    pub(crate) fn lower(&self, val: Box<Expr>) -> Box<Expr> {
+        match &self.function_pointer {
+            Some(ty) => transmute_expr(ty.clone(), Self::raw_pointer_type(), val),
+            None => val,
+        }
+    }
+
+    pub(crate) fn restore(&self, val: Box<Expr>) -> Box<Expr> {
+        match &self.function_pointer {
+            Some(ty) => transmute_expr(Self::raw_pointer_type(), ty.clone(), val),
+            None => val,
+        }
+    }
+
+    pub(crate) fn zero(&self) -> Box<Expr> {
+        if self.function_pointer.is_some() {
+            mk().call_expr(mk().abs_path_expr(vec!["core", "ptr", "null_mut"]), vec![])
+        } else {
+            mk().lit_expr(mk().int_lit(0, ""))
+        }
+    }
+}
+
 impl<'c> Translation<'c> {
+    pub(crate) fn atomic_value(&self, ptr_id: CExprId) -> TranslationResult<AtomicValue> {
+        let ptr_ty = self.ast_context[ptr_id].kind.get_qual_type().unwrap();
+        let mut value_ty = self
+            .ast_context
+            .get_pointee_qual_type(ptr_ty.ctype)
+            .unwrap();
+        if let CTypeKind::Atomic(inner) = self.ast_context.resolve_type(value_ty.ctype).kind {
+            value_ty = inner;
+        }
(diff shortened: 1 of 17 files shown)


c2rust-transpile/tests/snapshots/snapshots__transpile@opaque_pointer_casts.c.2024.clang15.snap
```diff
@@ -0,0 +1,44 @@
...
+    unused_mut
+)]
+#![feature(extern_types)]
+unsafe extern "C" {
+    pub type Opaque;
+    pub type OpaqueUnion;
+}
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sentinel() -> *mut Opaque {
+    return ::core::ptr::with_exposed_provenance_mut::<u8>(8 as ::core::ffi::c_int as usize)
```
ahomescu: If there isn't a matching call to `expose_provenance`, this is undefined behavior.

c2rust-transpile/src/translator/pointers.rs
```diff
@@ -593,14 +593,31 @@ impl<'c> Translation<'c> {
...
             };
-            let pointee_type_rs = self.convert_pointee_type(pointee_type_id.ctype)?;
+            // Extern types are unsized, but the exposed-provenance constructors
+            // require Sized pointees. Construct a byte pointer first and cast
+            // it to the opaque pointer type, which is still a thin pointer.
+            let is_opaque = self
+                .ast_context
+                .is_forward_declared_type(pointee_type_id.ctype);
+            let pointee_type_rs = if is_opaque {
+                mk().path_ty(vec!["u8"])
```
ahomescu: I'm not sure yet if this should be `u8` or `c_void`.

c2rust-transpile/src/translator/variadic.rs
```diff
@@ -250,8 +250,10 @@ impl<'c> Translation<'c> {
                         VaArgCastKind::Enum(enum_id) => {
                             self.enum_constructor_expr(enum_id, val, false)
                         }
-                        VaArgCastKind::Transmute => {
-                            transmute_expr(mk().infer_ty(), mk().infer_ty(), val)
+                        VaArgCastKind::Transmute(ty) => {
+                            // An enclosing cast or field access may prevent
+                            // inference of the transmute's destination type.
```
ahomescu, line 255: Could either this comment or the current commit message provide an example of the generated code?

c2rust-transpile/src/translator/pointers.rs
ahomescu: We use `c_void` in other places, so yeah we should keep it consistent.

c2rust-transpile/src/translator/atomics.rs
```diff
@@ -155,6 +155,37 @@ impl<'c> Translation<'c> {
             .transpose()?;
         let weak = weak_id.and_then(|x| self.convert_constant_bool(x));
 
+        // Rust represents nullable C function pointers as Option<fn>, which
```
ahomescu: This only covers the AST atomics, but not any of the `__sync_*` or `__atomic_*` builtins. You could refactor this to cover everything.

c2rust-transpile/tests/snapshots/snapshots__transpile@opaque_pointer_casts.c.2024.clang15.snap
thedataking: Does this apply when the pointer is only used as a sentinel? My agent claims that constructing it is
allowed; only accessing memory through it would require provenance.

thedataking:
Updated to use `c_void` and added a generated-code example explaining the varargs type-inference fix. The atomic
conversion is now shared with the legacy `__sync_*` paths, with coverage for exchange, compare-and-swap, and
release, including null callbacks.

c2rust-transpile/src/translator/variadic.rs
thedataking: Added a generated-code example explaining the existing fix.

c2rust-transpile/tests/snapshots/snapshots__transpile@opaque_pointer_casts.c.2024.clang15.snap
ahomescu: I'm not sure, check the Rust docs?

ahomescu (approved):

c2rust-transpile/tests/snapshots/snapshots__transpile@opaque_pointer_casts.c.2024.clang15.snap
thedataking: Checked: [Rust permits sentinel pointers without provenance](https://doc.rust-lang.org/std/ptr/index.html#provenance):

> Provenance can be absent, in which case the pointer does not have permission to access any memory.

and [`with_exposed_provenance` is equivalent to an integer-to-pointer cast](https://doc.rust-lang.org/std/ptr/fn.with_exposed_provenance.html).

PR #2002, opened 2026-09-08
tests: Add SQLite integration coverage from canonical sources

thedataking:
Add SQLite 3.53.4 to the integration harness and both Clang CI configurations. Build the official source submodule with `--disable-amalgamation` to exercise translation across individual source files. Generate sources before recording the compilation database to exclude host build tools.

Compare result counts and hashes for five upstream speedtest workloads (`main`, `cte`, `orm`, `fp`, and `rtree`) against native and translated libraries before and after refactoring. Always-active SQL assertions cover transactions, triggers, foreign keys, blobs, JSON/JSONB, large integers and floating-point parsing, window functions, integrity, and data and WAL journal-mode persistence after reopening. These explicit JSON and numeric checks replace upstream workloads that do not hash results or reliably report SQL errors. Both libraries use the same native C callers to exercise their public C ABI.

Use `jq` to read Cargo's configured output directory and rustc to determine the Rust archive's native link dependencies.

Depends on the transpiler fixes in #2001.

### DIFF
--- a/tests/integration/tests/sqlite/README.md
+++ b/tests/integration/tests/sqlite/README.md
@@ -1,0 +1,23 @@
+# SQLite
+
+The source submodule is pinned to SQLite 3.53.4 from the official Git mirror.
+Initialize it with:
+
+```sh
+git submodule update --init --checkout tests/integration/tests/sqlite/repo
+```
+
+Run from the C2Rust checkout with the integration harness prerequisites installed:
+
+```sh
+export PATH="$PWD/target/release:$PWD/c2rust-postprocess:$PATH"
+export C2RUST_DIR="$PWD"
+./tests/integration/test.py sqlite
+```
+
+The build uses `--disable-amalgamation` and the `libsqlite3.a` target to compile
+the library's individual translation units. SQLite's `sqlite3` and `speedtest1`
+make targets compile the amalgamation even with that configure option, so they
+are deliberately not used. Source generation runs before Bear to exclude build
+tools from the compilation database. R-tree support is enabled for its workload.
+
--- a/tests/integration/tests/sqlite/README.md
+++ b/tests/integration/tests/sqlite/README.md
@@ -7,7 +7,8 @@ Initialize it with:
 git submodule update --init --checkout tests/integration/tests/sqlite/repo
 ```
 
-Run from the C2Rust checkout with the integration harness prerequisites installed:
+Run from the C2Rust checkout with the integration harness prerequisites and `jq`
+installed:
 
 ```sh
 export PATH="$PWD/target/release:$PWD/c2rust-postprocess:$PATH"
@@ -21,3 +22,19 @@ make targets compile the amalgamation even with that configure option, so they
 are deliberately not used. Source generation runs before Bear to exclude build
 tools from the compilation database. R-tree support is enabled for its workload.
 
+After transpilation and again after refactoring, `test.sh` compiles SQLite's
+upstream `test/speedtest1.c` against each of the native and translated static
+libraries. It runs the `main`, `cte`, `orm`, `fp`, and `rtree` workloads with
+`--size 1 --verify` on temporary on-disk databases.
+Each workload must exit successfully and produce the same result byte count and
+verification hash as native SQLite. Timing output is ignored. The C test driver
+stays native so the same caller exercises both libraries' public C ABI.
+The Rust archive's native link dependencies come from rustc's
+`--print native-static-libs` output.
+
+`smoke.c` also checks fixed SQL results, transaction/savepoint rollback, triggers,
+foreign keys, blobs, JSON/JSONB operations, numeric parsing (including large
+integers and floating-point values), window functions, data and WAL journal-mode
+persistence after reopening, and an integrity check. The upstream `json` and
+`parsenumber` workloads do not hash result rows, and `parsenumber` ignores SQL
+errors, so JSON and numeric parsing use these explicit assertions instead.
--- a/tests/integration/tests/sqlite/conf.yml
+++ b/tests/integration/tests/sqlite/conf.yml
@@ -1,0 +1,6 @@
+transpile:
+  autogen: true
+  tflags: --reorganize-definitions --disable-refactoring
+
+cargo.transpile:
+  autogen: true
--- a/tests/integration/tests/sqlite/conf.yml
+++ b/tests/integration/tests/sqlite/conf.yml
@@ -1,3 +1,13 @@
+requirements:
+  generic:
+    programs:
+      in_path:
+        - jq
+  ubuntu:
+    apt:
+      packages:
+        - jq
+
 transpile:
   autogen: true
   tflags: --reorganize-definitions --disable-refactoring
--- a/tests/integration/tests/sqlite/conf.yml
+++ b/tests/integration/tests/sqlite/conf.yml
@@ -4,3 +4,16 @@ transpile:
 
 cargo.transpile:
   autogen: true
+
+refactor:
+  autogen: true
+  transforms:
+    - rename_unnamed
+    - reorganize_definitions
+    - shorten_imported_paths
+    - remove_unused_labels
+    - remove_redundant_casts
+    - remove_literal_suffixes
+
+cargo.refactor:
+  autogen: true
--- a/tests/integration/tests/sqlite/configure.sh
+++ b/tests/integration/tests/sqlite/configure.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+cd "$SCRIPT_DIR/repo"
+
+# Compile separate translation units to exercise cross-file translation.
+./configure --disable-amalgamation --disable-shared --disable-readline \
+    --disable-tcl --enable-rtree 2>&1 | tee "$SCRIPT_DIR/$(basename "$0").log"
--- a/tests/integration/tests/sqlite/smoke.c
+++ b/tests/integration/tests/sqlite/smoke.c
@@ -1,0 +1,91 @@
+#include "sqlite3.h"
+#include <stdio.h>
+#include <stdlib.h>
+
+#define CHECK(condition) do { \
+    if (!(condition)) { \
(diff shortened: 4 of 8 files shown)


tests/integration/tests/sqlite/test.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+cd "$SCRIPT_DIR/repo"
+
+{
+    TARGET_DIR=$(cargo ${TOOLCHAIN:+"$TOOLCHAIN"} metadata --no-deps --format-version 1 \
+        | python3 -c 'import json, sys; print(json.load(sys.stdin)["target_directory"])')
```
ahomescu: `jq` might actually be shorter here.

tests/integration/tests/sqlite/configure.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+cd "$SCRIPT_DIR/repo"
+
+./configure --disable-amalgamation --disable-shared --disable-readline \
```
ahomescu, line 8: I saw it in the README, but it might be useful to also explain` --disable-amalgamation` here.

ahomescu (approved):

tests/integration/tests/sqlite/test.sh
```diff
@@ -0,0 +1,38 @@
...
+    done
+
+    # Compile the same upstream workload against the native and Rust libraries.
+    # Passing the archives explicitly prevents picking up a system SQLite.
+    cc -O2 -DSQLITE_ENABLE_RTREE -I. test/speedtest1.c libsqlite3.a -lm -ldl -lpthread \
+        -o "$WORK_DIR/native"
+    cc -O2 -DSQLITE_ENABLE_RTREE -I. test/speedtest1.c "$TARGET_DIR/release/librepo.a" \
+        -lm -ldl -lpthread -o "$WORK_DIR/rust"
+
+    for testset in main cte orm fp json rtree parsenumber; do
```
ahomescu: **`json` and `parsenumber` testsets compare a constant hash.** Neither testset calls `speedtest1_run`, so both binaries always print the same trivial `Verification Hash` and the `diff` cannot fail. `parsenumber` additionally discards every `sqlite3_exec` return code (speedtest1.c:2890-2908). A translated library that errors on or mis-evaluates every `SELECT` in `parsenumber`, or corrupts JSON output in `json`, still passes this stage. Only `main`/`cte`/`orm`/`fp`/`rtree` hash rows. Either drop those two testsets or add explicit checks for them to `smoke.c`, and reword the README sentence about every workload's hash being compared.

tests/integration/tests/sqlite/conf.yml
```diff
@@ -0,0 +1,19 @@
+transpile:
+  autogen: true
+  tflags: --reorganize-definitions --disable-refactoring --fail-on-error
```
ahomescu: **`--fail-on-error` is unique to this project.** No other conf.yml passes it. With it, any `Skipping declaration ... due to error` (`translator/mod.rs:716` exits 1) or invalid exported Clang AST (`lib.rs:707` panics) aborts the whole transpile stage instead of stubbing and continuing. That class of warning is live on master today: libgit2's `transpile.gen.sh.log` shows `Exported Clang AST was invalid`. If one of the ~100 SQLite TUs hits an unexported node under either CI clang (15 or 18), sqlite fails hard while every other project degrades. Defensible as a regression guard for the #2001 fixes, but worth stating as a deliberate choice.

tests/integration/tests/sqlite/smoke.c
```diff
@@ -0,0 +1,67 @@
...
+    if (sqlite3_exec(db, sql, NULL, NULL, &error) != SQLITE_OK) {
+        fprintf(stderr, "%s: %s\n", sql, error);
+        sqlite3_free(error);
+        exit(1);
+    }
+}
+
+static void expect_int(sqlite3 *db, const char *sql, int expected) {
+    sqlite3_stmt *stmt = NULL;
+    assert(sqlite3_prepare_v2(db, sql, -1, &stmt, NULL) == SQLITE_OK);
```
ahomescu: **Side-effecting calls inside `assert()`.** `sqlite3_open`, `sqlite3_prepare_v2`, `sqlite3_step`, `sqlite3_finalize`, `sqlite3_close` and the FK-violation `sqlite3_exec` are evaluated only when `NDEBUG` is undefined, while the rest of the file uses an explicit `exit(1)` path. Anyone compiling with `-DNDEBUG` gets `db == NULL` passed to `sqlite3_exec` (misleading `SQLITE_MISUSE`), and the FK-constraint and finalized-statements checks silently disappear. Latent today since test.sh uses plain `cc -O2`, but a `CHECK` macro that cannot be compiled out would give the file one failure path.

tests/integration/tests/sqlite/test.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+cd "$SCRIPT_DIR/repo"
+
+{
+    TARGET_DIR=$(cargo ${TOOLCHAIN:+"$TOOLCHAIN"} metadata --no-deps --format-version 1 \
```
ahomescu, line 8: **`cargo metadata` + inline `python3` to find the target dir.** This adds an undeclared `python3` runtime dependency (the harness itself runs under `uv run --script`, which does not guarantee one on `PATH`) and a third convention for a lookup that `json-c/test.sh` and `python2/test.sh` do with `${CARGO_TARGET_DIR:-$SCRIPT_DIR/repo/target}/release/librepo.a` plus an explicit "Rust archive not found" check. Suggest the json-c expression, or at least `jq -r .target_directory`.

tests/integration/tests/sqlite/test.sh
```diff
@@ -0,0 +1,38 @@
...
+        | python3 -c 'import json, sys; print(json.load(sys.stdin)["target_directory"])')
+    WORK_DIR=$(mktemp -d)
+    trap 'rm -rf "$WORK_DIR"' EXIT
+
+    for archive in "$SCRIPT_DIR/repo/libsqlite3.a" "$TARGET_DIR/release/librepo.a"; do
```
ahomescu: **Four near-identical `cc` link lines.** `smoke.c` is built in a loop over the two archives, but the two `speedtest1.c` compiles are written out by hand, and the later loop re-derives the native/rust naming. The flags appear four times, and `libsqlite3.a` is referenced both relatively (line 22) and as `$SCRIPT_DIR/repo/libsqlite3.a` (here). A single `for impl in native rust` loop that picks the archive and builds both `smoke-$impl` and `$impl` covers everything, and per-implementation db names remove the `rm -f smoke.db*` cleanup.

tests/integration/tests/sqlite/README.md
```diff
@@ -0,0 +1,35 @@
...
+After transpilation and again after refactoring, `test.sh` compiles SQLite's
+upstream `test/speedtest1.c` against each of the native and translated static
+libraries. It runs the `main`, `cte`, `orm`, `fp`, `json`, `rtree`, and
+`parsenumber` workloads with `--size 1 --verify` on temporary on-disk databases.
+Each workload must exit successfully and produce the same result byte count and
+verification hash as native SQLite. Timing output is ignored. The C test driver
+stays native so the same caller exercises both libraries' public C ABI.
+
+`smoke.c` also checks fixed SQL results, transaction/savepoint rollback, triggers,
+foreign keys, blobs, JSON, numeric parsing, window functions, WAL persistence
```
ahomescu: **WAL persistence is claimed but not checked.** After reopening, `smoke.c` checks the `audit` row count and `integrity_check` but never `PRAGMA journal_mode`, so a translated pager that silently falls back to rollback-journal mode still passes. Either add `expect_int(db, "SELECT journal_mode = 'wal' FROM pragma_journal_mode", 1)` after the reopen, or soften this line.

tests/integration/tests/sqlite/test.sh
```diff
@@ -0,0 +1,38 @@
...
+        cc -O2 -I. "$SCRIPT_DIR/smoke.c" "$archive" -lm -ldl -lpthread \
```
ahomescu: **Hardcoded `-lm -ldl -lpthread` (low confidence).** `json-c/test.sh` derives the list from `RUSTFLAGS="--print native-static-libs"`; rustc reports `-lgcc_s -lutil -lrt -lpthread -lm -ldl -lc`. This links on CI because glibc >= 2.34 merged the rest and `cc` adds `libgcc_s` implicitly, but an older host or a future native dependency of the transpiled crate would fail with undefined symbols and no hint why.

ahomescu (commented):
Review findings (7), posted inline. The two that matter most: the `json` and `parsenumber` testsets compare a constant hash, and `--fail-on-error` is unique to this conf.yml.

tests/integration/tests/sqlite/conf.yml
thedataking: Removed. Shouldn't diverge from other tests here.

tests/integration/tests/sqlite/configure.sh
thedataking: Added a comment immediately above configure explaining that separate translation units exercise cross-file  translation.

tests/integration/tests/sqlite/test.sh
thedataking: Dropped those two workloads and expanded the explicit JSON/JSONB and numeric checks in smoke.c. Only the five workloads that hash result rows remain in the comparison loop.

tests/integration/tests/sqlite/test.sh
thedataking: Rust link dependencies now come from rustc’s native-static-libs output. The fixed flags apply only to the
native SQLite archive.

tests/integration/tests/sqlite/test.sh
thedataking: Switched to jq and declared the dependency. Kept cargo metadata so configured target directories work,
and added an explicit missing-archive check.

PR #2005, opened 2026-09-09
transpile: Do not consider `Bool` an integral type

Rua:
`CTypeKind::is_unsigned_integral_type` returns `true` for `Bool`, which is a bit of a footgun and probably not what most code (or coders) expect. So this removes it. I made sure to check that no code elsewhere relies on this including `Bool` and adjusted it where necessary. Most of the time, it was being used for wrapping arithmetic, which isn't possible on booleans anyway.

### DIFF
--- a/c2rust-transpile/src/c_ast/mod.rs
+++ b/c2rust-transpile/src/c_ast/mod.rs
@@ -2173,7 +2173,9 @@ impl CastKind {
 
             (CTypeKind::Function(..), CTypeKind::Pointer(..)) => CastKind::FunctionToPointerDecay,
 
-            (_, CTypeKind::Pointer(..)) if source_ty_kind.is_enum_or_integral_type() => {
+            (_, CTypeKind::Pointer(..))
+                if source_ty_kind.is_enum_or_integral_type() || source_ty_kind.is_bool() =>
+            {
                 CastKind::IntegralToPointer
             }
 
@@ -3224,7 +3226,7 @@ impl CTypeKind {
         use CTypeKind::*;
         matches!(
             self,
-            Bool | UChar
+            UChar
                 | UInt
                 | UShort
                 | ULong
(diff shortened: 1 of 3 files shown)


c2rust-transpile/src/c_ast/mod.rs
ahomescu, line 3221: Do we not want this to return `true` for `bool`?

ahomescu, line 3217: Do we not want `true` here for `bool`? Does it regress anything?

Rua: It's only used in two places, both of which explicitly account for `bool` separately anyway: `implicit_default_expr` and `literal_matches_ty`.

Rua: This function is only called in `CastKind::from_types`, where `Bool` is also handled separately. Possibly the `IntegralToPointer` case is not, so I've added handling for `bool` explicitly there now.

ahomescu: You might be introducing some regressions in `from_types`, can you test bool-to-unsigned casts and maybe bool-to-float? (I'm not sure how much we care about the latter, but it might still be worth to see what happens)

Rua: I'm trying to think of a situation where `from_types` is actually called and hitting these cases. It's not called for regular casts in the AST, which are passed on from C.

Rua: #2014 will remove the need for this to be correct in any case, because it selects based on the actual types instead of the generated `CastKind`.

ahomescu: What happens for this code?
```
double f(bool b, unsigned u) { b *= 2.0; b += u; return b; }
```

Rua: Ok, I see the problem now. Fixed.

ahomescu (approved):

PR #2010, opened 2026-09-12
transpile: Don't require `WithStmts` for `make_cast` and other cast functions

Rua:
The only thing here that actually required a `WithStmts` specifically was an assert for the `NullToPointer` case. That case was throwing out the existing `WithStmts` and replacing it with a new one, resulting in a double translation. `NullToPointer` is never returned by `CastKind::from_types`, and so that path is only ever hit when being called from `convert_cast`. So I just moved `NullToPointer` there instead.


c2rust-transpile/src/translator/mod.rs
```diff
@@ -4078,7 +4082,13 @@ impl<'c> Translation<'c> {
...
 
                 // if the context wants a different type, add a cast
-                return self.make_cast(ctx, source_ty.not_volatile(), target_ty, val);
+                return val.and_then_try(|val| {
+                    self.make_cast(ctx, source_ty.not_volatile(), target_ty, val)
+                });
+            }
+
+            CastKind::NullToPointer => {
+                return Ok(WithStmts::new_val(self.null_ptr(ty.ctype)?));
```
ahomescu: Should we use `target_ty.ctype` here instead?

Rua: Yes, fixed.

PR #2019, opened 2026-09-21
transpile: Remove `ConstCast`

Rua:
`ConstCast` was a "fake" cast kind, created within the C++ exporter. But there's no particular need for it; it can be implemented in Rust as well, using the more accurate (and less unsafe) tools available there. So this replaces it with a new function that does the same thing.


c2rust-transpile/src/c_ast/mod.rs
```diff
@@ -518,6 +518,48 @@ impl TypedAstContext {
         }
     }
 
+    /// Returns whether a cast from `source_type_id` to `target_type_id` is a cast from a `mut`
+    /// pointer to a `const` pointer, with the same pointee type and otherwise the same qualifiers.
+    pub(crate) fn is_mut_to_const_pointer_cast(
```
ahomescu, line 523: The old code only caught implicit casts, which was working as intended. Does this also handle explicit ones now?

Rua: Right, `make_cast_full` doesn't know about explicitness of casts. I've added a parameter for that now, and gated the early return to `!is_explicit`.

